PR #4, opened 2015-12-03
Add @noescape to Synchronous Enumeration, Predicate, and Comparator Blocks

Adlai-Holler:


parkera:
This is an interesting case, because if we do this then the API is going to be different between Darwin and Swift, but we want to make sure that we stay consistent. 

However, I think `@noescape` is the right answer for these APIs, so we should figure out a way to do this. clang has an attribute to express what we want in C/Objective-C: `__attribute__((noescape))`.

Since we can't fix up the Darwin Foundation headers at the same time as we do this, the best approach is going to be to use something like [API notes](https://github.com/apple/swift/tree/master/apinotes) to fix up the Foundation SDK on Darwin so it matches. Do we have support in API notes for `@noescape` yet?

parkera:
Following up on this - we think this is a great idea. Since this pull request contains an API change, so let’s follow the [Swift evolution process](https://swift.org/contributing/#participating-in-the-swift-evolution-process), starting with discussing the idea on the `swift-evolution` and `swift-corelibs-dev` lists.

Adlai-Holler:
@parkera OK cool.

PR #5, opened 2015-12-03
Verify contents of the file attributes dictionary

bolshedvorsky:


parkera:
This one didn't compile for me on Ubuntu 14.04:

```
TestFoundation/TestNSFileManager.swift:121:13: error: cannot invoke 'XCTAssertEqual' with an argument list of type '(NSString, String)'
            XCTAssertEqual(fileType!, NSFileTypeRegular)
            ^
```

There are actually some issues with the NSFileManager API itself that I think we need to fix up in addition to adding this test (which is a good idea, btw).

```
public func attributesOfItemAtPath(path: String) throws -> [String : AnyObject]
```

0. The return type should be [String : Any] (and we need to mark it as experimental since it will then be slightly different than Darwin, although compatible)
0. The implementation should be creating String instead of NSString. It should still use NSNumber and NSDate though.

Would you be willing to take a look at these issues and provide a new patch?

bolshedvorsky:
Yes I will take a look at these issues and I will provide a new patch.

PR #7, opened 2015-12-03
Use Correct Source Index in NSArray.getObjects

Adlai-Holler:


phausler:
actually this probably should insert instead of just indexing in the first place as well;

And a test for verifying the behavior would be useful as well:

```swift
    func test_getObjects() {
        let array : NSArray = ["foo", "bar", "baz", "foo1", "bar2", "baz3",].bridge()
        var objects = [AnyObject]()
        array.getObjects(&objects, range: NSMakeRange(1, 3))
        XCTAssertEqual(objects.count, 3)
        let fetched = [
            (objects[0] as! NSString).bridge(),
            (objects[1] as! NSString).bridge(),
            (objects[2] as! NSString).bridge(),
        ]
        XCTAssertEqual(fetched, ["bar", "baz", "foo1"])
    }
```

Adlai-Holler:
OK I've reimplemented it to be a little quicker, and to insert starting at 0 rather than replace. I can't get the tests to build because when I use `xcrun launch-with-toolchain` I get `unable to find utility "launch-with-toolchain"`. Using `xcrun 27`. Confirmed the toolchain is installed and usable, so is Xcode 7.2b4. If anyone knows of a quick solution I'd like to be able to run tests ASAP. Otherwise I've gone as far as I can with this PR.

bartekchlebek:
You need Xcode 7.2 beta installed, and use
```sh
xcode-select -switch /Applications/Xcode-beta.app/Contents/Developer
```
to change xcrun's tools path.
Then it should work

Adlai-Holler:
Nice @bartekchlebek! Everything's up and running. I've changed the implementation to append since that's what you do with buffers. Test is passing and I think it's ready for review!

phausler:
Tests look good on all platforms. Overall the code looks good to me.

PR #8, opened 2015-12-03
[WIP] Implement NSRangeFromString

harlanhaskins:
This is an initial implementation of NSRangeFromString that attempts to replicate the behavior of standard Foundation.

I'm still in the middle of setting up Xcode 7.2 for this version of Foundation, so I can't test this right away. This'll be a WIP until I can.


harlanhaskins:
Well, it works fine using Obj-C's foundation, but it relies heavily on `componentsSeparatedByCharactersInString`.

harlanhaskins:
Closing because there are certainly better ways to do all of this.

PR #12, opened 2015-12-03
Several NSString constructors implemented

bartekchlebek:
namely
```swift
public convenience init?(data: NSData, encoding: UInt)
public convenience init?(bytes: UnsafePointer<Void>, length len: Int, encoding: UInt)
public convenience init?(CString nullTerminatedCString: UnsafePointer<Int8>, encoding: UInt)
```


parkera:
Looks great, and thanks for the tests.

PR #13, opened 2015-12-03
Change NSURLRequest's HTTPMethod property stub to be readonly

angelodipaolo:
- Add several of `NSURLRequest`'s properties as stubs
- Change `NSURLRequest`'s `HTTPMethod` property stub to be readonly
- Add mutable property stub to `NSMutableURLRequest`


parkera:
I want to merge `ed31a34` (thanks for catching that), but the other one I have a few comments on. We made a conscious decision to leave off those particular properties and replace them with using the properties on NSURLSessionConfiguration.

angelodipaolo:
Thanks for the information. Using `NSURLSessionConfiguration` makes sense for those properties. I'm closing this PR in favor of https://github.com/apple/swift-corelibs-foundation/pull/14 which has the `ed31a34` cherry picked.

PR #15, opened 2015-12-04
[Need Feedback]: Added bounds checking to NSArray subscript

KaiCode2:
Completed bounds checking however I need feedback on what should should be done if the index is out of range.

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -344,8 +344,11 @@ public class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NS
     
     public subscript (idx: Int) -> AnyObject {
         get {
-            // TODO: Bounds checking
-            return objectAtIndex(idx)
+                guard idx < count && idx >= 0 else {
+                    fatalError("\(self): Index out of bounds")
+                }
+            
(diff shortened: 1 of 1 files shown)


KaiCode2:
@phausler I corrected the PR to incorporate the changes you mentioned. Let me know if there is anything else to add 🙂

parkera:
Thanks, this looks good. Could you rebase this into a single commit for us to merge in?

Adlai-Holler:
Would `precondition(idx >= 0 && idx < count)` be a cleaner equivalent?

KaiCode2:
@parkera finished the rebase, let me know if it's all good.

parkera:
@Adlai-Holler precondition sounds like it might be good too, since it takes the optimization level into account. We can consider that as another change (also perhaps moving this into the objectAtIndex call itself as discussed on mailing list).

KaiCode2:
@Adlai-Holler I think the issue with the precondition is the lack of an error message if the condition fails. Providing an error message is something I think is quite important to do.

gribozavr:
You can provide a message as the second parameter to `precondition()`.  The message is displayed in debug builds.

PR #19, opened 2015-12-04
Conform NSDate to Comparable

chrisamanse:


JiriTrecak:
Personally, I think compare function can be shortened so it does not introduce unnecessary overhead, to:

    func <(lhs: NSDate, rhs: NSDate) -> Bool {
        return lhs.timeIntervalSinceReferenceDate < rhs.timeIntervalSinceReferenceDate
    }

chrisamanse:
@JiriTrecak, good point.

parkera:

PR #20, opened 2015-12-04
Add stack collection type

AquaGeek:
This is mostly cribbed from the Swift book, but I figured it'd be nice to add a couple of these standard collection types to Foundation instead of everyone reimplementing them.

### DIFF
--- a/Foundation.xcodeproj/project.pbxproj
+++ b/Foundation.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


harlanhaskins:
I feel like this could just be an extension on Array<T>

```swift
extension Array {
    mutating func push(element: Element) {
        self.append(element)
    }

    mutating func peek() -> Element? {
        return self.last
    }

    mutating func pop() -> Element? {
        if !self.isEmpty {
            return storage.removeLast()
        } else {
            return nil
        }
    }
}
```

parkera:

PR #22, opened 2015-12-04
Prefer for-in loops. Avoid usage of ++ operator.

BryanHoke:


BryanHoke:
I am, however, puzzled by why these loops were written so non-idiomatically to begin with? Were they actually meant as intentional code smells to draw attention to something else?

PR #24, opened 2015-12-04
Prefer for-in loops. Avoid usage of ++ operator.

BryanHoke:


Foundation/NSIndexPath.swift
```diff
@@ -16,7 +16,7 @@ public class NSIndexPath : NSObject, NSCopying, NSSecureCoding {
     }
     public init(indexes: UnsafePointer<Int>, length: Int) {
         _indexes = []
-        for var idx = 0; idx < length; idx++ {
+        for idx in 0..<length {
             _indexes.append(indexes[idx])
```
gribozavr: This whole loop is just `_indexes = Array(UnsafeBufferPointer(start: indexes, count: length))`.

BryanHoke: Ah, so it is! Very cool.

PR #26, opened 2015-12-04
add implementation for NSRangeFromString and testNSRange.swift

jpsim:
Great job with the OSS release! :clap: 

Here's a first attempt at contributing something small.

All strings in the test cases were compared against the result of using `NSRangeFromString` in Foundation today. Some of that behavior I find odd, but I opted to keep it as-is. I'm happy to consider other implementation approaches too.

### DIFF
--- a/Foundation/NSRange.swift
+++ b/Foundation/NSRange.swift
@@ -99,5 +99,33 @@ public func NSStringFromRange(range: NSRange) -> String {
 }
 
 public func NSRangeFromString(aString: String) -> NSRange {
-    NSUnimplemented()
+    let emptyRange = NSMakeRange(0, 0)
+    if aString.isEmpty {
+        // fail early if the string is empty
+        return emptyRange
+    }
+    let scanner = NSScanner(string: aString)
+    let digitSet = NSCharacterSet.decimalDigitCharacterSet()
+    scanner.scanUpToCharactersFromSet(digitSet)
+    if scanner.atEnd {
+        // fail early if there are no decimal digits
+        return emptyRange
+    }
+    guard let location = scanner.scanInteger() else {
+        return emptyRange
+    }
+    let partialRange = NSMakeRange(location, 0)
+    if scanner.atEnd {
+        // return early if there are no more characters after the first int in the string
+        return partialRange
+    }
+    scanner.scanUpToCharactersFromSet(digitSet)
+    if scanner.atEnd {
+        // return early if there are no integer characters after the first int in the string
+        return partialRange
+    }
+    guard let length = scanner.scanInteger() else {
+        return partialRange
+    }
+    return NSMakeRange(location, length)
 }
--- a/Foundation.xcodeproj/project.pbxproj
+++ b/Foundation.xcodeproj/project.pbxproj
@@ -197,6 +197,7 @@
 		EA66F64E1BF1619600136161 /* TestNSIndexSet.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA66F63E1BF1619600136161 /* TestNSIndexSet.swift */; };
 		EA66F6501BF1619600136161 /* TestNSNumber.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA66F63F1BF1619600136161 /* TestNSNumber.swift */; };
 		EA66F6521BF1619600136161 /* TestNSPropertyList.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA66F6401BF1619600136161 /* TestNSPropertyList.swift */; };
+		E876A73E1C1180E000F279EC /* TestNSRange.swift in Sources */ = {isa = PBXBuildFile; fileRef = E876A73D1C1180E000F279EC /* TestNSRange.swift */; };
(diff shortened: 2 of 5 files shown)


Foundation/NSRange.swift
```diff
@@ -99,5 +99,33 @@ public func NSStringFromRange(range: NSRange) -> String {
...
+        return emptyRange
+    }
+    let scanner = NSScanner(string: aString)
+    let digitSet = NSCharacterSet.decimalDigitCharacterSet()
+    scanner.scanUpToCharactersFromSet(digitSet)
+    if scanner.atEnd {
+        // fail early if there are no decimal digits
+        return emptyRange
+    }
+    guard let location = scanner.scanInt() else {
```
phausler: scanInt is perhaps not exactly correct since that would only really be applicable for Int32; NSRange is defined as Int (aka long in c) so the size could technically be up to a 64 bit number on 64 bit platforms; so scanInteger probably is more correct here and also removes the nasty cast later on ;)

Foundation/NSRange.swift
```diff
@@ -99,5 +99,33 @@ public func NSStringFromRange(range: NSRange) -> String {
...
+    }
+    let partialRange = NSMakeRange(Int(location), 0)
+    if scanner.atEnd {
+        // return early if there are no more characters after the first int in the string
+        return partialRange
+    }
+    scanner.scanUpToCharactersFromSet(digitSet)
+    if scanner.atEnd {
+        // return early if there are no integer characters after the first int in the string
+        return partialRange
```
phausler: the failure cases for the objective-c version currently return -1 for anything not parseable. It is not an explicit goal of this to be in parity for failure mode but it is nice when we can easily; which probably can be done here too.

Foundation/NSRange.swift
jpsim: great point! thanks

parkera:
@jpsim Would you mind rebasing these into a single commit for merging? Thanks!

jpsim:
Rebased against `master` and fixed up the second commit into the first.

parkera:
Sorry, found another issue - TestNSRange needs to be added to the `build.py` file so that the test is found when building with ninja on Linux.

jpsim:
> Sorry, found another issue - TestNSRange needs to be added to the build.py file so that the test is found when building with ninja on Linux.

I did add it. Is it not at the right place?

phausler:
looks correct to me

parkera:
I see, I was compiling on Linux but forgot to reconfigure so the test file was not in the ninja script. We'll see about improving that process as well.

johankj:
I don’t yet have a build environment setup so I can’t test it, but does this implementation support hexadecimal ranges as input, like original implementation?

```swift
NSStringFromRange(NSRangeFromString(@"{0x10,0x20}")) // "{16, 32}"
```

jpsim:
> I don’t yet have a build environment setup so I can’t test it, but does this implementation support hexadecimal ranges as input, like original implementation?

No it does not currently, but I can confirm that "standard" Foundation behaves like that.

I was unaware of this behavior. Is that documented anywhere? I can't find it in the official documentation: https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Functions/#//apple_ref/c/func/NSRangeFromString.

phausler:
NSRangeFromString will take 0X and 0x for hexadecimal values, otherwise it is integer values.

jpsim:
> NSRangeFromString will take 0X and 0x for hexadecimal values, otherwise it is integer values.

Thanks for sharing! This discovery leads me to several comments and questions though:

I would have appreciated being informed of this discrepancy in my implementation by reviewers with source access to "standard" Foundation. Especially given that this is undocumented behavior (AFAICT).

Is it a stated goal for Swift Foundation to match undocumented behavior of Standard Foundation? If so, how are external contributors expected to know this? Is reverse engineering Standard Foundation encouraged/allowed?

PS: is "Swift Foundation" and "Standard Foundation" the appropriate way to refer to these projects?

parkera:
@jpsim We're here to help clarify the behavior expected of Swift Foundation (also, so far we're referring to the Objective-C implementation as Darwin Foundation). However, it is certain that we will sometimes also forget about behaviors in certain areas of the framework, so we'll just have to follow those up with additional changes and tests when we find them. Sometimes we may also choose to accept a partial implementation just to move things forward.

parkera:
Also, we can track those with issues on the [bug tracker](https://bugs.swift.org) when we find them, especially if it's a significant task.

dbrajkovic:
@parkera @jpsim This test ```test_NSRangeFromString``` keeps failing because it says ```rangeOfCharacterFromSet``` in NSString is unimplemented.

parkera:
It looks like @phausler  commented it out in e1c253f172d8f0a16bde67f6ebd253a0b7d034ed

jpsim:
Thanks for all these clarifications! 'Darwin Foundation' makes sense.

> This test test_NSRangeFromString keeps failing because it says rangeOfCharacterFromSet in NSString is unimplemented.

Ah, I apologize for this. I hadn't managed to set up a testing environment for this when I submitted it, but now I do, so that shouldn't happen again.

PR #27, opened 2015-12-04
Simplify implementation of `init(indexes:length:)`.

BryanHoke:
In `init(indexes:length:)`, initialize `_indexes` to an `Array` constructed with an `UnsafeBufferPointer` built from `indexes`, instead of initializing it to an empty array and adding each `idx` in `indexes` to it in a loop.


parkera:
Thanks!

PR #29, opened 2015-12-04
mark NSCache() initializer as public

krzyzanowskim:
mark NSCache() initializer as public to allow instantiate NSCache outside Foundation framework.


phausler:
looks good to me

PR #31, opened 2015-12-04
Make NSDate conform to Comparable protocol

tomaskraina:
It would be nice to provide more natural way to compare instances of `NSDate` -- using `<`, `<=`, `>`, and `>=` operators. This PR implements all methods declared in `Comparable` protocol for `NSDate` class.

### DIFF
--- a/Foundation/NSDate.swift
+++ b/Foundation/NSDate.swift
@@ -174,6 +174,28 @@ extension CFDateRef : _NSBridgable {
     internal var _nsObject: NSType { return unsafeBitCast(self, NSType.self) }
 }
 
+extension NSDate : Comparable { }
+
+public func ==(lhs: NSDate, rhs: NSDate) -> Bool {
+    return lhs === rhs || lhs.compare(rhs) == .OrderedSame
+}
+
+public func <(lhs: NSDate, rhs: NSDate) -> Bool {
+    return lhs.compare(rhs) == .OrderedAscending
+}
+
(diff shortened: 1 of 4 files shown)


Foundation/NSDate.swift
```diff
@@ -174,6 +174,28 @@ extension CFDateRef : _NSBridgable {
...
+public func <=(lhs: NSDate, rhs: NSDate) -> Bool {
+    return lhs < rhs || lhs == rhs
+}
+
+public func >(lhs: NSDate, rhs: NSDate) -> Bool {
+    return lhs.compare(rhs) == .OrderedDescending
+}
+
+public func >=(lhs: NSDate, rhs: NSDate) -> Bool {
+    return lhs > rhs || lhs == rhs
```
tevelee: Hi @tomaskraina. Great, simple and readable solution :+1: 
I'd like to suggest a small performance optimisation here: if `lhs` and `rhs` are equal you actually call the compare method 2 times on the same pair. Once inside the `>` method and then once again in `==`. Instead, you can save the result of compare and use `switch` to accept the valid cases or even use some pattern matching magic :-)
Since this is the foundation library, IMO we need to be conscious about performance decisions, even if they're small.

chrisamanse: @tomaskraina, isn't implementing `==` and `<` enough to make it conform to `Comparable`? Anyway, this is the same as pull request #19.

tomaskraina:
@lorentey good point! Updated.

parkera:

PR #32, opened 2015-12-04
Verify contents of the file attributes dictionary

bolshedvorsky:
1. NSString values replaced with String values
2. API return type changed to [String : Any]
3. NSFileOwnerAccountName and NSFileGroupOwnerAccountID values are not verified since they are optional. NSFileImmutable and NSFileAppendOnly values are not verified since they are used on iOS and OS X


parkera:
Thanks!

bolshedvorsky:
I will create a fixing PR.I've got both El Capitan and Linux.

bolshedvorsky:
I've pushed a PR #58 with a fix and comments for reference.

PR #33, opened 2015-12-04
Clarify where NSCalendar timezone is set and return values

syoutsey:
Flow is more logical and easy to follow by setting `timezone` immediately after declaration. Additionally, return values more closely follow idioms specified [here](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Functions.html#//apple_ref/doc/uid/TP40014097-CH10-ID528).

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -365,11 +365,7 @@ public class NSCalendar : NSObject, NSCopying, NSSecureCoding {
     public func dateFromComponents(comps: NSDateComponents) -> NSDate? {
         var (vector, compDesc) = _convert(comps)
         
-        let oldTz = timeZone
-        let tempTz = comps.timeZone
-        if let tz = tempTz {
-            timeZone = tz
-        }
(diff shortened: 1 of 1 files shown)


parkera:
This looks ok, but can you fix up the whitespace and rebase before we merge?

syoutsey:
Done, thanks!

parkera:
Great - just go ahead and rebase this into 1 commit and I'll merge it in. Thanks.

Foundation/NSCalendar.swift
```diff
@@ -368,7 +368,9 @@ public class NSCalendar : NSObject, NSCopying, NSSecureCoding {
         let oldTz = timeZone
```
jpsim: all these lines could be `self.timeZone = timeZone ?? comps.timeZone`

syoutsey:
@jpsim Great idea, though it looks like the operands are out of order. Fixed and rebased. Thanks!

parkera:
Looks good.

PR #34, opened 2015-12-04
Add some implementation of NSHTTPCookie

dbrajkovic:
Added some of the basic implementation of NSHTTPCookie

### DIFF
--- a/Foundation/NSHTTPCookie.swift
+++ b/Foundation/NSHTTPCookie.swift
@@ -11,79 +11,79 @@
     @const NSHTTPCookieName
     @discussion Key for cookie name
 */
-public let NSHTTPCookieName: String = "" // NSUnimplemented
+public let NSHTTPCookieName: String = "Name"
 
 /*!
     @const NSHTTPCookieValue
     @discussion Key for cookie value
 */
-public let NSHTTPCookieValue: String = "" // NSUnimplemented
+public let NSHTTPCookieValue: String = "Value"
 
 /*!
     @const NSHTTPCookieOriginURL
     @discussion Key for cookie origin URL
 */
-public let NSHTTPCookieOriginURL: String = "" // NSUnimplemented
+public let NSHTTPCookieOriginURL: String = "OriginURL"
 
 /*!
     @const NSHTTPCookieVersion
     @discussion Key for cookie version
 */
-public let NSHTTPCookieVersion: String = "" // NSUnimplemented
+public let NSHTTPCookieVersion: String = "Version"
 
 /*!
     @const NSHTTPCookieDomain
     @discussion Key for cookie domain
 */
-public let NSHTTPCookieDomain: String = "" // NSUnimplemented
+public let NSHTTPCookieDomain: String = "Domain"
 
 /*!
     @const NSHTTPCookiePath
     @discussion Key for cookie path
 */
-public let NSHTTPCookiePath: String = "" // NSUnimplemented
+public let NSHTTPCookiePath: String = "Path"
 
 /*!
     @const NSHTTPCookieSecure
     @discussion Key for cookie secure flag
 */
-public let NSHTTPCookieSecure: String = "" // NSUnimplemented
+public let NSHTTPCookieSecure: String = "Secure"
 
 /*!
     @const NSHTTPCookieExpires
     @discussion Key for cookie expiration date
 */
-public let NSHTTPCookieExpires: String = "" // NSUnimplemented
+public let NSHTTPCookieExpires: String = "Expires"
 
 /*!
     @const NSHTTPCookieComment
     @discussion Key for cookie comment text
 */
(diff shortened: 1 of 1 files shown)


Foundation/NSHTTPCookie.swift
```diff
@@ -277,21 +294,27 @@ public class NSHTTPCookie : NSObject {
         Version 1 maps to RFC2965 cookies. There may be future versions.
         @result the version of the receiver.
     */
-    public var version: Int { NSUnimplemented() }
+    public var version: Int {
+        return _properties[NSHTTPCookieVersion]
```
phausler: this won't work as written on linux since it is implicitly bridging to NSNumber

Foundation/NSHTTPCookie.swift
```diff
@@ -277,21 +294,27 @@ public class NSHTTPCookie : NSObject {
...
+    }
     
     /*!
         @method name
         @abstract Returns the name of the receiver.
         @result the name of the receiver.
     */
-    public var name: String { NSUnimplemented() }
+    public var name: String {
+        return _properties[NSHTTPCookieName]
```
phausler: this won't work as written on linux since it is implicitly bridging to NSString

dbrajkovic: I've changed my pull request. Is this the correct way to handle then?

```
    return _properties[NSHTTPCookieName] as! NSString)._swiftObject)
```

parkera: Actually, I think what we want to do here is keep the _properties dictionary holding a Swift.String object instead of NSString.

parkera:
@salgernon please review.

Foundation/NSHTTPCookie.swift
```diff
@@ -277,21 +290,27 @@ public class NSHTTPCookie : NSObject {
         Version 1 maps to RFC2965 cookies. There may be future versions.
         @result the version of the receiver.
     */
-    public var version: Int { NSUnimplemented() }
+    public var version: Int {
+        return (_properties[NSHTTPCookieVersion] as! NSNumber)._swiftObject)
```
ChrisChares: Each one of these closures has an extra parenthesis

Foundation/NSHTTPCookie.swift
```diff
@@ -11,79 +11,79 @@
...
 */
-public let NSHTTPCookieDiscard: String = "" // NSUnimplemented
+public let NSHTTPCookieDiscard: String = "Discard"
 
 /*!
     @const NSHTTPCookieMaximumAge
     @discussion Key for cookie maximum age (an alternate way of specifying the expiration)
 */
-public let NSHTTPCookieMaximumAge: String = "" // NSUnimplemented
+public let NSHTTPCookieMaximumAge: String = "MaximumAge"
```
ChrisChares: Should be "Max-Age" in accordance with rfc6265

Foundation/NSHTTPCookie.swift
```diff
@@ -219,7 +219,18 @@ public class NSHTTPCookie : NSObject {
...
+        func isValidProperty(property: String) -> Bool {
+            return property.length > 0 && (property as NSString).rangeOfString("/n") == NSNotFound
+        }
+        if  !isValidProperty(properties[NSHTTPCookiePath])
+            || !isValidProperty(properties[NSHTTPCookieDomain])
+            || !isValidProperty(properties[NSHTTPCookieName])
+            || !isValidProperty(properties[NSHTTPCookieValue]) {
+                return nil
+        }
+        self.properties = properties
```
ChrisChares: self.properties is defined as read only

Foundation/NSHTTPCookie.swift
```diff
@@ -370,7 +397,9 @@ public class NSHTTPCookie : NSObject {
         @result The comment of the receiver, or nil if the receiver has no
         comment.
     */
-    public var comment: String? { NSUnimplemented() }
+    public var comment: String? {
+        return (_properties[NSHTTPCookieComment] as! NSString)._swiftObject)
```
ChrisChares: Trying to access these such optional values would throw a runtime exception with the `as!` statement

ChrisChares:
What's the best practice for contributing to an active pull request?  Thanks.

dbrajkovic:
Updated code per comments.  Wanted to add unit tests, but can't seem to get the unit tests to run. Any advice?

salgernon:
I added a comment on the diff that I assume is the latest:

https://github.com/dbrajkovic/swift-corelibs-foundation/commit/5cb8bceef4b57e70e25dc9dadcb5c45056b9f83b

Which I believe is pull request #34; I'm not sure how that is going to get propagated to the discussion, but I would assume that it does.. somehow become part of the global conversation about this pull request.  (I'll wait and see.)

--sma

PR #35, opened 2015-12-05
Implement basic geometry functions

romanroibu:
Implemented some basic geometry functions

### DIFF
--- a/Foundation/NSGeometry.swift
+++ b/Foundation/NSGeometry.swift
@@ -50,6 +50,12 @@ public struct CGPoint {
     }
 }
 
+extension CGPoint: Equatable { }
+
+public func ==(lhs: CGPoint, rhs: CGPoint) -> Bool {
+    return (lhs.x == rhs.x) && (lhs.y == lhs.y)
+}
+
 public struct CGSize {
     public var width: CGFloat
     public var height: CGFloat
--- a/Foundation/NSGeometry.swift
+++ b/Foundation/NSGeometry.swift
@@ -69,6 +69,12 @@ public struct CGSize {
     }
 }
 
+extension CGSize: Equatable { }
+
+public func ==(lhs: CGSize, rhs: CGSize) -> Bool {
+    return (lhs.width == rhs.width) && (lhs.height == lhs.height)
+}
+
 public struct CGRect {
     public var origin: CGPoint
     public var size: CGSize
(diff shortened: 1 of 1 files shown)


Foundation/NSGeometry.swift
```diff
@@ -53,6 +69,12 @@ public struct CGSize {
     }
 }
 
+extension CGSize: Equatable { }
+
+public func ==(lhs: CGSize, rhs: CGSize) -> Bool {
+    return (lhs.width == rhs.width) && (lhs.height == lhs.height)
```
johankool: `lhs.height == ls.height` is always true. You want `lhs.height == rhs.height`

Foundation/NSGeometry.swift
```diff
@@ -40,6 +50,12 @@ public struct CGPoint {
     }
 }
 
+extension CGPoint: Equatable { }
+
+public func ==(lhs: CGPoint, rhs: CGPoint) -> Bool {
+    return (lhs.x == rhs.x) && (lhs.y == lhs.y)
```
johankool: You want `lhs.y == rhs.y` here instead.

Foundation/NSGeometry.swift
```diff
@@ -66,6 +88,12 @@ public struct CGRect {
     }
 }
 
+extension CGRect: Equatable { }
+
+public func ==(lhs: CGRect, rhs: CGRect) -> Bool {
+    return (lhs.origin == rhs.origin) && (lhs.size == lhs.size)
```
johankool: You want `lhs.size == rlhs.size` here instead.

romanroibu: Wow, can't believe I've missed those. Thanks!

parkera:
Let's rebase this after incorporating @johankool 's comments so we can merge it in. Thanks!

romanroibu:
Done, thanks!

Foundation/NSGeometry.swift
```diff
@@ -66,6 +88,12 @@ public struct CGRect {
     }
 }
 
+extension CGRect: Equatable { }
+
+public func ==(lhs: CGRect, rhs: CGRect) -> Bool {
```
phausler: This functionality on Darwin actually lives in CoreGraphics not in Foundation

phausler:
Looks fine to me.

PR #36, opened 2015-12-05
Add NSNull.null()

gshahbazian:
Returns kCFNull and bridges from CFNullRef
as described in Apple documentation.

### DIFF
--- a/Foundation.xcodeproj/project.pbxproj
+++ b/Foundation.xcodeproj/project.pbxproj
@@ -193,6 +193,7 @@
 		5BF7AEBF1BCD51F9008F214A /* NSURL.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5BDC3F4A1BCC5DCB00ED97BB /* NSURL.swift */; };
 		5BF7AEC01BCD51F9008F214A /* NSUUID.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5BDC3F4B1BCC5DCB00ED97BB /* NSUUID.swift */; };
 		5BF7AEC11BCD51F9008F214A /* NSValue.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5BDC3F4C1BCC5DCB00ED97BB /* NSValue.swift */; };
+		6C2CDAE51C12AB5200E52B3B /* TestNSNull.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6C2CDAE31C12AB3900E52B3B /* TestNSNull.swift */; };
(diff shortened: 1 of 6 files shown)


parkera:
This looks great, thanks. One small change - we need to add the new TestNSNull file to `build.py` so it gets included in our test suite on Linux.

It would also be nice to have a test which makes sure that this works:

Call Foundation API, which calls CF API internally, which returns kCFNull, which is returned from Foundation API, which is == to NSNull.

That is - a test which makes sure that the bridging from CF is working.

gshahbazian:
@parkera thanks for catching that. Added to `build.py` and changed the test.

phausler:
Actually i don't think there is a method NSNull.null() in Foundation - the importer seems to remove the factory pattern (which does not yet exist in swift). To mimic the behavior of the objc version the == and === operators could be overloaded for NSNull to make comparison always true.

The bridging however looks good (does that unit test actually work on linux? if so we have a module import leaking across)

parkera:
Yah unfortunately you can go ahead and import CF even though we want it to be private.

I think I was a bit unclear on my request. What I wanted to verify was that any CF code that produces kCFNull, which we then return from Foundation by using the bridging, will be seen as NSNull from Swift.

gshahbazian:
> Actually i don't think there is a method NSNull.null() in Foundation - the importer seems to remove the factory pattern (which does not yet exist in swift).

You're correct. What are the implications for this change given that fact?

> I think I was a bit unclear on my request. What I wanted to verify was that any CF code that produces kCFNull, which we then return from Foundation by using the bridging, will be seen as NSNull from Swift.

Gotcha, will update the test.

phausler:
The missing factory pattern is actually a language extension that would greatly benefit Foundation (and probably other libraries)

I have many times run across the need to have a factory init pattern like:

```swift
/// fictitious syntax
public factory init() -> NSNull {
    return unsafeBitCast(kCFNull, NSNull.self)
}
```

This would allow us to have proper abstraction of class clusters and the sort, but it is not yet available in the language.

For now the best way to combat this specific issue is to overload the operators of == and === to fake it out into being equal in addition to the bridging. Granted it will still allocate per NSNull but it will be better than the current standings.

parkera:
After looking at this again, @phausler is right that we shouldn't add the class method. The reason is that it doesn't exist on Darwin, so it would be incompatible if someone used it for Linux.

@gshahbazian can you try the suggestion of overloading the == and === operators? That is, all NSNull instances compare equal to each other.

parkera:
I think this PR has been superseded by #129.

PR #37, opened 2015-12-05
Implemented NSAffineTransform.init() with tests

mattrajca:
As described in the reference for NSAffineTransform, this initializer:

    Creates and returns a new NSAffineTransform object initialized to the identity matrix.

Part of this change:

- an initializer for the underlying `NSAffineTransformStruct`'s members has been implemented
- an initializer for creating `Double`s from `CGFloat`s has been added to aid testing. Its signature is identical to the one in `CoreGraphics` in the iOS 9.1 SDK. Tests are implemented in terms of `Double`s since `XCTAssertEqualWithAccuracy` only works with types that implement `FloatingPointType`, which `CGFloat` currently does not.


mattrajca:
The tests for this pass once https://github.com/apple/swift-corelibs-xctest/pull/7 is fixed.

mattrajca:
apple/swift-corelibs-xctest#7 has been fixed and the tests now pass.

parkera:
I had to fix up a couple of things (DBL_MAX not available on Linux, init should be `@_transparent`), but after that I merged it in.

PR #38, opened 2015-12-05
Added `NSPipe` implementation

terhechte:
The implementation itself is very simple. In the unlikely event that the
`pipe` system call fails, this implementation will stop with a fatalError.
This seems to be the canonical way of doing it based on my cursory reading of
the rest of foundation.

I've also added a first, simple, unit test to make sure the pipe works as expected.

### DIFF
--- a/TestFoundation/TestNSPipe.swift
+++ b/TestFoundation/TestNSPipe.swift
@@ -19,24 +19,32 @@ import SwiftXCTest
 
 class TestNSPipe : XCTestCase {
     
+    public var allTests: [(String, () -> ())] {
+        return [
+            ("test_NSPipe", test_NSPipe)
+        ]
+    }
+    
     func test_NSPipe() {
         let aPipe = NSPipe()
         let text = "test-pipe"
         
         // First write some data into the pipe
-        aPipe.fileHandleForWriting.writeData(text.dataUsingEncoding(NSUTF8StringEncoding)!)
+        let stringAsData = text.bridge().dataUsingEncoding(NSUTF8StringEncoding)
+        XCTAssertNotNil(stringAsData)
+        aPipe.fileHandleForWriting.writeData(stringAsData!)
         
         // Then read it out again
         let data = aPipe.fileHandleForReading.readDataOfLength(text.characters.count)
         
-        // Make sure we *did* get data
+        // Confirm that we did read data
         XCTAssertNotNil(data)
         
-        // Make sure the data can be converted
+        // Confirm the data can be converted to a String
         let convertedData = String(data: data, encoding: NSUTF8StringEncoding)
         XCTAssertNotNil(convertedData)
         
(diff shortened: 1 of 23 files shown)


TestFoundation/TestNSPipe.swift
```diff
@@ -0,0 +1,39 @@
...
+
+class TestNSPipe : XCTestCase {
+    
+    func test_NSPipe() {
+        let expectation = self.expectationWithDescription("Should read data")
+        let aPipe = NSPipe()
+        
+        let text = "test-pipe"
+        
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0)) { () -> Void in
```
phausler: the swift corelibs dispatch is not yet available for linux unfortunately so a thread or some other mechanism would have to be used to asynchronously read.

terhechte:
I've updated the test to remove the libdispatch requirement.

TestFoundation/TestNSPipe.swift
```diff
@@ -0,0 +1,42 @@
...
+import Foundation
+import XCTest
+#else
+import SwiftFoundation
+import SwiftXCTest
+#endif
+
+
+class TestNSPipe : XCTestCase {
+
```
phausler: This needs a method for allTests
```swift
public var allTests: [(String, () -> ())] { 
    return [
         ("test_NSPipe", test_NSPipe)
    ]
 }
```

TestFoundation/TestNSPipe.swift
```diff
@@ -0,0 +1,42 @@
...
+
+class TestNSPipe : XCTestCase {
+    
+    func test_NSPipe() {
+        let aPipe = NSPipe()
+        let text = "test-pipe"
+        
+        // First write some data into the pipe
+        aPipe.fileHandleForWriting.writeData(text.dataUsingEncoding(NSUTF8StringEncoding)!)
```
phausler: This is relying on implicit bridging and should be 
```swift
text.bridge().dataUsingEncoding(NSUTF8StringEncoding)
```

terhechte:
Thanks for these comments! I've updated the test and I've also included an additional assertion for the initial string to nsdata conversion.

phausler:
One other minor thing it needs is the TestNSPipe.swift added to build.py (or rebase from master which now globs; which is preferable) and in TestFoundation/main.swift add the test to the XCTMain

terhechte:
I did a rebase so that the `glob` in build.py could be used and added to main.swift.

phausler:
Builds correctly now; running the tests on linux fail due to missing functionality in NSString; but overall the implementation looks good to me, I am going to merge this in and just disable the test until NSString implementations get up to par to deal with this.

terhechte:
Awesome, thanks!

PR #40, opened 2015-12-05
Clarify GettingStarted.md

radex:
Two small clarifications:

1. peer directory -> sibling directory (I believe this is a standard, common way of saying this)
2. The Xcode target is called "SwiftFoundation", not "Foundation"


parkera:
Looks good to me. Sibling makes more sense.

PR #41, opened 2015-12-05
NSBundle additions

radex:
This adds a lot of basic NSBundle properties and methods that are simple ports of CoreFoundation to Swift, including:

- URL/path convenience properties
- infoDictionary and a few related things
- localization property and related

This also adds a bunch of basic tests to the functionality above (testing the TestFoundation bundle)


parkera:
Unfortunately this bridging pattern doesn't work without ObjC present (e.g. on Linux):

```
Foundation/NSBundle.swift:194:52: error: '_??' is not convertible to '[String : AnyObject]?'
        return CFBundleGetInfoDictionary(_bundle)?._swiftObject as! [String: AnyObject]?
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
Foundation/NSBundle.swift:197:57: error: '_??' is not convertible to '[String : AnyObject]?'
        return CFBundleGetLocalInfoDictionary(_bundle)?._swiftObject as! [String: AnyObject]?
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
Foundation/NSBundle.swift:206:75: error: '[_]' is not convertible to '@autoclosure () throws -> T'
        return (CFBundleCopyBundleLocalizations(_bundle)?._swiftObject ?? []) as! [String]
                                                                          ^~
Foundation/NSBundle.swift:210:108: error: '[_]' is not convertible to '@autoclosure () throws -> T'
        return (CFBundleCopyPreferredLocalizationsFromArray(localizationsArray._cfObject)?._swiftObject ?? []) as! [String]
```

In the meantime, for CFBundle's collection results we will have to use the `_expensivePropertyListConversion` function in NSPropertyListSerialization. The bridging story is something that we absolutely need to figure out in more depth in the very near term.

radex:
@parkera Hm, the problem is that `_expensivePropertyListConversion` returns `[String: Any]`, and NSBundle methods expect `[String: AnyObject]`. I tried a few different ways, but those `Any`s just don't want to convert to `AnyObject`.

I tried changing the `_expensivePropertyListConversion` function so that it produces `[String: AnyObject]`/`[AnyObject]` and that seemed to work for me (but obviously broke a bunch of other code expecting `Any`).

What should I do?

parkera:
In these cases what we are doing is changing the return type of the Foundation method to [String:Any] and marking it (in headerdoc) with `Experimental` so we can track these API differences and figure out what to do about them in Darwin Foundation.

If we're going to return actual String types from this API, and we lack bridging, it has to be Any instead of AnyObject because Swift.String is not an Object.

radex:
Thanks, will do that.

radex:
@parkera OK, changed it to use `_expensivePropertyListConversion` as requested. I have no way of testing on Linux right now, but hopefully will work.

Foundation/NSPropertyList.swift
```diff
@@ -113,6 +113,8 @@ internal func _expensivePropertyListConversion(input : AnyObject) -> Any {
         return data
     } else if let number = input as? NSNumber {
         return number
+    } else if let url = input as? NSURL {
```
phausler: NSURL is not a plist type so this is not exactly correct here

parkera: I guess I was thinking of using this as a general purpose conversion function until we get our bridging story straightened out.

radex: Hm, I'm pretty sure `NSBundle.infoDictionary` included a value somewhere that was an NSURL — triggering a fatalError without it... What should I do instead?

parkera: Can you track down where that is? I would be surprised if we inserted an NSURL there; since as @phausler says it is not a plist type.

TestFoundation/TestNSBundle.swift
```diff
@@ -0,0 +1,97 @@
...
+        // etc
+        #if os(OSX)
+        XCTAssertEqual("\(path)/Contents/Resources", bundle.resourcePath)
+        XCTAssertEqual("\(path)/Contents/MacOS/TestFoundation", bundle.executablePath)
+        XCTAssertEqual("\(path)/Contents/Frameworks", bundle.privateFrameworksPath)
+        XCTAssertEqual("\(path)/Contents/SharedFrameworks", bundle.sharedFrameworksPath)
+        XCTAssertEqual("\(path)/Contents/SharedSupport", bundle.sharedSupportPath)
+        #endif
+        
+        XCTAssertNil(bundle.pathForAuxiliaryExecutable("no_such_file"))
```
phausler: this unfortunately currently fails on linux due to a release of NULL in CFRelease(exeDirURL) @ CFBundle.c:1591

phausler:
this seems to build cleanly on linux; however there do seem to be a few failures.

Foundation/NSPropertyList.swift
radex: `NSBundle.mainBundle().infoDictionary!["CFBundleInfoPlistURL"]` returns this.

Looks like CoreFoundation adds this key to the actual plist contents here: https://github.com/apple/swift-corelibs-foundation/blob/4a6d90f0406c34e896bfd3e89585717496bb2c1f/CoreFoundation/PlugIn.subproj/CFBundle_InfoPlist.c#L534

TestFoundation/TestNSBundle.swift
radex: Right, exeDirURL is only set for Windows and Apple platforms. At the very least, CFRelease could be wrapped in `if (exeDirURL != NULL)` — but I don't know what the real solution is. A lot of bundle stuff is just undefined on CoreFoundation side for Linux at the moment.

I could also wrap this test temporarily in `#if os(OSX)`, but I don't know how many other tests will fail until there's proper Linux support done on CF side…

Foundation/NSPropertyList.swift
parkera: Interesting. I guess we get away with it because it's an "internal" key and no one who gets the contents of the Info.plist is supposed to see it.

Frankly, it's wrong. We should stash that URL inside the CFBundle struct instead. What that means is that _CFBundleCopyInfoDictionaryInDirectoryWithVersion should return the final URL (probably a byref arg), and CFBundleGetInfoDictionary should put it inside the CFBundle struct.

parkera: @radex do you mind taking on that change as well? With that we should remove the requirement to convert NSURL in the NSPropertyList method.

radex: Yes, I'll try :)

radex:
> Interesting. I guess we get away with it because it's an "internal" key and no one who gets the contents of the Info.plist is supposed to see it.
>
> Frankly, it's wrong. We should stash that URL inside the CFBundle struct instead. What that means is that _CFBundleCopyInfoDictionaryInDirectoryWithVersion should return the final URL (probably a byref arg), and CFBundleGetInfoDictionary should put it inside the CFBundle struct.

@parkera I did my best to make those changes, but I'm not super convinced in their correctness, especially that my C is a bit rusty, and I don't have much experience with CF-style memory management.

Take a look at the last two commits.

parkera:
I'll merge this in with a couple of minor changes once we get our tests passing on Linux again (an issue unrelated to this PR).

radex:
Awesome!

PR #43, opened 2015-12-05
Improve control flow and some fix

ken0nek:
- Use `guard let` and return early.
- Remove needless parenthesis.
- combiine several `if let`

### DIFF
--- a/Foundation/NSXMLParser.swift
+++ b/Foundation/NSXMLParser.swift
@@ -73,47 +73,39 @@ internal func _NSXMLParserExternalEntityWithURL(interface: _CFXMLInterface, urlS
     
     switch policy {
     case .ResolveExternalEntitiesSameOriginOnly:
-        if let url = parser._url {
-            if a == nil {
-                a = NSURL(string: String(urlStr))
+        guard let url = parser._url else { break }
+        
+        if a == nil {
+            a = NSURL(string: String(urlStr))
+        }
+        
+        guard let aUrl = a else { break }
+        
+        var matches: Bool
+        if let aHost = aUrl.host, host = url.host {
+            matches = host == aHost
+        } else {
+            return nil
+        }
+        
+        if matches {
+            if let aPort = aUrl.port, port = url.port {
+                matches = port == aPort
+            } else {
+                return nil
             }
-            if let aUrl = a {
-                var matches: Bool
-                if let aHost = aUrl.host {
-                    if let host = url.host {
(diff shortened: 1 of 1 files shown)


Foundation/NSXMLParser.swift
```diff
@@ -73,48 +73,40 @@ internal func _NSXMLParserExternalEntityWithURL(interface: _CFXMLInterface, urlS
...
-        if let url = parser._url {
-            if a == nil {
-                a = NSURL(string: String(urlStr))
+        guard let url = parser._url else { break }
+        
+        if a == nil {
+            a = NSURL(string: String(urlStr))
+        }
+        
+        guard let aUrl = a else { break }
```
kaandedeoglu: why don't you just return NSURL(string: String(urlStr)) in the guard's else case instead of checking for a's (non)existence twice?

Foundation/NSXMLParser.swift
```diff
@@ -329,13 +317,13 @@ internal func _NSXMLParserEndElementNs(ctx: _CFXMLInterface , localname: UnsafeP
     let reportQNameURI = parser.shouldProcessNamespaces
     let prefixLen = prefix == nil ? strlen(UnsafePointer<Int8>(prefix)) : 0
     let localnameString = (prefixLen == 0 || reportQNameURI) ? UTF8STRING(localname) : nil
-    let nilStr: String? = nil
+    let nilStr: String?
```
phausler: this is required to have an initialization by the compiler

Foundation/NSXMLParser.swift
```diff
@@ -73,48 +73,40 @@ internal func _NSXMLParserExternalEntityWithURL(interface: _CFXMLInterface, urlS
     
     switch policy {
     case .ResolveExternalEntitiesSameOriginOnly:
-        if let url = parser._url {
-            if a == nil {
-                a = NSURL(string: String(urlStr))
+        guard let url = parser._url else { break }
```
phausler: This case looks reasonable to me the original code was a bit of a rat's nest but I think all edges are taken care of here.

parkera:
@phausler what's left to do here?

phausler:
the only missing part is the `let nilStr: String? = nil` and it looks good to me

Foundation/NSXMLParser.swift
ken0nek: yes, I realized nil is needed.

PR #44, opened 2015-12-05
Fix Bugs in NSCalendar Date Calculations

brownleej:
I ran into a couple of crashers when writing tests for NSCalendar date calculations, and this PR is designed to address them.

The first crasher was in initializing an NSCalendar instance. It was attempting to compare the calendar identifiers to the kCF*Calendar constants, but that was getting routed through NSObject#isEqual, which prevented it from matching up the calendar identifiers. I was able to address this by adding an implementation of isEqual for NSString, but I don't know if this implementation is sufficient or correct, so any feedback on that would be welcome.

The second crasher was in _CFCalendarDecomposeAbsoluteTimeV, getting the value for the "leap month" component. The mapping for that component had not been defined yet.

### DIFF
--- a/TestFoundation/TestNSFileManager.swift
+++ b/TestFoundation/TestNSFileManager.swift
@@ -121,7 +121,7 @@ class TestNSFileManger : XCTestCase {
             XCTAssertEqual(fileType!, NSFileTypeRegular)
             
             let fileOwnerAccountID = attrs[NSFileOwnerAccountID] as? NSNumber
-            XCTAssertNotEqual(fileOwnerAccountID!.longLongValue, 0)
+            XCTAssertNotNil(fileOwnerAccountID)
             
         } catch let err {
             XCTFail("\(err)")


CoreFoundation/Locale.subproj/CFCalendar.c
```diff
@@ -422,6 +422,7 @@ static UCalendarDateFields __CFCalendarGetICUFieldCodeFromChar(char ch) {
     case 'G': return UCAL_ERA;
     case 'y': return UCAL_YEAR;
     case 'M': return UCAL_MONTH;
+    case 'l': return UCAL_IS_LEAP_MONTH;
```
phausler: This is something we are going to have to verify since it will upstream into CoreFoundation on darwin.

Foundation/NSString.swift
```diff
@@ -228,7 +228,12 @@ public class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, N
     
     override internal var _cfTypeID: CFTypeID {
         return CFStringGetTypeID()
-    }
+  }
+  
+  public override func isEqual(object: AnyObject?) -> Bool {
```
phausler: This probably could be split out into it's own request since it is relatively isolated from calendrical calculations

Foundation/NSString.swift
```diff
@@ -228,7 +228,12 @@ public class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, N
...
+    guard let string = object as? String else { return false }
```
phausler: object as? String however will not work on linux: it will need to be changed from the implicit type conversion to an explicit version ala (object as? NSString)._swiftObject

TestFoundation/main.swift
```diff
@@ -23,6 +23,7 @@ internal func testBundle() -> NSBundle {
 XCTMain([
     TestNSString(),
     TestNSArray(),
+    TestNSCalendar(),
```
phausler: TestNSCalendar.swift needs to be added to build.py in order to maintain building for linux as well.

CoreFoundation/Locale.subproj/CFCalendar.c
parkera: Looks correct to me.

parkera:
@brownleej This looks good, once the couple of issues @phausler brought up are addressed (splitting out `NSString` diffs and adding test file to `build.py`).

phausler:
I ran this through the tests by manually fixing up the calendar issue and it seems we have a second issue here that causes a crash in the tests the _identifier in the calendar is null which is attempted to be released in __CFCalendarDeallocate which crashes. It might be good to address that in this as well. (not certain if that is a setup or a implementation flaw currently without further debugging)

PR #45, opened 2015-12-05
Implemented NSData.description

harlanhaskins:
I'm not particularly a fan of `"0\(byteStr)"`, but I can't seem to find a way to pad Swift's hex strings.
Any better ideas?

`String(format: "%02X", bytes[i])` ends up being `"(\"%02X\", 4)"`, because of the whole arguments-as-tuples thing. I think `String(format: String, ...)` got removed?


Foundation/NSData.swift
```diff
@@ -145,10 +145,20 @@ public class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
...
-    override public var description: String {
-        get {
-            return "Fixme"
+    private var byteDescription: String {
+        var s = ""
+        let buffer = UnsafePointer<UInt8>(bytes)
+        for i in (0..<self.length) {
+            if i > 0 && i % 4 == 0 { s += " " }
+            var byteStr = String(buffer[i], radix: 16, uppercase: false)
+            if byteStr.characters.count == 1 { byteStr = "0\(byteStr)" }
```
phausler: if the buffer[i] > 0xF it is well known that it will be 2 characters in hex; that would alleviate the need to check the character count.

TestFoundation/main.swift
```diff
@@ -33,4 +33,5 @@ XCTMain([
     TestNSFileManger(),
     TestNSRange(),
     TestNSXMLParser(),
+    TestNSData(),
```
phausler: build.py needs to have TestNSData.swift added to it to make sure it builds on linux too.

Foundation/NSData.swift
harlanhaskins: Right! That's a better check.

parkera: I was also going to suggest we put a cap on self.length here. Data blobs could be arbitrarily long. One approach is to use the first N bytes and last N bytes if over a certain size.

harlanhaskins: What's the behavior in Foundation? I know there are some people that (misguidedly) parse the bytes from `NSData` as crude serialization/deserialization.

phausler: from my understanding description does a full hex dump and debugDescription does a truncated version.

harlanhaskins: Unrelated: Those two should be reversed 😅

TestFoundation/main.swift
harlanhaskins: Added! Thanks.

harlanhaskins:
So what should be the behavior for `.description`? Should we truncate or leave as-is?

parkera:
Let's do what Philippe suggested and implement description to show everything and debugDescription to truncate.

harlanhaskins:
So, in my Xcode 7.1, `description` and `debugDescription` are exactly the same until `data.length` is `65526`, at which point it becomes `<OS_dispatch_data: data[0x7fb97b737320] = { leaf, size = 65527, buf = 0x1127c5000 }>`

parkera:
That makes sense - at a certain size NSData changes the subclass it returns to you (the magic of class clusters). At 64k it effectively becomes a dispatch_data_t.

Is your question asking what the limit should be or if we should have a limit at all?

harlanhaskins:
I guess this is a place where our implementation will diverge from Foundation?
If that's the case, what should the truncation limit be, and how much data should we put after the truncation point?

Is this a good truncation scheme?
```
<ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ... ffffffff>
```

parkera:
There are case-by-case places where we can allow the implementation to differ. We should have a discussion like this about it for sure, so thanks for double-checking on the Darwin implementation.

In this case in particular, I'm ok with making changes to the `debugDescription` method which are incompatible with the wrong way of doing things on any platform. If we leave `description` with the full data then we are fairly compatible. If anyone is using `debugDescription` and expecting it not to truncate, then we may have to ask them to fix their code instead.

My opinion on `debugDescription` truncation length is that it should be manageable if you `po` in the debugger. Let's say 1k. Anything after that should show the first 512 bytes and last 512 bytes.

harlanhaskins:
I agree with that. I'll implement that now! :+1:

krzyzanowskim:
- Regarding building hex string out of NSData, this is how I did it some time ago, with `String(format: String, ...)`.

 below is sample of what I came up with some time ago to generate hex out of bytes:

 ```swift
    private func hex(limit limit: Int? = nil) -> String {
        var str = String()
        var totalCount = 0
        self.enumerateByteRangesUsingBlock { (bytes, byteRange, stop) -> Void in
            for var idx = 0; idx < min(byteRange.length, limit ?? byteRange.length); ++idx {
                str += String(format:"%02x", UnsafePointer<UInt8>(bytes + idx).memory)
            }
            totalCount += byteRange.length
            stop.memory = totalCount > limit ? true : false
        }
        return str
    }
 ```

- My second thought here is about if buffer is contiguous. Presence of `enumerateByteRangesUsingBlock()` suggest it may be not if so I don't think we shoud have this assumption.

harlanhaskins:
That's a great implementation, but the truncation logic relies on being able to skip potentially large amounts of data and print the last 512 bytes.

harlanhaskins:
Any ideas here?

parkera:
@harlanhaskins Yah, let's go ahead and close this one and merge a new PR.

PR #46, opened 2015-12-06
NSURLError CustomStringConvertable conformance

dlbuckley:
I've added conformance to CustomStringConvertable to NSURL. I want to make sure that this is the intended support before adding additional support throughout the framework or if we are just going to rely on overriding the getter for the description property?

This is less about getting API parity with Foundation and more about making it bridge better with Swift.

### DIFF
--- a/Foundation/NSURL.swift
+++ b/Foundation/NSURL.swift
@@ -28,7 +28,7 @@ private func _standardizedPath(path: String) -> String {
     return path
 }
 
-public class NSURL : NSObject, NSSecureCoding, NSCopying {
+public class NSURL : NSObject, NSSecureCoding, NSCopying, CustomStringConvertible {
     typealias CFType = CFURLRef
     internal var _base = _CFInfo(typeID: CFURLGetTypeID())
     internal var _flags : UInt32 = 0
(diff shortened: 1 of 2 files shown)


Foundation/NSURL.swift
```diff
@@ -342,6 +342,14 @@ public class NSURL : NSObject, NSSecureCoding, NSCopying {
     override internal var _cfTypeID: CFTypeID {
         return CFURLGetTypeID()
     }
+    
+    override public var description: String {
+        if let absoluteString = self.absoluteString {
```
phausler: does this handle relative urls? i.e. ones created via:
```swift
public convenience init(fileURLWithPath path: String, relativeToURL baseURL: NSURL?)
```

TestFoundation/TestNSURL.swift
```diff
@@ -343,4 +344,11 @@ class TestNSURL : XCTestCase {
         */
     }
 
+    func test_description() {
+        // test with file that exists
+        let path = TestNSURL.gFileExistsPath
+        let url = NSURL(fileURLWithPath: path)
+        
+        XCTAssertEqual(path, url.description, "expected description does not match")
```
phausler: on the darwin version of Foundation NSURLs come out for their description as "relativeString -- baseURL" if the base url is set, else they are presented as "relativeString -- relativeString"

parkera:
Let's drop the conformance with `CustomStringConvertible` and just focus on adding the correct description method implementation. We should consider adoption of the CustomStringConvertible protocol at a more global scope than just NSURL (e.g. NSObject).

phausler:
Just as a note: NSObject now conforms to CustomStringConvertible so descendant classes do not need to do anything but override description

dlbuckley:
Good to know! I will make the amendments soon but I've been rammed this week so haven't had chance to sit down and do anything with it.

phausler:
Going to close this one since it has not had any updates to the requested changes in a bit, feel free to re-open when ready.

PR #47, opened 2015-12-06
NSGeometry: add unit tests

jessesquires:
This PR implements the following:

- `NSMake*` functions from `NSGeometry`
    - `NSMakePoint`
    - `NSMakeSize`
    - `NSMakeRect` 
- Simplify the default initializers to delegate to the memberwise initializers
- Unit tests for all additions/changes

I decided to make `CGFloat` implement `Comparable` mostly to simplify testing. (Since `CGFloat(1.0).native` became very verbose and repetitive.) However, I noticed that a PR for [conforming `NSDate` to `Comparable`](https://github.com/apple/swift-corelibs-foundation/pull/19) was declined because it was considered an API change. It's not clear if this is also considered an API change that necessitates following the *Swift evolution process* since `CGFloat` is more or less a a basic type? In any case, I'm happy to make any necessary changes.


jessesquires:
Oh I just noticed #35. Ha, I should have looked more closely. Sorry for the dupe. :smile: 

This could still be considered for the unit tests alone. (Or considered altogether since it is a smaller change.)

parkera:
I'm going to ask for #35 to be rebased, then we can incorporate your unit tests as a PR on top of that.

jessesquires:
Thanks @parkera !

Updated:

- rebased
- some formatting

My original tests don't cover everything added from #35, but I'll submit another PR for those to keep this one small. :smile:

PR #48, opened 2015-12-06
NSString: uppercaseString, lowercaseString, capitalizedString

krzyzanowskim:
Implementation of uppercaseString, lowercaseString, capitalizedString (with localized versions) for NSString. Tests for methods.

I tested it with this tests: https://github.com/apple/swift/blob/master/test/1_stdlib/NSStringAPI.swift and noticed strange behaviour. 

this test fail:
```
XCTAssertEqual(NSString(stringLiteral: "\u{0069}").uppercaseStringWithLocale(NSLocale(localeIdentifier: "tr")), "\u{0130}")
```

this don't
```
XCTAssertEqual(String(stringLiteral: "\u{0069}").uppercaseStringWithLocale(NSLocale(localeIdentifier: "tr")), "\u{0130}")
```

Difference is `NSString(stringLiteral:)` vs. `String(stringLiteral:)`. The result of uppercaseStringWithLocale is correct, though comparison is failing. This may be some issue with NSString.

### DIFF
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -32,6 +32,9 @@ class TestNSString : XCTestCase {
             ("test_FromNullTerminatedCStringInASCII", test_FromNullTerminatedCStringInASCII ),
             ("test_FromNullTerminatedCStringInUTF8", test_FromNullTerminatedCStringInUTF8 ),
             ("test_FromMalformedNullTerminatedCStringInUTF8", test_FromMalformedNullTerminatedCStringInUTF8 ),
+            ("test_uppercaseString", test_uppercaseString ),
+            ("test_lowercaseString", test_lowercaseString ),
+            ("test_capitalizedString", test_capitalizedString ),
         ]
     }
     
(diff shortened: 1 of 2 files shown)


TestFoundation/TestNSString.swift
```diff
@@ -127,4 +130,33 @@ class TestNSString : XCTestCase {
         let string = NSString(CString: bytes.map { Int8(bitPattern: $0) }, encoding: NSUTF8StringEncoding)
         XCTAssertNil(string)
     }
+
+    func test_uppercaseString() {
+        XCTAssertEqual(NSString(stringLiteral: "abcd").uppercaseString, "ABCD")
+        XCTAssertEqual(NSString(stringLiteral: "абВГ").uppercaseString, "АБВГ")
+        XCTAssertEqual(NSString(stringLiteral: "たちつてと").uppercaseString, "たちつてと")
+
+        // Special casting (see swift/validation-tests/stdlib/NSStringAPI.swift)
```
gribozavr: 'casing' was the right term to use here.

krzyzanowskim: you ask to me to remove tests here or fix typo?

gribozavr: Fixing the typo, of course :)

krzyzanowskim: :+1: it is late night ;)

krzyzanowskim:
If we acknowledge there is bug in CFString<->NSString bridging I'd change it to version with simple `CFStringCreateMutableCopy`, wdyt @phausler ?

phausler:
I think that perhaps the problem is actually __CFStrIsUnicode is reporting incorrectly. I will look into this today and see what we can do about it

phausler:
I was able to delve a bit deeper into the failure here, there are actually a few missteps that are happening that lead to this being incorrectly handled. The patch I have for this changes a few things and has a bit of replication here but I would still like to merge this in.

krzyzanowskim:
thanks, will update this PR

krzyzanowskim:
ok, tests still not pass but now it's with `CFStringCreateMutableCopy` as it suppose to work from the beginning.

krzyzanowskim:
@phausler I see you commited this: https://github.com/apple/swift-corelibs-foundation/commit/fdd15b5287f81810422869c5dc30a97257fb4a20 with almost the same changes, though lack of tests hide the issue with unicode characters.

btw. what's the advantage of 

```swift
let string = CFStringCreateMutable(kCFAllocatorSystemDefault, 0)
CFStringReplaceAll(string, self._cfObject)
```
over 
```swift
CFStringCreateMutableCopy(kCFAllocatorSystemDefault, 0, self._cfObject)
```
?

phausler:
There is little to no differential that was my implementation that i was using for testing, I am still going to merge this pr since the tests do hit the correct cases for non ascii characters.

PR #49, opened 2015-12-06
[Need feedback]: Implemented isEqualToDictionary

KaiCode2:
Implemented isEqualToDictionary and would like some feedback.

### DIFF
--- a/Foundation/NSDictionary.swift
+++ b/Foundation/NSDictionary.swift
@@ -282,10 +282,10 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
             return false
         }
         
-        for (key, value) in _storage {
+        for key in keyEnumerator() {
             if otherDictionary[key] == nil {
                 return false
-            } else if otherDictionary[key] != value {
+            } else if otherDictionary[key] != objectForKey(key) {
                 return false
             }
         }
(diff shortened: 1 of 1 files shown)


amorde:
:+1:

Foundation/NSDictionary.swift
```diff
@@ -278,7 +278,19 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
     public func descriptionWithLocale(locale: AnyObject?, indent level: Int) -> String { NSUnimplemented() }
     
     public func isEqualToDictionary(otherDictionary: [NSObject : AnyObject]) -> Bool {
-        NSUnimplemented()
+        if count != otherDictionary.count {
+            return false
+        }
+        
+        for (key, value) in _storage {
```
gribozavr: Is NSDictionary subclassable?  Then iterating over _storage wouldn't use the data from the subclass.

gribozavr:
It would be great if you included tests.

Foundation/NSDictionary.swift
amorde: I think the solution here would be to use `keyEnumerator()`

KaiCode2: @gribozavr Just corrected this issue using @amorde suggestion.

KaiCode2:
> It would be great if you included tests.

Absolutely, I'll incorporate the tests now 😊

phausler:
This does not handle the CF bridged dictionary case or subclasses; _storage is only available when it is a NSDictionary constructed from swift and not a custom subclass etc.

phausler:
Looks good to me; a great follow up for this would definitely be some unit tests. However since it looks pretty cut and dry merging it in.

KaiCode2:
@phausler O, thanks 😊

PR #50, opened 2015-12-06
implement NSProcessInfo.operatingSystemVersion & add NSProcessInfo tests

jpsim:


Foundation/NSProcessInfo.swift
```diff
@@ -102,7 +102,17 @@ public class NSProcessInfo : NSObject {
     
     public var operatingSystemVersion: NSOperatingSystemVersion {
         get {
-            NSUnimplemented()
+            let osVersionDictionary = _CFCopySystemVersionDictionary().takeRetainedValue()._swiftObject as! [String: AnyObject]
+            guard let productVersion = osVersionDictionary[_kCFSystemVersionProductVersionKey._swiftObject] as? String else {
+                fatalError("Could not read operating system version. Please verify that /System/Library/CoreServices/SystemVersion.plist is intact.")
```
phausler: I am not certain that this will be present on linux targets

Foundation/NSProcessInfo.swift
```diff
@@ -102,7 +102,17 @@ public class NSProcessInfo : NSObject {
     
     public var operatingSystemVersion: NSOperatingSystemVersion {
         get {
-            NSUnimplemented()
+            let osVersionDictionary = _CFCopySystemVersionDictionary().takeRetainedValue()._swiftObject as! [String: AnyObject]
+            guard let productVersion = osVersionDictionary[_kCFSystemVersionProductVersionKey._swiftObject] as? String else {
```
phausler: the two calls here are doing implicit conversions between object types and struct types which will not work on linux

jpsim:
I'll revisit this shortly, I should be able to address the linux issues without too much trouble.

jpsim:
@phausler I think I addressed your concerns about implicit conversion by `unsafeBitCast`ing CF-types to NS-types.

parkera:
Two comments, one small and one large.

The first is that we have the `._nsObject` and `._swiftObject` methods on CF types (like CFStringRef) to convert them. Or we can just use the CF API directly instead of doing conversion on the dictionary.

The larger comment is that I think this path is going to be wrong for Linux. Do we expect clients to stick a plist at `/S/L/CoreServices/SystemVersion.plist` still? Is there a better way to get an answer for this that is reasonable on Linux?

parkera:
Also, @phausler just pushed a commit to make the `takeRetainedValue` unnecessary (thanks to seeing it in this diff we realized that it could be better).

jpsim:
This is now using `CFDictionaryGetValue()`, delaying using the Swift object until as late as possible. I also removed the reference to `/S/L/CoreServices/SystemVersion.plist`. Let me know how you'd like to proceed for Linux.

jpsim:
I've rebased this again.

Foundation/NSProcessInfo.swift
```diff
@@ -102,7 +102,18 @@ public class NSProcessInfo : NSObject {
...
         get {
-            NSUnimplemented()
+            let productVersionKey = unsafeBitCast(_kCFSystemVersionProductVersionKey, UnsafePointer<Void>.self)
+            let productVersion = unsafeBitCast(CFDictionaryGetValue(_CFCopySystemVersionDictionary(), productVersionKey), NSString!.self)
+            if productVersion == nil {
+                fatalError("Could not read operating system version.")
+            }
+            let versionComponents = productVersion._swiftObject.characters.split(".").flatMap(String.init).flatMap({ Int($0) })
+            guard versionComponents.count >= 3 else {
+                fatalError("Invalid operating system version")
```
parkera: This line will fail on a major release, which only has two version numbers (e.g. 10.11).

parkera: Here's what Darwin does in this case, and I think we should use it instead of a fatalError in either the null result case for Linux or count <3:

majorVersion = result or -1
minorVersion = result or 0
patchVersion = result or 0

jpsim: Updated to match Darwin Foundation. Thanks for the tip!

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -227,6 +227,8 @@ extern void _CFDataInit(CFMutableDataRef memory, CFOptionFlags flags, CFIndex ca
 extern int32_t _CF_SOCK_STREAM();
 
 extern CFStringRef CFCopySystemVersionString(void);
+extern CFDictionaryRef _CFCopySystemVersionDictionary(void);
+extern const CFStringRef _kCFSystemVersionProductVersionKey;
```
parkera: These shouldn't be required - CFPriv.h has them and this file imports CFPriv.h.

TestFoundation/TestNSProcessInfo.swift
```diff
@@ -0,0 +1,35 @@
...
+            ("test_operatingSystemVersion", test_operatingSystemVersion ),
+        ]
+    }
+    
+    func test_operatingSystemVersion() {
+        let processInfo = NSProcessInfo.processInfo()
+        let versionString = processInfo.operatingSystemVersionString
+        let version = processInfo.operatingSystemVersion
+        let reconstructedVersionString = "Version \(version.majorVersion).\(version.minorVersion).\(version.patchVersion)"
+        XCTAssert(versionString.hasPrefix(reconstructedVersionString))
```
parkera: I think the best approach for this test is to just verify that the 3 strings are non-zero length (each).

Foundation/NSProcessInfo.swift
```diff
@@ -102,7 +102,19 @@ public class NSProcessInfo : NSObject {
     
     public var operatingSystemVersion: NSOperatingSystemVersion {
         get {
-            NSUnimplemented()
+            let productVersionKey = unsafeBitCast(_kCFSystemVersionProductVersionKey, UnsafePointer<Void>.self)
+            let productVersion = unsafeBitCast(CFDictionaryGetValue(_CFCopySystemVersionDictionary(), productVersionKey), NSString!.self)
+            if productVersion == nil {
+                fatalError("Could not read operating system version.")
```
parkera: We shouldn't use a fatalError here because it will crash on Linux (where this dictionary does not exist).

Instead let's use a major/minor/patch version of 0. Even better, we should figure out what value we want to return on Linux and calculate it here. Kernel version maybe?

parkera:
Once we've got last few details worked out, can you rebase this into a single commit? Thanks!

TestFoundation/TestNSProcessInfo.swift
jpsim: I don't follow. `NSOperatingSystemVersion` has 3 *integers*, not strings. Although I agree that this test risks being incorrect, say for `10.11` where the patch is omitted from the version string.

I guess we could test that `major + minor + patch > 0`?

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
jpsim: You're partly right. `CFPriv.h` declares `_kCFSystemVersionProductVersionKey` but not `_CFCopySystemVersionDictionary`, so I just removed the key from here.

Foundation/NSProcessInfo.swift
jpsim: > Instead let's use a major/minor/patch version of 0.

To stay consistent with Darwin Foundation, I opted to fall back on a major/minor/patch of -1/0/0.

> Even better, we should figure out what value we want to return on Linux and calculate it here. Kernel version maybe?

Maybe in another PR? I feel like this simple change has already taken much more back and forth than I was expecting.

TestFoundation/TestNSProcessInfo.swift
jpsim: I could also extract the version string parsing functionality and test *that*.

TestFoundation/TestNSProcessInfo.swift
```diff
@@ -21,10 +21,19 @@ class TestNSProcessInfo : XCTestCase {
     
     var allTests : [(String, () -> ())] {
         return [
+            ("test_operatingSystemVersion", test_operatingSystemVersion ),
             ("test_processName", test_processName ),
         ]
     }
     
+    func test_operatingSystemVersion() {
```
kovpas: This test crashes whole test set for me on linux:

```
Test Case 'TestNSProcessInfo.test_operatingSystemVersion' started.
TestFoundation: /media/psf/Home/Projects/Swift-Develop/swift/include/swift/Runtime/../../../stdlib/public/SwiftShims/RefCount.h:207: bool StrongRefCount::doDecrementShouldDeallocate() [ClearPinnedFlag = false]: Assertion `newval + quantum >= RC_ONE && "releasing reference with a refcount of zero"' failed.
```

PR #51, opened 2015-12-06
implement NSString.boolValue

jpsim:
Gives same output as Standard Foundation for the same test values.

### DIFF
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -21,6 +21,7 @@ class TestNSString : XCTestCase {
     
     var allTests : [(String, () -> ())] {
         return [
+            ("test_boolValue", test_boolValue ),
             ("test_BridgeConstruction", test_BridgeConstruction ),
             ("test_isEqualToStringWithSwiftString", test_isEqualToStringWithSwiftString ),
             ("test_isEqualToObjectWithNSString", test_isEqualToObjectWithNSString ),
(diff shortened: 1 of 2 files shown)


TestFoundation/TestNSString.swift
```diff
@@ -34,6 +35,17 @@ class TestNSString : XCTestCase {
             ("test_FromMalformedNullTerminatedCStringInUTF8", test_FromMalformedNullTerminatedCStringInUTF8 ),
         ]
     }
+
+    func test_boolValue() {
+        let trueStrings: [NSString] = ["t", "true", "TRUE", "tRuE", "yes", "YES", "1"]
```
phausler: the value "+000009" should map to true

TestFoundation/TestNSString.swift
```diff
@@ -34,6 +35,17 @@ class TestNSString : XCTestCase {
...
+        for string in trueStrings {
+            XCTAssert(string.boolValue)
+        }
+        let falseStrings: [NSString] = ["false", "FALSE", "fAlSe", "no", "NO", "0", "<true>", "_true"]
```
phausler: the value "-00000" should map to false

harlanhaskins:
The NSString documentation says:

```
This property is YES on encountering one of "Y", "y", "T", "t", or a digit 1-9—the method ignores any trailing characters. This property is NO if the receiver doesn’t begin with a valid decimal text representation of a number.

The property assumes a decimal representation and skips whitespace at the beginning of the string. It also skips initial whitespace characters, or optional -/+ sign followed by zeroes.
```

It looks like, instead of checking the prefix for each character, it walks the string and checks if any of the characters match the predicates.
But it should also skip `+` or `-` and any number of 0's.

jpsim:
Thanks for the tip @harlanhaskins! This one's on me, I should have referred to the documentation.

jpsim:
I have an implementation in place, but I also had to implement `NSScanner.rangeOfCharacterFromSet (...)` which I'll package up in another PR.

jpsim:
This has been updated to address @phausler's feedback and abide by the documentation referred to by @harlanhaskins. This depends on #75.

harlanhaskins:
Looks great! :+1:

jpsim:
Now that #75 has been merged, I've dropped that commit from this branch, so now this just has the `NSString.boolValue` commit.

parkera:
Thanks!

PR #52, opened 2015-12-06
[SR-71] Repair NSNumber behaviour

argon:
During initalization store the number type in the CFInfo bitfield in the
same way that CFNumberCreate does.

NSNumber/CFNumber stores its value in a UInt64 (_pad) with associated type
information as part of the CFInfo bitfield. NSNUmber Initialization is
performed through _CFNumberInit which copies the value bytes into _pad but
does not store the type information. This leads to broken behaviour when
converting between type repesentations. i.e. Double -> NSNumber ->
Integer.

Fixes: https://bugs.swift.org/browse/SR-71


parkera:
Looks great, thanks.

PR #54, opened 2015-12-07
NSJSONSerialization JSONObjectWithData: implementation

argon:
#### Overview

This is a fully-tested [RFC4627](http://www.ietf.org/rfc/rfc4627.txt)-compliant implementation of `JSONObjectWithData(_:options:)`.

I would like to request feedback about whether this might be accepted in the near future, or conflicts with other plans before I invest more time in improving it.

#### Still needed

* Documentation
* Consistency improvements - my understanding of parsers evolved with the project
* `NSJSONReadingAllowFragments` implementation
* `JSONObjectWithStream(_:options:)` implementation - depends on unimplemented `NSInputStream`

#### Notes

The underlying parser is implemented as an internal type which would be well suited to reuse in possible-future API changes.

* Trailing commas are allowed, to maintain compatibility with Darwin Foundation
* Errors are defined as a Swift enum and closely follow Darwin Foundation error conditions

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -87,7 +87,7 @@ public class NSJSONSerialization : NSObject {
 internal extension NSJSONSerialization {
     
     static func JSONObjectWithString(string: String) throws -> AnyObject {
-        let parser = JSONDeserializer.UnicodeParser(view: string.unicodeScalars)
+        let parser = JSONDeserializer.UnicodeParser(viewSkippingBOM: string.unicodeScalars)
         if let (object, _) = try JSONDeserializer.parseObject(parser) {
             return object
         }
@@ -102,6 +102,7 @@ internal extension NSJSONSerialization {
 
 internal extension NSJSONSerialization {
     
+    /// Detect the encoding format of the NSData contents
     class func detectEncoding(data: NSData) -> NSStringEncoding {
         let bytes = UnsafePointer<UInt8>(data.bytes)
         let length = data.length
@@ -174,7 +175,11 @@ private struct JSONDeserializer {
             self.index = index
         }
         
-        init(view: String.UnicodeScalarView) {
+        init(viewSkippingBOM view: String.UnicodeScalarView) {
+            if view.startIndex < view.endIndex && view[view.startIndex] == UnicodeScalar(65279) {
+                self.init(view: view, index: view.startIndex.successor())
+                return
+            }
             self.init(view: view, index: view.startIndex)
         }
         
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -30,6 +30,7 @@ enum NSJSONSerializationError: ErrorType {
     case UnterminatedString(String.UnicodeScalarIndex.Distance)
     case MissingObjectKey(String.UnicodeScalarIndex.Distance)
     case InvalidValue(String.UnicodeScalarIndex.Distance)
+    case InvalidEscapeSequence(String.UnicodeScalarIndex.Distance)
     case BadlyFormedArray(String.UnicodeScalarIndex.Distance)
     case UnexpectedEndOfFile
 }
@@ -258,7 +259,7 @@ private struct JSONDeserializer {
         static let Escape        = UnicodeScalar(0x5C) // \
     }
     
-    static func parseEscapedString(input: UnicodeParser) throws -> (UnicodeScalar, UnicodeParser)? {
+    static func parseEscapeSequence(input: UnicodeParser) throws -> (UnicodeScalar, UnicodeParser)? {
         guard let (scalar, parser) = readScalar(input) else {
             throw NSJSONSerializationError.UnexpectedEndOfFile
         }
@@ -300,9 +301,13 @@ private struct JSONDeserializer {
             case StringScalar.QuotationMark:
                 return (String(value), UnicodeParser(view: view, index: index))
             case StringScalar.Escape:
-                if let (escaped, newParser) = try parseEscapedString(UnicodeParser(view: view, index: index)) {
+                let parser = UnicodeParser(view: view, index: index)
+                if let (escaped, newParser) = try parseEscapeSequence(parser) {
                     value.append(escaped)
                     index = newParser.index
+                }
+                else {
+                    throw NSJSONSerializationError.InvalidEscapeSequence(parser.distanceFromStart - 1)
                 }
             default:
                 value.append(scalar)
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -227,25 +227,19 @@ private struct JSONDeserializer {
     }
     
     static func consumeScalar(scalar: UnicodeScalar, input: UnicodeParser) throws -> UnicodeParser? {
-        guard input.index < input.view.endIndex else {
+        switch takeScalar(input) {
+        case nil:
             throw NSJSONSerializationError.UnexpectedEndOfFile
-        }
-        if scalar == input.view[input.index] {
-            return input.successor()
-        }
-        return nil
-    }
-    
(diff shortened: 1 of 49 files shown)


Foundation/NSJSONSerialization.swift
```diff
@@ -53,7 +67,11 @@ public class NSJSONSerialization : NSObject {
        The data must be in one of the 5 supported encodings listed in the JSON specification: UTF-8, UTF-16LE, UTF-16BE, UTF-32LE, UTF-32BE. The data may or may not have a BOM. The most efficient encoding to use for parsing is UTF-8, so if you have a choice in encoding the data passed to this method, use UTF-8.
      */
     public class func JSONObjectWithData(data: NSData, options opt: NSJSONReadingOptions) throws -> AnyObject {
-        NSUnimplemented()
+        
+        guard let string = NSString(data: data, encoding: detectEncoding(data)) else {
+            throw NSJSONSerializationError.InvalidStringEncoding
+        }
+        return try JSONObjectWithString(string as String)
```
phausler: this is an implicit conversion that needs to be written as string._swiftObject

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
     }
 }
+
+//MARK: - Deserialization
+internal extension NSJSONSerialization {
+    
+    static func JSONObjectWithString(string: String) throws -> AnyObject {
+        let parser = JSONDeserializer.UnicodeParser(viewSkippingBOM: string.unicodeScalars)
+        if let (object, _) = try JSONDeserializer.parseObject(parser) {
+            return object
```
phausler: object in this case does not conform to AnyObject since it is an implicit conversion

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
+        }
+        else if let (array, _) = try JSONDeserializer.parseArray(parser) {
+            return array
```
phausler: also the same failure case of implicit conversions Array is not an AnyObject

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
+        var index = input.index
+        var value = String.UnicodeScalarView()
+        while index < endIndex && numberScalars.contains(view[index]) {
+            value.append(view[index])
+            index = index.successor()
+        }
+        guard value.count > 0, let result = Double(String(value)) else {
+            return nil
+        }
+        return (result, UnicodeParser(view: view, index: index))
```
phausler: Double is not an AnyObject and cannot be converted implicitly

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
+    }
+
+    //MARK: - Value parsing
+    static func parseValue(input: UnicodeParser) throws -> (AnyObject, UnicodeParser)? {
+        if let (value, parser) = try parseString(input) {
+            return (value, parser)
```
phausler: String is not an AnyObject and cannot be converted implicitly

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
+        }
+        else if let parser = try consumeSequence("true", input: input) {
+            return (true, parser)
```
phausler: Bool is not an AnyObject and cannot be converted implicitly

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
+        }
+        else if let parser = try consumeSequence("false", input: input) {
```
phausler: Bool is not an AnyObject and cannot be converted implicitly here as well.

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
+            return (false, parser)
+        }
+        else if let parser = try consumeSequence("null", input: input) {
+            return (NSNull(), parser)
+        }
+        else if let (object, parser) = try parseObject(input) {
+            return (object, parser)
```
phausler: For some reason this is coming out as a return type of ([String : AnyObject], JSONDeserializer.UnicodeParser)
which Dictionary cannot be converted implicitly either.

Foundation/NSJSONSerialization.swift
```diff
@@ -68,3 +86,399 @@ public class NSJSONSerialization : NSObject {
...
+        }
+        else if let (array, parser) = try parseArray(input) {
+            return (array, parser)
```
phausler: Array is not an AnyObject and cannot be converted implicitly

parkera:
Just wanted to note that I think this is a great start, and is going in the direction we want.

argon:
@phausler I hadn't considered the return type conversions. Having looked at it now, I suppose the question here is whether to implement `JSONDeserializer` as purely as possible in Swift - with `Bool`, `Double` and `String` types, then returning `Any` instead of `AnyObject`. The final conversion to `AnyObject` could be performed explicitly before outputting from `JSONObjectWithData` to conform to existing Foundation API.

phausler:
I would suspect that something along the lines of what NSPropertyListSerialization had to do will be required as well - see _expensivePropertyListConversion). The route of Any is a bit perilous because Dictionaries and have to then be Any to Any which Any is not Hashable. So some sort of container will have to be used that is Hashable (like the NSObject classes; they are already basically Hashable boxes)

argon:
I see now that `NSSwiftRuntime` has changed. I'm merging changes now and going to try converting `JSONObjectWithString` to return value types. I'll also try with object types as well to see which makes the most sense. There's lots of casting involved either way so it depends whether it's more valuable to be as "pure" as possible or to stick with object types given this is Foundation?

argon:
@phausler I have merged in master and the `NSJSONSerialization` test suite passes again. I have opted to use Value types internally for now as `hashValue` is not implemented on `NSString` - meaning it's impossible to make the Object tests pass.

I did look at implementing `hashValue` but I see there is already some work on `__NSCFType` so I figured it would be landing soon?

phausler:
hashValue for NSString should be relatively easy to implement; CFStringHashNSString is the thing that would do it and it just needs a few CF_SWIFT_CALLV additions.

argon:
I'll take a look at that then. I had assumed __NSCFType was intended to take care of it somehow but if NSString needs its own implementation I'll try to get it done

argon:
The dictionaries here are actually String: Any given the JSON format. That would solve the hashable problem, no?

TestFoundation/TestNSJSONSerialization.swift
```diff
@@ -0,0 +1,341 @@
...
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+
+#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
+    @testable import Foundation
```
phausler: `@testable` does not yet seem to be available on Linux targets

TestFoundation/main.swift
```diff
@@ -37,7 +37,9 @@ XCTMain([
     TestNSRange(),
     TestNSXMLParser(),
     TestNSDate(),
+    TestNSJSONSerialization(),
     TestNSData(),
     TestNSTimeZone(),
     TestNSScanner(),
+    TestNSJSONSerialization(),
```
thii: Duplicate.

PR #55, opened 2015-12-07
Update NSIndexSet enumeration snippets to Swift

christopherweems:
The code snippets in NSIndexSet's comments are currently written in Objective-C. Even though this code does not execute on Swift Foundation, I have tried it with Foundation on OS X in a playground and it executes as you would expect from the Obj-C snippet.


phausler:
Actually we should probably update the comments in the objective-c Foundation to have both objc and swift examples for not just this but all examples.

PR #56, opened 2015-12-07
[WIP] NSHTTPCookie initializer, getters, and tests

tptee:
Most of version 0 cookies are covered in the test cases with version 1 tests on the way. Still some edge cases to iron out, esp. when the dictionary key accepts multiple types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,5 @@ project.xcworkspace
 
 Build
 .build
+.configuration
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+# Contributing
+
+Contributions to Foundation are welcome! This project follows the [contribution guidelines for the Swift project](https://swift.org/contributing/#contributing-code). A few additional details are outlined below.
+
+## Licensing
+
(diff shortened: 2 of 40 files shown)


ChrisChares:
@tptee that pull request rebases your commits with all the recent updates to upstream/master and adds implementation + tests for NSHTTPCookie.requestHeadersForCookies().  Please take a look and incorporate the changes into this pull request :)

ChrisChares:
@salgernon #34 and this take a different approach to storing cookie properties.  This uses a private cookie struct while #34 passes all requests to the properties dictionary.  Which is preferred?

tptee:
My two cents is that we shouldn't be force casting anywhere. Best to remove all our optionals in the initializer, making the getters trivial to implement.

tptee:
The type signatures of the getters constricts you to non-optional, non-throwing code, so I think the only way to avoid the force casts is check the validity of the entire cookie in the initializer, where you do have the freedom to check optionals, return nil for an invalid cookie, etc

ChrisChares:
I'm for that as well.  It seems better to set properties with all their respective fallback rules at the same time as doing validation.  

The alternative means a simplified `init?()` method with property fallback and casting logic spread amongst all the getters.  Not only does that seem more complicated conceptually, it may also lead to code duplication as it seems you'll have to walk through the property fallbacks in `init?()` to ensure a valid cookie and again in the respective getter.

parkera:
In order to try and get us all on the same page, I've merged #34 and this in.

PR #57, opened 2015-12-07
Implement -[NSData writeToURL:options:error:]

lowell:
I feel like I should feel bad for submitting a PR during a weekend, but this is *fun*.

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -431,9 +431,27 @@ extension NSData {
         }
         return false
     }
-    
+
+    ///    Write the contents of the receiver to a location specified by the given file URL.
+    ///
+    ///    - parameter url:              The location to which the receiver’s contents will be written.
+    ///    - parameter writeOptionsMask: An option set specifying file writing options.
+    ///
+    ///    - throws: This method returns Void and is marked with the `throws` keyword to indicate that it throws an error in the event of failure.
+    ///
+    ///      This method is invoked in a `try` expression and the caller is responsible for handling any errors in the `catch` clauses of a `do` statement, as described in [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/ErrorHandling.html#//apple_ref/doc/uid/TP40014097-CH42) in [The Swift Programming Language](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/index.html#//apple_ref/doc/uid/TP40014097) and [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID10) in [Using Swift with Cocoa and Objective-C](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/index.html#//apple_ref/doc/uid/TP40014216).
     public func writeToURL(url: NSURL, options writeOptionsMask: NSDataWritingOptions) throws {
-        NSUnimplemented()
+        if let path = url.path {
+            do {
+                try writeToFile(path, options: writeOptionsMask)
+            } catch let error {
+                throw error
+            }
+        } else {
+            let userInfo = [NSLocalizedDescriptionKey : "The folder “\(url.path)” doesn’t exist.", // NSLocalizedString() not yet available
+                            NSURLErrorKey             : url.absoluteString ?? ""]
+            throw NSError(domain: NSCocoaErrorDomain, code: 4, userInfo: userInfo)
+        }
     }
     
     internal func enumerateByteRangesUsingBlockRethrows(block: (UnsafePointer<Void>, NSRange, UnsafeMutablePointer<Bool>) throws -> Void) throws {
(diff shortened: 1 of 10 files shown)


Foundation/NSData.swift
```diff
@@ -431,9 +431,27 @@ extension NSData {
...
+    ///
+    ///      This method is invoked in a `try` expression and the caller is responsible for handling any errors in the `catch` clauses of a `do` statement, as described in [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/ErrorHandling.html#//apple_ref/doc/uid/TP40014097-CH42) in [The Swift Programming Language](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/index.html#//apple_ref/doc/uid/TP40014097) and [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID10) in [Using Swift with Cocoa and Objective-C](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/index.html#//apple_ref/doc/uid/TP40014216).
     public func writeToURL(url: NSURL, options writeOptionsMask: NSDataWritingOptions) throws {
-        NSUnimplemented()
+        if let path = url.path {
+            do {
```
okla: I think there is no need for `do {} catch {}` blocks, you can simply use `try writeToFile(path, options: writeOptionsMask)`

Foundation/NSData.swift
```diff
@@ -431,9 +431,27 @@ extension NSData {
...
+                try writeToFile(path, options: writeOptionsMask)
+            } catch let error {
+                throw error
+            }
+        } else {
+            let userInfo = [NSLocalizedDescriptionKey : "The folder “\(url.path)” doesn’t exist.", // NSLocalizedString() not yet available
```
okla: `url.path` will be `nil` in this block

Foundation/NSData.swift
```diff
@@ -431,9 +431,27 @@ extension NSData {
...
+    ///
+    ///      This method is invoked in a `try` expression and the caller is responsible for handling any errors in the `catch` clauses of a `do` statement, as described in [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/ErrorHandling.html#//apple_ref/doc/uid/TP40014097-CH42) in [The Swift Programming Language](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/index.html#//apple_ref/doc/uid/TP40014097) and [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID10) in [Using Swift with Cocoa and Objective-C](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/index.html#//apple_ref/doc/uid/TP40014216).
     public func writeToURL(url: NSURL, options writeOptionsMask: NSDataWritingOptions) throws {
-        NSUnimplemented()
+        if let path = url.path {
```
johankool: Should this also check if the url gives back true for `isFileURL`? So that it for example would fail (as expected) for the url http://www.apple.com/some/path/writable/by/the/user?

Foundation/NSData.swift
```diff
@@ -431,9 +431,22 @@ extension NSData {
...
     public func writeToURL(url: NSURL, options writeOptionsMask: NSDataWritingOptions) throws {
-        NSUnimplemented()
+        guard let path = url.path where url.filePathURL == true else {
+            let userInfo = [NSLocalizedDescriptionKey : "The folder at “\(url)” does not exist or is not a file URL.", // NSLocalizedString() not yet available
```
johankool: It seems to me that the error message also isn't quite correct about the first part. You aren't checking whether a folder exist, but wether the URL given contains a path. (But I am not sure exactly what the existing Foundation framework does/reports in this case.)

phausler: if the url is a file url then it checks to see if the url has a path, if it does not it claims there is no such file (which is kind-of true-ish)
then it will funnel to writeToFile in the case of a file url (which has it's own error cases)
else it claims the url is an unsupported scheme.

Foundation/NSDecimalNumber.swift
```diff
@@ -9,10 +9,10 @@
...
 
 /***************	Exceptions		***********/
-public let NSDecimalNumberExactnessException: String = "" // NSUnimplemented
-public let NSDecimalNumberOverflowException: String = "" // NSUnimplemented
-public let NSDecimalNumberUnderflowException: String = "" // NSUnimplemented
-public let NSDecimalNumberDivideByZeroException: String = "" // NSUnimplemented
+public let NSDecimalNumberExactnessException: String = "NSDecimalNumberExactnessException"
+public let NSDecimalNumberOverflowException: String = "NSDecimalNumberOverflowException"
+public let NSDecimalNumberUnderflowException: String = "NSDecimalNumberUnderflowException"
+public let NSDecimalNumberDivideByZeroException: String = "NSDecimalNumberDivideByZeroException"
```
phausler: These are not exactly part of NSData's url writing, it would be good to split them out to their own commits

Foundation/NSRunLoop.swift
```diff
@@ -8,8 +8,8 @@
 //
 
 
-public let NSDefaultRunLoopMode: String = "" // NSUnimplemented
-public let NSRunLoopCommonModes: String = "" // NSUnimplemented
+public let NSDefaultRunLoopMode: String = "NSDefaultRunLoopMode"
+public let NSRunLoopCommonModes: String = "NSRunLoopCommonModes"
```
phausler: These probably will need to be the same as their CF counterparts when run loops will be supported.

Foundation/NSDecimalNumber.swift
lowell: This and the one below `NSRunLoop` one are from previously a merged commit (cdde78cb7562ab927feb630a7cf5405ff30bc848) from different contributor. There's one other  commit from someone else that got attached to this when I pushed, but that has since been merged as well (65cdfacd1a4ca55874df563364cf146062d02712). Should I just close this and resubmit so it's not as messy?

parkera:
@lowell I'll take you up on your offer to close this one and rebase so we don't get things really confused.

parkera:
Oops, didn't mean to close it myself. Close when you're ready.

lowell:
Thanks @parkera, that was considerate. :D Resubmitted.

PR #58, opened 2015-12-07
Fixes test case TestNSFileManger.test_fileAttributes

bolshedvorsky:
Check only for the existence of file owner's account (user) identifier instead of comparing its numeric value because
- user identifier of `0` are assigned to the superuser in linux
- negative value are valid account identifier. e.g. `nobody` is `-2` (See https://en.wikipedia.org/wiki/User_identifier)
- There's an assumption made here that account identifier are numeric which is reasonable for most system but there's a slim chance that this might not be the case.

Thanks to @lxcid

### DIFF
--- a/TestFoundation/TestNSFileManager.swift
+++ b/TestFoundation/TestNSFileManager.swift
@@ -121,7 +121,7 @@ class TestNSFileManger : XCTestCase {
             XCTAssertEqual(fileType!, NSFileTypeRegular)
             
             let fileOwnerAccountID = attrs[NSFileOwnerAccountID] as? NSNumber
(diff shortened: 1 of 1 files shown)


lxcid:
Thanks! :D

phausler:
The assumption would probably just mean that we would not run this test on platforms like windows if/when it is ported to there.
Otherwise looks pretty reasonable to me.

bolshedvorsky:
Current implementation of the API returns NSNumber value for the NSFileOwnerAccountID key. This unit test checks if we are getting an NSNumber value for this key.

eosrei:
Thanks for fixing quick @bartekchlebek !

PR #61, opened 2015-12-07
[Feedback] Initial NSByteCountFormatter implementation

bolshedvorsky:
I used NSDateFormatter and some other classes as a reference point. Shall I continue with the implementation?

### DIFF
--- a/Foundation/NSObject.swift
+++ b/Foundation/NSObject.swift
@@ -127,3 +127,9 @@ extension NSObject : Equatable, Hashable {
 public func ==(lhs: NSObject, rhs: NSObject) -> Bool {
     return lhs.isEqual(rhs)
 }
+
+extension NSObject : CustomDebugStringConvertible {
+}
+
+extension NSObject : CustomStringConvertible {
+}


Foundation/NSByteCountFormatter.swift
```diff
@@ -40,49 +40,130 @@ public enum NSByteCountFormatterCountStyle : Int {
...
+    }
+
     /* Convenience method on stringForObjectValue:. Convert a byte count into a string without creating an NSNumber.
     */
     public func stringFromByteCount(byteCount: Int64) -> String { NSUnimplemented() }
     
     /* Specify the units that can be used in the output. If NSByteCountFormatterUseDefault, uses platform-appropriate settings; otherwise will only use the specified units. This is the default value. Note that ZB and YB cannot be covered by the range of possible values, but you can still choose to use these units to get fractional display ("0.0035 ZB" for instance).
     */
-    public var allowedUnits: NSByteCountFormatterUnits
+    internal var _allowedUnits: NSByteCountFormatterUnits = .UseDefault
```
phausler: technically this pattern is not needed and you could write it as the simpler version like:
```swift
public var allowedUnits: NSByteCountFormatterUnits = .UseDefault
```

phausler:
The default values look correct to me.

PR #64, opened 2015-12-07
Tests for NSDate

syoutsey:
Added basic testing for `NSDate` class.


parkera:
I'm really happy to see PRs that add new tests; that's something we really need to get started. @syoutsey would you mind rebasing this on top of current master so we can do a clean merge?

PR #66, opened 2015-12-07
Adds an implementation of isEqual to the NSString class.

brownleej:
This was extracted from https://github.com/apple/swift-corelibs-foundation/pull/44. This adds an implementation of isEqual for NSString, which is necessary for initializing NSCalendar instances.


phausler:
looks pretty reasonable, build and passes on mac and linux

PR #67, opened 2015-12-07
Implement -[NSData writeToURL:options:error:]

lowell:
Cleaned up resubmission of #57, powered by awesome feedback™

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -441,9 +441,9 @@ extension NSData {
     ///
     ///      This method is invoked in a `try` expression and the caller is responsible for handling any errors in the `catch` clauses of a `do` statement, as described in [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/ErrorHandling.html#//apple_ref/doc/uid/TP40014097-CH42) in [The Swift Programming Language](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/index.html#//apple_ref/doc/uid/TP40014097) and [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID10) in [Using Swift with Cocoa and Objective-C](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/index.html#//apple_ref/doc/uid/TP40014216).
     public func writeToURL(url: NSURL, options writeOptionsMask: NSDataWritingOptions) throws {
-        guard let path = url.path where url.filePathURL == true else {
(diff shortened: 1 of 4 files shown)


Foundation/NSData.swift
```diff
@@ -431,9 +431,22 @@ extension NSData {
...
+    ///
+    ///      This method is invoked in a `try` expression and the caller is responsible for handling any errors in the `catch` clauses of a `do` statement, as described in [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/ErrorHandling.html#//apple_ref/doc/uid/TP40014097-CH42) in [The Swift Programming Language](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/index.html#//apple_ref/doc/uid/TP40014097) and [Error Handling](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID10) in [Using Swift with Cocoa and Objective-C](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/index.html#//apple_ref/doc/uid/TP40014216).
     public func writeToURL(url: NSURL, options writeOptionsMask: NSDataWritingOptions) throws {
-        NSUnimplemented()
+        guard let path = url.path where url.filePathURL == true else {
```
phausler: Building on linux i get this failure
```
Foundation/NSData.swift:444:45: error: cannot convert value of type 'NSURL?' to expected argument type 'Bool'
        guard let path = url.path where url.filePathURL == true else {
                                        ~~~~^~~~~~~~~~~
```
which is really strange since i don't get that on mac

Foundation/NSData.swift
```diff
@@ -431,9 +431,22 @@ extension NSData {
...
+            let userInfo = [NSLocalizedDescriptionKey : "The folder at “\(url)” does not exist or is not a file URL.", // NSLocalizedString() not yet available
```
phausler: unfortunately the userInfo dictionary for NSError needs to be [String: AnyObject] not [String: String] due to implicit conversions.
This can easily be fixed via just creating NSString objects from the values.

parkera:
This is almost ready, but there is one thing to resolve here. The `as NSString` cast fails on Linux without the ObjC bridging enabled:

```
Foundation/NSData.swift:445:57: error: cannot convert value of type 'String' to type 'NSString' in coercion
            let userInfo = [NSLocalizedDescriptionKey : "The folder at “\(url)” does not exist or is not a file URL." as NSString, // NSLocalizedString() not yet available
```

we could add the explicit bridge cast here, but actually I think that NSError's userInfo should be `[String : Any]` instead of `[String : AnyObject]`. This is consistent with the changes we've been making elsewhere so we can return `Swift.String` instead of `Foundation.NSString`, since the lack of bridging otherwise makes it difficult to use by the caller.

parkera:
I pushed some other fixes for NSError so that we could split that problem from this one.

PR #68, opened 2015-12-07
NSCopying for NSDictionary

MaddTheSane:
This patch implements `NSCopying` and `NSMutableCopying` protocols in `NSDictionary`, as well as creating a couple of test methods for verifying the behavior of the copy methods.

### DIFF
--- a/Foundation/NSDictionary.swift
+++ b/Foundation/NSDictionary.swift
@@ -162,6 +162,11 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
         } else if self.dynamicType === NSMutableDictionary.self {
             // Otherwise, create a new NSDictionary object
             
+            let newDict = NSDictionary()
+            newDict._storage = _storage
+            
+            return newDict
+        } else {
             // TODO: speed up?
             var keys = [AnyObject]()
             var values = [AnyObject]()
@@ -171,8 +176,6 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
             }
             
             return NSDictionary(objects: values, forKeys: keys as! [NSObject])
-        } else {
-            NSRequiresConcreteImplementation()
         }
     }
     
@@ -180,6 +183,11 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
         if self.dynamicType === NSDictionary.self || self.dynamicType === NSMutableDictionary.self {
             //ALWAYS create and return an NSMutableDictionary
             
+            let newDict = NSMutableDictionary()
+            newDict._storage = _storage
+            
+            return newDict
+        } else {
             // TODO: speed up?
             var keys = [AnyObject]()
             var values = [AnyObject]()
@@ -189,8 +197,6 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
             }
             
             return NSMutableDictionary(objects: values, forKeys: keys as! [NSObject])
-        } else {
-            NSRequiresConcreteImplementation()
         }
     }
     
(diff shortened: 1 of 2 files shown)


Foundation/NSDictionary.swift
```diff
@@ -156,13 +156,44 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
...
     
     public func copyWithZone(zone: NSZone) -> AnyObject {
-        NSUnimplemented()
+        if self.dynamicType === NSDictionary.self {
+            // NSDictionary is immutable; just return ourself
+            return self
+        } else if self.dynamicType === NSMutableDictionary.self {
+            // Otherwise, create a new NSDictionary object
+            
+            // TODO: speed up?
```
phausler: technically in this case you can construct a NSMutableDictionary and assign it's storage directly

phausler: technically you could create a NSDictionary by returning one created via the storage directly.

Foundation/NSDictionary.swift
```diff
@@ -156,13 +156,44 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
...
+            var keys = [AnyObject]()
+            var values = [AnyObject]()
+            for (key, value) in self {
+                keys.append(key)
+                values.append(value)
+            }
+            
+            return NSDictionary(objects: values, forKeys: keys as! [NSObject])
+        } else {
+            NSRequiresConcreteImplementation()
```
phausler: the intent of the NSMutableDictionary case you have written is probably correct for here.

Foundation/NSDictionary.swift
```diff
@@ -156,13 +156,44 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
...
+        }
     }
     
     public func mutableCopyWithZone(zone: NSZone) -> AnyObject {
-        NSUnimplemented()
+        if self.dynamicType === NSDictionary.self || self.dynamicType === NSMutableDictionary.self {
+            //ALWAYS create and return an NSMutableDictionary
+            
+            // TODO: speed up?
```
phausler: again you could copy the storage directly into a NSMutableDictionary since Dictionary is mutable

Foundation/NSDictionary.swift
```diff
@@ -156,13 +156,44 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
...
+            var keys = [AnyObject]()
+            var values = [AnyObject]()
+            for (key, value) in self {
+                keys.append(key)
+                values.append(value)
+            }
+            
+            return NSMutableDictionary(objects: values, forKeys: keys as! [NSObject])
+        } else {
+            NSRequiresConcreteImplementation()
```
phausler: again the previous branch is probably appropriate here.

Foundation/NSDictionary.swift
```diff
@@ -156,13 +156,50 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
...
     public func copyWithZone(zone: NSZone) -> AnyObject {
-        NSUnimplemented()
+        if self.dynamicType === NSDictionary.self {
+            // NSDictionary is immutable; just return ourself
+            return self
+        } else if self.dynamicType === NSMutableDictionary.self {
+            // Otherwise, create a new NSDictionary object
+            
+            let newDict = NSDictionary()
+            newDict._storage = _storage
```
phausler: generally I would prefer a better isolation of control however the compiler has some issues with NSDictionary.init(dictionary: [NSObject, AnyObject])

phausler:
the change looks good however this is causing a crash in the compiler for me
```
1.	While emitting SIL for 'copyWithZone' at Foundation/NSDictionary.swift:150:12
Segmentation fault (core dumped)
```

phausler:
This functionality has been merged in already that did not have the compiler crashes.

PR #69, opened 2015-12-08
NSUUID improvements

0xced:
* Implement `isEqual` and `hash`
* Do not crash when initialised with `NSUUID(UUIDBytes: nil)`
* Fix `UUIDString` implementation
* Implement a sensible `description`

### DIFF
--- a/Foundation/NSUUID.swift
+++ b/Foundation/NSUUID.swift
@@ -60,11 +60,11 @@ public class NSUUID : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
     
     public required init?(coder: NSCoder) {
-        
+        NSUnimplemented()
     }
     
     public func encodeWithCoder(aCoder: NSCoder) {
-        
+        NSUnimplemented()
     }
     
     public override func isEqual(object: AnyObject?) -> Bool {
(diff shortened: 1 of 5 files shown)


Foundation/NSUUID.swift
```diff
@@ -32,7 +32,11 @@ public class NSUUID : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
     
     public init(UUIDBytes bytes: UnsafePointer<UInt8>) {
-        memcpy(unsafeBitCast(buffer, UnsafeMutablePointer<Void>.self), UnsafePointer<Void>(bytes), 16)
+        if (bytes != nil) {
+            memcpy(unsafeBitCast(buffer, UnsafeMutablePointer<Void>.self), UnsafePointer<Void>(bytes), 16)
+        } else {
+            memset(unsafeBitCast(buffer, UnsafeMutablePointer<Void>.self), 0, 16)
```
phausler: actually the darwin implementation crashes when nil is passed so perhaps a precondition assertion would be more correct

phausler:
overall this looks good, the behavior of the nil initializer is a bit undefined on darwin so that is just a minor issue if any; once conflicts are resolved by a rebase this looks good to go to me

0xced:
I rebased and resolved conflicts. I did not change the check against nil, so the `init(UUIDBytes:)` function returns the Nil UUID (00000000-0000-0000-0000-000000000000) instead of crashing when passed a nil argument.

parkera:
Thanks!

PR #71, opened 2015-12-08
basic implementation for NSURLResponse init and tests

cullenbmacdonald:


cullenbmacdonald:
Does this look ok or should we swap the interface to use `public let`?

phausler:
looks good to me; after seeing the style of public private(set), I think perhaps it might be something we should adopt in places that don't have side effects or locks instead of using ivars

cullenbmacdonald:
Cool. So are you saying that this is the right implementation here?

phausler:
yea - it seems like a pretty reasonable pattern in many cases (not just this one)

PR #75, opened 2015-12-08
implement NSString.rangeOfCharacterFromSet

jpsim:


Foundation/NSString.swift
```diff
@@ -405,15 +405,39 @@ extension NSString {
...
     }
     
     public func rangeOfCharacterFromSet(searchSet: NSCharacterSet, options mask: NSStringCompareOptions) -> NSRange {
-        NSUnimplemented()
+        return rangeOfCharacterFromSet(searchSet, options: mask, range: NSMakeRange(0, length))
     }
     
     public func rangeOfCharacterFromSet(searchSet: NSCharacterSet, options mask: NSStringCompareOptions, range searchRange: NSRange) -> NSRange {
-        NSUnimplemented()
+        if mask.contains(.RegularExpressionSearch) {
```
phausler: from my understanding this is actually ignored for character set searches

Foundation/NSString.swift
```diff
@@ -405,15 +405,39 @@ extension NSString {
...
+            NSUnimplemented()
+        }
+        if searchRange.length == 0 {
+            return NSMakeRange(NSNotFound, 0)
+        }
+        
+#if os(Linux)
+        var cfflags = CFStringCompareFlags(mask.rawValue)
+        if mask.contains(.LiteralSearch) {
+            cfflags |= UInt(kCFCompareNonliteral)
```
phausler: probably these constants would be good to expose as a uniform re-export on OS X/iOS targets so that the linux (and potentially other platforms) would all have a similar interface.

Foundation/NSString.swift
```diff
@@ -405,15 +405,39 @@ extension NSString {
...
+        }
+#else
+        var cfflags = CFStringCompareFlags(rawValue: mask.rawValue)
+        if mask.contains(.LiteralSearch) {
+            cfflags.unionInPlace(.CompareNonliteral)
+        }
+#endif
+        var result = CFRangeMake(kCFNotFound, 0)
+
```
phausler: there probably should be a precondition here but overall the implementation looks good to me.

PR #76, opened 2015-12-08
Implement NSAffineTransform.transform{Point,Size} with tests

mattrajca:
As part of this change:

- The addition and multiplication operators have been implemented for `CGFloat` so we can compute dot products.
- A private helper function for matrix-vector products has been implemented.
- Tests for the identity transformation have been added. Once we support more operations (translations, rotations, and scales), we can add more tests.


parkera:
Cool, thanks.

PR #78, opened 2015-12-08
[NSTimeZone] Add implementation of abbreviation property and tests

thii:
- Add implementation of `NSTimeZone`'s `abbreviation` property getter.
- Add a test case for the above implementation.


phausler:
looks good to me

PR #80, opened 2015-12-08
Implementation of NSByteCountFormatter default values

bolshedvorsky:
Default values are added.
Unit test for the NSByteCountFormatter added.

Based on comments from the closed pull request #61

### DIFF
--- a/Foundation/NSByteCountFormatter.swift
+++ b/Foundation/NSByteCountFormatter.swift
@@ -40,49 +40,57 @@ public enum NSByteCountFormatterCountStyle : Int {
 }
 
 public class NSByteCountFormatter : NSFormatter {
+    public override init() {
+        super.init()
+    }
+    
     public required init?(coder: NSCoder) {
         NSUnimplemented()
     }
     
     /* Shortcut for converting a byte count into a string without creating an NSByteCountFormatter and an NSNumber. If you need to specify options other than countStyle, create an instance of NSByteCountFormatter first.
     */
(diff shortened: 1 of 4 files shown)


Foundation/NSByteCountFormatter.swift
```diff
@@ -40,49 +40,57 @@ public enum NSByteCountFormatterCountStyle : Int {
...
     public required init?(coder: NSCoder) {
         NSUnimplemented()
     }
     
     /* Shortcut for converting a byte count into a string without creating an NSByteCountFormatter and an NSNumber. If you need to specify options other than countStyle, create an instance of NSByteCountFormatter first.
     */
-    public class func stringFromByteCount(byteCount: Int64, countStyle: NSByteCountFormatterCountStyle) -> String { NSUnimplemented() }
-    
+    public class func stringFromByteCount(byteCount: Int64, countStyle: NSByteCountFormatterCountStyle) -> String {
+        let byteCountFormatter = NSByteCountFormatter()
```
segiddins: contradicts the docs, which say `without creating an NSByteCountFormatter`

bolshedvorsky: From my understating this is a convenience method, when you are getting a result without manual creation of the NSByteCountFormatter, however this method suppose to use the NSByteCountFormatter logic in order to create a formatted string.

segiddins: then the docs here need to be updated. I'd also be tempted to disassemble foundation and see if it's using a cached singleton there

parkera: It just calls into a private function which is shared with the formatter object.

PR #62, opened 2015-12-07
Initial implementation of NSNotification support

kovpas:

PR #82, opened 2015-12-08
Implement NSNotification

harlanhaskins:
I'd love some guidance with implementing NSNotificationCenter.

Namely, since we don't have CFNotificationCenter in this version of CoreFoundation, should it just be a dictionary mapping notification names/objects to observers?

How do we handle automatically deregistering?

Thanks!


kaandedeoglu:
NSNotificationCenter doesn't automatically deregister it's targets, you usually have to manually deregister it's target. The main challenge is calling a Swift function with just a string without the Objective-C runtime, I believe this is still achievable, but you might have to bridge with C (via dlsym)

harlanhaskins:
["In OS X 10.11 and iOS 9.0 NSNotificationCenter and NSDistributedNotificationCenter will no longer send notifications to registered observers that may be deallocated. If the observer is able to be stored as a zeroing-weak reference the underlying storage will store the observer as a zeroing weak reference, alternatively if the object cannot be stored weakly (i.e. it has a custom retain/release mechanism that would prevent the runtime from being able to store the object weakly) it will store the object as a non-weak zeroing reference."](https://developer.apple.com/library/prerelease/mac/releasenotes/Foundation/RN-Foundation/index.html#10_11NotificationCenter)

parkera:
Looks like we have some overlap between this PR and #62.

kaandedeoglu:
@harlanhaskins I must have missed it :), great new improvement.

harlanhaskins:
Beat me to it! Alright, I'll close this then.

PR #83, opened 2015-12-08
[NSScanner] Fix a bug in the initialization of _NSStringBuffer that makes NSScanner unable to find any number

thii:
Copied from commit message:

After writing a simple test case for `NSScanner`, I found that `NSScanner` always returns 0 when scanning numbers. The problem is in `_NSStringBuffer`'s init function, `curChar` was assigned to `buffer[0]` before `buffer` gets characters from scanned string. Changing `curChar` to an optional variable and assign it to `buffer[0]` after `buffer` got all characters from scanned string fixed this.


jpsim:
Once this is merged, e1c253f172d8f0a16bde67f6ebd253a0b7d034ed should be reverted since those tests should then pass.

phausler:
good catch! looks good to me

phausler:
for some reason the NSRange one is still failing for me likely we have another bug lurking around

PR #84, opened 2015-12-08
[NSString] Add implementation of `integerValue` and tests

thii:
This depends on the bug fix in #83 and should be merged after merging that PR.


phausler:
looks pretty reasonable to me, as a later addendum perhaps Int64 min and max might be interesting unit tests

thii:
Thank you. I included Int32/Int64 min and max unit tests in #85 and #86.

PR #85, opened 2015-12-08
[NSString] Add implementation of intValue and tests

thii:


Foundation/NSString.swift
```diff
@@ -457,7 +457,10 @@ extension NSString {
     
     public var intValue: Int32 {
         get {
-            NSUnimplemented()
+            let scanner = NSScanner(string: _swiftObject)
```
jpsim: with the experimental API currently implemented, this could just be one line: `return NSScanner(string: _swiftObject).scanInt() ?? 0`

thii: Thank you! I'm rebasing this and #86 now.

hartbit:
Perhaps add a test to validate the 0/failing path?

thii:
@hartbit I think there is no need since `test_integerValue()` contains most of similar tests and the logic is the same.

hartbit:
I still think this is useful. You never know when someone in the future might change the implementation of the two methods independently.

thii:
Will add more tests in another PR.

PR #86, opened 2015-12-08
[NSString] Add implementation of longLongValue and tests

thii:


hartbit:
Perhaps add a test to validate the 0/failing path?

phausler:
looks good to me, builds and passes on both mac and linux targets
per the commentary on unit tests : Foundation ❤️tests!

PR #88, opened 2015-12-08
Implementation of NSCountedSet

codestergit:
This pull request contains the implementation of `NSCountedSet`. It includes implementation and test cases for `NSCountedSet`. Please let me know if you need any changes in this. Suggestion are welcome. I am not that much expert with system implementation. So please let me know your views or any other changes required.
Thanks to swift team

### DIFF
--- a/TestFoundation/TestNSSet.swift
+++ b/TestFoundation/TestNSSet.swift
@@ -31,6 +31,11 @@ class TestNSSet : XCTestCase {
             ("test_equality", test_equality),
             ("test_copying", test_copying),
             ("test_mutableCopying", test_mutableCopying),
+            ("test_CountedSetBasicConstruction", test_CountedSetBasicConstruction),
+            ("test_CountedSetObjectCount", test_CountedSetObjectCount),
+            ("test_CountedSetAddObject", test_CountedSetAddObject),
+            ("test_CountedSetRemoveObject", test_CountedSetRemoveObject),
+            ("test_CountedSetCopying", test_CountedSetCopying)
         ]
     }
     
@@ -78,10 +83,9 @@ class TestNSSet : XCTestCase {
     }
     
     func test_setOperations() {
-        // TODO: This fails because hashValue and == use NSObject's implementaitons, which don't have the right semantics
-//        let set = NSMutableSet(array: ["foo", "bar"])
-//        set.unionSet(["bar", "baz"])
-//        XCTAssertTrue(set.isEqualToSet(["foo", "bar", "baz"]))
+        let set = NSMutableSet(array: ["foo", "bar"].bridge().bridge())
+        set.unionSet(["bar".bridge(), "baz".bridge()])
+        XCTAssertTrue(set.isEqualToSet(["foo".bridge(), "bar".bridge(), "baz".bridge()]))
     }
 
     func test_equality() {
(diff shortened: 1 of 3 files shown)


codestergit:
@parkera Updated pull request to resolve the conflicts. Please have a look.
Thanks

phausler:
unfortunately this causes a SIL gen crasher on linux:
1.	While silgen constructor initializer SIL function @_TFC10Foundation12NSCountedSetcfT5arrayGSaPs9AnyObject___S0_ for 'init' at Foundation/NSSet.swift:407:24

codestergit:
Thanks @phausler  . Do not have linux machine will try to arrange something and run the code.

phausler:
I am not certain on why, we have had a few of these beasts lurking under the waters around NSSet and NSDictionary specifically when it comes to implementing initializers that take arrays

codestergit:
@phausler  @parkera  Getting the errors while testing on ubuntu 15.10 http://pastebin.com/9b9ubyit . Any idea?

parkera:
You'll need to make sure your Linux machine has lots (> 8 GB) of RAM in order for linking to finish. If it fails like that, you can also just try again.

parkera:
Also, pass -R to the build script - the release version of clang/swift uses less memory to link.

codestergit:
Thanks @parkera !! Do not have that much RAM also running in VMWare :( . Will try to arrange somewhere else :(

codestergit:
@parkera @phausler  Updated the pull request to resolve compiler bug. Run all the tests on linux ubuntu 15.10 and Mac. 
Thanks

parkera:
The general idea here is ok, but I have one major concern. The implementation limits itself to holding subclasses of NSObject. This is actually not required even on Darwin:

```swift
class Foo { }

let f = Foo()
var x = NSCountedSet()
x.addObject(f)
x.addObject(f)
x.addObject(f)

x.count // 1
x.countForObject(f) // 3
```

This may be one area where we are again bumping up against the difficulties of lacking a bridging story while maintaining cross-platform source compatibility...

codestergit:
@parkera  @phausler Updated the pull request to resolve conflict! Verified test are passing. Please review it.
I think we need bridging with `NSObject `and as `NSMutableSet` and other classes also have this limitation. it is very hard to do make limitation to `AnyObject` as our storage `Set` needs NSObject or some `Hashable` implementation. We can not limit Set to Set<Hashable> as Hashable is not concrete protocol. I think with `type erasure` also our api user need to know about this.
I think this pull request is good for now. Please let me know your view.

codestergit:
@phausler @parkera Updated the pull request to resolve conflicts.  Please review the pull request. All tests are passing. Please let me know if any changes required.
Thanks

codestergit:
@phausler @parkera Updated the pull request to resolve conflicts. Please review the pull request. All tests are passing. Please let me know if any changes required.
Thanks

codestergit:
@phausler Updated the pull request. Resolved conflicts. Please Review it.
Thanks

codestergit:
@phausler It's pending for while. Can you please review it. If you need any changes please let me know.
Thanks

TestFoundation/TestNSSet.swift
```diff
@@ -140,4 +144,87 @@ class TestNSSet : XCTestCase {
         }
     }
 
-}
+    func test_CountedSetBasicConstruction() {
+        let v1 = "v1".bridge()
+        let v2 = "v2".bridge()
+        let v3asv1 = "v1".bridge()
+        let set = NSCountedSet()
+        let set2 = NSCountedSet(array: [v1, v1, v2,v3asv1].bridge().bridge())
```
phausler: the double bridge here is probably not needed since the elements in the Array are of type NSString which should fit as AnyObject.

phausler:
@parkera this looks good to me, do you have any outstanding issues here?

TestFoundation/TestNSSet.swift
codestergit: @phausler  Thanks. Done the changes.

codestergit:
@phausler  Updated the pull request to address the comments. Thanks for reviewing it.

parkera:
Thanks!

PR #89, opened 2015-12-08
NSComparisonResult.OrderedAscending should be equal to -1

akolov:
To be in sync with value in Foundation on Apple platforms


phausler:
lgtm

PR #92, opened 2015-12-09
Implemented various path and info-related methods on NSBundle

insidegui:


phausler:
after resolving the merge conflicts; this has an issue in the regards that it has implicit conversions 
```swift
return dictionary._swiftObject as? [String: AnyObject]
```
```
'[NSObject : AnyObject]' is not convertible to '[String : AnyObject]'
```

parkera:
I changed the return type of the infoDictionary methods in a commit yesterday. Please go ahead and rebase this on top of current master. Thanks!

PR #93, opened 2015-12-09
Implement NSAffineTransform

nubbel:
Hi,

The changes in this PR improve the implementation of the transformation methods. The new implementation makes use of the fact that a 2D affine transformation matrix is not an arbitrary 3x3 matrix (see code comments), and is both simpler and more efficient by skipping unnecessary computation steps.
Also included is:
- Implementation of `NSAffineTransformStruct.init()` which initializes all entries to zeros.
- Implementation of `NSAffineTransform.init(transform:)`
- Implementation of `NSCopying` methods for `NSAffineTransform`

### DIFF
--- a/Foundation/NSAffineTransform.swift
+++ b/Foundation/NSAffineTransform.swift
@@ -15,7 +15,12 @@ public struct NSAffineTransformStruct {
     public var m22: CGFloat
     public var tX: CGFloat
     public var tY: CGFloat
-    public init() { NSUnimplemented() }
+    
+    public init() {
+        (self.m11, self.m12, self.m21, self.m22) = (CGFloat(), CGFloat(), CGFloat(), CGFloat())
+        (self.tX, self.tY) = (CGFloat(), CGFloat())
+    }
+    
     public init(m11: CGFloat, m12: CGFloat, m21: CGFloat, m22: CGFloat, tX: CGFloat, tY: CGFloat) {
         (self.m11, self.m12, self.m21, self.m22) = (m11, m12, m21, m22)
         (self.tX, self.tY) = (tX, tY)
--- a/Foundation/NSAffineTransform.swift
+++ b/Foundation/NSAffineTransform.swift
@@ -33,7 +33,11 @@ public class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
         NSUnimplemented()
     }
     public func copyWithZone(zone: NSZone) -> AnyObject {
-        NSUnimplemented()
+        return NSAffineTransform(transform: self)
+    }
+    // Necessary because `NSObject.copy()` returns `self`.
+    public override func copy() -> AnyObject {
+        return copyWithZone(nil)
     }
     public required init?(coder aDecoder: NSCoder) {
         NSUnimplemented()
--- a/Foundation/NSAffineTransform.swift
+++ b/Foundation/NSAffineTransform.swift
@@ -132,9 +132,9 @@ private extension NSAffineTransformStruct {
      Creates an affine transformation matrix from translation values.
      The matrix takes the following form:
      
-     [ 1  0  tX ]
-     [ 0  1  tY ]
-     [ 0  0   1 ]
+         [ 1  0  tX ]
+         [ 0  1  tY ]
+         [ 0  0   1 ]
      */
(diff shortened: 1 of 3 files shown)


Foundation/NSAffineTransform.swift
```diff
@@ -15,7 +15,12 @@ public struct NSAffineTransformStruct {
     public var m22: CGFloat
     public var tX: CGFloat
     public var tY: CGFloat
-    public init() { NSUnimplemented() }
+    
+    public init() {
+        (self.m11, self.m12, self.m21, self.m22) = (CGFloat(), CGFloat(), CGFloat(), CGFloat())
+        (self.tX, self.tY) = (CGFloat(), CGFloat())
```
jessesquires: this `init` could (should?) just call the full memberwise init below :smile:

Foundation/NSAffineTransform.swift
```diff
@@ -63,42 +76,30 @@ public class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
...
-        let vector = Vector3(aPoint.x, aPoint.y, CGFloat(1.0))
-        let resultVector = multiplyMatrix3x3(matrix, byVector3: vector)
-        return NSMakePoint(resultVector.m1, resultVector.m2)
+        /**
+         [ x' ]     [ x ]   [ m11  m12  tX ] [ x ]   [ m11*x + m12*y + tX ]
+         [ y' ] = T [ y ] = [ m21  m22  tY ] [ y ] = [ m21*x + m22*y + tY ]
+         [  1 ]     [ 1 ]   [  0    0    1 ] [ 1 ]   [           1        ]
+         */
+        let x = transformStruct.m11*aPoint.x + transformStruct.m12*aPoint.y + transformStruct.tX
+        let y = transformStruct.m21*aPoint.x + transformStruct.m22*aPoint.y + transformStruct.tY
```
jessesquires: i think readability can be improved here with some spacing and `()` :smile: 

```swift
let x = (transformStruct.m11 * aPoint.x) + (transformStruct.m12 * aPoint.y) + transformStruct.tX
```

(same below)

Foundation/NSAffineTransform.swift
nubbel: Ah, you're right. Thanks!

nubbel:
Thank you, @jessesquires!
I've incorporated your feedback and updated the PR.

nubbel:
I've updated the PR to include the following changes:
- Implementation of translation, scaling and rotation methods with tests
- Implementation for `prependTransform` and `appendTransform` methods

Tests for composed transformations are missing - working on it!
Implementation for `invert` is still in progress, I'm not happy with it yet.

nubbel:
I've updated the PR to include the following changes:
- Implementation for `invert`
- Added arithmetic operators for `CGFloat` which were needed for the matrix inversion.

Notes:
- The original implementation throws an exception. My implementation uses `preconditionFailure` in this case.

parkera:
Be sure to update the Status doc when we finish a new feature!

https://github.com/apple/swift-corelibs-foundation/blob/master/Docs/Status.md

Foundation/NSAffineTransform.swift
```diff
@@ -38,67 +51,246 @@ public class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
...
+    /**
+     Creates an affine transformation matrix from rotation value (angle in radians).
+     The matrix takes the following form:
+     
+         [ cos α   -sin α  0 ]
+         [ sin α    cos α  0 ]
+         [   0        0    1 ]
+     */
+    static func rotation(radians angle: CGFloat) -> NSAffineTransformStruct {
+        let α = Double(angle)
```
parkera: Love the use of unicode variable name here!

nubbel:
@parkera Sorry, I didn't know about that doc. I updated it in #117.

Foundation/NSAffineTransform.swift
nubbel: Cool, glad you like it. I wasn't sure if that is considered good style, though.

parkera: I think it makes sense to use it in limited places like this, where it actually conveys some meaning.

PR #97, opened 2015-12-09
NSAffineTransform: implemented support for translation with tests

mattrajca:
- A helper function for multiplying two 3x3 matrices has been implemented. This will also be helpful for implementing scaling and rotation.
- An initializer for creating `NSAffineTransformStruct`s with matrices has been created for convenience.
- Translation has been implemented as multiplication of the current transformation matrix with a translation matrix.

### DIFF
--- a/Foundation/NSAffineTransform.swift
+++ b/Foundation/NSAffineTransform.swift
@@ -44,7 +44,14 @@ public class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
     }
     
     // Translating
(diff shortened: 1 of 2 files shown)


phausler:
I will trust that the maths on this are correct (it has been a while since i have had to do matrix math). Looks good to me, passes on darwin and linux.

nubbel:
@phausler, @mattrajca please have an my PR: https://github.com/apple/swift-corelibs-foundation/pull/93
It is now a complete implementation for NSAffineTransform.

PR #99, opened 2015-12-09
NSJSONSerialization isValidJSONObject Implementation

cezarywojcik:
Added implementation of `isValidJSONObject`.

Testing is currently difficult since `NSArray`, `NSString,` and `NSDictionary` are not yet fully implemented, so I tested this by writing tests using the existing Foundation classes in another project. I added these tests here and commented them out since they cannot yet be used and added a `TODO:` to enable the tests once ready.

Let me know what you think!

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -28,20 +28,66 @@ public struct NSJSONWritingOptions : OptionSetType {
 /* A class for converting JSON to Foundation/Swift objects and converting Foundation/Swift objects to JSON.
    
    An object that may be converted to JSON must have the following properties:
-    - Top level object is an Array or Dictionary
-    - All objects are String, NSNumber, Array, Dictionary, or NSNull
-    - All dictionary keys are Strings
-    - NSNumbers are not NaN or infinity
+    - Top level object is a `Swift.Array` or `Swift.Dictionary`
+    - All objects are `Swift.String`, `Foundation.NSNumber`, `Swift.Array`, `Swift.Dictionary`,
+      or `Foundation.NSNull`
+    - All dictionary keys are `Swift.String`s
+    - `NSNumber`s are not NaN or infinity
 */
 
 public class NSJSONSerialization : NSObject {
     
-    /* Returns YES if the given object can be converted to JSON data, NO otherwise.
-    
-    Other rules may apply. Calling this method or attempting a conversion are the definitive ways to tell if a given object can be converted to JSON data.
+    /* Determines whether the given object can be converted to JSON.
+       Other rules may apply. Calling this method or attempting a conversion are the definitive ways
+       to tell if a given object can be converted to JSON data.
+       - parameter obj: The object to test.
+       - returns: `true` if `obj` can be converted to JSON, otherwise `false`.
      */
-    public class func isValidJSONObject(obj: AnyObject) -> Bool {
-        NSUnimplemented()
+    public class func isValidJSONObject(obj: Any) -> Bool {
+        // TODO: - revisit this once bridging story gets fully figured out
+        func isValidJSONObjectInternal(obj: Any) -> Bool {
+            // object is Swift.String or NSNull
+            if obj is String || obj is NSNull {
+                return true
+            }
+
+            // object is NSNumber and is not NaN or infinity
+            if let number = obj as? NSNumber {
+                let invalid = number.doubleValue.isInfinite || number.doubleValue.isNaN
+                    || number.floatValue.isInfinite || number.floatValue.isNaN
+                return !invalid
+            }
+
+            // object is Swift.Array
+            if let array = obj as? [Any] {
+                for element in array {
+                    guard isValidJSONObjectInternal(element) else {
+                        return false
+                    }
+                }
+                return true
+            }
+
+            // object is Swift.Dictionary
+            if let dictionary = obj as? [String: Any] {
+                for (_, value) in dictionary {
+                    guard isValidJSONObjectInternal(value) else {
+                        return false
+                    }
+                }
+                return true
+            }
+
+            // invalid object
+            return false
+        }
+
+        // top level object must be an Swift.Array or Swift.Dictionary
+        guard obj is [Any] || obj is [String: Any] else {
+            return false
+        }
+
+        return isValidJSONObjectInternal(obj)
     }
     
     /* Generate JSON data from a Foundation object. If the object will not produce valid JSON then an exception will be thrown. Setting the NSJSONWritingPrettyPrinted option will generate JSON with whitespace designed to make the output more readable. If that option is not set, the most compact possible JSON will be generated. If an error occurs, the error parameter will be set and the return value will be nil. The resulting data is a encoded in UTF-8.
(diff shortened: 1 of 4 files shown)


Foundation/NSJSONSerialization.swift
```diff
@@ -35,12 +35,48 @@ public struct NSJSONWritingOptions : OptionSetType {
...
+            if obj is String || obj is NSNull {
+                return true
+            }
+
+            // object is NSNumber and is not NaN or infinity
+            if let number = obj as? NSNumber {
+                if number.doubleValue.isInfinite || number.doubleValue.isNaN
+                    || number.floatValue.isInfinite || number.floatValue.isNaN {
+                    return false
+                }
```
radex: This would would probably be better implemented using `guard` to guarantee the early exit

Foundation/NSJSONSerialization.swift
```diff
@@ -35,12 +35,48 @@ public struct NSJSONWritingOptions : OptionSetType {
...
+                return true
+            }
+
+            // object is Array
+            if let array = obj as? Array<AnyObject> {
+                return array.filter({ isValidJSONObjectInternal($0) }).count == array.count
+            }
+
+            // object is Dictionary
+            if let dictionary = obj as? Dictionary<String, AnyObject> {
```
radex: Why not `[AnyObject]` and `[String: AnyObject]` ?

Foundation/NSJSONSerialization.swift
radex: Actually, scratch that, I would just write inside the outer if let:

```
let invalid = number.doubleValue.isInfinite || number.doubleValue.isNaN || ...
return !invalid
```

Foundation/NSJSONSerialization.swift
```diff
@@ -35,12 +35,48 @@ public struct NSJSONWritingOptions : OptionSetType {
...
+       Other rules may apply. Calling this method or attempting a conversion are the definitive ways
+       to tell if a given object can be converted to JSON data.
+       - parameter obj: The object to test.
+       - returns: `true` if `obj` can be converted to JSON, otherwise `false`.
      */
     public class func isValidJSONObject(obj: AnyObject) -> Bool {
-        NSUnimplemented()
+        func isValidJSONObjectInternal(obj: AnyObject) -> Bool {
+            // object is String or NSNull
+            if obj is String || obj is NSNull {
```
phausler: if obj is AnyObject it can never be String; it would be NSString in this case

Foundation/NSJSONSerialization.swift
```diff
@@ -35,12 +35,48 @@ public struct NSJSONWritingOptions : OptionSetType {
...
+            if let number = obj as? NSNumber {
+                if number.doubleValue.isInfinite || number.doubleValue.isNaN
+                    || number.floatValue.isInfinite || number.floatValue.isNaN {
+                    return false
+                }
+                return true
+            }
+
+            // object is Array
+            if let array = obj as? Array<AnyObject> {
```
phausler: obj can never be Array if it is AnyObject it would be NSArray

Foundation/NSJSONSerialization.swift
phausler: obj can never be Dictionary either, it would be NSDictionary if obj is AnyObject

cezarywojcik:
Addressed review comments.

Quick question about `Array`/`Dictionary`/`String` vs `NSArray`/`NSDictionary`/`NSString` though - aren't they interoperable? When I was testing this against the existing Foundation classes, it was working as intended. For example:

```
let arr = NSArray(array: [1])
arr is [AnyObject] // true
let dict = NSDictionary(dictionary: ["string" : 1])
dict is [String : AnyObject] // true
```

phausler:
code like this:
let arr = NSArray(array: [1])
arr is [AnyObject] // true
are implicit type conversions from the objc bridging code in the swift standard library and the compiler which do not exist on all platforms: see https://github.com/apple/swift-corelibs-foundation/blob/master/Docs/Issues.md

cezarywojcik:
Gotcha, thanks for the clarification.

cezarywojcik:
This means that I can't use the filter function as well, correct? I'll update that.

phausler:
you can if you use `_swiftObject` or `.bridge()`; but remember any bridges except the conversion between CF and NS types cost at least a function call overhead, sometimes they require a O(n) or worse traversal

cezarywojcik:
Makes more sense to avoid bridging if possible then. Updated to account for this.

parkera:
This one is tricky, but I think we need to change this API to take `[String:Any]`. A user of this API in Swift should be able to pass in `Swift.String` objects, and without the implicit conversion from `Swift.String` to `Foundation.NSString`, it would be difficult to use the method.

cezarywojcik:
I'm a bit confused on what you mean. If we pass in an `AnyObject`, then we can only have an `NSDictionary` passed in, which, in turn means that the keys are also all `AnyObject`s, so the keys couldn't be a `Swift.String` without implicit conversion, as you said.

Are you saying that the function signature should take in an `Any` instead of an `AnyObject` and test for both `Foundation.NSString, Foundation.NSArray, ...` types as well as `Swift.String, Swift.Array...`?

parkera:
The assumption we've decided to make here is that, in the absence of implicit bridging, we have to assume that clients are using the Swift collection and String types instead of the Foundation ones. Hopefully, once we get our bridging story fully figured out here, we can accept either String or NSString.

cezarywojcik:
Updated to check for `String`, `Array<Any>`, and `Dictionary<String, Any>` in place of `NSString`, `NSArray`, and `NSDictionary`.

Changed function signature of `isValidJSONObject` to take `Any` as input instead of `AnyObject` to allow for this.

Uncommented and reformatted tests.

Added `TODO:` to revisit the method once bridging gets fully figured out.

Let me know if this is in line with what you were thinking.

phausler:
```swift
func foo(t: AnyObject) { }
foo(nil)
```
on darwin results in `error: nil is not compatible with expected argument type 'AnyObject'`

So I would be inclined to claim that NSNull is more appropriate to represent nil

cezarywojcik:
Right, an optional type would probably have to be used, but since only the internal recursive method would be checking for `nil`/`NSNull` (since top-level objects can only be dictionaries or arrays), only the internal method's signature would have to change to support `nil`, so the actual `isValidJSONObject` would still only accept non-optional input.

What I mean:
```
public class func isValidJSONObject(obj: Any) -> Bool {
     func isValidJSONObjectInternal(obj: Any?) -> Bool {
                                            ^
...
```

parkera:
The current NSJSONSerialization only uses NSNull to represent nil. I don't think that we bridge a nil optional to NSNull in Objective-C mode, so if we accept an optional type here then I suspect we lose our cross-platform compatibility...

I think it makes total sense to use optionals for this (something we obviously could not do in Objective-C). But that would be an API change, so we'd have to discuss it separately.

cezarywojcik:
That makes sense - it's definitely something that I think is worth looking into in the long run.

Is there anything else that's blocking a merge for this?

phausler:
Not really, just need to finish running tests and verify a good state before merging anything else in this morning (last night there were some test breaks that got merged in that I am still trying to account for)

cezarywojcik:
Sounds good. As a quick related note, currently on master, in `main.swift`, `TestNSJSONSerialization` appears twice, but this PR removes one of those instances.

cezarywojcik:
The spec for the function says that any `NSNumber` that is not NaN or Infinity should be accepted. I don't think that we can make assumptions about what kind of value this is. Since this needs to be cross-platform, I think that it's better to check both. Feel free to submit another PR that removes the `floatValue` checks if you think that it doesn't matter - can get feedback about this in a new PR.

phausler:
The more tests we have the better even if they are slightly redundant; if anything it ensures consistency with little cost since these tests are rather short in execution time.

PR #100, opened 2015-12-09
NSAffineTransform: implemented support for scaling with tests

mattrajca:
Now that we have a generic matrix multiplication function, this becomes trivial. I have also verified these tests against the `NSAffineTransform` in the 10.11 SDK to ensure consistent results (e.g. order of oeprations).


nubbel:
Hi @mattrajca,
This is already implemented in https://github.com/apple/swift-corelibs-foundation/pull/93. 

It would be great in you could review it and give feedback.

PR #102, opened 2015-12-09
Always close file descriptor (eventually runs out of them)

johnno1962:
File descriptor is not closed if no error occurs. Eventually they get exhausted after 1000 NSData( contentsOfFile: path )


phausler:
good catch looks fine to me

PR #104, opened 2015-12-09
NSMutableURLRequest implementation

ChrisChares:
+ Basic initialization
+ Getters / setters
+ Set / add request header functions
+ Tests

I didn't implement this to avoid changing the interface, but I think it would make sense going forward to change NSURLRequest's init(NSURL) method from convenience to required and mark the URL property as non-optional.

### DIFF
--- a/Foundation/NSURLRequest.swift
+++ b/Foundation/NSURLRequest.swift
@@ -389,7 +389,13 @@ public class NSMutableURLRequest : NSURLRequest {
         if _httpHeaderFields == nil {
             _httpHeaderFields = [:]
         }
-        _httpHeaderFields?[field.lowercaseString] = value
+        if let existingHeader = _httpHeaderFields?.filter({ (existingField, _) -> Bool in
+            return existingField.lowercaseString == field.lowercaseString
+        }).first {
+            let (existingField, _) = existingHeader
+            _httpHeaderFields?.removeValueForKey(existingField)
+        }
+        _httpHeaderFields?[field] = value
     }
     
     /*! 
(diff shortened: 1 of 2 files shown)


Foundation/NSURLRequest.swift
```diff
@@ -348,7 +385,12 @@ public class NSMutableURLRequest : NSURLRequest {
         @param value the header field value. 
         @param field the header field name (case-insensitive). 
     */
-    public func setValue(value: String?, forHTTPHeaderField field: String) { NSUnimplemented() }
+    public func setValue(value: String?, forHTTPHeaderField field: String) {
+        if _httpHeaderFields == nil {
+            _httpHeaderFields = [:]
+        }
+        _httpHeaderFields?[field.lowercaseString] = value
```
johankool: The Obj-C Foundation framework preserves the capitalization as entered (as seen in what is returned from `allHTTPHeaderFields`), though is otherwise indeed case insensitive. Keeping the capitalization as is helps with readability during debugging network calls.

Foundation/NSURLRequest.swift
```diff
@@ -308,15 +308,39 @@ public class NSURLRequest : NSObject, NSSecureCoding, NSCopying, NSMutableCopyin
...
+    /*!
+     @method initWithURL:
+     @abstract Initializes an NSMutableURLRequest with the given URL.
+     @discussion Default values are used for cache policy
+     (NSURLRequestUseProtocolCachePolicy) and timeout interval (60
+     seconds).
+     @param URL The URL for the request.
+     @result An initialized NSMutableURLRequest.
+     */
+    public init(URL: NSURL) {
```
phausler: Doesn't the superclass handle this?

ChrisChares:
Updated code to store the values with supplied case but evaluate them for setting / adding as insensitive.

The API is a little strange since it provides case insensitive setters but no such getters.

Foundation/NSURLRequest.swift
ChrisChares: It doesn't and I'm not sure why.  NSMutableURLRequest implements all of NSURLRequest's designated initializers ( init(coder) ), yet isn't inheriting NSURLRequest's convenience initializers.

phausler: TIL: yea interesting - it does not seem to have the funnel behavior i would have guessed either. In short this seems fine to me

ChrisChares: Hey sorry I think I deleted that comment too quick, didn't know you'd get to it that fast.  NSMutableURLRequest wasn't overriding NSURLRequest's private blank initializer, so the convenience initializers weren't being passed down

PR #107, opened 2015-12-10
Added test for NSArray

joshuatbrown:
Just added a test for the `objectAtIndex` method.


phausler:
looks reasonable to me

joshuatbrown:
That makes sense - I'll make the change and update the PR. Thanks for the feedback!

joshuatbrown:
OK, this is ready for another review. I bridged the Strings to NSStrings - I hope that's what you meant, @phausler.

joshuatbrown:
Is there anything else I can do to move this along? I'm happy to make further changes if needed.

PR #108, opened 2015-12-10
implement the init(contentsOfFile) method in NSString

zixun:
implement the init(contentsOfFile) method in NSString


zixun:
CFStringCreateWithCString is return String!, so I delete the guard of it

Foundation/NSString.swift
```diff
@@ -1237,9 +1237,7 @@ extension NSString {
     }
     
     public convenience init?(CString nullTerminatedCString: UnsafePointer<Int8>, encoding: UInt) {
-        guard let cf = CFStringCreateWithCString(kCFAllocatorSystemDefault, nullTerminatedCString, CFStringConvertNSStringEncodingToEncoding(encoding)) else {
-            return nil
-        }
+        let cf = CFStringCreateWithCString(kCFAllocatorSystemDefault, nullTerminatedCString, CFStringConvertNSStringEncodingToEncoding(encoding))
```
phausler: CFStringCreateWithCString can and will return NULL when it cannot convert the cString to the requested encoding. So the call later to cf._nsObject will fail

zixun: @phausler  thx,i got it..

Foundation/NSString.swift
```diff
@@ -1253,7 +1253,20 @@ extension NSString {
     }
     
     public convenience init(contentsOfFile path: String, encoding enc: UInt) throws {
-        NSUnimplemented()    
+        let readResult = try NSData.readBytesFromFileWithExtendedAttributes(path, options: [])
+        guard let cf = CFStringCreateWithBytes(kCFAllocatorDefault, UnsafePointer<UInt8>(readResult.bytes), readResult.length, CFStringConvertNSStringEncodingToEncoding(enc), true) else {
+            throw NSError(domain: NSCocoaErrorDomain, code: NSCocoaError.CoderReadCorruptError.rawValue, userInfo: [
+                "NSDebugDescription" : "Unable to Creates a string from a buffer containing characters in a specified encoding."
```
johankool: Alternative error description suggestion: Unable to create a string using the specified encoding.

zixun: @johankool thank you~ I changed it

phausler:
Looking at this a bit further in comparison to the darwin behavior there are a few key notes:
this method is intended to read the file (mapped) 
then it will check the encoding and determine if it looks like unicode or UTF8 bom
If it cannot determine if it is unicode or UTF8, it should then fallback to the default string encoding.
and finally it should call out to initWithData:encoding:

joshuatbrown:
@zixun Would it be possible to add a test for this?

zixun:
@phausler  did you mean like this:
  
        guard let data = NSData(contentsOfFile: path) else {
            throw NSError(domain: NSCocoaErrorDomain, code: NSCocoaError.CoderReadCorruptError.rawValue, userInfo: [
                "NSDebugDescription" : "data in contentsOfFile is nil."
                ])
        }

        var encode = enc
        if encode != NSUTF8StringEncoding || encode != NSUnicodeStringEncoding {
            encode = String.defaultCStringEncoding()
        }
        self.init(data: data, encoding: encode)

zixun:
@joshuatbrown  I will try it

zixun:
@phausler  
  
     self.init(data: data, encoding: encode)

it maybe return nil,  how could I catch it while initWithData:encoding: is return nil and throw it out?

phausler:
I think what you have is fine; I will follow up with a slight correction to the error codes to match more similarly what happens on Darwin.

zixun:
@phausler OK, thank you~

zixun:
hi @joshuatbrown  please review my test code

TestFoundation/TestNSString.swift
```diff
@@ -227,6 +228,18 @@ class TestNSString : XCTestCase {
         let string = NSString(CString: bytes.map { Int8(bitPattern: $0) }, encoding: NSUTF8StringEncoding)
         XCTAssertNil(string)
     }
+    
+    func test_FromContentOfFile() {
+        let testFilePath = testBundle().pathForResource("NSStingTestData", ofType: "txt")
```
joshuatbrown: There appears to be a typo in the path name here; shouldn't it be `NSStringTestData`? Does this test pass when you run the test suite?

zixun: @joshuatbrown ,Im sorry for this,it's my stupid mistake~now,I fixed it,but it cant build while I run my test now, so I merge the newest code to my fork branch,but also not work, the error is in the 246 line of NSHTTPCookie.swift, error message is 

       /swift-corelibs-foundation/Foundation/NSHTTPCookie.swift:246:13: 
       All stored properties of a class instance must be initialized before returning nil from an initializer

zixun: @joshuatbrown  my xcode is 7.2

TestFoundation/TestNSString.swift
```diff
@@ -228,6 +229,18 @@ class TestNSString : XCTestCase {
         let string = NSString(CString: bytes.map { Int8(bitPattern: $0) }, encoding: NSUTF8StringEncoding)
         XCTAssertNil(string)
     }
+    
+    func test_FromContentOfFile() {
+        let testFilePath = testBundle().pathForResource("NSStringTestData", ofType: "txt")
```
phausler: this returns an optional

TestFoundation/TestNSString.swift
```diff
@@ -228,6 +229,18 @@ class TestNSString : XCTestCase {
...
+        
+        do {
+            let str = try NSString(contentsOfFile: testFilePath, encoding: NSUTF8StringEncoding)
```
phausler: this does not take optional

zixun: @phausler  thank you,I fix it at ccommit of 7c9d732. I cant build the test because of the eroor mentioned before ,even I merge the newest code to my fork branch . the error is in the 246 line of NSHTTPCookie.swift, error message is
  
     /swift-corelibs-foundation/Foundation/NSHTTPCookie.swift:246:13: 
    All stored properties of a class instance must be initialized before returning nil from an initializer

phausler:
seems to look good on linux, I will fix up a few error codes to be more congruent with the darwin version as a follow-up commit

PR #110, opened 2015-12-10
Ditch "-> ()" in favor of "-> Void"

thii:
This change should be done now in early stage, to prevent new code keep following the old convention.

### DIFF
--- a/TestFoundation/TestNSAffineTransform.swift
+++ b/TestFoundation/TestNSAffineTransform.swift
@@ -26,7 +26,7 @@
 class TestNSAffineTransform : XCTestCase {
     private let accuracyThreshold = 0.001
 
-    var allTests : [(String, () -> ())] {
+    var allTests : [(String, () -> Void)] {
         return [
             ("test_BasicConstruction", test_BasicConstruction),
             ("test_IdentityTransformation", test_IdentityTransformation),
--- a/TestFoundation/TestNSArray.swift
+++ b/TestFoundation/TestNSArray.swift
@@ -21,7 +21,7 @@ import SwiftXCTest
 
 class TestNSArray : XCTestCase {
     
-    var allTests : [(String, () -> ())] {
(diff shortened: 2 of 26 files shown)


a2:
Since `Void` is just a typealias for `()`, why do you prefer one over the other? (Just curious, not arguing with your decision.)

Catfish-Man:
I have a mild preference for Void as well, partly from C-family familiarity, partly because () really seems like it should be Unit rather than Void…

a2:
I definitely feel like `()` makes more sense as "this is a set of empty arguments" whereas, like @Catfish-Man says, `Void` feels like a "nothing return type".

thii:
It is considered standard now.

In [devforums](https://devforums.apple.com/message/1133616#1133616):
> FWIW, we've recently decided to standardize on () -> Void (generally, () for parameters and Void for return types) across all of our documentation.

Similar commit in swift repo: https://github.com/apple/swift/commit/6536edd68c3cf44f0c61eb69c117a69f8b4185c2

I also personal prefer `Void` over `()` since it's more readable when you have nest parentheses.

justMaku:
I agree with @thii, () might suggest that a method returns an empty tuple. Void makes it clear what's the expected return type. Obviously that doesn't make a difference for the compiler, but it's simpler to read in my opinion.

radex:
Same. I'm completely okay with `()`, just that it gets quite noisy when you have nested parentheses.

Extreme example:

```
func blah(f: (() -> ())?)
```

A bit easier to read with Void:

```
func blah(f: (() -> Void)?)
```

joshuatbrown:
I agree with @radex - it gets hard to read with lots of nested parentheses. I'd also vote for `-> Void`. This change looks good to me.

phausler:
I personally prefer Void as well, however in earlier development I did run across some hiccups with it not working correctly on linux targets for some reason; however that problem is probably resolved by now

PR #113, opened 2015-12-10
[NSDate] comparable protocol implementation.

NaughtyOttsel:
Added basic comparison operators.

Equatable has been explicitly declared as a conformed protocol, whilst this is implicitly declared by conforming to the Comparable protocol, adding the explicit declaration makes it more readable and understandable that this is the case.


thii:
Please rebase your branch before pushing it here. Do not merge back from the master branch.

PR #114, opened 2015-12-10
[NSDictionary] Implement description, descriptionWithLocale, descriptionWithLocale:indent and tests

thii:


Foundation/NSDictionary.swift
```diff
@@ -264,7 +264,17 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
         return matching
     }
 
-    public override var description: String { NSUnimplemented() }
+    public override var description: String {
+        get {
+            var _description: String = "{\n"
+            for key in self.allKeys {
+                _description += "    \(key) = \(objectForKey(key)!);\n"
```
phausler: what happens if you have a dictionary of strings to dictionaries? shouldn't this cascade with indentation?

tomj: @phausler also what would the expected behaviour be in this situation if someone has done something like make a dictionary of dictionaries with *many* levels of depth?  Some kind of termination at some point?

phausler: so description should probably call descriptionWithLocale:indent: which that I would not expect to iterate more than say 100 deep. most likely at that level there is probably an error in the caller's code.

tomj: 👍- perfect, thanks for the heads up and apologies for the rookie move on my part by not looking there first

thii:
What would be the expected behavior if a negative indent level is passed?

phausler:
id say that would definitely be in the territory of programmer error or undefined behavior for the darwin version. So perhaps precondition if you think it warrants it; the objc version's interface is an NSUInteger for the level which gets translated to Int in swift unfortunately.

thii:
It that okay to change the indent level type to UInt?

phausler:
UInt would diverge from the API that is currently exposed in swift on Darwin platforms. Even though it would be more type safe it would diverge usage patterns. So lets keep it as Int

thii:
I updated it as requested. Since most classes' `descriptionWithLocale` and `descriptionWithLocale:indent` are unimplemented, I just added some basic tests.

phausler:
looks like a pretty decent start to me, will merge once I can verify tests on linux

PR #115, opened 2015-12-10
Add full-width romaji testcases to uppercase/lowercase strings.

tetsuok:
[Review on Reviewable](https://reviewable.io/reviews/apple/swift-corelibs-foundation/115)


tetsuok:
Added testcases are important for the applications that use uppercase/lowercase functionalities.
The users that use the Japanese input method might want to type full-width romaji using Romaji input mode in the app, rather than Kana input mode.

phausler:
looks good to me

PR #118, opened 2015-12-10
Implement NSArray.indexOfObject(inSortedRange:options:usingComparator)

HeMet:
Implement NSArray.indexOfObject(inSortedRange:options:usingComparator) and add test cases for it.


phausler:
looks pretty reasonable to me passes on mac and linux

PR #119, opened 2015-12-10
[NSDate] Comparable Protocol Implementation

NaughtyOttsel:
Pull Request re-done after failing to rebase.

Added basic comparison operators.

Equatable has been explicitly declared as a conformed protocol, whilst this is implicitly declared by conforming to the Comparable protocol, adding the explicit declaration makes it more readable and understandable that this is the case.

### DIFF
--- a/Foundation/NSDate.swift
+++ b/Foundation/NSDate.swift
@@ -169,6 +169,18 @@ extension NSDate {
 
 extension NSDate : _CFBridgable { }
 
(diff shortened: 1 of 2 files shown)


parkera:
Hi @NaughtyOttsel, we are discussing this on the `swift-evolution` and `swift-corelibs-dev` mailing lists as a potential improvement to API. Feel free to jump in on the discussion there.

PR #122, opened 2015-12-11
For NSURL source compatibility with Darwin Foundation

johnno1962:
Not sure what to do here but to have code compile on both platforms without changing the original foundation this would be the easiest option. Would be nice if ! could be avoided but this still gives the client a chance to check for nil (were it possible.)


parkera:
Ah, you're right that the current API returns a non-optional. So `NSURL` instead of `NSURL!`, actually.

parkera:
Let's close this for now and re-open when the changes are made.

PR #123, opened 2015-12-11
Added test for NSMutableArray

joshuatbrown:


phausler:
looks pretty reasonable to me, I just need to run the build and test on linux before merging.

PR #124, opened 2015-12-11
NSHTTPCookie v1 tests and refactoring

tptee:
Added tests for initializing a v1 cookie. Also, broke a lot of the initializer logic into separate private functions for clarity.


phausler:
@salgernon this seems to build just fine on both mac and linux; do we have any issues here?

dbrajkovic:
@salgernon & @phausler Please see https://github.com/apple/swift-corelibs-foundation/pull/182 before accepting.

salgernon:
Finally looking at this, and with it #182; I think we need a synthesis of the two patches for the NSHTTPCookie changes.  Let me think about how we can best proceed for the NSHTTPCookie implementation itself with that change.  (I think everyone is basically on the same page.)

parkera:
Ok, I think we've subsumed this into #182.

PR #125, opened 2015-12-11
NSMutableArray replaceObjectsInRange:withObjectsFromArray:range:

joshuatbrown:
Tested and implemented NSMutableArray's `replaceObjectsInRange:withObjectsFromArray:range:`


Foundation/NSArray.swift
```diff
@@ -658,7 +658,10 @@ public class NSMutableArray : NSArray {
             }
         }
     }
-    public func replaceObjectsInRange(range: NSRange, withObjectsFromArray otherArray: [AnyObject], range otherRange: NSRange) { NSUnimplemented() }
+    public func replaceObjectsInRange(range: NSRange, withObjectsFromArray otherArray: [AnyObject], range otherRange: NSRange) {
+        replaceObjectsInRange(range, withObjectsFromArray: otherArray.bridge().subarrayWithRange(otherRange))
```
phausler: the bridge here is not needed since otherArray is already an `Array<AnyObject>`

joshuatbrown: Thanks! I'm on it.

joshuatbrown:
@phausler OK, I switched it to convert the `NSRange` to a `Range<Int>` instead of bridging the array. Let me know if I should make any more changes. Thanks!

joshuatbrown:
@phausler I just saw your comment on #139 (which is another implementation of the same method) saying it should work like this:

```swift
var list = [AnyObject]()
array.getObjects(&list, range:otherRange)
replaceObjectsInRange(range, withObjectsFromArray:list)
```

Should I update the implementation in this PR to follow that pattern?

phausler:
Personally I think that is the most reasonable pattern: I will likely try to merge all of the unit tests together from the desperate requests for this implementation

joshuatbrown:
Hmm... Just looked at it and noticed that it only works if we bridge `otherArray` to a `NSArray` since `getObjects` is defined on `NSArray`. Here's the (tested, passing) code, which I'm happy to add to the PR if calling `bridge()` is acceptable:

```swift
var list = [AnyObject]()
otherArray.bridge().getObjects(&list, range:otherRange)
replaceObjectsInRange(range, withObjectsFromArray:list)
```

phausler:
yea bridge is fine here, usually we have tried to use ._swiftObject internally to Foundation bridge is just a simple wrapper around that.

joshuatbrown:
Ready for review again.

phausler:
looks good to me

PR #126, opened 2015-12-11
Flips a conditional check when parsing time zone data.

brownleej:


CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1079,7 +1079,7 @@ Boolean _CFTimeZoneInit(CFTimeZoneRef timeZone, CFStringRef name, CFDataRef data
             CFTZPeriod *tzp = NULL;
             CFIndex cnt = 0;
             __CFTimeZoneLockGlobal();
-            if (!__CFParseTimeZoneData(kCFAllocatorSystemDefault, data, &tzp, &cnt)) {
+            if (__CFParseTimeZoneData(kCFAllocatorSystemDefault, data, &tzp, &cnt)) {
```
parkera: Nice catch.

parkera:
Unfortunately our handling of time zones on Ubuntu 14.04 at least is broken; CFTimeZone.c does not fall back to `localtime()` or similar if the file at `/etc/localtime` is not a symlink. So this test fails on Linux at the moment.

parkera:
I pushed a commit to start towards a fix of the last thing, but it's still not enough yet.

brownleej:
I think I've found the issue. The tryParseGMTName function was using a Boolean for something that needed be a String index, and this seems to behave differently on Linux than on Mac OS, and restricts the values to 0 or 1. I've updated the commit with a fix for this, and the tests are passing on my Linux install.

brownleej:
This commit also takes a lot of shorthands because the tests only use a GMT+XXXX time zone, rather than a proper named one. My main motivation for this patch is that it will provide enough time zone support to address SR-171, which causes test failures on both Mac and Linux when the system time is around GMT+11. I'm hoping to take some time to work on named time zone support next, but it'll be tricky because Ubuntu uses TZIF2 time zone files, so there will need to be new code for parsing that format.

parkera:
I have a fix pending for SR-190, then I'll revisit this with that in place.

parkera:
Thanks for getting to the bottom of that string encoding conversion bug.

PR #128, opened 2015-12-11
NSNotificationCenter: Added observers synchronization and support for custom observer queue.

ochococo:
Added `observers` synchronization and custom `NSOperationQueue` for observer.

Initial implementation of: `NSOperation`, `NSOperationQueue`, `NSBlockOperation`.

### DIFF
--- a/Foundation/NSNotification.swift
+++ b/Foundation/NSNotification.swift
@@ -49,7 +49,8 @@ private class NSNotificationReceiver : NSObject {
     private weak var object: NSObject?
     private var name: String?
     private var block: ((NSNotification) -> Void)?
-    private var sender: AnyObject?
+    private weak var sender: AnyObject?
+    private weak var queue: NSOperationQueue?
 }
 
 
@@ -57,25 +58,31 @@ private let _defaultCenter: NSNotificationCenter = NSNotificationCenter()
 
 public class NSNotificationCenter : NSObject {
     
-    private var observers: [NSNotificationReceiver]
-    
-    public required override init() {
-        observers = [NSNotificationReceiver]()
-    }
+    private var _observers = [NSNotificationReceiver]()
+    private let _observersLock = NSLock()
     
     public class func defaultCenter() -> NSNotificationCenter {
         return _defaultCenter
     }
     
+    private func synchronized<T>(lock: NSLock, closure: () -> T) -> T {
+        lock.lock()
+        defer { lock.unlock() }
+        return closure()
+    }
+    
     public func postNotification(notification: NSNotification) {
+
         let name = notification.name
         let sender = notification.object
-        
-        let sendTo = observers.filter { observer in
-            let sameName = (observer.name == nil || observer.name == name)
-            let sameSender = (observer.sender == nil || observer.sender === sender)
+
+        let sendTo = synchronized(_observersLock) {
+            return self._observers.filter { observer in
+                let sameName = (observer.name == nil || observer.name == name)
+                let sameSender = (observer.sender == nil || observer.sender === sender)
             
-            return sameSender && sameName
+                return sameSender && sameName
+            }
         }
         
         for observer in sendTo {
@@ -83,7 +90,17 @@ public class NSNotificationCenter : NSObject {
                 continue
             }
             
-            block(notification)
+            if let queue = observer.queue where queue != NSOperationQueue.currentQueue() {
+                
+                let operation = NSBlockOperation {
+                    block(notification)
+                }
+                
+                queue.addOperation(operation)
+                operation.waitUntilFinished()
+            } else {
+                block(notification)
+            }
         }
     }
 
@@ -102,22 +119,22 @@ public class NSNotificationCenter : NSObject {
     }
 
     public func removeObserver(observer: AnyObject, name aName: String?, object anObject: AnyObject?) {
-        guard let observer = observer as? NSObject else {
+        
+        guard let observerToRemove = observer as? NSObject else {
             return
         }
         
-        observers = observers.filter { curObserver in
-            return (curObserver.object !== observer)
-                || (aName != nil && curObserver.name != aName)
-                || (anObject != nil && curObserver.sender !== anObject)
+        _observersLock.lock()
+        defer { _observersLock.unlock() }
+        
+        _observers = _observers.filter { observer in
+            return (observer.object !== observerToRemove)
+                || (aName != nil && observer.name != aName)
+                || (anObject != nil && observer.sender !== anObject)
         }
     }
-
     
     public func addObserverForName(name: String?, object obj: AnyObject?, queue: NSOperationQueue?, usingBlock block: (NSNotification) -> Void) -> NSObjectProtocol {
-        if queue != nil {
-            NSUnimplemented()
-        }
 
         let object = NSObject()
         
@@ -126,8 +143,11 @@ public class NSNotificationCenter : NSObject {
         newObserver.name = name
         newObserver.block = block
         newObserver.sender = obj
+        newObserver.queue = queue
         
-        observers.append(newObserver)
+        _observersLock.lock()
+        _observers.append(newObserver)
+        _observersLock.unlock()
         
         return object
     }
--- a/Foundation/NSOperation.swift
+++ b/Foundation/NSOperation.swift
@@ -7,11 +7,16 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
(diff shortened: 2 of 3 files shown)


Foundation/NSOperation.swift
```diff
@@ -81,17 +86,56 @@ public enum NSOperationQueuePriority : Int {
...
+    }
+    
+    public func addExecutionBlock(block: () -> ()) {
+        _executionBlocksLock.lock()
+        _executionBlocks.append(block)
+        _executionBlocksLock.unlock()
+    }
+
+    public var executionBlocks: [() -> ()] {
+        return _executionBlocks
```
phausler: This should be guarded by the lock else the array is accessed unsafely

Foundation/NSOperation.swift
```diff
@@ -7,11 +7,16 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+public let NSOperationDefaultThreadPriority: Double                  = 0.5
+public let NSOperationDefaultQueuePriority: NSOperationQueuePriority = .Normal
+public let NSOperationDefaultQualityOfService: NSQualityOfService    = .Background
```
phausler: These are not part of the API exposed by NSOperation; they should be internal

Foundation/NSOperation.swift
```diff
@@ -99,7 +143,9 @@ public let NSOperationQueueDefaultMaxConcurrentOperationCount: Int = 0 // Unimpl
 public class NSOperationQueue : NSObject {
     
     public override init() {
-        NSUnimplemented()
+        maxConcurrentOperationCount = NSOperationQueueDefaultMaxConcurrentOperationCount
+        suspended = false
+        qualityOfService = .UserInitiated
```
phausler: This should be NSThread.currentThread().qualityOfService

Foundation/NSOperation.swift
ochococo: Fixed.

Foundation/NSOperation.swift
ochococo: Got it. Fixed.

Foundation/NSOperation.swift
ochococo: Yeah but... `Value of type 'NSThread' has no member 'qualityOfService'`.

Foundation/NSNotification.swift
```diff
@@ -57,32 +58,43 @@ private let _defaultCenter: NSNotificationCenter = NSNotificationCenter()
...
                 continue
             }
             
+            if let queue = observer.queue {
+                
+                let operation = NSBlockOperation {
+                    block(notification)
+                }
+                
+                queue.addOperation(operation)
```
dmcrodrigues: I think we need to wait for the execution of all blocks before proceeding/returning otherwise we will introduce a different behaviour relative to the current notification center which works synchronously.

Foundation/NSNotification.swift
```diff
@@ -57,32 +58,43 @@ private let _defaultCenter: NSNotificationCenter = NSNotificationCenter()
...
+
+                return
```
dmcrodrigues: Also we cannot return in here, otherwise we will only notify the first observer registered with a queue when they may be multiple or in another example if the first observer has a queue but the second one doesn't.

Foundation/NSOperation.swift
```diff
@@ -81,17 +86,58 @@ public enum NSOperationQueuePriority : Int {
...
+    public override var asynchronous: Bool {
+        NSUnimplemented()
+    }
+    
+    public override var ready: Bool {
+        NSUnimplemented()
+    }
+    
+    public func addExecutionBlock(block: () -> ()) {
+        _executionBlocksLock.lock()
```
dmcrodrigues: We cannot add a block if the operation is currently executing or has already finished. In this conditions maybe we should trigger a `fatalError` with a message.

Foundation/NSOperation.swift
```diff
@@ -99,7 +145,11 @@ public let NSOperationQueueDefaultMaxConcurrentOperationCount: Int = 0 // Unimpl
 public class NSOperationQueue : NSObject {
     
     public override init() {
-        NSUnimplemented()
+        maxConcurrentOperationCount = NSOperationQueueDefaultMaxConcurrentOperationCount
```
dmcrodrigues: To keep consistency within the source file, shouldn't we define this in the property itself? Also, the below assignment seems redundant since `suspend` is already initialized in the declaration.

Foundation/NSNotification.swift
```diff
@@ -57,32 +58,43 @@ private let _defaultCenter: NSNotificationCenter = NSNotificationCenter()
...
-        
-        let sendTo = observers.filter { observer in
+
+        _observersLock.lock()
+        let sendTo = _observers.filter { observer in
             let sameName = (observer.name == nil || observer.name == name)
             let sameSender = (observer.sender == nil || observer.sender === sender)
             
             return sameSender && sameName
         }
```
sgl0v: I would propose to use `defer` statement for releasing the lock. Just to be sure that the lock is released even if the exception exits prematurely.

Foundation/NSNotification.swift
ochococo: Good catch. Fixed.

Foundation/NSNotification.swift
ochococo: Right, fixed.

Foundation/NSNotification.swift
ochococo: Cannot agree. I should unlock as soon as possible.

Foundation/NSOperation.swift
ochococo: Agreed.

Foundation/NSOperation.swift
ochococo: True. Fixed.

Foundation/NSNotification.swift
```diff
@@ -57,33 +58,43 @@ private let _defaultCenter: NSNotificationCenter = NSNotificationCenter()
...
+        _observersLock.unlock()
         
         for observer in sendTo {
             guard let block = observer.block else {
                 continue
             }
             
-            block(notification)
+            if let queue = observer.queue {
```
dmcrodrigues: We should also check if we're already running in this queue, using `NSOperationQueue.currentQueue()`, and in that case we need to simply invoke the block otherwise we can enter in a deadlock if the queue is configured to be serial. This scenario may be unlikely but is in fact possible so I think we should have this extra validation.

Foundation/NSNotification.swift
sgl0v: In case of exception being thrown, the lines above lead to deadlock. The `defer` corresponds to the current scope. As an option, you can do smth like this:

```
var sendTo: [NSNotificationReceiver]!
{
	_observersLock.lock()
	defer {_observersLock.unlock()}
	sendTo = _observers.filter { ... }
}
```

but I would propose to define `synchronized` method and use it all over the class implementation:

```
private func synchronized<T>(lock: AnyObject, closure: () -> T) -> T {
    lock.lock()
    defer { lock.unlock() }
    return closure()
}

let sendTo = synchronized(_observersLock, {
	return _observers.filter { ... }
})
```

ochococo: Excellent idea. Will introduce the second suggestion.

Foundation/NSNotification.swift
```diff
@@ -57,25 +58,31 @@ private let _defaultCenter: NSNotificationCenter = NSNotificationCenter()
...
-        observers = [NSNotificationReceiver]()
-    }
+    private var _observers = [NSNotificationReceiver]()
+    private let _observersLock = NSLock()
     
     public class func defaultCenter() -> NSNotificationCenter {
         return _defaultCenter
     }
     
+    private func synchronized<T>(lock: NSLock, closure: () -> T) -> T {
```
phausler: probably could `@noescape` the closure here

Foundation/NSOperation.swift
```diff
@@ -81,16 +86,33 @@ public enum NSOperationQueuePriority : Int {
...
+    
+    public convenience init(block: () -> ()) {
+        self.init()
+        addExecutionBlock(block)
     }
     
-    public func addExecutionBlock(block: () -> Void) {
-        NSUnimplemented()
+    public func addExecutionBlock(block: () -> ()) {
+        guard !executing else {
```
phausler: this is a race by accessing executing's result value

ochococo:
This PR grew a bit too large. 
I will split changes to two smaller and more feature-focused PRs.

PR #129, opened 2015-12-11
Implement equality for NSNull

kaandedeoglu:
Implemented the `isEqual` method for `NSNull`, that always returns` true` for two `NSNull` instances - thus giving it singleton semantics. (too bad we don't have factory methods or can't return references from initializers) 

Added tests.


TestFoundation/TestNSNull.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    func test_alwaysEqual() {
+        let null_1 = NSNull()
+        let null_2 = NSNull()
+        
+        let null_3: NSNull? = NSNull()
+        let null_4: NSNull? = nil
+        
+        XCTAssertEqual(null_1, null_2)
+        XCTAssertEqual(null_1, null_3)
```
phausler: The NSNull exposed from objc should also have the property of a singleton. So I would also expect that XCTAssertTrue(null_1 === null_2)

kaandedeoglu: yes but wouldn't that break our NSNull.swift implementation? Should I do a conditional assert?

phausler: in this particular case it preserves the effective functioning behavior of the "singleton-ness" of NSNull without needing a factory pattern init method. I think it is in the spirit of the API to preserve this and add an overload of the === operator to satisfy the expectation that one construction of NSNull() is to be treated as exactly the same as another construction of NSNull()

kaandedeoglu:
@phausler `===` is now overridden for NSNull, it acts as a singleton. Also added tests to reflect this as suggested.

phausler:
looks great!

PR #130, opened 2015-12-11
Fix NSArray.indexOfObject(inSortedRange:options:usingComparator) for fringe cases

HeMet:


parkera:
@phausler you're in the middle of this NSArray work, I think you should take a look at this one

phausler:
looks ok to me.

PR #131, opened 2015-12-11
Start implementing NSPredicate

5sw:
Started implementing predicates. Currently the constant and block-based predicates are done.

Since the class-cluster approach from Objective-C is not possible in Swift I replaced it with an internal protocol PredicateEvaluator. Depending on the type of predicate (that is which initializer is used) a different implementation of PredicateEvaluator is chosen. (So basically the Pimpl idiom from C++)

If you guys agree with this I will finish implementing NSPredicate and NSExpression in that style.

One thing I don't know how to solve is implementing `predicateWithSubstitutionVariables`. Since the return type is `Self` it would have to use `self.dynamicType.init` with a required initializer, but there is no suitable initializer. I'd appreciate any hints how to implement this without having to change the public API.

### DIFF
--- a/Foundation.xcodeproj/project.pbxproj
+++ b/Foundation.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		22B9C1E11C165D7A00DECFF9 /* TestNSDate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 22B9C1E01C165D7A00DECFF9 /* TestNSDate.swift */; };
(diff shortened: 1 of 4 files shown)


parkera:
I accepted an earlier PR that started the Predicate implementation, so we should rebase this one on top of that if it's still adding more functionality.

`predicateWithSubstitutionVariables` may be impossible to implement without factory methods... I will take a closer look and see if I can figure something out too.

5sw:
Closing this, adds nothing new to the other PR.

PR #111, opened 2015-12-10
Rearrange list of all test cases alphabetically.

thii:
- Easier to know where to add new one.
- Avoiding duplications by mistakes (I found 2 `TestNSJSONSerialization `).

PR #132, opened 2015-12-11
Initial implementation of NSNumberFormatter.

chrisfsampaio:
- Added implementation for attributes.
- Added tests for initial implementation.
- Updated implementation status documentation.

### DIFF
--- a/Foundation/NSNumberFormatter.swift
+++ b/Foundation/NSNumberFormatter.swift
@@ -7,18 +7,34 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-
+import CoreFoundation
 
 public class NSNumberFormatter : NSFormatter {
     
-    public override init() { NSUnimplemented() }
-    public required init?(coder aDecoder: NSCoder) {
-        NSUnimplemented()
+    typealias CFType = CFNumberFormatterRef
+    private var _currentCfFormatter: CFType?
+    private var _cfFormatter: CFType {
+        get {
+            if let obj = _currentCfFormatter {
+                return obj
+            } else {
+                #if os(OSX) || os(iOS)
+                    let numberStyle = CFNumberFormatterStyle(rawValue: CFIndex(self.numberStyle.rawValue))!
+                #else
+                    let numberStyle = CFNumberFormatterStyle(self.numberStyle.rawValue)
+                #endif
+                
+                let obj = CFNumberFormatterCreate(kCFAllocatorSystemDefault, locale._cfObject, numberStyle)
+                _setFormatterAttributes(obj)
+                _currentCfFormatter = obj
+                return obj
+            }
+        }
     }
     
     // this is for NSUnitFormatter
     
-    public var formattingContext: NSFormattingContext // default is NSFormattingContextUnknown
+    public var formattingContext: NSFormattingContext = .Unknown // default is NSFormattingContextUnknown
     
     // Report the used range of the string and an NSError, in addition to the usual stuff from NSFormatter
     /// - Experiment: This is a draft API currently under consideration for official import into Foundation as a suitable alternative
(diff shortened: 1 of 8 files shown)


Foundation/NSNumberFormatter.swift
```diff
@@ -27,87 +43,796 @@ public class NSNumberFormatter : NSFormatter {
...
+        }
+        
+        set {
+            _reset()
+            _numberStyle = newValue
+        }
+    }
+    
+    internal var _locale: NSLocale = NSLocale.currentLocale()
+    /*@NSCopying*/ public var locale: NSLocale! {
```
cullenbmacdonald: another option is to drop the internal var here and use `willSet`

chrisfsampaio: That would be nice!
But I think that on a next step we should do something like this:
```swift
    internal var _numberStyle: NSNumberFormatterStyle = .NoStyle
    public var numberStyle: NSNumberFormatterStyle {
        get {
            let rawValue = CFNumberFormatterGetStyle(_cfObject).rawValue
            _numberStyle = NSNumberFormatterStyle(rawValue: UInt(rawValue))!
            return _numberStyle
        }
        
        set {
            _reset()
            _numberStyle = newValue
        }
    }
```
And then we would be sure that both our object and the internal CoreFoundation formatter are in sync, which currently isn't happening. And we could also take leverage of CoreFoundation number formatter default values.

I really dislike the interval vars too, but I can't think of some other option. :disappointed:
Do you see any other way that allows us to keep both worlds in sync?

chrisfsampaio:
Done, added the missing tests.
In terms of implementation, I guess this is enough for this PR (taking [incremental development](https://swift.org/contributing/#contributing-code) into account).
Please, let me know if I need to change anything else.

Foundation/NSNumberFormatter.swift
```diff
@@ -7,18 +7,34 @@
...
+                #else
+                    let numberStyle = CFNumberFormatterStyle(self.numberStyle.rawValue)
+                #endif
+                
+                let obj = CFNumberFormatterCreate(kCFAllocatorSystemDefault, locale._cfObject, numberStyle)
+                _setFormatterAttributes(obj)
+                __cfObject = obj
+                return obj
+            }
+        }
```
parkera: Two things here: 

1. We use the `_cfObject` method for types which are bridged. I think that the CFNumberFormatter/NSNumberFormatter relationship could instead be 'has-a', as in NSNumberFormatter has a CFNumberFormatterRef. This means we don't have to worry about things like matching up ivar layout. So I would call this internal var something besides _cfObject to avoid confusion.

2. Whatever we call the internal var backing the internal var, it'd be nice if it was not the same name with just an additional underscore... I feel like it's too easy to mess that up and use it accidentally somewhere in the rest of the class.

TestFoundation/main.swift
```diff
@@ -50,4 +50,6 @@ XCTMain([
     TestNSURLResponse(),
     TestNSUUID(),
     TestNSXMLParser(),
+    TestNSNotificationCenter(),
```
parkera: Seems like this line might have slipped in here accidentally.

parkera:
I'm going to review some more later but I think this is a great start.

TestFoundation/main.swift
chrisfsampaio: 😁👍 f05ce390

Foundation/NSNumberFormatter.swift
chrisfsampaio: 👍 ef329ed

Regarding (2), what do you think of making it `private`?  If we need access to the internal formatter somewhere else in the lib, one still could use the `_cfFormatter`.

chrisfsampaio:
> I'm going to review some more later but I think this is a great start.

Nice! 👍

Foundation/NSNumberFormatter.swift
parkera: I think it's probably the case that we want the _cfFormatter to only be used within NSDateFormatter itself; other internal-to-library clients should still use the public API (preferred) or a private swift method (if required for some reason).

parkera:
Let's go ahead and rebase this and squash into a single commit for merge. Thanks!

Foundation/NSNumberFormatter.swift
chrisfsampaio: Makes sense, I made both of them private then.

chrisfsampaio:
> Let's go ahead and rebase this and squash into a single commit for merge. Thanks!

👍 Done! Thanks!

I've also reordered the test case to be alphabetically sorted in favor of #111.

chrisfsampaio:
I've also updated the [status page](https://github.com/chrisfsampaio/swift-corelibs-foundation/blob/number-formatter-implementation/Docs/Status.md#implementation-status) to reflect the changes made on this PR.

parkera:
Hm, getting 5 test failures, all in the currency paths, but only on Linux.

parkera:
Ah, the currency stuff is only available on ICU 5.5 or later.

PR #133, opened 2015-12-11
Implement NSRunLoop

kovpas:
addPort and removePort are not implemented because NSPort implementation is missing.

### DIFF
--- a/CoreFoundation/RunLoop.subproj/CFRunLoop.c
+++ b/CoreFoundation/RunLoop.subproj/CFRunLoop.c
@@ -71,9 +71,18 @@ DISPATCH_EXPORT void _dispatch_main_queue_callback_4CF(void);
 
 #define AbsoluteTime LARGE_INTEGER 
 
+#elif DEPLOYMENT_TARGET_LINUX
+#include <dlfcn.h>
+#include <poll.h>
+#include <sys/epoll.h>
+#include <sys/eventfd.h>
+#include <sys/timerfd.h>
+
+#define _dispatch_get_main_queue_port_4CF dispatch_get_main_queue_eventfd_np
+#define _dispatch_main_queue_callback_4CF(x) dispatch_main_queue_drain_np()
 #endif
 
-#if DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_IPHONESIMULATOR
+#if DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_IPHONESIMULATOR || DEPLOYMENT_TARGET_LINUX
 CF_EXPORT pthread_t _CF_pthread_main_thread_np(void);
 #define pthread_main_thread_np() _CF_pthread_main_thread_np()
 #endif
@@ -115,6 +124,13 @@ static void _runLoopTimerWithBlockContext(CFRunLoopTimerRef timer, void *opaqueB
 static pthread_t kNilPthreadT = { nil, nil };
 #define pthreadPointer(a) a.p
 typedef	int kern_return_t;
+#define KERN_SUCCESS 0
+
+#elif DEPLOYMENT_TARGET_LINUX
+
+static pthread_t kNilPthreadT = (pthread_t)0;
+#define pthreadPointer(a) ((void*)a)
+typedef int kern_return_t;
 #define KERN_SUCCESS 0
 
 #else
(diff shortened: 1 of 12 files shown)


parkera:
The main issue here is that none of this will work on Linux because the classes are not yet ported.

kovpas:
@parkera I think I've got `CFRunLoop.c` working on ubuntu. Probably better to post it as a separate pull request, though. It's heavily based on `CFLite` by @nickhutchinson ( https://github.com/nickhutchinson/CFLite/tree/master ). Had to tweak a couple of things, but nothing too complex. Would be great if you could take a look and let me know if I could proceed with pull request.

https://github.com/kovpas/swift-corelibs-foundation/compare/nsrunloop...kovpas:cfrunloop_ubuntu

parkera:
I think you're on the right track; I had started on something similar, although I used `epoll` instead.

parkera:
Since we don't really have a mechanism in our tests for writing C unit tests of CF directly, I think it makes sense to consider this all together so we can verify that it's working.

kovpas:
@parkera makes sense to treat it as one thing, indeed. I'll update this PR. Also going to add some tests for `NSTimer`

One thing: for some reason `NSRunLoop.swift` doesn't compile for me on linux when I use `CFRunLoopRunResult`:

```
Foundation/NSRunLoop.swift:75:34: error: type of expression is ambiguous without more context
        return runloopResult == .HandledSource || runloopResult == .TimedOut
                                ~^~~~~~~~~~~~~
```

Do you have any idea why this could happen and how to fix that? Meanwhile I had to do that:

```
return runloopResult == 2 || runloopResult == 3
```

build.py
```diff
@@ -14,7 +14,7 @@
 foundation.GCC_PREFIX_HEADER = 'CoreFoundation/Base.subproj/CoreFoundation_Prefix.h'
 
 if Configuration.current.target.sdk == OSType.Linux:
-	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX '
+	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE '
```
parkera: This causes a build warning: 

```
[9/175] CompileC: closure/runtime.c
closure/runtime.c:16:9: warning: '__USE_GNU' macro redefined [-Wmacro-redefined]
#define __USE_GNU
        ^
/usr/include/features.h:322:10: note: previous definition is here
# define __USE_GNU      1
         ^
1 warning generated.
```

Foundation/NSRunLoop.swift
```diff
@@ -7,38 +7,51 @@
...
 
     public func removePort(aPort: NSPort, forMode mode: String) {
+//        CFRunLoopRemoveSource(CFRunLoopGetCurrent(), aPort._cfObject, mode._cfObject)
         NSUnimplemented()
     }
     
     public func limitDateForMode(mode: String) -> NSDate? {
-        NSUnimplemented()
+        let nextTimerFireAbsoluteTime = CFRunLoopGetNextTimerFireDate(CFRunLoopGetCurrent(), mode._cfObject)
+        return NSDate(timeIntervalSinceReferenceDate: nextTimerFireAbsoluteTime)
```
parkera: If `nextTimerFireAbsoluteTime` is == 0 then return NSDistantFuture here.

Foundation/NSRunLoop.swift
```diff
@@ -7,38 +7,51 @@
...
+        return CFRunLoopCopyCurrentMode(_cfObject)?._swiftObject
     }
     
     public func addTimer(timer: NSTimer, forMode mode: String) {
-        NSUnimplemented()
+        CFRunLoopAddTimer(CFRunLoopGetCurrent(), timer._cfObject, mode._cfObject)
     }
     
     public func addPort(aPort: NSPort, forMode mode: String) {
+//        CFRunLoopAddSource(CFRunLoopGetCurrent(), aPort._cfObject, mode._cfObject)
```
parkera: Let's leave the commented line out of the commit until we implement this.

Foundation/NSRunLoop.swift
```diff
@@ -7,38 +7,51 @@
...
+import CoreFoundation
 
-public let NSDefaultRunLoopMode: String = "NSDefaultRunLoopMode"
-public let NSRunLoopCommonModes: String = "NSRunLoopCommonModes"
+public let NSDefaultRunLoopMode: String = kCFRunLoopDefaultMode._swiftObject
+public let NSRunLoopCommonModes: String = kCFRunLoopCommonModes._swiftObject
 
 public class NSRunLoop : NSObject {
+    typealias CFType = CFRunLoopRef
+    internal var _cfObject : CFType!
```
parkera: Let's call this var `_cfRunLoop` or similar. `_cfObject` is reserved for our toll-free bridging.

CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -2256,6 +2346,94 @@ static Boolean __CFRunLoopServiceMachPort(mach_port_name_t port, mach_msg_header
...
+
+    } else {
+        struct epoll_event event;
+        do {
+            result = epoll_wait(portSet, &event, 1 /*numEvents*/, 0 /*timeout*/);
+        } while (result == -1 && errno == EINTR);
+        CFAssert2(result >= 0, __kCFLogAssertion, "%s(): error %d from epoll_wait", __PRETTY_FUNCTION__, errno);
+
+        if (result == 0)
+            return false;
```
parkera: style note - please use { } instead of dangling if statements here.

parkera:
Overall looks really good. I just have a few minor comments for you to address then we'll merge this in. Thanks!

build.py
kovpas: It's [not recommended](https://gcc.gnu.org/ml/fortran/2005-10/msg00365.html) to define `__USE_GNU` in code, better to use define `_GNU_SOURCE`. I just removed this define in `closure/runtime.c`, as all source files are now compiled with `_GNU_SOURCE`.

Foundation/NSRunLoop.swift
kovpas: Done

Foundation/NSRunLoop.swift
kovpas: Done

Foundation/NSRunLoop.swift
kovpas: Done

CoreFoundation/RunLoop.subproj/CFRunLoop.c
kovpas: Done.

PR #134, opened 2015-12-11
Implement contentsOfDirectoryAtPath

euantorano:
Implementing NSFileManager::contentsOfDirectoryAtPath() and NSFileManager::subpathsOfDirectoryAtPath().

Error messages aren't currently an exact match for the standard Foundation library, which throws:

```
Fatal error: Error raised at top level: Error Domain=NSCocoaErrorDomain Code=260 "The folder “notexist” doesn’t exist." UserInfo={NSFilePath=/Users/euan/Desktop/notexist, NSUserStringVariant=(
    Folder
), NSUnderlyingError=0x1007019f0 {Error Domain=NSOSStatusErrorDomain Code=-43 "fnfErr: File not found"}}: file /Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-700.1.101.15/src/swift/stdlib/public/core/ErrorType.swift, line 56
(lldb)
```

Whereas this commit throws:

```
fatal error: Error raised at top level: Error Domain=NSCocoaErrorDomain Code=260 "The file notexist couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/euan/Desktop/notexist}: file /Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-700.1.101.15/src/swift/stdlib/public/core/ErrorType.swift, line 56
```

However, I lack the knowledge to know what I'm missing with the `NSError`.

Otherwise, the test cases pass and this should hopefully work.

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -191,20 +191,115 @@ public class NSFileManager : NSObject {
         }
     }
     
-    /* contentsOfDirectoryAtPath:error: returns an NSArray of NSStrings representing the filenames of the items in the directory. If this method returns 'nil', an NSError will be returned by reference in the 'error' parameter. If the directory contains no items, this method will return the empty array.
+    /**
+     Performs a shallow search of the specified directory and returns the paths of any contained items.
      
-        This method replaces directoryContentsAtPath:
+     This method performs a shallow search of the directory and therefore does not traverse symbolic links or return the contents of any subdirectories. This method also does not return URLs for the current directory (“.”), parent directory (“..”) but it does return other hidden files (files that begin with a period character).
+     
+     The order of the files in the returned array is undefined.
+     
+     - Parameter path: The path to the directory whose contents you want to enumerate.
+     
+     - Throws: `NSError` if the directory does not exist, this error is thrown with the associated error code.
+     
+     - Returns: An array of String each of which identifies a file, directory, or symbolic link contained in `path`. The order of the files returned is undefined.
      */
     public func contentsOfDirectoryAtPath(path: String) throws -> [String] {
-        NSUnimplemented()
+        var contents : [String] = [String]()
+        
+        let dir = opendir(path)
+        
+        if dir == nil {
+            throw NSError(domain: NSCocoaErrorDomain, code: NSCocoaError.FileReadNoSuchFileError.rawValue, userInfo: [NSFilePathErrorKey: path])
+        }
+        
+        defer {
+            closedir(dir)
+        }
+        
+        var entry: UnsafeMutablePointer<dirent> = readdir(dir)
+        
+        while entry != nil {
+            if let entryName = withUnsafePointer(&entry.memory.d_name, { (ptr) -> String? in
+                let int8Ptr = unsafeBitCast(ptr, UnsafePointer<Int8>.self)
+                return String.fromCString(int8Ptr)
+            }) {
+                // TODO: `entryName` should be limited in length to `entry.memory.d_namlen`.
+                if entryName != "." && entryName != ".." {
+                    contents.append(entryName)
+                }
+            }
+            
+            entry = readdir(dir)
+        }
+        
+        return contents
     }
     
-    /* subpathsOfDirectoryAtPath:error: returns an NSArray of NSStrings representing the filenames of the items in the specified directory and all its subdirectories recursively. If this method returns 'nil', an NSError will be returned by reference in the 'error' parameter. If the directory contains no items, this method will return the empty array.
-     
-        This method replaces subpathsAtPath:
-     */
+    /**
+    Performs a deep enumeration of the specified directory and returns the paths of all of the contained subdirectories.
+    
+    This method recurses the specified directory and its subdirectories. The method skips the “.” and “..” directories at each level of the recursion.
+    
+    Because this method recurses the directory’s contents, you might not want to use it in performance-critical code. Instead, consider using the enumeratorAtURL:includingPropertiesForKeys:options:errorHandler: or enumeratorAtPath: method to enumerate the directory contents yourself. Doing so gives you more control over the retrieval of items and more opportunities to abort the enumeration or perform other tasks at the same time.
+    
+    - Parameter path: The path of the directory to list.
+    
+    - Throws: `NSError` if the directory does not exist, this error is thrown with the associated error code.
+    
+    - Returns: An array of NSString objects, each of which contains the path of an item in the directory specified by path. If path is a symbolic link, this method traverses the link. This method returns nil if it cannot retrieve the device of the linked-to file.
+    */
     public func subpathsOfDirectoryAtPath(path: String) throws -> [String] {
-        NSUnimplemented()
+        var contents : [String] = [String]()
+        
+        let dir = opendir(path)
+        
+        if dir == nil {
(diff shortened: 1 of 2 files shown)


parkera:
Overall this looks pretty good, with the minor comment on `._`, which is worth testing.

Can you rebase this on top of master and into a single commit?

euantorano:
Not a problem, I'll do so tonight once I've tested the `._` issue.

euantorano:
Rebased and squashed, hopefully that's all fixed! Your comments about the `._` seem to have been lost, but basically the docs [here](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSFileManager_Class/#//apple_ref/occ/instm/NSFileManager/contentsOfDirectoryAtPath:error:) seem to be wrong. The comments state that `This method also does not return URLs for the current directory (“.”), parent directory (“..”), or resource forks (files that begin with “._”)` when in actual fact files starting with `._` are indeed returned. This PR copies that behaviour, but removes that comment.

Foundation/NSFileManager.swift
```diff
@@ -191,20 +191,110 @@ public class NSFileManager : NSObject {
...
+        
+        let dir = opendir(path)
+        
+        if dir == nil {
+            throw NSError(domain: NSCocoaErrorDomain, code: NSCocoaError.FileReadNoSuchFileError.rawValue, userInfo: [NSFilePathErrorKey: path])
+        }
+        
+        defer {
+            closedir(dir)
+            dir.destroy()
```
phausler: I am not certain this is the correct usage here; destroy is usually reserved from the initialized from swift scope buffers.

euantorano: That may be true, but I wasn't sure. It doesn't cause any errors.

phausler: After looking into it - it seems to be almost a no-op in this case

TIL...
Interesting - did some digging around in the standard lib:
It claims
```swift
  /// Destroy the object the pointer points to.
  ///
  /// - Precondition: The memory is initialized.
  ///
  /// - Postcondition: The value has been destroyed and the memory must
  ///   be initialized before being used again.
```
and then promptly calls 
```
Builtin.destroy(Memory.self, _rawValue)
```
Which in turn ends up dealing with witness tables and seems to be primarily for object based buffers. So it would probably only matter if this was happening on UnsafeMutablePointer of CF types which are translated to swift retain/released objects.

euantorano: Interesting. Sounds like I can just remove it then without any downsides. I'll update this PR now.

Foundation/NSFileManager.swift
```diff
@@ -191,20 +191,110 @@ public class NSFileManager : NSObject {
...
+        }
+        
+        var entry: UnsafeMutablePointer<dirent> = readdir(dir)
+        
+        while entry != nil {
+            if let entryName = withUnsafePointer(&entry.memory.d_name, { (ptr) -> String? in
+                let int8Ptr = unsafeBitCast(ptr, UnsafePointer<Int8>.self)
+                return String.fromCString(int8Ptr)
```
phausler: isn't d_name not always null terminated and d_namlen should be used to limit the number of characters in d_name?

euantorano: Yes, but `String::fromCString` appears to handle this:

```
Creates a new `String` by copying the nul-terminated UTF-8 data referenced by a `Cstring`.
```

phausler: so given the bytes of "foobar\5\0" and the d_namelen of 3 the name should be "foo" but fromCString will make it "foobar\5"

euantorano: True, but the strings aren't user provided, they're coming from the OS kernel. I guess the best way to be 100% sure is to take a substring of the Swift string as it'll be much easier than trying to convert the `d_name` Tuple into an array and limit the entries.

phausler: I have seen this in the past that some altered forms of linux have done this type of thing. That type of thing is perfectly fine for a future fix. Not a blocker imho

euantorano: Fair enough, I'll try and sneak it in if the current version of Swift (mine is slightly out of date) has the substring operation.

euantorano:
Fixed the concerns regarding `.destroy()`. Any other possible blockers?

phausler:
looks pretty reasonable overall to me; just need to run tests before merging

Foundation/NSFileManager.swift
```diff
@@ -191,20 +191,106 @@ public class NSFileManager : NSObject {
...
+            }) {
+                // TODO: `entryName` should be limited in length to `entry.memory.d_namlen`.
+                if entryName != "." && entryName != ".." {
+                    contents.append(entryName)
+                    
+                    if let entryType = withUnsafePointer(&entry.memory.d_type, { (ptr) -> Int32? in
+                        let int32Ptr = unsafeBitCast(ptr, UnsafePointer<UInt8>.self)
+                        return Int32(int32Ptr.memory)
+                    }) {
+                        if entryType == DT_DIR {
```
phausler: on ubuntu I get this error:
```
Foundation/NSFileManager.swift:280:38: error: binary operator '==' cannot be applied to operands of type 'Int32' and 'Int'
                        if entryType == DT_DIR {
                           ~~~~~~~~~ ^  ~~~~~~
Foundation/NSFileManager.swift:280:38: note: overloads for '==' exist with these partially matching parameter lists: (Int32, Int32), (Int, Int)
                        if entryType == DT_DIR {
```

euantorano: Sounds like a bug elsewhere within the runtime, but I'll get round it anyway. The two types should be comparable as they are on OS X.

TestFoundation/TestNSFileManager.swift
```diff
@@ -239,4 +241,94 @@ class TestNSFileManger : XCTestCase {
...
+            fm.createFileAtPath(itemPath1, contents: NSData(), attributes: nil)
+            fm.createFileAtPath(itemPath2, contents: NSData(), attributes: nil)
+        } catch _ {
+            XCTFail()
+        }
+        
+        do {
+            let entries = try fm.contentsOfDirectoryAtPath(path)
+            
+            XCTAssertEqual(entries, expected)
```
phausler: this fails on ubuntu 14

["item2", "item"] should be ["item", "item2"]

Mac OS X returns the directory contents as sorted, ubuntu does not do this, so perhaps it should be an un-ordered collection comparison. e.g. verify the count and verify that the expected items are contained in the result array

TestFoundation/TestNSFileManager.swift
```diff
@@ -239,4 +241,94 @@ class TestNSFileManger : XCTestCase {
...
+            try fm.createDirectoryAtPath(path2, withIntermediateDirectories: false, attributes: nil)
+            fm.createFileAtPath(itemPath3, contents: NSData(), attributes: nil)
+        } catch _ {
+            XCTFail()
+        }
+        
+        do {
+            let entries = try fm.subpathsOfDirectoryAtPath(path)
+            
+            XCTAssertEqual(entries, expected)
```
phausler: also presumes order of the entries

euantorano:
Fixed. I haven't tested on Linux, I'm setting up a VM to do so now.

phausler:
yep, seems to be good now.

euantorano:
:+1:

PR #135, opened 2015-12-12
Add CGVector and a basic construction test

maxadamski:


parkera:
I know this seems like a fine line, but mostly we limited the CG types in Foundation to the ones that Foundation itself needed (e.g. for encoding). I'm not sure if I want to import all of the CG types here.

Is there an important use case you had in mind for bringing this in?

parkera:
I'm going to ahead and close this for now. Let's discuss on the mailing list if we want to reopen the issue.

PR #136, opened 2015-12-13
[NSMutableArray] replace objects in range docs/tests/impl

cullenbmacdonald:
I think I found a bug while writing these tests. Not sure what the process is for filing it, but I was getting a a complier error "Int does not conform to expected element type AnyObject"  when doing `NSMutableArray(array: [1, 2, 3])` which seems...wrong.

Tested on latests official xcode build from app store and it definitely works.

Will be happy to try and resolve that issue before this gets merged as the tests look a little verbose as a result of the bug.

### DIFF
--- a/Foundation.xcodeproj/project.pbxproj
+++ b/Foundation.xcodeproj/project.pbxproj
@@ -203,6 +203,7 @@
 		61F8AE7D1C180FC600FB62F0 /* TestNSNotificationCenter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61F8AE7C1C180FC600FB62F0 /* TestNSNotificationCenter.swift */; };
 		6E203B8D1C1303BB003B2576 /* TestNSBundle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6E203B8C1C1303BB003B2576 /* TestNSBundle.swift */; };
 		7A7D6FBB1C16439400957E2E /* TestNSURLResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A7D6FBA1C16439400957E2E /* TestNSURLResponse.swift */; };
+		7AD22BF01C1CE0AA00377EC0 /* TestNSMutableArray.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AD22BEF1C1CE0AA00377EC0 /* TestNSMutableArray.swift */; };
 		83712C8E1C1684900049AD49 /* TestNSURLRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 83712C8D1C1684900049AD49 /* TestNSURLRequest.swift */; };
 		844DC3331C17584F005611F9 /* TestNSScanner.swift in Sources */ = {isa = PBXBuildFile; fileRef = 844DC3321C17584F005611F9 /* TestNSScanner.swift */; };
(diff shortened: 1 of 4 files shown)


Foundation.xcodeproj/project.pbxproj
```diff
@@ -1041,36 +1043,37 @@
 		EA66F65A1BF1976100136161 /* Tests */ = {
 			isa = PBXGroup;
 			children = (
```
cullenbmacdonald: simply sorted the order here.

Foundation/NSArray.swift
```diff
@@ -658,7 +658,19 @@ public class NSMutableArray : NSArray {
             }
         }
     }
-    public func replaceObjectsInRange(range: NSRange, withObjectsFromArray otherArray: [AnyObject], range otherRange: NSRange) { NSUnimplemented() }
+    
+    /// Replaces the objects in the receiving array specified by one given range with the objects in another array specified by another range.
```
cullenbmacdonald: these docstrings are copy pasted from official docs.

TestFoundation/TestNSMutableArray.swift
```diff
@@ -0,0 +1,58 @@
...
+class TestNSMutableArray : XCTestCase {
+    
+    var allTests : [(String, () -> ())] {
+        return [
+            ("test_replaceObjectsInRangeWithObjectsFromArray", test_replaceObjectsInRangeWithObjectsFromArray),
+        ]
+    }
+    
+    func test_replaceObjectsInRangeWithObjectsFromArray() {
+        let existing = [NSNumber(int: 1), NSNumber(int: 2), NSNumber(int: 3), NSNumber(int: 4), NSNumber(int: 5), NSNumber(int: 6)]
```
cullenbmacdonald: this is what is more verbose as a result of the bug i found.

parkera:
What you're seeing is the lack of implicit bridging from Int to NSNumber. `AnyObject` requires that the type is a class type. Int is not a class type unless it's bridged to an NSNumber.

Unfortunately this is expected behavior until we can get implicit bridging figured out when the ObjC runtime is not present.

cullenbmacdonald:
got it. that makes a lot of sense. Do I need to change anything about this pull then?

Foundation/NSArray.swift
```diff
@@ -658,7 +658,19 @@ public class NSMutableArray : NSArray {
...
+    /// - parameter range: The range of objects to be replaced in (or removed from) the receiving array.
+    /// - parameter otherArray: The array of objects from which to select replacements for the objects in `range`.
+    /// - parameter otherRnage: The range of objects be selected from `otherArray` as replacements for the objects in `range`.
```
parkera: typo here in `otherRnage`

parkera:
Other than the typo, looks good to me.

cullenbmacdonald:
fixed

joshuatbrown:
It looks like there are at least three implementations of this method now:

1. PR #125: https://github.com/apple/swift-corelibs-foundation/pull/125
2. PR #136: https://github.com/apple/swift-corelibs-foundation/pull/136 (this one)
3. PR #139: https://github.com/apple/swift-corelibs-foundation/pull/139

thii:
From the [Contribution Guidelines](https://github.com/apple/swift-corelibs-foundation/blob/master/CONTRIBUTING.md):
> Before embarking on a large amount of work to implement missing functionality, please double-check with the community on the swift-corelibs-dev mailing list. Someone may already be working in this area, and we want to avoid duplication of work.

cullenbmacdonald:
My mistake. Dunno how I missed #125! ill be more careful next time

PR #139, opened 2015-12-13
Added NSArray.replaceObjectsInRange(range, otherArray, otherRange) 

terhechte:
Added the implementation of this method as well as a test for it

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -658,7 +658,10 @@ public class NSMutableArray : NSArray {
             }
         }
     }
-    public func replaceObjectsInRange(range: NSRange, withObjectsFromArray otherArray: [AnyObject], range otherRange: NSRange) { NSUnimplemented() }
+    public func replaceObjectsInRange(range: NSRange, withObjectsFromArray otherArray: [AnyObject], range otherRange: NSRange) {
+        self.replaceObjectsInRange(range, withObjectsFromArray: Array(otherArray[otherRange.toRange()!]))
+    }
+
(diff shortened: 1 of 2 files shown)


joshuatbrown:
There are now three implementations of this method:

1. PR #125
2. PR #136
3. PR #139 (this one)

Not sure which should be merged, but I'm guessing not all of them should be. :)

parkera:
We'll try to get this sorted out...

terhechte:
Oops sorry! I hadn't seen these other PRs. I'll do a better search next time.
Interestingly we're all using slightly different implementations:

```self.replaceObjectsInRange(range, withObjectsFromArray: Array(otherArray[otherRange.toRange()!]))```

vs

```let slicedOtherArray = Array(otherArray[otherRange.location...otherRange.length+otherRange.location])
replaceObjectsInRange(range, withObjectsFromArray: slicedOtherArray)```

vs

```replaceObjectsInRange(range, withObjectsFromArray: otherArray.bridge().subarrayWithRange(otherRange))```

**Comment by phausler** [comment]
so the expected call pattern in objc is:
```objective-c
[array getObjects:list range:otherRange];
[self replaceObjectsInRange:range withObjects:list count:otherRange.length];
```

this should probably hit the same type of call pattern:
```swift
var list = [AnyObject]()
array.getObjects(&list, range:otherRange)
replaceObjectsInRange(range, withObjectsFromArray:list)
```
this preserves subclasser expectations for performance since getObjects:range: should be the method to override as a "fast path"

phausler:
merged in for extra unit tests

PR #140, opened 2015-12-14
Fix: addIndexesInRange now correctly adds neighbor NSRanges

adomanico:
Fixing a bug where adding neighboring indices would result in an incorrect range being created in the index set.

Example:

```swift

let set = NSMutableIndexSet(index: 0)
set.addIndex(1)

```

In this case, the index set would contain one NSRange (location 2, length 0). 

The correct result should be that the index set contains one NSRange (location 0, length 2).

Added some tests as well.


phausler:
looks good to me

PR #141, opened 2015-12-14
Initial FreeBSD port of Foundation.

dcci:
Now that we got a working swiftc on FreeBSD, I started porting foundation (lots of tests failing because there's no support for that). There's still a lot to do, but this is a starting point.


CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -223,6 +228,32 @@ CF_INLINE size_t malloc_size(void *memblock) {
...
+int32_t OSAtomicIncrement32Barrier(volatile int32_t *dst);
+int32_t OSAtomicIncrement32(volatile int32_t *theValue);
+int32_t OSAtomicDecrement32(volatile int32_t *theValue);
+
+int32_t OSAtomicAdd32( int32_t theAmount, volatile int32_t *theValue );
+int32_t OSAtomicAdd32Barrier( int32_t theAmount, volatile int32_t *theValue );
+bool OSAtomicCompareAndSwap32Barrier( int32_t oldValue, int32_t newValue, volatile int32_t *theValue );
+
+void OSMemoryBarrier();
+
```
phausler: does FreeBSD have malloc_size? I was under the impression that it was similar to linux in that it defined malloc_usable_size instead.

phausler:
neat! overall this looks pretty decent of an initial `#ifdef` pass. What is the testing plan for this? I don't think there are any CI servers setup for this target.

dcci:
After I get this working (there are still some build failures), I'll try running the test and try to set up a buildbot for this (but I think we're still away from it).
More patches/pull request will follow in the next days/weeks, I'd just like to merge small patches rather than carrying a big patch around (which will be hard to review, as well). Please let me know if you're fine with this.

parkera:
In general we are happy to take portability diffs. This one will take me a bit to review but I have no problem merging it in. The only caveat is as @phausler mentioned - we may in the future take PRs or make changes ourselves which cause this port to fail to build or work correctly, since we are not testing it directly ourselves.

dcci:
I see, thanks. That's why I'm trying to upstream the bits ASAP so once I'm done with the porting I can setup a buildbot myself to catch failures. But at this point, I'm fine with submitting patches once I discovered some changes broke FreeBSD support, if you're fine with that.

parkera:
Looks good to me.

PR #142, opened 2015-12-14
NSStringFromPoint/Size/Rect and NSPoint/Size/RectFromString

kovpas:


Foundation/NSGeometry.swift
```diff
@@ -508,12 +508,84 @@ public func NSIntersectsRect(aRect: NSRect, _ bRect: NSRect) -> Bool {
...
+private func _scanTwoNumbersFromGeometryString(aString: String) -> (first: Double, second: Double) {
+    guard aString.hasPrefix("{") && aString.hasSuffix("}") else {
+        fatalError("Wrong format")
+    }
+    
+    var string: NSString = aString._nsObject
+    string = string.substringWithRange(NSMakeRange(1, aString.length - 2))._nsObject
+    string = string.stringByReplacingOccurrencesOfString(" ", withString: "")._nsObject;
+    
+    let numbers = string.componentsSeparatedByString(",")
```
phausler: What about non comma separated float values?

This passes on darwin
```swift
let point = NSPointFromString("4q5")
XCTAssertEqual(point, NSMakePoint(4, 5))
```

Foundation/NSGeometry.swift
```diff
@@ -508,12 +508,84 @@ public func NSIntersectsRect(aRect: NSRect, _ bRect: NSRect) -> Bool {
...
+    let parsedNumbers = _scanTwoNumbersFromGeometryString(aString)
+    
+    let x = parsedNumbers.first
+    let y = parsedNumbers.second
+    let result = NSMakeSize(CGFloat(x), CGFloat(y))
+    
+    return result
+}
+
+public func NSRectFromString(aString: String) -> NSRect {
```
phausler: and here is another tricky one that works on darwin:
```swift
let point = NSRectFromString("4q5d9e3;3")
XCTAssertEqual(point, NSMakeRect(4, 5, 9000, 3))
```

Foundation/NSGeometry.swift
kovpas: 😳. Could you please share the rest of tests? Or you just know that this can pass and created this assert yourself?

phausler: It is because I have run across silliness like that before in code. Specifically the one I saw in the wild was someone was creating points as the format 4x5; which is definitely in the territory of undocumented/undefined behavior.

parkera:
I'm ok with not accepting that garbage input in this version of Foundation.

Foundation/NSGeometry.swift
parkera: I'm ok with not accepting this garbage input here.

kovpas: @parkera what about `{` and `}`? I assumed that `NSRectFromString` is the opposite to `NSStringFromRect`, but apparently `NSRectFromString` is a lot more flexible. I think I could rewrite it with `NSScanner`, so `NSRectFromString` is a bit more tolerant to input. Not sure if it's worth the effort, though.

Foundation/NSCharacterSet.swift
```diff
@@ -168,6 +168,10 @@ public class NSCharacterSet : NSObject, NSCopying, NSMutableCopying, NSCoding {
         return CFCharacterSetCreateCopy(kCFAllocatorSystemDefault, self._cfObject)
     }
     
+    public override func mutableCopy() -> AnyObject {
```
kovpas: I've got a problem with `as?` statements on linux. I already noticed it twice: in `_NSObjectRepresentableBridge` function and now in `NSObject`'s `mutableCopy`:

```
        if let copyable = self as? NSMutableCopying {
```

this condition passes on OS X and fails on linux.

This override is a workaround for this problem.

Foundation/NSGeometry.swift
```diff
@@ -11,6 +11,7 @@
     import Darwin
 #elseif os(Linux)
     import Glibc
+    import CoreFoundation
```
kovpas: @phausler is there any other way to `#include <float.h>` on linux? `DBL_EPSILON` is visible on OS X, but not on ubuntu...

TestFoundation/TestNSGeometry.swift
```diff
@@ -447,4 +452,157 @@ class TestNSGeometry : XCTestCase {
...
+        
+        let expectedRect = NSMakeRect(CGFloat(-0.6), CGFloat(-5.4), CGFloat(-105.7), CGFloat(-24.3))
+        let rect = NSRectFromString(stringRect)
+        XCTAssertEqual(expectedRect, rect,
+                       "\(NSStringFromRect(rect)) is not equal to expected \(NSStringFromRect(expectedRect))")
+        
+    }
+    
+    func test_DecodeGarbageFromNSString() {
+        var stringRect = "-0.6a5.4das-105.7bfh24.3dfas;hk312}}"
```
kovpas: Garbage is now considered as a separator. Couldn't make it work with exponential format, though. Not sure if it should be supported on `NSGeomerty` side, probably it should be implemented in `NSScanner`.

Foundation/NSCharacterSet.swift
phausler: I have filed another related issue to this https://bugs.swift.org/browse/SR-310
it is probably the same underlying problem

Foundation/NSGeometry.swift
phausler: Seems reasonable to just import that for everyone; it is likely a bug in the Glibc module map

Foundation/NSGeometry.swift
```diff
@@ -36,7 +37,7 @@ public struct CGFloat {
 extension CGFloat: Comparable { }
 
 public func ==(lhs: CGFloat, rhs: CGFloat) -> Bool {
-    return lhs.native == rhs.native
+    return abs(lhs.native - rhs.native) < DBL_EPSILON
```
phausler: albeit that this is probably what most people mean:
the definition on darwin is
```swift
public func ==(lhs: CGFloat, rhs: CGFloat) -> Bool {
  return lhs.value == rhs.value
}
```
see https://github.com/apple/swift/blob/master/stdlib/public/SDK/CoreGraphics/CoreGraphics.swift

kovpas: you're right:
https://github.com/apple/swift/blob/master/stdlib/public/SDK/CoreGraphics/CGFloat.swift.gyb#L226

The problem then is that `NSScanner` reads `0.6` `Double` as `0.600000000000000004` or 
`0.599999999999999996`, which is essentially the same number. Shall I fix the test to compare rects, points and sizes with a certain precision?

phausler: I think it is more correct to change the tests here than in the definition.

Foundation/NSCharacterSet.swift
kovpas: It looks like the problem is bigger indeed: the problem with `_NSObjectRepresentableBridge` is not only `Int`, none of the objects, that adopt `_NSObjectRepresentable` return `object` for `object as? _NSObjectRepresentable` statement.

I faced with it when I was trying to fix https://bugs.swift.org/browse/SR-251 . Tests are not passing because of this issue with protocols.

phausler: I think there seems to be a failure in all as? testing protocol adoption on linux. It might be useful to help narrow that one down with some sort of reduced test case for the stdlib/compiler team to take a look at because this will mean that we have a number of nasty runtime lurking failures scattered all over the place that we have not yet found since we don't have full coverage yet.

Foundation/NSGeometry.swift
kovpas: @phausler https://github.com/apple/swift/blob/master/stdlib/public/Glibc/module.map.in#L51
Seems that this import is commented out intentionally: `note: supplied by compiler`. But for me it doesn't work. I've just updated pull request, reverted `CGFloat` `==` to it's initial state and changed tests, so they do comparison. Could you please try to remove `import CoreFoundation` from the `TestNSGeometry:15` and see if it works for you? May be there's something wrong with my Glibc library?

Foundation/NSGeometry.swift
kovpas: Done

Foundation/NSCharacterSet.swift
kovpas: I'll take a look if I can narrow it down.

PR #143, opened 2015-12-14
Enable CFUUIDCreate on linux

kovpas:


TestFoundation/TestNSProcessInfo.swift
```diff
@@ -53,4 +54,16 @@ class TestNSProcessInfo : XCTestCase {
         processInfo.processName = originalProcessName
         XCTAssertEqual(processInfo.processName, originalProcessName, "\"\(processInfo.processName)\" not equal to \"\(originalProcessName)\"")
     }
+    
+    func test_globallyUniqueString() {
+        let uuid = NSProcessInfo.processInfo().globallyUniqueString
+        let parts = uuid._bridgeToObject().componentsSeparatedByString("-")
```
parkera: There should be no need to use the internal _bridgeToObject method here. Is it just to get to the NSString API?

kovpas: Yep, it's to get to `componentsSeparatedByString`. What should I use instead?

Also `_bridgeToObject` is public:

```
extension String : _ObjectTypeBridgeable {
    public func _bridgeToObject() -> NSString {
        return NSString(self)
    }
...
}
```

parkera:
Hm, this is sometimes hanging on Linux for me here:

```
* thread #1: tid = 19079, 0x00007ffff6a87df0 libpthread.so.0`__libc_read + 16 at syscall-template.S:81, name = 'TestFoundation', stop reason = signal SIGSTOP
  * frame #0: 0x00007ffff6a87df0 libpthread.so.0`__libc_read + 16 at syscall-template.S:81
    frame #1: 0x00007ffff725e397 libFoundation.so`uuid_generate_random + 39
    frame #2: 0x00007ffff7518897 libFoundation.so`Foundation.NSUUID.init () -> Foundation.NSUUID + 55
    frame #3: 0x00007ffff75188df libFoundation.so`Foundation.NSUUID.__allocating_init () -> Foundation.NSUUID + 47
```

kovpas:
@parkera Here are results of fixe subsequent runs of the test suite:

```
Test Case 'TestNSUUID.test_Equality' passed (0.001 seconds).
Test Case 'TestNSUUID.test_Equality' passed (21.257 seconds).
Test Case 'TestNSUUID.test_Equality' passed (33.19 seconds).
Test Case 'TestNSUUID.test_Equality' passed (11.753 seconds).
Test Case 'TestNSUUID.test_Equality' passed (8.395 seconds).
```

```
Test Case 'TestNSUUID.test_description' passed (0.0 seconds).
Test Case 'TestNSUUID.test_description' passed (24.571 seconds).
Test Case 'TestNSUUID.test_description' passed (41.93 seconds).
Test Case 'TestNSUUID.test_description' passed (19.679 seconds).
Test Case 'TestNSUUID.test_description' passed (2.244 seconds).
```

Looks quite odd, seems that something is wrong with reading from `/dev/random`. Investigating the problem.

kovpas:
Seems, that `/dev/random` [can be very slow](http://superuser.com/questions/359599/why-is-my-dev-random-so-slow-when-using-dd) sometimes. How about using `/dev/urandom` instead? It'll give pseudo-random data even if enthropy pool is empty. Should be good enough for generating uuid.

phausler:
/dev/urandom is probably fine

TestFoundation/TestNSProcessInfo.swift
phausler: _bridgeToObject is the stub for the automatic type conversion from the compiler, if you need to bridge use the bridge() method

kovpas: @phausler Yep, that's what I ended up with. Thanks!

parkera:
Much better, thanks.

PR #144, opened 2015-12-14
Return copy of path

kovpas:
I'm not 100% sure about this one, primarily because it doesn't crash on OS X, however this fixes [this crash](https://github.com/apple/swift-corelibs-foundation/pull/50#discussion_r47550314) on linux for me

Also, looking at the line above (return within `#if TARGET_IPHONE_SIMULATOR`) I see that it actually creates and returns a new copy of a string - just what method's name says.


parkera:
This looks correct to me. The reason we get away with it on OS X is that the CFSTR() constant means the string cannot be overreleased. We don't have the same protections on Linux for various reasons.

PR #145, opened 2015-12-15
[NSDate] Fix descriptionWithLocale always returns the default description

thii:


Foundation/NSDate.swift
```diff
@@ -72,14 +72,51 @@ public class NSDate : NSObject, NSCopying, NSSecureCoding, NSCoding {
...
+     - Returns: A string representation of the receiver, using the given locale,
+       or if the locale argument is `nil`, in the international format
+       `YYYY-MM-DD HH:MM:SS ±HHMM`, where `±HHMM` represents the time zone
+       offset in hours and minutes from UTC (for example,
+       "2001-03-24 10:45:32 +0600")
+     */
     public func descriptionWithLocale(locale: AnyObject?) -> String {
+        let localeRef = (locale != nil) ? (locale as! NSLocale)._cfObject : CFLocaleCopyCurrent()
+        let dateFormatterRef = CFDateFormatterCreate(kCFAllocatorSystemDefault, localeRef, kCFDateFormatterFullStyle, kCFDateFormatterFullStyle)
+        CFDateFormatterSetProperty(dateFormatterRef, kCFDateFormatterTimeZone, CFTimeZoneCopySystem())
```
parkera: What Darwin Foundation does here is actually:

```c
CFTimeZoneRef = tz = CFTimeZoneCreateWithTimeIntervalFromGMT(kCFAllocatorSystemDefault, 0.0);
CFDateFormatterSetProperty(dateFormatter, kCFDateFormatterTimeZoneKey, tz);
CFDateFormatterSetFormat(dateFormatter, CFSTR("uuuu-MM-dd HH:mm:ss '+0000'"));
```

TestFoundation/TestNSDate.swift
```diff
@@ -41,6 +42,13 @@ class TestNSDate : XCTestCase {
         let d = NSDate()
         XCTAssertNotNil(d)
     }
+
+    func test_DescriptionWithLocale() {
+        let d = NSDate(timeIntervalSince1970: 0)
+        XCTAssertEqual(d.descriptionWithLocale(NSLocale(localeIdentifier: "ja_JP")), "1970年1月1日木曜日 0時00分00秒 GMT")
+        XCTAssertEqual(d.descriptionWithLocale(NSLocale(localeIdentifier: "de_DE")), "Donnerstag, 1. Januar 1970 um 00:00:00 GMT")
+        XCTAssertEqual(d.descriptionWithLocale(NSLocale(localeIdentifier: "ru_RO")), "четверг, 1 января 1970 г., 0:00:00 GMT")
+    }
```
parkera: Let's add a test for a nil locale as well. Obviously the result depends on user locale, so just verify that it returns a non-empty string.

thii:
Just updated and rebased the commit.

thii:
Whoops! Forgot to add the `test` prefix after copying the method name. Just fixed that now. Thank you for pointing it out.

Foundation/NSDate.swift
```diff
@@ -72,15 +72,51 @@ public class NSDate : NSObject, NSCopying, NSSecureCoding, NSCoding {
...
+     [NSDateFormatter](//apple_ref/occ/cl/NSDateFormatter) and
+     [Data Formatting Guide](//apple_ref/doc/uid/10000029i)),
+     and the `NSDate` methods `descriptionWithLocale:`,
+     `dateWithCalendarFormat:timeZone:`, and
+     `descriptionWithCalendarFormat:timeZone:locale:`.
+     */
     public override var description: String {
         get {
-            return CFCopyDescription(_cfObject)._swiftObject
+            let dateFormatterRef = CFDateFormatterCreate(kCFAllocatorSystemDefault, nil, kCFDateFormatterFullStyle, kCFDateFormatterFullStyle)
```
phausler: this really could just return descriptionWithLocale(nil)

Foundation/NSDate.swift
```diff
@@ -72,15 +72,51 @@ public class NSDate : NSObject, NSCopying, NSSecureCoding, NSCoding {
...
+       or if the locale argument is `nil`, in the international format
+       `YYYY-MM-DD HH:MM:SS ±HHMM`, where `±HHMM` represents the time zone
+       offset in hours and minutes from UTC (for example,
+       "2001-03-24 10:45:32 +0600")
+     */
     public func descriptionWithLocale(locale: AnyObject?) -> String {
-        return description
+        guard let aLocale = locale else { return description }
+        let dateFormatterRef = CFDateFormatterCreate(kCFAllocatorSystemDefault, (aLocale as! NSLocale)._cfObject, kCFDateFormatterFullStyle, kCFDateFormatterFullStyle)
+        CFDateFormatterSetProperty(dateFormatterRef, kCFDateFormatterTimeZoneKey, CFTimeZoneCopySystem())
```
phausler: the format is always "uuuu-MM-dd HH:mm:ss '+0000'"
and the timezone even in the description with locale is GMT

Foundation/NSDate.swift
thii: I actually did that in my previous implementation, but eventually chose this way since the 2 methods don't share much implementation.

Foundation/NSDate.swift
thii: Actually here is what I got with current Darwin Foundation (I tested with Xcode 7.2/Swift 2.1.1.). It actually returns with locale's language and time zone. Can you please check it again? 
```swift
let d = NSDate(timeIntervalSince1970: 0)
d.descriptionWithLocale(nil) // 1970-01-01 00:00:00 +0000
d.descriptionWithLocale(NSLocale(localeIdentifier: "en_US")) // Thursday, January 1, 1970 at 9:00:00 AM Japan Standard Time
d.descriptionWithLocale(NSLocale(localeIdentifier: "ja_JP")) // 1970年1月1日木曜日 9時00分00秒 日本標準時
```

phausler: you are correct. my mistake

phausler:
these funnel in reverse of the way they do on darwin but overall the implementation looks pretty reasonable to me; tests pass on linux and mac

PR #147, opened 2015-12-15
The `currentThread` method call returns different values

sgl0v:
The `currentThread` method call returns different values. We should check the `pthread_key_create` result for 0. According to documentation:

> If successful, the pthread_key_create() function shall store the newly created key value at *key and
> shall return zero. Otherwise, an error number shall be returned to indicate the error.


parkera:
Thanks!

PR #148, opened 2015-12-16
Fix pthread warnings on ubuntu

kovpas:
These warnings:
screen shot 2015-12-16 at 9 15 58 am

### DIFF
--- a/CoreFoundation/Base.subproj/CFPriv.h
+++ b/CoreFoundation/Base.subproj/CFPriv.h
@@ -563,9 +563,9 @@ CF_EXPORT CFMessagePortRef _CFMessagePortCreateLocalEx(CFAllocatorRef allocator,
 
 #endif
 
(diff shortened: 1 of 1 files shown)


phausler:
very strange that is the cause of the failure. I wonder if this is a module map bug in the glibc side.

kovpas:
@phausler no, this is not the reason, this just fixes bunch of warnings for Linux build. I decided to make a separate pull request, as this is not related to any other.

I think that the reason of the failure was a `/dev/random` problem. Can't prove it, though, as it never really failed for me. Tests just took a lot of time to pass.

phausler:
Technically the module map importer in swift should treat #include <pthread.h> to be only applied once, e.g. including string.h etc. I would definitely like to remove the warnings on linux but it means some follow-up will need to be done with the SwiftGlibc module map owner to verify that somehow this is correct behavior (it seems that there is something else lurking under the waters here)

phausler:
full clean build on linux seems to pass building.

PR #149, opened 2015-12-16
Change TestNSCharacterSet.test_Predefines

kovpas:
This test used to take 2 minutes on my (virtual) linux machine. On OS X it used to be 2.5 seconds, which is acceptable.
screen shot 2015-12-16 at 9 13 02 am

Also here's a list of tests that are also relatively slow to run:

```
Test Case 'TestNSURL.test_URLStrings' passed (7.307 seconds).
Test Case 'TestNSURLComponents.test_string' passed (14.938 seconds).
```

### DIFF
--- a/TestFoundation/TestNSCharacterSet.swift
+++ b/TestFoundation/TestNSCharacterSet.swift
@@ -52,12 +52,9 @@ class TestNSCharacterSet : XCTestCase {
         
         let mcset = NSMutableCharacterSet.whitespaceAndNewlineCharacterSet()
         let cset2 = NSCharacterSet.whitespaceAndNewlineCharacterSet()
-        let count = UInt32(0x10FFFF)
-        for var idx = UInt32(0); idx < count; idx++ {
-            let a = mcset.longCharacterIsMember(idx)
-            let b = cset2.longCharacterIsMember(idx)
-            XCTAssertEqual(a, b, "\(String(idx, radix: 16)) \(a ? "exists in" : "does not exist in") mutable and \(b ? "exists in" : "does not exist in") immutable character set")
-        }
+
+        XCTAssert(mcset.isSupersetOfSet(cset2))
+        XCTAssert(cset2.isSupersetOfSet(mcset))
         
         XCTAssertTrue(NSCharacterSet.whitespaceAndNewlineCharacterSet().isSupersetOfSet(NSCharacterSet.newlineCharacterSet()), "whitespace and newline should be a superset of newline")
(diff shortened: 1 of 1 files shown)


phausler:
I agree that we should not have long tests (anything over a second is suspect in my book)
That test was a transliteration from objc. it would be interesting to understand why it is slower on linux than on mac os x and why it is slower on both than in the objc version.

kovpas:
@phausler Well, partly this could be because I run Linux on a virtual machine. I'd assume, that if it was as slow for everyone as it was for me, someone would have fixed this before I did that...

phausler:
I am using a linux server which some pretty impressive stats and it is running slower than my mac running the objc version of that test so there seems to be a bug somewhere in there.

kovpas:
@phausler I suspect that something is wrong with `for` loop performance on linux. Even two empty loops take significantly different time to run. I'll take a closer look at this later today.

parkera:
Interesting, thanks @kovpas.

kovpas:
There's no problem with `for` loop or any other loop performance on Linux in comparison to OS X. I've just run `idx++` in loop, there's difference in performance.

There's a problem with `XCTAssert`, which adds some overhead on both platforms. Here are two tests:

```
        for var idx = UInt32(0); idx < count; idx++ {
            let a = idx + 1 > 0 // to make sure there's something real to compute 
            let b = idx + 2 > 0 // on every iteration and compiler doesn't optimize it
            XCTAssertEqual(a, b)
        }
```

and 

```
        for var idx = UInt32(0); idx < count; idx++ {
            let a = idx + 1 > 0
            let b = idx + 2 > 0
            if a != b {
                print("never happens")
            }
        }
```

Results:
OS X: `0.774` and `0.011` sec
Linux: `3.179` and `0.027`

In this case, I can believe that this ~3-4x difference in both cases could be related to the fact, that I run linux on a virtual machine.

However, there's a problem with strings: they are created on every single invocation of XCTAssertEqual. 

If we replace 
`XCTAssertEqual(a, b)` 
in the first test with 
`XCTAssertEqual(a, b, "\(String(idx, radix: 16)) \(a ? "exists in" : "does not exist in") mutable and \(b ? "exists in" : "does not exist in") immutable character set")` 

result will be `2.109` on OS X and `119.241` on Linux. Which is clearly not 3x difference.

kovpas:
@parkera @phausler I think i found a workaround: https://github.com/apple/swift-corelibs-xctest/pull/17

Still not sure why os x is sooooo much faster than linux. Ternary operator is handled differently may be? Really not sure.

But if the pull request for xctest will be accepted, I'll revert this change.

PR #150, opened 2015-12-16
Fix NSAffineTransform

nubbel:
Hi,

I just noticed that my implementation of `NSAffineTransform` (added in #93) was applying transformations in the opposite order than Darwin's.
My previous implementation assumed that the properties of `NSAffineTransform` represent a matrix that looks like this
```
[ m11  m12  tX ]
[ m21  m22  tY ]
[  0    0    1 ]
```
and applied transformations by `T * p` where `T` is the transformation matrix and `p` a column vector.

As it turns out, however, the matrix rather looks like that
```
[ m11  m12  0 ]
[ m21  m22  0 ]
[  tX   tY  1 ]
```
and transformation are applied by `p' * T` where `p'` is a row vector.

This has direct influence on the order of application, because `T * p = S * R * p` would apply `R` before `S`, whereas `p' * T = p' * S * R` would apply `S` before `R`.

I've also added tests for…
- `appendTransform` and `prependTransform`
- the direction of rotation (that was basically just a lucky guess)
- composed transformations

I apologise for that!

### DIFF
--- a/Foundation/NSAffineTransform.swift
+++ b/Foundation/NSAffineTransform.swift
@@ -68,19 +68,19 @@ public class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
     public func translateXBy(deltaX: CGFloat, yBy deltaY: CGFloat) {
         let translation = NSAffineTransformStruct.translation(tX: deltaX, tY: deltaY)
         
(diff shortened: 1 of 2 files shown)


parkera:
Thanks @nubbel!

PR #151, opened 2015-12-16
Update constants

kovpas:
As we're now using `_GNU_SOURCE` in the compiler flags for Linux, it's ok to remove `__USE_GNU` from `CFBundle_Binary.c`

Also, I replaced string constants with CF ones, so they could be defined in one place. There are still similar (`let NSConst: String = "kCFConst"`) defines in `NSStream`, but `CFStream` is not compatible with linux yet.

### DIFF
--- a/Foundation/NSLocale.swift
+++ b/Foundation/NSLocale.swift
@@ -168,7 +168,7 @@ public enum NSLocaleLanguageDirection : UInt {
     case BottomToTop
 }
 
-public let NSCurrentLocaleDidChangeNotification: String = "" // NSUnimplemented
+public let NSCurrentLocaleDidChangeNotification: String = "kCFLocaleCurrentLocaleDidChangeNotification"
 
 public let NSLocaleIdentifier: String = "kCFLocaleIdentifierKey"
 public let NSLocaleLanguageCode: String = "kCFLocaleLanguageCodeKey"
@@ -184,11 +184,11 @@ public let NSLocaleDecimalSeparator: String = "kCFLocaleDecimalSeparatorKey"
 public let NSLocaleGroupingSeparator: String = "kCFLocaleGroupingSeparatorKey"
 public let NSLocaleCurrencySymbol: String = "kCFLocaleCurrencySymbolKey"
 public let NSLocaleCurrencyCode: String = "currency"
-public let NSLocaleCollatorIdentifier: String = "" // NSUnimplemented // NSString
-public let NSLocaleQuotationBeginDelimiterKey: String = "" // NSUnimplemented // NSString
-public let NSLocaleQuotationEndDelimiterKey: String = "" // NSUnimplemented // NSString
-public let NSLocaleAlternateQuotationBeginDelimiterKey: String = "" // NSUnimplemented // NSString
-public let NSLocaleAlternateQuotationEndDelimiterKey: String = "" // NSUnimplemented // NSString
+public let NSLocaleCollatorIdentifier: String = "kCFLocaleCollatorIdentifierKey"
+public let NSLocaleQuotationBeginDelimiterKey: String = "kCFLocaleQuotationBeginDelimiterKey"
+public let NSLocaleQuotationEndDelimiterKey: String = "kCFLocaleQuotationEndDelimiterKey"
+public let NSLocaleAlternateQuotationBeginDelimiterKey: String = "kCFLocaleAlternateQuotationBeginDelimiterKey"
+public let NSLocaleAlternateQuotationEndDelimiterKey: String = "kCFLocaleAlternateQuotationEndDelimiterKey"
 
 extension CFLocaleRef : _NSBridgable {
     typealias NSType = NSLocale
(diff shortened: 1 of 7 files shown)


phausler:
There is a problem with the bridge usage here since it will add considerable warm up time for initializing these values. each one will have to traverse the bridging which may not be fully initialized at that point (it is a bit unclear when this exactly happens in relation to __CFInitializeSwift's call point. I would like to know more about what is the consequence here before moving forward with the constant values.
If you can pull the __USE_GNU into another pull request I can merge that in.

kovpas:
@phausler done. Changed back CF constants to strings.

Foundation/NSLocale.swift
```diff
@@ -177,18 +177,18 @@ public let NSLocaleScriptCode: String = "kCFLocaleScriptCodeKey"
 public let NSLocaleVariantCode: String = "kCFLocaleVariantCodeKey"
 public let NSLocaleExemplarCharacterSet: String = "kCFLocaleExemplarCharacterSetKey"
 public let NSLocaleCalendar: String = "kCFLocaleCalendarKey"
-public let NSLocaleCollationIdentifier: String = "collation"
+public let NSLocaleCollationIdentifier: String = "kCFLocaleCollationIdentifierKey"
```
phausler: kCFLocaleCollationIdentifierKey is defined as "collation"

Foundation/NSLocale.swift
```diff
@@ -177,18 +177,18 @@ public let NSLocaleScriptCode: String = "kCFLocaleScriptCodeKey"
...
 public let NSLocaleDecimalSeparator: String = "kCFLocaleDecimalSeparatorKey"
 public let NSLocaleGroupingSeparator: String = "kCFLocaleGroupingSeparatorKey"
 public let NSLocaleCurrencySymbol: String = "kCFLocaleCurrencySymbolKey"
-public let NSLocaleCurrencyCode: String = "currency"
-public let NSLocaleCollatorIdentifier: String = "" // NSUnimplemented // NSString
-public let NSLocaleQuotationBeginDelimiterKey: String = "" // NSUnimplemented // NSString
-public let NSLocaleQuotationEndDelimiterKey: String = "" // NSUnimplemented // NSString
-public let NSLocaleAlternateQuotationBeginDelimiterKey: String = "" // NSUnimplemented // NSString
-public let NSLocaleAlternateQuotationEndDelimiterKey: String = "" // NSUnimplemented // NSString
+public let NSLocaleCurrencyCode: String = "kCFLocaleCurrencyCodeKey"
```
phausler: kCFLocaleCurrencyCodeKey is defined as "currency"

Foundation/NSLocale.swift
kovpas: well, this is consistent and not confusing at all 😄

Foundation/NSLocale.swift
kovpas: I'm wondering if this should be covered with tests actually. Because I see other people doing the same mistake. Something like 
`XCTAssertEqual(NSLocaleCurrencyCode, kCFLocaleCurrencySymbolKey._swiftObject);`
What do you think?

phausler: Tests are almost always a good thing! I'd say that verifying the constants are quite reasonable to validate since we have a nearly concrete obligation to keep them the same (for archiving and other fringe cases) for the darwin version of Foundation.

kovpas: Added tests. Should be good now. The only thing: I couldn't get access for `kCFLocaleCurrentLocaleDidChangeNotification` from tests. It's accessible from `Foundation`, though. Do you know how to export it properly?

Foundation/NSRunLoop.swift
```diff
@@ -77,7 +77,7 @@ extension NSRunLoop {
     public func runMode(mode: String, beforeDate limitDate: NSDate) -> Bool {
         let runloopResult = CFRunLoopRunInMode(mode._cfObject, limitDate.timeIntervalSinceNow, false)
 #if os(Linux)
-        return runloopResult == 2 || runloopResult == 3
+        return runloopResult == kCFRunLoopRunHandledSource || runloopResult == kCFRunLoopRunTimedOut
```
phausler: getting this error:
```
Foundation/NSRunLoop.swift:80:30: error: binary operator '==' cannot be applied to operands of type 'CFRunLoopRunResult' (aka 'Int32') and 'Int'
        return runloopResult == kCFRunLoopRunHandledSource || runloopResult == kCFRunLoopRunTimedOut
               ~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~
Foundation/NSRunLoop.swift:80:30: note: overloads for '==' exist with these partially matching parameter lists: (Int32, Int32), (Int, Int)
        return runloopResult == kCFRunLoopRunHandledSource || runloopResult == kCFRunLoopRunTimedOut
```

kovpas: Crap. I remember I fixed that. Probably rebase/merge/revert problem. Sorry, will update.

PR #152, opened 2015-12-16
Fixes a shadowed variable in the system time zone initializer.

brownleej:
The shadowed variable that's being removed in this PR causes the name to fall out of scope as soon as it exits the "if (0 < ret)" block, and the outermost "name" field is left null. This causes the systemTimeZone to always report as GMT.

I'm not sure if the mechanism I'm using to test this is the best way to go. Calling into the C time functions seems like a hack, and something that might get broken by subtle changes to system configurations. It works on both my Mac and Linux installs, and it correctly gives an error when the change to CFTimeZone.c is removed.


phausler:
oops, hmm perhaps we should enable error on shadowing
Validating against the C api's is perfectly reasonable in a unit test.

PR #153, opened 2015-12-16
Forces the tests for calendar date conversion to use UTC instead of system time.

brownleej:
This is a fix for SR-171.


phausler:
looks good to me, as a good rule of thumb for the darwin foundation tests we prefer to have local machine independent tests if at all possible (unless we are testing the local machine setting of course)

PR #154, opened 2015-12-17
Remove ++ and -- operators

nRewik:
[Remove the ++ and -- operators](https://github.com/apple/swift-evolution/blob/master/proposals/0004-remove-pre-post-inc-decrement.md) proposal has been accepted.

++ and -- operators will be removed from Swift 3.0. So, let's change it now :smile:


lattner:
Awesome, thank you for doing this.  I will be turning on this warning soon, so this is very helpful.  The patch LGTM, but a corelibs reviewer should make the call.

parkera:
Thanks @nRewik!

PR #155, opened 2015-12-17
Remove empty line in NSObject.swift

onevcat:
Just a trivial fix. Sorry for my code mysophobia.


phausler:
I would have figured the fear of spaces would be grouped in agoraphobia.

onevcat:
I guess you are right. XD

PR #156, opened 2015-12-17
Implement NSString.completePathIntoString.

HeMet:
Implement NSString.completePathIntoString. Tests included.

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -360,26 +360,16 @@ public extension NSString {
         
         var isDirectory = false
         let isAbsolutePath = NSFileManager.defaultManager().fileExistsAtPath(_storage, isDirectory: &isDirectory)
+        let searchAllFilesInDirectory = isAbsolutePath && isDirectory
         
-        if isAbsolutePath && !isDirectory {
-            if types.isEmpty || types.contains(url.pathExtension ?? "") {
-                outputName = self
-                outputArray = [self]
-                return 1
-            } else {
-                return 0
-            }
-        }
-        
-        guard let urlWhereToSearch = isAbsolutePath ? url : url.URLByDeletingLastPathComponent else {
+        guard let urlWhereToSearch = searchAllFilesInDirectory ? url : url.URLByDeletingLastPathComponent else {
             return 0
         }
         
         var matches: [String] = []
-        var matchSuffixes: [String] = []
         
         let namePrefix = url.lastPathComponent ?? ""
-        
+
         let enumerator = NSFileManager.defaultManager().enumeratorAtURL(urlWhereToSearch, includingPropertiesForKeys: nil, options: .SkipsSubdirectoryDescendants, errorHandler: nil)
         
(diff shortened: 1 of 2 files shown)


Foundation/NSPathUtilities.swift
```diff
@@ -345,7 +345,100 @@ public extension NSString {
...
+        if let lcp = longestCommonPrefix(matchSuffixes, caseSensitive: flag) {
+           outputName = (urlWhereToSearch.absoluteString! + lcp).bridge()
+        }
+        
+        outputArray = matches.map({ $0.bridge() })
+        
+        return matches.count
+    }
+    
+    func longestCommonPrefix(strings: [String], caseSensitive: Bool) -> String? {
```
phausler: presuming this is just a utility method, it probably should be explicitly labeled as private/internal and guarded by a _ for immediate indication that it is an internal method

Foundation/NSPathUtilities.swift
```diff
@@ -345,7 +345,100 @@ public extension NSString {
     /// - Experiment: This is a draft API currently under consideration for official import into Foundation
     /// - Note: Since this API is under consideration it may be either removed or revised in the near future
     public func completePathIntoString(inout outputName: NSString?, caseSensitive flag: Bool, inout matchesIntoArray outputArray: [NSString], filterTypes: [String]?) -> Int {
```
phausler: so this method should handle a few cases: what if it starts with a "~"? consider both the case of "~/" and "~"
also isn't there a fast path case? where the input is ""

HeMet: I believe `NSURL(string: _storage)` really should be `NSURL(fileURLWithPath: _storage)` so it could be standardized, but core method of `NSURL` that does this is not implemented yet. There is `FIXME` comment but it is not very clear it seems. For now only paths without any symlinks etc supported.

TestFoundation/TestNSString.swift
```diff
@@ -337,6 +338,113 @@ class TestNSString : XCTestCase {
...
+            "/tmp/test_completePathIntoString_03.DAT"
+        ]
+        
+        guard ensureFiles(fileNames) else {
+            XCTAssert(false, "Could not create temp files for testing.")
+            return
+        }
+        
+        do {
+            let path: NSString = "/private/tmp/" // no symlink support yet
```
phausler: this will not work as expected on linux since /tmp is not a symlink there

HeMet:
Should work on Linux now.

phausler:
yep looks good on ubuntu; I am going to merge this in; we should probably make a ticket on bugs.swift.org about the NSURL thing 'cause i am certain that will come to bite us here soon.

phausler:
I did need to disable one of the asserts for the tests since it was failing on the CI system; I added https://bugs.swift.org/browse/SR-389 as a follow up issue

HeMet:
@phausler I made ticket about path standardizing and symlinks resolving https://bugs.swift.org/browse/SR-393

PR #158, opened 2015-12-17
Get rid of c-style for loops

kovpas:
This gets rid of eight (almost deprecated 😀) old style for loops.


phausler:
these look correct in that they don't seem to change any behavior

PR #159, opened 2015-12-18
Adding support for ARMv7 on linux

hpux735:
These changes begin support for ARMv7 and linux.  There are a few optimizations yet to be completed (it's falling back to soft floating point), and there seems to be assumptions about the size of ssize made in some swift code. (for example: Foundation/NSXMLParser.swift:238)

### DIFF
--- a/CoreFoundation/Base.subproj/CFBase.c
+++ b/CoreFoundation/Base.subproj/CFBase.c
@@ -875,14 +875,16 @@ void _CFRuntimeSetCFMPresent(void *addr) {
 
 
 extern void __HALT() {
-#if defined(__ppc__)
+#if defined(__ppc__) || defined(__arm__)
     __asm__("trap");
+
 #elif defined(__i386__) || defined(__x86_64__)
 #if defined(_MSC_VER)
     __asm int 3;
 #else
     __asm__("int3");
 #endif
+
 #endif
 }
 
--- a/CoreFoundation/Base.subproj/CFBase.c
+++ b/CoreFoundation/Base.subproj/CFBase.c
@@ -875,17 +875,7 @@ void _CFRuntimeSetCFMPresent(void *addr) {
 
 
 extern void __HALT() {
-#if defined(__ppc__) || defined(__arm__)
-    __asm__("trap");
-
-#elif defined(__i386__) || defined(__x86_64__)
-#if defined(_MSC_VER)
-    __asm int 3;
-#else
-    __asm__("int3");
-#endif
-
-#endif
+  __builtin_trap();
 }
 
 
--- a/Foundation/NSXMLNodeOptions.swift
+++ b/Foundation/NSXMLNodeOptions.swift
@@ -46,46 +46,46 @@
 public var NSXMLNodeOptionsNone: Int { return 0 }
 
 // Init
-public var NSXMLNodeIsCDATA: UInt { return 1 << 0 }
-public var NSXMLNodeExpandEmptyElement: UInt { return 1 << 1 } // <a></a>
-public var NSXMLNodeCompactEmptyElement: UInt { return 1 << 2 } // <a/>
-public var NSXMLNodeUseSingleQuotes: UInt { return 1 << 3 }
-public var NSXMLNodeUseDoubleQuotes: UInt { return 1 << 4 }
-public var NSXMLNodeNeverEscapeContents: UInt { return 1 << 5 }
+public var NSXMLNodeIsCDATA: Int { return 1 << 0 }
+public var NSXMLNodeExpandEmptyElement: Int { return 1 << 1 } // <a></a>
+public var NSXMLNodeCompactEmptyElement: Int { return 1 << 2 } // <a/>
+public var NSXMLNodeUseSingleQuotes: Int { return 1 << 3 }
+public var NSXMLNodeUseDoubleQuotes: Int { return 1 << 4 }
+public var NSXMLNodeNeverEscapeContents: Int { return 1 << 5 }
 
 // Tidy
-public var NSXMLDocumentTidyHTML: UInt { return 1 << 9 }
-public var NSXMLDocumentTidyXML: UInt { return 1 << 10 }
+public var NSXMLDocumentTidyHTML: Int { return 1 << 9 }
+public var NSXMLDocumentTidyXML: Int { return 1 << 10 }
 
 // Validate
-public var NSXMLDocumentValidate: UInt { return 1 << 13 }
+public var NSXMLDocumentValidate: Int { return 1 << 13 }
 
 // External Entity Loading
 // Choose only zero or one option. Choosing none results in system-default behavior.
-public var NSXMLNodeLoadExternalEntitiesAlways: UInt { return 1 << 14 }
-public var NSXMLNodeLoadExternalEntitiesSameOriginOnly: UInt { return 1 << 15 }
-public var NSXMLNodeLoadExternalEntitiesNever: UInt { return 1 << 19 }
+public var NSXMLNodeLoadExternalEntitiesAlways: Int { return 1 << 14 }
+public var NSXMLNodeLoadExternalEntitiesSameOriginOnly: Int { return 1 << 15 }
+public var NSXMLNodeLoadExternalEntitiesNever: Int { return 1 << 19 }
 
 // Parse
-public var NSXMLDocumentXInclude: UInt { return 1 << 16 }
+public var NSXMLDocumentXInclude: Int { return 1 << 16 }
 
 // Output
-public var NSXMLNodePrettyPrint: UInt { return 1 << 17 }
-public var NSXMLDocumentIncludeContentTypeDeclaration: UInt { return 1 << 18 }
+public var NSXMLNodePrettyPrint: Int { return 1 << 17 }
+public var NSXMLDocumentIncludeContentTypeDeclaration: Int { return 1 << 18 }
 
 // Fidelity
-public var NSXMLNodePreserveNamespaceOrder: UInt { return 1 << 20 }
-public var NSXMLNodePreserveAttributeOrder: UInt { return 1 << 21 }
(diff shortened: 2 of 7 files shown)


lib/target.py
```diff
@@ -223,8 +311,8 @@ class Vendor:
 
 class Target:
     triple = None
-    sdk = OSType.MacOSX
-    arch = ArchType.x86_64
+    sdk = OSType.UnknownOS
+    arch = ArchType.UnknownArch
     executable_suffix = ""
     dynamic_library_prefix = "lib"
     dynamic_library_suffix = ".dylib"
```
phausler: these are now incorrect if the sdk and arch are defaulted to unknown

lib/target.py
```diff
@@ -51,11 +51,99 @@ class ArchType:
...
     kalimba     = 38
     shave       = 39
-
+# Do not assume that these are 1:1 mapping.  This should follow
+# canonical naming conventions for arm, etc. architectures.
+# See apple/swift PR #608    
+    @staticmethod
+    def to_string(value):
+        if value == ArchType.arm:
+            return "armv7"
```
phausler: isn't armv7a the more common version?

lib/target.py
```diff
@@ -51,11 +51,99 @@ class ArchType:
...
+        return "unknown"
+# Not 1:1, See to_string
     @staticmethod
     def from_string(string):
         if string == "arm":
             return ArchType.arm
+        if string == "armv7":
+            return ArchType.arm
+        if string == "armv7l":
+            return ArchType.arm
```
phausler: i have previously used armv5 armv5tejl armv6 and armv7te; so perhaps this should actually just check a prefix

Foundation/NSXMLNodeOptions.swift
```diff
@@ -46,46 +46,46 @@
 public var NSXMLNodeOptionsNone: Int { return 0 }
 
 // Init
-public var NSXMLNodeIsCDATA: Int { return 1 << 0 }
-public var NSXMLNodeExpandEmptyElement: Int { return 1 << 1 } // <a></a>
-public var NSXMLNodeCompactEmptyElement: Int { return 1 << 2 } // <a/>
-public var NSXMLNodeUseSingleQuotes: Int { return 1 << 3 }
-public var NSXMLNodeUseDoubleQuotes: Int { return 1 << 4 }
-public var NSXMLNodeNeverEscapeContents: Int { return 1 << 5 }
+public var NSXMLNodeIsCDATA: UInt { return 1 << 0 }
```
phausler: This changes APIs can this be split out to a separate commit? They also present a problem in the regards that UInt does not properly translate to usage patterns in swift since all NSUInteger values are converted to Int not UInt by the importer

CoreFoundation/Base.subproj/CFBase.c
```diff
@@ -875,14 +875,16 @@ void _CFRuntimeSetCFMPresent(void *addr) {
 
 
 extern void __HALT() {
-#if defined(__ppc__)
+#if defined(__ppc__) || defined(__arm__)
```
phausler: actually i think we could probably generalize this into __builtin_trap() since we require clang

lib/target.py
hpux735: I think I added the appropriate logic to compensate below.  The only issue I have with setting default values is that it increased the time it took to identify the logic bug that "ArchType.from_string(comps[0])" wasn't actually assigning anything, it was essentially a no-op.  The default value made it look like everything was fine, just wrong.

phausler: fair point; I agree having an opinion about a "default" target is probably incorrect in a general sense. Just was thinking that perhaps it might be good to have these as None or some other thing to avoid any potential other trip-ups

lib/target.py
```diff
@@ -242,20 +328,30 @@ def __init__(self, triple):
...
 
     @staticmethod
     def default():
-        triple = platform.machine() + "-"
+        arch = ArchType.from_string(platform.machine())
+        triple = ArchType.to_string(arch)
         if platform.system() == "Linux":
-            triple += "linux-gnu"
+            if arch == ArchType.arm:
+                triple += "-linux-gnueabihf"
```
phausler: what about "-linux-armeabi" or "-linux-android" etc? I know these are other handled cases in llvm's triple calculations

lib/target.py
hpux735: Good call.  I'll do that.

lib/target.py
hpux735: Yes, I don't know the best way to handle that.  Do you know a good way to detect that part of the triple would be?

phausler: See http://clang.llvm.org/docs/CrossCompilation.html

arch = x86, arm, thumb, mips, etc.
sub = for ex. on ARM: v5, v6m, v7a, v7m, etc.
vendor = pc, apple, nvidia, ibm, etc.
sys = none, linux, win32, darwin, cuda, etc.
abi = eabi, gnu, android, macho, elf, etc.

So we probably need another field since triples are not three parts and instead five parts with three reasonably optional fields.
arch[sub][-vendor]-sys[-abi]

However this method is the script's host not the target. So I am not certain which hosts we want to really support here.

Foundation/NSXMLParser.swift
```diff
@@ -233,9 +233,9 @@ internal func _NSXMLParserStartElementNs(ctx: _CFXMLInterface, localname: Unsafe
     let parser = ctx.parser
     let reportQNameURI = parser.shouldProcessNamespaces
     let reportNamespaces = parser.shouldReportNamespacePrefixes
-    let prefixLen = prefix == nil ? strlen(UnsafePointer<Int8>(prefix)) : 0
+    let prefixLen = prefix == nil ? UInt(strlen(UnsafePointer<Int8>(prefix))) : 0
```
phausler: isn't the declaration of strlen
`size_t strlen(const char *s)`
which size_t should be unsigned in any reasonable environment typedefed by `__SIZE_TYPE__` hopefully or `unsigned long` if that isn't the case

Which os defines it as `int strlen(const char *s)`?

hpux735: In swift/stdlib/public/SwiftShims/LibcShims.h swift_ssize_t is defined
```
#if defined(__linux__) && defined (__arm__)
typedef      int __swift_ssize_t;
#else
 typedef long int __swift_ssize_t;
#endif
```

 and later in swift/stdlib/public/stubs/LibcShims.cpp it is checked with a static assert to ensure that swift_ssize_t matches whatever the C ssize_t is.  Because of this, the size of ssize_t cannot be assumed to be a long int.

```
static_assert(std::is_same<ssize_t, swift::__swift_ssize_t>::value,
              "__swift_ssize_t is wrong");
```

I definitely don't think that this commit is the right way to do it.  This is another case of trying to get it to finish building to check linking.  I can revert this to what's existing for the purposes of this pull request.

phausler: nah; if that is the current definition then lets roll with it and just submit a bug report on bugs.swift.org with perhaps a comment in the top of this file to explain why that is the case.

hpux735: Done: https://bugs.swift.org/browse/SR-314

phausler:
during the build it seems that it is spewing out a warning
```
[174/176] CompileSwift: Foundation/NSURLError.swift
<unknown>:0: warning: argument unused during compilation: '-mcpu=cortex-a15'
```
Is this because I am missing some other commit from the compiler?

hpux735:
There is a compiler issue at play there, but it isn't resolved yet.  I backed out.

phausler:
That warning was actually from a x86_64 host building for a x86_64 target; ubuntu 14

hpux735:
It's nice that the compiler was smart enough to ignore obviously contradictory information! :)

phausler:
full clean build for the entire toolchain + foundation seems to pass just fine on ubuntu and on mac os x; the only caveat here is that this falls under the same domain as the FreeBSD port in that this is not a regularly tested path so there may be things that break since we won't be testing there.

hpux735:
Yep, that's totally expected.  Thanks!!

Foundation/NSXMLParser.swift
hpux735: Apparently this is intended behavior.  Jordan Rose closed the bug report.

gribozavr: `strlen()` is returning `size_t`, not `ssize_t` (note the double s).  I think this issue is worth further investigation.

hpux735: Although it's true that size_t (and swift_size_t) is distinct from swift_ssize_t it still changes size based upon the underlying architecture.  It is ultimately typedef'd from __SIZE_TYPE__ which, although I couldn't find its actual definition) is 8 bytes on x86_64 (linux or mac) and 4 bytes on 32-bit arm.

I still don't think this is the best solution in this particular instance.  The error we're addressing here is:

error: cannot convert value of type 'UInt32' to expected argument type 'UInt'

So, it is my expectation that we're promoting a 32-bit unsigned into to a 64-bit unsigned int, or simply removing the bit-size specialization, with it remaining 32-bits.  Either way, it's safe in the 32-bit case.  In 64-bit code, this should be a no-op, because it's a cast from UInt to UInt.

gribozavr: `size_t` is supposed to be imported as `Int` in Swift, the importer has a special case for it.  Would you mind checking if there is an issue with that on Linux/armv7?

```Swift
import Glibc
print(size_t.self) // should print 'Int'
```

hpux735: Yes,  It is.

```
wdillon@tegra-ubuntu:~$ cat size_t.swift 
import Glibc
print(size_t.self) // should print 'Int'

wdillon@tegra-ubuntu:~$ swiftc size_t.swift 
wdillon@tegra-ubuntu:~$ ./size_t 
Int
wdillon@tegra-ubuntu:~$ uname -s -m
Linux armv7l
```

At some pint strlen's return value becomes Uint32.

PR #160, opened 2015-12-18
Fix `NSString.stringByTrimmingCharactersInSet` behavior

briancroom:
Previously it would always return just the first character in the string due
to broken location reporting in `_NSStringBuffer`


phausler:
oops, good catch!

PR #161, opened 2015-12-18
Implement NSNumber.compare

briancroom:
Including tests. Note the one `NSNumber(unsignedChar:)` assertion that is commented out for
the same reason as the series of assertions in test_numberWithUnsignedChar


phausler:
looks good to me; nice coverage!

PR #163, opened 2015-12-19
NSArray sorted array from range fix

HeMet:
Issue was method ignored range argument.

### DIFF
--- a/TestFoundation/TestNSArray.swift
+++ b/TestFoundation/TestNSArray.swift
@@ -13,7 +13,7 @@
 import Foundation
 import XCTest
 #else
-import SwiftFoundation
+@testable import SwiftFoundation
 import SwiftXCTest
 #endif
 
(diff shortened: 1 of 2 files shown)


HeMet:
Why result of `toRange` is optional? `Range(start:end:)` never fails.
```swift
extension _NSRange {    
    @warn_unused_result
    public func toRange() -> Range<Int>? {
        return Range<Int>(start: location, end: location + length)
    }
}
```

TestFoundation/TestNSArray.swift
```diff
@@ -13,7 +13,7 @@
 import Foundation
 import XCTest
 #else
-import SwiftFoundation
+@testable import SwiftFoundation
```
phausler: we cannot use `@tesable` currently due to CI limitations

phausler:
from the standard lib: (which should be in Foundation)
```swift
extension NSRange {
  public init(_ x: Range<Int>) {
    location = x.startIndex
    length = x.count
  }

  @warn_unused_result
  public func toRange() -> Range<Int>? {
    if location == NSNotFound { return nil }
    return Range(start: location, end: location + length)
  }
}
```

PR #164, opened 2015-12-19
Added implementation of NSPort Initializer

KBryan:


Foundation/NSString.swift
```diff
@@ -270,7 +270,7 @@ extension NSString {
     }
     
     public func compare(string: String, options mask: NSStringCompareOptions, range compareRange: NSRange) -> NSComparisonResult {
-        return compare(string, options: [], range:compareRange, locale: nil)
+        return compare(string, options: mask, range:compareRange, locale: nil)
```
phausler: The commit/pull request seems to be named differently than the implementation here, perhaps it should be named something along the lines of "pass mask options to locale based comparison" or something like that?

phausler:
So firstly initWithCoder should call super.init in most (if not all) of these cases and additionally elements like for example NSUUID should implement proper initializers.

In that specific case it should code bytes for the key "NS.uuidbytes" and should not tolerate decoders that do not allow keyed coding. and then when it has the bytes decoded it should then init with those bytes as a funnel to the initWithUUIDBytes method.

parkera:
Going to go ahead and close this for now; please open a new PR once comments are addressed.

PR #165, opened 2015-12-19
Implemented various sorting methods from the NSMutableArray class

sgl0v:
Implemented `sortUsingFunction`, `sortUsingComparator`, `sortWithOptions` methods from the NSMutableArray class.


TestFoundation/TestNSArray.swift
```diff
@@ -302,4 +304,48 @@ class TestNSArray : XCTestCase {
         let emptyArray = NSArray().sortedArrayWithOptions([]) { _,_ in .OrderedSame }
         XCTAssertTrue(emptyArray.isEmpty)
     }
+
+    func test_sortUsingFunction() {
+        let inputNumbers = [11, 120, 215, 11, 1, -22, 35, -89, 65]
+        let mutableInput = inputNumbers.bridge().mutableCopy() as! NSMutableArray
```
phausler: This currently fails due to a secondary failure of NSArray's mutable copy not returning the correct value on linux

phausler:
I pushed a fix for the mutableCopy case for linux; which is a work-around for the same underlying problem as SR-310

PR #166, opened 2015-12-20
Add override for `description` on NSNumber

rothomp3:
- Better match Darwin Foundation behavior


Foundation/NSNumber.swift
```diff
@@ -389,6 +389,13 @@ public class NSNumber : NSValue {
     override internal var _cfTypeID: CFTypeID {
         return CFNumberGetTypeID()
     }
+    
+    public override var description: String {
```
phausler: this should probably call descriptionWithLocale: nil and that should format the number according to it's "objcType"

PR #167, opened 2015-12-20
Sets the special flags in the CFType info on NSString to properly indicate its storage type.

brownleej:
Currently the CFInfo fields on NSString are being set to 0x80, 0x07, 0x00, 0x00. This means that the flag for __kCFContentsMask is being set to __kCFHasInlineContents, which is not an accurate description of how the data is laid out in an NSString that's created from a  Swift String. This in turn causes the result of `NSString.init(format:arguments:)` to be empty, since it's going through a CFString method that can't properly read the format string.

### DIFF
--- a/CoreFoundation/String.subproj/CFString.c
+++ b/CoreFoundation/String.subproj/CFString.c
@@ -6041,7 +6041,7 @@ static void __CFStringAppendFormatCore(CFMutableStringRef outputString, CFString
 
 
     formatLen = CFStringGetLength(formatString);
-    if (!CF_IS_OBJC(__kCFStringTypeID, formatString)) {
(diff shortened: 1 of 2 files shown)


phausler:
NSString should be falling into the case of CF_IS_SWIFT so it should dispatch out to the bridge methods to access the string. NSCFString is the one that should share layout to CF. the _storage parameter should NEVER be accessed by swift (and if it is it is a bug)

brownleej:
OK, I've pushed out a new commit that uses  CF_IS_SWIFT to send it down a path that calls the appropriate Swift functions for accessing the contents.

phausler:
that looks about right.

PR #168, opened 2015-12-20
Adds file and line context to more error functions.

brownleej:
This allows these error functions to report the failure as being at the call site, which will make it easier to debug errors caused by trying to call unimplemented methods.


phausler:
This seems reasonable as a behavioral differential; we are already divergent a bit from the darwin version on these but giving the line number seems quite reasonable to me to help aide identifying failures in CI etc.

PR #169, opened 2015-12-21
Improve ARMv7 support

hpux735:
This pull request addresses two issues.  The first is that Int is used in NSXMLNodeOptions, but it tries to "set the high 12 bits" to '1'.  The problem is that swift (correctly) identifies this as an overflow on a signed Int.  If it is necessary for this to be Int, I suppose that an overflow bitwise or could be used (does such an operator exist?).

The second bit is a hard-coding of the ABI on ARM.  This is a stopgap until it can either be provided by the swift/utils/build-script, or is correctly detected.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -177,6 +177,10 @@ struct _NSXMLParserBridge {
                            const unsigned char *SystemID);
 };
 
+struct _NSRunLoop {
+    _Nonnull CFTypeRef (*_Nonnull _new)(CFRunLoopRef rl);
+};
+
 struct _CFSwiftBridge {
     struct _NSObjectBridge NSObject;
     struct _NSArrayBridge NSArray;
@@ -188,6 +192,7 @@ struct _CFSwiftBridge {
     struct _NSStringBridge NSString;
     struct _NSMutableStringBridge NSMutableString;
     struct _NSXMLParserBridge NSXMLParser;
+    struct _NSRunLoop NSRunLoop;
 };
 
 __attribute__((__visibility__("hidden"))) extern struct _CFSwiftBridge __CFSwiftBridge;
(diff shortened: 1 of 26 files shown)


phausler:
unfortunately again this is an API change, the declared types for arm 32 bit and i386 are Int due to the conversion from NSUInteger to Int; the two ways to approach cross arch compatibility here are to either change these to be in the form
```swift
public let NSXMLNodePromoteSignificantWhitespace = 1 << 28
```
or 
```swift
#if arch(x86_64) // and other 64 bit arches go here
public var NSXMLNodePreserveCharacterReferences: Int { return 1 << 27 }
#else
public var NSXMLNodePreserveCharacterReferences: Int64 { return 1 << 27 }
#endif
```
I agree that the signed type is inappropriate for these values; but the swift importer changes all access of NSUInteger to Int so either a) we would need to change the declaration in NSXML* on darwin or b) make an exception for 32 bit targets on the open source version.

hpux735:
It seems like it’s not such a bad thing to define it to 64-bit on 32-bit platforms.  It’s a bit wasteful, but if on the off-chance that someone sends these values from one platform to another, at least they’ll match in size.

phausler:
perhaps it might be worthwhile to have a define for 32 bit platforms so that the #if statement can be simple (id guess this won't be the last place we are going to hit with this issue)

hpux735:
Yes, I suspect that would be useful.  For now, though, do you think a typealias would be appropriate here?  I would rather not have the entire file be copy & paste.  Something along the lines of:

```
#if arch(x86_64) 
typealias XMLNodeOptionRawType Int
#else
typealias XMLNodeOptionRawType Int64
#endif

public var NSXMLNodeOptionsNone: XMLNodeOptionRawType { return 0 }
```

The potential ugliness here (assuming that I understand it well enough that this would work) is that I suspect that alias would start showing up in autocompletes, leading to confusion.

phausler:
Type alias would be fine, however until we drop the NS prefix, perhaps this should be NSXMLNodeOptionRawType just for consistency sake when the mass change will be applied.

phausler:
looks good to me; merging in with one slight addendum to add arm64 checks too

hpux735:
Thanks so much!  Have a great break!

PR #170, opened 2015-12-21
[WIP] Initial implementation on NSAttributedString

kovpas:
Started with `NSAttributedString`. Question - I don't see Core Foundation counterpart. Does that mean that it's not needed? Or it should be implemented there first and then wrapped with `NSAttributedString`?

### DIFF
--- a/Foundation/NSAttributedString.swift
+++ b/Foundation/NSAttributedString.swift
@@ -7,6 +7,20 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+private struct _NSAttributedStringAttribute {
+    let range: Range<Int>
+    let name: String
+    let value: NSObject
+}
+
+extension _NSAttributedStringAttribute : Equatable {
+}
+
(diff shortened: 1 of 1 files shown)


phausler:
As of the current release we do not yet have an open source version of CFAttributedString. If it were present then I would claim that it should probably funnel to that.

Foundation/NSAttributedString.swift
```diff
@@ -23,28 +37,77 @@ public class NSAttributedString : NSObject, NSCopying, NSMutableCopying, NSSecur
     }
     
     public func copyWithZone(zone: NSZone) -> AnyObject {
-        NSUnimplemented()
+        return NSAttributedString(attributedString: self)
```
phausler: the copy of an immutable object is just self; no need to construct a new object

kovpas: Indeed, thanks.

kovpas:
Don't think it's smart to do what's already done, so it would make sense to decline this PR and wait for `CFAttributedString` release then.

PR #171, opened 2015-12-21
Initial implementation of the NSNotificationQueue

sgl0v:
Initial implementation of the NSNotificationQueue. Supports NSPostingStyle.PostNow style only.

### DIFF
--- a/Foundation/NSNotificationQueue.swift
+++ b/Foundation/NSNotificationQueue.swift
@@ -1,3 +1,11 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2015 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
 
 public enum NSPostingStyle : UInt {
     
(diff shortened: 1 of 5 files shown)


Foundation/NSNotificationQueue.swift
```diff
@@ -15,26 +23,128 @@ public struct NSNotificationCoalescing : OptionSetType {
...
+    private typealias NotificationQueueList = Array<NSNotificationQueue>
+    private typealias NSNotificationListEntry = (NSNotification, [String])
+    private typealias NSNotificationList = [NSNotificationListEntry]
+
+    private static let _notificationQueueThreadKey = "NSNotificationQueueThreadKey"
+    private static let _notificationListThreadKey = "NSNotificationListThreadKey"
+    private let notificationCenter: NSNotificationCenter
+    private var asapList = NSNotificationList()
+    private var idleList = NSNotificationList()
+    private var notificationQueueList: NotificationQueueList {
```
phausler: This seems to cause a compiler crash
```
1.	While emitting SIL for getter for notificationQueueList at Foundation/NSNotificationQueue.swift:37:9
```

Foundation/NSNotificationQueue.swift
```diff
@@ -15,26 +23,128 @@ public struct NSNotificationCoalescing : OptionSetType {
...
+        get {
+            let currentThread = NSThread.currentThread()
```
phausler: you could probably use a TLS specific storage via this internal class `NSThreadSpecific< NotificationQueueList >()`
see `NSThread.currentThread()` for an example of how to use that.

sgl0v:
Now `NSNotificationQueue` uses `NSThreadSpecific` object to store thread-related data instead of `NSThread`'s `threadDictionary` property. I changed the `NSThread` in order to set multiple thread-specific objects as well.

Foundation/NSNotificationQueue.swift
```diff
@@ -15,26 +23,116 @@ public struct NSNotificationCoalescing : OptionSetType {
...
+
+    // The NSNotificationQueue instance is associated with current thread.
+    // The _notificationQueueList represents a list of notification queues related to the current thread.
+    private static var _notificationQueueList = NSThreadSpecific<NSArray>()
+    internal class var notificationQueueList: NotificationQueueList {
+        get {
+            let array = _notificationQueueList.get({
+                return NSArray()
+            })
+            return array.allObjects as! NotificationQueueList
```
phausler: This really cannot work since NSArray must store NSObjects and NSNotificationListEntry is a tuple of a NSNotification and a swift array struct. The as! also causes a compiler crash on linux

phausler: err disregard, I misread the type aliasing; it still crashes the compiler on linux

This will do roughly the same thing except it won't crash the compiler.
```swift
internal static var notificationQueueList: NotificationQueueList {
        get {
            let array = _notificationQueueList.get() {
                return NSArray()
            }
	    return array._swiftObject.map { $0 as! NSNotificationQueue }
        }
        set {
            _notificationQueueList.set(newValue._nsObject)
        }
    }
```

Foundation/NSNotificationQueue.swift
```diff
@@ -15,26 +23,116 @@ public struct NSNotificationCoalescing : OptionSetType {
...
+
+        self.asapList = self.asapList.filter(predicate)
+        self.idleList = self.idleList.filter(predicate)
     }
 
-}
+    // MARK: Private
+
+    private class func registerQueue(notificationQueue: NSNotificationQueue) {
+        var notificationQueueList = self.notificationQueueList
```
phausler: if this was a NSMutableArray it would encapsulate the hoop jumping of fetching then mutating then re-assigning and the NSMutableArray would act as a box to the array container.

TestFoundation/TestNSNotificationQueue.swift
```diff
@@ -0,0 +1,151 @@
...
+        XCTAssertNil(notificationQueue)
+    }
+
+    // MARK: Private
+
+    private func executeInBackgroundThread(operation: () -> ()) {
+        var isFinished = false
+        let bgThread = NSThread() {
+            operation()
+            NSThread.exit()
```
phausler: direct calls to exit probably are not ideal here; the thread will exit on it's own accord when the block is finished.

sgl0v:
Fixed the compiler crash & replaced NSArray with NSMutableArray as proposed above. Removed the direct `NSThread.exit()` call as well.

Foundation/NSNotificationQueue.swift
```diff
@@ -15,26 +23,105 @@ public struct NSNotificationCoalescing : OptionSetType {
...
+            self.dequeueNotificationsMatching(notification, coalesceMask: coalesceMask)
+        }
+
+        switch postingStyle {
+        case .PostNow:
+            let currentMode = NSRunLoop.currentRunLoop().currentMode
+            if currentMode == nil || runloopModes.contains(currentMode!) {
+                self.notificationCenter.postNotification(notification)
+            }
+        case .PostASAP:
```
phausler: these two options could utilize a CFRunLoopObserver:
ASAP is basically kCFRunLoopBeforeTimers | kCFRunLoopExit
and idle would be kCFRunLoopBeforeWaiting

sgl0v: Thanks for the hint.

I implemented the idle posting style already via `CFRunLoopObserver` & `kCFRunLoopBeforeWaiting`. Will work on asap posting style then and update this PR or submit a new one.

phausler: if you want to do another that is fine, I was going to merge this since it looks like a great start and passes building/tests on linux

PR #172, opened 2015-12-25
Drop deprecated `++` from TestNSIndexSet

tomj:


phausler:
looks like this does not alter the behavior

PR #175, opened 2015-12-27
Fixes NSString.stringByDeletingLastPathComponent.

HeMet:
It deletes final path separator now. If the receiver represents the root path it is returned unaltered.

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -237,14 +237,28 @@ public extension NSString {
         if fixedSelf == "/" {
             return fixedSelf
         }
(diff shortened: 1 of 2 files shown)


phausler:
tests look good, I wonder if this was related to the failure that was detected in CI for the completion method.

HeMet:
@phausler I'm not sure. I checked tests again. Test that fails gets content of tmp directory by call to contentsOfDirectoryAtURL without any enumeration options specified. On the other hand, completion method uses enumerator with .SkipsSubdirectoryDescendants option specified. That might be the issue, but I could not reproduce it myself.

PR #177, opened 2015-12-28
NSNotificationCenter observers synchronization

ochococo:


Foundation/NSNotification.swift
```diff
@@ -105,20 +140,17 @@ public class NSNotificationCenter : NSObject {
...
-        }
-    }
 
+        _observersLock.synchronized({
+            self._observers = _observers.filterOutObserver(observer, name: name, object: object)
+        })
+    }
     
-    public func addObserverForName(name: String?, object obj: AnyObject?, queue: NSOperationQueue?, usingBlock block: (NSNotification) -> Void) -> NSObjectProtocol {
+    public func addObserverForName(name: String?, object obj: AnyObject?, queue: NSOperationQueue?, usingBlock block: (NSNotification) -> ()) -> NSObjectProtocol {
```
phausler: why the change to ()? Void seems more readable/appropriate to me

ochococo: You're right. It's more readable now. Fixed ✅

PR #180, opened 2015-12-28
Defined isEqual and hash value getters for various collection classes

sgl0v:
Defined isEqual and hash value getters for `NSArray`, `NSDictionary` and `NSSet`


phausler:
As a follow up it might be interesting to have unit tests to verify mutable versions of isEqual etc

PR #181, opened 2015-12-29
Speed up JSON Parsing

argon:
I ran 2 benchmarks against the current JSON parser implementation on `master`, comparing it to Darwin Foundation for [large inputs](https://github.com/zemirco/sf-city-lots-json/blob/master/citylots.json) and found it was 5-10x slower. (72s vs 8s for a single decode). This new code reduces this to a factor for 2-4x (17s)

While JSON is a text-based serialization, all control characters are ASCII characters. The data can be most efficiently parsed as a stream of bytes instead of the initial overhead of converting the bytes to a String.

The byte -> string conversion in the current code only accounts for 3s of time but parsing numbers is incredibly expensive (`Double.init?(_ text: String)` converts the string back to bytes to use [`strtod()`](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/strtod.3.html#//apple_ref/doc/man/3/strtod)). This technique uses `strtod` directly on the byte array. There is a penalty for non-UTF-8 encoded data which is inline with the Darwin implementation

> The most efficient encoding to use for parsing is UTF-8, so if you have a choice in encoding the data passed to this method, use UTF-8.

Another modification has been made to remove the intermediate `parser` values as the heap allocations for each intermediate parser was adding significant overhead.

This also includes a commit to implement `.AllowFragments`.

### DIFF
--- a/Docs/Status.md
+++ b/Docs/Status.md
@@ -48,7 +48,7 @@ As Foundation is a work in progress, not all methods and functionality are prese
 
     * `NSPropertyList` is mostly implemented.
     * `NSCoder` and `NSKeyedArchiver` are mostly not yet implemented (a few funnel methods are implemented).
-    * `NSJSONSerialization` is not yet implemented.
+    * `NSJSONSerialization` is partly implemented (serialization is not yet implemented)
 
 
 * **XML**: A group of classes for parsing and representing XML documents and elements.
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -251,12 +251,12 @@ private struct JSONDeserializer {
     }
     
     struct StructureScalar {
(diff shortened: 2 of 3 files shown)


phausler:
from a cursory glance this looks like a reasonable change; builds pass on ubuntu.

Do you happen to have the performance benchmarks handy? that might be something interesting to check in so that we could potentially build up a performance test suite.

Additionally what other hotspots are there in Foundation that you found by this?

argon:
The performance "benchmark" wasn't in any way automated unfortunately. I simply created a `main.swift` to read the JSON file from disk then decode it, then I profiled it using instruments.

The main hotspots in _this_ code now relate to

 - Array and Dictionary size increases, leading to a large number of copies
 - Bridging from NSString -> String in `parseString`

In general String parsing for the sf-city-lots-json bears a huge time cost ~(5000ms vs 250ms) when compared with Darwin Foundation.

PR #182, opened 2015-12-29
Updates to NSHTTPCookie

dbrajkovic:
Updates to NSHTTPCookie

## Changes in this pull request

* Fixed setting version from NSHTTPCookieVersion, was NSHTTPCookieSecure
* Fixed string format when setting expires date
* Fixed creating expiresDate from timeInterval. Docs state must be integer.
* Fixed setting discard to match documentation
* Replaced internal struct with private properties, serve the same purpose
* Created internal properties dictionary that is returned by properties get. Closer to Cocoa compatibility
* Added "Created" key and value to properties dictionary for Cocoa compatibility.

## Notes
The documentation of NSHTTPCookie and the Cocoa implementation of NSHTTPCookie differ. Not sure if it's a bug or if it's proper, yet undocumented behavior.  The question is, do we want to be compatible with Cocoa or more closely follow documentation.

### Examples
* Certain values that are string representations of numbers, such as ```NSHTTPCookieVersion``` are documented as accepting strings when using ```initWithProperties```. The Cocoa implementation actually accepts NSNumbers as well
* Along with the above example, the ```properties``` dictionary returns numbers instead of strings. The documentation states that the properties dictionary can be used to created a new cookie with the same values. This contradicts the documentation which states that all the numeric values of the properties dictionary are supposed to be strings.
* Created value in properties dictionary is undocumented.
* Cocoa implementation does not exclude any properties based on the cookie's version.
* Initializing with property values that should be invalid according to documentation returns a cookie, For example, comment and commentURL values should only be used with cookie version 1. Yet, you can initialize a cookie that is version 0 and with values in the comment and/or commentURL keys.

### DIFF
--- a/Foundation/NSHTTPCookie.swift
+++ b/Foundation/NSHTTPCookie.swift
@@ -95,23 +95,20 @@ public let NSHTTPCookiePort: String = "Port"
     attributes of a cookie.
 */
 public class NSHTTPCookie : NSObject {
-    private struct Cookie {
-        let comment: String?
-        let commentURL: NSURL?
-        let domain: String
-        let expiresDate: NSDate?
-        let HTTPOnly: Bool
-        let secure: Bool
-        let sessionOnly: Bool
-        let name: String
-        let path: String
-        let portList: [NSNumber]?
-        let properties: [String: Any]?
-        let value: String
-        let version: Int
-    }
-    private let cookieRepresentation: Cookie
-
+    let _comment: String?
+    let _commentURL: NSURL?
+    let _domain: String
+    let _expiresDate: NSDate?
+    let _HTTPOnly: Bool
+    let _secure: Bool
+    let _sessionOnly: Bool
+    let _name: String
+    let _path: String
+    let _portList: [NSNumber]?
+    let _value: String
+    let _version: Int
+    var _properties: [String : Any]
+    
     /*!
         @method initWithProperties:
         @abstract Initialize a NSHTTPCookie object with a dictionary of
@@ -245,7 +242,7 @@ public class NSHTTPCookie : NSObject {
         else {
             return nil
         }
-
+        
         let canonicalDomain: String
         if let domain = properties[NSHTTPCookieDomain] as? String {
             canonicalDomain = domain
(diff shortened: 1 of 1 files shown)


phausler:
builds cleanly on ubuntu, since this is not a component that is directly under my responsibility I am going to defer to @salgernon for commentary
If this one is preferred the one thing the other commit has is unit tests; which it might be worthwhile in that case to take both for the tests section.

dbrajkovic:
Agreed. Unit tests are lacking. I'll try to add them to this pull request while waiting for commentary from Steve. Then you all can pick one or the other.

I do feel that this PR can use more refactoring.  But I was torn on the best way to handle. It depends on whether we want to copy Cocoa's functionality. In the Cocoa Implementation, the properties property returns a dictionary with only the keys that were passed in the dictionary at initialization (plus Created). The values are either the original values or modified values if necessary (e.g. both Max-Age and Expires-Date are passed). Since all the other values can be derived from this dictionary, is it necessary to store those values internally, be it in variables or a struct, or just derive them when needed.

salgernon:
I'm looking at this in adjunct with pull request #124.  In general, the Cocoa approach here would be to be lenient in what we accept, and so the implementation would trump the documentation.  (There are *many* issues with the documentation in the NSURL loader headers.)  

We should add a property for HTTPOnly; it is lacking from Cocoa's properties as well and can't be inferred.

dbrajkovic:
@salgernon Thanks. Then I will update the code to feature match Cocoa's current implementation.

seabaylea:
Hi @dbrajkovic . Are you still working on this? One of the issues fixed by your PR has been raised by my colleague via [SR-948](https://bugs.swift.org/browse/SR-948).

If you're unlikely to get to this in the near future, would it be ok for me to fix (just) the issue that we currently are hitting?

Foundation/NSHTTPCookie.swift
```diff
@@ -95,23 +95,20 @@ public let NSHTTPCookiePort: String = "Port"
...
+    let _expiresDate: NSDate?
+    let _HTTPOnly: Bool
+    let _secure: Bool
+    let _sessionOnly: Bool
+    let _name: String
+    let _path: String
+    let _portList: [NSNumber]?
+    let _value: String
+    let _version: Int
+    var _properties: [String : Any]
```
jessesquires: hi @dbrajkovic ! :smile: 

>Replaced internal struct with private properties, serve the same purpose.

Aren't these `internal` now, not `private` ?

danieleggert:
Note: I've started working on `NSURLSession` — will try to integrate with this at some point.

PR #184, opened 2015-12-29
Implementation of NSKeyedArchiver/NSKeyedUnarchiver

lhoward:
Implementation of NSKeyedArchiver/NSKeyedUnarchiver.


phausler:
SR-404 is a pretty big deal... if we cannot use `as? NSCoding` that is kind-of a deal-breaker.
Can we bypass this with unsafeBitCast or some other way to fool the compiler into generating the correct code?

phausler:
per the unit tests I think the best way to go about this is to do at least two patterns - encode some archives from the Darwin version and then decode with this as well as a round trip.

CoreFoundation/Base.subproj/CFRuntime.c
```diff
@@ -1817,6 +1818,28 @@ void * objc_retainAutoreleasedReturnValue(void *obj) {
...
+} NominalTypeDescriptor;
+
+extern const NominalTypeDescriptor *_ZNK5swift8Metadata24getNominalTypeDescriptorEv(const void *);
+
+_Nullable CFStringRef _CFCopyNominalTypeNameForClass(_Nonnull CFTypeRef aClass) {
+    const NominalTypeDescriptor *ntd;
+
+    CFAssert1(aClass != NULL, __kCFLogAssertion, "%s(): invalid class", __PRETTY_FUNCTION__);
+
+    ntd = _ZNK5swift8Metadata24getNominalTypeDescriptorEv(aClass);
```
phausler: we will need a sign-off to support this type of access from the runtime team

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -291,6 +294,18 @@ extern void _cf_uuid_unparse(const _cf_uuid_t uu, _cf_uuid_string_t out);
 extern void _cf_uuid_unparse_lower(const _cf_uuid_t uu, _cf_uuid_string_t out);
 extern void _cf_uuid_unparse_upper(const _cf_uuid_t uu, _cf_uuid_string_t out);
 
+#if !__COREFOUNDATION_FORFOUNDATIONONLY__
```
phausler: this guard should not be needed

CoreFoundation/Collections.subproj/CFDictionary.c
```diff
@@ -500,7 +500,7 @@ void CFDictionaryGetValues(CFHashRef hc, const_any_pointer_t *keybuf) {
     const_any_pointer_t *valuebuf = 0;
 #endif
 #if CFDictionary
-    if (CFDictionary) CF_SWIFT_FUNCDISPATCHV(CFDictionaryGetTypeID(), void, (CFSwiftRef)hc, NSDictionary.getObjects, valuebuf, keybuf);
+    if (CFDictionary) CF_SWIFT_FUNCDISPATCHV(CFDictionaryGetTypeID(), void, (CFSwiftRef)hc, NSDictionary.getObjects, keybuf, valuebuf);
```
phausler: now that i look at it, perhaps it would be better to flip the other side

CoreFoundation/Parsing.subproj/CFPropertyList.h
```diff
@@ -87,7 +85,7 @@ typedef CF_ENUM(CFIndex, CFPropertyListFormat) {
 CF_EXPORT
 Boolean CFPropertyListIsValid(CFPropertyListRef plist, CFPropertyListFormat format);
 
-#if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_EMBEDDED
+#if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_EMBEDDED || DEPLOYMENT_RUNTIME_SWIFT
```
phausler: probably we can just remove this guard

CoreFoundation/Parsing.subproj/CFPropertyList.h
```diff
@@ -133,7 +131,7 @@ CF_ENUM(CFIndex) {
 CF_EXPORT
 CFPropertyListRef CFPropertyListCreateWithData(CFAllocatorRef allocator, CFDataRef data, CFOptionFlags options, CFPropertyListFormat *format, CFErrorRef *error) CF_AVAILABLE(10_6, 4_0);
 
-#if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_EMBEDDED
+#if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_EMBEDDED || DEPLOYMENT_RUNTIME_SWIFT
```
phausler: here too

Docs/Archiving.md
```diff
@@ -0,0 +1,68 @@
+# Archiving Notes
+
+There is a preliminary implementation of NSKeyedArchiver and NSKeyedUnarchiver which should be compatible with the OS X version.
+
+* The implementation of NSStringFromClass() and NSClassFromString() currently uses private Swift API and does not support encoding classes that are not exported (i.e. cannot be looked up with dlsym()
+
+* NSKeyedUnarchiver reads the entire plist into memory before constructing the object graph, it should construct it incrementally as does Foundation on OS X
+
+* NSConcreteValue is only partially implemented, it does not have encoding complex types such as NSRange, NSPoint, NSRect, etc
```
phausler: darwin special cases these types (primarily for speed) which I bet we can do the same easily

Foundation/NSCFDictionary.swift
```diff
@@ -139,8 +143,12 @@ internal func _CFSwiftDictionaryContainsValue(dictionary: AnyObject, value: AnyO
 internal func _CFSwiftDictionaryGetKeysAndValues(dictionary: AnyObject, keybuf: UnsafeMutablePointer<Unmanaged<AnyObject>?>, valuebuf: UnsafeMutablePointer<Unmanaged<AnyObject>?>) {
     var idx = 0
```
phausler: not that it is likely but if both the keybuf and valuebuf are nil you could early return before iterating

Foundation/NSCalendar.swift
```diff
@@ -106,12 +106,52 @@ public class NSCalendar : NSObject, NSCopying, NSSecureCoding {
         return unsafeBitCast(self, CFCalendarRef.self)
     }
     
-    public required init?(coder aDecoder: NSCoder) {
-        NSUnimplemented()
+    public convenience required init?(coder aDecoder: NSCoder) {
+        if aDecoder.allowsKeyedCoding {
```
phausler: I will follow up on this one to compare to the darwin version later

Foundation/NSCalendar.swift
```diff
@@ -1194,12 +1239,54 @@ public class NSDateComponents : NSObject, NSCopying, NSSecureCoding {
         return false
     }
     
-    public required init?(coder aDecoder: NSCoder) {
-        NSUnimplemented()
+    public convenience required init?(coder aDecoder: NSCoder) {
+        if aDecoder.allowsKeyedCoding {
```
phausler: i will follow up with this one as well to compare to darwin later

Foundation/NSCoder.swift
```diff
@@ -52,7 +52,7 @@ public class NSCoder : NSObject {
     }
     
     @warn_unused_result
-    public func decodeObjectOfClasses(classes: NSSet?, forKey key: String) -> AnyObject? {
+    public func decodeObjectOfClasses(classes: [AnyClass], forKey key: String) -> AnyObject? {
```
phausler: it would be good to annotate with headerdoc's the deviation from darwin here

Foundation/NSCoder.swift
```diff
@@ -72,7 +72,7 @@ public class NSCoder : NSObject {
     }
     
     @warn_unused_result
-    public func decodeTopLevelObjectOfClasses(classes: NSSet?, forKey key: String) throws -> AnyObject? {
+    public func decodeTopLevelObjectOfClasses(classes: [AnyClass], forKey key: String) throws -> AnyObject? {
```
phausler: this needs a comment as well

Foundation/NSCoder.swift
```diff
@@ -253,13 +253,34 @@ public class NSCoder : NSObject {
         NSUnimplemented()
     }
     
-    public var allowedClasses: Set<NSObject>? {
+    public var allowedClasses: [AnyClass]? {
```
phausler: and ditto for the comment

Foundation/NSCoder.swift
```diff
@@ -253,13 +253,34 @@ public class NSCoder : NSObject {
...
         get {
             NSUnimplemented()
         }
     }
     
     public func failWithError(error: NSError) {
-        NSUnimplemented()
+        if let debugDescription = error.userInfo["NSDebugDescription"] {
```
phausler: this is incorrect; failWithError is a method to halt the decoding of the archive such that the top level object will then return nil from the decoding; failure is not fatal on darwin it just is an error

Foundation/NSConcreteValue.swift
```diff
@@ -0,0 +1,199 @@
...
+            self.name = name
+        }
+    }
+
+    typealias TypeInfo = ManagedBuffer<TypeInfoHeader, Int8>
+    
+    static var _cachedTypeInfo = Dictionary<String, TypeInfo>()
+    
+    var _specialFlags : SpecialFlags = .NotSpecialType
+    var _typeInfo : TypeInfo
```
phausler: typeInfo probably could just be a String for storage

Foundation/NSConcreteValue.swift
```diff
@@ -0,0 +1,199 @@
...
+    override var classForKeyedArchiver: AnyClass? {
+        return NSValue.self
+    }
+    
+    override var description : String {
+        return NSData.init(bytes: self.value, length: self._size).description
+    }
+    
+    convenience required init?(coder aDecoder: NSCoder) {
+        if !aDecoder.allowsKeyedCoding {
```
phausler: I will take a look at this one later to compare to Darwin's implementation

Foundation/NSDecimalNumber.swift
```diff
@@ -51,6 +51,10 @@ public class NSDecimalNumber : NSNumber {
         NSUnimplemented()
     }
     
+    public required convenience init(bytes buffer: UnsafePointer<Void>, objCType: UnsafePointer<Int8>) {
+        NSUnimplemented()
```
phausler: to be honest I am not certain if this is even valid in objc for NSDecimalNumber

Foundation/NSError.swift
```diff
@@ -64,7 +91,16 @@ public class NSError : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
     
     public func encodeWithCoder(aCoder: NSCoder) {
-        
+        if aCoder.allowsKeyedCoding {
+            aCoder.encodeObject(_domain.bridge(), forKey: "NSDomain")
```
phausler: I know this would be an API change but I wonder if we could make an overload here to avoid needing to remember the .bridge(); not fully convinced that is a good idea but something to think about.

Foundation/NSKeyedArchiver.swift
```diff
@@ -7,183 +7,810 @@
...
+        keyedArchiver.encodeObject(rootObject, forKey: NSKeyedArchiveRootObjectKey)
+        keyedArchiver.finishEncoding()
+        
+        return data
     }
     
     public class func archiveRootObject(rootObject: AnyObject, toFile path: String) -> Bool {
-        NSUnimplemented()
+        // FIXME write to temp file first
+        // FIXME CFWriteStreamCreateWithFile() seems to be broken
```
phausler: what is broken about it? bug number?

Foundation/NSKeyedUnarchiver.swift
```diff
@@ -0,0 +1,970 @@
...
+                                                allowedClasses: [AnyClass]?,
+                                                inout classToConstruct: AnyClass?) -> Bool {
+        classToConstruct = nil
+        
+        func _classForClassName(codedName: String) -> AnyClass? {
+            var aClass : AnyClass?
+            
+            aClass = classForClassName(codedName)
+            if aClass == nil {
+                aClass = NSKeyedUnarchiver.classForClassName(codedName)
```
phausler: I think we should probably have the base Foundation types stuffed in the mappings backing this so that we can avoid the call to NSClassFromString as long as possible since that will require sign-off from the runtime team for us accessing their internal functions

Foundation/NSNumber.swift
```diff
@@ -39,6 +39,44 @@ extension Int : _ObjectTypeBridgeable {
...
+        result = x.intValue
+    }
+    
+    public static func _conditionallyBridgeFromObject(x: NSNumber, inout result: Int32?) -> Bool {
+        self._forceBridgeFromObject(x, result: &result)
+        return true
+    }
+}
+
+extension Int64 : _ObjectTypeBridgeable {
```
phausler: Are these types bridged in the current stdlib overlay for Foundation?

Foundation/NSObjCRuntime.swift
```diff
@@ -11,11 +11,139 @@
...
+import Glibc
+#endif
+
+#if os(OSX) || os(iOS)
 internal let kCFCompareLessThan = CFComparisonResult.CompareLessThan
 internal let kCFCompareEqualTo = CFComparisonResult.CompareEqualTo
 internal let kCFCompareGreaterThan = CFComparisonResult.CompareGreaterThan
 #endif
 
+internal enum NSObjCType : UnicodeScalar {
```
phausler: These are simple objc types because you can have for example `"^@"` or `"{dddd}"` etc, and probably we should prefix this with an underscore to drive home the point that this is internal

Foundation/NSObjCRuntime.swift
```diff
@@ -114,3 +242,106 @@ internal protocol _NSBridgable {
...
+        } else {
+            mangledName = mangledTypeNameNameForClass([_SwiftFoundationModuleName] + components)
+        }
+    }
+    
+    if !mangledName.hasPrefix("_Tt") {
+        fatalError("NSClassFromString() invalid mangled name \(mangledName)")
+    }
+
+    let accessorName = "_TMa" + mangledName.bridge().substringFromIndex(3)
```
phausler: We really need an official method to do this from the runtime. Lets create a few feature request bugs to track that and create a few bugs to improve this to be less fragile and block them on that. Since the start of this project we have already seen the mangler change once so it is reasonable to think that the mangled names may change again.

Foundation/NSObject.swift
```diff
@@ -118,8 +118,44 @@ public class NSObject : NSObjectProtocol {
     internal var _cfTypeID: CFTypeID {
         return 0
     }
-}
+    
+    // TODO move these back into extensions once extension methods can be overriden
+    
+    // TODO: Could perhaps be an extension of NSCoding instead. The reason it is an extension of NSObject is the lack of default implementations on protocols in Objective-C.
```
phausler: we provide default implementations on protocols for delegates already so why not these?

Foundation/NSOrderedSet.swift
```diff
@@ -32,10 +32,28 @@ public class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
...
+                aCoder.encodeObject(self.objectAtIndex(idx), forKey:"NS.object.\(idx)")
+            }
+        } else {
+            NSUnimplemented()
+        }
     }
     
-    public required init?(coder aDecoder: NSCoder) { NSUnimplemented() }
+    public required convenience init?(coder aDecoder: NSCoder) {
+        if aDecoder.allowsKeyedCoding {
```
phausler: I will take a look at this one later to compare to Darwin's version

Foundation/NSPropertyList.swift
```diff
@@ -117,6 +138,8 @@ internal func _expensivePropertyListConversion(input : AnyObject) -> Any {
         return true
     } else if input === kCFBooleanFalse {
         return false
+    } else if CFGetTypeID(input) == _CFKeyedArchiverUIDGetTypeID() {
```
phausler: @parkera do we want to extract this out? or will it be ok leaving it here since it is an internal class type?

phausler: CFGetTypeID assumes that the input is a CFTypeRef; non CF backed objects will potentially read past the allocations so we should use the property `_cfTypeID` instead

Foundation/NSSet.swift
```diff
@@ -117,11 +117,37 @@ public class NSSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
+                objects.append(object)
+                count += 1
+            }
+            self.init(array: objects)
+        }
     }
     
     public func encodeWithCoder(aCoder: NSCoder) {
-        NSUnimplemented()
+        self.allObjects._nsObject.encodeWithCoder(aCoder)
```
phausler: This probably deserves a comment along the lines of "NSSet shares the same data layout encoding as NSArray but the classes will be encoded differently" or something like that

Foundation/NSString.swift
```diff
@@ -331,7 +354,8 @@ public class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, N
     }
     
     public override var hash: Int {
-        return Int(CFStringHashNSString(self._cfObject))
+        return unsafeBitCast(CFStringHashNSString(self._cfObject), Int.self)
```
phausler: I think the underlying problem here is that hash is Int and CF/Foundation's hash ought to be UInt; you probably want Int(bitPattern:CFStringHashNSString(self._cfObject))

Foundation/NSConcreteValue.swift
```diff
@@ -0,0 +1,199 @@
...
+    
+    static func createTypeInfoNonSpecial(spec: String, objCType typep: UnsafePointer<Int8>) -> TypeInfo? {
+        var size: Int = 0
+        var align: Int = 0
+        var count : Int = 0
+        let extraBytes : Int = spec.length + 1 // extra bytes after TypeInfoHeader
+        
+        var type = NSObjCType(spec)
+        guard type != nil else {
+            NSLog("NSConcreteValue.createTypeInfoNonSpecial: unsupported type encoding spec '\(spec)'")
```
phausler: probably we can just use print instead; and is there a reason why this isn't fatal?

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
lhoward: Removing the guard results in the following warning:

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h:298:45: Redefinition of typedef 'CFKeyedArchiverUIDRef' is a C11 feature

as ForFoundationOnly.h is included when building CoreFoundation.

CoreFoundation/Collections.subproj/CFDictionary.c
lhoward: _CFSwiftDictionaryGetKeysAndValues() then becomes a very confusing prototype because the parameter order does not follow the function name. Are you proposing renaming _CFSwiftDictionaryGetKeysAndValues()?

Foundation/NSCFDictionary.swift
lhoward: right, I thought this was sufficiently unlikely as to not bother with the special case. but I will fix it.

Foundation/NSConcreteValue.swift
lhoward: was harmonising with Darwin but possibly too closely?

Foundation/NSDecimalNumber.swift
lhoward: prefer to leave this as NSRequireConcreteImplementation() in superclass?

Foundation/NSError.swift
lhoward: Docs/Issues.md says that bridge() is temporary so changing API permanently does not seem the right idea

Foundation/NSKeyedArchiver.swift
lhoward: comment is out of date, it works

Foundation/NSNumber.swift
lhoward: no, removing

Foundation/NSObjCRuntime.swift
lhoward: renaming to _NSSimpleObjCType

Foundation/NSObjCRuntime.swift
lhoward: https://bugs.swift.org/browse/SR-381

Foundation/NSObject.swift
lhoward: can you explain this in more detail, maybe it's just early but I can't parse this just yet

CoreFoundation/Collections.subproj/CFDictionary.c
phausler: yea; I think renaming it is totally reasonable here because the signature should resemble the CF_OBJC_FUNCDISPATCHV call imho

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
phausler: perhaps it is then time I started merging the two headers and leaving this as for the swift side only

Foundation/NSDecimalNumber.swift
phausler: yea; the objc version does this
`Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** initialization method -initWithBytes:objCType: cannot be sent to an abstract object of class NSDecimalNumberPlaceholder: Create a concrete instance!'`
when calling it via `NSDecimalNumber *dcn = [[NSDecimalNumber alloc] initWithBytes:&dcv objCType:@encode(NSDecimal)];`

Foundation/NSError.swift
phausler: yea; agreed.

Foundation/NSPropertyList.swift
lhoward: __NSCFType is lacking _cfTypeID, is this a bug? I have fixed for now just to check "input is __NSCFType" above. If you would prefer me to add _cfTypeID to __NSCFType in NSSwiftRuntime.swift let me know.

Foundation/NSString.swift
lhoward: fixed

Foundation/NSPropertyList.swift
phausler: Hmm that seems to be a mistake if it does not call CFGetTypeID on itself for NSCFType; I need to verify that bridging works correctly with that change - I think that can probably be improved outside of this pull request (and a follow up commit addresses this issue)

Foundation/NSConcreteValue.swift
lhoward: can't be fatal otherwise you may have a denial-of-service when decoding (unlike similar methods where type is caller supplied, here type is data supplied). will change to print.

lhoward:
> On 30 Dec 2015, at 5:25 AM, Philippe Hausler <notifications@github.com> wrote:
> 
> SR-404 is a pretty big deal... if we cannot use as? NSCoding that is kind-of a deal-breaker.
> Can we bypass this with unsafeBitCast or some other way to fool the compiler into generating the correct code
> 
I think it will be quicker/better to fix the underlying bug as my gut says hacking around it will be too intrusive.

Foundation/NSObject.swift
phausler: Since we don't have optional implementations of delegate methods we made all of the delegation methods have default implementations; so extending NSCoding with a default implementation is not unreasonable and really wouldn't incur a full API change (provided it really does make sense in the implementation)

CoreFoundation/Collections.subproj/CFDictionary.c
lhoward: fixed

Foundation/NSCoder.swift
lhoward: fixed

Foundation/NSDecimalNumber.swift
lhoward: is calling NSRequiresConcreteImplementation() from here OK? we can't call super.init() because it's a convenience initialiser and also NSNumber is the super class (which does have this initialiser)

phausler:
> On Dec 29, 2015, at 2:11 PM, Luke Howard <notifications@github.com> wrote:
> 
> 
> > On 30 Dec 2015, at 5:25 AM, Philippe Hausler <notifications@github.com> wrote:
> > 
> > SR-404 is a pretty big deal... if we cannot use as? NSCoding that is kind-of a deal-breaker.
> > Can we bypass this with unsafeBitCast or some other way to fool the compiler into generating the correct code
> > 
> I think it will be quicker/better to fix the underlying bug as my gut says hacking around it will be too intrusive.
> —
> Reply to this email directly or view it on GitHub <https://github.com/apple/swift-corelibs-foundation/pull/184#issuecomment-167888478>.
> 
Then we MUST get that bug fixed before proceeding any further. Because this will basically hit on any archival no? Which if that is the case we cannot have unit tests failing like that on linux builds; which we should probably get a set of tests up and rolling as a next order of business for this.

Foundation/NSObject.swift
lhoward: > On 30 Dec 2015, at 9:12 AM, Philippe Hausler <notifications@github.com> wrote:
> 
> Since we don't have optional implementations of delegate methods we made all of the delegation methods have default implementations; so extending NSCoding with a default implementation is not unreasonable and really wouldn't incur a full API change (provided it really does make sense in the implementation)
> 

Don’t we already have this already in NSKeyedArchiver.swift/NSKeyedUnarchiver.swift? This was actually from the origin code in master.

lhoward:
> On 30 Dec 2015, at 9:22 AM, Philippe Hausler <notifications@github.com> wrote:
> 
> Then we MUST get that bug fixed before proceeding any further. Because this will basically hit on any archival no? Which if that is the case we cannot have unit tests failing like that on linux builds; which we should probably get a set of tests up and rolling as a next order of business for this.

Right. Well, I can look into this bug next (although I have to get onto some other projects this week).

Foundation/NSKeyedUnarchiver.swift
lhoward: NSStringFromClass()/NSClassFromString() will work for Foundation classes without accessing internal API, so a less verbose approach would be to disable the call to _CFCopyNominalTypeNameForClass().

However I would prefer to get approval first – how does we move this forward?

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
lhoward: OK, will leave this for you

Docs/Archiving.md
lhoward: right, the NSConcreteValue implementation is structured to make this easy, I just figured it's better to start integrating before this branch diverges and then add this later

Foundation/NSKeyedUnarchiver.swift
phausler: I have reached out to the runtime team to see who we need to talk with to get this done; I think for now just working on Foundation classes would be a great step forward from the NSUnimplemented version ;)

lhoward: Can you verify if $classhints should be used directly or mapped via classForClassName()?

        
        if assertedClassHints != nil {
            for assertedClassHint in assertedClassHints! {
                // FIXME check whether class hints should be subject to mapping or not
                let assertedClass : AnyClass? = NSClassFromString(assertedClassHint)
                if _isClassAllowed(assertedClass, allowedClasses: allowedClasses) {
                    classToConstruct = assertedClass
                    return true
                }
            }
        }

CoreFoundation/Base.subproj/CFRuntime.c
lhoward: We can remove this but it will not be possible to encode >2 level classes (e.g. Foo.Bar.SomeClass). We could also mangle the name ourselves but it would be better to have a proper contract with the Swift runtime for transforming classes to names and the converse. If we do not care about interoperability with Darwin archives then such a putative API does not need to expose mangled names.

Foundation/NSCharacterSet.swift
```diff
@@ -45,6 +45,26 @@ public class NSCharacterSet : NSObject, NSCopying, NSMutableCopying, NSCoding {
         return unsafeBitCast(self, CFMutableCharacterSetRef.self)
     }
     
+    public override var hash: Int {
+        get {
+            return Int(CFHash(_cfObject))
```
phausler: don't these need to be a bitPattern?

phausler:
I know this might be some busy work; but I would like to merge in a few things out of this that are not associated with the coding changes - would it be reasonable to split out some of these into individual pull requests? that way I can start merging the non NSCoding associated versions (e.g. the linker script fix etc) because I need to ensure that we keep the CI happy. Alternatively I could manually branch out parts of this and cherry pick - your choice.

lhoward:
I've created SR branches for the ones I filed, once you've integrated those I can break out [SR-377] more finely grained if required.

phausler:
I think SR-377 is intrinsic to the NSCoder work. Thank you very much for breaking this up!

lhoward:
Compiler issue is fixed, so this is ready for integration pending final review.

phausler:
Builds cleanly on ubuntu; tests all pass
For Mac OS X builds this will require the latest stdlib i presume?

lhoward:
@phausler, yes current stdlib is required for _typeByName().

phausler:
I just merged in some other items that will require updates to swiftc so it will be reasonable to require a rev on the stdlib for Foundation; I will probably have a full review finished by tonight on this. Thanks for all of your hard work/patience with this.

Foundation/NSKeyedUnarchiver.swift
```diff
@@ -0,0 +1,955 @@
...
+    
+    private static func _supportsSecureCoding(clsv : AnyClass) -> Bool {
+        if let secureCodable = clsv as? NSSecureCoding.Type {
+            return secureCodable.supportsSecureCoding()
+        }
+        
+        return false
+    }
+    
+    // FIXME is there a better way to do this with Swift stdlib?
```
lhoward: Is there a way to do a dynamic cast given types (rather than instances) in the language itself? We could call swift_dynamicCastMetatype() but that's not public.

Foundation/NSKeyedArchiver.swift
```diff
@@ -7,183 +7,832 @@
...
+        self._equality = {
+            if let other = $0 as? T {
+                return hashableObject == other
+            }
+            return false
+        }
+    }
+    
+    init(_ object: AnyObject) {
+        // FIXME can't we check for Hashable directly?
```
lhoward: May want to document the implications of this (only NSObjects will be uniqued in the object graph because otherwise the pointer itself is used as the hash code).

Foundation/NSKeyedUnarchiver.swift
```diff
@@ -0,0 +1,955 @@
...
+            let assertedClass : AnyClass? = _classForClassName(assertedClassName!)
+            if _isClassAllowed(assertedClass, allowedClasses: allowedClasses) {
+                classToConstruct = assertedClass
+                return true
+            }
+        }
+        
+        if assertedClassHints != nil {
+            for assertedClassHint in assertedClassHints! {
+                // FIXME check whether class hints should be subject to mapping or not
```
lhoward: Please check against Foundation if you can @phausler

lhoward:
> On 18 Jan 2016, at 8:49 AM, Philippe Hausler <notifications@github.com> wrote:
> 
> I just merged in some other items that will require updates to swiftc so it will be reasonable to require a rev on the stdlib for Foundation; I will probably have a full review finished by tonight on this. Thanks for all of your hard work/patience with this.
> 

FWIW we require a rev on swiftc as well (in order to support type lookup for  types that don’t explicitly conform to protocols – no more need for dummy protocols). But obviously swiftc and stdlib come as one, so pointing this out as a curiosity only.

lhoward:
Just cleaned up the commit messages, no code changed.

lhoward:
Cool!

> On 27 Jan 2016, at 5:52 AM, Philippe Hausler <notifications@github.com> wrote:
> 
> Merged #184 <https://github.com/apple/swift-corelibs-foundation/pull/184>.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/apple/swift-corelibs-foundation/pull/184#event-526932528>.
> 

--
www.lukehoward.com
soundcloud.com/lukehoward

PR #185, opened 2015-12-29
Symlinks and standardizing

HeMet:
Implement `NSURL.URLByResolvingSymlinksInPath`, `NSURL.URLByStandardizingPath`, `NSString.stringByResolvingSymlinksInPath`, `NSString.stringByStandardizingPath` and `NSString.stringByExpandingTildeInPath`.

Single test fails on Linux in VirtualBox VM. See `TestNSString.swift`, line 670. Could not test it out of VM.


phausler:
This fails on ubuntu:
TestFoundation/TestNSString.swift:670: error: TestNSString.test_stringByExpandingTildeInPath : Return copy of reciver if home directory could no be resolved.

phausler:
this also fails as well for some reason (on actual linux hardware; not a vm)
TestFoundation/TestNSJSONSerialization.swift:312: error: TestNSJSONSerialization.test_deserialize_allowFragments :

HeMet:
@phausler What should we do with `test_stringByExpandingTildeInPath`? Also how do my PR affect JSON serialization?

phausler:
I have no clue why it was making the JSON serialization fail. I will try rebasing onto master and perhaps that fixes it.
Per the test_stringByExpandingTildeInPath issue we can disable the test for now since the code looks pretty reasonable to me; likely there is an edge case that we are not handling yet; most of my linux-fu is not user land code so I will have to do some research.

phausler:
Tested again against master (some extra things were fixed that were associated with the previous failures)

New failure occurs here on linux:
```swift
            let path: NSString = "/tmp/.."
            let result = path.stringByResolvingSymlinksInPath
            XCTAssertEqual(result, "/private", "For absolute paths, all symbolic links are guaranteed to be removed.")
```

HeMet:
@phausler It looks outdated. It should be
```swift
do {
    let path: NSString = "/tmp/.."
    let result = path.stringByResolvingSymlinksInPath
    
    #if os(OSX)
    let expected = "/private"
    #else
    let expected = "/"
    #endif
    
    XCTAssertEqual(result, expected, "For absolute paths, all symbolic links are guaranteed to be removed.")
}
```

https://github.com/HeMet/swift-corelibs-foundation/commit/29536c7819735cc93c9028654f12e7361db29b70
https://github.com/HeMet/swift-corelibs-foundation/blob/symlinks-and-standardizing/TestFoundation/TestNSString.swift#L568

phausler:
I was able to nail down the error, NSHomeDirectoryForUser was misbehaving and not using the result value correctly (now fixed and the tests pass)

PR #186, opened 2015-12-29
Initial implementation of NSOrderedSet

mcudich:
This is a fairly large class, so I've gone through and only implemented basic functionality. If the implementation approach makes sense, I can continue to fill this out.

Tested on OS X and Ubuntu 14.04.

### DIFF
--- a/Foundation/NSOrderedSet.swift
+++ b/Foundation/NSOrderedSet.swift
@@ -109,20 +109,6 @@ public class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
         _entries[object] = entry
     }
 
-    private func _removeEntry(object: AnyObject) {
-        guard containsObject(object), let object = object as? NSObject, entry = _entries[object] else {
-            return
-        }
-
-        entry.previousEntry?.nextEntry = entry.nextEntry
-        entry.nextEntry?.previousEntry = entry.previousEntry
-
-        _entries.removeValueForKey(object)
-        if entry === _headEntry! {
-            _headEntry = nil
-        }
-    }
-
     private func _insertEntries(objects: UnsafePointer<AnyObject?>, count cnt: Int) {
         let buffer = UnsafeBufferPointer(start: objects, count: cnt)
         for obj in buffer {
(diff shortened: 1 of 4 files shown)


Foundation/NSOrderedSet.swift
```diff
@@ -37,38 +48,151 @@ public class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
...
-    public func objectAtIndex(idx: Int) -> AnyObject { NSUnimplemented() }
-    public func indexOfObject(object: AnyObject) -> Int { NSUnimplemented() }
-    public override init() { NSUnimplemented() }
-    public init(objects: UnsafePointer<AnyObject?>, count cnt: Int) { NSUnimplemented() }
+    public var count: Int {
+        return _entries.count
+    }
+
+    public func objectAtIndex(idx: Int) -> AnyObject {
+        return _entryAtIndex(idx).object
```
phausler: so this is O(n)? could perhaps it be a Array/NSArray? those would be better access time perhaps.

mcudich: Yes. This was a conscious decision, where I opted to optimize for insertion/deletion time, rather than access-by-index time. With an array, you take a significant hit any time you need to insert or remove an entry from the set.

Do you have a sense of which usage type is better to optimize for?

sgl0v: Original implementation of the `NSOrderedSet`/`NSMutableOrderedSet` uses array and set internally (WWDC 2013, Session 224). This doubles memory usage, but make it possible to get the best of both worlds.

mcudich: That's right, except that using an array as a backing store is more costly when it comes to insertion and deletion (O(n) in most cases). The proposal I have guarantees O(1) performance for everything except getting an item by its index.

The trade-off to consider here is whether we want to optimize for accessing by index, at the cost of expensive mutation, or optimize for mutation at the cost of accessing by index.

I'm happy to go either route. I don't feel like I have enough data to state which case makes most sense to optimize for.

mcudich: I've gone ahead and done an implementation based on the original implementation - let me know if you prefer this approach over the linked-list version. Thanks!

phausler:
this seems much more of an analog of the darwin version behaviorally :+1:

PR #187, opened 2015-12-29
Implemented copyWithZone, mutableCopyWithZone methods for various collection classes

sgl0v:
Implemented `copyWithZone`, `mutableCopyWithZone` methods for the `NSArray`, `NSMutableArray`, `NSDictionary`, `NSMutableDictionary`, `NSSet`, `NSMutableSet` classes

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -84,7 +84,15 @@ public class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NS
     }
     
     public func copyWithZone(zone: NSZone) -> AnyObject {
-        return self
+        if self.dynamicType === NSArray.self {
+            // return self for immutable type
+            return self
+        } else if self.dynamicType === NSMutableArray.self {
+            let array = NSArray()
+            array._storage = self._storage
+            return array
+        }
+        return NSArray(array: self.allObjects)
     }
     
     public override func mutableCopy() -> AnyObject {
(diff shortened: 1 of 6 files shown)


Foundation/NSArray.swift
```diff
@@ -84,7 +84,7 @@ public class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NS
     }
     
     public func copyWithZone(zone: NSZone) -> AnyObject {
-        return self
+        return NSArray(array: self.allObjects)
```
phausler: NSArray is immutable so it just should return self, subclasses need to do the O(n) copy

Foundation/NSDictionary.swift
```diff
@@ -152,7 +152,7 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
     }
     
     public func copyWithZone(zone: NSZone) -> AnyObject {
-        NSUnimplemented()
+        return NSDictionary(objects: self.allValues, forKeys: self.allKeys as! [NSObject])
```
phausler: this should just return self, same thing as NSArray; NSMutableDictionary and other subclasses should use the O(n) copy and NSDictionary (immutable) should just be a O(1) return of self

sgl0v:
Updated PR according to the comments above.

phausler:
Unfortunately we have a compiler crash in this one:
```
1.	While emitting SIL for 'copyWithZone' at Foundation/NSDictionary.swift:154:12
```

sgl0v:
Fixed the compiler crash by replacing casting `self.allKeys as! [NSObject]` with `self.allKeys.map({ $0 as! NSObject})`

phausler:
looks like that got it! that one keeps cropping up sadly

PR #189, opened 2015-12-30
NSOperation initial implementation.

ochococo:
- `NSOperation` initial implementation.
- `NSOperationQueue` initial implementation.
- `NSBlockOperation` initial implementation.
- Added some tests.

### DIFF
--- a/Foundation/NSOperation.swift
+++ b/Foundation/NSOperation.swift
@@ -365,7 +365,7 @@ public class NSOperationQueue : NSObject {
     private func operationsReadyToExecute(ops: [NSOperation]) -> [NSOperation] {
         let readyOperations = ops.filter({ operation in operation.ready })
         let readyAndSortedOperations = readyOperations.sort({ operation1, operation2 in
-            operation1.qualityOfService.rawValue > operation2.qualityOfService.rawValue
+            operation1.queuePriority.rawValue > operation2.queuePriority.rawValue
         })
 
         return readyAndSortedOperations
--- a/Foundation/NSOperation.swift
+++ b/Foundation/NSOperation.swift
@@ -296,7 +296,15 @@ public class NSOperationQueue : NSObject {
 
     private var _suspendedLock = NSLock()
 
-    public var name: String?
+    public var name: String? {
+        didSet {
+            _underlyingQueue = concurrentQueueWithName(name)
+        }
+    }
+
+    private func concurrentQueueWithName(name: String?) -> DispatchQueueBridge {
+        return DispatchQueueBridge(name: name ?? "NSOperationQueue: \(NSUUID().UUIDString)", type: .Concurrent)
+    }
 
     public var qualityOfService: NSQualityOfService = .Default  // Unimplemented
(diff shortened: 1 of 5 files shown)


Foundation/NSOperation.swift
```diff
@@ -71,82 +140,392 @@ public class NSOperation : NSObject {
...
 
-    public var executionBlocks: [() -> Void] {
-        NSUnimplemented()
+    public var executionBlocks: [() -> ()] {
+        _executionBlocksLock.lock()
+        defer { _executionBlocksLock.unlock() }
+        return _executionBlocks
+    }
+
+    public override func start() {
```
phausler: start should actually be implemented in the abstract base class; since most operations will be subclasses of NSOperation

Foundation/NSOperation.swift
```diff
@@ -71,82 +140,392 @@ public class NSOperation : NSObject {
...
+    private func operationsReadyToExecute(ops: [NSOperation]) -> [NSOperation] {
+        let readyOperations = ops.filter({ operation in operation.ready })
+        let readyAndSortedOperations = readyOperations.sort({ operation1, operation2 in
+            operation1.queuePriority.rawValue > operation2.queuePriority.rawValue
+        })
+
+        return readyAndSortedOperations
+    }
+
+    private func executeOperationsAndWait(wait: Bool) {
```
phausler: this should actually just enqueue blocks to a dispatch queue; the behavior listed will incur an overhead to require re-calling the execution upon finishing an operation.

phausler:
After looking at this more there are some fundamental issues that i think would need to be addressed:
1) we need to build and link against libdispatch on linux for this to even be possible
2) I think the execution model could use a bit more refinement to be in parity for dependency graphs and inter-queue graphs; (this is a contrived example) 
an operation could have 2 dependencies that make it not ready, one of which could be in a completely different operation queue and the other could be made to be finished in the same thread that enqueued the primary operation later down the road.
3) for actual cases of operations that return immediately after main (but are not finished) we need property behaviors (a proposal currently on the slate for the evolution list)

I have been thinking on an implementation of NSOQ and NSOp for a while now of how we could implement it fully here; here couple of take aways from what I have learned and what may result in a good approximation to the behavior of NSOQ.

Each operation should have a dispatch_group. This group would meter the execution of the operation to it's group of dependencies; when a dep is added that dep will enter the group of it's parent; when it is cancelled or it is finished that dependent operation will leave the group. That way the parent operation would have then a "wait until all dependencies are finished/cancelled" method. After that waits it should check to see if the operation is ready (since that can be overridden by the subclassers) a property behavior or new API would then claim the operation to be ready. It would then execute.

The semaphore approach is the correct way to gate the max number of operations; however there are a few caveats to think about; what happens when you have a max concurrent operation count of 1? and what happens when you have a max concurrent operation count of "unlimited"? both of these cases can be described as a nil semaphore backing either a serial dispatch_queue or a concurrent dispatch_queue.

ochococo:
"we need to build and link against libdispatch on linux for this to even be possible"

PR #194, opened 2015-12-30
[SR-414] isEqual/hash/description for bridged CF objects

lhoward:
CF bridged objects are not correctly implementing Equatable because they do not provide an isEqual method. Whilst we're at it, implement hash and description.

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -149,7 +149,7 @@ public class NSCalendar : NSObject, NSCopying, NSSecureCoding {
     
     public override var hash: Int {
         get {
-            return Int(CFHash(_cfObject))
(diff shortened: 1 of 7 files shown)


Foundation/NSCalendar.swift
```diff
@@ -147,6 +147,26 @@ public class NSCalendar : NSObject, NSCopying, NSSecureCoding {
         }
     }
     
+    public override var hash: Int {
+        get {
+            return Int(CFHash(_cfObject))
```
phausler: I still think these need to be bit pattern for the same reason that NSCFString's hash needed it

lhoward: think this is a merge error, fixing

PR #197, opened 2015-12-31
[SR-404]: use linker script to export protocol conformances start

lhoward:


phausler:
doesn't -T replace the script entirely? that would mean that we are smashing the previous -Wl,@ script no?

phausler:
err further reading of the manual reveals that is safe since `@` gets expanded to additional arguments first

lhoward:
/usr/bin/x86_64-linux-gnu-ld: cannot find @: No such file or directory
/usr/bin/x86_64-linux-gnu-ld: warning: /home/lukeh/CVSRoot/apple/build/Ninja-DebugAssert/swift-linux-x86_64/lib/swift/linux/x86_64/swift.ld contains output sections; did you forget -T?
clang-3.8: error: linker command failed with exit code 1 (use -v to see invocation)

phausler:
This probably means we can remove some of the extra copy and mutableCopy overrides littered about since that was a failure of `as?` conformance checks

PR #198, opened 2015-12-31
Use the prefix for CoreFoundation headers/module data

ezephir:
This allows Foundation to be installed with a prefix that is not '/usr' and
resolves [SR-422](https://bugs.swift.org/browse/SR-422).

### DIFF
--- a/build.py
+++ b/build.py
@@ -25,12 +25,12 @@ elif Configuration.current.target.sdk == OSType.MacOSX:
 
 # For now, we do not distinguish between public and private headers (they are all private to Foundation)
 # These are really part of CF, which should ultimately be a separate target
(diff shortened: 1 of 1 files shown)


ezephir:
I've validated this test under Ubuntu 14.04 by running the command
`../swift/utils/build-script -R --foundation --xctest -- --install-foundation --install-destdir=<<some prefix>> --install-prefix=<<somedir>>`
as a full, clean build.

The tests for Foundation still pass, which I ran using `ninja test`

I haven't attempted any validation on OS X because I have the understanding that the Xcode project is used on that platform.

Is there anything else that I ought to do to validate this change?

phausler:
looks pretty reasonable to me.

PR #202, opened 2016-01-01
[SR-432] temporary file creation API

lhoward:
for NSKeyedArchiver


phausler:
NSData has `makeTemporaryFileInDirectory` we should funnel some of these together into one spot. Might be worth extracting it into it's own file (perhaps named NSPlatform since it is similar to the CF side) since I bet there are a few others that will be in this area of usage.

lhoward:
Noted will fix.

PR #206, opened 2016-01-02
Improve NSUserName NSHomeDirectoryForUser

HeMet:
* `getlogin_r` replaced with get `getpwuid_r->name`. According to the manpage it should be more reliable.
* use `sysconf(Int32(_SC_GETPW_R_SIZE_MAX))` to get recommended size of a buffer
* for the current user check `$HOME` environment variable first

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -603,7 +603,7 @@ public func NSHomeDirectory() -> String {
 public func NSHomeDirectoryForUser(user: String?) -> String? {
     let usr = user ?? NSUserName()
     var info = passwd()
-    let recommendedBufSize = sysconf(_SC_GETPW_R_SIZE_MAX)
+    let recommendedBufSize = sysconf(Int32(_SC_GETPW_R_SIZE_MAX))
     let bufSize = recommendedBufSize > 0 ? recommendedBufSize : Int(BUFSIZ * 10)
     var buffer = [Int8](count: bufSize, repeatedValue: 0)
     var result: UnsafeMutablePointer<passwd> = nil
@@ -620,7 +620,7 @@ public func NSUserName() -> String {
     let euid = geteuid()
     
     var info = passwd()
-    let recommendedBufSize = sysconf(_SC_GETPW_R_SIZE_MAX)
+    let recommendedBufSize = sysconf(Int32(_SC_GETPW_R_SIZE_MAX))
     let bufSize = recommendedBufSize > 0 ? recommendedBufSize : Int(BUFSIZ * 10)
     var buffer = [Int8](count: bufSize, repeatedValue: 0)
     var result: UnsafeMutablePointer<passwd> = nil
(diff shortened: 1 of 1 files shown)


Foundation/NSPathUtilities.swift
```diff
@@ -616,12 +628,22 @@ public func NSHomeDirectoryForUser(user: String?) -> String? {
...
+    var result: UnsafeMutablePointer<passwd> = nil
+    
+    let errno = getpwuid_r(euid, &info, &buffer, bufSize, &result)
+    if errno == 0 && result != nil && result.memory.pw_name != nil {
+        return String.fromCString(result.memory.pw_name)!
     }
-    fatalError("Could not get current logon name.")
+    
+    // maybe NSUserName should return optional string
+    print("Could not get current logon name. \(errno) Return `unknown`.")
```
HeMet: I placed workaround mentioned at https://bugs.swift.org/browse/SR-437 . If this PR fixes the issue it should be removed, I think.

Foundation/NSPathUtilities.swift
```diff
@@ -616,12 +628,22 @@ public func NSHomeDirectoryForUser(user: String?) -> String? {
 }
 
 public func NSUserName() -> String {
```
phausler: Looking at this further; it might be reasonable to actually just call out to CFCopyUserName

Foundation/NSPathUtilities.swift
```diff
@@ -601,9 +601,21 @@ public func NSHomeDirectory() -> String {
 }
 
 public func NSHomeDirectoryForUser(user: String?) -> String? {
```
phausler: also this might be better shifted to just call out to CFCopyHomeDirectoryURLForUser and any additional logic should be transitioned there

HeMet:
@phausler It's done.

Foundation/NSPathUtilities.swift
```diff
@@ -601,27 +601,18 @@ public func NSHomeDirectory() -> String {
...
 
 public func NSUserName() -> String {
-    let bufSize = Int(BUFSIZ)
-    var buffer = [Int8](count: bufSize, repeatedValue: 0)
-    if getlogin_r(&buffer, bufSize) == 0 {
-        return String.fromCString(buffer)!
-    }
-    fatalError("Could not get current logon name.")
+    let userName = CFCopyUserName().takeRetainedValue()
+    return String(userName)
```
phausler: Do we actually have an initializer for String that takes a CFStringRef? perhaps it might be better to use the `._swiftObject` property extension on CFStringRef instead.

HeMet:
@phausler It seems `public init<T>(_ instance: T)` was used. I replaced it with `._swiftObject`. It passes the tests, but crashes `SourceKit` repeatedly.

phausler:
seems to pass on linux just fine. I would suggest cleaning out your derived data if SourceKit is having issues. Sometimes it gets a cruft module map compilation that confuses it (primarily since we are re-defining CF and Foundation)

PR #207, opened 2016-01-02
[SR-439] Cached floating-point CF values do not bridge

lhoward:
If a bridged CF type is instantiated before __CFInitializeSwift() is called, then those types will not be bridged. This created unusual and unexpected results when initializing floating point numbers containing cached values such as 0.0, 1.0 and NaN: decoding a property list containing these would assert fail but other numbers would be fine.

This patch fixes this bug by moving the initialization of CFNumber constants to a new function, __CFNumberInitialize(), which is called after the Swift bridge is initialized. This patch is enabled but should not regress non-Swift builds.

CFNull and CFBooleans are instantiated before __CFInitializeSwift() is called; if these types are ever bridged then a similar fix must be made.


phausler:
seems pretty reasonable, passes on ubuntu and mac os x

PR #208, opened 2016-01-02
[SR-440] warnings cleanup

lhoward:
Fix a bunch of compile warnings.

### DIFF
--- a/CoreFoundation/Base.subproj/CFBase.h
+++ b/CoreFoundation/Base.subproj/CFBase.h
@@ -80,8 +80,10 @@
 #include <stdbool.h>
 #endif
 
-  #if (TARGET_OS_MAC && !(TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)) || (TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)
+  #if ((TARGET_OS_MAC && !(TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)) || (TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)) && !DEPLOYMENT_RUNTIME_SWIFT
     #include <libkern/OSTypes.h>
+  #else
+    #include <pthread.h>
   #endif
 
 #if !defined(__MACTYPES__)
@@ -122,6 +124,12 @@
     typedef UInt32                  UTF32Char;
     typedef UInt16                  UTF16Char;
     typedef UInt8                   UTF8Char;
+#endif
+
+#if DEPLOYMENT_RUNTIME_SWIFT
+    typedef _Bool                   CFBool; /* avoid DarwinBoolean confusion with Swift compiler */
+#else
+    typedef Boolean                 CFBool;
 #endif
 
 #if !defined(CF_EXTERN_C_BEGIN)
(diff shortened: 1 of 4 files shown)


CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -124,6 +124,16 @@
...
 
+/*
+ * The Swift compiler special cases Boolean on Darwin, this is necessary to avoid
+ * compile warnings.
+ */
+#if CF_BUILDING_CF || !DEPLOYMENT_RUNTIME_SWIFT
+typedef Boolean SWIFT_BOOL;
+#else
+typedef _Bool SWIFT_BOOL;
+#endif
```
phausler: I wonder if there is a better way to do this; I am worried that there will be a size differential in that _Bool may end up being different than Boolean and we could have corruption because of that. For DEPLOYMENT_RUNTIME_SWIFT targets it might be interesting to find the root import of Boolean and conditionally exclude that to define a CFBool similarly to how we define a CFIndex etc.

phausler: I traced down the include that hits it:
```c
  #if ((TARGET_OS_MAC && !(TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)) || (TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)) && !DEPLOYMENT_RUNTIME_SWIFT
    #include <libkern/OSTypes.h>
  #else
    #include <pthread.h>
  #endif
```

lhoward: > On 3 Jan 2016, at 7:48 AM, Philippe Hausler <notifications@github.com> wrote:
> 
>   #if ((TARGET_OS_MAC && !(TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)) || (TARGET_OS_EMBEDDED || TARGET_OS_IPHONE)) && !DEPLOYMENT_RUNTIME_SWIFT
>     #include <libkern/OSTypes.h>
>   #else
>     #include <pthread.h>
>   #endif

Is this a suggested fix? NSTask.swift breaks with this change.

lhoward: > I wonder if there is a better way to do this; I am worried that there will be a size differential in that _Bool may end up being different than Boolean and we could have corruption because of that. For DEPLOYMENT_RUNTIME_SWIFT targets it might be interesting to find the root import of Boolean and conditionally exclude that to define a CFBool similarly to how we define a CFIndex etc.
> 
> 

Agree my fix is ugly, will look into.

lhoward: > On 3 Jan 2016, at 7:05 AM, Philippe Hausler <notifications@github.com> wrote:
> 
> I wonder if there is a better way to do this; I am worried that there will be a size differential in that _Bool may end up being different than Boolean and we could have corruption because of that. For DEPLOYMENT_RUNTIME_SWIFT targets it might be interesting to find the root import of Boolean and conditionally exclude that to define a CFBool similarly to how we define a CFIndex etc.
> 


Have pushed an alternate fix. I don’t really understand what’s going on with DarwinBoolean so maybe there’s still a cleaner approach.

lhoward:
Fixed: Define CFBool when __MACTYPES__ and _OS_OSTYPES_H are defined

lhoward:
I don't know enough about DarwinBoolean to comment on whether it's an issue with the compiler/stdlib.

PR #209, opened 2016-01-03
NSKeyedArchiver/NSKeyedUnarchiver integration point

lhoward:
This branch contains all of the nscoding branch except for the actual implementation of NSKeyedArchiver and NSKeyedUnarchiver (which is awaiting stdlib API).

If you'd like to start integrating, you can use this, otherwise we can wait for [SR-381] to be approved and do it in one hit.


lhoward:
Closing for now, still need to fix some bugs.

PR #210, opened 2016-01-03
NSKeyedArchiver/NSKeyedUnarchiver integration point

lhoward:
Contains everything except for the archiver itself which is dependent on stdlib SPI approval [SR-381].


Foundation/NSGeometry.swift
```diff
@@ -91,6 +95,61 @@ public func ==(lhs: CGPoint, rhs: CGPoint) -> Bool {
...
 
+extension CGPoint: NSSpecialValueCoding {
+    init(bytes: UnsafePointer<Void>) {
+        let buffer = UnsafePointer<CGFloat>(bytes)
+
+        self.x = buffer.memory
+        self.y = buffer.advancedBy(1).memory
+    }
+    
+    public init?(coder aDecoder: NSCoder) {
```
phausler: These are API changes; id rather not have these be public if they are needed to make the special values work since it will mean that there is disparity; now perhaps with a later change we might want to consider this a first step for value-type coding but that should be considered  separately from the functionality of NSCoding

Foundation/NSGeometry.swift
```diff
@@ -91,6 +95,61 @@ public func ==(lhs: CGPoint, rhs: CGPoint) -> Bool {
...
+            return false
+        }
+    }
+    
+    var hash: Int {
+        return self.x.hash &+ self.y.hash
+    }
+}
+
+extension CGPoint: CustomStringConvertible {
```
phausler: this is an API change; granted pretty reasonable imho but this is separate from the NSCoding functionality and should be considered on it's own accord.

Foundation/NSGeometry.swift
```diff
@@ -187,6 +357,70 @@ public struct NSEdgeInsets {
     }
 }
 
+extension NSEdgeInsets: Equatable { }
```
phausler: also a API change

Foundation/NSRange.swift
```diff
@@ -50,6 +50,83 @@ extension NSRange {
     }
 }
 
+extension NSRange: Equatable { }
```
phausler: NSRange does not adhere to Equatable; this would be an API change

TestFoundation/TestNSValue.swift
```diff
@@ -0,0 +1,121 @@
...
+    }
+    
+    func test_valueWithNSRange() {
+        let range = NSMakeRange(1, 2)
+        let value = NSValue(range: range)
+        XCTAssertEqual(value.rangeValue, range)
+        
+        var expected = NSRange()
+        value.getValue(&expected)
+        XCTAssertEqual(expected, range)
```
phausler: Running this against darwin yields this error; so NSRange is not equatable in swift
```
error: cannot invoke 'XCTAssertEqual' with an argument list of type '(NSRange, NSRange)'
        XCTAssertEqual(value.rangeValue, range)
```

Foundation/NSGeometry.swift
lhoward: NSSpecialValueCoding is an internal protocol however it is true that the coding methods are public. How about I remove NSSpecialValueCoding inheriting from NSSecureCoding and make all these methods internal? It's not an API change if the methods are not exported, right?

Foundation/NSGeometry.swift
lhoward: these methods are internal, there is no way to access them except through NSSpecialValue. I can give them a different name if you think it is confusing with the similarly named NSObject methods.

Foundation/NSGeometry.swift
lhoward: fixed

TestFoundation/TestNSValue.swift
lhoward: fixed

Foundation/NSGeometry.swift
lhoward: if you think it'd be clearer, I can prefix all these internal methods with an underscore, e.g.

{{code}}
internal protocol NSSpecialValueCoding {
    static func _objCType() -> String
    init(bytes value: UnsafePointer<Void>)
    init?(coder aDecoder: NSCoder)
    func _encodeWithCoder(aCoder: NSCoder)
    func _getValue(value: UnsafeMutablePointer<Void>)
    func _isEqual(value: Any) -> Bool
    var _hash: Int { get }
}
{{code}}

phausler: eh i guess they are fine as is then

lhoward: cool, will leave as is

lhoward: let me know if you think adding CustomStringConvertible counts as an API change, I can see an argument that it does

lhoward:
Deleting this branch. Please see PR #184.

PR #211, opened 2016-01-03
Add a missing constructor for NSString

drewcrawford:
On Darwin we can do

    let argument: UInt8 = 75
    let string = NSString(format: "%02X", argument)

This is because Darwin NSString has constructor

    public convenience init(format: NSString, _ args: CVarArgType...)

This patch adds, implements, and tests this constructor.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -1232,6 +1232,13 @@ extension NSString {
         NSUnimplemented()    
     }
     
+    public convenience init(format: NSString, _ args: CVarArgType...) {
+        let str = withVaList(args) { (vaPtr) -> CFString! in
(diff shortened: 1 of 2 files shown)


Foundation/NSString.swift
```diff
@@ -1232,6 +1232,13 @@ extension NSString {
         NSUnimplemented()    
     }
     
+    public convenience init(format: NSString, _ args: CVarArgType...) {
```
phausler: it is very strange that the importer grabs this as NSString instead of String. But it is in parity with Darwin.

There is also the method
```swift
public convenience init(format: NSString, locale: NSLocale?, _ args: CVarArgType...)
```
as well

PR #214, opened 2016-01-04
[configure] Refactor to be compatible with Python 2 or 3

RLovelett:
Other projects have asked that their Python enabled scripts were refactored to be compatible with Python 2 or 3. This patch continues that vein in the `corelibs-foundation` project.


phausler:
looks good to me, seems to build and configure just fine with these changes

PR #216, opened 2016-01-04
[SR-458] add NSNull test to test list

lhoward:


phausler:
I wonder if main.swift should be generated...

lhoward:
Probably.

PR #217, opened 2016-01-06
slight refactor to remove redundancy

res1233:
first commit, since I've been told I need to commit more in my life. Minor change, removes a little redundancy. Shouldn't break anything.


phausler:
looks good to me

PR #218, opened 2016-01-06
Improve NSString.completePathIntoString

HeMet:
Now it works with relative paths, symlinks etc. 
Fix behavior when string point to a directory and there are other directories and(or) files with similar names. Previously it searches for all items in directory always. Now if there is no trailing `/` it starts from the parent directory.

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -451,10 +451,7 @@ public extension NSString {
             return 0
         }
         
-        guard let url = NSURL(string: _storage) else {
-            return 0
-        }
-        
+        let url = NSURL(fileURLWithPath: _storage)
         let normalizedTypes = flag ? filterTypes : filterTypes?.map { $0.lowercaseString }
         let types = Set<String>(normalizedTypes ?? [])
         let compareOptions = flag ? [] : NSStringCompareOptions.CaseInsensitiveSearch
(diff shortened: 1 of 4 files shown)


Foundation/NSPathUtilities.swift
```diff
@@ -446,57 +446,108 @@ public extension NSString {
...
         return matches.count
     }
+
+    internal func _stringIsPathToDirectory(path: String) -> Bool {
+        if !path.hasSuffix("/") {
+            return false
+        }
+        
+        var isDirectory = false
+        let isAbsolutePath = NSFileManager.defaultManager().fileExistsAtPath(path, isDirectory: &isDirectory)
```
phausler: `isAbsolutePath` seems like a misleading name; `exists` perhaps just to avoid future confusion?

Foundation/NSPathUtilities.swift
```diff
@@ -446,57 +446,108 @@ public extension NSString {
...
         }
         
-        guard let url = NSURL(string: _storage) else {
-            return 0
-        }
-        
-        let normalizedTypes = flag ? filterTypes : filterTypes?.map { $0.lowercaseString }
-        let types = Set<String>(normalizedTypes ?? [])
-        let compareOptions = flag ? [] : NSStringCompareOptions.CaseInsensitiveSearch
+        let url = NSURL(fileURLWithPath: _storage)
```
phausler: this probably shouldn't directly access _storage since it may be a NSCFString or a subclass, so _swiftObject is the better accessor here

PR #221, opened 2016-01-07
fix CFGetSystemUptime for linux

antoinepalazzolo:
The implementation of CFGetSystemUptime() for linux return incorrect values, which make NSThread.sleepForTimeInterval() behave incorrectly : 
- clock_gettime already return values in seconds and nano sec, there is no need to scale it by __CF1_TSRRate
- tv_nsec must be divided 1000000000 to get seconds, not multiplied


phausler:
good catch; __CF1_TSRRate is actually the cpu clock scale iirc which should have been 1 on linux targets.

PR #222, opened 2016-01-07
Remove redundant method.

DarrylBayliss:
Based on comments around the code. The _cfSetProductName method is not needed because people shouldn't have the ability to set the product name manually. I've therefore removed the call.


parkera:
It's not used inside this version but it is used by an external client in one specific circumstance, so we have to leave it in.

PR #223, opened 2016-01-07
NSXMLNode and Friends

rothomp3:
So this is probably larger and less organized than one would hope, but I'm not enough of a git expert to actually break up the changes properly.

There are many things that are still unimplemented or incomplete. Some things can't be done until we get factory methods or class-cluster-like behavior. That said, what is here is completely usable for parsing and creating XML documents of various sorts. As long as you don't really need DTDs, validation, or namespaces haha! So the main purpose of putting up this PR now is for comments and to break up at least a little bit what is obviously a large undertaking.

This has been tested on OS X and Linux.


CoreFoundation/Parsing.subproj/CFXMLInterface.c
```diff
@@ -216,4 +233,386 @@ void _CFXMLInterfaceSAX2UnparsedEntityDecl(_CFXMLInterfaceParserContext ctx, con
...
+    if (content == NULL) {
+        return NULL;
+    }
+    CFStringRef result = CFStringCreateWithCString(NULL, (const char*)content, kCFStringEncodingUTF8);
+    xmlFree(content);
+
+    return result;
+}
+
+const char* _getCStringFromCFString(CFStringRef string);
```
phausler: where is this used in here?

rothomp3: Whoops! That particular last line isn't used, it didn't get deleted when I realized I didn't need such a method. I'll clean that up as soon as I get to a computer.
Thanks,Robert Thompson

Sent from Outlook on my iPhone




On Thu, Jan 7, 2016 at 4:11 PM -0800, "Philippe Hausler" <notifications@github.com> wrote:












In CoreFoundation/Parsing.subproj/CFXMLInterface.c:
> +void _CFXMLNodeSetName(_CFXMLNodePtr node, const char* name) {
> +    xmlNodeSetName(node, (const xmlChar*)name);
> +}
> +
> +CFStringRef _CFXMLNodeGetContent(_CFXMLNodePtr node) {
> +    xmlChar* content = xmlNodeGetContent(node);
> +    if (content == NULL) {
> +        return NULL;
> +    }
> +    CFStringRef result = CFStringCreateWithCString(NULL, (const char*)content, kCFStringEncodingUTF8);
> +    xmlFree(content);
> +
> +    return result;
> +}
> +
> +const char* _getCStringFromCFString(CFStringRef string);



where is this used in here?

Foundation/NSXMLDocument.swift
```diff
@@ -53,147 +53,273 @@ public enum NSXMLDocumentContentKind : UInt {
...
+        }
+    }
     /*!
         @method initWithXMLString:options:error:
         @abstract Returns a document created from either XML or HTML, if the HTMLTidy option is set. Parse errors are returned in <tt>error</tt>.
     */
-    public convenience init(XMLString string: String, options mask: Int) throws { NSUnimplemented() }
-    
+    public convenience init(XMLString string: String, options mask: Int) throws {
+        let data: NSData = string.withCString {
```
phausler: The darwin version effectively just uses:
CFStringCreateExternalRepresentation with UTF8 encoding

Foundation/NSXMLElement.swift
```diff
@@ -7,165 +7,271 @@
...
-    
+    public var attributes: [NSXMLNode]? {
+        get {
+            var result: [NSXMLNode] = []
+            var attribute = _CFXMLNodeProperties(_xmlNode)
+            while attribute != nil {
+                result.append(NSXMLNode._objectNodeForNode(attribute))
+                attribute = _CFXMLNodeGetNextSibling(attribute)
+            }
+            return result.count > 0 ? result : nil // This appears to be how Darwin does it
```
phausler: effectively that is the case; the attributes are stored as an ivar which is lazily constructed

Foundation/NSXMLNode.swift
```diff
@@ -50,291 +51,832 @@ public enum NSXMLNodeKind : UInt {
...
+        if case .DocumentKind = kind { // documents have to be free'd explicitly as a document in order to not leak memory
+            _CFXMLFreeDocument(_CFXMLDocPtr(_xmlNode))
+        } else {
+            _CFXMLFreeNode(_xmlNode)
+        }
+    }
+
+    internal init(ptr: _CFXMLNodePtr) {
+        precondition(ptr != nil)
+        precondition(_CFXMLNodeGetPrivateData(ptr) == nil, "Only one NSXMLNode per xmlNodePtr allowed")
```
phausler: :+1: I like this precondition

Foundation/NSXMLNode.swift
```diff
@@ -50,291 +51,832 @@ public enum NSXMLNodeKind : UInt {
...
+    // see above
+    internal func _replaceChildAtIndex(index: Int, withNode node: NSXMLNode) {
+        let child = childAtIndex(index)!
+        _childNodes.remove(child)
+        _CFXMLNodeReplaceNode(child._xmlNode, node._xmlNode)
+        _childNodes.insert(node)
+    }
 }
 
+extension NSXMLNode: CollectionType {
```
phausler: albeit this is pretty neat; this isn't part of the current APIs

phausler:
overall this looks like a great start, I think with the limitations of missing functionality like factory initializers this is a great baseline to work from.

Foundation/NSXMLDocument.swift
rothomp3: So in trying to use this function, I believe I have found a bug in the conversion of Swift strings to CFStrings. `CFStringCreateExternalRepresentation` calls `__CFStrLength` which in turn uses `__CFStrIsInline` to determine where to find the length. Unfortunately, that function is returning `true` for a situation where it obviously isn't true! This leads to an erroneous length being returned, and obviously we get a crash as a result.

phausler: hmm yea that looks like a bug in CFStringCreateExternalRepresentation; I guess we can update this when that gets corrected to funnel out to swift correctly

rothomp3: In the meanwhile I'm changing it to use `dataUsingEncoding` from NSString, which seems to avoid `CFStringCreateExternalRepresentation` already.

phausler: seems pretty reasonable, I just pushed a fix for the CF function

Foundation/NSXMLNode.swift
phausler: the only thing really actually blocking this PR is the API addition of CollectionType; can we safely remove this? I didn't see anything directly relying on it and I would like to consider this part separately from the rest of the PR

rothomp3: So the thing is, I love this idea, the way it so neatly fit CollectionType with so little code. And I use it elsewhere in the implementation. I would have marked it as `internal`, but of course we can't use access control on extensions that declare protocol conformance, so I couldn't. Should I get rid of it and refactor the code that uses it, or should I put the "experimental" comment on it, since it's purely additive and doesn't change or replace any existing API?

phausler: experimental are currently things that the darwin component owners have signed off on to say this is something within reasonable scope; I have not yet had a chance to float this proposal yet - for now I would say that refactoring would be preferable here for now and I can follow up on this with an evolution proposal

rothomp3: Immediately after writing my note above, I hit upon a workaround:
```
internal protocol _NSXMLNodeCollectionType: CollectionType { }

extension NSXMLNode: _NSXMLNodeCollectionType {
…
```
This allows me to keep refactoring to a minimum (including the advantages of only creating new NSXMLNode objects lazily) while not exposing any new API to the outside world.

phausler:
lgtm

PR #226, opened 2016-01-11
Implement NSURL standardizedURL property

TETRA2000:


CoreFoundation/URL.subproj/CFURL.h
```diff
@@ -193,6 +193,9 @@ CFURLRef CFURLGetBaseURL(CFURLRef anURL);
  to request that none be.
  */
 
+CF_EXPORT
+CFURLRef CFGetStandardizedURL(CFURLRef url);
```
phausler: lets not expose this as API, it should be SPI for foundation only.
Naming wise it should be _CFCopyStandardizedURL because it should return a retain count +1 from this method unless it is a direct getter.

TestFoundation/TestNSURL.swift
```diff
@@ -212,13 +211,8 @@ class TestNSURL : XCTestCase {
...
             }
             if let url = url {
-
-                // TODO: NSURL.standardizedURL isn't implemented yet.
-                var modifiedExpectedNSResult = expectedNSResult as! [String: Any]
-                modifiedExpectedNSResult["standardizedURL"] = nil
-                
                 let results = generateResults(url, pathComponent: inPathComponent, pathExtension: inPathExtension)
-                let (isEqual, differences) = compareResults(url, expected: modifiedExpectedNSResult, got: results)
+                let (isEqual, differences) = compareResults(url, expected: expectedNSResult as! [String: Any], got: results)
```
phausler: linux probably won't like this; most likely it will have to be a mapped result (have not yet tested, but I will follow up with results on tests shortly)

CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -5150,6 +5150,70 @@ static Boolean _CFURLHasFileURLScheme(CFURLRef url, Boolean *hasScheme)
...
+    //FIXME: composeFromRFC1808 expects to get CFMutableString array as pathComponents
+    CFIndex resultCount = CFArrayGetCount(resultComponents);
+    for(CFIndex i=0; i<resultLen; i++) {
+        CFStringRef str =  CFArrayGetValueAtIndex(resultComponents, i);
+        CFArraySetValueAtIndex(resultComponents, i, CFStringCreateMutableCopy(CFGetAllocator(str), 0, str));
+    }
+    
+    components1808.pathComponents = resultComponents;
+    CFURLRef resultURL = composeFromRFC1808(alloc, &components1808);
+
```
phausler: resultComponents is leaked here

CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -5150,6 +5150,70 @@ static Boolean _CFURLHasFileURLScheme(CFURLRef url, Boolean *hasScheme)
...
+        } else {
+            ++pos;
+        }
+    }
+    
+    //FIXME: composeFromRFC1808 expects to get CFMutableString array as pathComponents
+    CFIndex resultCount = CFArrayGetCount(resultComponents);
+    for(CFIndex i=0; i<resultLen; i++) {
+        CFStringRef str =  CFArrayGetValueAtIndex(resultComponents, i);
+        CFArraySetValueAtIndex(resultComponents, i, CFStringCreateMutableCopy(CFGetAllocator(str), 0, str));
```
phausler: the string copy is leaked here: CFStringCreateMutableCopy returns retain count + 1

parkera:
Why is this not using _CFURLCopyComponents with kCFURLComponentDecompositionRFC1808?

parkera:
That is - use the existing public accessor for the function instead of adding a new entry point.

CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -5150,6 +5150,70 @@ static Boolean _CFURLHasFileURLScheme(CFURLRef url, Boolean *hasScheme)
...
+        }
+    }
+    
+    CFIndex len = CFArrayGetCount(pathComponents);
+    
+    CFMutableArrayRef resultComponents = CFArrayCreateMutableCopy(kCFAllocatorDefault ,len, pathComponents);
+    CFIndex resultLen = CFArrayGetCount(resultComponents);
+    CFIndex pos = 0;
+    
+    for (int i=0; i<len && 0 < resultLen; i++) {
```
mgrebenets: Why not using `CFIndex` here instead of `int`?

TETRA2000:
@phausler @parkera @mgrebenets 
Thank you very much for your reviews!!
It's really helpful for me.

I'll fix and update my patch.
But I think it'll take a bit more time.
I'm sorry.

HeMet:
What is a point to implement it in C and not in Swift?

phausler:
@HeMet I think with Tony's suggestion (which most closely mimics how the ObjC version funnels) it would be mostly in Swift.

HeMet:
@phausler Could you point me to that suggestion?

phausler:
`_CFURLCopyComponents` with `kCFURLComponentDecompositionRFC1808`
that is roughly where things should funnel into; it does most of the logic you need to implement this method.

HeMet:
@phausler Thanks.

TETRA2000:
@parkera @HeMet @phausler @mgrebenets 

Thank you for your reviews and comments.

I fixed this patch as much as possible I can.
I know there are still problems.

By the way, there is already swift implementation of path standardization like this by now.
Should I change to use those existing code?

parkera:
This one has been open for a while with no update, so unfortunately I'm going to have to close it now. Please feel free to reopen once comments are addressed.

PR #220, opened 2016-01-06
The NSMutableArray fixes

sgl0v:
* the `replaceObjectAtIndex` uses empty range, that is why replacement works as insertion
* typo in the `removeObjectsInArray`

PR #227, opened 2016-01-12
The NSMutableArray fixes

sgl0v:
* the `replaceObjectAtIndex` uses empty range, that is why replacement works as insertion
* typo in the `removeObjectsInArray`

The copy of this [PR](https://github.com/apple/swift-corelibs-foundation/pull/220), that I closed accidentally. Fixed the compiler error related to the `test_removeObjectsInArray` test.


phausler:
looks like this has the issues addressed for the compiler crash: builds and tests fine on ubuntu 14

sgl0v:
Yeah, I got the `Segmentation fault` before. The solution was to specify the variable type explicitly in `test_removeObjectsInArray`:

`let objectsToRemove: Array<AnyObject> = ...` instead of `let objectsToRemove = ...`

PR #228, opened 2016-01-13
Implemented init(contentsOfURL:encoding:) in NSString

pgw3:


parkera:
initWithContentsOfURL: should call initWithContentsOfFile: or the other way around. There's no reason to duplicate the logic.

pgw3:
You are of course right (just felt a bit hesitant to change another person's code for my first commit to this project...). 
I have refactored it now so initWithContentsOfFile calls initWithContentsOfURL to avoid the duplicated logic. And I hope I have commited the changed code properly so the pull request is updated, just let me know if I need to do something else with this pull request.

parkera:
Looks good to me! Can you actually rebase into a single commit? Then we'll merge.

Thanks for fixing this one.

pgw3:
I'm not really sure it worked out properly when I tried to rebase it but version b2eec72 does seem to contain everything needed. If that doesn't work for you I'll try to redo it.

phausler:
This has a introduced warning:
```
Foundation/NSCFString.swift:34:5: warning: declaring a public initializer for an internal class
    public required init(string aString: String) {
```

phausler:
Other than the warning: unit tests on Ubuntu seem to pass just fine.

pgw3:
Could the warning come from SR-568? It was fixed here:
https://github.com/apple/swift-corelibs-foundation/commit/5ffd28c6b902e0ab3bced774757a87117f831a91

Do you still get the warning after that fix has been merged?

phausler:
that fix ameliorated the warning

PR #229, opened 2016-01-14
Add string constructor for NSMutableString.

drewcrawford:
On Darwin we can say

```swift
import Foundation
let a = NSMutableString(string: "Foo")
```

This constructor is not present on corelibs-foundation.

* Implement the constructor
* Test coverage
* We have to move NSString's existing version of the constructor (which shares
  the name, so we're overriding) out of its extension and into the
  class.  This is because overriding an init declared in an extension is
  "not yet supported".  This may be why we don't have the constructor
  already.
* Marking required, so as to enforce all subclasses support this.

### DIFF
--- a/Foundation/NSCFString.swift
+++ b/Foundation/NSCFString.swift
@@ -30,6 +30,10 @@ internal class _NSCFString : NSMutableString {
     required init(capacity: Int) {
         fatalError()
     }
+
+    public required init(string aString: String) {
+        fatalError("init(string:) has not been implemented")
+    }
     
     deinit {
         _CFDeinit(self)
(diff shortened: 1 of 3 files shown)


Foundation/NSCFString.swift
```diff
@@ -83,6 +87,10 @@ internal final class _NSCFConstantString : _NSCFString {
     required init(capacity: Int) {
         fatalError()
     }
+
+    required init(string aString: String) {
+        fatalError("init(string:) has not been implemented")
```
phausler: the text for this probably should be "Constant strings cannot be constructed in code" or something along those lines

PR #232, opened 2016-01-17
Initial implementation of NSFileManager.enumeratorAtPath.

drewcrawford:
This implementation wraps the existing NSPathDirectoryEnumerator to do
all the heavy lifting.

Add test coverage.

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -642,7 +642,7 @@ public class NSFileManager : NSObject {
     /* enumeratorAtPath: returns an NSDirectoryEnumerator rooted at the provided path. If the enumerator cannot be created, this returns NULL. Because NSDirectoryEnumerator is a subclass of NSEnumerator, the returned object can be used in the for...in construct.
      */
     public func enumeratorAtPath(path: String) -> NSDirectoryEnumerator? {
-        NSUnimplemented()
+        return NSPathDirectoryEnumerator(path: path)
     }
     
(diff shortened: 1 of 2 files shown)


Foundation/NSFileManager.swift
```diff
@@ -842,6 +842,44 @@ public class NSDirectoryEnumerator : NSEnumerator {
...
+        }
+        self.innerEnumerator = ie
+    }
+    
+    override func nextObject() -> AnyObject? {
+        let o = innerEnumerator.nextObject()
+        guard let url = o as? NSURL else {
+            return nil
+        }
+        let path = url.path!.stringByReplacingOccurrencesOfString(baseURL.path!+"/", withString: "")
```
phausler: perhaps it might be safer to funnel to lastPathComponent instead of doing string manipulation?

drewcrawford: It is not as simple as calling `lastPathComponent`, because we want to preserve the (variable number of) path components that follow the base URL.

We could implement this by breaking up into arrays of `pathComponents` and doing the replacement, then rebuilding the path by joining the components.  But I'm not sure if we've gained from that (besides being portable to non-POSIX platforms like Windows).

phausler: ah good point, per the windows side of things id probably imagine that we have a number of other errors with the \ versus / issue in Foundation

TestFoundation/TestNSFileManager.swift
```diff
@@ -136,6 +137,38 @@ class TestNSFileManger : XCTestCase {
...
+        } catch _ {
+            XCTFail()
+        }
+        
+        if let e = NSFileManager.defaultManager().enumeratorAtPath(basePath) {
+            var foundItems : [String] = []
+            while let item = e.nextObject() as? NSString {
+                    foundItems.append(item.bridge())
+            }
+            XCTAssertEqual(foundItems, ["item","path2","path2/item"])
```
phausler: This expects the files to be sorted in a specific order which is OS dependent. Ubuntu does not give a promise that the files on disk are in the same order as Darwin; so the foundItems should be an un-ordered collection (ala Set).

PR #234, opened 2016-01-18
Tidy up NSDateFormatter

stephencelis:
General cleanup includes:

  - Remove internal properties in favor of property observers.
  - Demote remaining internal access control to private.
  - Use guard to enforce returns.


parkera:
Overall this looks good to me. I think we may want to actually just rename `__cfObject` and `_cfObject` since those names are usually reserved for our CF toll-free bridging, and I don't think this class is actually one of those that we want to bridge that way. That is, NSDateFormatter 'has-a' CFDateFormatter, not 'is-a' CFDateFormatter.

PR #235, opened 2016-01-19
implementation of NSData rangeOfData(options:range:)

antoinepalazzolo:
add the missing implementation of rangeOfData and tests : https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20160111/000352.html

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -505,6 +505,34 @@ extension NSData {
         try writeToFile(path, options: writeOptionsMask)
     }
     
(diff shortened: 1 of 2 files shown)


phausler:
Going to merge this, but just as a reference there is a CF function to do nearly exactly this: _CFDataFindBytes, if there proves to be behavior differences (which from what I can tell the _CFDataFindBytes function may have some perf benefits) we might want to eventually migrate to that.

antoinepalazzolo:
I didn't see the CF version, which is probably better. Should I make another PR to replace it?

phausler:
If you want, but what you have is relatively close.

PR #236, opened 2016-01-19
Add NSXMLDTD and NSXMLDTDNode support

rothomp3:
Namespaces are still basically completely unsupported. There are some hack-y things in this one because libxml2 doesn't quite handle things the way Darwin's NSXMLDocument seems to. That said, everything seems to work. There is one thing that is _technically_ an API change; `objectValue` for a `NSXMLNodeKind.ElementDeclarationKind` on Darwin returns a `NSXMLElementDeclarationContent`, but that is private API that I have no access to or knowledge of despite the lack of a leading underscore. Since it's (apparently, it isn't in any public header or documentation anyway) private, are we OK to not use such a structure here?

There are two test functions I had to disable on Linux because they were segfaulting in swift_dynamicCast. This has something to do with `throw`ing an NSError.

Also, on OS X, tests were spinning forever on the `NSTask` test, but that's happening for me on an unmodified master branch too, so it's nothing to do with this code!

### DIFF
--- a/TestFoundation/TestNSXMLDocument.swift
+++ b/TestFoundation/TestNSXMLDocument.swift
@@ -34,7 +34,8 @@ class TestNSXMLDocument : XCTestCase {
             ("test_comments", test_comments),
             ("test_processingInstruction", test_processingInstruction),
             ("test_parseXMLString", test_parseXMLString),
-            ("test_prefixes", test_prefixes)
+            ("test_prefixes", test_prefixes),
+            ("test_validation", test_validation)
         ]
     }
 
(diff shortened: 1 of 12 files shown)


TestFoundation/TestNSXMLDocument.swift
```diff
@@ -22,20 +22,47 @@ import CoreFoundation
...
+                ("test_processingInstruction", test_processingInstruction),
+                ("test_parseXMLString", test_parseXMLString),
+                ("test_prefixes", test_prefixes),
+                ("test_validation_success", test_validation_success),
+                ("test_validation_failure", test_validation_failure),
+                ("test_dtd", test_dtd),
+                ("test_documentWithDTD", test_documentWithDTD),
+                ("test_dtd_attributes", test_dtd_attributes)
+            ]
+        #else // On Linux, currently the tests that rely on NSError are segfaulting in swift_dynamicCast
```
phausler: is there a bugs.swift.org ticket for this?

rothomp3: There is now! https://bugs.swift.org/browse/SR-585
Possibly related to https://bugs.swift.org/browse/SR-429 also.

phausler:
Per the question of the internal class name, it is not a goal to match the internal/private APIs. Some times it may happen out of just inevitability of reasonable names or muscle memory/habit when re-writing it. As long as we have the functionality and the methods are named reasonably then they don't have to match.

That being said if there is a return value that claims lets say NSXMLElement and internally we return a NSSomeMadeUpXMLElement the only requirement is that the return value be a compatible class to a NSXMLElement.

PR #237, opened 2016-01-21
NSNotificationQueue: added the idle and asap posting style support

sgl0v:


Foundation/NSNotificationQueue.swift
```diff
@@ -31,6 +33,16 @@ public class NSNotificationQueue : NSObject {
     internal let notificationCenter: NSNotificationCenter
     internal var asapList = NSNotificationList()
     internal var idleList = NSNotificationList()
+    internal lazy var idleRunloopObserver: CFRunLoopObserverRef = {
+        return CFRunLoopObserverCreateWithHandler(kCFAllocatorDefault, CUnsignedLong(1) << 5, true, 0) {[weak self] observer, activity in
```
phausler: the `1 << 5` is not preferred; use kCFRunLoopBeforeWaiting instead and for mac export the constants similarly to the other enum work-arounds

Foundation/NSNotificationQueue.swift
```diff
@@ -31,6 +33,16 @@ public class NSNotificationQueue : NSObject {
...
+            self!.notifyQueues(.PostWhenIdle)
+        }
+    }()
+    internal lazy var asapRunloopObserver: CFRunLoopObserverRef = {
+        return CFRunLoopObserverCreateWithHandler(kCFAllocatorDefault, CUnsignedLong(1) << 1 | CUnsignedLong(1) << 7, true, 0) {[weak self] observer, activity in
```
phausler: same here use `kCFRunLoopBeforeTimers | kCFRunLoopExit`

Foundation/NSNotificationQueue.swift
```diff
@@ -53,10 +65,18 @@ public class NSNotificationQueue : NSObject {
         self.notificationCenter = notificationCenter
         super.init()
         NSNotificationQueue.registerQueue(self)
+        CFRunLoopAddObserver(NSRunLoop.currentRunLoop()._cfRunLoop, self.idleRunloopObserver, kCFRunLoopDefaultMode)
```
phausler: couldn't these be added lazily when a notification is enqueued?

Foundation/NSNotificationQueue.swift
phausler: I am not certain if we want to mimic this behavior but the weak self is "incorrect" in comparison to the darwin version... however this seems to be much better behavior.

We should run valgrind to make certain the block here is correct since the layout is a bit in dispute for blocks

Foundation/NSNotificationQueue.swift
sgl0v: It used to be like that. For some reason I got the next compile time error on Linux:

```
Foundation/NSNotificationQueue.swift:37:72: error: type 'CFRunLoopActivity' (aka 'UInt') has no member 'BeforeWaiting'
        return CFRunLoopObserverCreateWithHandler(kCFAllocatorDefault, CFRunLoopActivity.BeforeWaiting.rawValue, true, 0) {[weak self] observer, activity in
```

phausler: Thats because swift does not import the CF enums like it does on darwin. See https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSObjCRuntime.swift @ line 13 for an example of how to make it compatible.

sgl0v: Thx!

Foundation/NSNotificationQueue.swift
```diff
@@ -31,6 +33,16 @@ public class NSNotificationQueue : NSObject {
     internal let notificationCenter: NSNotificationCenter
     internal var asapList = NSNotificationList()
     internal var idleList = NSNotificationList()
+    internal lazy var idleRunloopObserver: CFRunLoopObserverRef = {
+        return CFRunLoopObserverCreateWithHandler(kCFAllocatorDefault, kCFRunLoopBeforeTimers, true, 0) {[weak self] observer, activity in
```
phausler: For linux this has to be `CFOptionFlags(kCFRunLoopBeforeTimers)`
```
Foundation/NSNotificationQueue.swift:37:72: error: cannot convert value of type 'Int' to expected argument type 'CFOptionFlags' (aka 'UInt')
        return CFRunLoopObserverCreateWithHandler(kCFAllocatorDefault, kCFRunLoopBeforeTimers, true, 0) {[weak self] observer, activity in
```

Foundation/NSNotificationQueue.swift
```diff
@@ -31,6 +33,16 @@ public class NSNotificationQueue : NSObject {
...
+            self!.notifyQueues(.PostWhenIdle)
+        }
+    }()
+    internal lazy var asapRunloopObserver: CFRunLoopObserverRef = {
+        return CFRunLoopObserverCreateWithHandler(kCFAllocatorDefault, kCFRunLoopBeforeWaiting | kCFRunLoopExit, true, 0) {[weak self] observer, activity in
```
phausler:
```
Foundation/NSNotificationQueue.swift:42:72: error: cannot convert value of type 'Int' to expected argument type 'UInt'
        return CFRunLoopObserverCreateWithHandler(kCFAllocatorDefault, kCFRunLoopBeforeWaiting | kCFRunLoopExit, true, 0) {[weak self] observer, activity in
```

PR #238, opened 2016-01-22
Add test for the Foundation version of hasPrefix and hasSuffix

glessard:

PR #239, opened 2016-01-23
Unicode-compatible (mostly) hasPrefix and hasSuffix

glessard:
The version of hasPrefix and hasSuffix originally found in Foundation for Linux does not do well with Unicode strings. The implementations is this PR reproduce those supplied in the standard library in the presence of the Objective-C runtime, namely swift_stdlib_NSStringHasPrefixNFD and swift_stdlib_NSStringHasSuffixNFD. The test results are largely the same, that is the issues appear to be the same (except the issues related to embedded nulls.)

Closing https://github.com/apple/swift-corelibs-foundation/pull/238, as it is superseded by this.

### DIFF
--- a/Foundation/String.swift
+++ b/Foundation/String.swift
@@ -1579,43 +1579,26 @@ extension String {
             ) != nil
         return r
     }
-    
-#if os(Linux)
+}
+
+#if !_runtime(_ObjC)
+import CoreFoundation
+
+extension String {
     public func hasPrefix(prefix: String) -> Bool {
-        let characters = utf16
-        let prefixCharacters = prefix.utf16
-        let start = characters.startIndex
-        let prefixStart = prefixCharacters.startIndex
-        if characters.count < prefixCharacters.count {
-            return false
-        }
-        for idx in 0..<prefixCharacters.count {
-            if characters[start.advancedBy(idx)] != prefixCharacters[prefixStart.advancedBy(idx)] {
-                return false
-            }
-        }
-        return true
+        let cfstring = self._cfObject
+        let range = CFRangeMake(0, CFStringGetLength(cfstring))
+        let opts = CFStringCompareFlags(kCFCompareAnchored + kCFCompareNonliteral)
+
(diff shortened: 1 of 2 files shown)


Foundation/String.swift
```diff
@@ -1579,43 +1579,26 @@ extension String {
...
-        }
-        for idx in 0..<prefixCharacters.count {
-            if characters[start.advancedBy(idx)] != prefixCharacters[prefixStart.advancedBy(idx)] {
-                return false
-            }
-        }
-        return true
+        let cfstring = self._cfObject
+        let range = CFRangeMake(0, CFStringGetLength(cfstring))
+        let opts = CFStringCompareFlags(kCFCompareAnchored + kCFCompareNonliteral)
```
phausler: +? shouldn't that be an |?
This might be cleaner just to funnel to the NS version of 
rangeOfString(str, options: .AnchoredSearch, range: NSMakeRange(0, length)).location != NSNotFound

glessard: Yes, a bitwise-or is more logical.

glessard:
@phausler 
An NS version wouldn't exhibit the same behaviour as the CF version. Either the behaviour of the stdlib version should be duplicated (as this PR does), or the behaviour should be as correct as possible.
https://github.com/apple/swift/pull/1031 has a ICU-based solution that behaves better, but also behaves better than the `_runtime(_ObjC)` version.

phausler:
The NS version does nearly exactly what you are doing:
public func hasPrefix(str: String) -> Bool {
        return rangeOfString(str, options: .AnchoredSearch, range: NSMakeRange(0, length)).location != NSNotFound
    }
which rangeOfString calls
CFStringFindWithOptionsAndLocale(_cfObject, searchString._cfObject, CFRange(searchRange), mask._cfValue(true), nil, ranges)

glessard:
I thought I remembered the test results being different when I tried it, but maybe I just decided in favor of the lower-level call. In any event, this version is very close to the one found at https://github.com/apple/swift/blob/master/stdlib/public/stubs/SwiftNativeNSXXXBase.mm.gyb#L131

PR #240, opened 2016-01-24
Made some upgrades to NSOperatingSystemVersion

rsmoz:
Added tests, too.

### DIFF
--- a/Foundation/NSProcessInfo.swift
+++ b/Foundation/NSProcessInfo.swift
@@ -20,6 +20,7 @@ public struct NSOperatingSystemVersion {
     public var majorVersion: Int
     public var minorVersion: Int
     public var patchVersion: Int
+	
     public init() {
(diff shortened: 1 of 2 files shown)


phausler:
This is an API change; any such change should be submitted as a API proposal and undergo review since it would have to be back-ported to the Darwin version of Foundation.

rsmoz:
@phausler Would the proposal go through Swift-evolution, or is that just for the actual language/compiler itself?

phausler:
Yes; swift-evolution is not just for compiler/language but also API changes and how they are exposed in swift. There is a bit more involved with that process since we have to take it to the component owners and have it be a maintained change in both the linux version as well as Mac OS X.

parkera:
I think we should probably do this, but we need to track it as an ER in JIRA first. Closing for now.

PR #242, opened 2016-01-28
Implement NSTemporaryDirectory

drewcrawford:


krzyzanowskim:
Temporary directory path may be unique for the user and may differ between operating systems. I'm not sure if it should be hardcoded as "/tmp".

nubbel:
How about [$TMPDIR](https://en.wikipedia.org/wiki/TMPDIR)?

drewcrawford:
What does it mean to say that it "may" be per-user?

I know that the Darwin Foundation documentation allows for this to be the case.  However FHS 3.17 and IEEE P1003.2 do not; they specify only one temporary directory (well also `/var/tmp` in the FHS case, but that is unhelpful here).

Darwin's documentation is not sufficiently detailed that we could match the implementation (without someone digging out sourcecode and throwing us a specification).  I'm not sure that is what is desired either, since on a standards-conforming Linux/Unix system, `/tmp/` is the appropriate directory for all temporary files.

krzyzanowskim:
@drewcrawford documentation for `NSTemporaryDirectory` says: 

> A string containing the path of the temporary directory for the current user. If no such directory is currently available, returns nil.

Swift is not targeted to Darwin solely, as such, in my opinion this value should take environment into consideration when calculate the value.

In particular, `NSTemporaryDirectory()` on OS X is not "/tmp"

drewcrawford:
I'm not opposed to taking the environment variable into account (it is also a standard) but then we are back to what to do when the environment variable is not set. It is not set in the situation that motivated this patch, and I can confirm that Darwin Foundation still returns a value in that case.

phausler:
Darwin should probably check against `confstr` using `_CS_DARWIN_USER_TEMP_DIR`
Then in all platforms check the environment variable TMPDIR
And then fall back to /tmp

That would approximate the behavior of the Darwin version

nubbel:
You might also want to take a look at the GNUStep implementation: https://github.com/timburks/gnustep-base/blob/master/Source/NSPathUtilities.m#L1908

phausler:
@nubbel no need, that implementation looks right on par. 
@swift-ci Please test

krzyzanowskim:
so the docs say:
> If no such directory is currently available, returns nil.

can't check ObjC foundation code... just wonder when it should return `nil` - and if return value should/shouldn't be Optional then.

@phausler Difference to GNUStep implementation is that it return `nil` if temporary directory does not exists - and that if it can't be created so... it cares about creating new temporary directory if missing or is not writable.

phausler:
The case of nil is a massive failure internally so my guess is that other things would start to fail if that was a nil return anyhow. Usually in these cases we claim the minuscule edge case failure as making this a non nil return value.

Darwin does not create the directory if it does not exist; we should maintain that behavior. The processes running with Foundation would rarely ever have the privs to actually create /tmp in the first place and we should not be requiring root to call that API.

phausler:
hmm swift-ci didn't run any tests for this... manual testing shows this works fine

krzyzanowskim:
@phausler understand. In that case documentation for NSTemporaryDirectory() need update in a way that it not suggest that returned value is usable. From now on developer should check if returned paths is usable for the application. 

It's tricky because (I think) for Objective-C it still may return nil value for unusable directory.
https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Functions/#//apple_ref/c/func/NSTemporaryDirectory

There is related radar for this: https://github.com/lionheart/openradar-mirror/issues/7356

m1entus:
Going deeper into Foundation framework, you can see that if `NSProcessInfo.processInfo().environment["TMPDIR"]` will fail, then `_NSTempDirectory` method is called which always return `tmp` and after that called is: `[[NSFileManager defaultManager] stringWithFileSystemRepresentation:[@"/tmp/" stringByStandardizingPath] length:strlen("/tmp/")]`

So the implementation should be like this
` let tmpdir = "/tmp/"
    
    return NSFileManager.defaultManager().stringWithFileSystemRepresentation(tmpdir.stringByStandardizingPath, length: Int(strlen(tmpdir)))` but the problem is that this may fail `(lldb) po [[NSFileManager defaultManager] stringWithFileSystemRepresentation:[@"/tmp" stringByStandardizingPath] length:5]` and -> String? should be returned from `public func stringWithFileSystemRepresentation(str: UnsafePointer<Int8>, length len: Int) -> String` instead of String.

PR #243, opened 2016-01-28
Implement setAttributesOfItemAtPath for NSFilePosixPermissions attribute

drewcrawford:
Add test coverage

### DIFF
--- a/TestFoundation/TestNSFileManager.swift
+++ b/TestFoundation/TestNSFileManager.swift
@@ -92,7 +92,7 @@ class TestNSFileManger : XCTestCase {
     
     func test_fileAttributes() {
         let fm = NSFileManager.defaultManager()
-        let path = "/tmp/testfile"
+        let path = "/tmp/test_fileAttributes\(NSUUID().UUIDString)"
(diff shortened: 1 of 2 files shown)


TestFoundation/TestNSFileManager.swift
```diff
@@ -137,6 +138,26 @@ class TestNSFileManger : XCTestCase {
         }
     }
     
+    func test_setFileAttributes() {
+        let path = "/tmp/testfile"
```
phausler: we might not want to use this since it would potentially conflict with other tests running on CI

drewcrawford: I stole this from the `get` test that was already there.

Can you recommend another design, and I can fix both of them?

phausler: using some sort of unique identifier per-run would be nice to have as a common accessor pattern. perhaps `/tmp/<some name identifying the test><uuid>` might be good as a pattern

phausler:
@swift-ci Please test

PR #245, opened 2016-01-31
Fix key's typo to NSFileGroupOwnerAccountName

RyanBreaker:
The key NSFileGroupOwnerAccountId is a typo as it is later overwritten by an NSNumber on line 381. The correct constant is NSFileGroupOwnerAccountName to correctly add the String of the account's name to the dictionary.

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -349,7 +349,7 @@ public class NSFileManager : NSObject {
         let grd = getgrgid(s.st_gid)
         if grd != nil && grd.memory.gr_name != nil {
             if let name = String.fromCString(grd.memory.gr_name) {
-                result[NSFileGroupOwnerAccountID] = name
+                result[NSFileGroupOwnerAccountName] = name
             }
         }
 


parkera:
Thanks!

parkera:
Swift CI detected a failure here, which doesn't look related to this change but I've rolled back just in case while we investigate.

RyanBreaker:
The failure is with NSNotificationQueue, which admittedly sparks my own curiosity:

`["TestFoundation/TestNSNotificationQueue.swift:127: error: TestNSNotificationQueue.test_notificationQueueLifecycle : XCTAssertNil failed: \"\" - "]

["Test Case \'TestNSNotificationQueue.test_notificationQueueLifecycle\' failed (0.0 seconds)."]
["Executed 6 tests, with 1 failure (0 unexpected) in 0.011 (0.012) seconds"`

parkera:
@phausler thinks this may be a race condition in the test and not related to your change (which I believe).

RyanBreaker:
Oh interesting, would this potentially be considered a bug then? I'm just curious as to what that means exactly.

phausler:
The bug is likely to either be exposed by NSThread or it is a bug in CFRunLoop: the thread issue has occurred before but I had thought I fixed it and the run loop backing is quite new and may have a corner case that is not addressed correctly yet. If it is NSThread there is a potential that the mistake is a incorrect usage or failure of withUnsafeMutablePointer. So I doubt this change had anything to do with it besides timing and/or a nondeterministic failure.

RyanBreaker:
I understand it's likely irrelevant to my change, I just can't help but be curious as to what that all means. :)

RyanBreaker:
Is it probably worth trying this pull again?

parkera:
@swift-ci please test

parkera:
Yah let's give it a shot.

PR #248, opened 2016-02-01
Implement NSFileManager.copyItemAtPath

drewcrawford:
Test coverage.

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -417,7 +417,38 @@ public class NSFileManager : NSObject {
     }
     
     public func copyItemAtPath(srcPath: String, toPath dstPath: String) throws {
-        NSUnimplemented()
+        let fd_from = open(srcPath, O_RDONLY)
+        if fd_from < 0 {
+            throw _NSErrorWithErrno(errno, reading: true, path: dstPath, url: nil, extraUserInfo: nil)
+        }
+        defer { precondition(close(fd_from) >= 0) }
+        let permission = (try! attributesOfItemAtPath(srcPath)[NSFilePosixPermissions] as! NSNumber).unsignedShortValue
+        let fd_to = open(dstPath, O_WRONLY | O_CREAT | O_EXCL, permission)
+        if fd_to < 0 {
+            throw _NSErrorWithErrno(errno, reading: false, path: dstPath, url: nil, extraUserInfo: nil)
+        }
+        defer { precondition(close(fd_to) >= 0) }
+        
+        var buf = [UInt8](count: 4096, repeatedValue: 0)
+        
+        while true {
+            let nread = read(fd_from, &buf, buf.count)
+            if nread < 0 { throw _NSErrorWithErrno(errno, reading: true, path: srcPath, url: nil, extraUserInfo: nil) }
+            if nread == 0 { break }
+            var writeSlice = buf[0..<nread]
+            
+            while true {
+                var nwritten: Int! = nil
+                writeSlice.withUnsafeBufferPointer({ (ptr) -> () in
+                    nwritten = write(fd_to, ptr.baseAddress, ptr.count)
+                })
+                if nwritten < 0 {
+                    throw _NSErrorWithErrno(errno, reading: false, path: dstPath, url: nil, extraUserInfo: nil)
+                }
+                writeSlice = writeSlice[writeSlice.startIndex.advancedBy(nwritten)..<writeSlice.endIndex]
+                if writeSlice.count == 0 { break }
+            }
+        }
     }
     
     public func moveItemAtPath(srcPath: String, toPath dstPath: String) throws {
--- a/TestFoundation/TestNSFileManager.swift
+++ b/TestFoundation/TestNSFileManager.swift
@@ -26,7 +26,8 @@ class TestNSFileManger : XCTestCase {
             ("test_directoryEnumerator", test_directoryEnumerator),
             ("test_pathEnumerator",test_pathEnumerator),
             ("test_contentsOfDirectoryAtPath", test_contentsOfDirectoryAtPath),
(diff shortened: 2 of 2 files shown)


Foundation/NSFileManager.swift
```diff
@@ -417,7 +417,38 @@ public class NSFileManager : NSObject {
...
     
     public func copyItemAtPath(srcPath: String, toPath dstPath: String) throws {
-        NSUnimplemented()
+        let fd_from = open(srcPath, O_RDONLY)
+        if fd_from < 0 {
+            throw _NSErrorWithErrno(errno, reading: true, path: dstPath, url: nil, extraUserInfo: nil)
+        }
+        defer { precondition(close(fd_from) >= 0) }
+        let permission = (try! attributesOfItemAtPath(srcPath)[NSFilePosixPermissions] as! NSNumber).unsignedShortValue
+        let fd_to = open(dstPath, O_WRONLY | O_CREAT | O_EXCL, permission)
```
phausler: what about directories?

Foundation/NSFileManager.swift
```diff
@@ -417,7 +417,38 @@ public class NSFileManager : NSObject {
...
+        if fd_to < 0 {
+            throw _NSErrorWithErrno(errno, reading: false, path: dstPath, url: nil, extraUserInfo: nil)
+        }
+        defer { precondition(close(fd_to) >= 0) }
+        
+        var buf = [UInt8](count: 4096, repeatedValue: 0)
+        
+        while true {
```
phausler: perhaps a `stat` or `fstat` might avoid the need for a potentially infinite loop?

Foundation/NSFileManager.swift
```diff
@@ -417,7 +417,38 @@ public class NSFileManager : NSObject {
...
+            let nread = read(fd_from, &buf, buf.count)
```
phausler: Darwin and BSDs have the API `copyfile` and linux has the similar API called `sendfile`

Foundation/NSFileManager.swift
```diff
@@ -417,7 +417,38 @@ public class NSFileManager : NSObject {
...
+            if nread < 0 { throw _NSErrorWithErrno(errno, reading: true, path: srcPath, url: nil, extraUserInfo: nil) }
```
phausler: what about EAGAIN?

Foundation/NSFileManager.swift
```diff
@@ -417,7 +417,38 @@ public class NSFileManager : NSObject {
     }
     
     public func copyItemAtPath(srcPath: String, toPath dstPath: String) throws {
-        NSUnimplemented()
+        let fd_from = open(srcPath, O_RDONLY)
+        if fd_from < 0 {
+            throw _NSErrorWithErrno(errno, reading: true, path: dstPath, url: nil, extraUserInfo: nil)
+        }
+        defer { precondition(close(fd_from) >= 0) }
```
phausler: shouldn't this be a throwing case?

drewcrawford: Defer can't throw.

I could rewrite without defer, but it will be more brittle.

Foundation/NSFileManager.swift
drewcrawford: On linux, sendfile only works for sockets, not files.

Foundation/NSFileManager.swift
drewcrawford: We could do an fstat on each file and compare their `off_t`.  The problem is, this introduces 2 new syscalls into a hot loop, and we pay that tax every 4096 bytes copied.

We already find out the bytes read and written from `read` and `write` respectively, so this also introduces a second source of truth.

Foundation/NSFileManager.swift
drewcrawford: I don't believe EAGAIN can be raised here.  The Darwin manpage has

>  [EAGAIN] The file was marked for non-blocking I/O, and no data were ready to be read.

In this case the file was not marked for non-blocking I/O, and the descriptor does not leak anywhere that someone might set it.

Linux manpage is similar.

Foundation/NSFileManager.swift
drewcrawford: I guess this changed recently; sendfile works for files in Linux kernels above 2.6.33.  [source](http://linux.die.net/man/2/sendfile)

In that case we should probably use sendfile/copyfile, as long as we're satisfied to take a dependency on that kernel or later.  Can you confirm we can do that?

Foundation/NSFileManager.swift
phausler: I was suggesting that the stat could be done before the loop to determine the file size ahead of time to know how much would need to be read.

Foundation/NSFileManager.swift
phausler: well `copyfile` we can definitely bank on for Darwin, for official support we support Ubuntu 14 & 15, so I believe `sendfile` is available on both of those versions (14 is Linux Kernel 3.13 so we should be in the clear here)

drewcrawford: Using `sendfile` on Linux results in error

```
error: use of unresolved identifier 'sendfile'
```

I gather that this doesn't come free with Glibc?

phausler: It should be; perhaps the import isn't in the Glibc overlay for linux for #include <sys/sendfile.h>
It could be re-exposed like _CFOpenFileWithMode has to be re-exported to swift.

it would be totally reasonable to make a platform abstraction for copyfile in CFPlatform.c since Darwin is different than linux and if ever there is a Windows port there is a different function there too.

drewcrawford:
Thanks for your help, but I think this has crossed the threshold of "I don't want it badly enough to take it further."  If that changes I'll PR again.

PR #249, opened 2016-02-02
Companion PR to Swift #1157 and 32-bit bug fixes

hpux735:
This PR is a necessary companion to https://github.com/apple/swift/pull/1157 .  swift.ld is used in the build of Foundation, and with the deletion of the script from Swift, Foundation fails to build.  The first commit uses the conformance begin and end objects generated in Swift for manual share library linking.  Also, this commit allows the Swift build scripts to indicate to Foundation whether it should use the gold linker.

**It's very important that this PR only be merged at the same time as the Swift#1157**

~~The second commit addresses some new regressions introduced in ```NSXML*``` on 32-bit systems.  The ```NSXMLNodeOptionRawType``` type was introduced for working with ```NSXMLNodeOption```s in a bit-width independent way.~~

### DIFF
--- a/build.py
+++ b/build.py
@@ -15,8 +15,7 @@ foundation.GCC_PREFIX_HEADER = 'CoreFoundation/Base.subproj/CoreFoundation_Prefi
 
 if Configuration.current.target.sdk == OSType.Linux:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE '
-	foundation.LDFLAGS = '-Wl,@./CoreFoundation/linux.ld -Xlinker -T ${SDKROOT}/lib/swift/linux/${ARCH}/swift.ld -lswiftGlibc `icu-config --ldflags` -Wl,-defsym,__CFConstantStringClassReference=_TMC10Foundation19_NSCFConstantString -Wl,-Bsymbolic '
-
+	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc `icu-config --ldflags` -Wl,-defsym,__CFConstantStringClassReference=_TMC10Foundation19_NSCFConstantString -Wl,-Bsymbolic '
 elif Configuration.current.target.sdk == OSType.FreeBSD:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_FREEBSD -I/usr/local/include -I/usr/local/include/libxml2 '
 	foundation.LDFLAGS = ''
--- a/lib/product.py
+++ b/lib/product.py
@@ -71,6 +71,8 @@ class Product(BuildAction):
         return Path.path(Configuration.current.build_directory.path_by_appending(self.name).absolute() + "/" + self.PROJECT_HEADERS_FOLDER_PATH)
 
 class Library(Product):
+    conformance_begin = ""
+    conformance_end = ""
     rule = None
     def __init__(self, name):
         Product.__init__(self, name)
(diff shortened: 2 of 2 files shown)


Foundation/NSXMLDTDNode.swift
```diff
@@ -62,7 +62,7 @@ public class NSXMLDTDNode : NSXMLNode {
         }
     } //primitive
     
-    public override init(kind: NSXMLNodeKind, options: Int) {
+    public override init(kind: NSXMLNodeKind, options: NSXMLNodeOptionRawType) {
```
phausler: This seems like an API change. It is declared in both 32 bit and 64 bit Darwin targets as `public convenience init(kind: NSXMLNodeKind, options: Int)` why can't it be that here too?

hpux735: To avoid changing the API, I can cast it to that type within the function if that's what you prefer.  We've hit this problem before in https://github.com/apple/swift-corelibs-foundation/pull/169#issuecomment-166961604   Ultimately, it seems like this stuff boils down to a CF type that resolves to UInt32.  I kinda don't understand why it's promoted to a 64-bit int in Foundation.

phausler: I think because NSUInteger and NSInteger were preferred for APIs for future proofing the storage size for parameters.

hpux735: That makes sense.  In that case, the type seems necessary. The reason being is that ```Int``` on 32-bit platforms is (as expected) a signed 32-bit integer.  In [NSXMLNodeOptions:105](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSXMLNodeOptions.swift#L105) the upper bits are set to '1' (including bit 31) which is an overflow in that case.  

Because ```NSXMLNodeOptionRawType``` is a typealias to ```Int``` on 64-bit machines and ```Int64``` on 32-bit machines Swift doesn't require casts (to my knowledge) on 64-bit.  On 32-bit machines, it guarantees that ```NSXMLNodePreserveAll``` won't overflow (among other issues).

hpux735: I went ahead and took this stuff out.  It's probably not appropriate to mix these into the same PR.  I'll make another one for the XML issues.

phausler:
@swift-ci Please test

phausler:
oh never mind apple/swift#1157 is required to build; else we get 
```
ninja: error: '/home/phausler/Public/build/Ninja-ReleaseAssert/swift-linux-x86_64/lib/swift/linux/x86_64/swift_begin.o', needed by '../build/Ninja-ReleaseAssert/foundation-linux-x86_64/Foundation/libFoundation.so', missing and no known rule to make it
```

hpux735:
Yep that's right.  We'll need to wait a bit. :)  Here's the new PR that handles XML https://github.com/apple/swift-corelibs-foundation/pull/252

phausler:
@swift-ci Please test

gribozavr:
@phausler It looks like we are going to merge apple/swift#1157, would you mind us merging this PR as well, assuming all tests pass on Linux/x86_64?

parkera:
I'm a little confused about why the XML change is required for this.

hpux735:
@parkera Oops!  It wasn't supposed to be there.  The XML stuff was required for Foundation to build on ARM, so I pulled it in prior to Philippe merging it into master.  I've removed it.

parkera:
Ok, this makes more sense to me now. @gribozavr , if you test this and it works you can merge it in.

gribozavr:
@parkera Thanks!  We're waiting for a companion PR for a similar build system change in xctest, but once that is ready, I'll merge all three PRs together.

PR #250, opened 2016-02-03
Add support for NSNumber to return a PlaygroundQuickLook

egranata:
This commit adds support for NSNumber to return a PlaygroundQuickLook object

On the OS X Foundation, this works due to a combination of the Swift runtime and the ObjC runtime cooperating (see Reflection.mm in the Swift runtime)

For corelibs Foundation there is no such support. This patch adds it.

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -465,13 +465,13 @@ extension NSNumber : CustomPlaygroundQuickLookable {
     public func customPlaygroundQuickLook() -> PlaygroundQuickLook {
         let type = CFNumberGetType(_cfObject)
         switch type {
-        case kCFNumberSInt8Type:
+        case .SInt8Type:
             fallthrough
-        case kCFNumberSInt16Type:
+        case .SInt16Type:
             fallthrough
-        case kCFNumberSInt32Type:
+        case .SInt32Type:
             fallthrough
-        case kCFNumberSInt64Type:
+        case .SInt64Type:
             fallthrough
         case kCFNumberCharType:
             fallthrough
@@ -481,21 +481,21 @@ extension NSNumber : CustomPlaygroundQuickLookable {
             fallthrough
         case kCFNumberLongType:
             fallthrough
-        case kCFNumberCFIndexType:
+        case .CFIndexType:
             fallthrough
-        case kCFNumberNSIntegerType:
+        case .NSIntegerType:
             fallthrough
-        case kCFNumberLongLongType:
+        case .LongLongType:
             return .Int(self.longLongValue)
-        case kCFNumberFloat32Type:
+        case .Float32Type:
             fallthrough
         case kCFNumberFloatType:
             return .Float(self.floatValue)
-        case kCFNumberFloat64Type:
+        case .Float64Type:
             fallthrough
         case kCFNumberDoubleType:
             return .Double(self.doubleValue)
-        case kCFNumberCGFloatType:
+        case .CGFloatType:
             if sizeof(CGFloat) == sizeof(Float32) {
                 return .Float(self.floatValue)
             } else {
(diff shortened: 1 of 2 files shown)


phausler:
Does this even do anything sensible in linux? 

On Darwin this has an error emitted:
```swift
func isPlaygroundQuickLook(ql: PlaygroundQuickLook) {
    
}

let num = NSNumber(float: 5.5)
isPlaygroundQuickLook(num)
```

So this would be new behavior?

egranata:
The error you get is because an NSNumber is not a PlaygroundQuickLook, it's a CustomPlaygroundQuicklookable

Also, yes, in my - admittedly limited - testing, this does something sensible on Linux

egranata:
Actually, before testing this, let me push one more commit. I found out that on OS X, CFNumberType gets imported as a proper Swift enum, but on Linux it only gets imported as Int. I need to rework my changes to work with that discrepancy. I'll give a heads up in this pull request when I am ready for getting it through CI testing

egranata:
Alright - this should be good to go. Feel free to get swift-ci testing.

phausler:
On Darwin:
```
main.swift:16:23: error: argument type 'NSNumber' does not conform to expected type 'CustomPlaygroundQuickLookable'
isPlaygroundQuickLook(num)
```

egranata:
Do you mean "on Darwin with Corelibs-Foundation" or "on Darwin with stock Foundation"?

If the former, that's weird, and I would need to investigate more.
If the latter, that is correct. On OS X, there is some magic code in the Swift runtime that generates a QuickLook for NSNumber without the protocol conformance involved. This piece of magic relies on the ObjC runtime, so it can't work on Linux, which means we actually need to explicitly implement CustomPlaygroundQuickLookable there. You might make the argument that we should do the same thing on Darwin via an SDK overlay for stock Foundation, and I am probably going to do that as well at some point, but that is a separate change from this diff.

phausler:
on Darwin with the objective-c foundation that ships with the operating system, changes like this would need to not only be applied here but also applied there; which require proposals to be passed along to the appropriate teams that own that portion.

So the change would need to be both here and in https://github.com/apple/swift/tree/master/stdlib/public/SDK/Foundation

I will follow up with what needs to be done here. 

My take on it is the concept makes sense to add more rich information to NSNumber but I am not the only one involved with that API.

gribozavr:
@phausler
 
> isPlaygroundQuickLook(num)

Sorry, but that's not the right way to use this conformance.  Please use:

```
PlaygroundQuickLook(reflecting: num)
```

This should work on all platforms.

phausler:
@gribozavr that unfortunately hits the bridge cases transforming NSNumber to it's other types, no?

egranata:
@gribozavr Correct, your code example works on Darwin

     3> PlaygroundQuickLook(reflecting: num)
     $R0: PlaygroundQuickLook = Int {
       Int = 1234
     }

egranata:
@phausler I suspect it's actually hitting this snippet of code in swift_ClassMirror_quickLookObject

```
  // NSNumbers quick-look as integers or doubles, depending on type.
  if ([object isKindOfClass:[NSNumber class]]) {
    NSNumber *n = object;
    
    switch ([n objCType][0]) {
    case 'd': // double
      result.payload.Double = [n doubleValue];
      result.payload.Kind = PlaygroundQuickLook::Tag::Double;
      break;
    case 'f': // float
      result.payload.Float = [n floatValue];
      result.payload.Kind = PlaygroundQuickLook::Tag::Float;
      break;
        
    case 'Q': // unsigned long long
      result.payload.UInt = [n unsignedLongLongValue];
      result.payload.Kind = PlaygroundQuickLook::Tag::Uint;
      break;

    // FIXME: decimals?
    default:
      result.payload.Int = [n longLongValue];
      result.payload.Kind = PlaygroundQuickLook::Tag::Int;
      break;
    }
    
    [object release];
    result.optional.isNone = false;
    return result;
  }

```

gribozavr:
@phausler No, it does not use bridging.

PR #251, opened 2016-02-03
Implement NSUserDefaults

MaddTheSane:
This pull requests implements some methods for `NSUserDefaults`, wrapping around the `CFPreferences` API.

Also included are a few basic tests.

What isn't implemented:
* Notification Center support
* Domain names
* Suite support is untested

### DIFF
--- a/Foundation/NSUserDefaults.swift
+++ b/Foundation/NSUserDefaults.swift
@@ -150,7 +150,7 @@ public class NSUserDefaults : NSObject {
         guard let aVal = objectForKey(defaultName), bVal = aVal as? NSArray else {
             return nil
         }
-        return (bVal._swiftObject as? [NSString])?.map({ return $0._swiftObject})
+        return _expensivePropertyListConversion(bVal) as? [String]
     }
     public func integerForKey(defaultName: String) -> Int {
         guard let aVal = objectForKey(defaultName), bVal = aVal as? NSNumber else {
--- a/Foundation/NSUserDefaults.swift
+++ b/Foundation/NSUserDefaults.swift
@@ -246,6 +246,9 @@ public class NSUserDefaults : NSObject {
     }
     
     public func dictionaryRepresentation() -> [String : AnyObject] {
+        NSUnimplemented()
+        /*
+        Currently crashes the compiler.
         guard let aPref = CFPreferencesCopyMultiple(nil, kCFPreferencesCurrentApplication, kCFPreferencesCurrentUser, kCFPreferencesCurrentHost),
             bPref = (aPref._swiftObject) as? [NSString: AnyObject] else {
                 return registeredDefaults
@@ -257,6 +260,7 @@ public class NSUserDefaults : NSObject {
         }
         
         return allDefaults
+        */
     }
     
     public var volatileDomainNames: [String] { NSUnimplemented() }
--- a/Foundation/NSUserDefaults.swift
+++ b/Foundation/NSUserDefaults.swift
@@ -7,46 +7,286 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+import CoreFoundation
 
 public let NSGlobalDomain: String = "NSGlobalDomain"
 public let NSArgumentDomain: String = "NSArgumentDomain"
 public let NSRegistrationDomain: String = "NSRegistrationDomain"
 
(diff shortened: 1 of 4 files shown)


Catfish-Man:
I've left some initial review feedback in the commits; thanks a ton for working on this! More tests would be great, and before we merge it in it'd be great to have it squashed down to one commit to avoid cluttering the history.

MaddTheSane:
Thanks for the feedback.

I have compressed the commits into one.

Foundation/NSUserDefaults.swift
```diff
@@ -7,46 +7,286 @@
...
+            cfType = NSNumber(bool: bType)._cfObject
+        } else if let bType = value as? [NSObject: AnyObject] {
+            cfType = bType._cfObject
+        } else if let bType = value as? [String: AnyObject] {
+            cfType = bType.map({ (str, obj) -> (NSString, AnyObject) in
+                return (str._nsObject, obj)
+            })._cfObject
+        } else if let bType = value as? [AnyObject] {
+            cfType = bType._cfObject
+        } else if let bType = value as? [String] {
```
phausler: if value is an `AnyObject` it cannot be `[String]` or `[AnyObject]` it can only be NSArray etc

Foundation/NSUserDefaults.swift
```diff
@@ -7,46 +7,286 @@
...
+    }
+    public func setObject(value: AnyObject?, forKey defaultName: String) {
+        guard let value = value else {
+            CFPreferencesSetAppValue(defaultName._cfObject, nil, suite?._cfObject ?? kCFPreferencesCurrentApplication)
+            return
+        }
+        
+        var cfType: CFTypeRef? = nil
+		
+		//FIXME: is this needed? Am I overcomplicating things?
```
phausler: Yea this is probably not needed at all since all of these types are already bridged to CF

Foundation/NSUserDefaults.swift
```diff
@@ -7,46 +7,286 @@
...
+        }
+        return bVal
+    }
+    public func stringArrayForKey(defaultName: String) -> [String]? {
+        guard let aVal = objectForKey(defaultName), bVal = aVal as? NSArray else {
+            return nil
+        }
+        return (bVal._swiftObject as? [NSString])?.map({ return $0._swiftObject})
+    }
+    public func integerForKey(defaultName: String) -> Int {
```
Austinate: Wouldn't it be a better idea to return optional values in all smthngForKey functions?

phausler: that would be divergent from the current API

Foundation/NSUserDefaults.swift
```diff
@@ -7,46 +7,257 @@
...
+        guard let aVal = objectForKey(defaultName), bVal = aVal as? NSData else {
+            return nil
+        }
+        return bVal
+    }
+    public func stringArrayForKey(defaultName: String) -> [String]? {
+        guard let aVal = objectForKey(defaultName), bVal = aVal as? NSArray else {
+            return nil
+        }
+        return (bVal._swiftObject as? [NSString])?.map({ return $0._swiftObject})
```
phausler: The compiler crashes on this:

```
1.	While emitting SIL for 'stringArrayForKey' at Foundation/NSUserDefaults.swift:149:12
Segmentation fault (core dumped)
```

phausler:
@swift-ci Please test

phausler:
@swift-ci Please test

MaddTheSane:
Any idea on why Swift is crashing at `dictionaryRepresentation`? Specifically, which line of code Swift is failing on?

phausler:
sorry for not getting back quicker on this:

after resolving some other issues I have a feeling it is because of this

```
TestFoundation/TestNSUserDefaults.swift:35:38: error: value of type 'Int' does not conform to expected dictionary value type 'AnyObject'
                defaults.registerDefaults(["key1": 5])
                                                   ^
```

MaddTheSane:
Ah. Forgot that.

Fix incoming.

phausler:
why is github showing this as un-merged? I merged this PR just a bit ago...

525ed7b9fa6eee042e2d1218623fceacc8c31436 is on master

Am I missing any commits from this?

MaddTheSane:
Don't know why it's not showing up as applied, but this does look like it can safely be closed.

PR #252, opened 2016-02-03
Addressing 32-bit issues with NSXMLNodeOptions

hpux735:
This patch aims to address the issues with NSXMLNodeOptions on 32-bit platforms in a way that doesn't impact the API.  The current problem (that ```NSXMLNodeOptionRawType``` aimed to fix) is that the assignment of the sign bit of an Int is flagged by Swift as an overflow.  By assigning this bit with the acknowledgement that it is becoming negative quiets the compiler while maintaining the effect.

### DIFF
--- a/Foundation/NSXMLNodeOptions.swift
+++ b/Foundation/NSXMLNodeOptions.swift
@@ -43,55 +43,49 @@
     @constant NSXMLDocumentIncludeContentTypeDeclaration Include a content type declaration for HTML or XHTML
 */
 
-#if arch(x86_64) || arch(arm64)
-public typealias NSXMLNodeOptionRawType = Int
-#else
-public typealias NSXMLNodeOptionRawType = Int64
-#endif
-
-public var NSXMLNodeOptionsNone: NSXMLNodeOptionRawType { return 0 }
+public var NSXMLNodeOptionsNone: Int { return 0 }
 
 // Init
-public var NSXMLNodeIsCDATA: NSXMLNodeOptionRawType { return 1 << 0 }
-public var NSXMLNodeExpandEmptyElement: NSXMLNodeOptionRawType { return 1 << 1 } // <a></a>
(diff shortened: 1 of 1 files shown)


Foundation/NSXMLNodeOptions.swift
```diff
@@ -43,55 +43,55 @@
     @constant NSXMLDocumentIncludeContentTypeDeclaration Include a content type declaration for HTML or XHTML
 */
 
-#if arch(x86_64) || arch(arm64)
-public typealias NSXMLNodeOptionRawType = Int
+#if arch(x86_64) || arch(arm64) || arch(powerpc64le)
```
phausler: so we have to keep bumping this when we get other 64 bit platforms? mips etc? is there not a swift defined or 64 bit?

Foundation/NSXMLNodeOptions.swift
```diff
@@ -43,55 +43,55 @@
...
+let NSXMLNodeOptionMask: Int = 0xFFF00000
 #else
-public typealias NSXMLNodeOptionRawType = Int64
+let NSXMLNodeOptionMask: Int = -1048576
```
phausler: can we do this with a `Int(bitPattern: )`?

Foundation/NSXMLNodeOptions.swift
hpux735: Yes, I believe so.  There are no configuration statements listed for bit-width in the Swift Programming Language book.  I would love for that to exist.

Foundation/NSXMLNodeOptions.swift
hpux735: Seems to work well.  I didn't know about that one.

hpux735:
I dont know what happened to your comment, but yes.  The bitmask seems to have worked as expected on 32 & 64 bit machines.  That's much better!

phausler:
:+1: I like this change, builds cleanly and tests pass

hpux735:
Thanks the the help, Philippe! 😀

PR #255, opened 2016-02-05
Changes related to SE-0028

Austinate:
Replaces old identifiers with new ones

### DIFF
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -888,7 +888,7 @@ struct ComparisonTest {
 
     init(
         _ lhs: String, _ rhs: String,
-        reason: String = "", line: UInt = __LINE__
+          reason: String = "", line: UInt = #line
     ) {
         self.lhs = lhs
         self.rhs = rhs
@@ -1039,8 +1039,8 @@ func checkHasPrefixHasSuffix(lhs: String, _ rhs: String, _ stack: [UInt]) -> Int
     }
 
     var failures = 0
-    failures += testFailure(expectHasPrefix, lhs.hasPrefix(rhs), stack + [__LINE__])
-    failures += testFailure(expectHasSuffix, lhs.hasSuffix(rhs), stack + [__LINE__])
+    failures += testFailure(expectHasPrefix, lhs.hasPrefix(rhs), stack + [#line])
+    failures += testFailure(expectHasSuffix, lhs.hasSuffix(rhs), stack + [#line])
     return failures
 }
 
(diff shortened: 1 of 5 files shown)


Austinate:
One thing i'm not sure about __PRETTY_FUNCTION__ which actually is define for __FUNCTION__ but is not covered in [SE-0028](https://github.com/apple/swift-evolution/blob/master/proposals/0028-modernizing-debug-identifiers.md) 
Should we rename it too or left same for compatibility?

lattner:
The changes to the .c and .h files seem very wrong to me, they won't build.

Austinate:
My bad, will fix

Austinate:
@lattner reverter those changes, sorry for that ¯\_(ツ)_/¯

phausler:
what corresponding commit for the compiler is this attached to?

Austinate:
[This one](https://github.com/apple/swift/commit/0619e57a61f27f721e273ab6f808ac81011aeb2c)

phausler:
thanks

TestFoundation/TestNSString.swift
```diff
@@ -1039,8 +1039,8 @@ func checkHasPrefixHasSuffix(lhs: String, _ rhs: String, _ stack: [UInt]) -> Int
     }
 
     var failures = 0
-    failures += testFailure(expectHasPrefix, lhs.hasPrefix(rhs), stack + [__LINE__])
-    failures += testFailure(expectHasSuffix, lhs.hasSuffix(rhs), stack + [__LINE__])
+    failures += testFailure(expectHasPrefix, lhs.hasPrefix(rhs), stack + [#line])
```
ianpartridge: This doesn't compile for me.

ianpartridge: Seems `[#line]` doesn't compile but `[ #line] does.  Intentional as part of SE-0028?

Austinate: "This includes keeping `__FUNCTION__`, and making `#line` have the dual behavior of being a directive when it is the first token on a line, but an expression otherwise." [From SE-0028](https://github.com/apple/swift-evolution/blob/master/proposals/0028-modernizing-debug-identifiers.md)
So if `#line` is taken as expression it looks like it should work. Or i'm missing something.

ianpartridge: Hmm - @lattner ?

lattner: This was fixed in a follow on patch.  Just put a space between the [ and # if you don't have a build with the fix yet.

phausler:
@swift-ci Please test

TestFoundation/TestNSString.swift
ianpartridge: Thanks!

PR #256, opened 2016-02-05
Next step in connecting libdispatch and foundation builds

dgrove-oss:
If the foundation build script is invoked from build-script-impl
with libdispatch enabled then add includes and define to CFLAGS.

Add another case in which __HAS_DISPATCH__ is set to 1 that will
be enabled by the CFLAGS set in build.py (staging while work in
progress).

### DIFF
--- a/CoreFoundation/Base.subproj/CFBase.h
+++ b/CoreFoundation/Base.subproj/CFBase.h
@@ -61,6 +61,10 @@
 #define __HAS_DISPATCH__ 1
 #endif
 #endif
+#if DEPLOYMENT_TARGET_LINUX && DEPLOYMENT_RUNTIME_SWIFT && DEPLOYMENT_ENABLE_LIBDISPATCH
+#define __HAS_DISPATCH__ 1
+#endif
+
 // Some compilers provide the capability to test if certain features are available. This macro provides a compatibility path for other compilers.
 #ifndef __has_feature
 #define __has_feature(x) 0
--- a/build.py
+++ b/build.py
@@ -61,6 +61,14 @@ if "XCTEST_BUILD_DIR" in Configuration.current.variables:
 		'-L${XCTEST_BUILD_DIR}',
 		'-I/usr/include/libxml2'
 	]
+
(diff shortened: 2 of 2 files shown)


parkera:
I assume eventually this would just always be set, right?

dgrove-oss:
I think once we get over the initial hump of getting things working it could make sense to enable dispatch by default for foundation.  Dispatch does currently drag in a dependency on libkqueue that is easily satisfiable on ubuntu 14 and ubuntu 15 via apt-get.  This may not be true on other Linux distros (at first glance I didn't find a libkqueue available via yum for CentOS or RHEL).

phausler:
What are the corresponding swift repo commits for this?

phausler:
Also i presume this is using the blocks runtime too? perhaps we should snippet out that portion of compilation as well in the build.py

dgrove-oss:
The related pull requests are in swift https://github.com/apple/swift/pull/1212 and in libdispatch https://github.com/apple/swift-corelibs-libdispatch/pull/43.  It would be nice to apply them all at about the same time, but they are not really dependent on each other since building libdispatch is left disabled by default.  

Note that, CFRunLoop won't compile against libdispatch on linux yet because the APIs are still using mach_ types.  We could try to get it all working at once, but I thought a series of smaller commits that leave things disabled would be easier to merge.  Should I try to do in a single step?

dgrove-oss:
Based on mailing list discussion today, I made some minor adjustments on the libdispatch side and raised a new pull request there (https://github.com/apple/swift-corelibs-libdispatch/pull/49).  After that clears, I'll make any adjustments needed to this pull request.  So, please hold off on merging this for now.  Thanks!

dgrove-oss:
No adjustments needed; the -I paths in this pull request are still valid (not impacted by https://github.com/apple/swift-corelibs-libdispatch/pull/49).  This could be merged.  We're going to need more changes in both dispatch (@seabaylea has https://github.com/apple/swift-corelibs-libdispatch/pull/46 which is being discussed) and  also in foundation depending on how libdisaptch-46 pans out before the C code will actually compile.

PR #257, opened 2016-02-08
Fix compilation issues with HAS_DISPATCH on Linux

seabaylea:
As part of the prep work I'm doing with @dgrove-oss to enable Dispatch, I've done some testing of compiling swift-corelibs-foundation with `__HAS_DISPATCH__` enabled. That's highlighted a number of compile errors which this PR resolves.

Enabling `__HAS_DISPATCH__` will still result in linker errors, which we'll look to resolve once we have the Dispatch overlay built into lib dispatch on Linux ([libdispatch PR #43](https://github.com/apple/swift-corelibs-libdispatch/pull/43), and made the necessary modifications to the foundation build scripts.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -805,7 +805,7 @@ CF_EXPORT void _NS_pthread_setname_np(const char *name);
 #define pthread_setname_np _NS_pthread_setname_np
 #endif
 
-#if DEPLOYMENT_TARGET_WINDOWS
+#if DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_LINUX
 // replacement for DISPATCH_QUEUE_OVERCOMMIT until we get a bug fix in dispatch on Windows
 // <rdar://problem/7923891> dispatch on Windows: Need queue_private.h
 #define DISPATCH_QUEUE_OVERCOMMIT 2
@@ -866,7 +866,6 @@ CF_INLINE long qos_class_main() {
 CF_INLINE long qos_class_self() {
 	return QOS_CLASS_DEFAULT;
 }
-
 #endif
 
 // Returns a generic dispatch queue for when you want to just throw some work
--- a/CoreFoundation/RunLoop.subproj/CFRunLoop.c
+++ b/CoreFoundation/RunLoop.subproj/CFRunLoop.c
@@ -2541,7 +2541,7 @@ static int32_t __CFRunLoopRun(CFRunLoopRef rl, CFRunLoopModeRef rlm, CFTimeInter
     }
     
 #if __HAS_DISPATCH__
-    mach_port_name_t dispatchPort = MACH_PORT_NULL;
+    __CFPort dispatchPort = CFPORT_NULL;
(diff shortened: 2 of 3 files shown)


CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -850,9 +850,6 @@ CF_INLINE const char *CFPathRelativeToAppleFrameworksRoot(const char *path, Bool
 #if (DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_EMBEDDED_MINI || DEPLOYMENT_TARGET_LINUX)
 #if __HAS_DISPATCH__
 
-#include <dispatch/dispatch.h>
-#include <dispatch/private.h>
```
phausler: why don't we have /usr/local/include/dispatch/private.h? can we just pass across private header paths if it isn't installed into a staging sysroot?

and I am not certain that /usr/include/dispatch/dispatch.h missing is a good thing. If that is not actually missing then perhaps we should just move the private.h out only.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -175,6 +176,11 @@ const char *_CFProcessPath(void) {
     }
     return __CFProcessPath;
 }
+
+Boolean _CFIsMainThread(void) {
+    return syscall(SYS_gettid) == getpid();
```
phausler: where is this function used? we already have a shim definition of pthread_main_np

seabaylea: This is redundant. I'll remove.

CoreFoundation/Base.subproj/CFInternal.h
seabaylea: We don't have a dispatch/private.h on Linux so it needs to be removed (at least for now). There's also no direct need to reference dispatch/dispatch.h from CFInternal.h, but I'll, update to retain the reference.

dgrove-oss: There's a workaround for dispatch/private.h in the pending overlay pull requests.  Basically it exploits the symlink that is made by configure from $(top_bindir)/tests/dispatch to $(top_srcdir)/private to allow an include of dispatch/private.h to succeed.  Personally, I think this is a bit quirky, but it works (and is consistent with how the include of dispatch/private.h is made to work when compiling libdispatch's test cases on linux).

seabaylea:
Sorry, I realised I'd created the pull request from and old working branch in my repo. I've updated the PR with the correct changes.

The reference to `dispatch/private.h` in `CFInternal.h` has been removed as there isn't an equivalent on Linux. There may be a few issues we'll need to work through on linking though once @dgrove-oss gets his build changes in.

seabaylea:
Thanks @dgrove-oss . Given that the reference to private.h is still protected under `__HAS_DISPATCH__` (which isn't enabled yet) then I'll revert to including `dispatch/private.h` as it won't cause a breakage out of the box, and will be ready for when your pending overlay pull requests come in.

seabaylea:
@phausler Changes made and commits squashed in place. Could you take another look?

phausler:
looks good to me

seabaylea:
Thanks @phausler !

PR #259, opened 2016-02-12
Made NSLock use direct pthread strucutres instead of raw pointers

richardjrossiii:
This should reduce memory usage, increase performance, and makes code
signifcantly easier to read.

### DIFF
--- a/Foundation/NSLock.swift
+++ b/Foundation/NSLock.swift
@@ -23,28 +23,26 @@ public protocol NSLocking {
 }
 
 public class NSLock : NSObject, NSLocking {
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t>.alloc(1)
+    internal var mutex = pthread_mutex_t()
     
     public override init() {
-        pthread_mutex_init(mutex, nil)
+        pthread_mutex_init(&mutex, nil)
     }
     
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci please test

richardjrossiii:
Whoops. Left out a cast. Will fix.

richardjrossiii:
PR Updated

richardjrossiii:
@swift-ci Please test

phausler:
unfortunately taking the address of these structures is not thread safe in swift the last time I checked, so the only way to make a safe reference to a pthread_mutex is to allocate the mutex

richardjrossiii:
@phausler Ah, so the compiler makes an actual temporary copy of the variable in order to pass it as a pointer when passed via inout? 

That seems like it could be improved at the compiler level, then.

Very good knowledge to have then, that '&' is not truly addressOf in swift when interfacing with C (which of course begs the question - what makes `unsafeAddressOf` special?).

gribozavr:
@phausler is right.  The compiler can insert a copy at any point.  This won't change, this is the Swift model for structs.  `withUnsafePointer` does not guarantee you anything.  `unsafeAddressOf` only works with class instances.

PR #262, opened 2016-02-14
Fixes for TestNSURL.swift on 32-bit (NFC)

hpux735:
This is a relatively minor patch that casts the result of sizeof to UInt so that UInt32 (from 32-bit machines) is promoted to match the other logic.


gribozavr:
Are you sure this is not a symptom of incorrectly importing some API on Linux/armv7, for example, strlen()?

hpux735:
@gribozavr sorry for the delay in getting back, I was focusing on that other PR.  I don't see anywhere where strlen is imported differently from other c functions.  Ultimately, size_t is being imported as a UInt32 rather than a UInt.  Looking at MappedTypes, it seems like it should be mapped as Int.

gribozavr:
> Ultimately, size_t is being imported as a UInt32 rather than a UInt.

Which is an issue that should be fixed on the compiler side.  It will cause a lot of other API inconsistencies.

PR #263, opened 2016-02-15
[swift-corelibs-foundation]-- Fixed JSON Deserialisation of string values with spaces at the start String

salmanjamil:
currently for strings that had spaces at the start, the spaces were ignored during deserialisation process. e.g "{\"title\" : \" hello world!!\" }" would deserialise into ["title" : "hello world!!"] instead of ["title" : " hello world!"] skipping the space at the start of hello world. This was due to `func consumeStructure(ascii: UInt8, input: Index) throws -> Index?` consuming white spaces at the start and end of every structure. The fix is  just to consume the starting whitespace sequence in case of QuotationMark Structure.

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -306,7 +306,12 @@ private struct JSONReader {
     }
 
     func consumeStructure(ascii: UInt8, input: Index) throws -> Index? {
-        return try consumeWhitespace(input).flatMap(consumeASCII(ascii)).flatMap(consumeWhitespace)
+        switch ascii {
+        case Structure.QuotationMark:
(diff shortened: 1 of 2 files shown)


argon:
I think it would be preferable to treat this as a special case within `parseString` rather than complicate the `consumeStructure` function

```swift
func parseString(input: Index) throws -> (String, Index)? {
    guard let beginIndex = try consumeWhitespace(input).flatMap(consumeASCII(Structure.QuotationMark)) else {
        return nil
    }
```

parkera:
Let's get the bug fix in now and fix up the style next.

argon:
Great stuff. I'll get a pull request with the change opened shortly.

argon:
☝️ 👉 #341

PR #266, opened 2016-02-25
Add required flags for using the new SwiftOnoneSupport library

swiftix:
This is a PR related to the following PR: https://github.com/apple/swift/pull/1429
They should be merged in sync, otherwise foundation builds will fail after the mentioned PR is merged.

### DIFF
--- a/build.py
+++ b/build.py
@@ -23,6 +23,9 @@ elif Configuration.current.target.sdk == OSType.MacOSX:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_MACOSX '
 	foundation.LDFLAGS = '-licucore -twolevel_namespace -Wl,-alias_list,CoreFoundation/Base.subproj/DarwinSymbolAliases -sectcreate __UNICODE __csbitmaps CoreFoundation/CharacterSets/CFCharacterSetBitmaps.bitmap -sectcreate __UNICODE __properties CoreFoundation/CharacterSets/CFUniCharPropertyDatabase.data -sectcreate __UNICODE __data CoreFoundation/CharacterSets/CFUnicodeData-L.mapping -segprot __UNICODE r r '
 
+if Configuration.current.build_mode == Configuration.Debug:
+        foundation.LDFLAGS += ' -lswiftSwiftOnoneSupport '
+
 # For now, we do not distinguish between public and private headers (they are all private to Foundation)
 # These are really part of CF, which should ultimately be a separate target
 foundation.ROOT_HEADERS_FOLDER_PATH = "${PREFIX}/lib/swift"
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -68,6 +71,15 @@
                 '-I'+Configuration.current.variables["LIBDISPATCH_BUILD_DIR"]+'/tests'  # for include of dispatch/private.h in CF
         ])
 
+if Configuration.current.build_mode == Configuration.Release:
+        foundation.LDFLAGS += ' -lswiftSwiftOnoneSupport '
+        # In the future, remove the LDFLAGS line above and enable
+        # the optimizations. But currently, it results in a segmentation
+        # fault on one of the tests.
```
phausler: if this causes a segfault in one of the tests and it is currently working that probably means there is an expectation failure that is not met in your library:

it is worth noting the lines below are incorrect and should not be modified at that point:

from lib/script.py:111

```python
        if Configuration.current.build_mode == Configuration.Debug:
            swift_flags += "-g -Onone "
        elif Configuration.current.build_mode == Configuration.Release:
            swift_flags += "-O "

```

that modification runs just fine without your changes; can you get me the crash point by attaching lldb?

```sh
cd swift-corelibs-foundation
ninja test
LD_LIBRARY_PATH=../build/Ninja-ReleaseAssert/foundation-linux-x86_64/Foundation/:/home/phausler/Public/build/Ninja-ReleaseAssert/xctest-linux-x86_64 ../build/Ninja-ReleaseAssert/lldb-linux-x86_64/bin/lldb ../build/Ninja-ReleaseAssert/foundation-linux-x86_64/TestFoundation/TestFoundation
```

presuming you built as release mode in the standard layouts

swiftix:
@phausler Actually, I was wrong. There is no crash anymore. Looks like I fixed it on the Swift side after I tested yesterday. Which means, that we can enable `-O` for Release, if we want. Or we can continue with  `foundation.LDFLAGS += ' -lswiftSwiftOnoneSupport '`. It is up to you what you prefer for foundation.

swiftix:
Let me know if you want me to submit the corresponding PR or if you want to commit this one-liner yourself.

phausler:
I will change the `-O` for release mode in a follow up

swiftix:
OK. Should I then just remove the Release related part altogether and have only the change for LDFLAGS in Debug mode?

swiftix:
Or should I remove the comment about `-O` which is obsolete by now.

phausler:
as soon as the corresponding swift commit is merged I will merge this and push corrections for the release builds

swiftix:
OK. Understood. I'm about to merge the Swift commit now.

swiftix:
Done.

swiftix:
With your `-O` commit, you probably want to remove the
`Configuration.current.build_mode == Configuration.Release:` block from the build script. It is not needed in `-O` builds. If you leave it there, it may result in `SwiftOnoneSupport` being linked and referred even though it is not used by an app or a dylib.

But let's first see that everything builds fine. Once everything is fine, we can remove this if-statement.

swiftix:
@phausler Thanks for your support!

PR #270, opened 2016-02-29
Update to support the latest Corelibs XCTest

briancroom:
This includes the necessary changes for the test suite to build and run with the `swift-corelibs-xctest` changes proposed in https://github.com/apple/swift-corelibs-xctest/pull/40 and should be merged in coordination with that PR.


TestFoundation/TestNSString.swift
```diff
@@ -81,7 +81,7 @@ class TestNSString : XCTestCase {
             ("test_ExternalRepresentation", test_ExternalRepresentation),
             ("test_mutableStringConstructor", test_mutableStringConstructor),
             ("test_PrefixSuffix", test_PrefixSuffix),
-            ("test_reflection", test_reflection),
+            ("test_reflection", { _ in test_reflection }),
```
phausler: unrelated to this change but we should consider moving this into an actual test scope.

phausler:
Overall looks good to me, as soon as  apple/swift-corelibs-xctest#40 is merged we should bring this in as well

PR #271, opened 2016-03-01
Migrating Foundation to Swift 3

moiseev:
This branch is a result of running migrator on master and polishing it manually until it compiled on Linux.
Attempt to compile tests results in a weird error that needs further investigation.


Foundation/NSFileManager.swift
```diff
@@ -340,17 +340,15 @@ public class NSFileManager : NSObject {
         result[NSFileSystemFileNumber] = NSNumber(unsignedLongLong: UInt64(s.st_ino))
         
         let pwd = getpwuid(s.st_uid)
-        if pwd != nil && pwd.memory.pw_name != nil {
-            if let name = String.fromCString(pwd.memory.pw_name) {
-                result[NSFileOwnerAccountName] = name
-            }
+        if pwd != nil && pwd.pointee.pw_name != nil {
+            let name = String(cString: pwd.pointee.pw_name)
+            result[NSFileOwnerAccountName] = name
```
phausler: why was this logic changed? couldn't a string from c potentially be corrupted? I have seen that type of failure in C before why can't it happen here? Is this perhaps the wrong API to call if we want a failable C string conversion (without bouncing to CFStringRef/NSString, which can handle a failable conversion)

Foundation/NSNumber.swift
```diff
@@ -11,22 +11,22 @@
...
-internal let kCFNumberShortType = CFNumberType.ShortType
-internal let kCFNumberIntType = CFNumberType.IntType
-internal let kCFNumberLongType = CFNumberType.LongType
-internal let kCFNumberLongLongType = CFNumberType.LongLongType
-internal let kCFNumberFloatType = CFNumberType.FloatType
-internal let kCFNumberDoubleType = CFNumberType.DoubleType
-internal let kCFNumberCFIndexType = CFNumberType.CFIndexType
-internal let kCFNumberNSIntegerType = CFNumberType.NSIntegerType
-internal let kCFNumberCGFloatType = CFNumberType.CGFloatType
+internal let kCFNumberSInt8Type = CFNumberType.sInt8Type
```
phausler: wow these look pretty poorly named now...

phausler:
What is the error you are running into? Is it a compilation failure or runtime?

phausler:
@swift-ci Please test

Foundation/NSFileManager.swift
moiseev: In general we're trying to avoid static factory methods and replace them with initializers. This particular initializer (`init(cString:)`) will try to repair ill-formed code units, which, we believe, is the behavior one would prefer by default (as opposed to the failing one). If you do prefer it to fail -- there is `String.init?(validatingUTF8:)`.
Please note though, that both initializers will trap on nil.

moiseev:
Rebased on top of current master to fix merge conflicts.

phausler:
Thanks, we have gotten a bit behind in our PRs. Overall the changes look like they applied just fine so far. Is this blocked on any specific swift or XCTest PRs?

moiseev:
I tried to reproduce the compilation error I mentioned above this morning, but it just wasn't there. Right now I'm migrating XCTest, after which will migrate Foundation tests and update this PR.

moiseev:
@phausler made tests compile after [migrating XCTest](https://github.com/apple/swift-corelibs-xctest/pull/60).
Seeing the following error now while running tests:
2__moiseev_rad_____ssh_

phausler:
@moiseev can you run that with lldb and get me a backtrace? is that the only failure if you just disable that test?

moiseev:
I was going to do just that, haven't got time yet. Will try tomorrow morning.

moiseev:
Ok. So, I've commented out the `test_xpath` test and it all runs smoothly now. Also squashed migration commits into one and rebase on top of current master. If commenting test is OK for now, I'll consider it done and move forward towards finally making Swift 3 a new master. Does that sound like a plan, @phausler?

parkera:
Good with me.

phausler:
yea, all looks good. Filed https://bugs.swift.org/browse/SR-878 to follow up on the test_xpath issue; probably would be a decent starter bug too since it would start off with just uncommenting and debugging.

moiseev:
@phausler that was a bit too early, Swift master is not Swift 3 yet, neither is XCTest. Could you please revert this one?

phausler:
ok sorry bout that, I thought we were good to go on this.

moiseev:
Thanks. I will send a new PR when we are actually ready.

moiseev:
Rebased. Fixed. Tested. Ready for merge now.

moiseev:
Oops. Commented here instead of #281.

PR #274, opened 2016-03-04
Initial NSDateFormatter tests

tfrank64:
This PR primarily contains tests, but I also added CF attribute setters. I tried to use a wide variety of dates, but more could be added in the future.

More `NSDateFormatter` work coming in future pull requests.

### DIFF
--- a/TestFoundation/TestNSDateFormatter.swift
+++ b/TestFoundation/TestNSDateFormatter.swift
@@ -1,0 +1,45 @@
+//
+//  TestNSDateFormatter.swift
+//  Foundation
+//
+//  Created by Taylor Franklin on 2/19/16.
+//  Copyright © 2016 Apple. All rights reserved.
+//
+
+#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
+    import Foundation
+    import XCTest
+#else
+    import SwiftFoundation
+    import SwiftXCTest
+#endif
+
+class TestNSDateFormatter: XCTestCase {
+    
+    var allTests : [(String, () throws -> Void)] {
+        return [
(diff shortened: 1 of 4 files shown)


parkera:
@swift-ci please test

parkera:
Thanks, always great to have more tests in place.

tfrank64:
Glad to help!

phausler:
lgtm

phausler:
wonder if this would prove to be useful for the person who made http://nsdateformatter.com

tfrank64:
Good question, I saw that yesterday and thought it would be helpful for me.

TestFoundation/TestNSDateFormatter.swift
```diff
@@ -0,0 +1,178 @@
...
+    }
+    
+    // FullStyle
+    // locale  stringFromDate  example
+    // ------  --------------  -------------------------
+    // en_US   EEEE, MMMM d, y  Friday, December 25, 2015
+    func test_dateStyleFull() {
+        
+        let timestamps = [
+            -31536000 : "Wednesday, January 1, 1969 at 12:00:00 AM GMT" , 0.0 : "Thursday, January 1, 1970 at 12:00:00 AM GMT",
```
bubski: Shouldn't the correct value be `Greenwich Mean Time` instead of `GMT`?

That's what ICU spec says (http://userguide.icu-project.org/formatparse/datetime)
Also, that's the behaviour of `f.timeStyle = .full` on macOS.

> It appears that for negative timestampts (e.g. -31536000) DateFormatter on macOS wrongly returns GMT instead of Greenwich Mean Time, which seems to be a bug.

PR #275, opened 2016-03-04
SR-794 pass supplied subpath through to URLForResource rather than nil

seabaylea:
[SR-794](https://bugs.swift.org/browse/SR-794) reports an issue with `NSBundle.pathForResource()` whereby it returns `nil` instead of the expected resource.

It looks like this is cause because the supplied `subpath` value isn't passed down to the underlying `URLForResource()` call.

I've fixed that and added an implementation of the test case to TestNSBundle. I've done that under the existing `test_URLsForResourcesWithExtension()` test as a separate call and check as its a similar test and reuses the same bundle setup/teardown. I can spin it off as a separate test though if it makes sense.

### DIFF
--- a/Foundation/NSBundle.swift
+++ b/Foundation/NSBundle.swift
@@ -205,7 +205,7 @@ public class NSBundle : NSObject {
     }
     
     public func pathForResource(name: String?, ofType ext: String?, inDirectory subpath: String?) -> String? {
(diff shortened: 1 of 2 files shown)


parkera:
Looks good. Will run through tests then we can merge.

parkera:
@swift-ci please test

PR #277, opened 2016-03-07
Implement missing stringValue and descriptionWithLocale APIs

seabaylea:
The [NSNumber reference](https://developer.apple.com/library/prerelease/ios/documentation/Cocoa/Reference/Foundation/Classes/NSNumber_Class/index.html#//apple_ref/occ/instm/NSNumber/descriptionWithLocale:) documents the following two APIs which appear not to be implemented (even as calls to `NSUnimplemented`):
```swift
var stringValue: String { get }
func descriptionWithLocale(_ locale: AnyObject?) -> String
```
Additionally `stringValue()` is described as a call to `descriptionWithLocale(nil)`.

I've therefore implemented `descriptionWithLocal()`, and `stringValue()` as a call through to it.

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -434,6 +434,10 @@ public class NSNumber : NSValue {
         return val
     }
     
+    public var stringValue: String {
+        return descriptionWithLocale(nil)
+    }
+    
     /// Create an instance initialized to `value`.
     public required convenience init(integerLiteral value: Int) {
         self.init(integer: value)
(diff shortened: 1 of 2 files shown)


Foundation/NSNumber.swift
```diff
@@ -434,6 +434,10 @@ public class NSNumber : NSValue {
         return val
     }
     
+    public var stringValue: String {
+        return descriptionWithLocale(nil);
```
ianpartridge: Semi-colon not needed.

seabaylea: Old habits ;-)

Foundation/NSNumber.swift
```diff
@@ -452,6 +456,12 @@ public class NSNumber : NSValue {
     public func compare(otherNumber: NSNumber) -> NSComparisonResult {
         return ._fromCF(CFNumberCompare(_cfObject, otherNumber._cfObject, nil))
     }
+
+    public func descriptionWithLocale(locale: AnyObject?) -> String {
+        guard let aLocale = locale else { return description }
```
phausler: from a subclassing perspective, iirc this is a bit inverted - descriptionWithLocale should be the funnel point and description should call descriptionWithLocale(nil) but that is a minor 'nit since this is pretty reasonable as it stands.

PR #278, opened 2016-03-08
Fix a false match bug in NSRegularExpression

pushkarnk:
Code changes to fix a false match returned by firstMatchInString().
Also include 35 out of 37 tests from  test_complexRegularExpressions
in TestNSRegularExpression. The two tests currently excluded from 
test_complexRegularExpressions() fail due to another defect @mamabusi 
is working on.

### DIFF
--- a/CoreFoundation/String.subproj/CFRegularExpression.c
+++ b/CoreFoundation/String.subproj/CFRegularExpression.c
@@ -226,7 +226,8 @@ CF_INLINE URegularExpression *prepareRegularExpression(void *internal, int32_t *
     UErrorCode errorCode = U_ZERO_ERROR;
     CFRange enclosingRange;
     UniChar *stringBuffer = NULL;
-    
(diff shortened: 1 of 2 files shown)


phausler:
this looks about right, but I am still a touch concerned about the edge case behavior, @parkera how much should we strive for bug-for-bug compatibility?

pushkarnk:
I discovered a small typo in complexRegularExpressionTest() which throws could lead to confusion if a test failed. 
```first.rangeAtIndex(1)``` which is the range of the first capture group match should be compared to ```firstMatchFirstCaptureRange``` and not ```firstMatchOverallRange```

Updated the pr.

parkera:
I don't think we need to be bug-for-bug compatible (at least not at this stage); let's go with what we have now and if we have issues then we'll keep iterating from there.

phausler:
@swift-ci Please test

PR #279, opened 2016-03-08
[SR-765] Implement NSNumberFormatter.stringForObjectValue()

seabaylea:
[SR 765](https://bugs.swift.org/browse/SR-765) reports that `NSNumberFormatter` is missing an implementation for `stringForObjectValue()`.

I've added an implementation for that which follows the [NSFormatter Class Reference](https://developer.apple.com/library/prerelease/ios/documentation/Cocoa/Reference/Foundation/Classes/NSFormatter_Class/index.html#//apple_ref/occ/instm/NSFormatter/stringForObjectValue:) by returning `nil` if the class is incorrect, otherwise returning a formatted string. The formatted string is obtained by calling through to `stringFromNumber()`, which also deals with the localisation of the string.


phausler:
looks good to me

PR #280, opened 2016-03-09
Fix API parity to ObjC Foundation NSURL.absoluteString

ketzusaka:
The Objective-C Foundation's NSURL.absoluteString returns non-optional, but SwiftFoundation's is currently an optional. This makes the APIs consistent between eachother so importing Foundation on either platform works the same with this method.


phausler:
looks good to me.

PR #281, opened 2016-03-10
Porting to Swift 3

moiseev:


moiseev:
@phausler, I've just merged changes to Swift, so this one is good to go.

jessesquires:
Hey @moiseev :smile: -- This looks nearly identical to what was merged in #271. Should this be rebased on master again?

gribozavr:
Merging the PR to bring CI back to green.

moiseev:
It is indeed the same thing. #271 was merged and then reverted.

moiseev:
Thanks, Dmitri! This one depends in the XCTest, so if you could merge that one as well...

moiseev:
Here is the xctest pr: https://github.com/apple/swift-corelibs-xctest/pull/60

jessesquires:
@moiseev -- ah, thanks for the clarification! :smile:

tfrank64:
Maybe it has been mentioned somewhere, but is there any instruction on running Foundation now with all these Swift 3 changes. I'm struggling to get things going on OSX or Linux.

moiseev:
@tfrank64 you should not have to change anything. On Linux I used `swift/utils/build-util --xctest --foundation -RT` and then `swift-corelibs-foundation/ninja test` both before and after applying changes. What are the exact problems you have?

tfrank64:
@moiseev I'm uncertain what kind of issue I have now. I can run `swift/utils/build-script --xctest --foundation -RT` on the `master` branch of swift in Linux. With foundation, everything works on the master branch, but when I switch to my branch, the tests stop at my changes saying `Trace/breakpoint trap (core dumped)`. So basically I'm not sure what that means or how to debug, because the debugging command `ninja test` gives me, needs lldb and apparently I don't have that

moiseev:
@tfrank64 hm.. Since PR is not a very good place for such discussion, I would suggest posting detailed description of your problem to swift-dev mailing list, perhaps somebody has the solution already. At the very least it will be brought to attention of a much wider audience. Feel free to CC me.

PR #282, opened 2016-03-10
Replace NSNotFound for kCFNotFound in NSRange

mamabusi:
CoreFoundation returns a "not found"  range as {-1, 0}.
This needs to be represented as {NSNotFound, 0} in Foundation.
Included all but four tests from test_complexRegularExpression
in TestNSRegularExpression (the excluded depend on PR 278)
Added four new tests as well.


phausler:
@swift-ci Please test

PR #283, opened 2016-03-12
Generalize trap intrinsic and allow aarch64

hpux735:
These are just some small changes that begin to enable building Foundation on 64-bit arm linux.


parkera:
@swift-ci please test

parkera:
I guess our CI is still in bad shape from the Swift 3 transition...

hpux735:
Thanks for checking anyway, It's helpful for me because it wouldn't have gotten that far if I had broken it. :)

phausler:
looks pretty reasonable to me, I wonder if we should just reduce the trap for all platforms down to the builtin trap.

hpux735:
I don't remember where it was mentioned, but from what I understand, the existing dependence on clang meant that relying on _builtin_trap did not introduce any extra dependencies.  It's clearly more portable, so it seems like a good idea to me.

PR #286, opened 2016-03-15
[SR-929] Implement NSString.init() with locale

seabaylea:
[SR-929](https://bugs.swift.org/browse/SR-929) covers the need to implement the following NSString API:
```swift
public convenience init(format: String, locale: AnyObject?, arguments argList: CVaListPointer)
```

This PR adds an implementation that:
- Allows the for `locale` parameter to be a `NSLocale`,  `NSDictionary` or `nil`.
The parameter type is checked, checking `NSLocale` first as its the expected type
- Casts the type to `CFDictionary` for use in `CFString`
Here I've used `unsafeBitCast` as casting to a CFDictionary directly doesn't work on Linux and there doesn't appear to be a "safer" was of casting that works. The underlying functions check to see if the actual type is a `CFLocale` or a `CFDictionary` and handles as appropriate.
- Throws a `fatalError()` for invalid locale parameters
The actual behaviour is undefined by the spec. As the init function isn't failable there needs to be some kind of error, and it looked better to fail here with a relevant message than further down the call stack.

The behaviour is the same as for Objective-C on Mac, except for the error handling. Let me know if there's any suggest changes to the implementation.

I've also added 4 tests for the implementation:
1. Checking the use of `nil`
2. Checking with the English locale using NSLocale, returning 1,000 (42.0)
3. Checking with the German locale using NSLocale, returning 1.000 (42,0)
4. Checking with a custom decimal separator using a NSDictionary, returning 1000 (42&0)

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -1214,7 +1214,17 @@ extension NSString {
     }
     
     public convenience init(format: String, locale: AnyObject?, arguments argList: CVaListPointer) {
-        NSUnimplemented()    
+        let str: CFString
+        if let loc = locale {
+            if loc.dynamicType === NSLocale.self || loc.dynamicType === NSDictionary.self {
+                str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, unsafeBitCast(loc, to: CFDictionary.self), format._cfObject, argList)
+            } else {
+                fatalError("locale parameter must be a NSLocale or a NSDictionary")
+            }
+        } else {
+            str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, nil, format._cfObject, argList)
+        }
+        self.init(str._swiftObject)
     }
     
     public convenience init(format: NSString, _ args: CVarArgType...) {
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -70,6 +70,7 @@ class TestNSString : XCTestCase {
             ("test_stringByTrimmingCharactersInSet", test_stringByTrimmingCharactersInSet),
             ("test_initializeWithFormat", test_initializeWithFormat),
             ("test_initializeWithFormat2", test_initializeWithFormat2),
+            ("test_initializeWithFormat3", test_initializeWithFormat3),
             ("test_stringByDeletingLastPathComponent", test_stringByDeletingLastPathComponent),
             ("test_getCString_simple", test_getCString_simple),
             ("test_getCString_nonASCII_withASCIIAccessor", test_getCString_nonASCII_withASCIIAccessor),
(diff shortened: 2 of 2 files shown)


parkera:
@swift-ci please test

parkera:
Hm, looks like we've got some Swift 3 cleanup still to do in our test cases...

seabaylea:
I'll make the changes and update this PR with them.

seabaylea:
@parkera updated and squashed in place.

Foundation/NSString.swift
```diff
@@ -1214,7 +1214,17 @@ extension NSString {
...
+        if let loc = locale {
+            if loc.dynamicType === NSLocale.self || loc.dynamicType === NSDictionary.self {
+                str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, unsafeBitCast(loc, to: CFDictionary.self), format._cfObject, argList)
+            } else {
+                fatalError("locale parameter must be a NSLocale or a NSDictionary")
+            }
+        } else {
+            str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, nil, format._cfObject, argList)
+        }
+        self.init(str._swiftObject)
```
phausler: The one possible issue I could see with this implementation is that it wont call the description with locale on the NS objects: _CFStringCreateWithFormatAndArgumentsAux2 can pass a function pointer to allow for fetching the descriptionWithLocale() off of the objects. However that would mean that we would need to have an internal protocol to identify those classes who respond to that method.

seabaylea: Ok - so we need to pass a function pointer to `_CFStringCreateWithFormatAndArgumentsAux2()` for a function that's capable of calling any implementation of `descriptionWithLocale()` on objects passed in as part of the argList. Does this only apply to CF/NS objects, you do we need to handle 'plain' Swift objects as well?

phausler: Technically on darwin the only reason why you can pass a swift class to this function is because they have root objc inheritance (that has the right methods implemented). However you cannot pass a struct to it even if it does conform to the descriptionWithLocale or the CustomStringConvertible protocol and I doubt there is any real way we can get struct passing to va_args to work in swift on either platform for formatting.

As it stands this implementation looks like a good start, if you want to make a bug on bugs.swift.org and have the format callbacks to be a later followup that is perfectly reasonable.

seabaylea: @phausler ok - I've raised [SR-1018 NSString.init() with locale doesn't handle descriptionWithLocale() callbacks on arguments](https://bugs.swift.org/browse/SR-1018)	

to cover the additional work.

https://bugs.swift.org/browse/SR-1018

gribozavr:
Sorry, but it looks like there is a CI failure: https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-15_10/3818/console

Could someone investigate or revert for now?

phausler:
actually it looks like the test is incorrect. will push a fix shortly

PR #287, opened 2016-03-15
Implement NSHTTPURLResponse

danieleggert:
This implements both the trivial attributes (`statusCode` and `allHeaderFields`), and also parses the headers in order to set the following attributes:
```
public var mimeType: String? { get }
public var expectedContentLength: Int64 { get }
public var textEncodingName: String? { get }
public var suggestedFilename: String? { get }
```
Added tests for these.

The
```
public class func localizedStringForStatusCode(statusCode: Int) -> String { NSUnimplemented() }
```
method on `NSHTTPURLResponse` is still unimplemented.

### DIFF
--- a/Foundation/NSURLResponse.swift
+++ b/Foundation/NSURLResponse.swift
@@ -52,11 +52,13 @@ public class NSURLResponse : NSObject, NSSecureCoding, NSCopying {
         @result The initialized NSURLResponse.
         @discussion This is the designated initializer for NSURLResponse.
     */
-    public init(URL: NSURL, MIMEType: String?, expectedContentLength length: Int, textEncodingName name: String?) {
-        self.URL = URL
-        self.MIMEType = MIMEType
+    public init(url: NSURL, mimeType: String?, expectedContentLength length: Int, textEncodingName name: String?) {
+        self.url = url
+        self.mimeType = mimeType
         self.expectedContentLength = Int64(length)
         self.textEncodingName = name
+        let c = url.lastPathComponent
+        self.suggestedFilename = (c?.isEmpty ?? true) ? "Unknown" : c
     }
     
     /*! 
@@ -64,7 +66,7 @@ public class NSURLResponse : NSObject, NSSecureCoding, NSCopying {
         @abstract Returns the URL of the receiver. 
         @result The URL of the receiver. 
     */
-    /*@NSCopying*/ public private(set) var URL: NSURL?
+    /*@NSCopying*/ public private(set) var url: NSURL?
 
     
     /*! 
@@ -78,7 +80,7 @@ public class NSURLResponse : NSObject, NSSecureCoding, NSCopying {
         be made if the origin source did not report any such information.
         @result The MIME type of the receiver.
     */
-    public private(set) var MIMEType: String?
+    public private(set) var mimeType: String?
     
     /*! 
         @method expectedContentLength
@@ -120,7 +122,7 @@ public class NSURLResponse : NSObject, NSSecureCoding, NSCopying {
         This method always returns a valid filename.
         @result A suggested filename to use if saving the resource to disk.
     */
-    public var suggestedFilename: String? { NSUnimplemented() }
+    public private(set) var suggestedFilename: String?
 }
 
 /*!
@@ -143,7 +145,17 @@ public class NSHTTPURLResponse : NSURLResponse {
       @result 	the instance of the object, or NULL if an error occurred during initialization.
       @discussion This API was introduced in Mac OS X 10.7.2 and iOS 5.0 and is not available prior to those releases.
     */
-    public init?(URL url: NSURL, statusCode: Int, HTTPVersion: String?, headerFields: [String : String]?) { NSUnimplemented() }
+    public init?(url: NSURL, statusCode: Int, httpVersion: String?, headerFields: [String : String]?) {
+        self.statusCode = statusCode
+        self.allHeaderFields = headerFields ?? [:]
+        super.init(url: url, mimeType: nil, expectedContentLength: 0, textEncodingName: nil)
+        expectedContentLength = getExpectedContentLength(fromHeaderFields: headerFields) ?? -1
+        suggestedFilename = getSuggestedFilename(fromHeaderFields: headerFields) ?? "Unknown"
+        if let type = ContentTypeComponents(headerFields: headerFields) {
+            mimeType = type.mimeType.lowercased()
+            textEncodingName = type.textEncoding?.lowercased()
+        }
+    }
     
     public required init?(coder aDecoder: NSCoder) {
         NSUnimplemented()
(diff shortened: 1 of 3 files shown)


danieleggert:
**Note**: This changes the type of `allHeaderFields` to `[String : String]` — I hope that's ok?

phausler:
That actually makes sense since the init method is NSDictionary<NSString *, NSString *> *)headerFields.

It seems like that might just be an oversight for the property. I will follow up with that component owner to determine if it is actually the case that it can never be anything but strings.

danieleggert:
@phausler The same goes for `NSURLSessionConfiguration` and `HTTPAdditionalHeaders`.

Foundation/NSURLResponse.swift
```diff
@@ -167,9 +178,9 @@ public class NSHTTPURLResponse : NSURLResponse {
         @result A dictionary containing all the HTTP header fields of the
         receiver.
     */
-    public var allHeaderFields: [NSObject : AnyObject] { NSUnimplemented() }
+    public let allHeaderFields: [String : String]
```
phausler: This should have a comment on why the differential and identifying it as an "experimental" change; albeit that I agree that this should be the signature, I have not yet heard back from the component owner on a definitive reason on either why it should or should not be marked as string to string.

Foundation/NSURLResponse.swift
```diff
@@ -178,4 +189,136 @@ public class NSHTTPURLResponse : NSURLResponse {
...
+    }
+    return nil
+}
+/// Parses the `Content-Type` header field
+///
+/// - Returns: Either a media type and `nil`, or a mesia type and a character set.
+///
+/// `Content-Type: text/html; charset=ISO-8859-4` would result in `("text/html", "ISO-8859-4")`, while
+/// `Content-Type: text/html` would result in `("text/html", nil)`.
+private func mimeTypeAndTextEncodingFromHeaderFields(headerFields: [String : String]?) -> (String,String?)? {
```
phausler: perhaps this would be better served with a struct return? or naming the tuple fields?

Foundation/NSURLResponse.swift
```diff
@@ -178,4 +189,136 @@ public class NSHTTPURLResponse : NSURLResponse {
...
+    return (parts.0, nil)
+}
 
+private extension String {
+    /// Split the string at each ";", remove any quoting.
+    /// 
+    /// The trouble is if there's a
+    /// ";" inside something that's quoted. And we can escape the separator and
+    /// the quotes with a "\".
+    var httpHeaderParts: (String, [(String,String?)])? {
```
phausler: definitely that structure here would be nice, I am not certain what this return value is representing structurally.

TestFoundation/TestNSURLResponse.swift
```diff
@@ -60,4 +61,208 @@ class TestNSURLResponse : XCTestCase {
...
+                   ("test_suggestedFilename_1", test_suggestedFilename_1),
+                   ("test_suggestedFilename_2", test_suggestedFilename_2),
+                   ("test_suggestedFilename_3", test_suggestedFilename_3),
+                   ("test_suggestedFilename_4", test_suggestedFilename_4),
+                   ("test_suggestedFilename_onlyTheLastPathComponent_1", test_suggestedFilename_onlyTheLastPathComponent_1),
+                   ("test_suggestedFilename_onlyTheLastPathComponent_2", test_suggestedFilename_onlyTheLastPathComponent_2),
+                   
+                   ("test_MIMETypeAndCharacterEncoding_1", test_MIMETypeAndCharacterEncoding_1),
+                   ("test_MIMETypeAndCharacterEncoding_2", test_MIMETypeAndCharacterEncoding_2),
+                   ("test_MIMETypeAndCharacterEncoding_3", test_MIMETypeAndCharacterEncoding_3),
```
phausler: I wish the swift-ci could run these the objc foundation. looks pretty comprehensive. I will have to run these tests manually later to fully review this.

danieleggert:
Thanks for the feedback. I’ve tried to address you suggesting with ae45629 and 82a818e. I’ll squash everything into a single commit if / once I get a green light.

phausler:
looks pretty good to me.

danieleggert:
Should this be assigned to @salgernon ?

danieleggert:
I rebased against master and squashed the 3 commits into a single commit.

phausler:
The tests seem to have some failures when it is tested against the objective-c version for me:

```
Test: test_contentLength_notAvailable_2() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional(997)") is not equal to ("Optional(-1)") - 
File: test.swift:174



Test: test_contentLength_notAvailable_3() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional(997)") is not equal to ("Optional(-1)") - 
File: test.swift:179



Test: test_contentLength_notAvailable_4() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional(997)") is not equal to ("Optional(-1)") - 
File: test.swift:184



Test: test_contentLength_notAvailable_5() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional(997)") is not equal to ("Optional(-1)") - 
File: test.swift:189



Test: test_MIMETypeAndCharacterEncoding_3() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional("iso-8859-4")") is not equal to ("Optional("ISO-8859-4")") - 
File: test.swift:265



Test: test_suggestedFilename_4() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional("wrong.ext")") is not equal to ("Optional("fname.ext")") - 
File: test.swift:234



Test: test_suggestedFilename_notAvailable_1() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertNil failed: "Unknown" - 
File: test.swift:208



Test: test_suggestedFilename_notAvailable_2() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertNil failed: "Unknown" - 
File: test.swift:213



Test: test_suggestedFilename_onlyTheLastPathComponent_1() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional("_a_b_name")") is not equal to ("Optional("name")") - 
File: test.swift:239



Test: test_suggestedFilename_onlyTheLastPathComponent_2() in TestNSHTTPURLResponse failed on My Mac
Assertions: XCTAssertEqual failed: ("Optional("a_.._b_name")") is not equal to ("Optional("name")") - 
File: test.swift:244

Test: test_suggestedFilename() in TestNSURLResponse failed on My Mac
Assertions: XCTAssertNil failed: "path" - 
File: test.swift:67

```

danieleggert:
The failure in `test_suggestedFilename_4` looks like a bug in Darwin Foundation to me.

danieleggert:
I've addressed the difference to Objective-C Foundation with the 4 above commits.

phausler:
I tend to agree that probably the failure of `test_suggestedFilename_4` on Darwin is a bit un-expected; I will follow up with the networking team via radar on this. For now I would probably claim if that is the only differential it is an edge-case enough that it shouldn't hamper this PR.

phausler:
@swift-ci Please test

danieleggert:
I’ll squashed the commits into a single commit…

danieleggert:
@swift-ci Please test

parkera:
@swift-ci please test

parkera:
@swift-ci please test

Foundation/NSURLResponse.swift
```diff
@@ -154,28 +166,192 @@ public class NSHTTPURLResponse : NSURLResponse {
...
+    /// the "raw" header information which was reported to the protocol
+    /// implementation by the HTTP server. This may be of use to
+    /// sophisticated or special-purpose HTTP clients.
+    ///
+    /// - Returns: A dictionary containing all the HTTP header fields of the
+    /// receiver.
+    ///
+    /// - Important: This is an *experimental* change from the
+    /// `[NSObject: AnyObject]` type that Darwin Foundation uses.
+    public let allHeaderFields: [String: String]
```
kballard: This doesn't match the behavior of the real Foundation. Specifically, `allHeaderFields` uses a custom `NSDictionary` subclass that's case-insensitive when looking up keys. That case-insensitive behavior is pretty important behavior when dealing with HTTP headers.

It also canonicalizes certain headers, e.g. `Content-Length`, although it doesn't document which headers (besides Content-Length), and the canonicalization really only affects the case where you're displaying the header name so it's not really particularly important.

Given that Swift dictionaries are a value type and not a class, the only way we're going to get the same behavior while still using a dictionary type is to rely on bridging support and using an `NSDictionary` subclass, like the real Foundation does.

kballard: Also note that this has to handle the case where the same header is provided with different capitalizations to the initializer (in that event, the resulting `allHeaderFields` dictionary should only contain one of the equivalent keys and its corresponding value, and the conflicts should be discarded).

danieleggert: Good point. Can you open a bug on that and assign it to me? Thanks.

PR #288, opened 2016-03-16
Implement remaining NSDateFormatter attributes, with more tests

tfrank64:
[SR-208](https://bugs.swift.org/browse/SR-208) should be resolved with the additions of this PR and recent commit 89556789cf1003dfa137eec9eed43ed965f3ea74 setting of dateFormat.

In this PR:
- Implemented getters and setters on remaining properties missing them
- Added tests for initial values on NSDateFormatter
- Added tests for setting custom `.dateFormat` values

Also, tests will need PR #285 to actually run.


parkera:
Hm, I'm a little nervous about commenting out some tests without fully understanding why they are failing.

tfrank64:
@parkera Thats understandable, I've sent an email out to the mailing list in order to figure out why I can't debug on Linux, but I have received no response.

Another option is I revert `.AMSymbol` and `.PMSymbol` to their previous implementation since all other asserts/tests are passing.

tfrank64:
So I finally figured out the issue, it happened when I casted a `CFTypeRef` to a `CFString` instead of an `NSString`. Regardless, all tests pass for me on OSX and Linux.

phausler:
@swift-ci Please test

PR #289, opened 2016-03-16
Add implementation of NSTimeZone.init() with abbreviation

seabaylea:
This adds an implementation of the following NSTimeZone.init() API:
```swift
 public convenience init?(abbreviation: String)
```
The implementation does a lookup of the abbreviation in the CFTimeZoneAbbreviationDictionary to generate a name and, if found, uses that to create a NSTimeZone instance. If no entry is found, we return nil.

I've also added two tests, one to check the negative (invalid abbreviation) case, and one to check the valid case using "AST" (the first entry in the dictionary).


parkera:
@swift-ci please test

PR #290, opened 2016-03-16
Add 5 missing attributes to NSURLRequest

danieleggert:
cachePolicy, timeoutInterval, networkServiceType, HTTPBody, and HTTPBodyStream
Rework how attributes are stored.
Refactor header field code into existingHeaderField().

### DIFF
--- a/Foundation/NSURLRequest.swift
+++ b/Foundation/NSURLRequest.swift
@@ -164,10 +164,6 @@ public enum NSURLRequestNetworkServiceType : UInt {
 */
 public class NSURLRequest : NSObject, NSSecureCoding, NSCopying, NSMutableCopying {
     
-    private var _URL : NSURL?
-    private var _mainDocumentURL: NSURL?
-    private var _httpHeaderFields: [String: String]?
-    
     public override func copy() -> AnyObject {
         return copyWithZone(nil)
     }
@@ -235,7 +231,7 @@ public class NSURLRequest : NSObject, NSSecureCoding, NSCopying, NSMutableCopyin
     */
     public convenience init(URL: NSURL) {
         self.init()
-        _URL = URL
+        self.URL = URL
     }
     
     /*!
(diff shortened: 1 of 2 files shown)


parkera:
@salgernon - please take a look

parkera:
I think we intentionally left these out, but hopefully @salgernon can comment.

danieleggert:
I think it makes sense to have `cachePolicy`, `timeoutInterval`, and `networkServiceType` on all platforms.

parkera:
If I recall correctly, the reason we left them out is that they are already properties on NSURLSession:

```swift
    /* default cache policy for requests */
    public var requestCachePolicy: NSURLRequestCachePolicy
    
    /* default timeout for requests.  This will cause a timeout if no data is transmitted for the given timeout value, and is reset whenever data is transmitted. */
    public var timeoutIntervalForRequest: NSTimeInterval
    
    /* default timeout for requests.  This will cause a timeout if a resource is not able to be retrieved within a given timeout. */
    public var timeoutIntervalForResource: NSTimeInterval
    
    /* type of service for requests. */
    public var networkServiceType: NSURLRequestNetworkServiceType
    
    /* allow request to route over cellular. */
    public var allowsCellularAccess: Bool
```

danieleggert:
That's why `NSURLSession` has API that takes `NSURL` and API that takes `NSURLRequest`: The API that takes a request let's you override the values set by the `NSURLSessionConfiguration`. As the documentation states:
> By creating a task based on a request object, you can tune various aspects of the task’s behavior, including the cache policy and timeout interval.

Foundation/NSURLRequest.swift
```diff
@@ -364,11 +360,25 @@ public class NSMutableURLRequest : NSURLRequest {
...
+    }
+    public override var timeoutInterval: NSTimeInterval {
+        get { return super.timeoutInterval }
+        set { super.timeoutInterval = newValue }
+    }
+    public override var networkServiceType: NSURLRequestNetworkServiceType {
+        get { return super.networkServiceType }
+        set { super.networkServiceType = newValue }
+    }
+    public override var allowsCellularAccess: Bool {
```
phausler: can this ever be satisfied? It seems like it would require very specific hardware access

danieleggert: I'm fine with leaving `allowsCellularAccess` out.

parkera:
Yes, it was at the request of the owner of this particular API that we do not have this override capability, because they were concerned about the complexity it introduces.

However, since we're not really getting any response on this right now, we may as well add this for parity since it's confusing people right now. Let's leave the cellular one out.

danieleggert:
I just noticed that `HTTPBody` and `HTTPBodyStream` are also missing. I’ll need these for https://github.com/apple/swift-corelibs-foundation/pull/299 in order to implement `POST` and `PUT` requests. Thoughts?

danieleggert:
I've removed `allowsCellularAccess`, but added `HTTPBody` and `HTTPBodyStream`.

danieleggert:
Only after adding `HTTPBody` and `HTTPBodyStream` did I realise that these should not have been added because the `NSURLSession` API has separate methods for uploads (e.g. `POST` / `PUT`). Will remove these again. Sorry for the noise.

phausler:
@swift-ci Please test

danieleggert:
I'll close this.

danieleggert:
C.f. #306

PR #294, opened 2016-03-21
Fix inconsistent UUIDString behavior between ObjC Foundation and Swift Foundation

eyeplum:
This pull request should fix [SR-865](https://bugs.swift.org/browse/SR-865).

The `UUIDString` property of `NSUUID` should behave the same as the Objective-C counterpart. As the Objective-C version returns an uppercase string, the Swift version should behave the same.

### DIFF
--- a/Foundation/NSUUID.swift
+++ b/Foundation/NSUUID.swift
@@ -45,7 +45,7 @@ public class NSUUID : NSObject, NSCopying, NSSecureCoding, NSCoding {
     
     public var UUIDString: String {
         let strPtr = UnsafeMutablePointer<Int8>(allocatingCapacity: 37)
-        _cf_uuid_unparse_lower(buffer, strPtr)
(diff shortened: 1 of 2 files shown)


phausler:
looks good to me.

rcugut:
Is it a good idea to prolong a "mistake" (deviation from the standard) made in Obj-C?

According to RFC 4122
> The hexadecimal values "a" through "f" are output as lower case characters and are case insensitive on input.

eyeplum:
Hi @rcugut, we had [a brief discussion](https://bugs.swift.org/browse/SR-865?focusedCommentId=12829&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12829) before, and we think the consistent experience between DarwinFoundation and CorelibsFoundation is the goal for now. That's why we made a change like this.

PR #295, opened 2016-03-22
[SR-999] Add dot to NSString.stringByAppendingPathExtension() result

seabaylea:
This is for [SR-999](https://bugs.swift.org/browse/SR-999) which reports that `stringByAppendingPathExtension()` and `stringByDeletingPathExtension()` are failing to add/remove the 'dot' before the extension.

I've implemented tests that match the examples given in the API spec, along with an additional test including a tilda, and the modified the relevant functions in `NSPathUtilities`.


phausler:
looks good to me, the tests seem to cover the issue well.

PR #296, opened 2016-03-22
[SR-789] Remove localization of NSNumber.description

seabaylea:
This is for [SR-789](https://bugs.swift.org/browse/SR-789) which reports that `NSNumber.description` is providing a localised string whereas it didn't do so previously.

In order to be behaviour compatible I've made a slight modification to the existing implementation to use a number formatter style of `kCFNumberFormatterNoStyle` rather than `kCFNumberFormatterDecimalStyle`.

At the same time I've done some refactoring to improve the delegation between `description` and `descriptionWithLocale()` as suggested in PR #277 and added tests for both `description` and `descriptionWithLocale()`

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -458,8 +458,15 @@ public class NSNumber : NSValue {
     }
 
     public func descriptionWithLocale(locale: AnyObject?) -> String {
-        guard let aLocale = locale else { return description }
-        let formatter = CFNumberFormatterCreate(nil, (aLocale as! NSLocale)._cfObject, kCFNumberFormatterDecimalStyle)
+        let aLocale = locale
+        let formatter: CFNumberFormatter
+        if (aLocale == nil) {
+            formatter = CFNumberFormatterCreate(nil, CFLocaleCopyCurrent(), kCFNumberFormatterNoStyle)
+            CFNumberFormatterSetProperty(formatter, kCFNumberFormatterMaxFractionDigits, 15._bridgeToObject())
+
+        } else {
+            formatter = CFNumberFormatterCreate(nil, (aLocale as! NSLocale)._cfObject, kCFNumberFormatterDecimalStyle)
+        }
         return CFNumberFormatterCreateStringWithNumber(nil, formatter, self._cfObject)._swiftObject
     }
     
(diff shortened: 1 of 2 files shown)


Foundation/NSNumber.swift
```diff
@@ -468,11 +475,13 @@ public class NSNumber : NSValue {
...
-        let formatter = CFNumberFormatterCreate(nil, locale, kCFNumberFormatterDecimalStyle)
-        CFNumberFormatterSetProperty(formatter, kCFNumberFormatterMaxFractionDigits, 15._bridgeToObject())
-        return CFNumberFormatterCreateStringWithNumber(nil, formatter, self._cfObject)._swiftObject
+        return descriptionWithLocale(nil)
     }
+//        let locale = CFLocaleCopyCurrent()
+//        let formatter = CFNumberFormatterCreate(nil, locale, kCFNumberFormatterNoStyle)
+//        CFNumberFormatterSetProperty(formatter, kCFNumberFormatterMaxFractionDigits, 15._bridgeToObject())
+//        return CFNumberFormatterCreateStringWithNumber(nil, formatter, self._cfObject)._swiftObject
+//    }
```
phausler: probably can remove the commented out code. git history is a beautiful thing ;)

Foundation/NSNumber.swift
```diff
@@ -458,8 +458,15 @@ public class NSNumber : NSValue {
...
+        let aLocale = locale
+        let formatter: CFNumberFormatter
+        if (aLocale == nil) {
+            formatter = CFNumberFormatterCreate(nil, CFLocaleCopyCurrent(), kCFNumberFormatterNoStyle)
+            CFNumberFormatterSetProperty(formatter, kCFNumberFormatterMaxFractionDigits, 15._bridgeToObject())
+
+        } else {
+            formatter = CFNumberFormatterCreate(nil, (aLocale as! NSLocale)._cfObject, kCFNumberFormatterDecimalStyle)
+        }
         return CFNumberFormatterCreateStringWithNumber(nil, formatter, self._cfObject)._swiftObject
```
phausler: Probably the easier and more compatible version would be to use the locale formatter method (this is how the objc version does it)

```swift
    public func descriptionWithLocale(locale: AnyObject?) -> String {
        switch objCType.memory {
        case 0x42: // B
            fallthrough
        case 0x63: // c
            return NSString(format: "%d", locale: locale, arguments: getVaList([Int32(self.charValue)])).bridge()
        case 0x43: // C
            return NSString(format: "%u", locale: locale, arguments: getVaList([UInt32(self.unsignedCharValue)])).bridge()
        case 0x73: // s
            return NSString(format: "%hi", locale: locale, arguments: getVaList([self.shortValue])).bridge()
        case 0x53: // S
            return NSString(format: "%hu", locale: locale, arguments: getVaList([self.unsignedShortValue])).bridge()
        case 0x69: // i
            return NSString(format: "%d", locale: locale, arguments: getVaList([self.intValue])).bridge()
        case 0x49: // I
            return NSString(format: "%u", locale: locale, arguments: getVaList([self.unsignedIntValue])).bridge()
        case 0x6C: // l
            return NSString(format: "%ld", locale: locale, arguments: getVaList([self.longValue])).bridge()
        case 0x4C: // L
            return NSString(format: "%lu", locale: locale, arguments: getVaList([self.unsignedLongValue])).bridge()
        case 0x66: // f
            return NSString(format: "%0.7g", locale: locale, arguments: getVaList([self.floatValue])).bridge()
        case 0x64: // d
            return NSString(format: "%0.16g", locale: locale, arguments: getVaList([self.doubleValue])).bridge()
        case 0x71: // q
            return NSString(format: "%lld", locale: locale, arguments: getVaList([self.longLongValue])).bridge()
        case 0x51: // Q
            return NSString(format: "%llu", locale: locale, arguments: getVaList([self.unsignedLongLongValue])).bridge()
        default:
            fatalError("Bad number type")
        }
    }
    
    public override var description: String {
        return descriptionWithLocale(nil)
    }
```

parkera:
I will merge this and we can follow up on @phausler's suggestion.

PR #298, opened 2016-03-23
Implementing reversedObjectEnumeration() and reversedOrderedSet in NSOrderedSet

pushkarnk:
Added implementation of reversedObjectEnumeration() (on the lines of objectEnumerator()) and read-only property reversedOrderedSet in the NSOrderedSet class along with a few tests exercising
the same with NSOrderedSet and subclass NSMutableOrderedSet.


parkera:
@swift-ci please test

PR #299, opened 2016-03-23
Basic NSURLSession implementation

danieleggert:
I'm creating this pull request to get feedback on my implementation so far.

There are still many *TODOs* in the code, but it’s far along enough to evaluate the architecture and design I have chosen for this task. I’ve tried to outline quite a bit in the code itself in form of comments. Please let me know:

* which parts need more explanation
* which parts should be redesigned
* which parts need more testing

The `NSURLSession.swift` file has a **NSURLSession API implementation overview**.

As for testing, I'm hoping to spend quite a bit of time on building an in-process HTTP server to better test some of the desired behaviour. Particularly when testing things such as different payloads, headers, and `POST` / `PUT` requests that's the only way to get good testing.

I'd like to split the `NSURLSession+libcurl.swift` file into multiple smaller files, but that'd break with the existing style. Any thoughts?

The `CFURLSessionInterface.{c,h}` files are very verbose in order to not expose libcurl directly. This was suggested by @phausler and @rothomp3 — but I'm not sure if this is what they were hoping for.

All unimplemented code should still be marked with `NSUnimplemented()` where applicable.

This does not at all try to implement so-called *background* session which would have
to rely on a system daemon to perform HTTP requests white the app is not running (either
suspended or terminated). Such code would have to rely on IPC and the launchd equivalent
on the platform it is running on.


TestFoundation/main.swift
```diff
@@ -21,52 +21,57 @@ internal func testBundle() -> NSBundle {
...
+    testCase(TestNSHTTPURLResponse.allTests),
+    testCase(TestNSURLSession.allTests),
+    testCase(TestNSURLSessionConfiguration.allTests),
+    testCase(TestNSURLSession_libcurlHelpers.allTests),
+//    testCase(TestNSNull.allTests),
+//    testCase(TestNSUUID.allTests),
+//    testCase(TestNSValue.allTests),
+//    testCase(TestNSUserDefaults.allTests),
+//    testCase(TestNSXMLParser.allTests),
+//    testCase(TestNSXMLDocument.allTests),
```
phausler: obviously these will all have to be re-enabled before merging is considered.

danieleggert:
I should note that this relies on https://github.com/apple/swift-corelibs-libdispatch — and it has an extremely rudimentary version of NSOperationQueue to support the delegate queue on `NSURLSession`.

CoreFoundation/URL.subproj/CFURLSessionInterface.h
```diff
@@ -0,0 +1,731 @@
...
+    CFURLSession_EasyCodeSSL_ISSUER_ERROR, // CURLE_SSL_ISSUER_ERROR
+    CFURLSession_EasyCodeFTP_PRET_FAILED, // CURLE_FTP_PRET_FAILED
+    CFURLSession_EasyCodeRTSP_CSEQ_ERROR, // CURLE_RTSP_CSEQ_ERROR
+    CFURLSession_EasyCodeRTSP_SESSION_ERROR, // CURLE_RTSP_SESSION_ERROR
+    CFURLSession_EasyCodeFTP_BAD_FILE_LIST, // CURLE_FTP_BAD_FILE_LIST
+    CFURLSession_EasyCodeCHUNK_FAILED, // CURLE_CHUNK_FAILED
+    CFURLSession_EasyCodeNO_CONNECTION_AVAILABLE, // CURLE_NO_CONNECTION_AVAILABLE
+    CFURLSession_EasyCodeSSL_PINNEDPUBKEYNOTMATCH, // CURLE_SSL_PINNEDPUBKEYNOTMATCH
+    CFURLSession_EasyCodeSSL_INVALIDCERTSTATUS, // CURLE_SSL_INVALIDCERTSTATUS
+} CFURLSession_EasyCode;
```
phausler: it is worth noting that when porting to Linux these will not be exported in the same manner as enums on Darwin. We have numerous other examples that show the `#if os(OSX) || os(iOS)` re-export as constant pattern that must be done.  it might be worthwhile to modify that script that generates these to generate those as well.

phausler:
per the NSOQ stub (and other dispatch usage) we are still waiting on libdispatch to completely land; hopefully before then I will have some more things for NSOQ that will give a decent implementation for something more complete than what you have so far.

CoreFoundation/URL.subproj/CFURLSessionInterface.h
danieleggert: Makes sense. I’ll deal with that once libdispatch lands for Linux — until then it won’t compile either way.

danieleggert:
Question: What changes should be done on this before it can be merged back into master?
One thing I'm aware of: this needs libdispatch, and that's not there, yet, on Linux. What’s a reasonable workaround? Or should we wait?

Also, I'd like feedback on the fact that `NSURLSession.swift` and `NSURLSession+libcurl.swift` are now 2090 and 1741 lines respectively, and that it may make sense to break these into multiple smaller files. Any thoughts?

CoreFoundation/URL.subproj/CFURLSessionInterface.h
danieleggert: I changed this to avoid using `enum` in C.

parkera:
We're still working on dispatch; it's certainly the plan for it to be available. We should proceed as if it'll be there in the end.

I'm ok with splitting implementation into multiple files as well.

Thanks for your work on this @danieleggert.

danieleggert:
I’ve split it into these files
```
NSURLSession.swift
NSURLSessionConfiguration.swift
NSURLSessionDelegate.swift
NSURLSessionTask.swift
Configuration.swift
EasyHandle.swift
HTTPBodySource.swift
HTTPMessage.swift
MultiHandle.swift
TaskRegistry.swift
TransferState.swift
libcurlHelpers.swift
```

Foundation.xcodeproj/project.pbxproj
```diff
@@ -2150,6 +2226,7 @@
 					r,
 					r,
 				);
+				OTHER_SWIFT_FLAGS = "-DDEPLOYMENT_ENABLE_LIBDISPATCH";
```
parkera: We should probably leave this out until everything is ready to go right?

Foundation.xcodeproj/xcshareddata/xcschemes/TestFoundation.xcscheme
```diff
@@ -11,8 +11,7 @@
             buildForRunning = "YES"
             buildForProfiling = "YES"
             buildForArchiving = "YES"
-            buildForAnalyzing = "YES"
-            hideIssues = "NO">
+            buildForAnalyzing = "YES">
```
parkera: Can we omit the changes to the scheme?

Foundation.xcodeproj/xcshareddata/xcschemes/TestFoundation.xcscheme
```diff
@@ -92,6 +89,18 @@
...
+            key = "NSURLSessionDebugLibcurl"
+            value = ""
+            isEnabled = "YES">
+         </EnvironmentVariable>
+         <EnvironmentVariable
+            key = "NSURLSessionDebug"
+            value = ""
+            isEnabled = "YES">
+         </EnvironmentVariable>
+      </EnvironmentVariables>
```
parkera: These should probably be removed from the scheme as well.

parkera:
I'd like to double check a few things:

1. Is this disabled by default on Linux, since dispatch is not yet integrated?
2. I'm unsure about enabling by default on Darwin. I suppose that's fine, since it's for our own purposes.
3. The tests should only run if it's enabled (obviously).
4. Can we remove some of the scheme settings?

Once we're set, we'll squash merge this in so we can work on it iteratively after that (including enabling on Linux).

danieleggert:
@parkera This all assumes that libdispatch is available on Linux — we’ll have to hold off until then. I’ll do a few more tweaks to this tomorrow and the day after and then squash everything into a single commit.

danieleggert:
It’s all squash into a single commit now, but I might keep working on this.

danieleggert:
Note that this includes #306 and #287

czechboy0:
@danieleggert Conflicts :(

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -0,0 +1,747 @@
...
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file contains wrappes / helpers to import libcurl into Swift.
```
dtorres: typo here (wrappers)

danieleggert:
I’ll make sure conflicts get resolved once libdispatch is working.

mominul:
@danieleggert Is this pull request ready to merge?

explicitcall:
any updates on this PR?

parkera:
@danieleggert Thanks for your work on this. I believe it's been merged as part of the later work.

parkera:
It's in Swift 3 actually. Give it a shot and let us know on the mailing lists if you have any problems.

seabaylea:
@sgharms The main content went into Swift 3, and there's been some addition bug fixes in the 3.0.1 previews. There's still some missing function - particularly things like task cancellation. Having people "kick the tires" and provide tests for anything that doesn't work would be really useful.

sgharms:
@seabaylea 

My apologies for slow reply here. I went on vacation 😎 .

Let me admit that there's a great chance I'm exposing my **profound** ignorance here, so I hope I'll not embarrass myself too much.

## My Environment

```text
➜  octopress git:(master) ✗ swift --version
Swift version 3.0 (swift-3.0-RELEASE)
Target: x86_64-unknown-linux-gnu
➜  octopress git:(master) ✗ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.1 LTS"
```

## Code


```swift
        import Foundation
        let config = URLSessionConfiguration.default // Session Configuration
        let session = URLSession(configuration: config) // Load configuration into Session
        let url = URL(string: "http://apple.com")!
        
        let task = session.dataTask(with: url, completionHandler: {
            (data, response, error) in
            print("Callback ran")
            
        })
        task.resume()
        
        print("End of line")
```

## Error

I never see "Callback ran"

## Conclusion

Based on the stack trace it would seem to me that using Swift to accomplish basic HTTP operations is not supported on my platform.

sgharms:
As supplement, I asked someone with 3.0 Mac to run the same code and got the following:

screen shot 2016-10-26 at 4 30 46 pm

dtorres:
@sgharms I am assuming you are running this in the CLI in linux. well, the request doesn't get fired because the program has already exited. You need to maintain the program running until (at least) the request has had a chance to run to see the results. Look into NSRunLoop or dispatch_main(). At least one of them should be available in Linux's Swift environment

sgharms:
Ah! OK. I had wrongly assumed that there was a default run loop that would make sure all the spawned tasks closed. Thanks.

PR #300, opened 2016-03-23
Fix failure in setter for NSURLComponents.port

seabaylea:
Once you've created a `NSURLComponents` object, setting the `port` value has no effect. This is caused by why looks like a copy/paste issue in `_CFURLComponentsSetPort` where the code is calling `_passwordComponentValid` rather than `_portComponentValid`

I've fixed this, and added a test for the setter. Note that the test case sets the expected string explicitly rather than concatenating the `urlString` and the `port` value. This is because of the issue that PR #296 fixes - when converting the port to a string it currently localises it into 8,080.


parkera:
Thanks!

PR #303, opened 2016-03-31
Basic usage of ObjectiveCBridgeable protocol on Linux

dgrove-oss:
Add extensions to Bridgeable types to define
their conformance to the ObjectiveCBridgeable
protocol in terms of the existing bridging
support.  Since the compiler understands the
semantics of ObjectiveCBridgeable this allows
'as' conversions to be performed to/from various
NS types on platforms without ObjectiveC.

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -34,6 +34,44 @@ extension Array : _ObjectTypeBridgeable {
     }
 }
 
+extension Array : _ObjectiveCBridgeable {
+
+  public static func _isBridgedToObjectiveC() -> Bool {
+    return Swift._isBridgedToObjectiveC(Element.self)
+  }
+
+  @_semantics("convertToObjectiveC")
+  public func _bridgeToObjectiveC() -> NSArray {
+    return _bridgeToObject()
+  }
+
+  public static func _forceBridgeFromObjectiveC(
+    _ source: NSArray,
+    result: inout Array?
+  ) {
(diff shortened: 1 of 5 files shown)


dgrove-oss:
Relies on swift pull request https://github.com/apple/swift/pull/1994 which makes the protocol available.

parkera:
Thanks Dave. Let's hold on to this and make sure the other PR goes through.

phausler:
I guess the other followup part is to test via removing the calls to `.bridge()` and replace them with `as` conversions. If that works then we should see if the implicit conversions will work too (giving us source compatibility with struct <-> object conversion)

dgrove-oss:
Agreed.  I can start going through the tests.  I can structure either as a separate pull request or as an addendum to this one.  Either is fine with me.

phausler:
I think it may be best to have them separate just in case we get compiler crashes from the as conversion

dgrove-oss:
Test suite update in #310

dgrove-oss:
rebased to master as of 4/25; updated for Swift change for first argument labels; squashed to single commit.

parkera:
I'm going to follow up on this with some swift compiler engineers and get back to you on it. The recent back and forth with implicit bridging going away has confused things...

dgrove-oss:
ok, thanks Tony.   I'm happy with any outcome.  If we need this, I will rebase it.  If we don't need it after all, no problem...I learned a few things doing it.

dgrove-oss:
not needed; closing.

PR #304, opened 2016-04-05
Update for SE-0055: Making pointer nullability explicit.

jrose-apple:
See apple/swift#1878. Note: this needs a coordinated merge, i.e. it must go in ASAP after that request!

@phausler or someone else, would you mind reviewing? I tried to keep APIs matching the nullability audit of the Darwin Foundation, and attempted to change as little behavior as possible besides that, but there were definitely a few rough spots.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -54,7 +54,7 @@ struct _NSObjectBridge {
 struct _NSArrayBridge {
     CFIndex (*_Nonnull count)(CFTypeRef obj);
     _Nonnull CFTypeRef (*_Nonnull objectAtIndex)(CFTypeRef obj, CFIndex index);
-    void (*_Nonnull getObjects)(CFTypeRef array, CFRange range, CFTypeRef _Nonnull * _Nonnull values);
+    void (*_Nonnull getObjects)(CFTypeRef array, CFRange range, CFTypeRef _Nullable *_Nonnull values);
 };
 
 struct _NSMutableArrayBridge {
@@ -73,11 +73,11 @@ struct _NSDictionaryBridge {
     CFIndex (*countForKey)(CFTypeRef dictionary, CFTypeRef key);
     bool (*containsKey)(CFTypeRef dictionary, CFTypeRef key);
     _Nullable CFTypeRef (*_Nonnull objectForKey)(CFTypeRef dictionary, CFTypeRef key);
-    bool (*_getValueIfPresent)(CFTypeRef dictionary, CFTypeRef key, CFTypeRef _Nonnull * _Nullable value);
+    bool (*_getValueIfPresent)(CFTypeRef dictionary, CFTypeRef key, CFTypeRef _Nullable *_Nullable value);
     CFIndex (*__getValue)(CFTypeRef dictionary, CFTypeRef value, CFTypeRef key);
     bool (*containsObject)(CFTypeRef dictionary, CFTypeRef value);
     CFIndex (*countForObject)(CFTypeRef dictionary, CFTypeRef value);
-    void (*getObjects)(CFTypeRef dictionary, CFTypeRef _Nonnull * _Nonnull valuebuf, CFTypeRef _Nonnull * _Nonnull keybuf);
+    void (*getObjects)(CFTypeRef dictionary, CFTypeRef _Nullable *_Nullable valuebuf, CFTypeRef _Nullable *_Nullable keybuf);
     void (*__apply)(CFTypeRef dictionary, void (*applier)(CFTypeRef key, CFTypeRef value, void *context), void *context);
 };
 
@@ -104,7 +104,7 @@ struct _NSStringBridge {
     CFIndex (*length)(CFTypeRef str);
     UniChar (*characterAtIndex)(CFTypeRef str, CFIndex idx);
     void (*getCharacters)(CFTypeRef str, CFRange range, UniChar *buffer);
-    CFIndex (*__getBytes)(CFTypeRef str, CFStringEncoding encoding, CFRange range, uint8_t *buffer, CFIndex maxBufLen, CFIndex *usedBufLen);
+    CFIndex (*__getBytes)(CFTypeRef str, CFStringEncoding encoding, CFRange range, uint8_t *_Nullable buffer, CFIndex maxBufLen, CFIndex *_Nullable usedBufLen);
     const char *_Nullable (*_Nonnull _fastCStringContents)(CFTypeRef str);
     const UniChar *_Nullable (*_Nonnull _fastCharacterContents)(CFTypeRef str);
     bool (*_getCString)(CFTypeRef str, char *buffer, size_t len, UInt32 encoding);
@@ -123,7 +123,7 @@ struct _NSMutableStringBridge {
 
 struct _NSXMLParserBridge {
     _CFXMLInterface _Nullable (*_Nonnull currentParser)();
-    _CFXMLInterfaceParserInput _Nonnull (*_Nonnull _xmlExternalEntityWithURL)(_CFXMLInterface interface, const char *url, const char * identifier, _CFXMLInterfaceParserContext context, _CFXMLInterfaceExternalEntityLoader originalLoaderFunction);
+    _CFXMLInterfaceParserInput _Nullable (*_Nonnull _xmlExternalEntityWithURL)(_CFXMLInterface interface, const char *url, const char * identifier, _CFXMLInterfaceParserContext context, _CFXMLInterfaceExternalEntityLoader originalLoaderFunction);
     
     _CFXMLInterfaceParserContext _Nonnull (*_Nonnull getContext)(_CFXMLInterface ctx);
     
@@ -131,7 +131,7 @@ struct _NSXMLParserBridge {
     int (*isStandalone)(_CFXMLInterface ctx);
     int (*hasInternalSubset)(_CFXMLInterface ctx);
     int (*hasExternalSubset)(_CFXMLInterface ctx);
-    _CFXMLInterfaceEntity _Nonnull (*_Nonnull getEntity)(_CFXMLInterface ctx, const unsigned char *name);
+    _CFXMLInterfaceEntity _Nullable (*_Nonnull getEntity)(_CFXMLInterface ctx, const unsigned char *name);
     void (*notationDecl)(_CFXMLInterface ctx,
                          const unsigned char *name,
                          const unsigned char *publicId,
@@ -156,16 +156,16 @@ struct _NSXMLParserBridge {
     void (*endDocument)(_CFXMLInterface ctx);
     void (*startElementNs)(_CFXMLInterface ctx,
                            const unsigned char *localname,
-                           const unsigned char *prefix,
+                           const unsigned char *_Nullable prefix,
                            const unsigned char *URI,
                            int nb_namespaces,
-                           const unsigned char *_Nonnull *_Nonnull namespaces,
+                           const unsigned char *_Nullable *_Nonnull namespaces,
                            int nb_attributes,
                            int nb_defaulted,
-                           const unsigned char *_Nonnull *_Nonnull attributes);
+                           const unsigned char *_Nullable *_Nonnull attributes);
     void (*endElementNs)(_CFXMLInterface ctx,
                          const unsigned char *localname,
-                         const unsigned char *prefix,
+                         const unsigned char *_Nullable prefix,
                          const unsigned char *URI);
     void (*characters)(_CFXMLInterface ctx,
                        const unsigned char *ch,
@@ -239,7 +239,7 @@ extern CFWriteStreamRef _CFWriteStreamCreateFromFileDescriptor(CFAllocatorRef al
 extern _Nullable CFDateRef CFCalendarCopyGregorianStartDate(CFCalendarRef calendar);
 extern void CFCalendarSetGregorianStartDate(CFCalendarRef calendar, CFDateRef date);
 
-CF_PRIVATE CF_EXPORT char *_Nonnull*_Nonnull _CFEnviron(void);
+CF_PRIVATE CF_EXPORT char *_Nullable *_Nonnull _CFEnviron(void);
 
 CF_EXPORT void CFLog1(CFLogLevel lev, CFStringRef message);
 
--- a/Foundation/NSCoder.swift
+++ b/Foundation/NSCoder.swift
@@ -130,12 +130,15 @@ public class NSCoder : NSObject {
         encodeValueOfObjCType("[\(count)\(String(cString: type))]", at: array)
     }
     
-    public func encodeBytes(_ byteaddr: UnsafePointer<Void>, length: Int) {
+    public func encodeBytes(_ byteaddr: UnsafePointer<Void>?, length: Int) {
         var newLength = UInt32(length)
         withUnsafePointer(&newLength) { (ptr: UnsafePointer<UInt32>) -> Void in
             encodeValueOfObjCType("I", at: ptr)
         }
-        encodeArrayOfObjCType("c", count: length, at: byteaddr)
+        var empty: [Int8] = []
+        withUnsafePointer(&empty) {
+            encodeArrayOfObjCType("c", count: length, at: byteaddr ?? UnsafePointer($0))
+        }
     }
     
     public func decodeObject() -> AnyObject? {
@@ -251,7 +254,7 @@ public class NSCoder : NSObject {
         NSRequiresConcreteImplementation()
     }
     
-    public func decodeBytesForKey(_ key: String, returnedLength lengthp: UnsafeMutablePointer<Int>) -> UnsafePointer<UInt8> { // returned bytes immutable!
+    public func decodeBytesForKey(_ key: String, returnedLength lengthp: UnsafeMutablePointer<Int>?) -> UnsafePointer<UInt8>? { // returned bytes immutable!
         NSRequiresConcreteImplementation()
     }
     
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -75,20 +75,21 @@ public class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     private var _base = _CFInfo(typeID: CFDataGetTypeID())
     private var _length: CFIndex = 0
     private var _capacity: CFIndex = 0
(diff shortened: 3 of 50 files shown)


Foundation/NSCoder.swift
```diff
@@ -126,11 +126,12 @@ public class NSCoder : NSObject {
         encodeObject(object)
     }
     
-    public func encodeArrayOfObjCType(type: UnsafePointer<Int8>, count: Int, at array: UnsafePointer<Void>) {
-        encodeValueOfObjCType("[\(count)\(String(cString: type))]", at: array)
+    public func encodeArrayOfObjCType(type: UnsafePointer<Int8>, count: Int, at array: UnsafePointer<Void>?) {
+        // FIXME: What about empty arrays at null addresses?
+        encodeValueOfObjCType("[\(count)\(String(cString: type))]", at: array!)
```
jrose-apple: I wasn't sure what to do in this case; this method can be called from `encodeBytes(_:length:)` with a null array. Maybe they should both have a separate helper function?

Foundation/NSData.swift
```diff
@@ -75,20 +75,20 @@ public class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
...
             return unsafeBitCast(self, to: CFType.self)
         } else {
-            return CFDataCreate(kCFAllocatorSystemDefault, unsafeBitCast(self.bytes, to: UnsafePointer<UInt8>.self), self.length)
+            return CFDataCreate(kCFAllocatorSystemDefault, UnsafePointer<UInt8>(self.bytes), self.length)
         }
     }
     
     public override required convenience init() {
-        self.init(bytes: nil, length: 0, copy: false, deallocator: nil)
+        self.init(bytes: nil, length: 0, copy: true, deallocator: nil)
```
jrose-apple: This allocation is unfortunate, but necessary to preserve the property that NSData always has non-null `bytes`. There are several alternate solutions here, but this one requires the fewest changes elsewhere.

Foundation/NSData.swift
```diff
@@ -75,20 +75,20 @@ public class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
...
     private var _deallocHandler: _NSDataDeallocator? = _NSDataDeallocator() // for Swift
-    private var _bytes: UnsafeMutablePointer<UInt8> = nil
+    private var _bytes: UnsafeMutablePointer<UInt8>? = nil
     
     internal var _cfObject: CFType {
         if self.dynamicType === NSData.self || self.dynamicType === NSMutableData.self {
             return unsafeBitCast(self, to: CFType.self)
         } else {
-            return CFDataCreate(kCFAllocatorSystemDefault, unsafeBitCast(self.bytes, to: UnsafePointer<UInt8>.self), self.length)
+            return CFDataCreate(kCFAllocatorSystemDefault, UnsafePointer<UInt8>(self.bytes), self.length)
```
jrose-apple: I left most of these pointer-to-pointer conversions alone, but I strongly recommend switching over to using converting initializers, which will now preserve pointer nullability through the conversion (and which are shorter and less scary than `unsafeBitCast`).

Foundation/NSData.swift
```diff
@@ -104,21 +104,21 @@ public class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
...
     
-    internal init(bytes: UnsafeMutablePointer<Void>, length: Int, copy: Bool, deallocator: ((UnsafeMutablePointer<Void>, Int) -> Void)?) {
+    internal init(bytes: UnsafeMutablePointer<Void>?, length: Int, copy: Bool, deallocator: ((UnsafeMutablePointer<Void>, Int) -> Void)?) {
         super.init()
         let options : CFOptionFlags = (self.dynamicType == NSMutableData.self) ? __kCFMutable | __kCFGrowable : 0x0
         if copy {
             _CFDataInit(unsafeBitCast(self, to: CFMutableData.self), options, length, UnsafeMutablePointer<UInt8>(bytes), length, false)
-            if let handler = deallocator {
-                handler(bytes, length)
+            if let handler = deallocator, allocatedBytes = bytes {
```
jrose-apple: I believe this is correct based on the behavior in the deinitializer.

Foundation/NSDictionary.swift
```diff
@@ -122,7 +122,8 @@ public class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCodin
     }
     
     public override convenience init() {
-        self.init(objects: nil, forKeys: nil, count: 0)
+        // FIXME: This isn't really safe. Either the required initializer handles optional pointers, or it doesn't.
+        self.init(objects: unsafeBitCast(nil as UnsafePointer<AnyObject>?, to: UnsafePointer<AnyObject>.self), forKeys: unsafeBitCast(nil as UnsafePointer<NSObject>?, to: UnsafePointer<NSObject>.self), count: 0)
```
jrose-apple: Here I didn't have a great answer; another option would be to pass empty Swift arrays, which are at least fairly cheap to construct. (They all wrap a preallocated global constant.)

jrose-apple: I want to note that this is NOT SAFE in the general case. I don't even want to promise that this won't cause problems in optimized builds today.

Foundation/NSFileManager.swift
```diff
@@ -590,10 +585,10 @@ public class NSFileManager : NSObject {
         return self.fileExistsAtPath(path, isDirectory: nil)
     }
     
-    public func fileExistsAtPath(path: String, isDirectory: UnsafeMutablePointer<ObjCBool>) -> Bool {
+    public func fileExistsAtPath(path: String, isDirectory: UnsafeMutablePointer<ObjCBool>?) -> Bool {
         var s = stat()
         if lstat(path, &s) >= 0 {
-            if isDirectory != nil {
+            if let isDirectory = isDirectory {
```
jrose-apple: I introduced a few of these `if let x = x`. This has been a controversial idiom in the Swift community, so please let me know if you'd prefer alternate names for the unwrapped value every time.

Foundation/NSLocale.swift
```diff
@@ -13,9 +13,9 @@ import CoreFoundation
 public class NSLocale : NSObject, NSCopying, NSSecureCoding {
     typealias CFType = CFLocale
     private var _base = _CFInfo(typeID: CFLocaleGetTypeID())
```
jrose-apple: A general warning for the future: Swift property layout is not guaranteed to be the same as C property layout. I don't actually know of a safe way to do this, but the safest way there is today is to declare all the properties in a C struct and have a single property Swift-side of that type.

Foundation/NSObjCRuntime.swift
```diff
@@ -116,24 +116,21 @@ internal func _NSGetSizeAndAlignment(type: _NSSimpleObjCType,
...
     
     if !_NSGetSizeAndAlignment(type, &size, &align) {
-        return nil
-    }
-    
-    if sizep != nil {
-        sizep.pointee = size
+        // FIXME: This used to return nil, but the corresponding Darwin
+        // implementation is defined as returning a non-optional value.
+        fatalError("invalid type encoding")
```
jrose-apple: Here's a place where the Darwin API disagrees with the open-source implementation. The one call I saw, though, didn't even check the result.

Foundation/NSOrderedSet.swift
```diff
@@ -351,8 +351,12 @@ public class NSMutableOrderedSet : NSOrderedSet {
         }
     }
 
-    public init(capacity numItems: Int) {
-        super.init(objects: nil, count: 0)
+    public convenience init(capacity numItems: Int) {
+        self.init(array: [])
+    }
+
+    public override init(objects: UnsafePointer<AnyObject?>, count cnt: Int) {
```
jrose-apple: This designated initializer was missing, which meant the other convenience initializers were silently missing as well. That probably points to some potential area of improvement in the compiler.

Foundation/NSSet.swift
```diff
@@ -43,12 +43,12 @@ extension Set : _ObjectTypeBridgeable {
...
             let cnt = CFSetGetCount(cf)
             
-            let objs = UnsafeMutablePointer<UnsafePointer<Void>>(allocatingCapacity: cnt)
+            let objs = UnsafeMutablePointer<UnsafePointer<Void>?>(allocatingCapacity: cnt)
             
             CFSetGetValues(cf, objs)
             
             for idx in 0..<cnt {
-                let obj = unsafeBitCast(objs.advanced(by: idx), to: AnyObject.self)
+                let obj = unsafeBitCast(objs.advanced(by: idx).pointee!, to: AnyObject.self)
```
jrose-apple: This was a straight-up bug. More tests needed!

Foundation/NSTask.swift
```diff
@@ -40,24 +40,24 @@ private var managerThreadRunLoopIsRunningCondition = NSCondition()
 internal let kCFSocketDataCallBack = CFSocketCallBackType.dataCallBack.rawValue
 #endif
 
-private func emptyRunLoopCallback(context : UnsafeMutablePointer<Void>) -> Void {}
+private func emptyRunLoopCallback(context : UnsafeMutablePointer<Void>!) -> Void {}
 
 
 // Retain method for run loop source
-private func runLoopSourceRetain(pointer : UnsafePointer<Void>) -> UnsafePointer<Void> {
+private func runLoopSourceRetain(pointer : UnsafePointer<Void>!) -> UnsafePointer<Void>! {
```
jrose-apple: This is an unfortunate consequence of the compiler being strict about `@convention(c)` callback types.

Foundation/NSURL.swift
```diff
@@ -369,7 +369,10 @@ public class NSURL : NSObject, NSSecureCoding, NSCopying {
             return UnsafePointer(_fsrBuffer)
         }
 
-        return nil
+        // FIXME: This used to return nil, but the corresponding Darwin
+        // implementation is marked as non-nullable.
+        fatalError("URL cannot be expressed in the filesystem representation;" +
+                   "use getFileSystemRepresentation to handle this case")
```
jrose-apple: Here's another one I wasn't sure what to do with.

Foundation/NSUUID.swift
```diff
@@ -32,11 +32,7 @@ public class NSUUID : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
     
     public init(UUIDBytes bytes: UnsafePointer<UInt8>) {
-        if (bytes != nil) {
-            memcpy(unsafeBitCast(buffer, to: UnsafeMutablePointer<Void>.self), UnsafePointer<Void>(bytes), 16)
-        } else {
-            memset(unsafeBitCast(buffer, to: UnsafeMutablePointer<Void>.self), 0, 16)
-        }
+        memcpy(unsafeBitCast(buffer, to: UnsafeMutablePointer<Void>.self), UnsafePointer<Void>(bytes), 16)
```
jrose-apple: This one was strange. There's an explicit test for the removed behavior, but the Darwin API declares that the parameter is non-optional.

Foundation/NSCoder.swift
phausler: this is not nullable

Foundation/NSXMLNode.swift
```diff
@@ -229,7 +226,7 @@ public class NSXMLNode : NSObject, NSCopying {
         @abstract Returns an element, attribute, entity, or notation DTD node based on the full XML string.
     */
     public var kind: NSXMLNodeKind  {
-        guard _xmlNode != nil else { return .InvalidKind }
+//        guard _xmlNode != nil else { return .InvalidKind }
```
jrose-apple: I'm not sure what to do here. Pretty much nothing else in NSXMLNode can handle `_xmlNode` being `nil`. I suppose we could have asserting and non-asserting properties?

Foundation/NSData.swift
phausler: that is really bad... I don't understand why that requires any sort of change; if so we can pass in a non null placeholder dummy address instead

Foundation/NSValue.swift
```diff
@@ -96,7 +96,7 @@ public class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
         return NSSpecialValue._typeFromObjCType(type) != nil
     }
     
-    public convenience required init(bytes value: UnsafePointer<Void>, objCType type: UnsafePointer<Int8>) {
+    public convenience required init(bytes value: UnsafePointer<Void>?, objCType type: UnsafePointer<Int8>) {
```
jrose-apple: I'm not very happy with this. Do we care about zero-byte types enough to make this initializer optional?

Foundation/NSData.swift
phausler: nope, that is incorrect behavior, if the passed pointer to the deallocator is null it is perfectly valid in the objc sense

Foundation/NSSpecialValue.swift
```diff
@@ -10,7 +10,7 @@
 internal protocol NSSpecialValueCoding {
     static func objCType() -> String
     
-    init(bytes value: UnsafePointer<Void>)
+    init(bytes value: UnsafePointer<Void>?)
```
jrose-apple: Are there any "special values" with a zero-byte encoding? Or should this just be non-optional?

Foundation/NSDictionary.swift
phausler: can we use placeholder memory here too? e.g. a static UnsafePointer<Void> bit pattern or something like NSData and promise that it will never be de-referenced internally? We are breaking subclassing patterns by doing this btw; internally the values are nullable and externally vended they are not.

Foundation/NSFileManager.swift
phausler: how is this associated with the commit? id really like to just keep this as isolated as possible

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -73,11 +73,11 @@ struct _NSDictionaryBridge {
     CFIndex (*countForKey)(CFTypeRef dictionary, CFTypeRef key);
     bool (*containsKey)(CFTypeRef dictionary, CFTypeRef key);
     _Nullable CFTypeRef (*_Nonnull objectForKey)(CFTypeRef dictionary, CFTypeRef key);
-    bool (*_getValueIfPresent)(CFTypeRef dictionary, CFTypeRef key, CFTypeRef _Nonnull * _Nullable value);
+    bool (*_getValueIfPresent)(CFTypeRef dictionary, CFTypeRef key, CFTypeRef _Nullable *_Nonnull value);
```
phausler: from a pedantic standpoint iirc this is doubly nullable

CoreFoundation/Collections.subproj/CFData.c
```diff
@@ -361,7 +361,10 @@ CF_SWIFT_EXPORT void _CFDataInit(CFMutableDataRef memory, CFOptionFlags flags, C
         }
         
         __CFDataSetNeedsToZero(memory, !cleared);
-        CFDataReplaceBytes(memory, CFRangeMake(0, 0), bytes, length);
+        // Avoid copying from null pointers.
+        if (length > 0) {
+            CFDataReplaceBytes(memory, CFRangeMake(0, 0), bytes, length);
```
phausler: lets not change this... this is extraneous to the pull request and would require synchronization with CF.

Foundation/NSData.swift
jrose-apple: Hm. That's problematic, because the deallocator normally isn't required to handle a null pointer.

Foundation/NSFileManager.swift
jrose-apple: There are a lot of conversions from `!= nil` tests to `if let`, because otherwise you end up with `!` scattered around the body of the `if`. The question is really just about the names; using the same name would be the most minimal diff.

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
jrose-apple: Ah, are you permitted to use this to ask if a value is present? Okay, will change. Good catch!

CoreFoundation/Collections.subproj/CFData.c
jrose-apple: This is technically a correctness issue (the same portability issue I alluded to above). I figured since it was about knowing which pointers were nullable that it was worth including, but I'll take it out.

phausler:
This does not seem to build at all for me from the Xcode project, there are some errors in the interface for NSXMLParser. We need to have this compile for both darwin and linux

jrose-apple:
Oooops. Will get on the Darwin side today.

jrose-apple:
Fixed the Darwin side. Now doing a second pass to (a) keep the public API in sync with the SDK shipped with Xcode 7.3, and (b) make sure no behavior changes snuck in. Also planning to rebase after the update for the new parameter naming rules.

jrose-apple:
I'm sorry, this is as small as the patch is going to get. This is the minimal *nice* patch to get swift-corelibs-foundation to build (on both Linux and Darwin) under the compiler branch above. What I mean by a "nice" patch is preferring `if let` and `guard let` to putting `!` in bodies, if there was nothing else interesting going on.

There are three actual behavior changes here:

- NSUUID's initializer does not take an optional input, and so the special behavior for null pointers has been removed.

- NSNumber non-keyed decoding was logically broken; the old code would have immediately led to a null pointer dereference.

- _NSXMLParserStartElementNs had a number of logic inversions (`x == nil` instead of `x != nil`); previously these were silently ignored, but they started being caught by tests. These have been fixed.

All public APIs involving pointers ought to match their Darwin equivalents now, except for those in the following classes:
- NSAttributedString
- NSFormatter subclasses
- NSScanner
- NSStream
- `NSDivideRect(_:_:_:_:_:)` (not a class)

Foundation/NSSet.swift
jrose-apple: I've restored this to being broken in the newer commit.

Foundation/NSOrderedSet.swift
jrose-apple: I dropped this in the newer commit.

parkera:
@swift-ci please test

jrose-apple:
Unfortunately that won't work; this depends on the branch in apple/swift#1878.

parkera:
Ok, I think we decided to hold off on this for the time being then.

parkera:
We're merging this to unblock the compiler work on this feature. We may have to iterate in master if it causes unintended fallout. @jroose-apple will coordinate.

jrose-apple:
I reverted for now so that we can get all related PRs in at roughly the same time.

PR #308, opened 2016-04-07
[SE-0046] Fix Foundation build

modocache:
The changes applied for SE-0046 appear to have missed a few spots. Fix the build by naively adopting all fix-it hints.


modocache:
Oops! Looks like I should have pulled first. 😇

PR #309, opened 2016-04-08
Use XCTest's async waiting functionality instead of a hand-rolled method

briancroom:
The tests for `NSNotificationQueue` predate the introduction of swift-corelibs-xctest's own API for asynchronous testing, and instead use a private method in the test case for spinning the run loop and waiting for an event to have occurred. It would be nice to migrate to the built-in XCTest functionality for this.

I have done my best to stress-test the test suite on my machine in both OSX and Linux environments and haven't encountered a single failure, however I do still have a slight concern that this could re-introduce test flakiness because it remove the explicit locking added by @phausler in https://github.com/apple/swift-corelibs-foundation/commit/735d33bfed54157c66f7ae89823d8837643bd902. If it turns out to be problematic, I would consider it to be a shortcoming in XCTest, though, and would like to see it fixed there.

### DIFF
--- a/TestFoundation/TestNSNotificationQueue.swift
+++ b/TestFoundation/TestNSNotificationQueue.swift
@@ -213,41 +213,22 @@ class TestNSNotificationQueue : XCTestCase {
     // MARK: Private
 
     private func scheduleTimer(withInterval interval: NSTimeInterval) {
-        var isInvoked = false
+        let e = expectation(withDescription: "Timer")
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci please test

parkera:
Everything should be in place for this right?

briancroom:
@parkera Yes! Thanks to @modocache

parkera:
I'm celebrating this by using the new Squash and Merge feature.

PR #311, opened 2016-04-10
Clean up Xcode warnings

briancroom:
Currently, building `SwiftFoundation.framework` in Xcode generates a couple of distracting warnings. In an attempt to reduce pain points for contributors working on corelibs Foundation and XCTest using Xcode, these changes fix the warnings, so the all targets build without producing any issues.


parkera:
Okeydokey.

PR #313, opened 2016-04-13
Add comparator between two dates ignoring time

mauriciomeirelles:
In many situations we need a way of comparing if two dates are the same
but ignoring time. It’ll be really nice if we provide an easy way of
doing that


parkera:
This would be new API, which are are not accepting through PR at this time. We can have a discussion about this on the corelibs mailing list.

parkera:
As a side note, I don't think this implementation is going to be correct for all calendars...

mauriciomeirelles:
Ok! Thanks :)

PR #326, opened 2016-04-16
fix warnings for SE-0049 (Move @noescape ...)

scanf:
Not sure if this change is unwelcome, considering that [proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0049-noescape-autoclosure-type-attrs.md) says:

> The Swift 3 migrator should move these over, and has the information it needs to do a perfect migration in this case.

Anyways the warnings are annoying and I wanted to silence them.


parkera:
Thanks! Not unwelcome at all. =)

PR #327, opened 2016-04-16
[gardening] remove unused variable in TestNSURLRequest

scanf:


parkera:
I think I fixed this in another merge.

PR #329, opened 2016-04-17
Parse non-floating point numbers into Int instead of Double

tomquist:
Fix for my bug report https://bugs.swift.org/browse/SR-1250


parkera:
@swift-ci please test

parkera:
Hm, interesting. I think eventually we'll need to figure out how to type this more strongly (as with plists), but I think this implementation does match the Objective-C one more closely.

PR #333, opened 2016-04-18
Fix NSData to use cross platform write function

mbvreddy:
Fixed it in same way as Darwin/Glibc is imported

`#if os(OSX) || os(iOS)
    bytesWritten = Darwin.write(fd, buf.advanced(by: length - bytesRemaining), bytesRemaining)
#elseif os(Linux)
    bytesWritten = Glibc.write(fd, buf.advanced(by: length - bytesRemaining), bytesRemaining)
#endif
`

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -411,7 +411,11 @@ extension NSData {
         while bytesRemaining > 0 {
             var bytesWritten : Int
             repeat {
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci please test

parkera:
Looks reasonable.

rzhikharevich:
Won't this break on tvOS?

parkera:
Technically this code only really runs on OS X and linux in the first place.

PR #297, opened 2016-03-23
Implement basic initializers of NSAttributedString

eyeplum:
Implemented and tested the following methods (and properties):

- `public var string: String`
- `public var length: Int`
- `public init(string str: String)`
- `public init(string str: String, attributes attrs: [String : AnyObject]?)`
- `public init(attributedString attrStr: NSAttributedString)`

The implementation is based on CFAttributedString.

PR #334, opened 2016-04-19
Basic implementations of NSAttributedString

eyeplum:
Different from #297, this time I'm trying to implement the NSAttributedString to be a is-a relationship to CFAttributedString.

### DIFF
--- a/Foundation/NSAttributedString.swift
+++ b/Foundation/NSAttributedString.swift
@@ -12,7 +12,7 @@ import CoreFoundation
 public class NSAttributedString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     
     private let _cfinfo = _CFInfo(typeID: CFAttributedStringGetTypeID())
-    private let _string: String
+    private let _string: NSString
     private let _attributeArray: CFRunArrayRef
     
     public required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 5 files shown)


Foundation/NSAttributedString.swift
```diff
@@ -7,9 +7,14 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+import CoreFoundation
 
 public class NSAttributedString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     
+    private let _cfinfo = _CFInfo(typeID: CFAttributedStringGetTypeID())
+    private let _string: NSString
```
eyeplum: I tried to use `private let _string: String` here, but encountered EXC_BAD_ACCESS at CFString function calls.
I didn't figure out why, but found out that with a `NSString`, everything seemed to be fine.

TestFoundation/TestNSAttributedString.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+    
+    func test_initWithString() {
+        let string = "Lorem 😀 ipsum dolor sit amet, consectetur adipiscing elit. ⌘ Phasellus consectetur et sem vitae consectetur. Nam venenatis lectus a laoreet blandit. ಠ_ರೃ"
+        let attrString = NSAttributedString(string: string)
+        XCTAssertEqual(attrString.string, string)
+        XCTAssertEqual(attrString.length, string.utf16Count)
+        
+        var range = NSRange()
+        let attrs = attrString.attributesAtIndex(0, effectiveRange: &range)
```
eyeplum: It seems that this will cause a compiler error with current swift toolchain:

```
let attrs = attrString.attributesAtIndex(0, effectiveRange: nil)
```

But the darwin counterpart can actually accept `nil` as a valid argument.
Not sure if a change is needed in the function signature.

TestFoundation/TestNSAttributedString.swift
```diff
@@ -0,0 +1,65 @@
...
+    func test_initWithStringAndAttributes() {
+        let string = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus consectetur et sem vitae consectetur. Nam venenatis lectus a laoreet blandit."
+        let attributes: [String : AnyObject] = ["attribute.placeholder.key" : "attribute.placeholder.value" as NSString]
+        
+        let attrString = NSAttributedString(string: string, attributes: attributes)
+        XCTAssertEqual(attrString.string, string)
+        XCTAssertEqual(attrString.length, string.utf16Count)
+        
+        var range = NSRange()
+        let attrs = attrString.attributesAtIndex(0, effectiveRange: &range)
```
eyeplum: Same as line 38.

Foundation/NSAttributedString.swift
modocache: Do you think https://bugs.swift.org/browse/SR-1129 could be related?

eyeplum: SR-1129 seems to be a compile time error, what I encountered is a runtime error. 
So, maybe they are not related.

parkera:
Thanks!

PR #337, opened 2016-04-20
Remove the description of dropping 'NS' prefix

JustinJiaDev:
Remove " We will also drop the 'NS' prefix from all Foundation classes." according to https://github.com/apple/swift-evolution/commit/bf2e4c4bcb1ad3dd072920f33514546e75cf59bb

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ In our first year, we are not looking to make major API changes to the library. 
 
 ### API Naming and Foundation
 
-One of the goals of the Swift 3 project is [a new set of naming guidelines](https://swift.org/documentation/api-design-guidelines/). The Foundation project will soon update all of its names to match the new guidelines. We will also drop the 'NS' prefix from all Foundation classes.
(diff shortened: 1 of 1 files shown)


parkera:
We can discuss the goals of the project on the mailing list.

JustinJiaDev:
Hi! In my understand, Swift 3 are not going to drop the NS prefix. README.md was written before this change. Therefore, this sentence is no longer valid and should be removed. It will confuse people.

parkera:
It is still part of the plan, we are iterating on exactly how to do it.

ianpartridge:
Thanks for this information @parkera - do you have links to the discussion/iteration? I think there's a lot of interest in how the Foundation APIs evolve.

parkera:
Not just yet, but soon.

tomquist:
I think this proposal is one step in removing "NS" prefixes: https://github.com/apple/swift-evolution/blob/master/proposals/0069-swift-mutability-for-foundation.md

PR #338, opened 2016-04-26
Further integration of Dispatch into Foundation

seabaylea:
This makes some minor changes that gets us closer to having Dispatch built and included in CoreFoundation and Foundation, including:
* Switching CFRunLoop to use the new `4CF` functions in libdispatch  
Note that these APIs are only available in the `experimental/foundation' branch of libdispatch.
* Adding a basic implementation for `pthread_main_np()`  
There isn't an implementation in Glibc and there isn't a shim today. We should potentially move the implementation to the Glibc overlay. 
* Updated the build scripts to include Dispatch in Foundation as well as CoreFoundation  
This is still commented out as there's a number of compile errors in NSOperation and the incoming NSURLSession code - we should fix those up before uncommenting this code.

The changes should allow anyone to build a toolchain with Dispatch enabled by:
1. Extracting the `experimental/foundation` branch of `swift-corelibs-libdispatch`
2. Uncommenting the code in `build.py`
3. Running a build with the `libdispatch` and `install-libdispatch` flags added

### DIFF
--- a/build.py
+++ b/build.py
@@ -64,14 +64,18 @@ if "XCTEST_BUILD_DIR" in Configuration.current.variables:
 		'-I/usr/include/libxml2'
 	]
 
-# Disable until changes are merged into dispatch.
+# Configure use of Dispatch in CoreFoundation and Foundation if libdispatch is being built
 #if "LIBDISPATCH_SOURCE_DIR" in Configuration.current.variables:
-#        foundation.CFLAGS += " "+" ".join([
-#                '-DDEPLOYMENT_ENABLE_LIBDISPATCH',
-#                '-I'+Configuration.current.variables["LIBDISPATCH_SOURCE_DIR"],
-#                '-I'+Configuration.current.variables["LIBDISPATCH_BUILD_DIR"]+'/tests'  # for include of dispatch/private.h in CF
-#        ])
-
+#	foundation.CFLAGS += " "+" ".join([
+#		'-DDEPLOYMENT_ENABLE_LIBDISPATCH',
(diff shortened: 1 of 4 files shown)


build.py
```diff
@@ -64,14 +64,18 @@
...
-# Disable until changes are merged into dispatch.
-#if "LIBDISPATCH_SOURCE_DIR" in Configuration.current.variables:
-#        foundation.CFLAGS += " "+" ".join([
-#                '-DDEPLOYMENT_ENABLE_LIBDISPATCH',
-#                '-I'+Configuration.current.variables["LIBDISPATCH_SOURCE_DIR"],
-#                '-I'+Configuration.current.variables["LIBDISPATCH_BUILD_DIR"]+'/tests'  # for include of dispatch/private.h in CF
-#        ])
-
+# Configure use of Dispatch in CoreFoundation and Foundation if libdispatch is being built
+if "LIBDISPATCH_SOURCE_DIR" in Configuration.current.variables:
```
modocache: I don't have a ton of context on the changes in this pull request, but thought I'd chime in with a tiny nit-pick: Is there a reason you uncommented this line? That's fine of course, but Python requires `if` to contain at least one expression. To do nothing, you may use `pass`:

```py
if "LIBDISPATCH_SOURCE_DIR" in Configuration.current.variables:
    pass
```

I think if you run this as-is, it would raise a syntax error of some kind.

seabaylea: Thanks for picking that up. Looks like I missed a line when I re-commented out the block as building Dispatch into Foundation currently hits compile errors in NSOperation (due to some Dispatch overlay work that needs fixing up).

PR #340, opened 2016-04-27
NSOperationQueue.operationCount always return 0

mbvreddy:
NSOperationQueue maintain different queues based on priority and a queue 'all' to maintain operations of all priorities. 

NSOperationQueue.operationCount is nothing but the no.of operations in 'all' queue. However, addOperation() doesn't add Operations to 'all' queue and thereby NSOperationQueue.operationCount is always 0.

Currently, there is no testsuite for NSOperationQueue. I've added test case for this scenario (partially, as it would require Dispatch support to add more scenarios to it).

### DIFF
--- a/Foundation/NSOperation.swift
+++ b/Foundation/NSOperation.swift
@@ -204,6 +204,7 @@ internal struct _OperationList {
     var all = [NSOperation]()
     
     mutating func insert(_ operation: NSOperation) {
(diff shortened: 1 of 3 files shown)


mbvreddy:
small correction in dequeue() method. Will push changes shortly.

mbvreddy:
done

PR #341, opened 2016-04-28
Treat JSON string parsing as a special case

argon:
`consumeStructure` should be simple with no knowledge of the special
requirements for string parsing. We'll handle it directly in `parseString`
instead.


parkera:
@swift-ci please test

PR #343, opened 2016-05-03
Fix up swift-corelibs-foundation for abolition of IUO type in SE-0054

cwillmor:
This commit fixes up swift-corelibs-foundation for when SE-0054 is merged into swift master. See apple/swift#2322. The two PRs are to be merged in conjunction.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -85,7 +85,7 @@ extension String : _ObjectTypeBridgeable {
             let cf = unsafeBitCast(x, to: CFString.self)
             let str = CFStringGetCStringPtr(cf, CFStringEncoding(kCFStringEncodingUTF8))
             if str != nil {
-                result = String(cString: str)
+                result = String(cString: str!)
             } else {
                 let length = CFStringGetLength(cf)
                 let buffer = UnsafeMutablePointer<UniChar>(allocatingCapacity: length)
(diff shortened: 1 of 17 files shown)


Foundation/NSString.swift
```diff
@@ -971,28 +971,28 @@ extension NSString {
     }
     
     public var decomposedStringWithCanonicalMapping: String {
-        let string = CFStringCreateMutable(kCFAllocatorSystemDefault, 0)
+        let string = CFStringCreateMutable(kCFAllocatorSystemDefault, 0)!
```
parkera: It's not clear to me why the result here is nullable in the first place...

Foundation.xcodeproj/project.pbxproj
```diff
@@ -340,6 +340,7 @@
 		EADE0BCB1BD15E0000C49C64 /* NSXMLNode.swift in Sources */ = {isa = PBXBuildFile; fileRef = EADE0B8D1BD15DFF00C49C64 /* NSXMLNode.swift */; };
 		EADE0BCC1BD15E0000C49C64 /* NSXMLNodeOptions.swift in Sources */ = {isa = PBXBuildFile; fileRef = EADE0B8E1BD15DFF00C49C64 /* NSXMLNodeOptions.swift */; };
 		EADE0BCD1BD15E0000C49C64 /* NSXMLParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = EADE0B8F1BD15DFF00C49C64 /* NSXMLParser.swift */; };
+		F2C623AF1CD80D9400E1A419 /* CFSocket.c in Sources */ = {isa = PBXBuildFile; fileRef = 5B5D88E01BBC9B0300234F36 /* CFSocket.c */; };
```
parkera: Why did we add this as part of this PR?

cwillmor: It was necessary for me to get the SwiftFoundation target to link. But you're right, it shouldn't have been in this PR. I'll remove this part.

Foundation/NSString.swift
cwillmor: CFStringCreateMutable appears to have IUO return type imported. Maybe it hasn't been annotated for nullability.

parkera:
Merge when ready

cwillmor:
Thanks!

PR #344, opened 2016-05-03
fix NSOperationQueue to retain operation in 'all' queue once it is dispatched

mbvreddy:
After https://github.com/apple/swift-corelibs-foundation/pull/340,  NSOperationQueue.operationCount is decremented as soon as task is dispatched for execution. This is because dequeue() method is modified to delete operation from 'all' queue.


parkera:
@swift-ci please test

PR #349, opened 2016-05-06
[NSTask] Cleanup environment after task is launched

Bouke:
Currently the environment is deallocated before the task is launched. This results in memory being overwritten, and a corrupt environment. @ddunbar suggested moving the `defer` up a scope. I've added a few tests to verify the correctness of the environment.


ddunbar:
@swift-ci please test

PR #350, opened 2016-05-07
NSFileManager.moveItem - destination path should not exist before moving

Bouke:
(cc @ddunbar)


parkera:
Nice

parkera:
@swift-ci please test

PR #352, opened 2016-05-07
Adding support for ARMv6 on linux

pj4533:
This change expands on the work by @hpux735 to support ARMv6, not just ARMv7.

While making the change I noticed the ArchSubType class (which already has a bunch of ARM architectures listed).  That might be a more robust change...but ArchSubType doesn't look like its used anywhere, so I did this quicker solution that I know works.

Thoughts on whether it is worth it to try an implementation with the `arm` ArchType, and manage the v6 or v7 in the `ArchSubType`?  It makes things more complicated (have to add a `subarch` to `Target`, etc), so maybe better just use ArchType only (as I have done here)?

### DIFF
--- a/lib/target.py
+++ b/lib/target.py
@@ -12,7 +12,7 @@ import platform
 
 class ArchType:
     UnknownArch = 0
-    arm         = 1
+    armv7       = 1
     armeb       = 2
     aarch64     = 3
     aarch64_be  = 4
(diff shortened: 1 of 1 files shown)


hpux735:
This successfully built on the un-official swift-arm CI infrastructure.  It *should* be NFC on any other arch.   LGTM.

parkera:
@swift-ci please test

parkera:
Just to make sure we didn't mess something up for the other build platforms, I'll kick off a CI run.

PR #353, opened 2016-05-09
[NSTask] Change argv[0] to full path + fixes combining stdout+stderr

Bouke:
* Helper function to reduce TestNSTask test case duplication
* Fix issue capturing stdout and stderr to same pipe + test case
* Pass complete path as argv[0]  
  This allows executables to derive their own absolute path. SwiftPM relies on this value to derive the compiler path and fake toolchain path.

### DIFF
--- a/TestFoundation/TestNSTask.swift
+++ b/TestFoundation/TestNSTask.swift
@@ -164,7 +164,7 @@ class TestNSTask : XCTestCase {
             XCTFail("Could not read stdout")
             return
         }
-        XCTAssertEqual(string, "cat: invalid_file_name: No such file or directory\n")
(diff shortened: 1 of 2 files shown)


TestFoundation/TestNSTask.swift
```diff
@@ -27,7 +27,7 @@ class TestNSTask : XCTestCase {
                    ("test_pipe_stdin", test_pipe_stdin),
                    ("test_pipe_stdout", test_pipe_stdout),
                    ("test_pipe_stderr", test_pipe_stderr),
-                   ("test_file_stdout", test_file_stdout),
```
ddunbar: Did you mean to remove this line?

Bouke: No, will put it back in.

parkera:
@swift-ci please test

PR #354, opened 2016-05-09
Creation of a serial queue in NSOperationQueue fails on setting property maxConcurrentOperationCount to "1"

mamabusi:
Recreate:
----------------------------------------------------------------------------------------------------

     let operation1 : NSBlockOperation = NSBlockOperation (block: {
            sleep(1)
            print("Operation1")
        })
        let operation2 : NSBlockOperation = NSBlockOperation (block: {
            sleep(1)
            print("Operation2”)
        })
        var operations = [NSOperation]()
        operations.append(operation1)
        operations.append(operation2)
        let queue = NSOperationQueue()
        queue.maxConcurrentOperationCount = 1
        queue.addOperations(operations, waitUntilFinished: true)

----------------------------------------------------------------------------------------------------

Above code was tested with OpenSource Foundation on Mac.

### DIFF
--- a/Foundation/NSOperation.swift
+++ b/Foundation/NSOperation.swift
@@ -318,7 +318,7 @@ public class NSOperationQueue : NSObject {
             } else {
                 effectiveName = "NSOperationQueue::\(unsafeAddress(of: self))"
             }
-            let attr: dispatch_queue_attr_t
+            let attr: dispatch_queue_attr_t?
             if maxConcurrentOperationCount == 1 {
                 attr = DISPATCH_QUEUE_SERIAL
             } else {
--- a/TestFoundation/TestNSOperationQueue.swift
+++ b/TestFoundation/TestNSOperationQueue.swift
@@ -20,6 +20,8 @@ import SwiftXCTest
 class TestNSOperationQueue : XCTestCase {
     static var allTests: [(String, (TestNSOperationQueue) -> () throws -> Void)] {
         return [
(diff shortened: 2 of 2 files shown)


parkera:
Is this just a test case to recreate the bug or an actual fix?

mamabusi:
The test-case mentioned in the comment is only for recreating the bug. Fix is also included in the PR. Foundation class 'NSOperation' is modified to handle the unwrapping of an option.  XCTest format test-case is also added to the TestFoundation.

parkera:
Sorry, I feel like I'm missing something. How does making `attr` optional fix the bug? It was already assigned in either side of that if statement.

mamabusi:
I agree that this seems to be confusing as the variable 'attr' to be optional should have been caught by the compiler itself. I had raised a query on the mailing list and here is what Philippe Hausler had to say: 

"I have a feeling this is associated with the changes for IUO types that recently landed. I am very worried that this actually could happen on Darwin targets as well. Perhaps it is the swift overlay for dispatch that is correcting the failure on Darwin and the dispatch on Linux is missing that annotation. Worth looking into."

The complete conversation of this issue is on the mailing list with the subject line <Setting property "maxConcurrentOperationCount = 1" of NSOperation fails to execute operations using OpenSource Foundation.>

parkera:
@swift-ci please test

parkera:
Ok.

phausler:
Looks good to me

PR #355, opened 2016-05-10
Updated remaining NSCharacterSet API to match Darwin version

tomquist:


parkera:
@swift-ci please test

PR #358, opened 2016-05-11
Remove use of pthread qos on Linux

seabaylea:
The use of QOS isn't supported on Linux, so needs to be wrapped in a guard.

### DIFF
--- a/CoreFoundation/Stream.subproj/CFStream.c
+++ b/CoreFoundation/Stream.subproj/CFStream.c
@@ -1761,7 +1761,9 @@ static CFRunLoopRef _legacyStreamRunLoop()
             pthread_attr_t attr;
             pthread_attr_init(&attr);
             pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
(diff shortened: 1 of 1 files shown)


CoreFoundation/Stream.subproj/CFStream.c
```diff
@@ -1761,7 +1761,9 @@ static CFRunLoopRef _legacyStreamRunLoop()
             pthread_attr_t attr;
             pthread_attr_init(&attr);
             pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
+#if !DEPLOYMENT_TARGET_LINUX
```
parkera: Let's actually change this to DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED. We prefer the positive-sense style for deployment target macros in CF.

seabaylea:
@parkera Thanks for the review. Updated and squashed in place.

parkera:
Looks good to me.

PR #359, opened 2016-05-11
Use _CFIsMainThread as a shim for pthread_main_np on Linux

seabaylea:
`NSOperation.currentQueue()` makes a call to `pthread_main_np()`, which isn't supported on Linux. There's already a shim in CoreFoundation that implements the function under `_CFIsMainThread()`, so this PR:
1. Externalizes `_CFIsMainThread()`
2. Imports `CoreFoundation` and creates a `pthread_main_np` function that calls it

The implementation of the `pthread_main_np()` function in `NSOperation` is done as a top level function - that's probably not the right thing to do, but it allows it to be groups with the import of `CoreFoundation`, which is only required to expose `_CFIsMainThread()`. This way its a little more clear why the function is there, and makes it easier to clean up the code if we implement the `pthread_main_np()` shim a better way.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -806,7 +806,6 @@ CF_EXPORT void _NS_pthread_setname_np(const char *name);
 #endif
 
 #if DEPLOYMENT_TARGET_LINUX
-CF_EXPORT Boolean _CFIsMainThread(void);
 #define pthread_main_np _CFIsMainThread
 #endif
 
--- a/Foundation/NSOperation.swift
+++ b/Foundation/NSOperation.swift
@@ -9,6 +9,12 @@
 
 #if DEPLOYMENT_ENABLE_LIBDISPATCH
 import Dispatch
+#if os(Linux)
+import CoreFoundation
+private func pthread_main_np() -> Int32 {
+    return _CFIsMainThread() ? 1 : 0
+}
+#endif
 #endif
 
 public class NSOperation : NSObject {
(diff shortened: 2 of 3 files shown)


CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -623,6 +623,11 @@ CFIndex CFAllocatorGetPreferredSizeForSize(CFAllocatorRef allocator, CFIndex siz
 CF_EXPORT
 void CFAllocatorGetContext(CFAllocatorRef allocator, CFAllocatorContext *context);
 
+#if DEPLOYMENT_TARGET_LINUX
+CF_EXPORT
+int _CFIsMainThread(void);
+#endif
+
```
parkera: This should go into `ForSwiftFoundationOnly.h`.

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -806,7 +806,7 @@ CF_EXPORT void _NS_pthread_setname_np(const char *name);
 #endif
 
 #if DEPLOYMENT_TARGET_LINUX
-CF_EXPORT Boolean _CFIsMainThread(void);
+CF_EXPORT int _CFIsMainThread(void);
```
parkera: ... And be removed from CFInternal.h.

Also I think we should leave the result as Boolean.

Foundation/NSOperation.swift
```diff
@@ -9,6 +9,12 @@
 
 #if DEPLOYMENT_ENABLE_LIBDISPATCH
 import Dispatch
+#if os(Linux)
+import CoreFoundation
+private func pthread_main_np() -> Int32 {
+    return _CFIsMainThread()
+}
+#endif
```
parkera: Here I would turn the boolean result of _CFIsMainThread into the Int32 expected by the pthread_main_np stub.

seabaylea:
@parkera Thanks for the review. I've made the changes requested.

parkera:
@swift-ci please test

PR #361, opened 2016-05-13
NSProgress

hartbit:


hartbit:
First question for @parkera (or anybody else who can answer): are we allowed to make very small API modifications? For example, I was envisioning moving some string values into enums with raw value strings:

```
public let NSProgressFileOperationKindDownloading: String = "NSProgressFileOperationKindDownloading"
public let NSProgressFileOperationKindDecompressingAfterDownloading: String = "NSProgressFileOperationKindDecompressingAfterDownloading"
public let NSProgressFileOperationKindReceiving: String = "NSProgressFileOperationKindReceiving"
public let NSProgressFileOperationKindCopying: String = "NSProgressFileOperationKindCopying"
```

Would become:

```
public enum NSProgressFileOperationKind: String {
    case downloading = "NSProgressFileOperationKindDownloading"
    case decompressingAfterDownloading = "NSProgressFileOperationKindDecompressingAfterDownloading"
    case receiving = "NSProgressFileOperationKindReceiving"
    case copying = "NSProgressFileOperationKindCopying"
```

Its more of a "namespacing" change really.

parkera:
For now, let's leave the API identical to the one in Foundation. That will make it easier to update in the future. You probably already noticed that the swift-evolution "drop NS" proposal listed this particular enum as being hoisted into the Progress class itself anyway. More details on that will be available a little bit later.

hartbit:
Ok, sounds great. Two extra questions:

1) Objective-C Foundation has `NSProgress.init()` defined, but it's inheriting it from `NSObject`. I guess we should add an `init` definition in `NSProgress` to have identical APIs. Correct?

2) I start by writing my tests against Objective-C Foundation and copy them to corelibs-foundation once they pass. While doing that, I found two surprises:

```
let progress = NSProgress(parent: nil, userInfo: ["key": "value"])
progress.userInfo["key"]
```

Had to be rewritten to:

```
let NSProgress(parent: nil, userInfo: ["key".bridge(): "value" as! AnyObject])
progress.userInfo["key".bridge()]
```

I understand that `bridge()` is currently necessary because `userInfo` expects a `[NSObject:AnyObject]` and corelibs-foundation doesn't automatically bridge `String` to `NSString` yet. But why is `as! AnyObject` necessary on "value" when `AnyObject` is expected?

hartbit:
@parkera New question :) The comments for isIndeterminate say:

```
Whether the progress being made is indeterminate. -isIndeterminate returns YES when the value of the totalUnitCount or completedUnitCount property is less than zero. Zero values for both of those properties indicates that there turned out to not be any work to do after all; -isIndeterminate returns NO and -fractionCompleted returns 1.0 in that case. NSProgress is by default KVO-compliant for these properties, with the notifications always being sent on the thread which updates the property.
```

But after unit-testing Darwin Foundation, `fractionCompleted` does not return `1.0` when both `totalUnitCount` and `completedUnitCount` are equal to `0`, it's actually equal to `0.0`:

```
progress.completedUnitCount = 0
progress.totalUnitCount = 0
XCTAssertEqual(progress.fractionCompleted, 0)
```

I'm going to go towards following the implementation rather than the documentation. But I thought that was worth mentioning.

parkera:
I think I had to change the definition of `isIndeterminate` at some point in the ObjC version. Here is what it uses:

```
    // An uninitialized NSProgress ( completed == total == 0 ) is not finished, it is indeterminate
    return _completed < 0 || _total < 0 || (_completed == 0 && _total == 0);
```

Foundation/NSProgress.swift
```diff
@@ -118,11 +137,21 @@ public class NSProgress : NSObject {
     
     /* Whether the progress being made is indeterminate. -isIndeterminate returns YES when the value of the totalUnitCount or completedUnitCount property is less than zero. Zero values for both of those properties indicates that there turned out to not be any work to do after all; -isIndeterminate returns NO and -fractionCompleted returns 1.0 in that case. NSProgress is by default KVO-compliant for these properties, with the notifications always being sent on the thread which updates the property.
     */
-    public var indeterminate: Bool { NSUnimplemented() }
+    public var isIndeterminate: Bool {
+        return completedUnitCount <= 0 && totalUnitCount <= 0
```
parkera: as described in the other comment:

```
completedUnitCount < 0 || totalUnitCount < 0 || (completedUnitCount == 0 && totalUnitCount == 0)
```

parkera: by the way, isFinished is:

```
((_completed >= _total) && _completed > 0 && _total > 0) || (_completed > 0 && _total == 0);
```

Foundation/NSProgress.swift
```diff
@@ -118,11 +137,21 @@ public class NSProgress : NSObject {
...
+    }
     
     /* The fraction of the overall work completed by this progress object, including work done by any children it may have.
     */
-    public var fractionCompleted: Double { NSUnimplemented() }
+    public var fractionCompleted: Double {
+        if completedUnitCount <= 0 || totalUnitCount < 0 {
```
parkera: first case should call to isIndeterminate and return 0.0 if that's true

parkera:
This is a great start.

hartbit:
@parkera I've taken up work on `Progress` again. I've started by adapting to the new APIs, and doing some documentation at the same time.

Is it OK to copy/paste some documentation from https://developer.apple.com/reference/foundation/nsprogress ?

hartbit:
Other questions:

* I saw that `localizedDescription` and `localizedAdditionalDescription` return a IUO String even if their documentation states that they return an empty string when they can't compute a description. I think that's a relic from the Objective-C headers not having the nullability attributes. I transformed those into straight Strings. Any complaints?

* Some functions have weird looking secondary parameter labels like `func addChild(_ child: Progress, withPendingUnitCount inUnitCount: Int64)`'s `inUnitCount`. As those are only used in the documentation and are not part of the signature of the function, I modified them to be more natural. Any complaints?

hartbit:
> I saw that localizedDescription and localizedAdditionalDescription return a IUO String even if their documentation states that they return an empty string when they can't compute a description. I think that's a relic from the Objective-C headers not having the nullability attributes. I transformed those into straight Strings. Any complaints?

In fact, after more consideration, this is a weird case. Technically, if I want to be 100% compatible, I can't remove the IUO because the property's setter is public: I can set the `localizedDescription` to `nil` under Darwin Foundation. Which makes no sense. And doesn't even do anything because the getter still returns the precomputed value. This is an API issue. To be correct, the property should become a computed property (or have its setter made private) on Darwin Foundation. But I don't know if that's possible.

parkera:
The documentation is a bit misleading - these should never return an empty string. They are not really IOU, they are null-resettable.

PR #365, opened 2016-05-14
Fix crash when running test_xmlpath

NaughtyOttsel:
Resolve crash when _CFXMLNodeGetName returns an empty string.

For some reason and empty pointer causes a crash during the String init.

### DIFF
--- a/Foundation/NSXMLNode.swift
+++ b/Foundation/NSXMLNode.swift
@@ -262,7 +262,11 @@ public class NSXMLNode : NSObject, NSCopying {
     */
     public var name: String? {
         get {
-            return String(cString: _CFXMLNodeGetName(_xmlNode))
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci please test

NaughtyOttsel:
@parkera Looking at the CI Console Output, it appears that the test this aims to fix is not run...

Whilst I'm happy to see a green tick on this PR, doesn't seem right to merge if the CI Tests don't include the test it fixed on Darwin.

parkera:
Thanks @NaughtyOttsel! I'll re-run the tests.

parkera:
@swift-ci please test

parkera:
double checking CI after recent nastiness

parkera:
@swift-ci please test

PR #366, opened 2016-05-15
Foundation: clear up some unused return value warnings

compnerd:
This just cleans up a set of unused-return-value warnings in the Foundation
implementation.  NFC.

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -392,12 +392,12 @@ extension NSData {
         }
         return NSData(bytes: bytes.advanced(by: range.location), length: range.length)
     }
-    
+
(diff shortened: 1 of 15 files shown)


modocache:
@swift-ci please test

parkera:
@swift-ci please test

phausler:
Some of these are actually hiding potential errors or error handling. I am actually very surprised that the glibc functions are being imported with a warning (since they are not annotated with `__attribute__((warn_unsed_result))`. I was under the assumption that swift functions were now by default with this warning, but un-annotated methods imported from external languages (objc/c) were suppressed.

compnerd:
I agree that error checking would be good (e.g. on stat).  However, this is maintaining the status quo.  I think that we should figure out which of these are interesting cases to error check and then do that in a follow up change.

PR #369, opened 2016-05-15
CFStorage: use the reserved spelling of restrict

compnerd:
`__restrict` is the reserved extension.  It avoids the need for the use of the
preprocesor on Windows, but has the same results on compilers which support
restrict (though, some like MSVC support __restrict though not restrict).


compnerd:
CC @phausler

phausler:
I like it! Looks good to me

PR #370, opened 2016-05-15
Base: merge two identical CPP cases

compnerd:
The two cases are identical.  Merge them to avoid the unnecessary duplication.


compnerd:
CC @phausler @parkera

phausler:
This might pose some merging issues for us in the actual Darwin version. I will have to double check if the branches collapse correctly (they used to not)

phausler:
After looking at this; I think we can easily do this. Looks like this was unified in the most recent major releases of those deployment targets.

PR #372, opened 2016-05-18
Halt

compnerd:


CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -158,22 +158,17 @@ CF_PRIVATE CFIndex __CFActiveProcessorCount();
 #endif
 
 #if DEPLOYMENT_TARGET_WINDOWS
-#define __builtin_unreachable() do { } while (0)
+#if !defined(__GNUC__)
```
parkera: Why not use _MSC_VER here as below, instead of looking for !__GNUC__?

compnerd: Because you could be using a GNU compatible compiler on Windows (e.g. clang).  *-windows-* other than windows-msvc will define __GNU__ even though the code is being compiled for Windows.

parkera:
@swift-ci please test

compnerd:
The failure seems unrelated to this change.

compnerd:
Ping?

parkera:
I'm reluctant to merge anything in while we're still trying to figure out the root cause of the CFNumber failure currently hitting CI.

compnerd:
I didn't realize that there were still CI issues.  Sorry for the noise (I thought they had been resolved and this was just forgotten).

parkera:
I think the CI problem is cleared up now (compiler bug). I'll try this again.

parkera:
@swift-ci please test

PR #378, opened 2016-05-20
For SE-0017, remove use of OpaquePointer

russbishop:
This is related to the changes in https://github.com/apple/swift/pull/2537


russbishop:
Confirmed that Foundation builds properly with the changes in the Swift PR.

PR #379, opened 2016-05-21
CoreFoundation: use C99 macro expansion

compnerd:
Rather than defining the various flavours of CFAssert requiring the user to
count the number of substituted parameters, collapse them to CFAssert.  The C99
variadic macro expansion permits us to ensure that the arguments are all
passable.  At the same time, the compiler validation of the printf conversion
will still catch the missing arguments (which could still be possible to have in
the previous case by just using the wrong count).

### DIFF
--- a/CoreFoundation/Base.subproj/CFBase.c
+++ b/CoreFoundation/Base.subproj/CFBase.c
@@ -762,7 +762,7 @@ void CFAllocatorGetContext(CFAllocatorRef allocator, CFAllocatorContext *context
 #else
     __CFGenericValidateType(allocator, __kCFAllocatorTypeID);
 #endif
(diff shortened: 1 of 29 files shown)


phausler:
On the surface this looks pretty good; however I am going to need to make certain that it will work for merging this. Will this prevent MSVC builds for older versions of VisualStudio?

compnerd:
It should work with the older releases of MSVC.  At least VS2013+, though, I expect this to work with even older releases.

phausler:
So I did some research on merging this: tentatively I think we can do this (the only issue that will happen is that when I merge this back into the mainline we will need to retain the CFAssertN macros due to other usage points outside of the open source CF. But that can be addressed at a later point.

PR #380, opened 2016-05-21
[SR-1552] Fix memory leak in NSMutableData and NSData

mbvreddy:
This is fix for bug https://bugs.swift.org/browse/SR-1552

Memory held by NSData._bytes never gets reclaimed as __kCFDontDeallocate is set on NSData always.


phausler:
This looks about right. I need to double check some other interactions before merging this.

ianpartridge:
Hi @phausler - any update?  Also, could we get this fix into `swift-3.0-preview-1-branch`?

djones6:
Resolving this leak is important to server side applications on Linux.
@phausler - were you able to complete your checks?

mbvreddy:
@phausler can this be merged?

phausler:
From what I can tell we actually had a similar fix already applied to the internal versions of CF, unfortunately I was bogged down with other tasks so there was a bit of delay on getting this in - so it will have to hit in the next cut of prebuilt toolchains

PR #381, opened 2016-05-21
Porting to Cygwin

tinysun212:
This is the initial porting.

A simple function like below code is working in Cygwin.
```
import Foundation
let swifty = NSURLComponents(string: "https://swift.org")!
print("\(swifty.host!)")
```

Maybe there are many bugs in this port, but it is sufficient us to review and taste the Foundation in Cygwin.

The Swift compiler option "-DCYGWIN" will be added if companion PR #541 is merged.

The macro might be changed to some other macro function like os(Cygwin) later.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -156,6 +156,8 @@ const char *_CFProcessPath(void) {
 #endif
 
 #if DEPLOYMENT_TARGET_LINUX
+#if TARGET_OS_CYGWIN
+#else
 #include <unistd.h>
 #if __has_include(<syscall.h>)
 #include <syscall.h>
@@ -166,6 +168,7 @@ const char *_CFProcessPath(void) {
 Boolean _CFIsMainThread(void) {
     return syscall(SYS_gettid) == getpid();
 }
+#endif
 
 const char *_CFProcessPath(void) {
     if (__CFProcessPath) return __CFProcessPath;
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -182,7 +182,7 @@ typedef int		boolean_t;
 typedef unsigned long fd_mask;
 #endif
 
-#ifndef __ANDROID__
+#if !defined(__ANDROID__) && !TARGET_OS_CYGWIN
 CF_INLINE size_t
 strlcpy(char * dst, const char * src, size_t maxlen) {
     const size_t srclen = strlen(src);
@@ -210,7 +210,9 @@ strlcat(char * dst, const char * src, size_t maxlen) {
 }
 #endif
 
+#if !TARGET_OS_CYGWIN
 #define issetugid() 0
+#endif
     
 // Implemented in CFPlatform.c 
 bool OSAtomicCompareAndSwapPtr(void *oldp, void *newp, void *volatile *dst);
@@ -269,7 +271,14 @@ void OSMemoryBarrier();
 
 #endif
 
-#if DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_LINUX    
+#if TARGET_OS_CYGWIN
+#define HAVE_STRUCT_TIMESPEC 1
+#define strncasecmp_l(a, b, c, d) strncasecmp(a, b, c)
+#define _NO_BOOL_TYPEDEF
+#undef interface
+#endif
+
+#if DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_LINUX
 #if !defined(MIN)
 #define MIN(A,B)	((A) < (B) ? (A) : (B))
 #endif
--- a/CoreFoundation/Base.subproj/SwiftRuntime/TargetConditionals.h
+++ b/CoreFoundation/Base.subproj/SwiftRuntime/TargetConditionals.h
@@ -46,6 +46,7 @@
     
         TARGET_OS_WIN32           - Generated code will run under 32-bit Windows
         TARGET_OS_UNIX            - Generated code will run under some Unix (not OSX) 
+           TARGET_OS_CYGWIN           - Generated code will run under 64-bit Cygwin
         TARGET_OS_MAC             - Generated code will run under Mac OS X variant
            TARGET_OS_IPHONE          - Generated code for firmware, devices, or simulator 
               TARGET_OS_IOS             - Generated code will run under iOS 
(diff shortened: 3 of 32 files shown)


CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -155,13 +155,15 @@ const char *_CFProcessPath(void) {
 }
 #endif
 
-#if DEPLOYMENT_TARGET_LINUX
+#if DEPLOYMENT_TARGET_LINUX || DEPLOYMENT_TARGET_CYGWIN
 #include <unistd.h>
+#if !DEPLOYMENT_TARGET_CYGWIN
```
parkera: Generally we avoid the negative-sense deployment target macros. Keeping them consistently positive-sense and using #if/#else instead makes it easier to reason about them (and there are lots).

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -263,7 +265,15 @@ void OSMemoryBarrier();
 
 #endif
 
-#if DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_LINUX    
+#if DEPLOYMENT_TARGET_CYGWIN
+#define HAVE_STRUCT_TIMESPEC 1
+#define strncasecmp_l(a, b, c, d) strncasecmp(a, b, c)
+#define _NO_BOOL_TYPEDEF
+#include <windows.h>
```
parkera: So, what Windows SDK is required here?

build.py
```diff
@@ -22,6 +22,9 @@
 elif Configuration.current.target.sdk == OSType.MacOSX:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_MACOSX '
 	foundation.LDFLAGS = '-licucore -twolevel_namespace -Wl,-alias_list,CoreFoundation/Base.subproj/DarwinSymbolAliases -sectcreate __UNICODE __csbitmaps CoreFoundation/CharacterSets/CFCharacterSetBitmaps.bitmap -sectcreate __UNICODE __properties CoreFoundation/CharacterSets/CFUniCharPropertyDatabase.data -sectcreate __UNICODE __data CoreFoundation/CharacterSets/CFUnicodeData-L.mapping -segprot __UNICODE r r '
+elif Configuration.current.target.sdk == OSType.Win32:
```
parkera: Does this mean that cygwin is the only way to build on Windows?

The current Windows deployment targets were designed to allow CF to build using Visual Studio against the Windows SDK, with no cygwin required.

parkera:
I like the general direction of this, but I wonder if we can simplify the patch a bit by lumping cygwin under linux in general (since it seems to closely related) and then having just a few places where they differ.

Including the Windows SDK (I'm assuming this is the "windows.h" we imported) is a pretty big change though. Is there any compatibility shim provided by cygwin that will allow us to skip that?

tinysun212:
@parkera, Thanks your reviews. I will reply your review or modify the code, as soon as I fix my machine which does not boot from two days ago.

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
tinysun212: Functions ```eventfd(), epoll(), timerfd_create()``` used for Linux are not supported by Cygwin because they are not POSIX standard. I used the Windows implementation using ```#elif DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_CYGWIN```.
Win32 API is used only for ```CFRunLoop.c```.

build.py
tinysun212: I didn't know the method to distinguish the Cygwin from Windows (MSVC). I'll separate this file from PR.

tinysun212:
@parkera,
I changed the ```#if !DEPLOYMENT_TARGET_CYGWIN``` to ```#if/#else``` pattern.
The python scripts will be modified in other PR considering Windows.
The ```#include <windows.h>``` is moved from ```CoreFoundation_Prefix.h```  to ```CFRunLoop.c```.

CoreFoundation/Base.subproj/SwiftRuntime/TargetConditionals.h
```diff
@@ -76,21 +77,31 @@
...
+#define TARGET_OS_CYGWIN       0
 #elif __linux__
 #define TARGET_OS_DARWIN       0
 #define TARGET_OS_LINUX        1
 #define TARGET_OS_WINDOWS      0
 #define TARGET_OS_BSD          0
+#define TARGET_OS_CYGWIN       0
+#elif __CYGWIN__
+#define TARGET_OS_DARWIN       0
+#define TARGET_OS_LINUX        0
```
parkera: So - how different is cygwin from linux? cygwin's idea is to provide a Linux-like environment for Windows.

I think we should have TARGET_OS_LINUX here set to 1 for cygwin, and also define DEPLOYMENT_TARGET_LINUX when building cygwin.

We can also define TARGET_OS_CYGWIN here (as you have done) and use that in the rare places where cygwin is truly different from Linux. That should dramatically simplify this patch, and also keep us from having to specify that deployment target everywhere (or forgetting to, and breaking the build for Cygwin).

Foundation/NSConcreteValue.swift
```diff
@@ -11,7 +11,7 @@ import CoreFoundation
 
 #if os(OSX) || os(iOS)
     import Darwin
-#elseif os(Linux)
+#elseif os(Linux) || IS_CYGWIN
```
parkera: If we do what I said above, would os(Linux) just work for cygwin? I could understanding needing the platform difference for CF (which has a lot of platform specific stuff in it), but I'm not sure we need to distinguish between the two from the Swift code.

Foundation/NSGeometry.swift
```diff
@@ -36,7 +36,7 @@ public struct CGFloat {
 #if arch(i386) || arch(arm)
         return Int(Float(self.native).bitPattern)
 #else
-        return Int(self.native.bitPattern)
+        return Int(self.native.bitPattern())
```
parkera: I don't think this diff should be in here (perhaps a toolchain mismatch?)

CoreFoundation/Base.subproj/SwiftRuntime/TargetConditionals.h
tinysun212: Cygwin provides the POSIX system calls and environment, and is similar to Linux because Linux also confirms to POSIX. But some functions like ```eventfd```, ```epoll```, ```timerfd_create```  seems to be never implemented in Cygwin because they are not POSIX standard.
Anyway, I agree your method will simplify this patch and I will modify if others has no opposing opinion, although there was similar opposing discussions for compiler. [swift #2351](https://github.com/apple/swift/pull/2351#discussion_r61667167)

Foundation/NSConcreteValue.swift
tinysun212: Some headers ```aio.h, cpio.h, fmtmsg.h, nl_types.h, ulimit.h, ``` are not exist in Cygwin. If we can not distinguish Cygwin from Linux, the Swift code on Cygwin can not know if the Glibc function is implemented.

Current swift compiler is implemented as ```os(Windows)``` is true for ```Windows (MSVC)``` and ```Cygwin```. Since ```os()``` is exclusive, ```os(Linux)``` is false in Cygwin. There was a not finished (holded) discussion. ([swift-evolution ](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160502/thread.html#16381)).
I think ```IS_CYGWIN``` is best until the discussion.

Foundation/NSGeometry.swift
tinysun212: Sorry, this will be revoked.

tinysun212:
@parkera, I changed definitions for Cygwin, `(DEPLOYMENT_TARGET_CYGWIN, TARGET_OS_CYGWIN)` to `(DEPLOYMENT_TARGET_LINUX, TARGET_OS_LINUX, TARGET_OS_CYGWIN)` as you suggested.

mominul:
ping @parkera

tinysun212:
@parkera, I think the requested changes are applied.

parkera:
Thanks - we'll take a look at this in the post 3.0 timeframe.

parkera:
We should see about merging this now. @tinysun212 can you rebase against master?

tinysun212:
@parkera, I rebased and squashed to one commit.

parkera:
@swift-ci please test

parkera:

parkera:
Don't know why the CI won't kick this off, but it passed before so I don't see what the problem would be now. Merging manually.

tinysun212:
Thanks for your review and merging.

PR #382, opened 2016-05-21
Add implementation for NSURL copy()

mbvreddy:
This PR has below implementations along with testcase

NSURL.copyWithZone()
NSURLQueryItem.copyWithZone()

### DIFF
--- a/TestFoundation/TestNSURL.swift
+++ b/TestFoundation/TestNSURL.swift
@@ -60,7 +60,8 @@ class TestNSURL : XCTestCase {
             // TODO: these tests fail on linux, more investigation is needed
             ("test_fileURLWithPath", test_fileURLWithPath),
             ("test_fileURLWithPath_isDirectory", test_fileURLWithPath_isDirectory),
-            ("test_URLByResolvingSymlinksInPath", test_URLByResolvingSymlinksInPath)
+            ("test_URLByResolvingSymlinksInPath", test_URLByResolvingSymlinksInPath),
+            ("test_copy", test_copy)
         ]
     }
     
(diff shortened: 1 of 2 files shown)


TestFoundation/TestNSURL.swift
```diff
@@ -417,6 +418,16 @@ class TestNSURL : XCTestCase {
             XCTAssertEqual(result, "file:///tmp/")
         }
     }
+
+    func test_copy() {
+        let url = NSURL(string: "https://www.swift.org")
+        let urlCopy = url!.copy() as! NSURL
+        XCTAssertTrue(url === urlCopy)
```
parkera: This requires that the url instance is the same as the other; we may change that in the future. I would prefer the test checks a property of the URL for equality, or even better calls into URL's isEqual

TestFoundation/TestNSURL.swift
```diff
@@ -417,6 +418,16 @@ class TestNSURL : XCTestCase {
...
+
+        let queryItem = NSURLQueryItem(name: "id", value: "23")
+        let queryItemCopy = queryItem.copy() as! NSURLQueryItem
+        XCTAssertTrue(queryItem === queryItemCopy)
```
parkera: Same here

mbvreddy:
@parkera I've squashed suggested changes. Please check.

mbvreddy:
@parkera 

I've resolved conflicts.

NSURL.copy is already taken care by @phausler with commit https://github.com/apple/swift-corelibs-foundation/commit/6ce843275247e5fb1b42c66bb882cec0e3f7c8af

This PR now has implementation for NSURLQueryItem.copy() along with testcase.

mbvreddy:
@parkera can this be merged?

parkera:
Yes, this should be good.

parkera:
@swift-ci please test

PR #385, opened 2016-05-24
Changed the NSOrderedSet to use NSMutableArray and NSMutableSet internally

sgl0v:
`NSOrderedSet` API contains lots of methods similar to the ones we have in `NSSet` & `NSArray`. I'd consider using these collections instead of the Swift ones in the `NSOrderedSet`/`NSMutabeOrderedSet` implementation.

### DIFF
--- a/Foundation/NSOrderedSet.swift
+++ b/Foundation/NSOrderedSet.swift
@@ -9,8 +9,8 @@
 
 /****************       Immutable Ordered Set   ****************/
 open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, ExpressibleByArrayLiteral {
-    internal var _storage: Set<NSObject>
-    internal var _orderedStorage: [NSObject]
+    internal var _storage: NSMutableSet
+    internal var _orderedStorage: NSMutableArray
     
     open override func copy() -> Any {
         return copy(with: nil)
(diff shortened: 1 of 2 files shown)


Foundation/NSOrderedSet.swift
```diff
@@ -243,25 +232,61 @@ extension NSOrderedSet {
...
     // appear to change spontaneously, since a copy of the ordered set is not
     // being made.
-    public var array: [AnyObject] { NSUnimplemented() }
-    public var set: Set<NSObject> { NSUnimplemented() }
-    
-    public func enumerateObjectsUsingBlock(_ block: (AnyObject, Int, UnsafeMutablePointer<ObjCBool>) -> Void) { NSUnimplemented() }
-    public func enumerateObjectsWithOptions(_ opts: NSEnumerationOptions, usingBlock block: (AnyObject, Int, UnsafeMutablePointer<ObjCBool>) -> Void) { NSUnimplemented() }
-    public func enumerateObjectsAtIndexes(_ s: NSIndexSet, options opts: NSEnumerationOptions, usingBlock block: (AnyObject, Int, UnsafeMutablePointer<ObjCBool>) -> Void) { NSUnimplemented() }
+    public var array: [AnyObject] {
+        return _orderedStorage.bridge()
```
parkera: This is probably good enough for now, because it's busted on Darwin too - but note that the comment here describes this class as returning a proxy object for this method and also `set`. The bridge does a copy, so no proxying is happening.

parkera:
@swift-ci please test

Foundation/NSOrderedSet.swift
sgl0v: Yeah. I didn't come up with any solution here. Thus return just a copy of the `array`/`set`. Will work on implementing other `NSOrderedSet` methods and think over about improving these ones then.

parkera:
Unfortunately this became out of date. Is it possible to rebase?

sgl0v:
Finally did rebase on current master branch.

alblue:
This has gone stale. Please rebase and reopen if necessary.

PR #386, opened 2016-05-25
Support Linux on z

vivkong:
- Add initial support for Linux on IBM z Systems
- Fix endianess issues in CFString (https://bugs.swift.org/browse/SR-1548)

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -343,6 +343,13 @@ CF_PRIVATE Boolean __CFProcessIsRestricted();
 CF_EXPORT void * __CFConstantStringClassReferencePtr;
 CF_EXPORT void *__CFConstantStringClassReference[];
 
+#if __CF_BIG_ENDIAN__
+#define CFINFO {0x00, 0x00, 0x07, 0xc8}
+#elif __CF_LITTLE_ENDIAN__
+#define CFINFO {0xc8, 0x07, 0x00, 0x00}
+#endif
+
+
 #ifdef __CONSTANT_CFSTRINGS__
 
 #if DEPLOYMENT_RUNTIME_SWIFT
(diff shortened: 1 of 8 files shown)


CoreFoundation/String.subproj/CFString.h
```diff
@@ -165,7 +165,11 @@ struct __CFConstStr {
         uint8_t _pad[4];
     } _base;
     uint8_t *_ptr;
+#if __LP64__
+    uint64_t _length;
+#else
```
parkera: Hm, this is always a uint32_t on other platforms, even in 64 bit.

vivkong: I have changed the code to use unsigned long instead (same as the length field in __CFString struct which is of type CFIndex)

CoreFoundation/String.subproj/CFString.h
```diff
@@ -165,7 +165,7 @@ struct __CFConstStr {
         uint8_t _pad[4];
     } _base;
     uint8_t *_ptr;
-    uint32_t _length;
+    signed long _length;
```
ianpartridge: Signed?

vivkong: I meant to change it to be the same as the length field in __CFString struct which is of type CFIndex.   CFIndex is signed long.

CoreFoundation/String.subproj/CFString.h
```diff
@@ -165,7 +165,7 @@ struct __CFConstStr {
         uint8_t _pad[4];
     } _base;
     uint8_t *_ptr;
-    uint32_t _length;
+    unsigned long _length;
```
phausler: This must remain the same since the emission is determined by what clang itself emits (and has emitted for previously compiled libraries) this layout is paired not only to the emission of `@""` but also `CFSTR("")`

phausler: If it is needed for this platform to change then it should have an ifdef for gating the change only for that specific platform.

CoreFoundation/String.subproj/CFUnicodeDataB.S
```diff
@@ -0,0 +1,20 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+    .global __CFUnicodeDataB
```
phausler: These asm files should be preprocessed so perhaps we should #ifdef them out to nothing on the platform it is appropriate for e.g. Little endian only has little endian data slices and big endian only has big endian slices. It might make it simpler for detecting duplicate symbols and the such.

CoreFoundation/String.subproj/CFString.h
```diff
@@ -174,10 +174,17 @@ struct __CFConstStr {
 #define CONST_STRING_LITERAL_SECTION
 #endif
 
+#if __BIG_ENDIAN__
+#define CFSTR(cStr)  ({ \
+static struct __CFConstStr str CONST_STRING_LITERAL_SECTION = {{(uintptr_t)&__CFConstantStringClassReference, _CF_CONSTANT_OBJECT_STRONG_RC, 0, {0x00, 0x00, 0x07, 0xc8}, {0x00, 0x00, 0x00, 0x00}}, (uint8_t *)(cStr), sizeof(cStr) - 1}; \
+(CFStringRef)&str; \
+})
```
phausler: This is not the only macro btw, there is another in CFInternal.h called `CONST_STRING_DECL` that probably should be double checked if it is still correct; it has been a while since CF ran on big endian.

vivkong:
Thank you everyone for the reviews.  I appreciate the feedback and have made changes based on the suggestions.

CoreFoundation/String.subproj/CFString.h
bryanpkc: The aliasing of a 32-bit _length field with a 64-bit CFIndex field will not work on any 64-bit big-endian system (e.g. s390x and powerpc64). So I believe the guard should be:

```
#if defined(__LP64__) && defined(__BIG_ENDIAN__)
```

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -177,6 +177,12 @@ CF_PRIVATE CFIndex __CFActiveProcessorCount();
     #else
         #error Compiler not supported
     #endif
+#elif defined(__s390x__)
+    #if defined(__GNUC__)
+        #define HALT do {__builtin_trap(); kill(getpid(), 9); __builtin_unreachable(); } while (0)
+    #else
+        #error Compiler not supported
+    #endif
```
compnerd: BTW, it would be nice to rebase this on https://github.com/apple/swift-corelibs-foundation/pull/372 .  It collapses the duplication.

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -363,6 +363,13 @@ CF_PRIVATE Boolean __CFProcessIsRestricted();
 CF_EXPORT void * __CFConstantStringClassReferencePtr;
 CF_EXPORT void *__CFConstantStringClassReference[];
 
+#if __CF_BIG_ENDIAN__
+#define CFINFO {0x00, 0x00, 0x07, 0xc8}
+#elif __CF_LITTLE_ENDIAN__
+#define CFINFO {0xc8, 0x07, 0x00, 0x00}
+#endif
```
compnerd: I also have a PR https://github.com/apple/swift-corelibs-foundation/pull/392 which should address the CFInfo field for endianness.  It actually collapses the various definitions into a single one as well.

CoreFoundation/Base.subproj/CFInternal.h
vivkong: Thanks.  I've rebased on top of it.

vivkong:
Wondering if there are other comments or suggestions for this PR?  Thanks.

parkera:
@swift-ci please test

PR #387, opened 2016-05-25
add missing implementation for NSLocale

mbvreddy:
Only UnImplemented api is copy(). This PR added implementation for the same.

### DIFF
--- a/Foundation/NSLocale.swift
+++ b/Foundation/NSLocale.swift
@@ -55,7 +55,7 @@ public class NSLocale : NSObject, NSCopying, NSSecureCoding {
         return copyWithZone(nil)
     }
     
-    public func copyWithZone(_ zone: NSZone) -> AnyObject { NSUnimplemented() }
+    public func copyWithZone(_ zone: NSZone) -> AnyObject { return self }
     
     public func encodeWithCoder(_ aCoder: NSCoder) {
         if aCoder.allowsKeyedCoding {
--- a/TestFoundation/TestNSLocale.swift
+++ b/TestFoundation/TestNSLocale.swift
@@ -19,6 +19,7 @@ class TestNSLocale : XCTestCase {
     static var allTests: [(String, (TestNSLocale) -> () throws -> Void)] {
         return [
             ("test_constants", test_constants),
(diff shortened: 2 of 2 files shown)


Foundation/NSLocale.swift
```diff
@@ -55,7 +55,7 @@ public class NSLocale : NSObject, NSCopying, NSSecureCoding {
         return copyWithZone(nil)
     }
     
-    public func copyWithZone(_ zone: NSZone) -> AnyObject { NSUnimplemented() }
+    public func copyWithZone(_ zone: NSZone) -> AnyObject { return self }
```
parkera: How is this correct? Shouldn't we create a copy of the CFLocaleRef?

parkera: n.b. the CFLocale header:

```
	// Having gotten a CFLocale from somebody, code should make a copy
	// if it is going to use it for several operations
	// or hold onto it.  In the future, there may be mutable locales.
```

mbvreddy:
@parkera NSLocale is immutable. And, on Mac OS X, NSLocale === NSLocale.copy().

parkera:
Guess it's wrong there too; we shouldn't provide a copy function on CFLocale if we don't intend for people to use it. =) I guess this one works for now.

parkera:
@swift-ci please test

parkera:
I think this file manager test needs to make sure it creates unique paths. I'm going to try that again.

parkera:
@swift-ci please test

Foundation/NSLocale.swift
phausler: The base abstract implementation of NSLocale in objc does exactly this:
```
- (id)copyWithZone:(NSZone *)zone {
    return [self retain];
}
```
So it stands to reason that since Locale's are immutable that it is just a return of self

parkera: Yah, I think that's probably a mistake in NSLocale. But anyway, we can proceed with this for now.

parkera:
@swift-ci please test

parkera:
Get this test going again

PR #388, opened 2016-05-27
test: address a number of unused result warnings

compnerd:
Address the various set of unused result warnings in the test suite.  NFC.

### DIFF
--- a/TestFoundation/TestNSBundle.swift
+++ b/TestFoundation/TestNSBundle.swift
@@ -115,12 +115,12 @@ class TestNSBundle : XCTestCase {
             
             // Put some resources in the bundle
             for n in _bundleResourceNames {
-                NSFileManager.defaultManager().createFile(atPath: bundlePath + "/" + n, contents: nil, attributes: nil)
+                _ = NSFileManager.defaultManager().createFile(atPath: bundlePath + "/" + n, contents: nil, attributes: nil)
             }
             // Add a resource into a subdirectory
             let subDirPath = bundlePath + "/" + _subDirectory
             try NSFileManager.defaultManager().createDirectory(atPath: subDirPath, withIntermediateDirectories: false, attributes: nil)
-            NSFileManager.defaultManager().createFile(atPath: subDirPath + "/" + _main + "." + _type, contents: nil, attributes: nil)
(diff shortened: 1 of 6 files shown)


TestFoundation/TestNSBundle.swift
```diff
@@ -115,12 +115,12 @@ class TestNSBundle : XCTestCase {
             
             // Put some resources in the bundle
             for n in _bundleResourceNames {
-                NSFileManager.defaultManager().createFile(atPath: bundlePath + "/" + n, contents: nil, attributes: nil)
+                _ = NSFileManager.defaultManager().createFile(atPath: bundlePath + "/" + n, contents: nil, attributes: nil)
```
parkera: I wonder if we should annotate this API as a discardable result instead. It doesn't have the error handling anyway...

TestFoundation/TestNSString.swift
```diff
@@ -401,7 +401,7 @@ class TestNSString : XCTestCase {
         var buf : [UInt8] = []
         buf.reserveCapacity(reservedLength)
         var usedLen : CFIndex = 0
-        buf.withUnsafeMutableBufferPointer { p in
+        _ = buf.withUnsafeMutableBufferPointer { p in
```
parkera: I think the right place to ignore this result is actually the CFStringGetBytes (which is returning a value passed through the closure here)

TestFoundation/TestNSBundle.swift
compnerd: Im happy to make such a change.  I imagine that a) that would need to go through swift evolution b) should be a follow up change.

TestFoundation/TestNSString.swift
compnerd: Hmm, I think that I agree with you.  I will adjust the patch and upload it again.  Good thought!

TestFoundation/TestNSBundle.swift
parkera: Let's file a follow up JIRA for it.

compnerd: https://bugs.swift.org/browse/SR-1638 ... can't seem to add you to it :-(.

compnerd:
ping

phausler:
Overall the changes look reasonable to me. I will take a look at this later today to see if we can get it merged manually

compnerd:
Awesome!  Thanks!

PR #390, opened 2016-05-28
CoreFoundation: name tags for externally visible types

compnerd:
When building LLDB with modules, CF headers cause some problems due to the tags
being unnamed.  The anonymous tag gets internal linkage which is then changed by
the subsequent typedef.  Name all the tags, which allows these headers to be
used with modules.

### DIFF
--- a/CoreFoundation/Base.subproj/CFBase.h
+++ b/CoreFoundation/Base.subproj/CFBase.h
@@ -475,9 +475,9 @@ static const CFIndex kCFNotFound = -1;
 
 
 /* Range type */
-typedef struct {
-    CFIndex location;
-    CFIndex length;
+typedef struct CFRange {
+  CFIndex location;
+  CFIndex length;
 } CFRange;
 
 #if defined(CF_INLINE)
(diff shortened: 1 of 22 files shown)


parkera:
Why wouldn't this just be considered a bug in lldb, or clang?

compnerd:
I don't think it would be a bug in lldb, it is only using the headers.  However, it could be argued that it is a clang issue.  It would then be an issue with how we construct modules.  The thing is that because you have an anonymous structure, it is defined to have internal linkage (since you could not construct an instance of that struct since it has no name).  The typedef aliases a struct that you cannot construct, so, it is an issue.  However, I could see an argument for relaxing constraints on what type of structures we permit in modules.

phausler:
Does this problem not exist on Darwin because we have a module for CoreFoundation? If so perhaps should we consider doing that?

compnerd:
I *think* I ran into it on Darwin.  It was building lldb from source against the system headers, with modules.  I figured that if I sent this PR here, it would eventually make its way to that set.

parkera:
I'm still unsure about this one, because we certainly have modules for CF on Darwin.

alblue:
This has gone stale; closing.

PR #391, opened 2016-05-28
Base: always define HALT

compnerd:
HALT is used in various places without a check for the target.  There is nothing
which is architecture specific.  Remove the unnecessary architecture check.


phausler:
Looks simpler to me

PR #392, opened 2016-05-28
Base: collapse the various CF_CONST_STR macros

compnerd:
Use a bit more preprocessor to collapse the 6(!!) definitions into 2.  The only
difference is the name used for internal structure definitions on non-swift
builds without constant CFStrings.


compnerd:
ping

PR #393, opened 2016-05-29
Removed uses of deprecated @warn_unused_result attr

tanadeau:
The `@warn_unused_result` attribute is no longer needed as its behavior is now the default. A future change to the compiler (https://github.com/apple/swift/pull/2760) will make its use an error. This PR removes the uses of this attribute.


tanadeau:
@phausler did this in a later commit (9cd298bd57a39129209183a92946ede85a4f1118)

PR #395, opened 2016-05-30
Implementation of NSObject's 'description' property (overriding) for NSURLHttpResponse object.

mamabusi:


phausler:
Looks sensible to me.

PR #396, opened 2016-05-30
Change the NSFormatter API to match the Darwin version

pushkarnk:


pushkarnk:
FYI. After rebasing, the diff bears only a change in a comment.

PR #398, opened 2016-05-31
Support cross-compiling Foundation

karwa:
Support for cross-compiling:
- Replace `icu-config` with `pkg-config`. Even ICU [recommends](http://userguide.icu-project.org/howtouseicu) to use pkg-config these days.
- Include sysroot in include directories
- Added `linker` flag, which gets passed through as the `-fuse-ld` option (so it accepts either a linker name or absolute path, as clang does)
- Added `resource-dir` flag to swift. Similarly to https://github.com/apple/swift/commit/f7353a73ceb2e23af1404155dc0db85adc40200b
- Check for presence of binaries in the toolchain folder before appending `bin`. Our toolchain might just be a folder full of binaries.

### DIFF
--- a/build.py
+++ b/build.py
@@ -15,7 +15,7 @@ foundation.GCC_PREFIX_HEADER = 'CoreFoundation/Base.subproj/CoreFoundation_Prefi
 
 if Configuration.current.target.sdk == OSType.Linux:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE '
-	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc `icu-config --ldflags` -Wl,-defsym,__CFConstantStringClassReference=_TMC10Foundation19_NSCFConstantString -Wl,-Bsymbolic '
+	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc `pkg-config icu-uc icu-i18n --libs` -Wl,-defsym,__CFConstantStringClassReference=_TMC10Foundation19_NSCFConstantString -Wl,-Bsymbolic '
 elif Configuration.current.target.sdk == OSType.FreeBSD:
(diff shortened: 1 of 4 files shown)


phausler:
Looks pretty good to me

build.py
```diff
@@ -15,7 +15,7 @@
 
 if Configuration.current.target.sdk == OSType.Linux:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE '
-	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc `icu-config --ldflags` -Wl,-defsym,__CFConstantStringClassReference=_TMC10Foundation19_NSCFConstantString -Wl,-Bsymbolic '
+	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc `pkg-config icu-uc icu-i18n --libs` -Wl,-defsym,__CFConstantStringClassReference=_TMC10Foundation19_NSCFConstantString -Wl,-Bsymbolic '
```
compnerd: We should use ${PKG_CONFIG} instead.  pig-config can be target prefixed for cross-compiling (e.g. armv7-unknown-linux-gnueabihf-pkg-config).

karwa: I don't think that's necessary - pkg-config isn't architecture-specific; it relies on .pc text files it finds in the `PKG_CONFIG_PATH`. I also have the prefixed version on my Raspberry Pi, but it's just a symlink to unprefixed `pkg-config`. It's not required for cross-compiling either; I've been cross-compiling for the RPi using the `pkg-config` I got from OSX homebrew. It just reads the .pc files from the RPi sysroot, and I've cross-compiled basically all of the swift tools like that.

compnerd: The tool itself reads a file.  However, the path in which the tool searches is platform dependent as the libraries (e.g. on ELF) would be different.  So, you end up linking against the libraries for the wrong architecture if you do not use the right path (which is hard) or use the target triple prefixed version.  The reason that the version you looked at was symlinked is because the unprefixed version was for the target that you were targeting.  Set up a full image for a foreign architecture and you will find it to be more than a symlink.

karwa:
@compnerd Added `--pkg-config` flag which lets you set a prefixed pkg-config binary
@phausler anything else?

PR #399, opened 2016-05-31
Fix building on linux (arm)

karwa:
Without this, we're unable to build on Debian (armv7) because of a conflicting va_list definition.

The include doesn't seem to be required, and removing needless includes is usually a good idea anyway.

See: https://bugs.swift.org/browse/SR-1412

### DIFF
--- a/CoreFoundation/Base.subproj/SwiftRuntime/CoreFoundation.h
+++ b/CoreFoundation/Base.subproj/SwiftRuntime/CoreFoundation.h
@@ -34,7 +34,6 @@
 #include <setjmp.h>
 #include <signal.h>
 #include <stddef.h>
-#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>


hpux735:
Thanks for working on this, @karwa.  It's weird that this appeared.  I'm curious what caused it to happen.

phausler:
I don't think we can merge this change as is; the include is needed by Foundation and other projects that include that on Darwin. It would be better that if this is somehow a failure on linux to include this header that perhaps we gate that include somehow? That way the change is actually mergeable into the Darwin builds of Foundation/CoreFoundation.

karwa:
Actually, turns out we only need to remove it from the `SwiftRuntime/CoreFoundation.h` umbrella header. Basically - we shouldn't export stdio from `CoreFoundation`. For some reason when it's in the module it causes conflicts between clang builtin headers and system headers.

gribozavr:
> because of a conflicting va_list definition

What is defining the second va_list?

karwa:
It's in the bug report. `stdarg.h` has a comment about some kind of Ubuntu 'hack', and I've been [finding occurrences](https://www.google.de/search?q=clang+_G_va_list) of what looks like the same problem going back to 2012/13.

```
/home/ryan/Source/build/Ninja-ReleaseAssert/swift-linux-armv7/lib/swift/clang/include/stdarg.h:30:27: note: candidate found by name lookup is 'va_list'
typedef __builtin_va_list va_list;
                          ^
/usr/include/stdio.h:79:20: note: candidate found by name lookup is 'va_list'
typedef _G_va_list va_list;
```

parkera:
Hm, I'm still not sure about this one.

rfdickerson:
Yeah- I can confirm the duplicate definition of va_args is causing Foundation and libDispatch to fail when building on ARM (like Ubuntu Mate for Raspberry Pi). Commenting out stdio.h fixes it. As an alternative, adding the following to `/usr/lib/swift/clang/include/stdarg.h` seems to fix it also:

```c
#ifndef _VA_LIST
#include <_G_config.h>  //Adding this line 
typedef __builtin_va_list va_list;
#define _VA_LIST
#endif
```

hpux735:
I think I have some more context to add to this.

```
<module-includes>:1:10: note: in file included from <module-includes>:1:
#include "dispatch.h"
         ^
/mnt/build/swift-sources/swift-corelibs-libdispatch/dispatch/dispatch.h:68:10: note: in file included from /mnt/build/swift-sources/swift-corelibs-libdispatch/dispatch/dispatch.h:68:
#include <dispatch/object.h>
         ^
/mnt/build/swift-sources/swift-corelibs-libdispatch/dispatch/object.h:554:64: error: reference to 'va_list' is ambiguous
dispatch_debugv(dispatch_object_t object, const char *message, va_list ap);
                                                               ^
/mnt/build/swift-sources/build/Ninja-ReleaseAssert/swift-linux-armv7/lib/swift/clang/include/stdarg.h:30:27: note: candidate found by name lookup is 'va_list'
typedef __builtin_va_list va_list;
                          ^
/usr/include/stdio.h:79:20: note: candidate found by name lookup is 'va_list'
typedef _G_va_list va_list;
                   ^
/mnt/build/swift-sources/swift-corelibs-libdispatch/src/swift/Block.swift:13:8: error: could not build Objective-C module 'CDispatch'
import CDispatch
       ^
```

Is the complete text of the same error in libdispatch (see https://github.com/apple/swift-corelibs-libdispatch/pull/155 ).  If you look at the system /usr/include/stdio.h it defines va_list, but it makes a feeble attempt to make sure it's only defined if using GNU C:

```
#if defined __USE_XOPEN || defined __USE_XOPEN2K8
# ifdef __GNUC__
#  ifndef _VA_LIST_DEFINED
typedef _G_va_list va_list;
#   define _VA_LIST_DEFINED
#  endif
# else
#  include <stdarg.h>
# endif
#endif
```

Before `#include <stdio.h>` I tried `#undef __GNUC__` but it didn't work for some reason.  I'm a little surprised it's being redefined in such a way.

parkera:
This has been superseded by #603.

PR #400, opened 2016-06-01
Fix the CFAssert macro when used with no format string parameters

briancroom:
https://github.com/apple/swift-corelibs-foundation/pull/379 simplified the usage of the `CFAssert` macro, however it introduced a regression when Foundation is built in the `DEBUG` configuration such that `CFAssert` can no longer be used if no variable arguments are passed after the assertion description, because a trailing comma ends up being produced once the macro is expanded. This is causing build failures for the Corelibs XCTest CI on OS X.

Using `##__VA_ARGS__` (already being used elsewhere in this CoreFoundation) invokes compiler magic which removes the trailing comma in this situation.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -176,7 +176,7 @@ CF_PRIVATE CFIndex __CFActiveProcessorCount();
     #define __CFAssert(cond, prio, desc, ...)                                  \
         do {                                                                   \
           if (!(cond)) {                                                       \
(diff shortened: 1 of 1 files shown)


briancroom:
@swift-ci please test

briancroom:
Considering that CI passed this, and it is interfering with CI builds for XCTest PRs, I'm going to go ahead and merge this. @parkera @phausler feel free to re-open if you are concerned about using `##__VA_ARGS__` (I understand that it may have some portability implications.)

parkera:
Works for me, thanks @briancroom.

PR #401, opened 2016-06-01
Fix TestFoundation build order issue in Xcode 

briancroom:
Currently when building the `TestFoundation` scheme in Xcode with a clean build directory, the build fails because it attempts to build `SwiftXCTest` before `SwiftFoundation`. This fixes the build order in the scheme.


briancroom:
AFAIK CI never touches anything related to this, so there's no point in requesting a build from it.

parkera:
Ok

PR #403, opened 2016-06-02
change NSFileManager.defaultManager() to NSFileManager.default()

mbvreddy:
This change is to align with Darwin version.

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -402,7 +402,7 @@ extension NSData {
         if fd == -1 {
             throw _NSErrorWithErrno(errno, reading: false, path: dirPath)
         }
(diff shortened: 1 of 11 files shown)


parkera:
This is becoming a class property, but there is no way we want the name of default with backticks.

mbvreddy:
default is a keyword. So, I've used backticks to use it as function name. And, signature in Dawrin version is similar to the change I make here.

`public class func `default`() -> NSFileManager`

mbvreddy:
Ah.. I see the changes for this PR is already ported by @phausler in https://github.com/apple/swift-corelibs-foundation/commit/70fd876618025f8e44ecbff311972bc29010ed96

This PR is no longer required.

PR #409, opened 2016-06-10
implementation for NSCalendar.copy()

mbvreddy:
Implementation for NSCalendar.copy() is required to use API NSCalendar.components(in timezone:fromDate:)

since NSCalendar is mutable, we cannot simply return 'null'. Create new copy, update fields like 'locale', 'timezone' and return copy.

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -167,7 +167,13 @@ public class Calendar: NSObject, NSCopying, NSSecureCoding {
     }
     
     public func copy(with zone: NSZone? = nil) -> AnyObject {
-        NSUnimplemented()
+        let copy = Calendar(calendarIdentifier: calendarIdentifier)!
+        copy.locale = locale
+        copy.timeZone = timeZone
+        copy.firstWeekday = firstWeekday
+        copy.minimumDaysInFirstWeek = minimumDaysInFirstWeek
+        copy._startDate = _startDate
+        return copy
     }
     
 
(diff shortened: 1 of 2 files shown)


Foundation/NSCalendar.swift
```diff
@@ -163,7 +163,10 @@ public class NSCalendar : NSObject, NSCopying, NSSecureCoding {
     }
     
     public func copyWithZone(_ zone: NSZone) -> AnyObject {
-        NSUnimplemented()
+        let copy = NSCalendar(calendarIdentifier: calendarIdentifier)!
+        copy.locale = locale
+        copy.timeZone = timeZone
+        return copy
```
parkera: What about firstWeekday and minimumDaysInFirstWeek?

mbvreddy: @parkera 

CFCalendarGetFirstWeekday, CFCalendarGetMinimumDaysInFirstWeek which are called to populate firstWeekday, minimumDaysInFirstWeek fields derives it from 

> _cfobject->cal

 initialised based on identifier, locale and timezone. So, it is not required to set these explicitly in copyWithZone()

mbvreddy:
resolved conflicts after rebasing.

Foundation/NSCalendar.swift
phausler: I think the question was if a consumer of the api mutates those values via assigning to the setter of firstWeekday etc, the copy should reflect those mutations.

It seems to be missing
```swift
copy.firstWeekday = firstWeekday
copy.minimumDaysInFirstWeek = minimumDaysInFirstWeek
copy._startDate = _startDate
```

mbvreddy: ah, right! I missed mutation of those fields. I've changed implementation to consider above point and added unit testcase to explicitly check those fields.

parkera:
@swift-ci please test

parkera:
Looks like the build failure is tangential. This should be ok to go soon.

mbvreddy:
@parkera build failure with ci doesn't seems to be due to this fix. is it solved?

parkera:
Let's go ahead, this one should be ok.

PR #410, opened 2016-06-12
CoreFoundation: correct export attributes

compnerd:
This adjusts the headers and the definitions to ensure that the export
decorations match.  This enables building CoreFoundation with hidden visibility
by default on Linux targets.

### DIFF
--- a/CoreFoundation/Base.subproj/CFBase.c
+++ b/CoreFoundation/Base.subproj/CFBase.c
@@ -779,7 +779,7 @@ void CFAllocatorGetContext(CFAllocatorRef allocator, CFAllocatorContext *context
     context->preferredSize = __CFAllocatorGetPreferredSizeFunction(&allocator->_context);
 }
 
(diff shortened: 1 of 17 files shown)


modocache:
@swift-ci please test

compnerd:
CC @parkera

parkera:
So this changes the default visibility to hidden as well?

compnerd:
Yes, it does change the default visibility to hidden as well.  Should I change that separately?

parkera:
Yah, let's split that out. Traditionally we haven't done it, but I'm not sure if there was any really strong reasoning that applies to this port.

compnerd:
@parkera okay, rebased and removed the build side support.  This now only corrects the declaration/definitions for the export'ed visibility.

parkera:
Thanks! Let's get a test going.

parkera:
@swift-ci please test

compnerd:
@parkera anything else that needs to be done for this?

parkera:
Nope. Thanks!

PR #411, opened 2016-06-12
CoreFoundation: clean up CF_CONST_STR macro definition

compnerd:
There were six different variants with minor tweaks to the macro definition.
Use a few more macros to abstract out the retain count initializer, the tag
name, and the const-ness of the struct.  This allows the variants to be
collapsed into a single definition.


modocache:
@swift-ci please test

compnerd:
CC @parkera

PR #413, opened 2016-06-13
`fileprivate` updates

CodaFi:
Updates Foundation to use the new `fileprivate` access control level to ensure compatibility with SE-0025 and apple/swift#3000.


parkera:
This was just a find/replace, right? Can we wait on this until after we merge the gigantic swift-3 branch?

CodaFi:
Essentially.  I couldn't find you guys using inner classes and any chicanery that would require full private cases like SwiftPM.

parkera:
I think we have some on the swift-3 branch, so we'll need to revisit this after merge.

CodaFi:
Definitely.  The proposal needs to be amended in the meantime, so the timing is perfect.

PR #414, opened 2016-06-14
Fix performance bottleneck with __CFGetConverter

mbvreddy:
This PR is to resolve performance bottleneck while converting strings using NSString.data(using:) API concurrently. Performance degradation is exponential to no.of concurrent conversion. This API boils down to __CFGetConverter() function of CFStringEncodingConverter.c

__CFGetConverter returns converter which is costly operation based on the encoding passed. Hence, this function cache converter(s) in static map/dictionary and static double pointer for common case like kCFStringEncodingUTF8, kCFStringEncodingMacRoman | kCFStringEncodingWindowsLatin1, __defaultEncoding. So, it make use of lock(s) to avoid multiple threads working with static variables.

```
const _CFEncodingConverter **commonConverterSlot = NULL;
static _CFEncodingConverter *commonConverters[3] = {NULL, NULL, NULL}; // UTF8, MacRoman/WinLatin1, and the default encoding*
static CFMutableDictionaryRef mappingTable = NULL;
```

Below is the high-level operations under lock(s) :

```
lock
    lookup converter for common case converters (commonConverterSlot)
    lookup converter from static MutableDictionary (mappingTable)
unlock

if converter is null, 
    lock
        lookup converter for common case converters (commonConverterSlot)
        lookup converter from static MutableDictionary (mappingTable)
        if converter is null,
            compute converter and update MutableDictionary or static double pointer
        end if
    unlock
end if
```

Performance bottleneck is due to first lookup for converter. This should be relaxed and allow read operation without lock since lookup is done anyway within lock before computing converter. So, it now becomes

```
//lock
    lookup converter for common case converters (commonConverterSlot)
    lookup converter from static MutableDictionary (mappingTable)
//unlock

if converter is null, 
    lock
        lookup converter for common case converters (commonConverterSlot)
        lookup converter from static MutableDictionary (mappingTable)
        if converter is null,
            compute converter and update MutableDictionary or static double pointer
        end if
    unlock
end if
```

Let me know if any comments.

### DIFF
--- a/CoreFoundation/StringEncodings.subproj/CFStringEncodingConverter.c
+++ b/CoreFoundation/StringEncodings.subproj/CFStringEncodingConverter.c
@@ -585,7 +585,7 @@ static const _CFEncodingConverter *__CFGetConverter(uint32_t encoding) {
     const _CFEncodingConverter **commonConverterSlot = NULL;
     static _CFEncodingConverter *commonConverters[3] = {NULL, NULL, NULL}; // UTF8, MacRoman/WinLatin1, and the default encoding*
     static CFMutableDictionaryRef mappingTable = NULL;
-    static OSSpinLock lock = OS_SPINLOCK_INIT;
+    static pthread_rwlock_t rwlock = PTHREAD_RWLOCK_INITIALIZER;
 
     switch (encoding) {
 	case kCFStringEncodingUTF8: commonConverterSlot = (const _CFEncodingConverter **)&(commonConverters[0]); break;
@@ -603,15 +603,15 @@ static const _CFEncodingConverter *__CFGetConverter(uint32_t encoding) {
 	default: if (CFStringGetSystemEncoding() == encoding) commonConverterSlot = (const _CFEncodingConverter **)&(commonConverters[2]); break;
     }
 
-    OSSpinLockLock(&lock);
+    pthread_rwlock_rdlock(&rwlock);
     converter = ((NULL == commonConverterSlot) ? ((NULL == mappingTable) ? NULL : (const _CFEncodingConverter *)CFDictionaryGetValue(mappingTable, (const void *)(uintptr_t)encoding)) : *commonConverterSlot);
-    OSSpinLockUnlock(&lock);
+    pthread_rwlock_unlock(&rwlock);
 
     if (NULL == converter) {
         const CFStringEncodingConverter *definition = __CFStringEncodingConverterGetDefinition(encoding);
 
         if (NULL != definition) {
-            OSSpinLockLock(&lock);
+            pthread_rwlock_wrlock(&rwlock);
             converter = ((NULL == commonConverterSlot) ? ((NULL == mappingTable) ? NULL : (const _CFEncodingConverter *)CFDictionaryGetValue(mappingTable, (const void *)(uintptr_t)encoding)) : *commonConverterSlot);
 
             if (NULL == converter) {
@@ -625,7 +625,7 @@ static const _CFEncodingConverter *__CFGetConverter(uint32_t encoding) {
 		    *commonConverterSlot = converter;
 		}
             }
-            OSSpinLockUnlock(&lock);
+            pthread_rwlock_unlock(&rwlock);
         }
     }
 


michael-lehew:
Hi Bhaktavatsal,

Thanks for the PR.  I am a member of the Foundation/CoreFoundation team, and have a few comments: 

1. This PR is essentially changing this code to be an example of the double-checked-locking pattern, which is rife with peril in the simplest cases; but especially so when it involves mutable collections
2. It is not safe to read from a mutable collection that might be being mutated, especially hash based structures 
3. I am dubious of the perf impact here; in the non-contended case, the cost is minimal; in the contended case, it will depend on platform specifics, but on Darwin platforms, its not that bad.  This can be mitigated further on Darwin platforms where we can/should switch this spin lock away to the new os_.* locks announced this year.  
4. Have you written a microbench that can demonstrate worst case performance and substantiate perf impact of this change? (along with correctness, since any test that was sufficiently powerful enough to confirm perf impact, likely is a good stress test of correctness).  Such a bench would likely crash a fair amount due to 2.
5. I don't see the need to do the double lock in this code -- if it fails, it ought to just do the create/insert right then. Between that, and de-one-lining/ternarying this code, it could be a lot clearer :)

All that said, I think there is value in producing such a micro bench however for this, because it may provide insight into other places we may be able to do better.

mbvreddy:
Hi Michael,

Thank you for your review comments.

For point 1 and 2 - fix in this PR is a case of double-checked-locking pattern. Yes, you are right. It is not safe to read from hash based collection (Dictionary) when other thread is mutating it. However, fix can be modified to read from memory without lock and read Dictionary inside lock. relaxed to read Dictionary under lock and perform read from memory without lock. This will not have any side effects.

```
if commonConverterSlot != NULL
    lookup converter for common case converters
else
    lock
        lookup converter from static MutableDictionary (mappingTable)
    unlock
endif
```

For point 3, Performance impact is visible only in contended case, multiple blocks accessing this code concurrently.

For point 4, simple benchmark which tests conversion using 'NSUTF8StringEncoding' is uploaded to https://github.com/mbvreddy/swift-performance-benchmark/blob/master/ConverterTest/ConvTest.swift. This shows significant time spent in __CFGetConverter especially when converting small strings concurrently. This benchmark do not test lookup from Dictionary as encoding is 'NSUTF8StringEncoding'.

For point 5, I think, double lock is there to avoid other threads from waiting on lock to retrieve different converter (which might have already available in cache) while preparing definition for missing converter. Restricting lookup from Dictionary under lock would fetch avoid contention for common encoders like NSUTF8StringEncoding

Please let me know your views on this approach.

mbvreddy:
```
perf record numactl --cpunodebind=0 --membind=0 ./ConvTest -c 16 -l 20
Queueing 16 blocks
Go!
^C[ perf record: Woken up 33 times to write data ]
[ perf record: Captured and wrote 10.105 MB perf.data (264105 samples) ]

perf report | grep CFGetConverter
     50.95%  ConvTest  libFoundation.so              [.] __CFGetConverter
```

After code change as proposed in my earlier update (use locks only to read from Dictionary)
```
perf record numactl --cpunodebind=0 --membind=0 ./ConvTest -c 16 -l 20
Queueing 16 blocks
Go!
^C[ perf record: Woken up 36 times to write data ]
[ perf record: Captured and wrote 10.773 MB perf.data (281603 samples) ]

perf report | grep CFGetConverter
     0.78%  ConvTest  libFoundation.so              [.] __CFGetConverter
```

michael-lehew:
Hi, 

I think the memory access is mostly OK to sans the lock -- I see that for the common cases we are going through a lot of effort to ensure that each of the three slots are lazily initialized-once.  

So though it'd be uncommon, there remains a concurrency hazard on the bootstrap of that look-aside table.  I think we need a memory barrier, or switch this to make use of C11 atomics.  As I alluded above, I feel this function in general could use a modernization/clarity refactor, perhaps not for this PR.  Friends don't let friends deeply nest ternaries! :)

To possibly uncover the need for the barrier, you'd need to restart the process (since its write-once, and the window for the issue will be very narrow). I also see that perf is using a single encoding.   Which is probably a pretty standard use-case for most cases, but as far as good 'all-case' improvement bench could stand to be improved. 

That said, the improvement/numbers do look good for what it is measuring, so that's promising. 

Thank you again for steadfastness on this, I think we will get somewhere here that will make a great impact, and I think we are close.

mbvreddy:
Working to improve benchmark for multiple encoding.

mbvreddy:
@michael-lehew I made progress on this PR. Improved benchmark did not show expected improvement with any of the fix made/proposed in this PR i.e., open access while reading (or) open access while reading from memory and read from mutable collection under lock.

Since the write into memory/mutable collection happens only once per encoder, amount of writes are very very less compared to reads. Ofcourse, reads should be synchronized with writes. So, I've used pthread_rwlocks which is perfect fit here. Multiple threads can read simultaneously when there is no write operation and thereby contention is less.

Results of benchmark with pthread_rwlocks:

```
perf record ./ConvTest --concurrency 16 --length 10
Queueing 16 blocks
Go!
^C[ perf record: Woken up 190 times to write data ]
[ perf record: Captured and wrote 49.960 MB perf.data (1309274 samples) ]


perf report | grep __CFGetConverter
     0.09%  ConvTest  libFoundation.so              [.] __CFGetConverter
```

Test results without any fix:

```
record ./ConvTest_old --concurrency 16 --length 10
Queueing 16 blocks
Go!
^C[ perf record: Woken up 70 times to write data ]
[ perf record: Captured and wrote 18.934 MB perf.data (495994 samples) ]

perf report | grep __CFGetConverter
    46.90%  ConvTest_old  libFoundation.so              [.] __CFGetConverter 
```

mbvreddy:
I'll squash the changes to use read-write locks instead of spin locks soon.

mbvreddy:
on a side note, the benchmark written for this issue has unearthed a bug with __CFStringEncodeByteStream() which return incorrect value due to uninitialized local variables being used in arithmetic operations. I'll open separate PR for new issue. :)

parkera:
Thanks @mbvreddy !

mbvreddy:
Changes are squashed and ready for review.

mbvreddy:
@michael-lehew - Any comments on the new changes?

Also, are there any guidelines to contribute this benchmark testcase to swift/benchmark?

michael-lehew:
Sorry for missing this -- my notifications are borked (and I was heads down on a few other things and didn't notice).  I'll look this over now.

michael-lehew:
Changes LGTM lets make this so.  I think we should take this for Darwin too -- @parkera  how do changes best go in that direction?

michael-lehew:
@swift-ci please test linux and merge

michael-lehew:

michael-lehew:
(I guess the regex didn't like the linux part :))

ianpartridge:
Can we get this in Swift 3? Is a separate PR against `swift-3.0-branch` needed?

michael-lehew:
swift-corelibs-foundation hasn't branched yet, so this is sufficient for Swift 3.

parkera:
@michael-lehew Best way to make sure we get this for Darwin inside the system CF is to file a radar to make sure this change is picked up.

PR #416, opened 2016-06-17
Add TestNSTextCheckingResult file to Xcode project

wilmerpaulino:
I've also reordered some of the files in the Tests group within TestFoundation to keep the alphabetical order consistent.

@modocache Also, somewhat unrelated, there are a lot of unused value warnings when running these tests. Is this something we don't usually worry about?

### DIFF
--- a/TestFoundation/TestNSTextCheckingResult.swift
+++ b/TestFoundation/TestNSTextCheckingResult.swift
@@ -18,7 +18,7 @@
 #endif
 
 class TestNSTextCheckingResult: XCTestCase {
-    static var allTests: [(String, TestNSTextCheckingResult -> () throws -> Void)] {
+    static var allTests: [(String, (TestNSTextCheckingResult) -> () throws -> Void)] {
         return [
            ("test_textCheckingResult", test_textCheckingResult),
         ]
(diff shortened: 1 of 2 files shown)


TestFoundation/TestNSTextCheckingResult.swift
```diff
@@ -18,7 +18,7 @@
 #endif
 
 class TestNSTextCheckingResult: XCTestCase {
-    static var allTests: [(String, TestNSTextCheckingResult -> () throws -> Void)] {
+    static var allTests: [(String, (TestNSTextCheckingResult) -> () throws -> Void)] {
```
modocache: Was this change also necessary? Just want to make sure this didn't accidentally slip in.

modocache:
@swift-ci please test

TestFoundation/TestNSTextCheckingResult.swift
wilmerpaulino: Oops, sorry I forgot to mention that! Without the parentheses, there's the following warning:

"Single argument function types require parentheses"

modocache: Cool, thanks!

PR #418, opened 2016-06-21
Base: share the module map

compnerd:
The modulemap is identical across Linux and non-Linux (not surprising given that
it reflects the project itself).  Rather than duplicating the module, share it.


phausler:
Looks good to me

phausler:
@swift-ci Please test

compnerd:
@phausler anything else that needs to be done for this?

PR #419, opened 2016-06-22
Update README.md

matthew-an:
Correct a sentence.


parkera:
I'm not sure I understand the problem with the original sentence.

PR #422, opened 2016-06-24
Enable libdispatch usage in TestFoundation

seabaylea:
This adds a further update to build.py to make the libdispatch libraries available to TestFoundation.

This is done using adding the libdispatch build path to `LD_LIBRARY_PATH`, extending the approach that's already used to provide access to the XCTest libraries during build time.


modocache:
Excited to see progress on this, thank you!

@swift-ci please test

modocache:
I don't always merge pull requests to corelibs-foundation, but when I do they're NFC's to the build system that unblock pull requests like https://github.com/apple/swift/pull/3191. 😉

PR #423, opened 2016-06-24
Implementation of HTTPCookie.cookies(withResponseHeaderFields:forURL:)

pushkarnk:
This is for an initial implementaion of HTTPCookie.cookies(withResponseHeaderFields:forURL:).
In addition to the contributed unit tests, I have also tested this code against cookies
returned by some of the twenty most popular websites. 

Kindly review and provide feedback.

### DIFF
--- a/Foundation/NSHTTPCookie.swift
+++ b/Foundation/NSHTTPCookie.swift
@@ -69,6 +69,11 @@ public class HTTPCookie : NSObject {
     let _version: Int
     var _properties: [String : Any]
 
+    static let _attributes: [String] = [NSHTTPCookieName, NSHTTPCookieValue, NSHTTPCookieOriginURL, NSHTTPCookieVersion,
+                                        NSHTTPCookieDomain, NSHTTPCookiePath, NSHTTPCookieSecure, NSHTTPCookieExpires, 
+                                        NSHTTPCookieComment, NSHTTPCookieCommentURL, NSHTTPCookieDiscard, NSHTTPCookieMaximumAge,
+                                        NSHTTPCookiePort] 
+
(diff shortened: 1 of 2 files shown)


Foundation/NSHTTPCookie.swift
```diff
@@ -327,16 +332,106 @@ public class HTTPCookie : NSObject {
...
+            }
+        }
     
+        return httpCookies
+    } 
+
+    //Bake a cookie
+    private class func createHttpCookie(url: URL, pairs: [String], start: Int, end: Int) -> HTTPCookie? {
+        var properties: [String:Any] = [:]
+        for index in start..<end {
```
parkera: Any reason not to just use `for aPair in pairs`?

Foundation/NSHTTPCookie.swift
```diff
@@ -459,3 +554,19 @@ public class HTTPCookie : NSObject {
         return _portList
     }
 }
+
+//utils for cookie parsing
+internal extension String {
```
parkera: A style note, but I'd rather just put these functions inline at the call site than extending String in the entire project.

parkera:
This looks great, thanks. I'm going to kick off a test run while I review further.

parkera:
@swift-ci please test

Foundation/NSHTTPCookie.swift
parkera: Oh I see, we're parsing a subset of the array. I think it might be cleaner to just pass an ArraySlice into this.

pushkarnk:
Thanks, I'll continue working on the improvements you suggested

PR #424, opened 2016-06-26
Activate the ability to integrate libdispatch

seabaylea:
@parkera @phausler This uncomments the block that configures the use of libdispatch by Foundation in build.py, activating the `__HAS_DISPATCH__` guard.

The block includes guards to check that libdispatch is being built, which means that the toolchain builds in the CI environment won't be affected by this yet. 

It will affect the libdispatch CI build, which also builds Foundation, XCTest and Swift PM, so we'll start to exercise the use of libdispatch via TestFoundation and the XCTest and Swift PM unit tests.

If the libdispatch CI builds run clean/stable, we'll then look at adding libdispatch into the toolchain builds via the "buildbot_linux" preset.

### DIFF
--- a/build.py
+++ b/build.py
@@ -68,19 +68,19 @@ if "XCTEST_BUILD_DIR" in Configuration.current.variables:
 foundation.LDFLAGS += '-lpthread -ldl -lm -lswiftCore -lxml2 '
 
 # Configure use of Dispatch in CoreFoundation and Foundation if libdispatch is being built
-#if "LIBDISPATCH_SOURCE_DIR" in Configuration.current.variables:
-#	foundation.CFLAGS += " "+" ".join([
-#		'-DDEPLOYMENT_ENABLE_LIBDISPATCH',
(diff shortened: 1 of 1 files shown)


parkera:
Ok - so we don't need to wait for libdispatch to be in Swift CI before merging this right?

parkera:
@swift-ci please test

phausler:
My latest pulls this afternoon had it building for Linux iirc

parkera:
Cool, sounds like we're almost there.

seabaylea:
Merging this will add libdispatch into Foundation (and thereby into Swift PM and XCTest) in the following Swift CI builds:
    https://ci.swift.org/view/All/job/oss-swift-incremental-RA-libdispatch-linux-ubuntu-15_10/

If those run clean (i.e., the tests for Foundation, XCTest and SwiftPM pass), within just need to add libdispatch into the full package builds.

parkera:
Thanks @seabaylea!

PR #429, opened 2016-06-29
Fix use of LIBS_DIRS in build.ninja file for ninja test

seabaylea:
PR#424 introduced the use of the `LIBS_DIRS` variable as part of the configuration for TestFoundation. Unfortunately as a local variable it doesn't get passed down to ninja.build.

Adding `LIBS_DIRS` to the configuration dictionary adds it to ninja.build so that `ninja test` works correctly.

### DIFF
--- a/build.py
+++ b/build.py
@@ -434,11 +434,13 @@ foundation.add_phase(plutil)
 
 script.add_product(foundation)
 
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci please test

PR #432, opened 2016-07-04
Update protocol compositions to SE-0095 syntax

joewillsher:
Update swift-corelibs-foundation to syntax introduced from [SE-0095](https://bugs.swift.org/browse/SR-1938), merge once [the compiler support is merged](https://github.com/apple/swift/pull/3293).


parkera:
LGTM, someone can merge when the other support has landed.

PR #437, opened 2016-07-08
[NSLocation] Drop needless word and change to class var

damuellen:
Changed from .currentLocation() to .current to match darwin import

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -177,7 +177,7 @@ public class Calendar: NSObject, NSCopying, NSSecureCoding {
     }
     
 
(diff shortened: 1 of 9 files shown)


parkera:
@swift-ci please test

parkera:
Looks like NSString needs an update:

```
Foundation/NSString.swift:509:146: error: '!=' produces 'Bool', not the expected contextual result type 'Locale?'
        return range(of: str, options: [.caseInsensitive, .diacriticInsensitive], range: NSMakeRange(0, length), locale: Locale.current.location != NSNotFound)
```

damuellen:
Sorry but I have create a new pr to fix this, I messed up my fork.

PR #438, opened 2016-07-08
[NSFileManager] Updated to match Darwin version

damuellen:
I hope that I have this time do not forget anything, and the tests will pass.

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -404,7 +404,7 @@ extension NSData {
         if fd == -1 {
             throw _NSErrorWithErrno(errno, reading: false, path: dirPath)
         }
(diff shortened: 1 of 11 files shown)


Foundation/NSData.swift
```diff
@@ -468,7 +468,7 @@ extension NSData {
                 } catch let err {
                     if let auxFilePath = auxFilePath {
                         do {
-                            try FileManager.default().removeItem(atPath: auxFilePath)
+                            try FileManager.default.removeItem(atPath: auxFilePath)
```
parkera: Does this work or do you actually need to use the backticks here?

damuellen:
#484 does the same, so I close this one.

PR #442, opened 2016-07-11
Implementing (initial) of Stream and InputStream.

mamabusi:


Foundation/NSStream.swift
```diff
@@ -118,19 +139,34 @@ public class InputStream: Stream {
     
     // returns YES if the stream has bytes available or if it impossible to tell without actually doing the read.
     public var hasBytesAvailable: Bool {
-        NSUnimplemented()
+        return CFReadStreamHasBytesAvailable(_stream)
     }
     
     public init(data: Data) {
-        NSUnimplemented()
+        let nsData = data._bridgeToObjectiveC()
```
mamabusi: Using NSData as an intermediate to convert the Data object to the expected CFData. Currently, this is the available option. Found a work item which says so: https://bugs.swift.org/browse/SR-1797

mamabusi:
This PR has common changes with:
https://github.com/apple/swift-corelibs-foundation/pull/443 for the below mentioned files.

TestFoundation/TestNSStream.swift
TestFoundation/main.swift
Foundation/NSStream.swift

CoreFoundation/Stream.subproj/CFStream.h
```diff
@@ -78,6 +78,8 @@ const CFStringRef kCFStreamPropertyDataWritten;
 CF_EXPORT
 CFReadStreamRef CFReadStreamCreateWithBytesNoCopy(CFAllocatorRef alloc, const UInt8 *bytes, CFIndex length, CFAllocatorRef bytesDeallocator);
 
+CF_EXPORT CFReadStreamRef CFReadStreamCreateWithData(CFAllocatorRef alloc, CFDataRef data);
```
phausler: This probably should be a private method, putting it here makes it public API which I am not certain we really want to do just yet. Can you move this declaration to ForFoundationOnly.h or ForSwiftFoundationOnly.h?

Foundation/NSStream.swift
```diff
@@ -60,17 +64,17 @@ public func <(lhs: Stream.PropertyKey, rhs: Stream.PropertyKey) -> Bool {
...
-    
+
     public override init() {
-        
+
     }
     
     public func open() {
-        NSUnimplemented()
+        streamOpen()
```
phausler: This can be NSRequiresConcreteImplementation and calls can just be simplified to open and close without needing to override an internal method

Foundation/NSStream.swift
phausler: There is an internal helper property for this: `data._cfObject`

mamabusi:
@phausler Thanks for reviewing the changes. Have addressed the comments. Please take a look at the modified code.

Foundation/NSStream.swift
```diff
@@ -7,6 +7,10 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+#if os(Linux)
+import CoreFoundation
```
phausler: This should be for all operating systems not just Linux

phausler:
Other than the #if it looks good

mamabusi:
@phausler #if has been removed.

phausler:
@swift-ci Please test

LoganWright:
@mamabusi really excited to see this work in upcoming snapshots!

PR #445, opened 2016-07-12
Update for SE-0025 ('private' and 'fileprivate')

jrose-apple:
[SE-0025](https://github.com/apple/swift-evolution/blob/master/proposals/0025-scoped-access-level.md) makes `private` restrict access to a lexical scope; the Swift 2 behavior is now named `fileprivate`. At the time of this pull request, the compiler treats both `private` and `fileprivate` as having `fileprivate` semantics, but that will soon change.

Supersedes #413; the latest amendment to SE-0025 permits some accesses that the original model (or at least our interpretation of it) would have forbidden. This was done by hand, so it's possible I missed a few cases, but it's at least a smaller diff.

### DIFF
--- a/Foundation/Boxing.swift
+++ b/Foundation/Boxing.swift
@@ -14,7 +14,7 @@
 ///
 /// Note: This assumes that the result of calling copy() is mutable. The documentation says that classes which do not have a mutable/immutable distinction should just adopt NSCopying instead of NSMutableCopying.
 internal final class _MutableHandle<MutableType : NSObject where MutableType : NSCopying> {
-    private var _pointer : MutableType
+    fileprivate var _pointer : MutableType
     
     init(reference : MutableType) {
         _pointer = reference.copy() as! MutableType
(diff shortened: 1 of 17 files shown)


phausler:
Looks good to me, I wonder if we should actually consider some of these for `internal` versus `fileprivate` is there a performance differential (either compilation time or execution time?)

phausler:
Good to merge as soon as the binary drops of the toolchain have this feature

jrose-apple:
Both `fileprivate` and `private` are a bit more optimizable than `internal`, though that matters less under WMO. Compilation may also be slightly faster, but probably not in any significant way.

jrose-apple:
Today's toolchain should have 'fileprivate' support. Okay to merge?

parkera:
@swift-ci please test

parkera:
CI is in an awful state; let's make sure we are not making things worse.

jrose-apple:
Good call. Looks like it passed—good to go?

PR #446, opened 2016-07-12
Update for SE-0112: Rename ErrorProtocol to Error

DougGregor:


parkera:
LGTM as a first step; we should probably just rename that error enum in the test to something else.

PR #449, opened 2016-07-15
Foundation: address a number of warnings

compnerd:
This is a mechanical change addressing a number of
  "warning: expected 'let' in conditional"
warnings from the current compiler.  NFC.


parkera:
@swift-ci please test

PR #450, opened 2016-07-15
Fix IndexSet intersection and symmetric difference

xwu:
Earlier in the week, I opened a [PR](https://github.com/apple/swift/pull/3486) to fix IndexSet intersection and symmetric difference in the apple/swift repo. I guess it should be fixed here too. I've ported over the salient tests as well.


parkera:
Thanks @xwu.

parkera:
@swift-ci please test

PR #451, opened 2016-07-15
Updating to account for SE-0118

dabrahams:
See [SE-0118](https://github.com/apple/swift-evolution/blob/master/proposals/0118-closure-parameter-names-and-labels.md).

I expect this patch may not fix all the problems.


parkera:
@swift-ci please test

dabrahams:
The commits in swift that do this have been reverted because I didn’t get Foundation completely updated, so I wouldn’t expect it to work anymore.  I have to set up a linux testing environment so I can get everything right.  At that point I’ll probably push the whole wad at once without pull requests, if you don’t mind.  Otherwise, there’s little hope of making it through CI.

dabrahams:
Committed, hopefully synchronously with other necessary changes, in 1d26dfd5df837d25c40ac1f25fd0cf31fa8536a7

PR #457, opened 2016-07-20
Fix _length access on CF_CONST_STRING for s390x

garyliu1:
Previous s390x PR #386 has a change to make the `_length` field of `CF_CONST_STRING` to `uint64_t`.  This requires a subsequent change to access the field correctly in `_NSCFConstantString`.

This change has been tested and verified to have no regression on x86-64.  On s390x, we see a good pass rate jump including TestNSURL, TestNSCalendar and TestNSNotificationQueue

### DIFF
--- a/Foundation/NSCFString.swift
+++ b/Foundation/NSCFString.swift
@@ -63,11 +63,24 @@ internal final class _NSCFConstantString : _NSCFString {
         let ptr = Unmanaged.passUnretained(self).toOpaque()
         return ptr.load(fromByteOffset: offset, as: UnsafePointer<UInt8>.self)
     }
-    internal var _length : UInt32 {
-        let offset = MemoryLayout<OpaquePointer>.size + MemoryLayout<Int32>.size + MemoryLayout<Int32>.size + MemoryLayout<_CFInfo>.size + MemoryLayout<UnsafePointer<UInt8>>.size
-        let ptr = Unmanaged.passUnretained(self).toOpaque()
-        return ptr.load(fromByteOffset: offset, as: UInt32.self)
+
+    private var _lenOffset : Int {
+        return MemoryLayout<OpaquePointer>.size + MemoryLayout<Int32>.size + MemoryLayout<Int32>.size + MemoryLayout<_CFInfo>.size + MemoryLayout<UnsafePointer<UInt8>>.size
     }
+
+    private var _lenPtr :  UnsafeMutableRawPointer {
+        return Unmanaged.passUnretained(self).toOpaque()
+    }
+
+#if arch(s390x)
+    internal var _length : UInt64 {
+        return _lenPtr.load(fromByteOffset: _lenOffset, as: UInt64.self)
+    }
+#else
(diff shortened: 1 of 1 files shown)


Foundation/NSCFString.swift
```diff
@@ -62,11 +62,20 @@ internal final class _NSCFConstantString : _NSCFString {
...
+    internal var _length : UInt64 {
+        let offset = sizeof(OpaquePointer.self) + sizeof(Int32.self) + sizeof(Int32.self) + sizeof(_CFInfo.self) + sizeof(UnsafePointer<UInt8>.self)
+        let ptr = unsafeAddress(of: self) + offset
+        return UnsafePointer<UInt64>(ptr).pointee
+    }
+#else
     internal var _length : UInt32 {
         let offset = sizeof(OpaquePointer.self) + sizeof(Int32.self) + sizeof(Int32.self) + sizeof(_CFInfo.self) + sizeof(UnsafePointer<UInt8>.self)
         let ptr = unsafeAddress(of: self) + offset
         return UnsafePointer<UInt32>(ptr).pointee
```
parkera: Can we more tightly scope this conditional code, so as not to duplicate the rest?

garyliu1: Thanks for the comment @parkera, how about something like this instead?
```
    private var _lenPtr : UnsafePointer<Void> {
        let offset = sizeof(OpaquePointer.self) + sizeof(Int32.self) + sizeof(Int32.self) + sizeof(_CFInfo.self) + sizeof(UnsafePointer<UInt8>.self)
        return unsafeAddress(of: self) + offset
    }

#if arch(s390x)
    internal var _length : UInt64 {
        return UnsafePointer<UInt64>(_lenPtr).pointee
    }
#else
    internal var _length : UInt32 {
        return UnsafePointer<UInt32>(_lenPtr).pointee
    }
#endif
```

parkera:
I guess that's better. What makes me nervous is having fundamental things like the size of that field depend on platform. It's 32 bits even on other 64 bit platforms, so if we start adding a lot of special cases like this all over the place then maintenance will be extremely difficult.

Do you anticipate this difference in the definition of the constant CFString size to affect any other places?

parkera:
@swift-ci please test

garyliu1:
Thanks @parkera.   s390x and ppc64 (not ppc64le) are the ones I know that need 64 bit.  I did a sanity search for other uses of the _length field and didn't spot any other occurrences that needs attention.

parkera:
@swift-ci please test

parkera:
Looks like an unrelated failure in swiftpm.

garyliu1:
@parkera, I've resolved the conflict with a modified change.  Can you re-issue a test request please? 

Thanks

parkera:
@swift-ci please test linux

parkera:
yup!

parkera:
Tested manually

garyliu1:
@parkera  Thanks Tony!

PR #458, opened 2016-07-21
[SR-2133] archive to streams in XML, not OpenStep, format

lhoward:
_Note: I have not had a chance to test this fix, I just noticed it whilst glancing through the code recently._

_writeXMLData() in NSKeyedArchiver invokes CFPropertyListWrite() with kCFPropertyListOpenStepFormat, which not only does not make sense (given the name of the calling function) but will also assert fail in CF. Presumably it should be changed to kCFPropertyListXMLFormat_v1_0.


parkera:
@swift-ci please test

parkera:
LGTM

PR #459, opened 2016-07-21
[SE-0121] remove uses of optional comparison operators

jtbandes:
These operators will be removed as part of SE-0121 and apple/swift#3637.

This use site was discovered by CI: https://ci.swift.org/job/swift-PR-Linux/2900/console

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -831,8 +831,8 @@ public class Calendar: NSObject, NSCopying, NSSecureCoding {
         let comp2 = components(reducedUnits, from: date2)!
     
         for unit in units {
-            let value1 = comp1.value(forComponent: unit)
-            let value2 = comp2.value(forComponent: unit)
+            let value1 = comp1.value(forComponent: unit)!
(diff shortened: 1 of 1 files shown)


jtbandes:
FYI for whoever works on this code, I think there might be a bug: it seems this function never actually returns nil, because this code doesn't use `_getter()`: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/DateComponents.swift#L224-L226

jtbandes:
Hi @parkera — what's the etiquette for getting stuff merged into corelibs-foundation? Is there a CI setup that can build/test for us?

parkera:
@swift-ci please test

parkera:
Same CI as the rest of swift.

parkera:
That file is mostly shared with the Swift overlay - does the same error exist there?

jtbandes:
It doesn't look like this method exists in the overlay. https://github.com/apple/swift/blob/eb27bb65a7c17bd9b4255baee5c4e4f9c214bde6/stdlib/public/SDK/Foundation/DateComponents.swift

jtbandes:
Thanks!

PR #461, opened 2016-07-21
implementation for NSURLComponents.copy(with:)

mbvreddy:
This PR has following:

1. Implementation for NSURLComponents.copy(with:). Approach is to create default NSURLComponent and update fields one by one.

2. Override isEqual() function. Since copy is not 'self', it is required to override isEqual to compare properties instead of object references. 

3. Unit test for copy(). I've used NSURLComponents class in testcase instead of URLComponents structure because it doesn't have mapping for copy() function.

### DIFF
--- a/Foundation/NSURL.swift
+++ b/Foundation/NSURL.swift
@@ -929,9 +929,49 @@ open class NSURLComponents: NSObject, NSCopying {
     open override func copy() -> Any {
         return copy(with: nil)
     }
-    
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        if let other = object as? NSURLComponents {
+            if scheme != other.scheme {
+                return false
+            }
+            if user != other.user {
+                return false
+            }
+            if password != other.password {
+                return false
+            }
+            if host != other.host {
+                return false
+            }
+            if port != other.port {
+                return false
+            }
+            if path != other.path {
+                return false
+            }
+            if query != other.query {
+                return false
+            }
+            if fragment != other.fragment {
+                return false
+            }
+            return true
+        }
+        return false
+    }
+
     open func copy(with zone: NSZone? = nil) -> Any {
-        NSUnimplemented()
+        let copy = NSURLComponents()
+        copy.scheme = self.scheme
+        copy.user = self.user
+        copy.password = self.password
+        copy.host = self.host
+        copy.port = self.port
+        copy.path = self.path
+        copy.query = self.query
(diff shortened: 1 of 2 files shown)


Foundation/NSURL.swift
```diff
@@ -858,9 +858,49 @@ public class NSURLComponents: NSObject, NSCopying {
...
+            if query != other.query {
+                return false
+            }
+            if fragment != other.fragment {
+                return false
+            }
+            return true
+        }
+        return false
+    }
```
parkera: CFURLComponentsRef implements an equality checking function. Can we just use CFEqual?

Foundation/NSURL.swift
```diff
@@ -858,9 +858,49 @@ public class NSURLComponents: NSObject, NSCopying {
...
+        let copy = NSURLComponents()
+        copy.scheme = self.scheme
+        copy.user = self.user
+        copy.password = self.password
+        copy.host = self.host
+        copy.port = self.port
+        copy.path = self.path
+        copy.query = self.query
+        copy.fragment = self.fragment
+        return copy
```
parkera: CFURLComponentsRef also has `_CFURLComponentsCreateCopy`. Can we just use that and set the CFURLComponentsRef ivar directly?

mbvreddy: @parkera

_components are set in init's. There is no variant of init that takes CFURLComponentsRef. 

So, I tried creating NSURLComponents using default init() and updated _components with _CFURLComponentsCreateCopy (after changing _components to var). But, this results in memory leak.

Foundation/NSURL.swift
mbvreddy: I tried to use it. But, couldn't figure out how to cast    CFURLComponentsRef to CFTypeRef thay CFEqual expect. NSURLComponent doesn't support toll-free bridge.

phausler: NSURLComponents is a has-a not an is-a class so for subclasses this approach is by far safer imho

Foundation/NSURL.swift
phausler: Again the same issue above applies to subclassers; @mbvreddy's approach is correct for the abstract implementation.

mbvreddy:
rebased

parkera:
@swift-ci please test linux

CodaFi:
#590 Fixed at least one of these failures

@swift-ci please test Linux.

PR #465, opened 2016-07-22
Implementing JSONSerialization.writeJSONObject

saiHemak:


Foundation/NSJSONSerialization.swift
```diff
@@ -159,7 +159,13 @@ public class JSONSerialization : NSObject {
     /* Write JSON data into a stream. The stream should be opened and configured. The return value is the number of bytes written to the stream, or 0 on error. All other behavior of this method is the same as the dataWithJSONObject:options:error: method.
      */
     public class func writeJSONObject(_ obj: AnyObject, toStream stream: NSOutputStream, options opt: WritingOptions) throws -> Int {
-        NSUnimplemented()
+        do {
```
saiHemak: This API differs from the Darwin version,the latter takes a fourth  argument of type ```NSErrorPointer``` which is  a typealias unavailable on Linux .The Darwin version also does not throw any error.The spec says that 0 needs to be returned on error.If we return 0 we cannot throw the error.

parkera: I would have expected this code to just call 'try' and not 'catch', therefore propagating the error outward if it happens (both from `data(withJSONObject:options:)` and `stream.write`).

saiHemak:
@parkera ,I have addressed your comments please review .

parkera:
Thanks!

parkera:
Unfortunately this crossed paths with the stdlib change to rename all of the pointer types.

parkera:
cc @atrick

atrick:
I can checkin a quick fix if no one has claimed it.

atrick:
commit dd9e4385c7f69dd58d55f2618b8970eed1054f5d (HEAD -> master, origin/master, origin/HEAD)
Author: Andrew Trick <atrick@apple.com>
Date:   Tue Jul 26 15:30:12 2016

    Quick fix for conflicting commits.

PR #466, opened 2016-07-22
Implementing jsonObject with streams api of NSJSONSerialization.

mamabusi:


Foundation/NSJSONSerialization.swift
```diff
@@ -125,8 +125,8 @@ public class JSONSerialization : NSObject {
        The data must be in one of the 5 supported encodings listed in the JSON specification: UTF-8, UTF-16LE, UTF-16BE, UTF-32LE, UTF-32BE. The data may or may not have a BOM. The most efficient encoding to use for parsing is UTF-8, so if you have a choice in encoding the data passed to this method, use UTF-8.
      */
     /// - Experiment: Note that the return type of this function is different than on Darwin Foundation (Any instead of AnyObject). This is likely to change once we have a more complete story for bridging in place.
-    public class func jsonObject(with data: Data, options opt: ReadingOptions = []) throws -> Any {
```
mamabusi: Implementation NSJSONSerialization.jsonObject(with: stream) internally needs to call the api NSJSONSerialization.jsonObject(with: data). Currently, return type of NSJSONSerialization.jsonObject(with: data) returns 'Any' instead of 'AnyObject' and it is not possible to cast 'Any' to 'AnyObject'. 

Wrote a query to the mailing list regarding the above hiccup. Following is the response from Philippe Hausler:
"We had changed the return type of JSONSerialization to be Any such that it avoided the caller needing to bridge types (which works poorly on linux). Even though this is an API change technically, just change the return type to match the other JSONSerialization methods."

@phausler: In accordance with the above suggestion, modified the return type and the related code (including the existing test cases) to resolve errors caused by the change.

parkera:
I think we wound up with two implementations of the JSON stream methods. @mamabusi , I like the test cases from this - can you rebase vs master so we can merge this?

mamabusi:
@parkera , I see NSJSONSerialization.writeObject(with streams) implementation is merged but not NSJSONSerialization.jsonObject(with streams). This PR implements the latter. 
Is there an implementation already for this as well? 

I have resolved the conflicts. Please review. 
Thanks!

mamabusi:
Rebased to resolve the conflicts.

mamabusi:
@parkera This PR implements an unimplemented method in NSJSONSerialization. Could you please review.

Foundation/NSJSONSerialization.swift
```diff
@@ -184,8 +184,20 @@ open class JSONSerialization : NSObject {
...
+        var bytesRead: Int?
+        repeat {
+            bytesRead = 0
+            if stream.hasBytesAvailable {
+                bytesRead = stream.read(&buffer, maxLength: buffer.count)
+                if bytesRead! > 0 {
+                    data.append(&buffer, count: bytesRead!)
+                }
+            }
+        } while bytesRead == buffer.count
```
phausler: shouldn't this condition be `while stream.status == .open || stream.status == .reading`?

Foundation/NSJSONSerialization.swift
```diff
@@ -184,8 +184,20 @@ open class JSONSerialization : NSObject {
...
-    open class func jsonObject(with stream: InputStream, options opt: ReadingOptions = []) throws -> AnyObject {
-        NSUnimplemented()
+    open class func jsonObject(with stream: InputStream, options opt: ReadingOptions = []) throws -> Any {
+        var buffer = [UInt8](repeating: 0, count: 1024)
+        var data = Data()
+        var bytesRead: Int?
+        repeat {
+            bytesRead = 0
+            if stream.hasBytesAvailable {
+                bytesRead = stream.read(&buffer, maxLength: buffer.count)
```
phausler: if the bytesRead is less than zero it probably should throw an error derived from the streamError

parkera:
@swift-ci please test

mamabusi:
streamError implementation is now available. Will work on the review comments and resolution of conflicts.

mamabusi:
@parkera @phausler Have addressed the previous review comments. Please have a look.
Thanks!

Foundation/NSJSONSerialization.swift
```diff
@@ -202,8 +202,24 @@ open class JSONSerialization : NSObject {
     
     /* Create a JSON object from JSON data stream. The stream should be opened and configured. All other behavior of this method is the same as the JSONObjectWithData:options:error: method.
      */
-    open class func jsonObject(with stream: InputStream, options opt: ReadingOptions = []) throws -> AnyObject {
-        NSUnimplemented()
+    open class func jsonObject(with stream: InputStream, options opt: ReadingOptions = []) throws -> Any {
+        var data = Data()
+        guard stream.streamStatus == .open || stream.streamStatus == .reading else {
+            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: [
```
parkera: This should actually be `fatalError` to match the Darwin behavior.

Foundation/NSJSONSerialization.swift
```diff
@@ -202,8 +202,24 @@ open class JSONSerialization : NSObject {
...
+                "NSDebugDescription" : "Stream is not available for reading."
+                ])
+        }
+        repeat {
+            var buffer = [UInt8](repeating: 0, count: 1024)
```
parkera: Is it possible to use a `Data` here instead?

TestFoundation/TestNSJSONSerialization.swift
```diff
@@ -956,11 +956,524 @@ extension TestNSJSONSerialization {
...
+    fileprivate func removeTestFile(_ location: String) {
         do {
             try FileManager.default.removeItem(atPath: location)
         } catch _ {
             
         }
     }
 }
+
+//MARK: - JSONObjectWithStreams
```
parkera: Ok, I appreciate the extended test coverage here - but there has got to be a better way than copy/pasting all the existing test cases and just passing a stream with the same content.

Can we perhaps make a helper function which tests both data and stream versions and modify the existing tests?

Foundation/NSJSONSerialization.swift
mamabusi: Ok..

Foundation/NSJSONSerialization.swift
mamabusi: I tried to use Data instead of the buffer here,  but got stuck at the point where I am supposed to send in the UnsafeMutablePointer and an Int of maxLength to read the contents of the stream:

var dataChunk = Data()
try dataChunk.withUnsafeMutableBytes {(bytes: UnsafeMutablePointer<UInt8>)->Void in
                bytesRead = stream.read(bytes, maxLength: ??)

Is there a way I could read the InputStream to Data without the buffer?

TestFoundation/TestNSJSONSerialization.swift
mamabusi: Modifying the existing test functions to include both JSONObject with data and with stream would look something like this:

` //MARK: - Object Deserialization
    func test_deserialize_emptyObject() {
        let subject = "{}"
        do {
            guard let data = subject.data(using: .utf8) else {
                XCTFail("Unable to convert string to data")
                return
            }
            // Test with Data
            let t = try JSONSerialization.jsonObject(with: data, options: [])
            let resultWithData = t as? [String: Any]
            //Test with Stream
            let inputStream: InputStream = InputStream(data: data)
            inputStream.open()
            let resultWithStream = try JSONSerialization.jsonObject(with: inputStream, options: []) as? [String: Any]
            inputStream.close()
            XCTAssertEqual(resultWithData?.count, 0)
            XCTAssertEqual(resultWithStream?.count, 0)
        } catch {
            XCTFail("Error thrown: \(error)")
        }
    }
`

Would something like this be ok?

mamabusi:
@parkera Have addressed the changes requested. 
Thanks!

parkera:
Thanks!

parkera:
@swift-ci test and merge

PR #470, opened 2016-07-25
Reserve capacity when creating the storage Dictionary for NSDictionary.

gonzalolarralde:
The allocation of `NSDictionary` will now make use of the capacity and count parameters provided in the constructors.

For `NSMutableDictionary`'s `init(capacity:)` we need to reinstantiate the storage with a new `minimumCapacity` due to the lack of a `reserveCapacity` instance method like the one that `Array` provides.

An option considered was to change the required initializer of `NSDictionary` to a method that creates the underlying `Array` with the desired capacity, but this won't make any sense in the immutable `NSDictionary` context.


parkera:
Seems reasonable.

parkera:
@swift-ci please test

gonzalolarralde:
@parkera rebased.

parkera:
Tested manually

PR #471, opened 2016-07-26
Fixed compilation errors:

amraboelela:
1. Converting @noreturn to -> Never
2. for NSCalender.swift added ! at the end to unwrap the value1 and value2


amraboelela:
I realized that those 2 fixes are already fixed by other developers, but I just didn't fetch the upstream repo.

PR #472, opened 2016-07-26
Update for SE-0107: Migrate Void->Raw.

atrick:
Unsafe[Mutable]Pointer is replaced by Unsafe[Mutable]RawPointer.

Migrating often means simply removing casts to/from UnsafePointer
and working directly with UnsafeRawPointer instead.

Unfortunately some API's, like NSData still want UnsafePointer,
which requires "binding" memory to UInt8 before passing it to the API.

### DIFF
--- a/Foundation/Data.swift
+++ b/Foundation/Data.swift
@@ -16,11 +16,11 @@ import Darwin
 import Glibc
 #endif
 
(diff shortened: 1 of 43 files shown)


atrick:
Depends on Swift PR:
https://github.com/apple/swift/pull/3724

PR #473, opened 2016-07-26
[SE-0127] Remove calls to withUnsafe[Mutable]Pointers

xwu:
Groundwork for removing withUnsafe[Mutable]Pointers from stdlib.


parkera:
@swift-ci please test

PR #474, opened 2016-07-26
 [Starter Bug] [SR-1608] NSStream

DaveLiu888:
What's in this pull request?
NSStream "is-a" implementation with tests
Input/Output streams "has-a" implementation with tests

Resolved bug number: 1608 (https://bugs.swift.org/browse/SR-1608)

I have some questions i have outlined in the PR as well


Foundation/NSStream.swift
```diff
@@ -109,24 +107,71 @@ public class Stream: NSObject {
...
     /*@NSCopying */public var streamError: NSError? {
-        NSUnimplemented()
+        NSRequiresConcreteImplementation()
     }
+    
+}
+
+internal final class _SwiftInputStream : NSInputStream, _SwiftNativeFoundationType {
+    internal typealias ImmutableType = NSInputStream
+    internal typealias MutableType = NSInputStream
```
DaveLiu888: not sure about setting MutableType to NSInputStream but only way to satisfy _SwiftNativeFoundationType's constraints

Foundation/NSStream.swift
```diff
@@ -154,17 +205,97 @@ public class InputStream: Stream {
...
+    public override func setProperty(_ property: AnyObject?, forKey key: String) -> Bool {
+        return CFReadStreamSetProperty(_stream, key._cfObject, property)
+    }
+    
+    public override func propertyForKey(_ key: String) -> AnyObject? {
+        return CFReadStreamCopyProperty(_stream, key._cfObject)
     }
+    
+    public func copy(with zone: NSZone?) -> AnyObject{
+        if let data = _data {
```
DaveLiu888: not sure if i am handling the copy and mutableCopy responsibilities correctly could not figure out a way with out caching url and data which leads to extra storage on the heap

Foundation/NSStream.swift
```diff
@@ -202,43 +337,62 @@ public class NSOutputStream : Stream {
...
     public  override func setProperty(_ property: AnyObject?, forKey key: String) -> Bool {
         return CFWriteStreamSetProperty(_stream, key._cfObject, property)
     }
+    
+    public override func propertyForKey(_ key: String) -> AnyObject? {
+        return CFWriteStreamCopyProperty(_stream, key._cfObject)
+    }
+    
+    public func copy(with zone: NSZone?) -> AnyObject{
+        if let url = _url , let shouldAppend = _append {
```
DaveLiu888: not sure if i am handling the copy and mutableCopy responsibilities correctly could not figure out a way with out caching url and shouldAppend and the other instance variables which leads to extra storage on the heap

Foundation/NSStream.swift
```diff
@@ -154,17 +205,97 @@ public class InputStream: Stream {
...
+    
+    public func mutableCopy(with zone: NSZone?) -> AnyObject{
+        return copy(with: nil)
+    }
+
+}
+
+internal final class _SwiftOutputStream : NSOutputStream, _SwiftNativeFoundationType {
+    internal typealias ImmutableType = NSOutputStream
+    internal typealias MutableType = NSOutputStream
```
DaveLiu888: not sure how to satisfy _SwiftNativeFoundationType with out making the MutableType to OutputStream

TestFoundation/TestIOStream.swift
```diff
@@ -0,0 +1,422 @@
...
+
+#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
+    import Foundation
+    import XCTest
+#else
+    import SwiftFoundation
+    import SwiftXCTest
+#endif
+
+class TestIOStream : XCTestCase {
```
DaveLiu888: this is tests for  "has-a"  value type wrapper a lot of the tests overlaps the "is-a" implementation
opinions on weather they are needed are welcome. 

Questions should we test at the wrapper level or at the reference instance level?
Also even at the reference instance level it just wrap another reference instance

Tho testing the wrapper yielded bugs in the existing subsystem

TestFoundation/TestNSStream.swift
```diff
@@ -16,34 +16,43 @@
 #endif
 
 class TestNSStream : XCTestCase {
```
DaveLiu888: this is tests for "is-a" value type wrapper a lot of the tests overlaps the "has-a" implementation
opinions on weather they are needed are welcome. whether to keep both or get rid of some from this set or the "has-a" set

Foundation/NSStream.swift
```diff
@@ -202,43 +337,62 @@ public class NSOutputStream : Stream {
...
+    public func mutableCopy(with zone: NSZone?) -> AnyObject{
+        return copy(with: nil)
+    }
 }
 
 // Discussion of this API is ongoing for its usage of AutoreleasingUnsafeMutablePointer
 #if false
-extension Stream {
+extension NSStream {
     public class func getStreamsToHost(withName hostname: String, port: Int, inputStream: AutoreleasingUnsafeMutablePointer<InputStream?>?, outputStream: AutoreleasingUnsafeMutablePointer<NSOutputStream?>?) {
```
LoganWright: I know the discussion of this api is ongoing, but I think it would be better to not prevent implementation due to a non-finalized api. I know many of us would love to be able to use this functionality on Linux. Here's a simple version I threw together that might satisfy the use case.

```
public class func getStreamsToHost(withName hostname: String, port: Int, inputStream: AutoreleasingUnsafeMutablePointer<InputStream?>?, outputStream: AutoreleasingUnsafeMutablePointer<NSOutputStream?>?) {
        var read: Unmanaged<CFReadStream>? = nil
        var write: Unmanaged<CFWriteStream>? = nil
        CFStreamCreatePairWithSocketToHost(kCFAllocatorDefault, host, UInt32(port), &read, &write)
        inputStream = read?.takeRetainedValue()
        outputStream = write?.takeRetainedValue()
    }
```

DaveLiu888: Thanks for the feed back i will try to work it in to the PR. How ever i accidentally submitted it un rebased and squashed to apple/master instead of my forks's master. will close and resubmit after

DaveLiu888:
i accidentally submitted it un rebased and squashed to apple/master instead of my forks's master. will close and resubmit shortly

Foundation/NSStream.swift
DaveLiu888: Thanks for the feedback ill try to work this in to the PR. i accidentally submitted it un rebased and squashed to apple/master instead of my forks's master. will close and resubmit shortly

PR #485, opened 2016-07-28
change ProcessInfo.processinfo() class func to processinfo static let

jpsim:
to align with recent changes on Darwin (Xcode 8)

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -23,7 +23,7 @@ public func NSTemporaryDirectory() -> String {
         return String(cString: buf, encoding: .utf8)!
     }
     #endif
-    if let tmpdir = ProcessInfo.processInfo().environment["TMPDIR"] {
+    if let tmpdir = ProcessInfo.processInfo.environment["TMPDIR"] {
         if !tmpdir.hasSuffix("/") {
             return tmpdir + "/"
         } else {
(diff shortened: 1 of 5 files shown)


parkera:
@swift-ci please test

jpsim:
FYI I'm sure there are many more of these than the 3 PRs I just filed. A good starter task could be to audit all of the changes in Foundation from Xcode 8 betas and align Swift Foundation with them.

jpsim:
The CI failure seems due to `swift-corelibs-xctest` relying on `processInfo` being a function. Looks like these two changes would have to be made simultaneously. How would you suggest we do this @parkera?

parkera:
Can you put up a PR for XCTest? I'll merge them together.

jpsim:
Sure thing! apple/swift-corelibs-xctest#144

parkera:
Hm, is it possible to temporarily stage this by having both names in Foundation?

jpsim:
> Hm, is it possible to temporarily stage this by having both names in Foundation?

I don't think so, since referring to `ProcessInfo.processInfo` would then be ambiguous because that's also a valid way to refer to the `processInfo()` function.

parkera:
Ok - let's hold off on this particular change for a few days. The rest of Swift is scrambling to land source breaking changes, but per my email to swift-corelibs-dev earlier today, we have a few more weeks. That will keep us out of the line of fire while everything else finishes up.

parkera:
Then we can do a simultaneous merge of both PRs.

jpsim:
happy to wait if it helps

jpsim:
rebased

parkera:
Hopefully we can un-stick the Linux builder today and land this. Thanks for your patience.

parkera:
Going to go ahead and merge these.

PR #486, opened 2016-07-28
Use utf8CString instead of nulTerminatedUTF8

xwu:
Needs to be committed simultaneously with apple/swift PR implementing SE-0134.

### DIFF
--- a/TestFoundation/TestNSXMLParser.swift
+++ b/TestFoundation/TestNSXMLParser.swift
@@ -26,9 +26,9 @@ class TestNSXMLParser : XCTestCase {
     }
     
     func test_data() {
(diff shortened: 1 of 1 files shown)


jrose-apple:
This should use `nullTerminatedUTF8`, not `utf8CString`, no?

xwu:
Core team decision was to remove nullTerminatedUTF8 altogether.

atrick:
@jrose-apple nullTerminatedUTF8 was removed. I always merge the foundation change right before the swift change so the CI passes. But I just found out I'm not authorized to merge anything and couldn't find anyone who could do it.
https://github.com/apple/swift/pull/3816

PR #487, opened 2016-07-28
UnsafePointer conversion fixes.

atrick:
Update for SE-0107: UnsafeRawPointer.

Some of this is pretty rough, but it's necessary to
keep the build working and land the raw vs. typed pointer feature.

As expected there are two major pain points:

1. CString vs. UTF8 types

2. NSData and company use pointer to UInt8 for untyped memory.


atrick:
@parkera This is just a stop-gap to allow the build to succeed when we're ready to pull in the stdlib changes sometime today. I plan to go through over all these changes again later and find a cleaner way to handle things.

atrick:
I'll update this after
https://github.com/apple/swift/pull/3832

Also, I noticed a stdlib dependency which hasn't been checked in yet, so this should not be merged.

parkera:
Ok.

atrick:
This passes 'build-script -R --llbuild --foundation --xctest -T' with the latest swift master.
I need this to proceed with changes on master.
Merging and keeping an eye on the bots.

PR #488, opened 2016-07-28
More updates for SE-0025 ('fileprivate' and 'private')

jrose-apple:
Members satisfying protocol requirements cannot be private, since they are necessarily available to call from outside the type.


jrose-apple:
@swift-ci Please test

jrose-apple:
It fits the Swift 2 notion of private (i.e. fileprivate): the members match the access of the type.

PR #496, opened 2016-07-31
Support internal null characters in NSString initializers [SR-2225]

xwu:
This PR aims to fix a discrepancy in NSString between corelibs-foundation and Apple Foundation. Namely, initialization of an NSString from a byte sequence currently calls `CFStringCreateWithBytes(_:_:_:_:_:)`, which cannot accept null characters. Instead, switch to using `CFStringCreateFromExternalRepresentation(_:_:_:)`.

Resolves SR-2225.

Note: I'm having trouble getting TestFoundation to compile in Xcode at the moment, but I've checked that previous builds of corelibs-foundation fail the test supplied, while Apple Foundation passes that test.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -1221,10 +1221,7 @@ extension NSString {
     }
     
     public convenience init?(data: Data, encoding: UInt) {
-        guard let cf = data.withUnsafeBytes({ (bytes: UnsafePointer<UInt8>) -> CFString? in
-            return CFStringCreateWithBytes(kCFAllocatorDefault, bytes, data.count, CFStringConvertNSStringEncodingToEncoding(encoding), true)
-        }) else { return nil }
-        
+        guard let cf = CFStringCreateFromExternalRepresentation(kCFAllocatorDefault, data._cfObject, CFStringConvertNSStringEncodingToEncoding(encoding)) else { return nil }
         var str: String?
         if String._conditionallyBridgeFromObject(cf._nsObject, result: &str) {
             self.init(str!)
@@ -1234,8 +1231,8 @@ extension NSString {
     }
     
     public convenience init?(bytes: UnsafeRawPointer, length len: Int, encoding: UInt) {
-        let bytePtr = bytes.bindMemory(to: UInt8.self, capacity: len)
-        guard let cf = CFStringCreateWithBytes(kCFAllocatorDefault, bytePtr, len, CFStringConvertNSStringEncodingToEncoding(encoding), true) else {
+        let extRep = NSData(bytesNoCopy: UnsafeMutableRawPointer(mutating: bytes), length: len, freeWhenDone: false)._cfObject
+        guard let cf = CFStringCreateFromExternalRepresentation(kCFAllocatorDefault, extRep, CFStringConvertNSStringEncodingToEncoding(encoding)) else {
             return nil
         }
         var str: String?
@@ -1265,12 +1262,10 @@ extension NSString {
             return nil
         }
     }
-
+    
     public convenience init(contentsOf url: URL, encoding enc: UInt) throws {
         let readResult = try NSData(contentsOf: url, options: [])
-
-        let bytePtr = readResult.bytes.bindMemory(to: UInt8.self, capacity: readResult.length)
-        guard let cf = CFStringCreateWithBytes(kCFAllocatorDefault, bytePtr, readResult.length, CFStringConvertNSStringEncodingToEncoding(enc), true) else {
+        guard let cf = CFStringCreateFromExternalRepresentation(kCFAllocatorDefault, readResult._cfObject, CFStringConvertNSStringEncodingToEncoding(enc)) else {
             throw NSError(domain: NSCocoaErrorDomain, code: NSCocoaError.FileReadInapplicableStringEncodingError.rawValue, userInfo: [
                 "NSDebugDescription" : "Unable to create a string using the specified encoding."
                 ])
@@ -1284,7 +1279,7 @@ extension NSString {
                 ])
         }
     }
-
+    
     public convenience init(contentsOfFile path: String, encoding enc: UInt) throws {
         try self.init(contentsOf: URL(fileURLWithPath: path), encoding: enc)
     }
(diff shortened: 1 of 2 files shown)


DaveLiu888:
@xwu how are you building AppFoundation and Tests? Which snapshot are you using and Xcode version. I am having trouble building and testing as well after i update to master from 7/29?

Foundation/NSString.swift
```diff
@@ -1221,10 +1221,7 @@ extension NSString {
     }
     
     public convenience init?(data: Data, encoding: UInt) {
-        guard let cf = data.withUnsafeBytes({ (bytes: UnsafePointer<UInt8>) -> CFString? in
-            return CFStringCreateWithBytes(kCFAllocatorDefault, bytes, data.count, CFStringConvertNSStringEncodingToEncoding(encoding), true)
-        }) else { return nil }
-        
+        guard let cf = CFStringCreateFromExternalRepresentation(kCFAllocatorDefault, data._cfObject, CFStringConvertNSStringEncodingToEncoding(encoding)) else { return nil }
```
parkera: The story of what happens on Darwin here is fairly complicated, which is probably why this was different in the first place, but it boils down to this:

```objc
- initWithData:(NSData *)data encoding:(NSStringEncoding)encoding {
    return (id)CFStringCreateFromExternalRepresentation(kCFAllocatorDefault, (CFDataRef)data, CFStringConvertNSStringEncodingToEncoding(encoding));
}
```

So this change looks correct.

Foundation/NSString.swift
```diff
@@ -1234,8 +1231,8 @@ extension NSString {
     }
     
     public convenience init?(bytes: UnsafeRawPointer, length len: Int, encoding: UInt) {
-        let bytePtr = bytes.bindMemory(to: UInt8.self, capacity: len)
-        guard let cf = CFStringCreateWithBytes(kCFAllocatorDefault, bytePtr, len, CFStringConvertNSStringEncodingToEncoding(encoding), true) else {
+        let extRep = NSData(bytesNoCopy: UnsafeMutableRawPointer(mutating: bytes), length: len, freeWhenDone: false)._cfObject
```
parkera: But this one should remain using the bytes:

```objc
- initWithBytes:(const void *)bytes length:(NSUInteger)len encoding:(NSStringEncoding)encoding {
    return (id)CFStringCreateWithBytes(kCFAllocatorDefault, (const UInt8 *)bytes, len, CFStringConvertNSStringEncodingToEncoding(encoding), true);
}
```

Foundation/NSString.swift
```diff
@@ -1265,12 +1262,10 @@ extension NSString {
...
         }
     }
-
+    
     public convenience init(contentsOf url: URL, encoding enc: UInt) throws {
         let readResult = try NSData(contentsOf: url, options: [])
-
-        let bytePtr = readResult.bytes.bindMemory(to: UInt8.self, capacity: readResult.length)
-        guard let cf = CFStringCreateWithBytes(kCFAllocatorDefault, bytePtr, readResult.length, CFStringConvertNSStringEncodingToEncoding(enc), true) else {
+        guard let cf = CFStringCreateFromExternalRepresentation(kCFAllocatorDefault, readResult._cfObject, CFStringConvertNSStringEncodingToEncoding(enc)) else {
```
parkera: This actually calls through to `initWithBytes:length:encoding:`.

Foundation/NSString.swift
xwu: That's interesting. I'm not in front of a computer at the moment, but IIRC, testing on OS X showed that Apple Foundation `NSString.init(bytes:length:encoding:)` proceeded past an internal null character. I will check again later in the day and report back.

parkera: It may depend on exactly how you're calling it. Those are from the placeholder string, which is the result of calling alloc on NSString, but not on a subclass of NSString.

xwu: I'm not quite sure I understand what you mean by placeholder string. However, it's clear that Apple Foundation reads past the internal null character with `NSString.init(bytes:length:encoding:)`, as demonstrated below:

```bash
echo -e "This is a string.\0This is another string." > /Users/xwu/Desktop/test.txt
```

```swift
let url = URL(fileURLWithPath: "/Users/xwu/Desktop/test.txt")
let data = try Data(contentsOf: url)
data.withUnsafeBytes { (bytes: UnsafePointer<UInt8>) in
  let string = NSString(bytes: bytes, length: data.count, encoding: String.Encoding.ascii.rawValue)
  // "This is a string.\0This is another string."
}
```

Not sure what it would take to achieve this result if we are to use `CFStringCreateWithBytes` in corelibs-foundation. Likewise with the original bug regarding `init(contentsOf: URL, encoding: UInt)`: I'm not sure how to interpret the discrepancy if Apple Foundation is doing nothing more than calling through to `CFStringCreateWithBytes`.

parkera:
I debugged this a bit. I'm not finished yet, but I think the discrepancy has to do with the bridging happening after the CFString initialization. The CFString call returns the same result on both platforms (reading past the NULL).

xwu:
Wow, that's neat. `CFStringGetLength()` is showing that there's more characters returned from `CFStringCreateWithBytes` than what's printed after bridging.

For now, I'll create another PR with the change for `init?(data:encoding:)`. I can think of maybe some hacky ways to get the rest to work, but since you're already debugging I'll defer to you.

xwu:
@parkera I bet you've found it too, but things go awry during bridging because, if `type(of: x) == _NSCFString.self`, bridging is performed by calling `CFStringGetCStringPtr(_:_:)` followed by `String(cString:)`. If the cost isn't too great, one solution would be to use this result only if `result.utf16.count == CFStringGetLength(cf)`.

xwu:
@DaveLiu888 Sorry for the delayed reply. I put off hacking on corelibs-foundation directly for a few days. Everything is now building fine for me with the 8/7 snapshot (although I did have to disable the NSString reflection test, which segfaults).

parkera:
I actually didn't get too much further on this (we've been busy getting all of those API changes landed).

DaveLiu888:
@xwu thanks for the update yea everything is working for me as well. 👍

xwu:
@parkera OK, I will forge ahead and see how far I get with this. Since the solution is taking a different turn, I'll close this PR and open another one once I have something presentable. I'll roll in the `init?(data:encoding:)` change there as well.

parkera:
Cool, thanks.

PR #499, opened 2016-08-01
Mechanically add 'open' to all public classes.

jrose-apple:
This applies `open` to all public classes and all of their non-final public methods, properties, and subscripts, preserving the existing behavior when [SE-0117](https://github.com/apple/swift-evolution/blob/master/proposals/0117-non-public-subclassable-by-default.md) is implemented. This transformation was performed mechanically using the following sed script (run with `sed -E`):

    /(init|mutating|struct|enum|typealias|final|let|static)/ {
      p
      d
    }
    /public class/,/^}/ s/public/open/

This should probably be refined soon, but for now this allows us to implement SE-0117 without immediately requiring updates in all clients. Specifically, there are two areas where this change may not be correct to match Foundation on Apple platforms:

- Members in extensions cannot yet be overridden, so none of them have been marked `open`.
- Some classes are *not* intended to be overridden, even in Objective-C. Specifying this requires input from the Foundation team.

Groundwork for SE-0117 "Allow distinguishing between public access and public overridability"

### DIFF
--- a/Foundation/NSAffineTransform.swift
+++ b/Foundation/NSAffineTransform.swift
@@ -31,16 +31,16 @@ public struct NSAffineTransformStruct {
     }
 }
 
-public class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
+open class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
     
-    public func encode(with aCoder: NSCoder) {
+    open func encode(with aCoder: NSCoder) {
         NSUnimplemented()
     }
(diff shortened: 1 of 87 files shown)


jrose-apple:
cc @parkera, @rjmccall 

@swift-ci Please test

jrose-apple:
@swift-ci Please test

jrose-apple:
Ah, this may not work without apple/swift-corelibs-xctest#150.

rjmccall:
I've pulled the xctest change, so we'll just merge this as well.  Thanks!

PR #503, opened 2016-08-05
Sets correct locale for date parsing in HTTPCookie

itaiferber:
As mentioned in https://developer.apple.com/library/ios/qa/qa1480/_index.html, the correct locale to use when parsing "Internet-style dates" is "en_US_POSIX", which is strongly specified, and date- and time-invariant. We should *not* be defaulting to using the user's current locale, which is what DateFormatter currently does.

This fix also side-steps an issue where the test and cookie expiration parsing were failing on macOS because the bundled version of CFLite defaults to a locale identifier of "", failing to parse the "MMM" format specifier. This test did not fail on Linux or on the regular CoreFoundation project because on those platforms, CFDateFormatter defaults to a locale of "en_US" (still wrong, but it works); here, a locale of "" caused the failing test. A fix for this incorrect default is forthcoming in CF itself.

### DIFF
--- a/Foundation/NSHTTPCookie.swift
+++ b/Foundation/NSHTTPCookie.swift
@@ -259,6 +259,7 @@ open class HTTPCookie : NSObject {
                 _expiresDate = date
             } else if let dateString = expiresProperty as? String {
                 let formatter = DateFormatter()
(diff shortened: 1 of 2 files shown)


phausler:
@swift-ci Please test Linux

phausler:
@swift-ci Please test Linux

PR #504, opened 2016-08-05
More value types

phausler:
Value types for Calendar, Locale, and TImeZone as well as a build script change for building CF standalone


phausler:
@swift-ci please test linux

parkera:
This was done in 1d1ddba.

PR #506, opened 2016-08-07
Move where clauses to end to adjust to new syntax

parkera:
Apply the fix-its from the compiler to adopt the new `where` clause syntax.


parkera:
@swift-ci smoke test linux

parkera:
@swift-ci test linux

PR #523, opened 2016-08-10
Additional fixes for macOS test failure in TestHTTPCookie.

parkera:
The root cause of the problem is missing logic in the CFTimeZone.c file
to handle "GMT" (no offset) as a time zone, which Darwin Foundation
handles. Also, CFLocale.c was using an empty string for the current locale, due
to not checking the AppleLocale preference. For now, we hard code en_US
as the current locale, although obviously this should be fixed to use
CFPreferences if possible.

### DIFF
--- a/CoreFoundation/Locale.subproj/CFLocale.c
+++ b/CoreFoundation/Locale.subproj/CFLocale.c
@@ -13,8 +13,6 @@
     Responsibility: David Smith
 */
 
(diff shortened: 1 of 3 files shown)


parkera:
@swift-ci please test linux

seabaylea:
@parkera FYI, I think the test run for this got lost in yesterdays shutdown - theres no sign of it in the build queue.

parkera:
I think I'll just merge this.

PR #524, opened 2016-08-10
NSKeyedArchiver, NSKeyedUnarchiver now use Any instead of AnyObject.

parkera:
This also contains another kind of fix for the NSKeyedArchiver tests for macOS, which were failing because of assumptions made in the implementation about AnyObject. To help reduce the amount of casting in the implementation, this commit adds a private NS "toll free bridged" type for the private
CF type used as a keyed archiver UID in CFPropertyList.

### DIFF
--- a/Foundation/NSKeyedArchiverHelpers.swift
+++ b/Foundation/NSKeyedArchiverHelpers.swift
@@ -1,0 +1,47 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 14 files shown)


parkera:
@phausler can you code review?

Foundation/NSKeyedArchiverHelpers.swift
```diff
@@ -0,0 +1,47 @@
...
+import CoreFoundation
+
+extension CFKeyedArchiverUIDRef : _NSBridgable {
+    typealias NSType = _NSKeyedArchiverUID
+    
+    internal var _nsObject: NSType { return unsafeBitCast(self, to: NSType.self) }
+}
+
+internal class _NSKeyedArchiverUID : NSObject {
+    typealias CFType = CFURL
```
parkera: Well this is wrong. I'll fix that shortly.

parkera:
Tested manually

PR #536, opened 2016-08-12
Updates to enable "swift test" in Linux

amraboelela:
This update is to enable the use of “swift test” in Linux


parkera:
Is there a way to do this without moving all of our tests into another directory?

amraboelela:
That's why I am making links to the test files whenever possible.

parkera:
But then we also have things like TestNSBundle, which looks like it just moved into a different location?

I'm a bit confused about what the goal is here. Is `swift test` really ready to support our test scenarios? Bundle, for example, requires a specific layout of the resulting app.

parkera:
Also, we need to maintain the ability to run tests on Mac using our Xcode project.

amraboelela:
Yes that's why I copied TestNSBundle (and couple other files) and commented out the parts which give compilation errors.

amraboelela:
The current TestFoundation folder I believe is for testing using Xcode in MacOS
That should still work

So TestFoundation folder for MacOS and Tests for Linux or command line swift test in general

And files in Tests are links to files in TestFoundation

parkera:
I don't want to have copies of our test files in two different locations in the repo. I also don't want to put a bunch of stub files in another directory so `swift test` can work. What we need is the ability to tell `swift test` (via configuration) where the test files actually are. If it doesn't have that capability, then we should block adoption on that. I also don't want stub structs with "hello world" in them (why do we need that?).

Even after we did all of this, what's the concrete benefit of using `swift test` vs our existing infrastructure? We should have a clear net positive from reworking our infrastructure around it, but right now I don't see what it is, exactly. We don't use `swift build` to build, so `swift test` would not be smart enough to rebuild the dependencies as required. Our current scripts do that, at least.

amraboelela:
Using swift test will make it easy for developers to test the framework in Linux. This solution can get improved later. And I can remove the "hello world" line.

parkera:
@amraboelela I see you're continuing with the same approach, for example copying more files into another subdirectory. I want to be clear: I won't merge this without addressing some of the concerns I listed above.

parkera:
Let's have a further discussion about this on the mailing list. Once we have a direction, we'll re-open the PR and figure out the right patch.

PR #538, opened 2016-08-13
Fix Data(capacity:) and Data(count:) to return non-optional

spevans:
- This matches the Foundation API in Darwin


parkera:
@swift-ci please test linux

parkera:
CI is finally green!

PR #539, opened 2016-08-13
Reduce warnings when building Foundation

parkera:
Also, upgrade Xcode project file to silence a project warning there.


parkera:
Tested manually.

PR #542, opened 2016-08-14
Add missing Data init() methods to match Foundation API

spevans:
- Add init(buffer: UnsafeMutableBufferPointer<SourceType>)

- Replace internal init(_bridged data: NSData) with
  public init(referencing reference: NSData)


michael-lehew:
This is subsumed by: https://github.com/apple/swift-corelibs-foundation/pull/553    We wanted to get all of Data/NSData converted in one go (most of the team right now is going through all of the Darwin wift 3 and bringing it over)

PR #543, opened 2016-08-14
Fix CoreFoundation standalone build syntax errors

copumpkin:
Not sure how these trailing quotes crept in, but the file doesn't parse unless you remove them.

Also, the standalone build still doesn't work for me unless I add some hackery to bring `<CoreFoundation/CFAsmMacros.h>` into the search path. I've worked around it in my build but I don't know what the proper solution is so I'm only including the syntax fixes until that's sorted out.

cc @phausler @CodaFi

### DIFF
--- a/CoreFoundation/build.py
+++ b/CoreFoundation/build.py
@@ -13,10 +13,10 @@ elif Configuration.current.target.sdk == OSType.MacOSX:
 	cf.CFLAGS = '-DDEPLOYMENT_TARGET_MACOSX '
 
 cf.ASFLAGS = " ".join([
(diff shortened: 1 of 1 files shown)


CodaFi:
You can insert `Base.subproj/CFAsmMacros.h` into the private part of CoreFoundation's `build.py`.

copumpkin:
@CodaFi oh thanks! that did it. Should I submit a separate pull request to fix that?

CodaFi:
Naw, this one works just fine.

copumpkin:
@CodaFi done 😄 thanks again!

CodaFi:
I've verified that this fix builds locally on OS X and Linux.  Merging.

phausler:
Awesome, thanks for the cleanup!

PR #544, opened 2016-08-15
[Swift 3.0] Fix _length access on CF_CONST_STRING for s390x

vivkong:
This is the same fix from https://github.com/apple/swift-corelibs-foundation/pull/457.  We'd like to include this in Swift 3.0.  Thanks.

@garyliu1 FYI


parkera:
We're going to mass merge master into swift-3.0-branch, or figure out some other integration strategy. The work that's going on right now in master is for Swift 3.

vivkong:
Ah I see.  Thanks for clarifying @parkera!  I'll close this.

PR #550, opened 2016-08-16
Fix heap overflow in TestNSJSONSerialization

spevans:
- The capacity passed was larger than the buffer length causing
  NSOutputStream.write() to write past the end of the buffer

- Found using AddressSanitizer


parkera:
@swift-ci please test linux

parkera:
Thanks!

PR #553, opened 2016-08-17
Swift 3 API Parity:  Data & NSData

michael-lehew:
Brings the Data and NSData API in s-cl-f to match the public version of Swift3 on Darwin. Reorganizes the implementation of Data/NSData to be a bit saner (versus somewhat vestigially matching the Objective-C categories on Darwin).  Also brings a whole host of tests over from our overlay (some disabled for now, since the real focus of this PR is the API parity, which we should have now).


michael-lehew:
This got a bit larger than I wanted because the disorganization in NSData.swift was making it hard to keep track of what I had already done. Further, the initializers defined in extensions was impacting ability to define subclasses and the like.

michael-lehew:
@swift-ci please test linux

michael-lehew:
@swift-ci please test linux

michael-lehew:
I'm going to redo this in a different way + branch (this commit was getting too messy, and now that Philippe has landed his id  -> Any changes) a few things end up different. Standby.

PR #555, opened 2016-08-17
Id to any

phausler:
Migrate collection classes and assocated APIs from id to Any with a boxing storage.

Convert NSArray, NSDictionary, and NSSet for AnyObject to Any.
Remove the `.bridge()` function!
Re-work the tests to no longer need `.bridge()` calls. This is over 400 unique places where `.bridge()` was removed.
Bring in a number of NSString APIs for String since it helps avoid calls to `.bridge()`.
remove _expensivePropertyListConversion since we now have a better tool for the job: _SwiftValue.fetch


phausler:
@swift-ci Please test Linux

Foundation/NSAffineTransform.swift
```diff
@@ -406,3 +406,11 @@ extension AffineTransform {
...
     }
 }
+
+extension NSAffineTransform : _StructTypeBridgeable {
+    public typealias _StructType = AffineTransform
+    
+    public func _bridgeToSwift() -> AffineTransform {
+        return AffineTransform._unconditionallyBridgeFromObjectiveC(self)
+    }
+}
```
itaiferber: Need newline at EOF here

Foundation/NSArray.swift
```diff
@@ -219,15 +201,15 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
-                descriptions.append(string._swiftObject)
-            } else if let array = obj as? NSArray {
-                descriptions.append(array.description(withLocale: locale, indent: level + 1))
-            } else if let dict = obj as? NSDictionary {
-                descriptions.append(dict.description(withLocale: locale, indent: level + 1))
+            let obj = self[idx]
+            if let string = obj as? String {
+                descriptions.append(string)
+            } else if let array = obj as? [Any] {
+                descriptions.append(NSArray(array: array).description(withLocale: locale, indent: level + 1))
```
itaiferber: This seems inefficient — is there a way to improve on it, or was this change necessary because of type/bridging changes?

Foundation/NSArray.swift
```diff
@@ -171,44 +151,46 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
     open func componentsJoined(by separator: String) -> String {
         // make certain to call NSObject's description rather than asking the string interpolator for the swift description
-        return bridge().map() { ($0 as! NSObject).description }.joined(separator: separator)
+        return allObjects.map { "\($0)" }.joined(separator: separator)
     }
 
-    open func contains(_ anObject: AnyObject) -> Bool {
-        let other = anObject as! NSObject
+    open func contains(_ anObject: Any) -> Bool {
+        guard let other = anObject as? AnyHashable else {
```
itaiferber: Why the requirement on AnyHashable here? Is it not enough to just use Equatable?

Foundation/NSArray.swift
```diff
@@ -247,99 +229,111 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
         }
         
-        
-        
         objects += range.toCountableRange()!.map { self[$0] }
     }
     
-    open func index(of anObject: AnyObject) -> Int {
+    open func index(of anObject: Any) -> Int {
+        guard let val = anObject as? AnyHashable else {
```
itaiferber: Same as the AnyHashable/Equatable comment above

Foundation/NSArray.swift
```diff
@@ -247,99 +229,111 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
+                    return idx
+                }
             }
         }
         return NSNotFound
     }
     
-    open func index(of anObject: AnyObject, in range: NSRange) -> Int {
+    open func index(of anObject: Any, in range: NSRange) -> Int {
+        guard let val = anObject as? AnyHashable else {
```
itaiferber: Same as the AnyHashable/Equatable comment above (marking all of these for easier reference later, if need be)

Foundation/NSArray.swift
```diff
@@ -247,99 +229,111 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
         
         for idx in 0..<count {
-            let obj1 = object(at: idx) as! NSObject
-            let obj2 = otherArray[idx] as! NSObject
-            if obj1 === obj2 {
-                continue
-            }
-            if !obj1.isEqual(obj2) {
-                return false
+            if let val1 = object(at: idx) as? AnyHashable,
```
itaiferber: Same as the AnyHashable/Equatable comment above

Foundation/NSAttributedString.swift
```diff
@@ -186,11 +186,11 @@ private extension AttributedString {
     }
 }
 
-extension AttributedString: _CFBridgable {
+extension NSAttributedString: _CFBridgable {
```
itaiferber: `NSAttributedString:` → `NSAttributedString :`

Foundation/NSCFDictionary.swift
```diff
@@ -24,19 +24,19 @@ internal final class _NSCFDictionary : NSMutableDictionary {
...
 
-    required init(objects: UnsafePointer<AnyObject>, forKeys keys: UnsafePointer<NSObject>, count cnt: Int) {
+    required init(objects: UnsafePointer<AnyObject>!, forKeys keys: UnsafePointer<NSObject>!, count cnt: Int) {
         fatalError()
     }
     
-    required convenience init(dictionaryLiteral elements: (NSObject, AnyObject)...) {
-        fatalError()
+    required public convenience init(dictionaryLiteral elements: (Any, Any)...) {
+        fatalError("init(dictionaryLiteral:) has not been implemented")
```
itaiferber: Perhaps better to leave as `NSUnimplemented()`?

Foundation/NSDictionary.swift
```diff
@@ -407,15 +346,18 @@ open class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
...
         if count != otherDictionary.count {
             return false
         }
         
         for key in keyEnumerator() {
-            if let otherValue = otherDictionary[key as! NSObject] as? NSObject {
-                let value = objectForKey(key as! NSObject)! as! NSObject
-                if otherValue != value {
+            if let otherValue = otherDictionary[key as! AnyHashable] as? AnyHashable {
+                if let value = object(forKey: key)! as? AnyHashable {
```
itaiferber: Is it not possible to write:

```swift
if let otherValue = otherDictionary[key as! AnyHashable] as? AnyHashable,
   let value = object(forKey: key) as? AnyHashable,
   otherValue == value {
   continue
} else {
    return false
}
```

Foundation/NSAttributedString.swift
michael-lehew: Typical swift style is actually   TypeName: Protocol   which differs from Objective-C.    

I'd rather we take on style on the whole for slcf in a separate pr. I think @mgrebenets is right -- styles is all over the place, and we'd benefit from something consistent.

itaiferber: I was under the impression that the style tended to be
```swift
TypeName : OtherType {
    ...
}
```

for type declarations and `var name: Type = ...` for variable declarations, but I could be out of date. Across all of our code, this seems to be the most consistent styling, and AFAIR, I think this is how sample Swift code is styled, no? (Again, could certainly be wrong here.)

mgrebenets: In early days I used Apple's own Swift book as a reference. So far the use of `:` seems to be just like in natural English: no space before `:`, one space after. Same as `,`. The only exception being ternary operator `a ? b : c`. Latest documentation is still following this style, e.g. https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Enumerations.html#//apple_ref/doc/uid/TP40014097-CH12-ID145

But then Swift was open sourced and I noticed that there's not much consistency. So these days I wonder what approach the core team and open source community takes. I guess there's tons of really important stuff to do :)

We personally were early adopters of SwiftLint and so far it does a great job for us. I'm curious if open source Swift code will use third party tools or if Apple is planning to come up with this functionality built-in in Xcode.

phausler:
Per the AnyHashable versus equatable; there doesn't seem to be currently a way to hoist the protocol conformance as a concrete implemented type that can be compared. So AnyHashable will have to do for now until we can find a better answer.

phausler:
@swift-ci Please test Linux

Foundation/NSArray.swift
parkera: I agree with Itai - this does not seem correct. On Darwin you can put any struct in there and it is boxed in the SwiftValue box.

phausler: There is currently no facility from the language to do that because you would need to reach into the protocol and grab the `==` function to compare two items with `Self` types which has been destroyed by this point. the only way to obtain a clue approximation is to use `AnyHashable`.

PR #556, opened 2016-08-17
Fix refcount issue in CFDateFormatter

spevans:
- __CFDateFormatterDeallocate() always releases _TimeZone but
  __CFDateFormatterSetProperty() only retains it if the value
  is different

- Triggered by NSDate.description() if used enough times, simple
  regression test to expose it

I think I have the CFRetain() in the correct place but it will definitely need a review by someone more familiar with CFDateFormatter.c. 

Without this fix, calls to NSDate.description will eventually cause a segfault

### DIFF
--- a/Foundation/NSDate.swift
+++ b/Foundation/NSDate.swift
@@ -115,8 +115,8 @@ open class NSDate : NSObject, NSCopying, NSSecureCoding, NSCoding {
      */
     open override var description: String {
         let dateFormatterRef = CFDateFormatterCreate(kCFAllocatorSystemDefault, nil, kCFDateFormatterFullStyle, kCFDateFormatterFullStyle)
-        let timeZone = CFTimeZoneCreateWithTimeIntervalFromGMT(kCFAllocatorSystemDefault, 0.0)
-        CFDateFormatterSetProperty(dateFormatterRef, kCFDateFormatterTimeZoneKey, timeZone)
+        CFDateFormatterSetProperty(dateFormatterRef, kCFDateFormatterTimeZoneKey,
+            CFTimeZoneCreateWithTimeIntervalFromGMT(kCFAllocatorSystemDefault, 0.0))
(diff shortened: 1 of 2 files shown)


lifeissweetgood:
Thanks for this, @spevans! After evaluating this pull request, though, we found that there actually isn't a need to retain value in an else clause because value would only need to be retained if it was being stored in formatter->_property._TimeZone.  Retaining value in an else clause as you have it would up value's refcount unnecessarily; if the caller of __CFDateFormatterSetProperty() released value afterwards, there would be a leak.

However, we have reason to believe an overrelease may in fact be happening, though this may not be the fix.  If you could attach the dump from the segfault, we could help further investigate.

spevans:
@lifeissweetgood yes I wasn't really sure where the CFRetain should go as it was quite tricky to follow the retain/releases. I have attached a log from asan, there wasn't a coredump
[timezone-refcnt.txt](https://github.com/apple/swift-corelibs-foundation/files/423750/timezone-refcnt.txt)
 
thanks

lifeissweetgood:
Thanks @spevans!

lifeissweetgood:
So upon reviewing the log, it looks like the same time zone instance is being reference across multiple instances of CFDateFormatter.  The code for CFDateFormatter looks solid, so I would recommend checking out the description method in NSDate.swift to see what could be going on there.

spevans:
@lifeissweetgood you were correct about the issue being in NSDate.swift, it seems that swift runtime is double releasing the timeZone variable. I did a workaround that fixes the segfault but I haven't found the underlying issue yet.

I have only seen this on linux, Ive not checked Darwin yet as Im still building the latest toolchain to test against

Of course NSDate.description should probably be rewritten to use NSDateFormatter instead of CFDateFormatter but that is a separate issue, but I will see if I can isolate the underlying problem

spevans:
Closing this as I think the underlying issue is a compiler bug, opened https://bugs.swift.org/browse/SR-2462

PR #557, opened 2016-08-17
CharacterSet: minor improvements to match darwin

michael-lehew:
1) we don't actually impl NSCoding methods, so NSUnimplement them
2) punctuationCharacters was incorrectly migrated as punctuation


michael-lehew:

PR #565, opened 2016-08-18
Sort and enumeration labels

phausler:
Update the labels for methods using options in the collections to match Darwin. Additionally rename SortOptions to NSSortOptions to match Darwin names


phausler:

PR #576, opened 2016-08-19
convert all isEqual methods to take Any? instead of AnyObject?

phausler:


phausler:
@swift-ci Please test linux

PR #580, opened 2016-08-22
NSData and struct Data should not have a function called bridge

phausler:
This function was removed on purpose and should not be re-introduced.


phausler:
@swift-ci Please test

PR #581, opened 2016-08-22
Foundation API Conformance Update: Errors

itaiferber:
### Changes
* `NSCocoaError` → `CocoaError`
* Reimplement top-level error values in terms of consistent definitions
* Import new error types (`RecoverableError`, `LocalizedError`) from overlay
* Import URL errors and POSIX errors from overlay
* Import error bridging from overlay
* Update error code usage throughout project

### DIFF
--- a/Foundation/NSError.swift
+++ b/Foundation/NSError.swift
@@ -1,3 +1,5 @@
+//===----------------------------------------------------------------------===//
+//
 // This source file is part of the Swift.org open source project
 //
 // Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
@@ -6,18 +8,25 @@
 // See http://swift.org/LICENSE.txt for license information
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
+//===----------------------------------------------------------------------===//
 
 
 import CoreFoundation
 
+public typealias NSErrorDomain = NSString
+
+/// Predefined domain for errors from most Foundation APIs.
 public let NSCocoaErrorDomain: String = "NSCocoaErrorDomain"
 
+// Other predefined domains; value of "code" will correspond to preexisting values in these domains.
 public let NSPOSIXErrorDomain: String = "NSPOSIXErrorDomain"
 public let NSOSStatusErrorDomain: String = "NSOSStatusErrorDomain"
 public let NSMachErrorDomain: String = "NSMachErrorDomain"
 
+// Key in userInfo. A recommended standard way to embed NSErrors from underlying calls. The value of this key should be an NSError.
 public let NSUnderlyingErrorKey: String = "NSUnderlyingError"
 
+// Keys in userInfo, for subsystems wishing to provide their error messages up-front. Note that NSError will also consult the userInfoValueProvider for the domain when these values are not present in the userInfo dictionary.
 public let NSLocalizedDescriptionKey: String = "NSLocalizedDescription"
 public let NSLocalizedFailureReasonErrorKey: String = "NSLocalizedFailureReason"
 public let NSLocalizedRecoverySuggestionErrorKey: String = "NSLocalizedRecoverySuggestion"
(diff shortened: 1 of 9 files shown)


itaiferber:
@swift-ci Please test

itaiferber:
@DougGregor Any thoughts/concerns about this change? I ported the `NSError` changes almost wholesale from the overlay (sans features requiring silgen and forms of bridging which are not present on Linux).

itaiferber:
Hmm. Looks like a serious compiler crash. Can't replicate on Linux myself, but looking into it.

itaiferber:
Perhaps there was a timing issue involved here. I want to see if re-testing makes any sort of difference. @swift-ci Please test

itaiferber:
(For reference, the failure is still available to view [here](https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/241/console) while CI re-runs)

Foundation/NSError.swift
```diff
@@ -187,29 +196,218 @@ extension CFError : _NSBridgable {
...
+    return result
+}
+
+// NSError and CFError conform to the standard Error protocol. Compiler
+// magic allows this to be done as a "toll-free" conversion when an NSError
+// or CFError is used as an Error existential.
+
+extension NSError {
+    /// The "embedded" NSError is itself.
+    public func _getEmbeddedNSError() -> AnyObject? {
```
DougGregor: `_getEmbeddedNSError` isn't actually used when we don't have an Objective-C runtime, so you can remove this bit.

Foundation/NSError.swift
```diff
@@ -187,29 +196,218 @@ extension CFError : _NSBridgable {
...
+    public var _code: Int {
+        return CFErrorGetCode(self)
+    }
+
+    public var _userInfo: Any? {
+        return CFErrorCopyUserInfo(self) as Any
+    }
+
+    /// The "embedded" NSError is itself.
+    public func _getEmbeddedNSError() -> AnyObject? {
```
DougGregor: Same here

Foundation/NSError.swift
```diff
@@ -187,29 +196,218 @@ extension CFError : _NSBridgable {
...
+        return self
+    }
+}
+
+// An error value to use when an Objective-C API indicates error
+// but produces a nil error object.
+public enum _GenericObjCError : Error {
```
DougGregor: This is only used when bridging NSError, so I don't think you need this in the corelibs Foundation.

Foundation/NSError.swift
```diff
@@ -244,7 +445,909 @@ public extension __BridgedNSError where RawValue: UnsignedInteger {
...
+            return nil
+        }
+
+        self.init(_nsError: error)
+    }
+}
+
+/// Implementation of CustomNSError for all _BridgedStoredNSErrors.
+public extension _BridgedStoredNSError {
+    // FIXME: Would prefer to have a clear "extract an NSError
```
DougGregor: Oops, we should kill this FIXME in the overlay; this is what `getEmbeddedNSError` does.

Foundation/NSError.swift
itaiferber: Figured, but wasn't 100% sure (about this and about `_GenericObjCError`); will remove.

itaiferber:
I doubt this will fix the CI build error, but it's worth a shot. @swift-ci Please test

itaiferber:
@swift-ci Please test

DougGregor:
The compiler crash looks like we're trying to do some bridging without checking for ObjC interop.

DougGregor:
For reference, the rest of this PR looks good to me. Thank you!

itaiferber:
@DougGregor Figured; trying to diagnose, though haven't been able to reproduce on Linux yet, which will make testing expensive if I have to push to CI every time.

Thanks for the help!

itaiferber:
Pushed a change that might avoid the compilation crash we see here.

itaiferber:
@swift-ci Please test

itaiferber:
Looks like that fixed the `hashValue` failure point; the only other location is `isEqual`. Hopefully this change will correct things. @swift-ci Please test

itaiferber:
@swift-ci Please test

itaiferber:
Perfect. Avoiding the (missing) bridging by overriding `NSObject` methods avoids the silgen issue.

PR #582, opened 2016-08-22
Correct the expectation of _SwiftValue isEqual to properly handle the Any case

phausler:


phausler:
@swift-ci Please test Linux

PR #583, opened 2016-08-23
Re-import a few things for struct Data

phausler:
There was a missing protocol and a few overrides that didnt come in right before


phausler:
@swift-ci Please test Linux

PR #584, opened 2016-08-23
Enable DEPLOYMENT_ENABLE_LIBDISPATCH on Darwin hosts for both Foundation and TestFoundation

phausler:


phausler:
@swift-ci Please test

PR #585, opened 2016-08-23
Extract CGFloat definition into it’s own file and re-build the full conformance list for CGFloat

phausler:
There were a few new protocol conformances and annotations to CGFloat that had not yet been imported to swift-corelibs-foundation. This expands the definitions to be more in-line with how CGFloat is defined in the standard library.


phausler:
@swift-ci Please test

PR #586, opened 2016-08-23
Make Darwin targets buildable after merge

phausler:
un-buildable changes were pushed to the project were introduced from URLSession. This broke the builds for Darwin hosts since the Xcode project could no longer build. This adds the URLSession and associated files in as per required for the Xcode project, links libcurl, and renames the nested class which does not work on the latest binary drop of the swift toolchain (the support for which is still pending on master)

### DIFF
--- a/Foundation/NSURLSession/EasyHandle.swift
+++ b/Foundation/NSURLSession/EasyHandle.swift
@@ -22,56 +22,55 @@ import Dispatch
 
 
 
-extension URLSessionTask {
-    /// Minimal wrapper around the [curl easy interface](https://curl.haxx.se/libcurl/c/)
-    ///
-    /// An *easy handle* manages the state of a transfer inside libcurl.
-    ///
-    /// As such the easy handle's responsibility is implementing the HTTP
-    /// protocol while the *multi handle* is in charge of managing sockets and
-    /// reading from / writing to these sockets.
-    ///
-    /// An easy handle is added to a multi handle in order to associate it with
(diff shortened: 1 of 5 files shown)


phausler:
@swift-ci Please test

Foundation/NSURLSession/EasyHandle.swift
```diff
@@ -391,35 +390,35 @@ extension URLSessionTask._EasyHandle {
...
-    static func from(callbackUserData userdata: UnsafeMutableRawPointer?) -> URLSessionTask._EasyHandle? {
+fileprivate extension _EasyHandle {
+    static func from(callbackUserData userdata: UnsafeMutableRawPointer?) -> _EasyHandle? {
         guard let userdata = userdata else { return nil }
-        return Unmanaged<URLSessionTask._EasyHandle>.fromOpaque(userdata).takeUnretainedValue()
+        return Unmanaged<_EasyHandle>.fromOpaque(userdata).takeUnretainedValue()
     }
 }
 
-fileprivate extension URLSessionTask._EasyHandle {
```
phausler: Note: this pattern causes build failures on Darwin with the latest binary toolchain drops.
We cannot adopt this until the toolchain revs to contain this.

parkera: Does the Swift 3.0 toolchain have it or is it only on master?

phausler: I believe it is only master. However I could be mistaken.

PR #587, opened 2016-08-24
Allow C++ to be built from the python build script configuration

phausler:


phausler:
@swift-ci Please test

phausler:
That test failure looks nondeterminisitic. There probably is a bug in the priority ordering

phausler:
@swift-ci please test

PR #589, opened 2016-08-24
[SR-2431] getStreamsToHost

DaveLiu888:
Old pull: #489
Old Pull request was closed and split due to the uncertain nature and unknown dependencies  of this pull 

Discussion on resolved bug number: 2431 (https://bugs.swift.org/browse/SR-2431)
- The potential API change to 

```
open class func getStreamsToHost(withName hostName: String, port: Int, inputStream: inout InputStream?, outputStream: inout NSOutputStream?)
```
- The dependencies on Linux
  @LoganWright mentioned that these [functions](https://github.com/apple/swift-corelibs-foundation/blob/master/CoreFoundation/Stream.subproj/CFSocketStream.c#L158-L160) will need to be enabled/implemented.
  help would be much appreciated on figuring out the dependencies, as i have not had the time to get a Linux build working locally 
- [ ] @phausler 
- [ ] @parkera

### DIFF
--- a/CoreFoundation/Stream.subproj/CFSocketStream.c
+++ b/CoreFoundation/Stream.subproj/CFSocketStream.c
@@ -110,7 +110,7 @@ static struct {
 
 #define CFNETWORK_CALL(sym, args)		((CFNetworkSupport.sym)args)
 
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED
+#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_LINUX
 #define CFNETWORK_LOAD_SYM(sym)   __CFLookupCFNetworkFunction(#sym)
 #elif DEPLOYMENT_TARGET_WINDOWS
 #define CFNETWORK_LOAD_SYM(sym)   (void *)GetProcAddress(CFNetworkSupport.image, #sym)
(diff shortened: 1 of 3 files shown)


phausler:
Why is the test in TestNSNotificationQueue? Seems like it belongs in TestNSStream

DaveLiu888:
@phausler good call it was late last night 😄 totally put it in the wrong place

DaveLiu888:
updated the tests to be in the right place

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -110,7 +110,7 @@ static struct {
```
DaveLiu888: taking a guess here on adding the Linux targets, but assuming ```CFNETWORK_LIBRARY_PATH``` is found from ```__CFgetenvIfNotRestricted```  it should work?
if not we may need to do a 
```
const char *path = __CFgetenvIfNotRestricted("CFNETWORK_LIBRARY_PATH");
	if (!path) {
#if DEPLOYMENT_TARGET_MACOSX  
        path = "/System/Library/Frameworks/CFNetwork.framework/CFNetwork";
#elif DEPLOYMENT_TARGET_LINUX 
        path = ".../Linux/CFNetwork";
#endif
	}
```
@LoganWright can you test this to see if this works, i still don't have Linux build working locally

parkera: Where is CFNetwork itself coming from on Linux?

DaveLiu888: @parkera my apologies i think i am confused with the CFLite framework

DaveLiu888: @parkera so there is no way to get this functionality on Linux? 
 CFNetwork is no longer available as open source https://openradar.appspot.com/7195752

phausler:
The old versions can be found here http://opensource.apple.com//source/CFNetwork/ but that is considerably older than the CF we imported to swift corelibs. It might be a decent reference but I am dubious if it is immediately usable here without a bit of overhauling. And since CF is an implementation detail of swift-corelibs-foundation we could just implement the exact functionality we need to get this done instead of pulling in a few dozen other items.

DaveLiu888:
@phausler thanks for the tip i will use the old open source as a reference and see how far i can get.

robertjpayne:
@DaveLiu888 the only other thing is this API specifically has been deprecated on OSX and has never been available on iOS.

In saying that we need these CFNetwork patches to do any of the CFStreamPair API's.

alblue:

PR #591, opened 2016-08-24
Adding s390x to CGFloat

vivkong:
https://github.com/apple/swift-corelibs-foundation/pull/585 added CGFloat.swift.  I'd like to add s390x as a platform so we can build it.


phausler:
looks good to me, we likely have a few other places that assume these are the only arches.
Seems like we really need a `#if bits(64)` or something like that

PR #593, opened 2016-08-25
Swift 3 API Parity: Drop NS prefix from NSHTTPURLResponse

seabaylea:
The Swift 3 API changes dropped the `NS` prefix from `NSHTTPURLResponse`, but that was missed in the `URLSession` code drop.

This remove the `NS` prefix from the declaration and where its used in both Foundation and TestFoundation.


parkera:
@swift-ci please test linux

seabaylea:
@parkera Is this good to merge? It looks like there might be some other API changes required, so it would be good to get this merged to avoid any change conflicts.

parkera:
Yup, thanks.

PR #594, opened 2016-08-25
Correct NSOrderedSet filtering by unboxing objects

itaiferber:
Our updated collections store boxed values as `_SwiftValue` objects. All implementations of `filtered(using:)` for our other collections use methods which automatically unbox those stored objects (e.g. `self.enumerated()` in mutable variants, `allObjects` in `NSArray` and `NSSet`) before passing to a predicate. The method for `NSOrderedSet` did not, causing the predicate to receive an unexpected box instead of the contained value (crashing when the predicate attempted to force cast the box instead of the value).

This was caught while uncommenting unit tests for `NSPredicate`; there's now no longer a reason to keep them commented out.

### DIFF
--- a/Foundation/NSOrderedSet.swift
+++ b/Foundation/NSOrderedSet.swift
@@ -120,6 +120,16 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
             _insertObject(obj)
         }
     }
+    
+    internal var allObjects: [Any] {
+        if type(of: self) === NSOrderedSet.self || type(of: self) === NSMutableOrderedSet.self {
+            return _orderedStorage.map { _SwiftValue.fetch($0) }
+        } else {
+            return (0..<count).map { idx in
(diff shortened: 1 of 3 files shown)


itaiferber:
@phausler What do you think would be better here — leaving this implementation as-is (unboxing values in the `filter` method only), or instead adding such an auto-unboxing internal method (à la `NSArray`'s `internal var allObjects: [Any]`) for consistency with `NSSet` and `NSArray`?

itaiferber:
@swift-ci Please test

phausler:
I think it might be best for us to funnel all to the same type of method; so it would make sense to use (and implement) the SPI here for NSOrderedSet

itaiferber:
@swift-ci Please test

itaiferber:
Fails in `TestNSOperationQueue.test_OperationPriorities`, which appears to be unrelated/inconsistent anyway. The affected tests pass.

PR #595, opened 2016-08-25
Remove the dep on CUUID for Foundation

phausler:
Resolves https://bugs.swift.org/browse/SR-2470


phausler:
@swift-ci Please test

PR #596, opened 2016-08-25
Enable NSLocale keyed archiving unit tests

itaiferber:
`NSLocale`'s `encode(with:)` encodes the locale's identifier as a `CFString`, which causes a crash in `NSKeyedArchiver`'s `_haveVisited` ref map, since `CFString` cannot be forcibly cast to `AnyHashable`. Encoding the identifier as an `NSString` fixes the issue.

In decoding, a different `NSLocale` instance is returned. Since `NSLocale` no longer has value semantics (which were ported to `Locale`), it needs an `isEqual` implementation to compare two `NSLocale` instances as equal.

These changes combined allows us to uncomment the keyed archiving test and have it pass.

### DIFF
--- a/Foundation/NSLocale.swift
+++ b/Foundation/NSLocale.swift
@@ -61,7 +61,7 @@ open class NSLocale: NSObject, NSCopying, NSSecureCoding {
     
     open func encode(with aCoder: NSCoder) {
         if aCoder.allowsKeyedCoding {
-            let identifier = CFLocaleGetIdentifier(self._cfObject)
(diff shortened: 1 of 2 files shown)


itaiferber:
@swift-ci Please test

itaiferber:
@swift-ci Please test

itaiferber:
The build timed out because of `TestURLSession.test_downloadTaskWithURL`
Going to try re-running and hoping that it won't hang this time

itaiferber:
@swift-ci Please test

PR #597, opened 2016-08-25
ensure the concurrency gate is present for all variations of OperationQueue (irregardless of the queue width)

phausler:
Testing with malloc scribble, guard edges, and guard malloc revealed this failure on Darwin builds. Evidently it alters the timing enough to show the race on un-gated operations


phausler:
@swift-ci Please test

PR #600, opened 2016-08-26
Swift 3 API Parity: Use URLRequest rather than NSURLRequest in APIs

seabaylea:
A number of the URLSession APIs should now take `URLRequest` rather than `NSURLRequest` or `NSMutableURLRequest` as parameters. Examples of this include:
```
open func dataTask(with request: URLRequest) -> URLSessionDataTask
open func uploadTask(with request: URLRequest, from bodyData: Data) -> URLSessionUploadTask
open func downloadTask(with request: URLRequest) -> URLSessionDownloadTask
```
This PR switches over to using `URLRequest` rather than the existing use of `NSURLRequest`.

I've additionally added tests for `URLRequest` in addition to `NSURLRequest` (which still exists) by duplicating the existing `NSURLRequest` tests. 

Whilst doing this I hit a crash when trying to mutate a value in a copied `URLRequest`. This is caused because, when we call `_appyMutation` for a non-uniquely referenced value, we create a new `_MutableHandle`. The `_MutableHandle` constructor calls `copy()` on the referent (in our case a `NSMutableURLSession`). Calling `NSMutableURLSession.copy()` actually returns a `NSURLSession` (as it inherits copy() from NSURLSession), which fails because its not a `MutableType`. I've fixed this by overriding copy in `NSMutableURLSession` to return `mutableCopy()`.


parkera:
@swift-ci please test linux

phausler:
we should consider adding these structural tests into the overlay unit tests as well

PR #601, opened 2016-08-26
[Docs] Mention installing libcurl on Ubuntu 14.04

modocache:
Installing libcurl is necessary as of c26f465e9a.


parkera:
Isn't this covered by the dependencies listed in the general documentation?

phausler:
that documentation change for swift was merged in https://github.com/apple/swift/pull/4407
However it is a notable enough change that perhaps it isn't such a bad idea to have it in multiple places

modocache:
Oh! Sorry, I hadn't noticed. I'm fine with either location, and in any case, this pull request should probably match the one in the README, so I'll close this.

PR #602, opened 2016-08-28
Swift 3 API parity and new implementations for NSTimeZone

xwu:
Done:

* Renamed members for `NSTimeZone` to align with Darwin Foundation API.
* Implemented `init(forSecondsFromGMT:)`, `localizedName(_:locale:)`.
* Implemented properties `abbreviationDictionary` (getter only), `secondsFromGMT`, `isDaylightSavingTime`, `daylightSavingTimeOffset`, `nextDaylightSavingTimeTransition`.
* Removed `fatalError()` from helper function for `TimeZone.current` and plumbed through to `NSTimeZone.system`.
* Aligned `TimeZone` API to Swift 3, restoring/adding `TimeZone.autoupdatingCurrent` (now plumbed through to `NSTimeZone.local`, which is unimplemented) and updating conformance to `Equatable`, `CustomStringConvertible`, etc., by copying over newer code from apple/swift.
 
Not done:

* `NSTimeZone.local` should return an auto-updating member; this one is non-obvious to implement and its lack of support here is in line with other auto-updating facilities exclusive to Darwin (e.g., Locale, Calendar).

* The `NSTimeZone.abbreviationDictionary` setter remains unimplemented because the underlying CF function hits an assertion failure when trying to release non-CF objects while clearing an internal cache; it is a problem independent of the overlying Foundation methods, since the same failure is triggered by making a mutable copy of the existing abbreviation dictionary, then setting the abbreviation dictionary to that mutable copy using only CF functions.

* No attempt has been made to align `NSTimeZone.description` with Darwin, which appears no longer to use the underlying CF description as corelibs-foundation continues to do.

* Darwin Foundation initializes time zones with fixed GMT offsets between -1800 and 1800, but the open-sourced CF functions limit the range to between -1400 and 1400; I assume this is a case where the open-sourced CF could simply use an update.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -287,6 +287,8 @@ typedef pthread_t _CFThreadRef;
 
 CF_EXPORT _CFThreadRef _CFThreadCreate(const _CFThreadAttributes attrs, void *_Nullable (* _Nonnull startfn)(void *_Nullable), void *restrict _Nullable context);
 
+CF_EXPORT Boolean _CFTimeZoneInitWithTimeIntervalFromGMT(CFTimeZoneRef result, CFTimeInterval ti);
+
 CF_EXPORT Boolean _CFCharacterSetIsLongCharacterMember(CFCharacterSetRef theSet, UTF32Char theChar);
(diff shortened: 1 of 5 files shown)


itaiferber:
LGTM

itaiferber:
@swift-ci Please test

parkera:
Let's get this PR up for the swift-3.0-branch as well.

itaiferber:
@parkera Yep, working on it

itaiferber:
Testing in #609

PR #606, opened 2016-08-29
Add Dispatch to the library list and library search path for linking

vivkong:
While building on s390x, the linker cannot find the symbols from libdispatch while building TestFoundation.  I'd like to add libdispatch to the library list and search path for linking.


vivkong:
The linker error was actually caused by not using the gold linker on s390x. I have a fix and I'll create a pull request for Swift.

PR #605, opened 2016-08-29
Fixing a typo in internal bridging protocol names (NFC)

xwu:
"Bridgeable" not "bridgable".

PR #588, opened 2016-08-24
[SR-1608] NSStream unimplemented features PT1

DaveLiu888:
Old pull: https://github.com/apple/swift-corelibs-foundation/pull/489
Old Pull request was closed due to churn about experimental un official API particularly 

```
class func getStreamsToHost(withName hostname: String, port: Int, inputStream: AutoreleasingUnsafeMutablePointer<InputStream?>?, outputStream: AutoreleasingUnsafeMutablePointer<NSOutputStream?>?) 
```

I have decoupled these 2 issues and created a separate Jira ticket for it here and will be filing a pr for soon https://bugs.swift.org/browse/SR-2431

**What's in this pull request?**
1st part of Resolved bug number: 1608 (https://bugs.swift.org/browse/SR-1608)
NSStream implementation of unimplemented features with tests

SR-1608 has been split in 2 pulls in effort to keep pulls small
- [ ] @phausler 
- [ ] @parkera

PR #607, opened 2016-08-29
Bridging to CoreFoundation types for InputStream/OutputStream

apocolipse:
Relies on #605 for fixes to spelling.

Adds proper bridging to `InputStream` and `OutputStream` to their CF counterparts.

### DIFF
--- a/Foundation/NSStream.swift
+++ b/Foundation/NSStream.swift
@@ -119,7 +119,7 @@ open class Stream: NSObject {
 // Subclassers are required to implement these methods.
 open class InputStream: Stream {
 
-    private var _stream: CFReadStream!
+    fileprivate var _stream: CFReadStream!
 
     // reads up to length bytes into the supplied buffer, which must be at least of size len. Returns the actual number of bytes read.
     open func read(_ buffer: UnsafeMutablePointer<UInt8>, maxLength len: Int) -> Int {
@@ -147,6 +147,10 @@ open class InputStream: Stream {
     public convenience init?(fileAtPath path: String) {
         self.init(url: URL(fileURLWithPath: path))
     }
+    
+    internal init(reference: CFReadStream) {
+        _stream = reference
+    }
 
     open override func open() {
         CFReadStreamOpen(_stream)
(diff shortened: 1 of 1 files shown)


phausler:
This seems ok, but what is the exact use case here? the CF bridge protocols are internal only. Is this somehow intending to allow Foundation's OutputStream to be passed/fetched directly to/from CF as a CFWriteStream?

apocolipse:
@phausler Yes, intending to allow direct access to CFWriteStream (and read), Currently there's a discrepancy between CFStreamPropertyKeys and Stream.PropertyKeys such that if you want to set specific keys it must be done on a CF stream, and vice versa.  i.e. Stream.PropertyKey.socketSecuritylevelKey is implemented but not for CFStreams.  As well as things like CFStreamCreatePairWithSocketToHost  are not implemented for Stream (yet).

Ideally, all of the Stream.PropertyKeys should eventually align with all kCFStreamPropertyKeys, and utilities like getStreamsToHost in Stream will be fully implemented (#588 #589) so that lower level CFStream , but for now bridging helps get around some of the current inconsistencies.

apocolipse:
@phausler are you indicating this won't allow for external bridging?  If not, is there any way that will?  Otherwise I'd opt to just expose the internal `_cfObject` and `init(reference:)`

phausler:
CF is an internal implementation detail of swift-corelibs-foundation. I often attempt to dissuade Swift developers from using CF - because it is quite painful to deal with typing unsafe so many times and often the Foundation counterparts are faster anyhow. 

In this particular case I was curious if the intent is to directly allow APIs from CF to pass transparently across to Foundation - if that is the case then we need to make certain we have a pattern more similar to NSCFString and such to do toll free bridging. What you have here is a "has-a" pattern bridge, I was wanting to make certain you were not intending for it to be a "is-a" pattern bridge.

tl;dr To hand out a CFWriteStream and allow use as a NSStream it needs to be the "is-a" pattern.

parkera:
I'm going to go ahead and close this for now. Please re-open when changes requested are made. Thanks.

PR #608, opened 2016-08-29
Foundation API Conformance Update: Calendar

itaiferber:
To go along with #602

PR #609, opened 2016-08-29
Swift 3 API parity for [NS]TimeZone & [NS]Calendar

itaiferber:
Pulls #602 and #608 into swift-3.0-branch

### DIFF
--- a/Foundation/NSTimeZone.swift
+++ b/Foundation/NSTimeZone.swift
@@ -294,5 +294,5 @@ extension NSTimeZone {
 }
 
 extension NSNotification.Name {
-    public static let NSSystemTimeZoneDidChange = NSNotification.Name(rawValue: "NSSystemTimeZoneDidChangeNotification") // NSUnimplemented
+    public static let NSSystemTimeZoneDidChange = NSNotification.Name(rawValue: "") // NSUnimplemented
 }
--- a/Foundation/NSTimeZone.swift
+++ b/Foundation/NSTimeZone.swift
@@ -174,7 +174,7 @@ extension NSTimeZone {
 
     open class func resetSystemTimeZone() {
         CFTimeZoneResetSystem()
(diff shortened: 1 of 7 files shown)


itaiferber:
@swift-ci Please test

itaiferber:
This failure is unrelated to these changes. I think something is going on with CI ATM — we had a separate, spurious build failure in CI a few minutes ago that seemed related to a configuration issue

itaiferber:
Going to try again soon

Foundation/NSTimeZone.swift
```diff
@@ -249,3 +292,7 @@ extension NSTimeZone {
     }
 
 }
+
+extension NSNotification.Name {
+    public static let NSSystemTimeZoneDidChange = NSNotification.Name(rawValue: "NSSystemTimeZoneDidChangeNotification") // NSUnimplemented
```
xwu: @itaiferber, could you do me a favor and remove the raw value from this line? I think it's not supposed to be here (my mistake!). I'll create a PR to have it removed from master as well.

itaiferber: @xwu As in, the string value itself, or `rawValue:`?

itaiferber:
@swift-ci Please test

Foundation/NSTimeZone.swift
xwu: As in make it read `public static let NSSystemTimeZoneDidChange = NSNotification.Name(rawValue: "")`, since that seems to be what the project does.

itaiferber: @xwu No problem. I'm just going to let this CI pass run first before pushing the changes.

itaiferber:
CI fails when trying to `update-checkouts` from Swift. This might have to do with the fact that this is on the swift-3.0-branch
Investigating

itaiferber:
CI job fixed and passed [here](https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/286/console)
Merging

PR #613, opened 2016-08-31
Loopback tests for URLSession

pushkarnk:
This is a very basic HTTP server for loopback tests for URLSession. I have borrowed some implementation aspects from the HTTPServer.swift in PR 299. The server currently, listens, reads and writes only once. It is possible to test data tasks and download tasks. Opening this PR for review and comments. 

I also had to make a small change to build.py to get TestFoundation built with Dispatch.

### DIFF
--- a/TestFoundation/HTTPServer.swift
+++ b/TestFoundation/HTTPServer.swift
@@ -1,0 +1,245 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+
+//This is a very rudimentary HTTP server written plainly for testing URLSession. 
+//It is not concurrent. It listens on a port, reads once and writes back only once.
+//We can make it better everytime we need more functionality to test different aspects of URLSession.
+
+import Dispatch
+
+#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
+    import Foundation
+    import Glibc
+#else
+    import CoreFoundation
+    import SwiftFoundation
+    import Darwin
+#endif
+
+public let globalDispatchQueue = DispatchQueue.global()
+
+struct _HTTPUtils {
+    static let CRLF = "\r\n"
+    static let VERSION = "HTTP/1.1"
+    static let SPACE = " "
+    static let CRLF2 = CRLF + CRLF
+    static let EMPTY = ""
+}
+
+class _TCPSocket {
+  
+    private var listenSocket: Int32!
+    private var socketAddress = UnsafeMutablePointer<sockaddr_in>.allocate(capacity: 1) 
+    private var connectionSocket: Int32!
+    
+    private func isNotNegative(r: CInt) -> Bool {
+        return r != -1
+    }
+
+    private func isZero(r: CInt) -> Bool {
+        return r == 0
+    }
+
+    private func attempt(_ name: String, file: String = #file, line: UInt = #line, valid: (CInt) -> Bool,  _ b: @autoclosure () -> CInt) throws -> CInt {
+        let r = b()
+        guard valid(r) else { throw ServerError(operation: name, errno: r, file: file, line: line) }
+        return r
+    }
+
+    init(port: UInt16) throws {
+        #if os(Linux)
+            let SOCKSTREAM = Int32(SOCK_STREAM.rawValue)
+        #else
+            let SOCKSTREAM = SOCK_STREAM
+        #endif
+        listenSocket = try attempt("socket", valid: isNotNegative, socket(AF_INET, SOCKSTREAM, Int32(IPPROTO_TCP)))
+        var on: Int = 1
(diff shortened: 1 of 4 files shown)


build.py
```diff
@@ -458,8 +458,11 @@
 # TODO: Probably this should be another 'product', but for now it's simply a phase
 foundation_tests = SwiftExecutable('TestFoundation', [
 	'TestFoundation/main.swift',
+        'TestFoundation/HTTPServer.swift',
 ] + glob.glob('./TestFoundation/Test*.swift')) # all TestSomething.swift are considered sources to the test project in the TestFoundation directory
 
+Configuration.current.extra_ld_flags = '-L'+Configuration.current.variables["LIBDISPATCH_BUILD_DIR"]+'/src/.libs'
```
pushkarnk: @seabaylea :  In build.py, TestFoundation is built as a SwiftExecutable without the linker flags, this results in ld.gold failing to locate libdispatch. I can see that a the rule for SwiftExecutable used EXTRA_LD_FLAGS. Hence added this line. Can you please comment on this?

seabaylea: That makes sense - its what we have to do when building Foundation as Dispatch isn't yet in the right place (as we're still building the toolchain)

TestFoundation/HTTPServer.swift
```diff
@@ -0,0 +1,203 @@
...
+//
+
+
+//This is a very rudimentary HTTP server written plainly for testing URLSession. 
+//It is not concurrent. It listens on a port, reads once and writes back only once.
+//We can make it better everytime we need more functionality to test different the aspects of URLSession.
+
+import Foundation
+import Dispatch
+import Glibc
```
pushkarnk: Still working on Darwin

TestFoundation/TestNSURLSession.swift
```diff
@@ -101,13 +158,7 @@ class TestURLSession : XCTestCase {
...
             let httpResponse = response as! HTTPURLResponse?
             XCTAssertEqual(200, httpResponse!.statusCode, "HTTP response code is not 200")
-            do {
-                let json = try JSONSerialization.jsonObject(with: data!, options: [])
-                let arr = json as? Array<Any>
-                let first = arr![0]
-                let result = first as? [String : Any]
-                expectedResult = result!["capital"] as! String
-            } catch { }
+            expectedResult = NSString(data: data!, encoding: String.Encoding.utf8.rawValue)!._bridgeToSwift()
```
parkera: I believe we shouldn't be using the `_bridgeToSwift` function outside of Foundation's implementation itself.

pushkarnk: Thanks. Corrected.

TestFoundation/HTTPServer.swift
```diff
@@ -0,0 +1,203 @@
...
+
+
+//This is a very rudimentary HTTP server written plainly for testing URLSession. 
+//It is not concurrent. It listens on a port, reads once and writes back only once.
+//We can make it better everytime we need more functionality to test different aspects of URLSession.
+
+import Foundation
+import Dispatch
+import Glibc
+
```
pushkarnk: wip on Darwin

pushkarnk:
Added support to test download tasks. Working on making this work on Darwin now.

parkera:
How's this going?

pushkarnk:
I'm need to make this build and run on Darwin.

pushkarnk:
@parkera : I've updated the Xcode project. Can you please run a test?

TestFoundation/TestNSURLSession.swift
```diff
@@ -164,7 +259,13 @@ class TestURLSession : XCTestCase {
     }
 }
 
-class DataTask: NSObject {
+#if os(Linux)
+typealias FoundationObject = NSObject
+#else
+typealias FoundationObject = SwiftFoundation.NSObject
+#endif
```
pushkarnk: This looks awkward. But I found the use of NSObject causing a "NSObject ambiguous for type lookup in this context" compiler error with Xcode.

parkera: I'd like to understand more about the root cause of this error. Clients of Foundation should absolutely be able to subclass NSObject without this kind of hackery.

pushkarnk:
The ambiguity error comes from the ```import Dispatch``` statement in TestFoundation/TestNSURLSession. I have created https://bugs.swift.org/browse/SR-2752 to investigate this. Looks like Dispatch also exposes NSObject. I removed the dirty hack and did some refactoring to work around the reported problem.

TestFoundation/TestNSURLSession.swift
pushkarnk: Thanks. Removed this.

build.py
```diff
@@ -467,8 +467,11 @@
 # TODO: Probably this should be another 'product', but for now it's simply a phase
 foundation_tests = SwiftExecutable('TestFoundation', [
 	'TestFoundation/main.swift',
+        'TestFoundation/HTTPServer.swift',
```
pushkarnk: I am not sure. But ```swift_sources```  seems to be holding all the swift source files that are compiled into Foundation. ```HTTPServer.swift``` needs to be compiled into TestFoundation.

parkera: You're right, the other test files are found via the glob. I guess this is ok. I'm going to kick off at est.

parkera:
@swift-ci please test

PR #614, opened 2016-08-31
Partial implementation of NSSearchPathForDirectoriesInDomains

amraboelela:


Foundation/NSPathUtilities.swift
```diff
@@ -574,6 +574,18 @@ extension FileManager {
...
+    if domainMask == .userDomainMask {
+        switch directory {
+        case .libraryDirectory:
+            return [bundlePath + "/Library"]
+        case .documentDirectory:
+            return [bundlePath + "/Documents"]
+        default:
+            break
+        }
+    }
```
parkera: Shouldn't the user domain be in the home directory (`~`)? The search functionality on `Bundle` itself is used for finding resources inside the main bundle; and bundles typically do not have a Library and Documents directory inside them.

amraboelela: In iOS this function returns Library or Documents directory inside the main bundle

Foundation/NSPathUtilities.swift
```diff
@@ -574,6 +574,7 @@ extension FileManager {
 }
 
 public func NSSearchPathForDirectoriesInDomains(_ directory: FileManager.SearchPathDirectory, _ domainMask: FileManager.SearchPathDomainMask, _ expandTilde: Bool) -> [String] {
+    #if os(iOS)
```
parkera: This doesn't quite make sense either - this code will only run on Linux.

amraboelela: Then why do we have this everywhere?

#if os(OSX) || os(iOS)
    import Darwin
#elseif os(Linux)
    import Glibc
#endif

parkera: We run on Mac for testing purposes but never on iOS.

Anyway, I'm pretty sure that this does not return a location inside the main bundle even on iOS. It returns something inside the container, which is not the same as the main bundle.

parkera:
I'm going to ahead and close this for now; please re-open when changes requested are addressed.

PR #615, opened 2016-08-31
Swift 3 API parity: Change type of HTTPURLResponse.allHeaderFields

pushkarnk:
Change type of HTTPURLResponse.allHeaderFields to [AnyHashable : Any] to match Darwin.

### DIFF
--- a/Foundation/NSURLResponse.swift
+++ b/Foundation/NSURLResponse.swift
@@ -154,7 +154,7 @@ open class HTTPURLResponse : URLResponse {
     ///
     /// - Important: This is an *experimental* change from the
     /// `[NSObject: AnyObject]` type that Darwin Foundation uses.
-    public let allHeaderFields: [String: String]
+    public let allHeaderFields: [AnyHashable : Any]
     
     /// Convenience method which returns a localized string
     /// corresponding to the status code for this response.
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci please test linux

Foundation/NSURLResponse.swift
```diff
@@ -233,7 +233,8 @@ open class HTTPURLResponse : URLResponse {
     /// This property is intended to produce readable output.
     override open var description: String {
         var result = "<\(type(of: self)) \(Unmanaged.passUnretained(self).toOpaque())> { URL: \(url!.absoluteString) }{ status: \(statusCode), headers {\n"
-        for(key, value) in allHeaderFields {
+        for(aKey, aValue) in allHeaderFields {
+            let (key,value) = (aKey as! String, aValue as! String) //allHeaderFields is either assigned a [String:String] or [:]
```
parkera: If we allow the header fields to be anything besides a string as part of the API, then shouldn't we use `as?` here? Otherwise we'll just assert in some cases.

pushkarnk: Yes, though allHeaderFields is always derived from a ```[String:String]``` we can have subclasses storing other types in it. I've updated the PR. Thanks.

parkera:
@swift-ci please test linux

parkera:

PR #617, opened 2016-09-02
Contributing test-cases for NSOperation/NSOperationQueue.

mamabusi:


TestFoundation/TestNSOperationQueue.swift
```diff
@@ -36,17 +41,21 @@ class TestNSOperationQueue : XCTestCase {
 
     func test_OperationPriorities() {
         var msgOperations = [String]()
-        let operation1 : BlockOperation = BlockOperation(block: {
+        let operation1 = BlockOperation(block: {
             msgOperations.append("Operation1 executed")
+            sleep(1)
```
mamabusi: Added time-lag to resolve the test-case failures seen sometimes on TestFoundation.

parkera: Sorry - sleep cannot be a fix for issues like this. Instead we should use a semaphore to signal completion of whatever work we are waiting on.

mamabusi: Ok. Will modify to use semaphores. Thank you.

erg:
The test is correct and exposed a real bug in the underlying concurrency code that is fixed by #621.

This PR is not needed now and can be closed.

PR #618, opened 2016-09-02
Changed coefficients of UnitAngle to more exact values

Qbyte248:
The provided values were not as exact as they could be. In fact the value for "revolutions" was completely wrong (2π instead of 360°).

### DIFF
--- a/Foundation/Unit.swift
+++ b/Foundation/Unit.swift
@@ -277,11 +277,11 @@ open class UnitAngle : Dimension {
     
     private struct Coefficient {
         static let degrees      = 1.0
-        static let arcMinutes   = 0.016667
-        static let arcSeconds   = 0.00027778
-        static let radians      = 57.2958
+        static let arcMinutes   = 1.0 / 60.0
+        static let arcSeconds   = 1.0 / 3600.0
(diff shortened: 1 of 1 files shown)


Foundation/Unit.swift
```diff
@@ -277,11 +277,11 @@ open class UnitAngle : Dimension {
...
         static let degrees      = 1.0
-        static let arcMinutes   = 0.016667
-        static let arcSeconds   = 0.00027778
-        static let radians      = 57.2958
+        static let arcMinutes   = 1.0 / 60.0
+        static let arcSeconds   = 1.0 / 3600.0
+        static let radians      = 180.0 / .pi
         static let gradians     = 0.9
-        static let revolutions  = 6.28319
+        static let revolutions  = 360.0
```
3a4oT: 2π rad = 360° so why do you think it was wrong?

Qbyte248: That's true but the base unit is degrees so the conversion factor has to be 360.

parkera:
@swift-ci please test linux

lifeissweetgood:
Good catch, thanks!

PR #619, opened 2016-09-04
Fix crash in NSValue.isEqual() when it is passed an NSNumber

gribozavr:
NSNumbers are not added to the side table, so we can't find them.

I don't think that `NSValue.isEqual()` is correct in general, but this gets us closer to the correct implementation and unblocks https://github.com/apple/swift/pull/4621 . For example, subclasses that have custom storage -- like NSNumber -- should compare equal to an NSValue as long as they have the same bytes and the same type, but the current implementation of `NSValue.isEqual()` can only return true if two references are equal.

Note: this issue blocks https://github.com/apple/swift/pull/4621 (after we change the hashing algorithm, the test TestNSKeyedArchiver.test_archive_set starts to compare NSValue to NSNumber where it previously wasn't doing that just because we got lucky in the hash table order).

### DIFF
--- a/Foundation/NSValue.swift
+++ b/Foundation/NSValue.swift
@@ -58,10 +58,12 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
             } else {
                 // bypass _concreteValue accessor in order to avoid acquiring lock twice
                 let (lhs, rhs) = NSValue.SideTableLock.synchronized {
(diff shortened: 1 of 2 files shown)


gribozavr:
@phausler @itaiferber Could you review, please?

gribozavr:
@swift-ci Please test

PR #620, opened 2016-09-06
Implement missing functions in Date

alblue:
The two NSUnimplemented functions in Date relate to a class property
`timeIntervalSinceReferenceDate` which matches an instance property
of the same name, and methods relating to the unkeyed coder.

Implement the `timeIntervalSinceReferenceDate` by instantiating a
new instance and delegating to the instance property, and change
the unkeyed coder implementations to a fatalError indicating that
they are unsupported.


parkera:
@swift-ci please test

PR #621, opened 2016-09-06
resolve inserting operations as a batch race to the execution of operations

phausler:


phausler:
@swift-ci Please test

PR #622, opened 2016-09-07
Port of Foundation to Android

johnno1962:
This PR contains the relatively minor changes to have Foundation _compile_ with the Android NDK. There is an associated pull request with the small change https://github.com/apple/swift/pull/4662 to build-script-impl in the swift project to pass through the required arguments to the ninja scritpt generator. The focus has been on getting the code to  compile but preliminary testing of Regular expressions, NSHost, fetching Strings from URLs has been successful so far. There is a known issue with needing to hook thread exit to call JNI method DetachCurrentThread which I’ll have to look at before making a PR on swift-corelibs-libdispatch (resolved, see below).

Issues found and not resolved:
Optimising the creation of a string from zero length data - optimisation patched out for now
Fetching the current Locale language for web fetch request headers - also patched out

Changes tested on Linx and OSX build. Included is a minor change fixing a missing -lcurl when building "util/build-script —foundation" on Linux.

As foundation has binary dependencies on libxml2 and libcurl and it's not been able to automate the libdispatch crosscompile the build is dependent on a pre-built binary & header package downloaded by a series of scripts in the swift-corelibs-foundation/android directory. These binaries were built from the public repos pretty much as-is after some Makefilery. Details are in the README.md in this directory.


CoreFoundation/String.subproj/CFString.c
```diff
@@ -1264,12 +1264,15 @@ CF_PRIVATE CFStringRef __CFStringCreateImmutableFunnel3(
 	contentsDeallocator = __CFGetDefaultAllocator();
     }
 
+    #ifndef DEPLOYMENT_TARGET_ANDROID
+    // crashes String.swift, line 51 when creating String from zero length Data
```
modocache: It'd be great to file a https://bugs.swift.org report to track this crash, and include the report number here. That way, we'll know to remove this `#ifdef` once the report is fixed.

Foundation/CGFloat.swift
```diff
@@ -611,6 +611,7 @@ public func %=(lhs: inout CGFloat, rhs: CGFloat) {
     fatalError("%= is not available.")
 }
 
+#if !os(Android)
```
modocache: IIRC tgmath is available on Android, and I think the apple/swift tests for tgmath pass when run on Android as well. I must be missing something, but why the `#if !os(Android)` here? And do the Foundation tests pass this way?

Foundation/NSData.swift
```diff
@@ -930,7 +930,11 @@ open class NSMutableData : NSData {
     }
     
     open func resetBytes(in range: NSRange) {
+        #if os(Android)
+        memset(mutableBytes.advanced(by: range.location), 0, range.length)
+        #else
         bzero(mutableBytes.advanced(by: range.location), range.length)
+        #endif
```
modocache: nit-pick: Maybe indent the `memset(...)` lines here?

johnno1962:
I’ll have another look at it. There are some tricky type conversions to get right.

Foundation/NSData.swift
johnno1962: I’ll have another look at it. There are some tricky type conversions to get right.

Foundation/CGFloat.swift
johnno1962: I’ll have another look at it. There are some tricky type conversions to get right.
 Add a line note

Foundation/NSGeometry.swift
```diff
@@ -460,7 +475,7 @@ public func NSIntegralRectWithOptions(_ aRect: NSRect, _ opts: NSAlignmentOption
         width = 0
     }
     
-
+#if !os(Android)
```
modocache: nit-pick: Looks like some trailing whitespace here.

Foundation/NSData.swift
johnno1962: hmm. not sure about that!

modocache:
> Issues found and not resolved:
> Fetching the current Locale language for web fetch request headers - also patched out

[I think this might be related?](https://github.com/apple/swift/blob/016bc249981088194d6216aa92a8b616825f1570/stdlib/public/stubs/Stubs.cpp#L43-L63)

Foundation/NSGeometry.swift
johnno1962: zapped

Foundation/NSGeometry.swift
```diff
@@ -9,10 +9,25 @@
...
+#elseif os(Linux) || os(Android)
     import Glibc
 #endif
 
+#if os(Android)
+private func floor( value: CGFloat ) -> Double {
+  return floor( Double(value ) )
+}
+private func ceil( value: CGFloat ) -> Double {
+  return floor( Double(value ) )
```
mekjaer: ceil here?

CoreFoundation/Base.subproj/CoreFoundation.h
```diff
@@ -33,6 +35,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <netdb.h>
```
phausler: These are a global change to both the swift-corelibs, CFLite and Darwin versions, changes to these types of headers come at definite risk. Can this be done without these modifications in CoreFoundation.h? (and perhaps somewhere else instead?)

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -157,7 +157,11 @@ const char *_CFProcessPath(void) {
 
 #if DEPLOYMENT_TARGET_LINUX
 #include <unistd.h>
+#if __has_include(<syscall.h>)
 #include <syscall.h>
+#else
```
phausler: If the previous header does not exist what is to say the sys/syscall version exists? Perhaps that should be gated as well and #error'd on the else claiming that syscall.h is required

CoreFoundation/String.subproj/CFString.c
phausler: IIRC android's memcpy and memmove are not the same as Darwin. That may be the culprit of this issue.

Foundation/NSAffineTransform.swift
```diff
@@ -9,7 +9,7 @@
 
 #if os(OSX) || os(iOS)
     import Darwin
-#elseif os(Linux)
+#elseif os(Linux) || os(Android)
```
phausler: why is Android named Glibc? shouldn't it be `import Bionic`?

Foundation/NSData.swift
phausler: might actually be reasonable to just implement a CF inline shim for bzero; it is used a fair amount.

Foundation/NSLog.swift
```diff
@@ -41,3 +41,11 @@ public func NSLog(_ format: String, _ args: CVarArg...) {
         NSLogv(format, $0) 
     }
 }
+
+// hook for System.out.println()
+#if os(Android)
+public var debug = {
```
phausler: can we name this something else? debug is potentially a very generically named symbol it could be used for storing strings, bools etc that might be something completely different and stomping upon that name might be a bit confusing to developers.

I would expect that NSLog actually hooks to android_log_print so that NSLog's show up in adb

Foundation/NSURLSession/NSURLSessionTask.swift
```diff
@@ -624,9 +624,12 @@ fileprivate extension URLSessionTask {
         var result = [("Connection", "keep-alive"),
                       ("User-Agent", userAgentString),
                       ]
+        #if !os(Android)
+        // Crashes on Android
```
phausler: please file a ticket on this. and perhaps a stack trace.

Foundation/NSData.swift
compnerd: Why not change all uses of `bzero` to `memset`?  `bzero` is considered deprecated, and POSIX1-2008 removes the specification of `bzero` entirely.

android/install.sh
```diff
@@ -0,0 +1,33 @@
...
+
+SWIFT_ROOT="$(dirname $(dirname $(which swiftc)))"
+BUILD_DIR="$(dirname $SWIFT_ROOT)"
+
+if [[ "$1" == "-u" ]]; then
+    rm -rf "${SWIFT_ROOT}/lib/swift/CoreFoundation"
+    exit
+fi
+
+\cp -v "${BUILD_DIR}/foundation-linux-x86_64/Foundation/libFoundation.so" "${SWIFT_ROOT}/lib/swift/android" &&
```
phausler: why is the configuration being passed as a x86_64 triple? shouldn't it be armv7a-linux-androideabi? which would emit the build path as foundation-android-armv7

android/install.sh
```diff
@@ -0,0 +1,33 @@
...
+
+\cp -v "${BUILD_DIR}/foundation-linux-x86_64/Foundation/Foundation.swift"* "${SWIFT_ROOT}/lib/swift/android/armv7" &&
+
+rsync -arv "${BUILD_DIR}/foundation-linux-x86_64/Foundation/usr/lib/swift/CoreFoundation" "${SWIFT_ROOT}/lib/swift/"
```
phausler: how are these packagable with JNI apps?
can they be just dropped in as a pre-built?

phausler:
For the Android port does this need to have a JNI_OnLoad or is that entry point higher?

CoreFoundation/Base.subproj/CoreFoundation.h
johnno1962: Which head is which of the two? I can remove it from the Darwin version and add #if __has_include(). The symbols are used in NSHost.swift.

CoreFoundation/Base.subproj/CFPlatform.c
johnno1962: won’t it error anyway? It doesn’t fail silently or anythhing anti-social. This just improves the odds. I can gate it if you want.

Foundation/NSLog.swift
johnno1962: This is a temporary convenience. I can rename it or put it in the NSLog class if you like.

Foundation/NSURLSession/NSURLSessionTask.swift
johnno1962: ok.

CoreFoundation/Base.subproj/CoreFoundation.h
phausler: So there are two CoreFoundation.h's one for swift and one for Darwin. this one is the Darwin version.

Having alterations to this header could prove to be a non starter because it would prevent us from merging changes back from CoreFoundation. That being said if the reason is compelling enough for us to change the header for ALL platforms then perhaps it can be done, but I think it would be best to tackle it separately if possible.

android/install.sh
johnno1962: I’ve no idea how build-script-impl works in detail TBH. I may need some help here.

android/install.sh
johnno1962: yes, I have a modified gradle plugin which copies them out of buidl/…/lib/swift/android along with the swift shared lbs.

johnno1962:
The JNI_OnLoad is in https://github.com/SwiftJava/java_swift/blob/master/Sources/JavaJNI.swift#L23

phausler:
Will that mean we will need a hook in Thread.swift for VM context destruction?

johnno1962:
Absolutely. Android requires you to detach the thread and won’t allow you to do it pre-emptively claiming you are stilll processing. Some help required here also. I haven’t made much headway understanding libdispatch, kqueues, etc

Foundation/NSData.swift
johnno1962: I’ll remove bzero altogether.

Foundation/NSAffineTransform.swift
johnno1962: That would make for a lot of #if’s all over peoples codebase!

CoreFoundation/Base.subproj/CoreFoundation.h
johnno1962: I’ve added netdb.h in the copy of this header in the SwiftRuntime directory with a #if __has_include(..)

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -203,6 +206,7 @@ strlcat(char * dst, const char * src, size_t maxlen) {
     }
     return dstlen + srclen;
 }
+#endif
```
parkera: Why is fd_mask defined in an else clause with definitions of strlcpy/strlcat? Are the two related in some way that I am missing?

CoreFoundation/Base.subproj/SwiftRuntime/CoreFoundation.h
```diff
@@ -14,6 +14,8 @@
  
 */
 
+#include <stdarg.h>
```
parkera: Nothing should be above the defined macro.

Foundation/NSAffineTransform.swift
parkera: Probably true, but what does that accomplish for us besides making us put another import in our files?

Foundation/NSData.swift
```diff
@@ -930,7 +930,7 @@ open class NSMutableData : NSData {
     }
     
     open func resetBytes(in range: NSRange) {
-        bzero(mutableBytes.advanced(by: range.location), range.length)
+        memset(mutableBytes.advanced(by: range.location), 0, range.length)
```
parkera: What was wrong with `bzero` here? I think I would rather have a compatibility shim for Android than excise it from our code, if it's missing.

Foundation/NSFileManager.swift
```diff
@@ -9,10 +9,16 @@
 
 #if os(OSX) || os(iOS)
     import Darwin
-#elseif os(Linux)
+#elseif os(Linux) || os(Android)
     import Glibc
 #endif
 
+#if os(Android)
```
parkera: I'd really like any special case for other platforms to at least come with a comment explaining why it's needed and what could be done to elide it in the future. Even better would be to avoid as much platform-specific code as possible, or put it into a central location so we can count on its presence everywhere in the project.

Foundation/NSFileManager.swift
```diff
@@ -279,8 +285,10 @@ open class FileManager : NSObject {
 
 #if os(OSX) || os(iOS)
         let ti = (TimeInterval(s.st_mtimespec.tv_sec) - kCFAbsoluteTimeIntervalSince1970) + (1.0e-9 * TimeInterval(s.st_mtimespec.tv_nsec))
-#else
+#elseif !os(Android)
```
parkera: I would prefer `#elseif os(Android)` -- that is, keep the #if in the positive sense.

Foundation/NSGeometry.swift
```diff
@@ -9,10 +9,27 @@
...
+    return ceil( Double(value) )
+}
+private func round( value: CGFloat.NativeType ) -> Double {
+    return round( Double(value) )
+}
+#if false
+private func +( left: CGFloat.NativeType, right: CGFloat.NativeType ) -> Double {
+    return Double(left) + Double(right)
+}
+#endif
```
parkera: No dead code please.

Foundation/NSGeometry.swift
```diff
@@ -9,10 +9,27 @@
...
+private func floor( value: CGFloat.NativeType ) -> Double {
+    return floor( Double(value) )
+}
+private func ceil( value: CGFloat.NativeType ) -> Double {
+    return ceil( Double(value) )
+}
+private func round( value: CGFloat.NativeType ) -> Double {
+    return round( Double(value) )
+}
+#if false
```
parkera: This and the other changes to CGFloat implementations aren't very clear to me. CGFloat should actually be Float on 32-bit and Double on 64-bit.

Foundation/NSGeometry.swift
```diff
@@ -459,8 +476,7 @@ public func NSIntegralRectWithOptions(_ aRect: NSRect, _ opts: NSAlignmentOption
     if aRect.size.width.native < 0 {
         width = 0
     }
-    
-
+#if !os(Android) // FIXME!!!
```
parkera: I think we need to understand what's going on here; I don't really want to see a // FIXME in code. JIRA number, or a fatalError of some kind (like our NSUnimplemented).

Foundation/NSLog.swift
parkera: I agree with @phausler - please no additional public symbols, and especially not one named `debug`.

Foundation/NSTask.swift
```diff
@@ -7,6 +7,7 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+#if !os(Android)
```
parkera: I don't want to comment out whole files for specific platforms like this. Is it impossible to support? If so we should put a fatalError in a useful place (like init).

Either that or we need a better availability strategy for the API (e.g., marking the class as unavailable on android only).

Foundation/NSURLSession/NSURLSessionTask.swift
```diff
@@ -838,9 +841,14 @@ extension URLSessionTask {
...
             data.append(Data(bytes: bodyData!.bytes, count: bodyData!.length)) 
 
+            #if !os(Android)
             s.delegateQueue.addOperation {
                 completion(data, response, nil)
             }
+            #else
+            // HACK: Operation Queues not working on Android
+            completion(data, response, nil)
+            #endif
```
parkera: What's the plan to make sure dispatch and operation queues are actually working?

Foundation/NSXMLNode.swift
```diff
@@ -83,7 +83,11 @@ open class XMLNode: NSObject, NSCopying {
         public static let nodePromoteSignificantWhitespace = Options(rawValue: 1 << 28)
         public static let nodePreserveEmptyElements = Options([.nodeExpandEmptyElement, .nodeCompactEmptyElement])
         public static let nodePreserveQuotes = Options([.nodeUseSingleQuotes, .nodeUseDoubleQuotes])
+        #if !os(Android)
         public static let nodePreserveAll = Options(rawValue: Options([.nodePreserveNamespaceOrder, .nodePreserveAttributeOrder, .nodePreserveEntities, .nodePreservePrefixes, .nodePreserveCDATA, .nodePreserveEmptyElements, .nodePreserveQuotes, .nodePreserveWhitespace, .nodePreserveDTD, .nodePreserveCharacterReferences]).rawValue | UInt(bitPattern: 0xFFF00000))
+        #else
+        public static let nodePreserveAll = Options(rawValue: Options([.nodePreserveNamespaceOrder, .nodePreserveAttributeOrder, .nodePreserveEntities, .nodePreservePrefixes, .nodePreserveCDATA, .nodePreserveEmptyElements, .nodePreserveQuotes, .nodePreserveWhitespace, .nodePreserveDTD, .nodePreserveCharacterReferences]).rawValue | UInt(bitPattern: 0x7FF00000))
+        #endif
```
parkera: Why is this #if required?

Foundation/NSGeometry.swift
johnno1962: I’m still arguing with the compiler about exactly what is going on in NSGeiometry.swift re types at the moment a signifiant peice of code is effectively commented out.

CoreFoundation/Base.subproj/SwiftRuntime/CoreFoundation.h
johnno1962: This is the only way it would build for some reason.. I’ll take another look

Foundation/NSFileManager.swift
johnno1962: I’ll change this. I wanted the cases to be in order of likelyhood.

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
johnno1962: I’m just saving a line. I’ll separate them.

Foundation/NSTask.swift
johnno1962: This file does not compile at all on Android the way things are

Foundation/NSXMLNode.swift
johnno1962: I’ll check this. the 0xFFF00000 wouldn’t compile before.

Foundation/NSURLSession/NSURLSessionTask.swift
johnno1962: Operation queue is probably the first thing to look at as it is in Foundation. I have the suspicion that the dispatch thread problem is going to be difficult to resolve. I means that you can not make a call to Java from a background thread at the moment. There may have to be other strategies such as a barrier of some sort.

Any hints you can provide here would be greatly appreciated. I can’t even detemine the code path. Should I be looking at the libkqueue or libpwq code for where threads for queues are created and time out??

Foundation/NSFileManager.swift
johnno1962: agreed but this specific case is difficult to void/centralise.

Foundation/NSData.swift
johnno1962: How do I create a compatibility shim? In C?

Foundation/NSGeometry.swift
phausler: this is likely generating incorrect code since the host OS is leaking into the build product (per my commentary on the triples) where it is building with the flags of 64 bit but emitting 32 bit. Which is probably incorrect.

Foundation/NSLog.swift
johnno1962: ANLog?

Foundation/NSGeometry.swift
johnno1962: The generated build.ninja seemed correct. I'm thinking the file paths are a red herring.

Foundation/NSURLSession/NSURLSessionTask.swift
johnno1962: Right now I’m thinking a barrier for calls from background threads to Java through a thread that never exits is the way to go. Anything else is going to be very messy given the constraints Android imposes.

johnno1962:
New commit with responses to comments pushed. NSGeometry fixed and Operation queues work if you make sure  -DDEPLOYMENT_ENABLE_LIBDISPATCH is defined when you build. Less common code pathh does not seem to work. I’ll raise a ticket. Focus turns to the problem of detching threads if they have intereacted with JNI before they exit.

johnno1962:
Jira tickets creates SR 2587-2590

Foundation/CGFloat.swift
```diff
@@ -615,174 +619,178 @@ public func %=(lhs: inout CGFloat, rhs: CGFloat) {
...
 //===----------------------------------------------------------------------===//
 
+#if os(Android)
+import Glibc
+#endif
+
 @_transparent
 public func acos(_ x: CGFloat) -> CGFloat {
-    return CGFloat(acos(x.native))
+    return CGFloat(acos(x.double))
```
xwu: What is going on here? Doesn't this change the result of `acos(_: CGFloat)` for all 32-bit platforms?

Foundation/IndexPath.swift
```diff
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import CoreFoundation
```
xwu: This seems to be unused?

Foundation/NSURLSession/NSURLSessionTask.swift
```diff
@@ -838,7 +841,7 @@ extension URLSessionTask {
             var data = Data(capacity: bodyData!.length)
             data.append(Data(bytes: bodyData!.bytes, count: bodyData!.length)) 
 
-            s.delegateQueue.addOperation {
+	    s.delegateQueue.addOperation {
```
xwu: Nit: unintentional outdent?

Foundation/IndexPath.swift
johnno1962: It’s there for a free() in the code.

Foundation/NSURLSession/NSURLSessionTask.swift
johnno1962: sorted

Foundation/CGFloat.swift
johnno1962: acos() takes a double whereas CGFloat.native is a Float on 32 bit platforms

xwu: D'oh, of course. So that means these functions never did compile on 32-bit platforms.

I'm guessing you didn't mean to add `double` as part of the `CGFloat` public API, though. OTOH, I suppose it ought to be `public` if you're invoking from an `@_transparent` function. Would maybe `public var _doubleValue` be appropriate? And shouldn't it be `@_transparent` itself?

Foundation/CGFloat.swift
```diff
@@ -143,6 +143,10 @@ public struct CGFloat {
     
     /// The native value.
     public var native: NativeType
+
+    public var double: Double {
```
phausler: This is an addition to API surface area, can this not be done in the call sites?

Foundation/CGFloat.swift
phausler: This file was pulled mostly from the overlay which means that we need to keep it in sync as the overlay changes (and for note that builds 32 bit targets so it seems we are still missing something here)

Foundation/IndexPath.swift
phausler: free is not defined in CoreFoundation, it is defined in either Darwin or Glibc so the import should be of one of those.

Foundation/NSAffineTransform.swift
phausler: so perhaps it should be import Libc or System not Darwin then huh?

Foundation/NSLog.swift
```diff
@@ -41,3 +41,11 @@ public func NSLog(_ format: String, _ args: CVarArg...) {
         NSLogv(format, $0) 
     }
 }
+
+// hook for System.out.println()
+#if os(Android)
+public var ANLog = {
```
xwu: Does this need to be public? Also, aren't two-letter prefixes reserved by Apple?

Foundation/NSLog.swift
```diff
@@ -41,3 +41,11 @@ public func NSLog(_ format: String, _ args: CVarArg...) {
...
+  (msg: String) in
+  NSLog( msg )
```
xwu: Nit: code style--no spaces inside parens.

Foundation/NSSwiftRuntime.swift
```diff
@@ -14,10 +14,16 @@ import CoreFoundation
...
 #if os(OSX) || os(iOS)
 @_exported import Darwin
-#elseif os(Linux)
+#elseif os(Linux) //// || os(Android) // Causes ambiguity in NSScanner.swift
 @_exported import Glibc
 #endif
 
+#if os(Android) // shim required for bzero
+func bzero( _ ptr: UnsafeMutableRawPointer, _ size: size_t ) {
+    memset( ptr, 0, size )
```
xwu: Same nit: no spaces inside parens.

Foundation/NSFileManager.swift
```diff
@@ -9,10 +9,16 @@
 
 #if os(OSX) || os(iOS)
     import Darwin
-#elseif os(Linux)
+#elseif os(Linux) || os(Android)
     import Glibc
 #endif
 
+#if os(Android) // struct stat.st_mode is UInt32
+func &( left: UInt32, right: mode_t ) -> mode_t {
```
xwu: Best to annotate explicitly as `internal`? Also, same nit about space inside parens.

Foundation/NSXMLNode.swift
```diff
@@ -83,7 +83,12 @@ open class XMLNode: NSObject, NSCopying {
         public static let nodePromoteSignificantWhitespace = Options(rawValue: 1 << 28)
         public static let nodePreserveEmptyElements = Options([.nodeExpandEmptyElement, .nodeCompactEmptyElement])
         public static let nodePreserveQuotes = Options([.nodeUseSingleQuotes, .nodeUseDoubleQuotes])
+        #if !os(Android)
         public static let nodePreserveAll = Options(rawValue: Options([.nodePreserveNamespaceOrder, .nodePreserveAttributeOrder, .nodePreserveEntities, .nodePreservePrefixes, .nodePreserveCDATA, .nodePreserveEmptyElements, .nodePreserveQuotes, .nodePreserveWhitespace, .nodePreserveDTD, .nodePreserveCharacterReferences]).rawValue | UInt(bitPattern: 0xFFF00000))
+        #else
+	//// 0xFFF00000 is not a valid Int literal on 32 bit systems
```
xwu: If it's not valid on 32-bit systems generally, can we do better than `os(Android)`?

Foundation/CGFloat.swift
johnno1962: There are an swful lot of call sites.. I’ve made it internal

Foundation/NSXMLNode.swift
johnno1962: Better version on the way

Foundation/NSLog.swift
johnno1962: It’s used for debugging. It’s replaced by a closure calling System.out.println() by the app.

Foundation/CGFloat.swift
johnno1962: Noted & it’s not @_transparent. I tried making it internal and got a compier assertion:
SIL verification failed: function_ref inside fragile function cannot reference a private or hidden symbol: (SingleFunction && RefF->isExternalDeclaration()) || RefF->hasValidLinkageForFragileRef()

Foundation/IndexPath.swift
johnno1962: fixed

Foundation/CGFloat.swift
xwu: @_transparent functions aren't supposed to invoke internal methods. See: https://github.com/apple/swift/blob/master/docs/TransparentAttr.rst

It is *doable* to do this manually, yes? After all, it's just writing `Double(foo.native)` instead of `foo.double`.

Foundation/NSLog.swift
xwu: Maybe `AndroidLog` then, since `AN` is a two-letter prefix? But does this hook have to live in NSLog.swift?

Foundation/CGFloat.swift
johnno1962: There are many call sites. I wanted to centralise this code.

Foundation/NSLog.swift
johnno1962: It’s gotta live somewhere!

Foundation/NSFileManager.swift
johnno1962: done

Foundation/NSLog.swift
phausler: My suggestion is to in the backing for NSLog use android_log_print and similarly for the print function. 

Technically you could `funopen` a new descriptor for stdout to pipe logging on Android (I have done that on previous projects that I have worked on and it worked reasonably well for logging)

android/install.sh
phausler: So the build-script-impl needs to pass a target triple to the foundation configuration script. If that is not being passed you are building a strange hodgepodge beast of potentially incompatible compiled code.

Foundation/CGFloat.swift
xwu: Understandable. Having been bitten by a 10x regression when `&&` accidentally lost its `@_transparent` for a few weeks, my worry here is about the cost of an additional non-`@_transparent` function call at each of these conversions. Is that silly?

android/install.sh
johnno1962: “hodgepodge”? Pretty sure this is ok or it wouldn’t even link. https://github.com/SwiftJava/swift/blob/master/utils/build-script-impl#L2269 has been patched to pass in the triple and there is some extra code to make sure C and Swift get the same one: https://github.com/SwiftJava/swift-corelibs-foundation/blob/master/lib/script.py#L47

Foundation/CGFloat.swift
johnno1962: In the end it is @_transparent public

Foundation/NSLog.swift
johnno1962: Thanks for the pointer. The unpopular ANLog is no more and NSLog now works.

johnno1962:
Current status: Just about there. Found where the worker queues time out after 15 seconds to put in the hook https://github.com/SwiftJava/libpwq/blob/master/src/posix/manager.c#L295. Wrestling with Makefiles at the moment trying to get libpwq to rebuild.

johnno1962:
The above PR on libpwq has been prepared countaing the hook required and the changes to libdispatch are ready to submit https://github.com/SwiftJava/swift-corelibs-libdispatch/commit/ad8a41f4cbf385ad7123ed570b3154b87716c59e. It’ll wait to submit that until libpwq has merged and there is anothere PR related to an Android port in the pipline https://github.com/apple/swift-corelibs-libdispatch/pull/162.

This completes the port. How are you felling about merging this?

johnno1962:
I’ve made a pre-bilt toolchain (build directory) available http://johnholdsworth.com/android_toolchain.tgz. Hopefully, this will help get a bit more coverage on testing the port. Example code used in testing: https://github.com/SwiftJava/swift-android-samples/blob/master/swifthello/src/main/swift/Sources/main.swift

Foundation/CGFloat.swift
```diff
@@ -143,6 +143,10 @@ public struct CGFloat {
     
     /// The native value.
     public var native: NativeType
+
+    @_transparent public var double: Double {
+        return Double(native)
+    }
```
parkera: This still doesn't feel right to me for two reasons:

1. I don't like phrasing this as a type conversion hidden as a property.
2. It cannot be public.
3. It feels like a separate issue than about Android, and probably should be its own PR so we can figure it out independently.

Foundation/NSAffineTransform.swift
parkera: Anyone want to file a bug for the Swift project? :)

Foundation/NSSwiftRuntime.swift
```diff
@@ -14,10 +14,16 @@ import CoreFoundation
 // This mimics the behavior of the swift sdk overlay on Darwin
 #if os(OSX) || os(iOS)
 @_exported import Darwin
-#elseif os(Linux)
+#elseif os(Linux) //// || os(Android) // Causes ambiguity in NSScanner.swift
```
parkera: I would like a further explanation of the comment, and/or a bug to go with it and noted here, so that future maintainers can understand why this is commented out.

Foundation/NSTask.swift
```diff
@@ -7,6 +7,7 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+#if !os(Android) // not available
```
parkera: Still not happy about this.

Is there a way, instead, to mark the class as unavailable on the Android platform?

Foundation/NSURLSession/NSURLSessionTask.swift
parkera: And put the JIRA number in the comment so we can find it later.

parkera:
I added a few more comments; if we can address those we're pretty close to merging this.

Primarily I would like to split up the 32-bit CGFloat thing into a separate PR.

Foundation/CGFloat.swift
johnno1962: Can you tell me what you would be happy with here? Potentially brittle casts at all call sites? A function? It can be internal but not also @_transparent. I can then prepare a separate PR and look at merging it into this one.

Foundation/NSSwiftRuntime.swift
johnno1962: Glibc should be included here to make import Foundation be an implicit import of Glibc so I’ll change this. It causes ambiguity on FLT/DBL_MAX/MIN in NSScanner.swift

// these might be good to have in the stdlib
extension Float : _FloatLike {
    static var max: Float { return FLT_MAX }
    static var min: Float { return FLT_MIN }
}

Foundation/NSTask.swift
johnno1962: Can you tell me the syntax to make specify a NSTask unavailable for android please.

johnno1962:
OK, I’ll focus on CGFloat etc in a separte PR and pick up the peices once that has been merged.

Foundation/CGFloat.swift
johnno1962: Reflecting on this it seems people expect a cast to Double at the call sites. I’ll do this.

Foundation/NSSwiftRuntime.swift
xwu: These constants are known in the standard library as greatestFiniteMagnitude and leastFiniteMagnitude. The spellings max and min were considered and explicitly rejected, and any such additions to the public API would have to go through the evolution process.

johnno1962: Do’h. That was already in the NSScanner.swift code before I came along.

johnno1962:
Separate PR raised for the CGFloat/NSGeometry.swift conversions (see above)

Foundation/NSTask.swift
johnno1962: Had a quick look at removing the #if !os(Android), got the following:

Foundation/NSTask.swift:95:2: warning: unknown platform 'Android' for attribute 'available'
@available(Android, unavailable)
 ^

Foundation/NSSwiftRuntime.swift
xwu: You misunderstand. That code is fine in NSScanner as an internal extension. But it cannot be a public API in the stdlib.

xwu: I should rephrase: if the code had no other issues, it would be fine where it is, but it cannot be a public API because there is a public API in the stdlib that provides the same facility.

In fact, the code in NSScanner is there because it pre-dates the revised FloatingPoint protocols and works around a common limitation of floating point protocols prior to their current state. The internal protocol _FloatLike could actually be ripped out entirely and replaced with the new FloatingPoint, and probably should be if it's causing you grief here. That would be a rather straightforward thing to implement.

johnno1962: Foundation builds with your new version of NSScanner.swift, thanks!

Foundation/NSTask.swift
parkera: What does it take to add Android as a platform for the compiler to understand here? That may be a prereq to anyone doing anything meaningful on Android in the first place.

modocache: > Can you tell me the syntax to specify a NSTask unavailable for android please.

I think this means adding Android to [the list of platforms that the Swift compiler knows about](https://github.com/apple/swift/blob/bfc7924995d4de543e7b685faf49e8d25d563ce1/include/swift/AST/PlatformKinds.def). Adding that (and maybe some other tweaks) should allow us to write `@available(Android, unavailable)`. Does that sound right to you, @parkera?

Foundation/NSAffineTransform.swift
modocache: I [started this discussion](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160307/012085.html) on the swift-evolution mailing list, but was too lazy to push it forward. @compnerd is now [leading the charge](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160704/023534.html). 👍

johnno1962:
Closed in favour of https://github.com/apple/swift-corelibs-foundation/pull/632 which builds for me on Android & https://github.com/apple/swift-corelibs-foundation/pull/631. CGFloat.swift changes are not required after complete rebuild of toolchain from master.

Foundation/NSTask.swift
johnno1962: It’s odd. The compiler knows about #if os(Android) giving an error if you mistype it but not @available(Android, unavailable)

Foundation/NSGeometry.swift
```diff
@@ -884,22 +882,20 @@ extension NSCoder {
 
 private extension NSCoder {
     func _encodeCGFloat(_ value: CGFloat) {
-        if let keyedArchiver = self as? NSKeyedArchiver {
-            keyedArchiver._encodeValue(NSNumber(value: value.native))
-        } else {
-            NSUnimplemented()
+        guard let keyedArchiver = self as? NSKeyedArchiver else {
+            preconditionFailure("Unkeyed coding is unsupported.")
```
xwu: I think these should be NSUnimplemented()

parkera: Putting aside this oddity of the abstract protocol knowing about NSKeyedArchiver for a moment, instead of letting NSKeyedArchiver implement this correctly -- we've decided across the project to just not support unkeyed coding instead of haphazardly implementing in some places and not others.

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -1265,7 +1277,7 @@ CFDictionaryRef __CFGetEnvironment() {
         extern char **environ;
         char **envp = environ;
 #elif DEPLOYMENT_TARGET_LINUX
-#ifndef environ
+#if !defined(environ) && !defined(__ANDROID__)
```
xwu: Is this right? Do you really intend to leave environ undefined on Android?

Foundation/IndexPath.swift
```diff
@@ -10,6 +10,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if os(Android)
+import Glibc // for free()
```
xwu: Out of curiosity, why is this only needed for Android but not for Linux?

Foundation/NSFileManager.swift
```diff
@@ -9,10 +9,16 @@
 
 #if os(OSX) || os(iOS)
     import Darwin
-#elseif os(Linux)
+#elseif os(Linux) || os(Android)
     import Glibc
 #endif
 
+#if os(Android) // struct stat.st_mode is UInt32
```
xwu: Here too, does this also apply to Linux?

Foundation/NSGeometry.swift
```diff
@@ -282,25 +282,23 @@ extension NSEdgeInsets: NSSpecialValueCoding {
...
     init?(coder aDecoder: NSCoder) {
-        if aDecoder.allowsKeyedCoding {
-            self.top = aDecoder._decodeCGFloatForKey("NS.edgeval.top")
-            self.left = aDecoder._decodeCGFloatForKey("NS.edgeval.left")
-            self.bottom = aDecoder._decodeCGFloatForKey("NS.edgeval.bottom")
-            self.right = aDecoder._decodeCGFloatForKey("NS.edgeval.right")
-        } else {
-            NSUnimplemented()
+        guard aDecoder.allowsKeyedCoding else {
+            preconditionFailure("Unkeyed coding is unsupported.")
```
xwu: NSUnimplemented instead of preconditionFailure, I think.

CoreFoundation/Base.subproj/CFUtilities.c
modocache: I believe `environ` is already defined on Android: https://github.com/apple/swift/blob/e6dec58cc5e08a59028310aac14d1e8fbc46c2d4/stdlib/private/SwiftPrivateLibcExtras/Subprocess.swift#L292

So in theory this additional `!defined(__ANDROID__)` check may not be necessary at all...?

Foundation/NSTask.swift
modocache: Yeah, that's a good point. I'm not 100% sure -- I just grepped around, found that spot, and thought it looked interesting. :)

If adding Android to that list is indeed what @parkera had in mind, we could email the swift-dev mailing list to learn more about it.

Foundation/NSFileManager.swift
johnno1962: Seems to be an Android NDK thing

Foundation/IndexPath.swift
johnno1962: I’ll removed this. There is an @_exported import of Glibc in Foundation/NSSwiftRuntime.swift

CoreFoundation/Base.subproj/CFUtilities.c
johnno1962: environ is already defined on Android. No need to redefine it to __environ.

Foundation/NSGeometry.swift
johnno1962: I’ve refeshed this file from master.

CoreFoundation/Base.subproj/CFUtilities.c
xwu: So this change can be reverted, yes?

Foundation/NSAffineTransform.swift
johnno1962: This boilerplate goes away for most users as if they import Foundation as there is an @_exported import of the right C library in NSSwiftRuntime.swift.

CoreFoundation/Base.subproj/CFUtilities.c
johnno1962: I’m saying it’s needed, honest! Full code is:

#elif DEPLOYMENT_TARGET_LINUX
#if !defined(environ) && !defined(__ANDROID__)
#define environ __environ
#endif
        char **envp = environ;
#endif

xwu: I'm confused, if environ is already defined in Android, then the second half of your #if is never evaluated; if environ is undefined in Android, then what is envp on Android?

Am I missing something?

johnno1962: envrion is a C symbol not a #define for the preprocessor so you can’t test for whether it exists with a #ifdef.

xwu: Got it.

CoreFoundation/Base.subproj/SwiftRuntime/CoreFoundation.h
johnno1962: This problem has gone away since I completely rebuilt the toolchain.

johnno1962:
I’m thinking this PR is ready again now thanks to @xwu and @rintaro resolving the various typing problems and I’ve checked macOS and Linux still build. The build is far from automated at the moment, waiting on https://github.com/apple/swift-corelibs-libdispatch/pull/162 after which I’ll add back in the thread handler support which as been merged into libpwq upstream https://github.com/mheily/libpwq/pull/18. That would still leave automating the build of libxml and libcurl for the Android platform somehow but this PR to Foundation itself is ready to go.

Foundation/NSFileHandle.swift
```diff
@@ -15,6 +15,12 @@ import Darwin
 import Glibc
 #endif
 
+#if os(Android) // legacy constants
+let L_SET  = SEEK_SET
+let L_INCR = SEEK_CUR
+let L_XTND = SEEK_END
+#endif
```
rintaro: Let's use `SEEK_` variant all together: `lseek(_fd, off_t(offset), SEEK_SET)`
AFAIK, all supported platform has these macro.

xwu:
@johnno1962, would you mind rebasing? I'm having a little trouble studying the diffs :)

johnno1962:
@xwu, I would if I knew how to do it! Create a PR from apple/swift-corelibs-foundation/master back to SwiftJava/swift-corelibs-foundation/master and merge it??

xwu:
Nah, see:
https://git-scm.com/book/en/v2/Git-Branching-Rebasing

Looks like you're working on the master branch you checked out, so you'll need to do a little musical chairs so that you can pull apple/swift-corelibs-foundation master into your repo.

johnno1962:
Musical chairs, Welp! I was looking at that page. Simple instructions please!
Lesson seems to be I shouldn’t be working on master.

rintaro:
Could you give us `git remote -v` result?

johnno1962:
johnno@ubuntu:~/Swift/swift-corelibs-foundation$ git remote -v
origin	https://github.com/SwiftJava/swift-corelibs-foundation.git (fetch)
origin	https://github.com/SwiftJava/swift-corelibs-foundation.git (push)
johnno@ubuntu:~/Swift/swift-corelibs-foundation$

johnno1962:
How about I start again with a new fork, new branch, copy everything changed across, then new PR?

rintaro:
```
$ git checkout master
$ git remote add apple https://github.com/apple/swift-corelibs-foundation.git
$ git fetch apple
$ git rebase apple/master
$ git push -f origin
```

rintaro:
Hm, I checked out your repo, and tried rebasing.
I see some conflicts on `git rebase apple/master` step.
To abort rebasing, `git rebase --abort`

johnno1962:
How does it look now?

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -25,6 +25,9 @@
 #if DEPLOYMENT_TARGET_WINDOWS
 #include <process.h>
 #endif
+#ifdef __ANDROID__
```
phausler: @parkera should we perhaps instead of using macros defined by the compiler use a DEPLOYMENT_TARGET_ANDROID for these?

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -764,7 +767,16 @@ void CFLog(CFLogLevel lev, CFStringRef format, ...) {
 #if DEPLOYMENT_RUNTIME_SWIFT
 // Temporary as Swift cannot import varag C functions
 void CFLog1(CFLogLevel lev, CFStringRef message) {
+#ifdef __ANDROID__
+    UInt8 buffer[4096];
+    CFIndex usedBufLen = sizeof buffer-1;
+    CFStringGetBytes(message, CFRangeMake(0, CFStringGetLength(message)),
+		     kCFStringEncodingUTF8, ' ', FALSE, buffer, usedBufLen, &usedBufLen);
+    buffer[usedBufLen] = '\000';
+    __android_log_print(ANDROID_LOG_INFO, "Swift", "%s", buffer);
```
phausler: For a tag it should probably be something other than "Swift". the process name would probably be more what users are expecting. Additionally the __android_log_print level enum can map reasonably well onto CFLogLevel.

Note: this is just window dressing imho it just makes the experience better and is not a blocking item for this PR

CoreFoundation/String.subproj/CFString.c
```diff
@@ -1264,12 +1264,16 @@ CF_PRIVATE CFStringRef __CFStringCreateImmutableFunnel3(
 	contentsDeallocator = __CFGetDefaultAllocator();
     }
 
+    #ifndef __ANDROID__
+    // https://bugs.swift.org/browse/SR-2587
+    // crashes String.swift, line 51 when creating String from zero length Data
```
phausler: IIRC bionic's malloc is built with error checking enabled such that free is a bit more pedantic on heap corruption (and has some gnarly behavioral differences) 

It would be nice to see a backtrace and values of locals at the crash point in the bug, not everyone who might be versed to fix this has easy access to an Android device and development environment for them ;)

android/install.sh
phausler: Hmm it looks like the "error" in the script actually is:
`SWIFT_BUILD_PATH="$(build_directory ${host} swift)"`
etc.

That is claiming to emit the build directory as the host.

lib/script.py
```diff
@@ -39,14 +39,15 @@ def generate_products(self):
...
             verbose_flags += "-v"
         verbose_flags += "\n"
-        swift_triple = Configuration.current.target.swift_triple
+        triple = Configuration.current.target.triple
         base_flags = """
-TARGET                = """ + Configuration.current.target.triple + """
+TARGET                = """ + triple + """
 DSTROOT               = """ + Configuration.current.install_directory.absolute() + """
 """
+        swift_triple = triple if triple == "armv7-none-linux-androideabi" else Configuration.current.target.swift_triple
```
phausler: It seems like this should actually not be logic here but instead live in `lib/target.py` instead

Foundation/NSTask.swift
phausler: @parkera  is having `@available(Android, unavailable)` a blocker for this PR? IMHO it would be the ideal solution here.

Foundation/NSURLSession/NSURLSessionTask.swift
```diff
@@ -630,9 +630,12 @@ fileprivate extension URLSessionTask {
         var result = [("Connection", "keep-alive"),
                       ("User-Agent", userAgentString),
                       ]
+        #if !os(Android)
+        // Crashes on Android https://bugs.swift.org/browse/SR-2588
```
phausler: The bug claims no stack trace, without one the chances of this being fixed is low. Does lldb not work for Android? it should be able to interface with the gdbserver loaded with the apk.

android/README.md
```diff
@@ -0,0 +1,51 @@
...
+downloading the swift sources use the script "prepare.sh" in this
+directory to install prebuilt binaries and headers for libxml,
+libcurl and libdispatch then run the script "builder.sh".
+
+This build requires the path to a [r12 Android NDK](http://developer.android.com/ndk/downloads/index.html) in the
+`ANDROID_NDK_HOME` environment variable and the path to the
+android port of the "icu" libraries in `ANDROID_ICU_UC`
+downloaded from [here](https://github.com/SwiftAndroid/libiconv-libicu-android/releases/download/android-ndk-r12/libiconv-libicu-armeabi-v7a-ubuntu-15.10-ndk-r12.tar.gz).
+The port was tested against api 21 on a Android v5.1.1 LG K4 Phone (Lollipop)
+and requires an Ubuntu 15 host with the Android NDK Gold linker from
```
phausler: So this won't build as a darwin host? that seems subpar; what would it take to fix that to actually build on a primary development platform of swift?

phausler:
So this only supports armv7 (no neon right?) and not arm64 yet?
Also it is worth noting that currently the CI for swift does not build anything other than Ubuntu and Darwin based platforms. What can we expect as per CI for this as a target? I would expect that as we iterate there may be portions where the Android target might break due to changes in the other sides of code-branches etc.

android/README.md
johnno1962: I’ll look into it. As far as I know the original Android swift compiler port which this is based on specifies Ubuntu 15. https://github.com/apple/swift/blob/master/docs/Android.md

johnno1962:
Yes armv7 only as per the compiler though some 32/64 bit wrinkles have been ironed out along the way. We’re a long way from CI for Android though that would be the longer tem goal. Most notably the libdispatch build has an outstanding PR https://github.com/apple/swift-corelibs-libdispatch/pull/162 then there is ibxml2 and libcurl to work into a build for Android somehow. It will take some time for a cross compile to bed into the build process. Right now these dependencies are provided as binaries for anyone wanting to work on Foundation for Android itself.

modocache:
> What can we expect as per CI for this as a target?

I would love to add CI for Android. I had been meaning to email @tkremenek but haven't had the time -- apologies. I'll send that email today.

If Apple isn't interested in supporting an Android target in @swift-ci, I can set up a @swift-android-ci GitHub bot that can be used to test Android.

> I would expect that as we iterate there may be portions where the Android target might break due to changes in the other sides of code-branches etc.

Breakages occur (albeit these days pretty rarely) for apple/swift due to the lack of CI as well. I fix them when they pop up. CI would be ideal, but I this works well enough for now -- there are only a handful of consumers of the Swift stdlib for Android, so hardly anyone notices when it breaks. I imagine this is the same for Linux ARM targets.

> So this won't build as a darwin host? that seems subpar; what would it take to fix that to actually build on a primary development platform of swift?

As luck would have it, I started working on cross-compiling the Swift stdlib for Android from a macOS host this morning. I think it's possible, in large part thanks to work like https://github.com/apple/swift/pull/2497 and https://github.com/apple/swift/pull/2560. However, there are still quite a few  assumptions baked into Swift's CMake build system: "If my host machine is macOS, then pass the macOS SDK path to all clang++ invocations...", etc. I'm working on fixing these now.

CoreFoundation/Base.subproj/CFUtilities.c
johnno1962: Mapped the log levels but I couldn’t find a way to get the app name without reching back to Java through JNI.

lib/script.py
johnno1962: Moved into target.py

tkremenek:
@modocache The idea we are pursuing is to add another Jenkins master that could be used to add additional machines for different platforms.  The second master would be for security reasons, but the intent would be to enable more CI testing on additional platforms the community cares about, including Android.  Things are still being investigated which is why it hasn't been communicated on swift-dev yet.

modocache:
Thanks, @tkremenek! Please loop me in on those talks at whatever stage you think is appropriate. I'd be happy to contribute developer time, hardware, or other resources.

> What can we expect as per CI for this as a target?

In the short term, I propose no official CI. If the Android build breaks, its users can follow up with patches. That's how I've been handling Swift's Android support since April. We can add CI for Foundation on Android once the mechanism @tkremenek described is in place.

Foundation/NSURLSession/NSURLSessionTask.swift
johnno1962: I’ll look into it. Do you have any pointers? Is this the sort of thing: https://github.com/mapbox/mapbox-gl-native/wiki/Android-debugging-with-remote-GDB

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -764,7 +767,28 @@ void CFLog(CFLogLevel lev, CFStringRef format, ...) {
 #if DEPLOYMENT_RUNTIME_SWIFT
 // Temporary as Swift cannot import varag C functions
 void CFLog1(CFLogLevel lev, CFStringRef message) {
+#ifdef __ANDROID__
+    android_LogPriority priority = ANDROID_LOG_UNKNOWN;
+    switch ( lev ) {
```
xwu: Nit (sorry!)--spaces inside parens differs from code style in this file.

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -764,7 +767,28 @@ void CFLog(CFLogLevel lev, CFStringRef format, ...) {
...
+        case kCFLogLevelCritical: priority = ANDROID_LOG_ERROR; break;
+        case kCFLogLevelError: priority = ANDROID_LOG_ERROR; break;
+        case kCFLogLevelWarning: priority = ANDROID_LOG_WARN; break;
+        case kCFLogLevelNotice: priority = ANDROID_LOG_WARN; break;
+        case kCFLogLevelInfo: priority = ANDROID_LOG_INFO; break;
+        case kCFLogLevelDebug: priority = ANDROID_LOG_DEBUG; break;
+    }
+    CFIndex blen = message ? CFStringGetMaximumSizeForEncoding(CFStringGetLength(message), kCFStringEncodingUTF8) + 1 : 0;
+    char *buf = message ? (char *)malloc(blen) : 0;
+    if ( buf ) {
```
xwu: Nit--here too, and once more inside the block.

android/README.md
```diff
@@ -0,0 +1,51 @@
...
+[https://github.com/mheily/libkqueue](https://github.com/mheily/libkqueue)
+
+To build these the recipe was generally the same. autogen or configure
+for Linux then alter their Makefiles to have CFLAGS = include: --target=armv7-none-linux-androideabi --sysroot=$(ANDROID_NDK_HOME)/platforms/android-21/arch-arm.
+
+There is a known issue when libdispatch background tasks exit
+they will cause an exception as DetachCurrentThread has not
+been called. To avoid this your app must include the line:
+
+   DispatchGroup.threadCleanupCallback = JNI_DetachCurrentThread
```
xwu: Not primarily related to this commit, but if a thread cleanup hook could be useful for platforms other than Android, it'd be nice to have a Swift evolution proposal to add it as part of the official public API of the Swift libdispatch overlay.

Otherwise, I'd imagine you'll be asked to prefix the name with an underscore, i.e.: `_threadCleanup[Callback|Handler]` (and I think "handler" is the terminology used in the overlay).

android/package.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#
+# prepare binay package used to build Foundtion for Android
```
xwu: Last nit: spelling and capitalization here :P

android/install.sh
johnno1962: Trying not to tangle with “build-script-impl”. I understand it’s being refactored to take into account cross compiles https://github.com/apple/swift/pull/4662

CoreFoundation/Base.subproj/CFUtilities.c
johnno1962: I can change it back if you like.

CoreFoundation/String.subproj/CFString.c
johnno1962: I think I’ve been able to resolve this problem. Constant strings have a fake isa which wasn’t being picked up or linked correctly so I created an alias in the source. The Locale problem was just a manifestation of the same problem as it used CONST_STRING_DECL().

#ifdef __ANDROID__
// Avoids crashes on Android
// https://bugs.swift.org/browse/SR-2587
// https://bugs.swift.org/browse/SR-2588
// Seemed to be a linker/relocation? problem.
// CFStrings using CONST_STRING_DECL() were not working
// Applies reference to _NSCFConstantString's isa here
// rather than using a linker option to create an alias.
#define __CFConstantStringClassReference _TMC10Foundation19_NSCFConstantString
#endif

Foundation/NSURLSession/NSURLSessionTask.swift
johnno1962: This is resolved. See above

android/package.sh
johnno1962: Nit removed. Obviously a file written in a hurry.

johnno1962:
Closed https://bugs.swift.org/browse/SR-2587 & https://bugs.swift.org/browse/SR-2588 so there are now no known issues apart from whether waiting for support for @available(Android, unavailable) in the compiler should be a blocker.

tkremenek:
@modocache Let's dicsuss extending the CI on `swift-dev@swift.org`, not in a pull request.

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -787,7 +787,13 @@ void CFLog1(CFLogLevel lev, CFStringRef message) {
     if (maxLength > sizeof(stack_buffer) / sizeof(stack_buffer[0])) {
         buffer = calloc(sizeof(char), maxLength);
     }
-    CFStringGetCString(message, buffer, maxLength, encoding);
+    if ( maxLength == 1 )
```
Coeur: no spaces inside parenthesis

Foundation/NSTask.swift
johnno1962: Jira https://bugs.swift.org/browse/SR-2670 has a response - currently not implemented.

android/README.md
johnno1962: I’m waiting for https://github.com/apple/swift-corelibs-libdispatch/pull/162 to land. Then we can finalise the name. Handler sounds good to me.

johnno1962:
Ready to go my end. All issues resoved, @available(Android, unavailable) will have to wait.

Foundation/NSObjCRuntime.swift
```diff
@@ -196,6 +196,9 @@ internal func NSRequiresConcreteImplementation(_ fn: String = #function, file: S
 }
 
 internal func NSUnimplemented(_ fn: String = #function, file: StaticString = #file, line: UInt = #line) -> Never {
+    #if os(Android)
+    NSLog("\(fn) is not yet implemented. \(file):\(line)")
+    #endif
```
parkera: Why did we add an NSLog here?

parkera:
@swift-ci please test

parkera:
Let's give this a shot.

Foundation/NSObjCRuntime.swift
johnno1962: fatalError() doesn’t print anything on Android (though asserts do). Just giving people a pointer.

phausler: shouldn't fatalError call __android_log_print on android?

johnno1962: It should but this is a Foundation PR. I’ll take a look. Annoyingly using __android_log_print requires a -llog when you link so it wouldn’t be a trivial change.

parkera:
We can follow up with separate changes for the rest. Thanks for tackling this @johnno1962 !

johnno1962:
Thanks Tony 👍

Foundation/NSObjCRuntime.swift
johnno1962: Jira https://bugs.swift.org/browse/SR-2734

johnno1962:
Reminder, Precompiled toolchain here http://johnholdsworth.com/android_toolchain.tgz

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -764,7 +767,45 @@ void CFLog(CFLogLevel lev, CFStringRef format, ...) {
...
+
+    if (message == NULL) message = CFSTR("NULL");
+
+    char stack_buffer[1024] = { 0 };
+    char *buffer = &stack_buffer[0];
+    CFStringEncoding encoding = kCFStringEncodingUTF8;
+    CFIndex maxLength = CFStringGetMaximumSizeForEncoding(CFStringGetLength(message), encoding) + 1;
+
+    if (maxLength > sizeof(stack_buffer) / sizeof(stack_buffer[0])) {
+        buffer = calloc(sizeof(char), maxLength);
```
lhoward: should check calloc() return value

johnno1962: Too late now.

parkera: Not too late - open up another PR. =)

johnno1962: Alright already, new PR https://github.com/apple/swift-corelibs-foundation/pull/655

zayass:
Hi @johnno1962 

> it's not been able to automate the libdispatch crosscompile 

Can you little explain why it is hard to automate and how you build this libdispatch binary?

johnno1962:
Things have moved on a little since this PR. Some detail about how to build libdispatch are on this PR https://github.com/apple/swift-corelibs-libdispatch/pull/162

PR #623, opened 2016-09-07
Implementing NSDictionary.descriptionInStringsFileFormat

saiHemak:
Also includes fix for displaying nested dictionaries properly


parkera:
I've always been suspicious of this API; feels legacy and better suited for NSPropertyListSerialization. But anyway it's there and we may as well implement it.

parkera:
@swift-ci please test

PR #625, opened 2016-09-09
Allow httpAdditionalHeaders to contain Strings or NSStrings

seabaylea:
`NSURLSessionConfiguration.httpAdditonalHeaders` is declared to accept `[AnyHashable:Any]`, but internally this maps to `[String : String]`.

Currently the code assumes that the `[AnyHashable:Any]` actually only contains `NSString`s, whereas on macOS/Darwin its acceptable for either `NSString`s or `String`s to be used.

I've therefore updated `URLSession._Configuration.convertToStringString()` to allow either and coerce to String if needed.


parkera:
Cool

parkera:
@swift-ci please test

PR #626, opened 2016-09-12
Remove non-keyed encoding for data objects

alblue:
Instead of having non-keyed encoding supported by some but not
all foundation types, remove the non-keyed encoding everywhere
and make passing a non-keyed encoder a preconditionFailure.


parkera:
@swift-ci please test linux

PR #630, opened 2016-09-13
NSSet/CFSet bridging

henrybetts:
Implements the bridge between NSSet and CFSet in the same way as the other NSCF classes.
I believe that NSSet member() now behaves as expected, returning the actual object stored in the set, rather than the compared value.

Also includes a fix that allows bridged dictionaries to be copied.


parkera:
@phausler can you take a look at this?

parkera:
@henrybetts In the meantime, it would be great to add some test cases here.

phausler:
as a cursory glance this looks about right

@swift-ci Please test

parkera:
Thanks!

PR #631, opened 2016-09-13
Update NSScanner to use BinaryFloatingPoint

xwu:
Migrate from internal `_FloatLike` protocol to `BinaryFloatingPoint`.

Incidentally fix a subtle bug and align to Darwin Foundation:

* We don't want to return `FLOAT_MIN` when the string has a negative number with magnitude too great to be representable in a `Float` (`FLOAT_MIN` isn't like `Int.min`!)
* Actually, we don't even want `-Float.greatestFiniteMagnitude`, since Darwin Foundation returns `-Float.infinity`

I've adjusted the behavior accordingly and written a test.

Annoyingly, one test is currently failing on my machine, but it fails both with and without this patch and must be unrelated (i.e., it is failing with a freshly checked-out copy of swift-corelibs-foundation on XCode 8 GM):

```
error: TestNSData.test_replaceBytes : XCTAssertEqual failed: ("<0008090a 00>") is not equal to ("<0008090a 00>")
```

### DIFF
--- a/Foundation/NSScanner.swift
+++ b/Foundation/NSScanner.swift
@@ -369,7 +369,7 @@ extension String {
             let limit3 = T(numeral - (neg ? 1 : 0)) >= T.greatestFiniteMagnitude - localResult * T(10)
             if (limit1) && (limit2 || limit3) {
                 // apply the clamps and advance past the ending of the buffer where there are still digits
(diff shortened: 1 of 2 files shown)


parkera:
Thanks @xwu!

parkera:
@swift-ci please test

PR #635, opened 2016-09-13
Fix for a URLSession crash (SR-2630)

pushkarnk:
I am abstaining from adding the tests from SR-2630 to TestFoundation because the reach out to external URLs and I saw them hang on a few instances.

### DIFF
--- a/Foundation/NSURLSession/NSURLSessionTask.swift
+++ b/Foundation/NSURLSession/NSURLSessionTask.swift
@@ -833,10 +833,13 @@ extension URLSessionTask {
             guard case .inMemory(let bodyData) = bodyDataDrain else {
                 fatalError("Task has data completion handler, but data drain is not in-memory.")
             }
+
             guard let s = session as? URLSession else { fatalError() }
-           
-            var data = Data(capacity: bodyData!.length)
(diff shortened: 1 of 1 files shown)


Foundation/NSURLSession/NSURLSessionTask.swift
```diff
@@ -833,10 +833,13 @@ extension URLSessionTask {
...
+
             guard let s = session as? URLSession else { fatalError() }
-           
-            var data = Data(capacity: bodyData!.length)
-            data.append(Data(bytes: bodyData!.bytes, count: bodyData!.length)) 
+
+            var data = Data()
+            if let body = bodyData {
+                data.append(Data(bytes: body.bytes, count: body.length))
+            }
```
parkera: Shouldn't we be able to use `public init(bytes: UnsafeRawPointer, count: Int)` on Data here?

pushkarnk:
Thanks @parkera . The ```Data.append()``` was unnecessary. Also, if there is no body, we need to call the handler with an empty Data object (not nil) for compatibility with OS X. I've updated the patch

parkera:

PR #636, opened 2016-09-13
SR-2683 URLSessionDelegate didCompleteWithError callback never triggered

seabaylea:
SR-2638 reports that the following URLSessionDelegate callback is never triggered:
```
func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?)
```
This is actually caused by a function signature issue - I'm not sure why the compiler isn't picking it up.

On Linux, the function signature for the callback is currently using the old style:
```
internal func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: NSError?)
```
ie. its specified as an `NSError` instead of `Error`.

This PR switches over the API, and coerces the `NSError` to an `Error` before calling the callback.


parkera:

PR #639, opened 2016-09-15
Implementation of NSMutableDictionary.init(contentsOfFile:)

saiHemak:


Foundation/NSDictionary.swift
```diff
@@ -563,8 +563,20 @@ open class NSMutableDictionary : NSDictionary {
...
-    public convenience init?(contentsOfFile path: String) { NSUnimplemented() }
-    public convenience init?(contentsOfURL url: URL) { NSUnimplemented() }
+    public convenience init?(contentsOfFile path: String) {
+        self.init(contentsOfURL: URL(fileURLWithPath: path))
+    }
+    
+    public convenience init?(contentsOfURL url: URL) {
+        do {
+            guard let plistDoc = try? Data(contentsOf: url) else { return nil }
+            let plistDict = try PropertyListSerialization.propertyList(from: plistDoc, options: [], format: nil) as? Dictionary<AnyHashable,Any>
```
phausler: NSMutableDictionary technically will use NSPropertyListMutableContainers however I am not fully certain how much value that adds since all struct Dictionaries are mutable via var declarations

parkera:

PR #640, opened 2016-09-15
Fix for SR-2631 and implementing URLSession.finishTasksAndInvalidate

pushkarnk:
This PR primarily addresses the premature URLSession object cleanup reported in[ SR-2631](https://bugs.swift.org/browse/SR-2631). 

As a solution to SR-2631, we now maintain a reference to the session object from the task object creating a retain cycle. So, this PR also takes care of breaking such cycles on task completion. 

Another type of retain cycle can exist between session objects and the delegates registered with them (see point 15 [here](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/URLLoadingSystem/NSURLSessionConcepts/NSURLSessionConcepts.html)), the ```finishTasksAndInvalidate``` method breaks these cycles per the URLSession specification.

### DIFF
--- a/Foundation/NSURLSession/NSURLSession.swift
+++ b/Foundation/NSURLSession/NSURLSession.swift
@@ -189,8 +189,9 @@ open class URLSession : NSObject {
     /// This queue is used to make public attributes on `URLSessionTask` instances thread safe.
     /// - Note: It's a **concurrent** queue.
     internal let taskAttributesIsolation: DispatchQueue 
-    fileprivate let taskRegistry = URLSession._TaskRegistry()
+    internal let taskRegistry = URLSession._TaskRegistry()
     fileprivate let identifier: Int32
+    fileprivate var invalidated = false
     
     /*
      * The shared session uses the currently set global NSURLCache,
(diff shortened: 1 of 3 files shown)


Foundation/NSURLSession/NSURLSession.swift
```diff
@@ -258,7 +259,27 @@ open class URLSession : NSObject {
      * session with the same identifier until URLSession:didBecomeInvalidWithError: has
      * been issued.
      */
-    open func finishTasksAndInvalidate() { NSUnimplemented() }
+    open func finishTasksAndInvalidate() {
+       //we need to return immediately
+       workQueue.async {
+           //don't allow creation of new tasks from this point onwards
+           self.invalidated = true
```
parkera: This seems a bit racy since invalidated is not checked on the same queue elsewhere, but since it's effectively a debugging tool I guess that's ok,

parkera:

Foundation/NSURLSession/NSURLSession.swift
pushkarnk: Thanks for pointing this out. I'll do something about it.

PR #643, opened 2016-09-16
Allow numeric types in NSNumberFormatter.string(for:)[SR-2477]

pushkarnk:
Reported by https://bugs.swift.org/browse/SR-2477

In the absence of coercion of Swift types to Foundation types on Linux, this is to bring about functional similarity with Darwin.


phausler:
@swift-ci Please test

PR #644, opened 2016-09-16
Fix for [SR-2486] Crash in NSDictionary.description with NSSet inside

saiHemak:


phausler:

Foundation/Set.swift
```diff
@@ -40,8 +40,13 @@ extension Set : _ObjectTypeBridgeable {
                 if let o = obj as? Element {
                     set.insert(o)
                 } else {
-                    failedConversion = true
-                    stop.pointee = true
+                    // here obj must be a swift type
+                    if let nsObject = _SwiftValue.store(obj) as? Element {
```
xwu: Nit: can you make this an "else if" to avoid nested "if"?

xwu: Edit--never mind, since it's being merged

PR #645, opened 2016-09-17
Change RegularExpression back to NSRegularExpression

pushkarnk:


phausler:

pushkarnk:
@phausler : The URLSession data task test timed out here :-( Can you please trigger the test again?

parkera:
@swift-ci please test

parkera:

colemancda:
@parkera Any explanation why this was changed? I don't see any JIRA ticket for Swift evolution discussed related to this. Now we have to include a bunch of `typealias` for Swift 3.0 backwards compatibility. e.g https://github.com/OpenKitten/BSON/commit/9f3dfc8baf007eabdf2aaf76afdf95bb059345e2

ikesyo:
@colemancda See https://github.com/apple/swift-evolution/blob/master/proposals/0086-drop-foundation-ns.md#proposed-solution

> If the class is planned to have a value-type equivalent in the near future, then keep the `NS` prefix. Examples: `NSAttributedString`, `NSRegularExpression`, `NSPredicate`.

The class was named as `RegularExpression` mistakenly in swift-corelibs-foundation (Linux). So I think this is a bug fix.

parkera:
Yah, the darwin version is `NSRegularExpression` as of Swift 3.0 and our charter here is to match that.

PR #646, opened 2016-09-18
Fixes for associated types of Data

gribozavr:
- Set Foundation.Data.SubSequence to be RangeReplaceable

  RangeReplaceableCollection.SubSequence should be a RangeReplaceableCollection.  Data implements RangeReplaceableCollection, but sets SubSequence to MutableRandomAccessSlice that is not a RangeReplaceableCollection.

  This commit changes Foundation.Data.SubSequence to MutableRangeReplaceableRandomAccessSlice.

  Fixes rdar://problem/28330713.

- Change Data.Indices to be a CountableRange<Int>, that is much faster than the default.


gribozavr:
@swift-ci Please test

PR #647, opened 2016-09-18
Fix for a crash in String.substring(with:) [SR-2483]

pushkarnk:
I think this is just one way to fix: https://bugs.swift.org/browse/SR-2483 
I've documented some additional findings in the bug report.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -346,7 +346,49 @@ extension NSString {
             let start = _storage.utf16.startIndex
             let min = start.advanced(by: range.location)
             let max = start.advanced(by: range.location + range.length)
-            return String(_storage.utf16[min..<max])!
+            if let substr = String(_storage.utf16[min..<max]) {
+                return substr
+            }
+            //If we come here, then the range has created unpaired surrogates on either end.
+            //An unpaired surrogate is replaced by OXFFFD - the Unicode Replacement Character.
+            //The CRLF ("\r\n") sequence is also treated like a surrogate pair, but its constinuent
+            //characters "\r" and "\n" can exist outside the pair!
+
+            let replacementCharacter = String(describing: UnicodeScalar(0xFFFD)!)
+            let CR: UInt16 = 13  //carriage return
+            let LF: UInt16 = 10  //new line
+
+            //make sure the range is of non-zero length
+            guard range.length > 0 else { return "" }
+
+            //if the range is pointing to a single unpaired surrogate
+            if range.length == 1 {
+                switch _storage.utf16[min] {
+                case CR: return "\r"
+                case LF: return "\n"
+                default: return replacementCharacter
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


phausler:
could you possibly add a unit test to verify this behavior?

pushkarnk:
@phausler : I am no longer comfortable with this fix. The API reference says ```String.init?(_ utf16)``` returns nil only when the utf16 contains unpaired surrogates - so, the case of it returning nil for "\r\n"[1..<1] does not match the spec and we cannot simply return "". This needs more investigation on the stdlib part. If you don't mind, I'll keep this pull request open until we have a better understanding of how this should be fixed.

phausler:
Certainly, it might be useful to verify the ground truth of how that behaves in objc - since that is our baseline here

pushkarnk:
When using ```substring(with:)``` if either end of the range breaks a surrogate pair, we crash on Linux but not on Darwin. On Darwin, I found the broken surrogate pair being replaced by the[ Unicode replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character). We need to do the same on Linux. Also, for ranges of the type ```t..<t``` we need to return "".

pushkarnk:
@swift-ci please test

pushkarnk:
Updated this to include a handful of corner cases with surrogate pairs and `\r\n`

pushkarnk:
@swift-ci please test

Foundation/NSString.swift
```diff
@@ -346,7 +346,49 @@ extension NSString {
...
+
+            //if the range breaks a surrogate pair at the end of the string
+            if let substrPrefix = String(_storage.utf16[min..<max.advanced(by: -1)]) {
+                return substrPrefix + suffix
+            }
+
+            //the range probably breaks surrogate pairs at both the ends
+            guard min.advanced(by: 1) <= max.advanced(by: -1) else { return prefix + suffix }
+
+            let substr =  String(_storage.utf16[min.advanced(by: 1)..<max.advanced(by: -1)])!
```
pushkarnk: After all the prior checks, I think this can be safely unwrapped.

pushkarnk:
@phausler @parkera Sorry for the delay on this. 

I've tried to make `substring(with: NSRange)` more robust (and behave like Darwin) here, in the process of fixing SR-2483. Can you please review? Thanks.

parkera:
@swift-ci please test

pushkarnk:
@parkera @phausler is there anything else you'd want me to consider here?

parkera:
This took me a while to review because I was digging through the Darwin version, which really defers this operation to CFString. That's where the replacement character substitution happens. In the long term we may want to make NSString.swift do something similar so we don't end up replicating this logic in two places, but in the short term this fix seems appropriate.

parkera:
@swift-ci test and merge

PR #650, opened 2016-09-21
Support JSONSerialization of more number types

seabaylea:
Currently the JSON Serializer only handles the "common" numbers types: `Int`, `Double` and `NSNumber`.

Trying to serialise the other number types: `UInt`, `Int8`, `UInt8`, etc as well as `Float` results in an Error.

This PR adds support for the wider set of standard types.

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -273,11 +273,7 @@ private struct JSONWriter {
         
         if let str = obj as? String {
             try serializeString(str)
-        } else if let num = obj as? Int {
-            try serializeNumber(NSNumber(value: num))
-        } else if let num = obj as? Double {
-            try serializeNumber(NSNumber(value: num))
-        } else if let num = obj as? NSNumber {
+        } else if let num = _SwiftValue.store(obj) as? NSNumber {
             try serializeNumber(num)
         } else if let array = obj as? Array<Any> {
             try serializeArray(array)


phausler:
This is incompatible with the Darwin version:

```swift
let json = ["foo" : UInt8(4)]
let data = try! JSONSerialization.data(withJSONObject: json, options: [])

print("\(String(data: data, encoding: .utf8)!)")
```

fails with the exception:

```
2016-09-21 12:34:16.233614 JSONTest[50614:1325824] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid type in JSON write (_SwiftValue)'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff81f4442b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00007fff9689ca2a objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff81fc2c05 +[NSException raise:format:] + 197
	3   Foundation                          0x00007fff838e973a _writeJSONValue + 698
	4   Foundation                          0x00007fff83946636 ___writeJSONObject_block_invoke + 226
	5   CoreFoundation                      0x00007fff81eb863e __NSDictionaryEnumerate + 862
	6   Foundation                          0x00007fff8394642a _writeJSONObject + 430
	7   Foundation                          0x00007fff838e965f _writeJSONValue + 479
	8   Foundation                          0x00007fff838e942c -[_NSJSONWriter dataWithRootObject:options:error:] + 124
	9   Foundation                          0x00
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) 
```

parkera:
Why does Int8 bridge to _SwiftValue instead of NSNumber in the first place? I thought this was something they were looking into fixing.

phausler:
The changes for numeric bridging have not yet landed, but to replicate that behavior then we should funnel the top level input to `_SwiftValue.store()` and traverse references instead of structures.

seabaylea:
I've updated to use `_SwiftValue.store(obj)` for the various number types, which matches the behaviour on Darwin.

phausler:
@swift-ci please test

parkera:
@phausler you ok with this? should we merge now?

phausler:
Yea, this seems like a good first step to gaining more pairity

PR #651, opened 2016-09-21
Mark Operation's internal fields as public to avoid link errors when subclassing

seabaylea:
SR-1901 (and others) documents an issue where subclassing Foundation classes that have hidden (private or internal" field and functions causes link time errors.

This particularly affects the users of Operation, where its common to subclass to override start/complete etc. This in particular is what's being done for the Swift Cloudant package:
https://github.com/cloudant/swift-cloudant/blob/linux-support/Source/Operation.swift

In order to temporarily work around the issue, this PR marks the fields as public. Additionally, we may well need to permanently make `finish()` a public API as Swift can't use KVO to determine the end of async operations - that would give a mechanism for the application code to notify that the operation was complete.

### DIFF
--- a/Foundation/NSOperation.swift
+++ b/Foundation/NSOperation.swift
@@ -19,16 +19,16 @@ private func pthread_main_np() -> Int32 {
 
 open class Operation : NSObject {
     let lock = NSLock()
-    internal weak var _queue: OperationQueue?
-    internal var _cancelled = false
-    internal var _executing = false
-    internal var _finished = false
-    internal var _ready = false
-    internal var _dependencies = Set<Operation>()
+    public weak var _queue: OperationQueue?
+    public var _cancelled = false
+    public var _executing = false
+    public var _finished = false
+    public var _ready = false
+    public var _dependencies = Set<Operation>()
 #if DEPLOYMENT_ENABLE_LIBDISPATCH
-    internal var _group = DispatchGroup()
-    internal var _depGroup = DispatchGroup()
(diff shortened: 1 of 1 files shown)


modocache:
Looks good to me, and thanks for linking to [SR-1901](https://bugs.swift.org/browse/SR-1901).

In swift-corelibs-xctest, we place FIXMEs above symbols that are marked as `public` due to this bug, with explanations why they must be marked `public` for now. Maybe do the same here? It may be less confusing to other contributors that way.

phausler:
I don't think we really can do this since it would change vastly the surface area of the APIs.

parkera:
I'm following up with the compiler team on this to see if we can get some eyes on the problem ASAP.

parkera:
The underlying bug claims to be fixed. @seabaylea would you mind verifying that this PR is no longer needed?

parkera:
https://github.com/apple/swift/commit/fd61338444ac42d8715fd9e637b238e035f17f8d

seabaylea:
Sure - I'll give it a test.

seabaylea:
The underlying change in Swift has fixed the linker errors.

The only outstanding question is how we handle notifying the completion of async operations as Swift can't use KVO - i.e., whether we mark `finish()` as public to allow the operation to notify that it is complete.

parkera:
Yes, I think we'll have to come up with a temporary solution for that. I'll close this PR and we can raise a new one for that piece.

PR #652, opened 2016-09-21
String: fix potential unaligned access

compnerd:
A potential unaligned access was identified by UBSAN.  It is unclear if this is
a correctness issue, but may be a source for some inefficient code.


compnerd:
@swift-ci please test

CC: @parkera @phausler

phausler:
The only calling points are from __CFStringCreateImmutableFunnel3, which would mean that a caller of CFStringCreateWithCString would have to pass an un-aligned buffer somehow

compnerd:
@phausler this is motivated by the desire to avoid UBSAN warnings, and why I call out in the commit message that this isn't really a correctness issue.

compnerd:
@parkera @phausler ping?  This can also be a slight optimization since we will pre-process to get into a wider stride.

alblue:
This pull request has gone stale, closing. If required, please re-base onto master and reopen.

PR #653, opened 2016-09-22
Fix a missing test case in NSAttributedString

eyeplum:
Previously, the `test_longestEffectiveRange` was not added to `allTests`.
Also some of the asserts were miscalculated.


parkera:

PR #656, opened 2016-09-26
[NSFileManager] Cast file types as Int32

modocache:
On macOS, `DT_DIR` is imported as an `Int`. `Int` and `Int32` cannot be compared via `==`, so on macOS the comparison results in a build failure. Cast the variables to `Int32` to resolve.

---

This fixes a Foundation build failure on XCTest's macOS CI. Unblocks https://github.com/apple/swift-corelibs-xctest/pull/172 and https://github.com/apple/swift-corelibs-xctest/pull/174.


modocache:
@swift-ci please test

phausler:
looks good to me

parkera:
Thanks @modocache!

PR #658, opened 2016-09-27
Fix Any? -> Any build warnings

e78l:
Addressing build-time warnings of possibly nil values being used as Any [SR-2737] (https://bugs.swift.org/browse/SR-2737)
Some whitespace fixes too

### DIFF
--- a/Foundation/NSHTTPCookie.swift
+++ b/Foundation/NSHTTPCookie.swift
@@ -37,39 +37,42 @@ extension HTTPCookiePropertyKey {
 
     /// Key for cookie value
     public static let value = HTTPCookiePropertyKey(rawValue: "Value")
-    
+
     /// Key for cookie origin URL
     public static let originURL = HTTPCookiePropertyKey(rawValue: "OriginURL")
 
     /// Key for cookie version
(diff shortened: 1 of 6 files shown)


e78l:
Realized I should include changes to NSHTTPCookie too; I had to change the time format to 'z' in my environment (Amazon Linux, ICU 57.1).
"O" doesn't work to parse "GMT" some reason...
The problem was caught by existing test. 'z' works OK and on my Mac too.
Heads up to @dbrajkovic and @pushkarnk in case something breaks but http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table says 'O' is a fallback for 'z', so I don't expect any problems.

parkera:
We had at least one situation with Any/Any? where it would warn either on Darwin or on Linux. @phausler should double check that.

Also, I would prefer we split up the HTTPCookie format changes into a separate PR.

e78l:
Updated - split off the HTTPCookie format change; HTTPCookie changes to silence warnings about optional values in [String:Any] remain

e78l:
@phausler @parkera Chance for a review? I think this still applies though there've been more warnings introduced since (re: optional string interpolation).

phausler:

PR #659, opened 2016-09-27
Implement Decimal

alblue:
Although not yet fully covered by tests, this is a nearly complete implementation of NSDecimal. There are cases where dealing with large and small numbers will overflow that isn't implemented yet (there is one remaining NSUnimplemented) but it can be used for simple mathematical operations.

### DIFF
--- a/Foundation/NSDecimal.swift
+++ b/Foundation/NSDecimal.swift
@@ -15,6 +15,7 @@ public var NSDecimalMaxSize: Int32 { return 8 }
 public var NSDecimalNoScale: Int32 { return Int32(Int16.max) }
 
 public struct Decimal {
+    fileprivate static let maxSize: UInt32 = UInt32(NSDecimalMaxSize)
     fileprivate var __exponent: Int8
     fileprivate var __lengthAndFlags: UInt8
     fileprivate var __reserved: UInt16
@@ -74,6 +75,14 @@ public struct Decimal {
         self.__exponent = 0
         self.__lengthAndFlags = 0
         self.__reserved = 0
+    }
+
+    fileprivate init(length: UInt32, mantissa: (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)) {
+        self._mantissa = mantissa
+        self.__exponent = 0
+        self.__lengthAndFlags = 0
+        self.__reserved = 0
+        self._length = length
     }
 
     public init(_exponent: Int32, _length: UInt32, _isNegative: UInt32, _isCompact: UInt32, _reserved: UInt32, _mantissa: (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)){
@@ -207,12 +216,12 @@ extension Decimal {
     }
     public static func +(lhs: Decimal, rhs: Decimal) -> Decimal {
         var answer = lhs
-        answer.add(lhs)
+        answer.add(rhs)
         return answer;
     }
     public static func -(lhs: Decimal, rhs: Decimal) -> Decimal {
         var answer = lhs
-        answer.subtract(lhs)
+        answer.subtract(rhs)
         return answer;
     }
     public static func /(lhs: Decimal, rhs: Decimal) -> Decimal {
@@ -381,28 +390,28 @@ extension Decimal {
             while mantissa != 0 && i < NSDecimalMaxSize {
                 switch i {
                 case 0:
-                    _mantissa.0 = UInt16(mantissa & 0xffff)
+                    _mantissa.0 = UInt16(truncatingBitPattern:mantissa)
                     break
                 case 1:
-                    _mantissa.1 = UInt16(mantissa & 0xffff)
+                    _mantissa.1 = UInt16(truncatingBitPattern:mantissa)
                     break
                 case 2:
-                    _mantissa.2 = UInt16(mantissa & 0xffff)
+                    _mantissa.2 = UInt16(truncatingBitPattern:mantissa)
                     break
                 case 3:
-                    _mantissa.3 = UInt16(mantissa & 0xffff)
+                    _mantissa.3 = UInt16(truncatingBitPattern:mantissa)
                     break
                 case 4:
-                    _mantissa.4 = UInt16(mantissa & 0xffff)
+                    _mantissa.4 = UInt16(truncatingBitPattern:mantissa)
                     break
                 case 5:
-                    _mantissa.5 = UInt16(mantissa & 0xffff)
+                    _mantissa.5 = UInt16(truncatingBitPattern:mantissa)
                     break
                 case 6:
-                    _mantissa.6 = UInt16(mantissa & 0xffff)
+                    _mantissa.6 = UInt16(truncatingBitPattern:mantissa)
                     break
                 case 7:
-                    _mantissa.7 = UInt16(mantissa & 0xffff)
+                    _mantissa.7 = UInt16(truncatingBitPattern:mantissa)
                     break
                 default:
                     fatalError("initialization overflow")
@@ -543,59 +552,95 @@ extension Decimal : CustomStringConvertible {
     }
 }
 
-fileprivate func divideByShort(_ d: UnsafeMutablePointer<Decimal>, _ divisor:UInt16) -> (UInt16,NSDecimalNumber.CalculationError) {
+fileprivate func divideByShort(_ d: inout BiggerDecimal, _ divisor:UInt16) -> (UInt16,NSDecimalNumber.CalculationError) {
     if divisor == 0 {
-        d.pointee._length = 0
+        d._length = 0
         return (0,.divideByZero)
     }
     // note the below is not the same as from length to 0 by -1
     var carry: UInt32 = 0
-    for i in stride(from: 0, to: d.pointee._length, by: 1).reversed() {
-        let accumulator = UInt32(d.pointee[i]) + carry * (1<<16)
-        d.pointee[i] = UInt16(accumulator / UInt32(divisor))
+    for i in stride(from: 0, to: d._length, by: 1).reversed() {
+        let accumulator = UInt32(d[i]) + carry * (1<<16)
+        d[i] = UInt16(accumulator / UInt32(divisor))
         carry = accumulator % UInt32(divisor)
     }
-    while d.pointee._length != 0 && d.pointee[d.pointee._length - 1] == 0 {
-        d.pointee._length -= 1
+    while d._length != 0 && d[d._length - 1] == 0 {
+        d._length -= 1
     }
     return (UInt16(carry),.noError)
 }
 
-fileprivate func multiplyByShort(_ d: UnsafeMutablePointer<Decimal>, _ mul:UInt16) -> NSDecimalNumber.CalculationError {
+
+fileprivate func divideByShort(_ d: inout Decimal, _ divisor:UInt16) -> (UInt16,NSDecimalNumber.CalculationError) {
+    if divisor == 0 {
+        d._length = 0
+        return (0,.divideByZero)
+    }
+    // note the below is not the same as from length to 0 by -1
+    var carry: UInt32 = 0
+    for i in stride(from: 0, to: d._length, by: 1).reversed() {
+        let accumulator = UInt32(d[i]) + carry * (1<<16)
+        d[i] = UInt16(accumulator / UInt32(divisor))
+        carry = accumulator % UInt32(divisor)
+    }
+    while d._length != 0 && d[d._length - 1] == 0 {
+        d._length -= 1
+    }
+    return (UInt16(carry),.noError)
+}
+
+fileprivate func multiplyByShort(_ d: inout Decimal, _ mul:UInt16) -> NSDecimalNumber.CalculationError {
     if mul == 0 {
-        d.pointee._length = 0
+        d._length = 0
         return .noError
     }
     var carry: UInt32 = 0
     // FIXME handle NSCalculationOverflow here?
-    for i in 0..<d.pointee._length {
-        let accumulator: UInt32 = UInt32(d.pointee[i]) * UInt32(mul) + carry
+    for i in 0..<d._length {
+        let accumulator: UInt32 = UInt32(d[i]) * UInt32(mul) + carry
         carry = accumulator >> 16
-        d.pointee[i] = UInt16(truncatingBitPattern: accumulator)
+        d[i] = UInt16(truncatingBitPattern: accumulator)
     }
     if carry != 0 {
-        if Int32(d.pointee._length) == NSDecimalMaxSize {
+        if d._length == Decimal.maxSize {
             return .overflow
         }
-        d.pointee[d.pointee._length] = UInt16(truncatingBitPattern: carry)
-        d.pointee._length += 1
+        d[d._length] = UInt16(truncatingBitPattern: carry)
+        d._length += 1
     }
     return .noError
 }
 
-fileprivate func addShort(_ d: UnsafeMutablePointer<Decimal>, _ add:UInt16) -> NSDecimalNumber.CalculationError {
+fileprivate func addShort(_ d: inout BiggerDecimal, _ add:UInt16) -> NSDecimalNumber.CalculationError {
     var carry:UInt32 = UInt32(add)
-    for i in 0..<d.pointee._length {
-        let accumulator: UInt32 = UInt32(d.pointee[i]) + carry
+    for i in 0..<d._length {
+        let accumulator: UInt32 = UInt32(d[i]) + carry
         carry = accumulator >> 16
-        d.pointee[i] = UInt16(truncatingBitPattern: accumulator)
+        d[i] = UInt16(truncatingBitPattern: accumulator)
     }
     if carry != 0 {
-        if Int32(d.pointee._length) == NSDecimalMaxSize {
+        if d._length == Decimal.maxSize {
             return .overflow
         }
-        d.pointee[d.pointee._length] = UInt16(truncatingBitPattern: carry)
-        d.pointee._length += 1
+        d[d._length] = UInt16(truncatingBitPattern: carry)
+        d._length += 1
+    }
+    return .noError
+}
+
+fileprivate func addShort(_ d: inout Decimal, _ add:UInt16) -> NSDecimalNumber.CalculationError {
+    var carry:UInt32 = UInt32(add)
+    for i in 0..<d._length {
+        let accumulator: UInt32 = UInt32(d[i]) + carry
+        carry = accumulator >> 16
+        d[i] = UInt16(truncatingBitPattern: accumulator)
+    }
+    if carry != 0 {
+        if d._length == Decimal.maxSize {
+            return .overflow
+        }
+        d[d._length] = UInt16(truncatingBitPattern: carry)
+        d._length += 1
     }
     return .noError
 }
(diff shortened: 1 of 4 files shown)


Foundation/NSDecimal.swift
```diff
@@ -1,26 +1,89 @@
...
+    public init() {
+        self._mantissa = (0,0,0,0,0,0,0,0)
+        self.__exponent = 0
+        self.__lengthAndFlags = 0
+        self.__reserved = 0
+    }
+
+    public init(_exponent: Int32, _length: UInt32, _isNegative: UInt32, _isCompact: UInt32, _reserved: UInt32, _mantissa: (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)){
+        self._mantissa = _mantissa
+        self.__exponent = Int8(_exponent & Int32(0xff))
```
alblue: There is a bug with this, in that negative exponents don't work as expected. It doesn't dial the tests but there needs some additional work before this goes through, as well as more tests.

alblue:
It looks like Int8(truncatingBitPattern:) is required here

Foundation/NSDecimal.swift
```diff
@@ -1,42 +1,159 @@
...
+    }
+    public var significand: Decimal {
+        get {
+            return Decimal(_exponent: 0, _length: _length, _isNegative: _isNegative, _isCompact: _isCompact, _reserved: 0, _mantissa: _mantissa)
+        }
+    }
+    public init(sign: FloatingPointSign, exponent: Int, significand: Decimal) {
+        self.init(_exponent: Int32(exponent) + significand._exponent, _length: significand._length, _isNegative: sign == .plus ? 0 : 1, _isCompact: significand._isCompact, _reserved: 0, _mantissa: significand._mantissa)
+    }
+    public init(signOf: Decimal, magnitudeOf magnitude: Decimal) {         self.init(_exponent: magnitude._exponent, _length: magnitude._length, _isNegative: signOf._isNegative, _isCompact: magnitude._isCompact, _reserved: 0, _mantissa: magnitude._mantissa)
```
xwu: Nit: missing newline before `self.init`

Foundation/NSDecimal.swift
```diff
@@ -56,74 +186,285 @@ extension Decimal {
...
+    public var description: String {
+        var val = self
+        return NSDecimalString(&val, nil)
     }
-    public var description: String { NSUnimplemented() }
 }
 
-public func NSDecimalIsNotANumber(_ dcm: UnsafePointer<Decimal>) -> Bool { NSUnimplemented() }
-
+public func NSDecimalIsNotANumber(_ dcm: UnsafePointer<Decimal>) -> Bool { return dcm.pointee.isNaN
```
xwu: Nit: missing new line before `return`

Foundation/NSDecimal.swift
```diff
@@ -153,3 +494,30 @@ public func NSDecimalPower(_ result: UnsafeMutablePointer<Decimal>, _ number: Un
...
+    fileprivate var isCompact: Bool {
+        get {
+            return _isCompact != 0
+        }
+        set {
+            if newValue {
+                _isCompact = 1
+            } else {
+                _isCompact = 0
+            }
```
xwu: Nit: could this be more terse? `_isCompact = newValue ? 1 : 0`

Foundation/NSDecimal.swift
```diff
@@ -153,3 +494,30 @@ public func NSDecimalPower(_ result: UnsafeMutablePointer<Decimal>, _ number: Un
...
+    }
+    fileprivate mutating func compact() {
+        if isCompact || isNaN || _length == 0 {
+            return
+        }
+        // FIXME implement the multiply/divide
+        // Divide by 10 as much as possible
+        // Put the non-empty remainder in place
+        // Set the new exponent
+        isCompact = true
```
xwu: I'm confused by this stub. It doesn't do anything, so why not leave `NSDecimalCompact(_:)` unimplemented?

Foundation/NSDecimal.swift
alblue: Can you explain what you mean? Self.init is on line 143 and the public init is on line 142.

Foundation/NSDecimal.swift
alblue: Fair enough, can fix

Foundation/NSDecimal.swift
alblue: Yes, it could be but I don't know if there's aversion to using a ternary operator.

Foundation/NSDecimal.swift
alblue: Mainly because the NSDecimalCompact call is invoked from the init(Double) constructor, without which you couldn't test this. However that part of it is still a work in progress, so this could be left out of this patch set for now.

Foundation/NSDecimal.swift
alblue: Ah, sorry - the comment was on line 145.

Foundation/NSDecimal.swift
alblue: Fixed in subsequent push

Foundation/NSDecimal.swift
```diff
@@ -1,42 +1,160 @@
...
+        get {
+            return UInt32(UInt32(__lengthAndFlags & 0b1100_0000) << 10 | UInt32(__reserved))
+        }
+        set {
+            __lengthAndFlags =
+                (__lengthAndFlags & 0b0011_1111) |
+                UInt8(UInt32(newValue & (0b11 << 16)) >> 10)
+            __reserved = UInt16(newValue & 0b1111_1111_1111_1111)
+        }
+    }
```
phausler: these accessors change the structural layout and size of Decimal, it will probably make it incompatible when archiving and serializing it. Is there a compelling reason to re-layout the structure? To port addition and division and compaction it might be nice to be able to send these over to C functions.

Foundation/NSDecimal.swift
```diff
@@ -56,74 +187,286 @@ extension Decimal {
...
+            }
+        }
+        return _isNegative != 0 ? -d : d
+    }
+    public var hashValue: Int {
+        return Int(bitPattern: __CFHashDouble(doubleValue))
+    }
+    public static func ==(lhs: Decimal, rhs: Decimal) -> Bool {
+        if lhs.isNaN || rhs.isNaN {
+            return false
```
phausler: To match the behavior of the objective-c version:
```objc
    // NaN is a special case and is arbitrary ordered before everything else
    // Conceptually comparing with NaN is bogus anyway but raising or 
    // always returning the same answer will confuse the sorting algorithms
   if(NSDecimalIsNotANumber(leftOperand)) {
        if(NSDecimalIsNotANumber(rightOperand)) {
            return NSOrderedSame;
        }
        return NSOrderedAscending;
    }
```

so if both are NaN then they are ordered same which is considered equal (irrespective of any other behavior standard of floating point numbers)

Foundation/NSDecimal.swift
```diff
@@ -56,74 +187,286 @@ extension Decimal {
...
+        }
+        if lhs.__exponent == rhs.__exponent && lhs.__lengthAndFlags == rhs.__lengthAndFlags && lhs.__reserved == rhs.__reserved {
```
phausler: negative values fall into here?

Foundation/NSDecimal.swift
```diff
@@ -56,74 +187,286 @@ extension Decimal {
...
+                lhs._mantissa.4 == rhs._mantissa.4 &&
+                lhs._mantissa.5 == rhs._mantissa.5 &&
+                lhs._mantissa.6 == rhs._mantissa.6 &&
+                lhs._mantissa.7 == rhs._mantissa.7 {
+                return true
+            }
+        }
+        var lhsVal = lhs
+        var rhsVal = rhs
+        return NSDecimalCompare(&lhsVal, &rhsVal) == .orderedSame
```
phausler: The Darwin version just calls this.

Foundation/NSDecimal.swift
```diff
@@ -56,74 +187,286 @@ extension Decimal {
...
+	return dcm.pointee.isNaN
+}
 
 /***************	Operations		***********/
-public func NSDecimalCopy(_ destination: UnsafeMutablePointer<Decimal>, _ source: UnsafePointer<Decimal>) { NSUnimplemented() }
+public func NSDecimalCopy(_ destination: UnsafeMutablePointer<Decimal>, _ source: UnsafePointer<Decimal>) {
+    destination.pointee.__lengthAndFlags = source.pointee.__lengthAndFlags
+    destination.pointee.__exponent = source.pointee.__exponent
+    destination.pointee.__reserved = source.pointee.__reserved
+    destination.pointee._mantissa = source.pointee._mantissa
```
phausler: The objective-c version does not preserve reserved (but perhaps it should)

Foundation/NSDecimal.swift
```diff
@@ -56,74 +187,286 @@ extension Decimal {
...
+}
 
-public func NSDecimalCompact(_ number: UnsafeMutablePointer<Decimal>) { NSUnimplemented() }
+public func NSDecimalCompact(_ number: UnsafeMutablePointer<Decimal>) {
+    number.pointee.compact()
```
phausler: This is upside down from the objective-c implementation, for consistency sake we might want to put the workhorse logic here instead of member functions (and perhaps treat the NSDecimal* functions as "implementation details"

Foundation/NSDecimal.swift
alblue: OK, can fix that

Foundation/NSDecimal.swift
alblue: These bring the structure to be the same format as the Objective-C structure. At the moment both Linux and Objective-C are different. Secondly, there is a difference in behaviour visible, in that the results are truncated on Darwin but not on Linux (hence the tests for isCompact:3). That's because the structure on the Objective-C is packed. Note that I have run these tests on both Linux and Objective-C to verify the same behaviour. Trivially, on Darwin at the moment:

  1> import Foundation
  2> MemoryLayout<Decimal>.size
$R0: Int = 20

and on Linux:

  1> import Foundation
  2> MemoryLayout<Decimal>.size
$R0: Int = 36

I would argue that to be performant we would want to have these as the same data structure.

Foundation/NSDecimal.swift
alblue: The 'lengthAndFlags' field contains both the length (upper nibble) and the compact/negative flags. So this will test for negative as well. We could explicitly test the flags individually but then we're doing a lot of byte wrapping for no real reason. The 'lengthAndFlags' looks like:

llllxxxx
xxxxCxxx
xxxxxNxx

Foundation/NSDecimal.swift
alblue: Sorry, I don't understand the comment on its own. Is it related to other comments or code paths?

Foundation/NSDecimal.swift
alblue: I'm happy to just leave setting reserved to a no-op and treating them as zero if you want me to change the Swift version

Foundation/NSDecimal.swift
alblue: Using the UnsafeMutablePointer is actually pretty painful. You can't (for example) have 'var d = number.pointee; d.doStuff()' because it ends up with a different copy. So everywhere you have the functions you need to do number.pointee to use the passed in version. In addition, it's not being safe for the sake of it, but rather adopting an older API.

Finally you can't tell what is being mutated and what isn't with these APIs; you're just passing an opaque pointer. This is tedious when doing non-mutating code (e.g. the NSDecimalIsNotANumber) because you have can't pass in a 'let' constant:

  1> import Foundation
  2> let one = Decimal(1)
one: Decimal = {
  _mantissa = {
    0 = 8448
    1 = 0
    2 = 1
    3 = 0
    4 = 0
    5 = 0
    6 = 0
    7 = 0
  }
}
  3> NSDecimalIsNotANumber(one)
error: repl.swift:3:23: error: cannot convert value of type 'Decimal' to expected argument type 'UnsafePointer<Decimal>'
NSDecimalIsNotANumber(one)
                      ^~~


  3> NSDecimalIsNotANumber(&one)
error: repl.swift:3:23: error: cannot pass immutable value as inout argument: 'one' is a 'let' constant
NSDecimalIsNotANumber(&one)
                      ^~~~

Having those as the API is probably a bad thing to do; in fact, I'd argue that the NSDecimalXxxx functions should be marked as deprecated. Right now the simple (non-mutating) ones can be more efficiently implemented (read: not taking an unsafe pointer) by doing it the reverse way from the way that Objective-C did it, and having the functions as backwards compatible for those that need it.

Foundation/NSDecimal.swift
phausler: Hmm i guess the importer is claiming them to be different sizes than the actual bit-packed size. So i guess your solution is probably the best we can do for now sadly.

Foundation/NSDecimal.swift
phausler: got it; perhaps worth either a comment or a ivar name change to reflect that it it stored there

Foundation/NSDecimal.swift
phausler: The only part that might matter is for verification of behavior. To API consumers it ought to not matter.

Foundation/NSDecimal.swift
phausler: We can't mark them as deprecated and still use them to power the extended methods on NSDecimal (perhaps we could rename them so that they are not publicly autocompleted; but that is not going to happen likely for Swift 3.X).

Often when making the overlay I had to use a temp copy to transact upon instead. And we can't make them hoisted since they transact upon an address of a Decimal not the Decimal itself.

Foundation/NSDecimal.swift
alblue: Any suggestions as to what to call it? __lengthAndNegativeAndCompact?

Foundation/NSDecimal.swift
alblue: By the way there's some inconsistency going on here because of this. On Darwin:

  1> import Foundation
  2> let a = Decimal.nan
  3> let b = Decimal.nan
  4> a.isEqual(to:b)
$R0: Bool = true
  5> a.isTotallyOrdered(belowOrEqualTo: b)
$R1: Bool = false

You also have the fact that:

  6> Double.nan == Double.nan
$R2: Bool = false

which makes it doubly confusing, though slightly off-topic for here.

I'm happy to implement/make this into bug compatible with Darwin but arguably the fact that you have a different behaviour between 'isEqual' and 'isTotallyOrdered' is a little strange.

alblue:
This isn't yet complete (test coverage and normalisation NSUnimplemented still to go) but is in a state where the content can be seriously reviewed ahead of merging.

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1837 @@
...
+            error = integerMultiply(&big,result,p10)
+        } else {
+            error = integerDivide(&big,result,p10)
+        }
+
+        if error != .noError && error != .lossOfPrecision {
+            return error;
+        }
+
+        for i in 0..<Decimal.maxSize {
```
alblue: This should probably be big._length rather than Decimal.maxSize, otherwise we'll lose precision

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1837 @@
...
+        aa = a
+        bb = b
+    }
+
+    //
+    // Build a backup for aa
+    //
+    var backup = Decimal()
+
+    NSDecimalCopy(&backup,aa)
```
alblue: NB this seems redundant at the moment but will be used in the NSUnimplemented below

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1837 @@
...
+
+                // Just to be sure!
+                if newExponent > Int32(Int8.max) {
+                    result.pointee.setNaN()
+                    return .overflow
+                }
+                result.pointee._exponent = newExponent
+            }
+            result.pointee._length = big._length
+            for i in 0..<Decimal.maxSize {
```
alblue: This should be big._length as well, I think

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1837 @@
...
+
+    var newExponent = leftOperand.pointee._exponent + rightOperand.pointee._exponent
+
+    if big._length > Decimal.maxSize {
+        var exponent:Int32 = 0
+        calculationError = fitMantissa(&big, &exponent, roundingMode)
+        newExponent += exponent
+    }
+
+    for i in 0..<Decimal.maxSize { // bigSize?
```
alblue: Think this should be big._length as well

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1837 @@
...
+    return dcm.pointee.description
+}
+
+fileprivate protocol VariableLengthNumber {
+    var _length: UInt32 { get set }
+    subscript(index:UInt32) -> UInt16 { get set }
+    var isZero:Bool { get }
+}
+
+extension Decimal: VariableLengthNumber {
```
alblue: Not sure if we can apply fileprivate here but we should if we could

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1837 @@
...
+        result._length = self._length
+        for i in 0..<_length {
+            result[i] = self[i]
+        }
+        return result
+    }
+}
 
-public func NSDecimalMultiplyByPowerOf10(_ result: UnsafeMutablePointer<Decimal>, _ number: UnsafePointer<Decimal>, _ power: Int16, _ roundingMode: NSDecimalNumber.RoundingMode) -> NSDecimalNumber.CalculationError { NSUnimplemented() }
+// == Internal (Swifty) functions ==
```
alblue: It would be good to make these public API and have the same on the Darwin SDK - thoughts?

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1838 @@
...
+            i -= 1
+            buffer[i] = MINUS
+        }
+        return String(cString: Array(buffer.suffix(from:i)))
     }
-    public var description: String { NSUnimplemented() }
 }
 
-public func NSDecimalIsNotANumber(_ dcm: UnsafePointer<Decimal>) -> Bool { NSUnimplemented() }
+fileprivate func divideByShort<T:VariableLengthNumber>(_ d: inout T, _ divisor:UInt16) -> (UInt16,NSDecimalNumber.CalculationError) {
```
xwu: Divide and carry, multiply and carry, etc, are best off using the new double-width divide and multiply facilities now available with new integer protocols. Those were exposed precisely for implementation of types such as Decimal.

xwu: Shouldn't these be internal methods on Decimal, not free functions?

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1838 @@
...
-public func NSDecimalCompare(_ leftOperand: UnsafePointer<Decimal>, _ rightOperand: UnsafePointer<Decimal>) -> ComparisonResult { NSUnimplemented() }
 // NSDecimalCompare:Compares leftOperand and rightOperand.
+public func NSDecimalCompare(_ leftOperand: UnsafePointer<Decimal>, _ rightOperand: UnsafePointer<Decimal>) -> ComparisonResult {
+    let left = leftOperand.pointee
+    let right = rightOperand.pointee
+    return left.compare(to: right)
+}
+
+fileprivate extension UInt16 {
+    func compareTo(_ other: UInt16) -> ComparisonResult {
```
xwu: Why do you need this?

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1838 @@
...
+    // D2: initialize j
+    // On each pass, build a single value for the quotient.
+    for j in 0..<ql {
+        var q: UInt32 // Quotient digit. could be a short.
+
+        // D3: calculate q^
+        // This formula and test for q gives at most q+1; See Knuth for proof.
+
+        let ul = u._length
+        let uu = u // work around compiler bug
```
xwu: What compiler bug is this?

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1838 @@
...
+     * than that of the right operand (for example,
+     * 20 and 0.112314123094856724234234572), then the
+     * difference in their exponents is large and a lot of
+     * precision will be lost below. This is particularly
+     * true as the difference approaches 38 or larger.
+     * Normalizing here looses some precision on the
+     * individual operands, but often produces a more
+     * accurate result later. I chose 19 arbitrarily
+     * as half of the magic 38, so that normalization
+     * doesn't always occur. */
```
xwu: This seems incorrect. What is the behavior of Darwin Foundation?

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1838 @@
...
-public func NSDecimalAdd(_ result: UnsafeMutablePointer<Decimal>, _ leftOperand: UnsafePointer<Decimal>, _ rightOperand: UnsafePointer<Decimal>, _ roundingMode: NSDecimalNumber.RoundingMode) -> NSDecimalNumber.CalculationError { NSUnimplemented() }
 // Exact operations. result may be a pointer to same space as leftOperand or rightOperand
 
-public func NSDecimalSubtract(_ result: UnsafeMutablePointer<Decimal>, _ leftOperand: UnsafePointer<Decimal>, _ rightOperand: UnsafePointer<Decimal>, _ roundingMode: NSDecimalNumber.RoundingMode) -> NSDecimalNumber.CalculationError { NSUnimplemented() }
+public func NSDecimalSubtract(_ result: UnsafeMutablePointer<Decimal>, _ leftOperand: UnsafePointer<Decimal>, _ rightOperand: UnsafePointer<Decimal>, _ roundingMode: NSDecimalNumber.RoundingMode) -> NSDecimalNumber.CalculationError {
+    var r = rightOperand.pointee
+    if r._length != 0 {
+        r.negate()
+    }
+    return NSDecimalAdd(result, leftOperand, &r, roundingMode)
```
xwu: If you negate the operand, then the rounding mode may change.

alblue: The implementation is based on Foundation

Foundation/NSDecimal.swift
alblue: To make it easier to implement the code

Foundation/NSDecimal.swift
alblue: These were based on the implementation from Core Foundation for ease of review and acceptance. I am fine with the implementations being modified subsequently to use alternative methods of calculation. I would caution, though, that the NSDecimal* top level functions are public API and are a real nightmare to implement as sane methods on Decimal, so it is not trivial to do as such.

Foundation/NSDecimal.swift
xwu: It looks like it's used only once and can be trivially inlined. Doing so would have performance benefit and improve clarity at the use site.

Foundation/NSDecimal.swift
phausler: might be good to attach bugs.swift.org url here for a filed ticket just so it is known what bug it is

Foundation/NSDecimal.swift
phausler: When we added functionality to Decimal to have methods it became quite clear that it is nearly impossible to mimic the behavior and avoid needing to call out to the freestanding functions. They will most likely need to remain API in swift to service the hoisted method implementations. The divideByShort and fellow methods are relatively complex and hold some quirky behavior that these seem to attempt to mimic - we should consider porting some of the edge case tests from Foundation to swift-corelibs

Foundation/NSDecimal.swift
```diff
@@ -1,155 +1,1838 @@
...
+        aa.pointee._length = result._length
+        for i in 0..<result._length {
+            aa.pointee[i] = result[i]
+        }
+        aa.pointee._isCompact = 0
+        aa.pointee._exponent = bb.pointee._exponent
+        return .noError;
+    }
+
+    NSUnimplemented() // work in progress
```
phausler: might be good to comment on the remaining work here that needs to be done as a reminder ;)

Foundation/NSDecimal.swift
phausler: That would have to be something that we would need to pass through API review and decide on if they have the merit to maintain etc. Can they be re-written to use the free floating functions? it seems like there is some duplication of effort here. 

Ideally we should have these to be sharable code from swift-corelibs-foundation to the Darwin overlay for Foundation.

phausler:

PR #660, opened 2016-09-28
Implemented copyItem method

amraboelela:


pushkarnk:
Would it be possible to have a test for this method?

parkera:
I agree with @pushkarnk on getting a test for new code.

parkera:
Closing in favor of #847

PR #662, opened 2016-09-29
[SR-2798] Fix deadlock in URLSession.

aleksgapp:
The problem is in the current implementation of `finishTasksAndInvalidate`, if there're executing tasks in session - it will create a semaphore, that will block session work queue, but to release, it has to enter that queue again.

Resolves [SR-2798](https://bugs.swift.org/browse/SR-2798).

### DIFF
--- a/Foundation/NSURLSession/NSURLSession.swift
+++ b/Foundation/NSURLSession/NSURLSession.swift
@@ -265,18 +265,20 @@ open class URLSession : NSObject {
            //don't allow creation of new tasks from this point onwards
            self.invalidated = true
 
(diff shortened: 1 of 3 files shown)


pushkarnk:
Oh, I see the problem. ```taskRegistry.remove()``` was called on the same queue. Thanks.

pushkarnk:
@swift-ci please test

parkera:
LGTM, @pushkarnk go ahead and merge when you're happy.

pushkarnk:
I agree with the fix.

PR #663, opened 2016-09-29
Don't use #import when #include will do

kastiglione:
This changes a few `#import` statements to `#include`. This is because Clang's MSVC compatibility does not handle `#import` statements outside of of objc.

    #import of type library is an unsupported Microsoft feature


modocache:
@swift-ci please smoke test

modocache:
@swift-ci Please test

modocache:
@swift-ci please test

PR #664, opened 2016-09-30
Allow ; appear in path

frogcjn:
Due to https://openradar.appspot.com/24434533
E.g.
input string:
```swift
URLComponents(string: "http://example.com/api/group;id=2/list;id=1/detail")
```
Actual Results:
```console
http://example.com/api/group%3Bid=2/list%3Bid=1/detail
```

Expected Results:
```console
http://example.com/api/group;id=2/list;id=1/detail
```

### DIFF
--- a/CoreFoundation/URL.subproj/CFURLComponents_URIParser.c
+++ b/CoreFoundation/URL.subproj/CFURLComponents_URIParser.c
@@ -28,11 +28,11 @@ typedef CF_ENUM(CFIndex, URLPredefinedCharacterSet) {
     kURLAllowedCharacterSetIllegal  = 6
 };
 
-// IMPORTANT: the kURLxxxxAllowedCharacters definitions MUST match sURLAllowedCharacters (except for the '[', ':' and ']' characters in kURLHostAllowedCharacters are not kURLHostAllowed, and ';' is not in kURLPathAllowedCharacters, but is special cased in kURLPathAllowed)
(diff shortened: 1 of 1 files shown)


parkera:
This is from a response that bug:

"The param component (described in RFC1808 and RFC2396) was marked as obsolete (deprecated) in the URL spec (RFC3986/STD66). See appendix D.2. of that document — it’s listed as a obsolete rule. NSURLComponents percent-encodes the semicolon (’;’) character on purpose older parsers don’t confuse semicolon characters in paths. Since the param component is obsolete, you should not use it."

frogcjn:
@parkera 
RFC1808 URL:

https://johnny:p4ssw0rd@www.example.com:443/script.ext;param=value?query=value#ref

Results:

https://johnny:p4ssw0rd@www.example.com:443/script.ext%3Bparam=value?query=value#ref

frogcjn:
RFC3986/STD66 allows semicolon in path.

http://example.com/api/group;id=2/list;id=1/detail

frogcjn:
Side evidence:
HTTP: The Definitive Guide: Page 29

parkera:
The issue here is that the compatibility API in question is NSURL/URL, and we want to maintain compatibility between it and NSURLComponents/URLComponents. Therefore this code chooses to percent encode the `;`.

If/when we change the behavior of NSURL/URL itself (which we have to do carefully, as many apps out there rely on its nuances), then we can change this one as well. This code is shared on all of our platforms so we can't forget about the consequences for darwin API as well.

I think the best place for this patch right now is in a JIRA, so we can track updating these together in the future.

frogcjn:
Mozilla discussion: https://bugzilla.mozilla.org/show_bug.cgi?id=665706

frogcjn:
This is serious bug, YOU SHOULD FIX IT.

PR #665, opened 2016-09-30
Remove XML validation from NSArray/NSDictionary tests

alblue:
Calling XML validation results in hitting the Apple web servers
to load a DTD for plists http://www.apple.com/DTDs/PropertyList-1.0.dtd
to correctly validate the XML file.

Given that this test is not about whether it is possible to read
or write property lists, but they're just being used to validate
that they can read and write arrays from them, remove this part
of the test. The property list gets parsed from the subsequent
PropertyListSerialization call anyway, which will do the same error
checking as if the document is valid (but probably with a hard-coded
DTD instead of a network accessible one)

Issue: SR-2793


parkera:

PR #666, opened 2016-10-03
Formatted Very long horizontal comment to multiline.

kandelvijaya:
I believe this makes comments more easier to read rather than scrolling all horizontally just for one line.


parkera:
I'm ok with splitting up lines if there is a natural break in the English there. I don't want to enforce some kind of column limit on lines, though. I would rather leave it to people's editors to reflow as they choose.

parkera:
Going to go ahead and close this one for now.

kandelvijaya:
👍

PR #667, opened 2016-10-04
Add support for ISO8601 calendar and others

alblue:
The list of valid keys did not include the ISO8601 calendar, which
meant that code running on Linux that used it would fail. Fix this
by adding the kCFISO8601Calendar and others into the list of known
constants valid for calendars, and add tests for the Gregorian
calendar, as well as for creating all other calendars.

Issue: SR-2551

### DIFF
--- a/CoreFoundation/Locale.subproj/CFCalendar.c
+++ b/CoreFoundation/Locale.subproj/CFCalendar.c
@@ -236,14 +236,23 @@ CFCalendarRef CFCalendarCopyCurrent(void) {
 }
 
 Boolean _CFCalendarInitWithIdentifier(CFCalendarRef calendar, CFStringRef identifier) {
-    if (identifier != kCFGregorianCalendar && identifier != kCFBuddhistCalendar && identifier != kCFJapaneseCalendar && identifier != kCFIslamicCalendar && identifier != kCFIslamicCivilCalendar && identifier != kCFHebrewCalendar && identifier != kCFChineseCalendar) {
+    if (identifier != kCFGregorianCalendar && identifier != kCFBuddhistCalendar && identifier != kCFJapaneseCalendar && identifier != kCFIslamicCalendar && identifier != kCFIslamicCivilCalendar && identifier != kCFHebrewCalendar && identifier != kCFRepublicOfChinaCalendar && identifier != kCFPersianCalendar && identifier != kCFCalendarIdentifierCoptic && identifier != kCFCalendarIdentifierEthiopicAmeteMihret && identifier != kCFCalendarIdentifierEthiopicAmeteAlem && identifier != kCFChineseCalendar && identifier != kCFISO8601Calendar && identifier != kCFIslamicTabularCalendar && identifier != kCFIslamicUmmAlQuraCalendar) {
         if (CFEqual(kCFGregorianCalendar, identifier)) identifier = kCFGregorianCalendar;
         else if (CFEqual(kCFBuddhistCalendar, identifier)) identifier = kCFBuddhistCalendar;
         else if (CFEqual(kCFJapaneseCalendar, identifier)) identifier = kCFJapaneseCalendar;
         else if (CFEqual(kCFIslamicCalendar, identifier)) identifier = kCFIslamicCalendar;
         else if (CFEqual(kCFIslamicCivilCalendar, identifier)) identifier = kCFIslamicCivilCalendar;
         else if (CFEqual(kCFHebrewCalendar, identifier)) identifier = kCFHebrewCalendar;
+        else if (CFEqual(kCFRepublicOfChinaCalendar, identifier)) identifier = kCFRepublicOfChinaCalendar;
+        else if (CFEqual(kCFPersianCalendar, identifier)) identifier = kCFPersianCalendar;
+        else if (CFEqual(kCFIndianCalendar, identifier)) identifier = kCFIndianCalendar;
+        else if (CFEqual(kCFCalendarIdentifierCoptic, identifier)) identifier = kCFCalendarIdentifierCoptic;
+        else if (CFEqual(kCFCalendarIdentifierEthiopicAmeteMihret, identifier)) identifier = kCFCalendarIdentifierEthiopicAmeteMihret;
+        else if (CFEqual(kCFCalendarIdentifierEthiopicAmeteAlem, identifier)) identifier = kCFCalendarIdentifierEthiopicAmeteAlem;
         else if (CFEqual(kCFChineseCalendar, identifier)) identifier = kCFChineseCalendar;
+        else if (CFEqual(kCFISO8601Calendar, identifier)) identifier = kCFISO8601Calendar;
+        else if (CFEqual(kCFIslamicTabularCalendar, identifier)) identifier = kCFIslamicTabularCalendar;
+        else if (CFEqual(kCFIslamicUmmAlQuraCalendar, identifier)) identifier = kCFIslamicUmmAlQuraCalendar;
         else return false;
     }
     
--- a/CoreFoundation/Locale.subproj/CFLocaleKeys.c
+++ b/CoreFoundation/Locale.subproj/CFLocaleKeys.c
@@ -126,7 +126,7 @@ CONST_STRING_DECL(kCFCalendarIdentifierChinese, "chinese");
 CONST_STRING_DECL(kCFCalendarIdentifierRepublicOfChina, "roc");
 CONST_STRING_DECL(kCFCalendarIdentifierPersian, "persian");
 CONST_STRING_DECL(kCFCalendarIdentifierIndian, "indian");
-CONST_STRING_DECL(kCFCalendarIdentifierISO8601, "");
+CONST_STRING_DECL(kCFCalendarIdentifierISO8601, "iso8601");
 CONST_STRING_DECL(kCFCalendarIdentifierCoptic, "coptic");
 CONST_STRING_DECL(kCFCalendarIdentifierEthiopicAmeteMihret, "ethiopic");
 CONST_STRING_DECL(kCFCalendarIdentifierEthiopicAmeteAlem, "ethiopic-amete-alem");
--- a/TestFoundation/TestNSCalendar.swift
+++ b/TestFoundation/TestNSCalendar.swift
@@ -20,15 +20,41 @@ class TestNSCalendar: XCTestCase {
     
     static var allTests: [(String, (TestNSCalendar) -> () throws -> Void)] {
         return [
+            ("test_allCalendars", test_allCalendars),
             ("test_gettingDatesOnGregorianCalendar", test_gettingDatesOnGregorianCalendar ),
             ("test_gettingDatesOnHebrewCalendar", test_gettingDatesOnHebrewCalendar ),
             ("test_gettingDatesOnChineseCalendar", test_gettingDatesOnChineseCalendar),
+            ("test_gettingDatesOnISO8601Calendar", test_gettingDatesOnISO8601Calendar),
             ("test_copy",test_copy),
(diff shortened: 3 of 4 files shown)


CoreFoundation/Locale.subproj/CFCalendar.c
```diff
@@ -236,14 +236,23 @@ CFCalendarRef CFCalendarCopyCurrent(void) {
 }
 
 Boolean _CFCalendarInitWithIdentifier(CFCalendarRef calendar, CFStringRef identifier) {
-    if (identifier != kCFGregorianCalendar && identifier != kCFBuddhistCalendar && identifier != kCFJapaneseCalendar && identifier != kCFIslamicCalendar && identifier != kCFIslamicCivilCalendar && identifier != kCFHebrewCalendar && identifier != kCFChineseCalendar) {
+    if (identifier != kCFGregorianCalendar && identifier != kCFBuddhistCalendar && identifier != kCFJapaneseCalendar && identifier != kCFIslamicCalendar && identifier != kCFIslamicCivilCalendar && identifier != kCFHebrewCalendar && identifier != kCFRepublicOfChinaCalendar && identifier != kCFPersianCalendar && identifier != kCFCalendarIdentifierCoptic && identifier != kCFCalendarIdentifierEthiopicAmeteMihret && identifier != kCFCalendarIdentifierEthiopicAmeteAlem && identifier != kCFChineseCalendar && identifier != kCFISO8601Calendar && identifier != kCFIslamicTabularCalendar && kCFIslamicUmmAlQuraCalendar) {
```
alblue: I don't know at what point is stops being an optimisation to look for calendars here. Perhaps it should just be a subset of the available calendars?

Also, there's a `_CFCalendarInitWithIdentifier` and `CFCalendarCreateWithIdentifier` -- I'm not sure of the difference between them (but I've only implemented this one).

CoreFoundation/Locale.subproj/CFLocaleKeys.c
```diff
@@ -126,7 +126,7 @@ CONST_STRING_DECL(kCFCalendarIdentifierChinese, "chinese");
 CONST_STRING_DECL(kCFCalendarIdentifierRepublicOfChina, "roc");
 CONST_STRING_DECL(kCFCalendarIdentifierPersian, "persian");
 CONST_STRING_DECL(kCFCalendarIdentifierIndian, "indian");
-CONST_STRING_DECL(kCFCalendarIdentifierISO8601, "");
+CONST_STRING_DECL(kCFCalendarIdentifierISO8601, "iso8601");
```
alblue: Without this line change, the lookup for ISO8601 failed.

TestFoundation/TestNSCalendar.swift
```diff
@@ -20,15 +20,41 @@ class TestNSCalendar: XCTestCase {
...
+            Calendar.Identifier.islamicCivil,
+            Calendar.Identifier.islamicTabular,
+            Calendar.Identifier.islamicUmmAlQura,
+            Calendar.Identifier.iso8601,
+            Calendar.Identifier.japanese,
+            Calendar.Identifier.persian,
+            Calendar.Identifier.republicOfChina
+            ] {
+                let calendar = Calendar(identifier: identifier)
+                XCTAssertEqual(identifier,calendar.identifier)
```
alblue: I'm not sure if there's a generic property I should be looking up here for all calendars to test that they are created appropriately? At the moment it's just checking the identifier and initialiser.

pushkarnk: We can have tests that create calendars for all types and then verify the year, month and day (something like ```gettingDatesOnGregorianCalendar```).

CoreFoundation/Locale.subproj/CFCalendar.c
pushkarnk: Looks like there's a typo at the end of the condition: 
```identifier != kCFIslamicTabularCalendar **&& kCFIslamicUmmAlQuraCalendar**```

Comment (pushkarnk) [inline CoreFoundation/Locale.subproj/CFCalendar.c]: Also, now that we are supporting all the calendar identifiers, do we need this if statement?

**Comment by pushkarnk** [comment]
@alblue - Do the tests pass on Ubuntu 14.04 as well? I just want to make sure the libicu there supports all these calendar types. Thanks.

File: TestFoundation/TestNSCalendar.swift (hunk 3 above)
Comment (alblue) [inline TestFoundation/TestNSCalendar.swift]: Not all calendars have the same concept of year, though.

File: CoreFoundation/Locale.subproj/CFCalendar.c [hunk 4]
```diff
@@ -236,14 +236,23 @@ CFCalendarRef CFCalendarCopyCurrent(void) {
...
         if (CFEqual(kCFGregorianCalendar, identifier)) identifier = kCFGregorianCalendar;
         else if (CFEqual(kCFBuddhistCalendar, identifier)) identifier = kCFBuddhistCalendar;
         else if (CFEqual(kCFJapaneseCalendar, identifier)) identifier = kCFJapaneseCalendar;
         else if (CFEqual(kCFIslamicCalendar, identifier)) identifier = kCFIslamicCalendar;
         else if (CFEqual(kCFIslamicCivilCalendar, identifier)) identifier = kCFIslamicCivilCalendar;
         else if (CFEqual(kCFHebrewCalendar, identifier)) identifier = kCFHebrewCalendar;
+        else if (CFEqual(kCFRepublicOfChinaCalendar, identifier)) identifier = kCFRepublicOfChinaCalendar;
+        else if (CFEqual(kCFPersianCalendar, identifier)) identifier = kCFPersianCalendar;
+        else if (CFEqual(kCFIndianCalendar, identifier)) identifier = kCFIndianCalendar;
+        else if (CFEqual(kCFCalendarIdentifierCoptic, identifier)) identifier = kCFCalendarIdentifierCoptic;
```
alblue: I'm open to that idea, but will wait for the recommended solution before I make any further changes here.

alblue:
The tests build and pass on Linux, and (for example) you can identify certain components from the calendar, such as the identifier itself and the firstWeekday (which is 1 for all of them). It's not clear if that's the default though; for example, .amSymbol throws for both .gregorian and .iso8601. However, whether the ICU library supports the calendars or not I would expect the Swift code to be the same; that is, if the ICU library is upgraded to support new calendars we should expect to be able to sue them without having to change the Swift code for Calendar again. If we want to dynamically figure out which ones work then we should probably have a separate run-time initialisation step that builds up a list of known calendars from the ICU library.

parkera:
We really only should support the calendars identified by the constants exported in struct Calendar. It used to be the case that the identifier argument to init was a string, so you could pass whatever you liked. But in reality we only supported the known set anyway, which is why I changed it to an enum.

parkera:
@swift-ci please test

alblue:
I checked the set of identifiers at https://developer.apple.com/reference/foundation/nscalendar/1652687-calendar_identifiers against this set; I believe they are the same ones.

alblue:
The test in CI failed on the newly added calendars test:

```
Test Suite 'TestNSCalendar' started at 21:47:18.879
Test Case 'TestNSCalendar.test_allCalendars' started at 21:47:18.879
TestFoundation: /home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift/include/swift/Runtime/../../../stdlib/public/SwiftShims/RefCount.h:252: bool StrongRefCount::doDecrementShouldDeallocate() [ClearPinnedFlag = false]: Assertion `newval + quantum >= RC_ONE && "releasing reference with a refcount of zero"' failed.
/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift/utils/build-script-impl: line 249: 12314 Aborted                 "$@"
/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift/utils/build-script: fatal error: command terminated with a non-zero exit status 134, aborting
Build step 'Execute shell' marked build as failure

```
I'll investigate further as to the cause, although I did not see this in the local Swift build that I tested it against.

alblue:
Rebuilt with latest HEAD and do not see the same failure as shown by the CI. This output from Linux:

```
Test Suite 'TestNSCalendar' started at 11:04:07.958
Test Case 'TestNSCalendar.test_allCalendars' started at 11:04:07.958
Test Case 'TestNSCalendar.test_allCalendars' passed (0.0 seconds).
Test Case 'TestNSCalendar.test_gettingDatesOnGregorianCalendar' started at 11:04:07.958
Test Case 'TestNSCalendar.test_gettingDatesOnGregorianCalendar' passed (0.0 seconds).
Test Case 'TestNSCalendar.test_gettingDatesOnHebrewCalendar' started at 11:04:07.959
Test Case 'TestNSCalendar.test_gettingDatesOnHebrewCalendar' passed (0.0 seconds).

> :version
lldb-local-2016-10-05 (LLDB 99c98286cf, LLVM b9bd56d1b8, Clang 66cbe896f7, Swift-3.0 a05e8412fa)
```

I wonder if this is to do with the latest beta being installed onto the CI servers?

CoreFoundation/Locale.subproj/CFCalendar.c
alblue: The if statement is about optimisation as much as anything else. The `CFEqual` check performs the equality comparison as well but is more details if that fails, so by doing a first-pass through known good identifiers the fast path can be optimised. Whether it's worth doing that for every calendar or not is debatable.

parkera:
Let me try again...

parkera:
@swift-ci please test

alblue:
Same problem. I can remove the failing test which was added here but it's not clear to me why the overrelease is happening, nor why I can't reproduce the error on my Linux build locally.

alblue:
OK, I'm able to reproduce this locally by running `swift/utils/build-script -R -T --llbuild --swiftpm --xctest --foundation --libdispatch -- --reconfigure` which shows the assertion failure.

alblue:
So the bug occurs when iterating across the fourth calendar item. Using this in the for loop:

```
 42.                 let calendar = Calendar(identifier: identifier) 
 43.                 print("identifier: \(identifier)") 
 44.                 print("\(identifier == calendar.identifier)") 
```

 20.                 let calendar = Calendar(identifier: identifier) 
 21.                 print("identifier: \(identifier)") 
 22.                 print("\(identifier == calendar.identifier)") 
 23.             } 
identifier: buddhist
true
identifier: chinese
true
identifier: coptic
true
identifier: ethiopicAmeteAlem
true
repl_swift: /home/ablewitt/swift-build/swift/include/swift/Runtime/../../../stdlib/public/SwiftShims/RefCount.h:252: bool StrongRefCount::doDecrementShouldDeallocate() [ClearPinnedFlag = false]: Assertion `newval + quantum >= RC_ONE && "releasing reference with a refcount of zero"' failed.

but then when running again, it goes through that calendar without any issues:

identifier: buddhist
true
identifier: chinese
true
identifier: coptic
true
identifier: ethiopicAmeteAlem
true
identifier: ethiopicAmeteMihret
true
identifier: gregorian
true
identifier: hebrew
true

pushkarnk:
@alblue : Can we please keep this PR open while the apparent ARC issue is looked into?

alblue:
Sorry, didn't mean to close the request - hit the wrong button.

I have raised https://bugs.swift.org/browse/SR-2879 to look into the release issues, and in the meantime, commented out some of the list contents so that we don't see the four calendars in a row issue that we were seeing earlier. This may allow the test to pass on the Swift CI in the meantime.

alblue:
Looks like I can't reopen this one for some reason. I'll create a new pull request.

alblue:
Since I can't re-open this, I've created https://github.com/apple/swift-corelibs-foundation/pull/677 instead. I've rebased this on the calendar fix, so hopefully the test won't fail like it did last time.

PR #669, opened 2016-10-05
Duplicate HTTP headers should be joined with a comma

NobodyNada:
`headersAsDictionary` didn't support duplicate header fields; if there are multiple occurrences of one field, only one of them was returned.   However, [duplicate header fields are valid](http://stackoverflow.com/q/4371328/3476191), and are equivalent to one comma-separated field.

### DIFF
--- a/Foundation/NSURLSession/HTTPMessage.swift
+++ b/Foundation/NSURLSession/HTTPMessage.swift
@@ -141,12 +141,17 @@ extension URLSessionTask {
 
 extension URLSessionTask._HTTPMessage {
     var headersAsDictionary: [String: String] {
-        var result: [String: String] = [:]
-        headers.forEach {
-            result[$0.name] = $0.value
-        }
-        return result
-    }
+		var result: [String: String] = [:]
+		headers.forEach {
+			if result[$0.name] == nil {
+				result[$0.name] = $0.value
+			}
+			else {
+				result[$0.name]! += (", " + $0.value)
+			}
+		}
+		return result
+	}
 }
 extension URLSessionTask._HTTPMessage {
     /// A single HTTP message header field


Foundation/NSURLSession/HTTPMessage.swift
```diff
@@ -141,12 +141,17 @@ extension URLSessionTask {
 
 extension URLSessionTask._HTTPMessage {
     var headersAsDictionary: [String: String] {
-        var result: [String: String] = [:]
-        headers.forEach {
-            result[$0.name] = $0.value
-        }
-        return result
-    }
+		var result: [String: String] = [:]
```
xwu: Drive-by nit: these lines should use the same indenting as the rest of the file (4 spaces, not tab).

pushkarnk:
Looks good to me. Its in accordance with http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

pushkarnk:
@swift-ci please test

Foundation/NSURLSession/HTTPMessage.swift
NobodyNada: Oops; I forgot I had set Xcode to tabs.  Should I change it?

pushkarnk: Hi @NobodyNada , it'd be great if you could change the identation as @xwu pointed out. Thanks.

JALsnipe: Should be fixed with 5bc8808.

parkera:
@swift-ci please test

swizzlr:
I don't mean to necro a PR that's good and done but my litmus test here would be "what does Foundation do"?

Using a playground and Charles Proxy I discovered that Foundation automatically coalesces the values into a comma separated list. This is probably to better support bad server side parsers that won't recognise the above discussed edge case. I think the Swift implementation here ought to ensure the same thing happens; however, if this already happens for free in CFNetwork, then there's nothing to worry about.

NobodyNada:
@swizzlr I'm  not quite sure what you're saying.  What exactly does Foundation do that this PR doesn't?

swizzlr:
@nobodynada I misread the implementation behaviour - i thought it would result in multiple instances of the header key, but now I see more closely that it does coalesce them with commas - so ignore me! Total mistake on my part

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci please test

PR #671, opened 2016-10-06
Implemented setLocalizedDateFormatFromTemplate

JGiola:
I've implemented the missing method in DateFormatter.

I've only two questions:
1. I've called the relative class method using explicitly the `DateFormatter` class, it is better using the new `type(of:)`?
2. There are no tests for the class method `dateFormat(fromTemplate:opts:locale:)`to get inspiration any guidance on how to add for both the class and the instance method?

### DIFF
--- a/TestFoundation/TestNSDateFormatter.swift
+++ b/TestFoundation/TestNSDateFormatter.swift
@@ -27,7 +27,8 @@ class TestNSDateFormatter: XCTestCase {
             ("test_dateStyleMedium",   test_dateStyleMedium),
             ("test_dateStyleLong",     test_dateStyleLong),
             ("test_dateStyleFull",     test_dateStyleFull),
-            ("test_customDateFormat", test_customDateFormat)
+            ("test_customDateFormat", test_customDateFormat),
+            ("test_setLocalizedDateFormatFromTemplate", test_setLocalizedDateFormatFromTemplate),
         ]
     }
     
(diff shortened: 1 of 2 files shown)


JGiola:
I've change a mistake in which locale use in the implementation and add a simple test

TestFoundation/TestNSDateFormatter.swift
```diff
@@ -277,5 +278,17 @@ class TestNSDateFormatter: XCTestCase {
...
     }
-    
+
+    func test_setLocalizedDateFormatFromTemplate() {
+        let locale = Locale(identifier: DEFAULT_LOCALE)
+        let template = "EEEE MMMM d y hhmmss a zzzz"
+
+        let f = DateFormatter()
+        f.locale = locale
+        f.setLocalizedDateFormatFromTemplate(template)
```
lifeissweetgood: What's the benefit of calling `f.setLocalizedDateFormatFromTemplate(template)` over doing
`f.dateFormat = DateFormatter.dateFormat(fromTemplate: dateFormatTemplate, options: 0, locale: locale)`?

JGiola: Well, here I have to test that the setLocalizedDataFormatFromTemplate will generate the same format as DateFormatter.dateFormat(fromTemplate: dateFormatTemplate, options: 0, locale: locale) if I call directly the latter how can I test it?

JGiola: Oh, right, my bad :)

lifeissweetgood: I am one of the Foundation "guys" 😉.  Looking at the implementation of the obj-c version of this method, the method essentially is meant to be purely a convenience so what you've implemented here is fine.

lifeissweetgood:

TestFoundation/TestNSDateFormatter.swift
JGiola: I will write here because it seems the correct place, I'm sorry for all the people who are working on Foundation and on every other project around the world to have used the term "guys" for reference the group working on Foundation. There are many words to indicate a group of people (like, you know "people") and I'm very sorry to have choose and write the less inclusive of all, I would like to apologise to every person who have read my comment. Please forgive me, and keep up with your amazing work.

PR #673, opened 2016-10-06
Update API of NSProgress to Progress

parkera:
NFC. Also, not much attempt to fix up the comments yet.

No tests because nothing is actually implemented yet.


parkera:

PR #675, opened 2016-10-07
Retain localeID when constructing a Calendar

alblue:
The Calendar destructor releases stored values when the
Calendar is being deallocated, including releasing the localeID.

By default this is an empty string (kCFEmptyString) which results
in multiple Calendars causing a fault when it is over-released.
Ensure that instead the localeID is retained to balance the
release when it is cleaned up.

Issue: SR-2879


parkera:

PR #677, opened 2016-10-07
Add support for ISO8601 calendar and others

alblue:
The list of valid keys did not include the ISO8601 calendar, which
meant that code running on Linux that used it would fail. Fix this
by adding the kCFISO8601Calendar and others into the list of known
constants valid for calendars, and add tests for the Gregorian
calendar, as well as for creating all other calendars.

Issue: SR-2551


parkera:
@swift-ci please test

parkera:
Ack. Hit the wrong button. =)

alblue:
Yeah I did that with the other pull request accidentally as well :-) Still, at least the tests worked out this time.

alblue:
Thanks!

PR #679, opened 2016-10-10
Implementation of most of the API for Progress.

parkera:
This implementation is based on the Darwin NSProgress implementation. One important detail is that swift-corelibs-foundation we do not have Key-Value Observing (KVO) yet. In the meantime, it is suggested to use polling or some other mechanism to read the values.

Updates along the parent/child relationship in Progress are working.

One additional thing that remains is an implementation of the localizedDescription methods. These are customized based on the contents of the userInfo dictionary.


parkera:
@swift-ci please test

parkera:
@swift-ci please test

PR #680, opened 2016-10-10
[SR-2503] Implement subtraction for CharacterSet

xwu:
This is a PR that parallels apple/swift#5201.

Execution is interrupted when `CharacterSet.subtracting(_:)` is invoked. The underlying reason is that `subtracting(_:)` is a default implementation that calls `symmetricDifference(_:)`, but `CharacterSet.symmetricDifference(_:)` invokes `subtracting(_:)`. This PR implements `CharacterSet.subtracting(_:)` and `CharacterSet.subtract(_:)`.

A new test has been added for this code; however, as reported in [SR-2509](https://bugs.swift.org/browse/SR-2509), `CharacterSet.union(_:)` is not currently working in corelibs-foundation, so the test must be disabled in this repo until that entirely separate issue is resolved.

Resolves [SR-2503](https://bugs.swift.org/browse/SR-2503).

### DIFF
--- a/Foundation/CharacterSet.swift
+++ b/Foundation/CharacterSet.swift
@@ -434,29 +434,29 @@ public struct CharacterSet : ReferenceConvertible, Equatable, Hashable, SetAlgeb
         return result
     }
     
-    /// Sets the value to the intersection of the `CharacterSet` with another `CharacterSet`.
+    /// Sets the value to an intersection of the `CharacterSet` with another `CharacterSet`.
     public mutating func formIntersection(_ other: CharacterSet) {
         _applyUnmanagedMutation {
             $0.formIntersection(with: other)
         }
     }
 
-    /// Returns the set of elements in the `CharacterSet` but not in another `CharacterSet`.
+    /// Returns a `CharacterSet` created by removing elements in `other` from `self`.
     public func subtracting(_ other: CharacterSet) -> CharacterSet {
         return intersection(other.inverted)
     }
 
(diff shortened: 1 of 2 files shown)


parkera:
I didn't mention it in the other review, but thanks for finding/fixing this @xwu.

parkera:
Seems like we should get to the bottom of SR-2509 before merging this, so we can have the test enabled.

xwu:
@parkera I agree that'd be preferable.

SR-2509 is tricky, though, because it's probably related to another bug (SR-2510) about `NSMutableCharacterSet` in corelibs-foundation. It seems that it's generally borked: try to invert an instance of `NSMutableCharacterSet` and execution is interrupted; ask if "abc" is a superset of "cde" and you get `true`. I'm not getting anywhere fixing that one.

parkera:
Well at least that one is easy to repro. Looks like we are failing to initialize the right callback into Swift. I'll take a closer look.

parkera:
I talked about this with @phausler and he has a fix for 2510 planned soon.

xwu:
@parkera With the new year and SR-2509/10 being still unresolved, any thoughts on maybe merging this first? While it's true the tests would stay disabled, arguably there's value in aligning with changes already merged in the apple/swift repo?

parkera:
Sure.

parkera:
@swift-ci test and merge

parkera:
I have some changes for CFCharacterSet pending which will help with some extant struct CharacterSet bugs as well.

PR #681, opened 2016-10-11
Fix for SR-2913 - HTTP headers not picked up

pushkarnk:
In `startNewTransfer(with:)` though we first set currentRequest property to the incoming request [here](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURLSession/NSURLSessionTask.swift#L462), the update happens asynchronously on the `taskAttributesIsolation` queue. Subsequently, in `curlHeaders(for: URLRequest)` we use the `currentRequest` property assuming it has been set to get the HTTP headers. 

If we end up picking up the HTTP headers before async block runs, we can get into problems. For new requests, the `currentRequest` property could still be nil, meaning the expected headers will not passed to libcurl. For redirections, we may pick up headers from the previous request. 

I think it must be acceptable to use the incoming URLRequest parameter instead of the property. After all, the `currentRequest` property is expected to be set to the incoming URLRequest.

### DIFF
--- a/Foundation/NSURLSession/NSURLSessionTask.swift
+++ b/Foundation/NSURLSession/NSURLSessionTask.swift
@@ -88,7 +88,7 @@ open class URLSessionTask : NSObject, NSCopying {
         originalRequest = nil
         body = .none
         workQueue = DispatchQueue(label: "URLSessionTask.notused.0")
-        taskAttributesIsolation = DispatchQueue(label: "URLSessionTask.notused.1")
+        taskAttributesIsolation = DispatchQueue(label: "URLSessionTask.notused.1", attributes: DispatchQueue.Attributes.concurrent)
         let fileName = NSTemporaryDirectory() + NSUUID().uuidString + ".tmp"
         _ = FileManager.default.createFile(atPath: fileName, contents: nil)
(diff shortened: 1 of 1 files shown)


pushkarnk:
@swift-ci please test

pushkarnk:
Another problem reported ```currentRequest``` equal to ```nil``` [here](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURLSession/NSURLSessionTask.swift#L794).

The ```taskAttributesIsolation``` queue which is used to make the public attributes thread safe should be a concurrent queue and writes should use barriers. [Line 74](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURLSession/NSURLSessionTask.swift#L74) makes this intention clear but this wasn't implemented. 

I've made suitable changes to the PR.

pushkarnk:
@swift-ci please test

parkera:
LGTM

pushkarnk:

pushkarnk:
@swift-ci please test

pushkarnk:
Looks like a compiler crash. Trying one more time.
@swift-ci please test

pushkarnk:
@swift-ci please test

PR #682, opened 2016-10-11
Use #include over #import in C souce

kastiglione:
This is because Clang's MSVC compatibility does not handle `#import` statements outside of of objc.

    #import of type library is an unsupported Microsoft feature

This is a follow up to a similar change in #663.


modocache:
@swift-ci Please test

modocache:
Looks good to me! I defer to @parkera or @phausler, though.

modocache:
@swift-ci please test

modocache:
@swift-ci Please test.

parkera:
Thanks!

PR #683, opened 2016-10-12
Suppress string interpolation warnings

e78l:
Suppresses new warnings from https://github.com/apple/swift/pull/5195

- Avoid optional where no impact w/IUO
- Mostly cast to Optional
- Change in Notification description() to avoid issue

### DIFF
--- a/Foundation/NSKeyedArchiver.swift
+++ b/Foundation/NSKeyedArchiver.swift
@@ -241,8 +241,7 @@ open class NSKeyedArchiver : NSCoder {
     }
     
     private func _validateObjectSupportsSecureCoding(_ objv : Any?) {
-        if objv != nil &&
-            self.requiresSecureCoding &&
+        if let objv = objv, self.requiresSecureCoding &&
             !NSKeyedArchiver._supportsSecureCoding(objv) {
             fatalError("Secure coding required when encoding \(objv)")
         }
--- a/Foundation/Notification.swift
+++ b/Foundation/Notification.swift
@@ -40,9 +40,12 @@ public struct Notification : ReferenceConvertible, Equatable, Hashable {
     public var hashValue: Int {
         return name.rawValue.hash
     }
-    
+
     public var description: String {
-        return "name = \(name.rawValue),  object = \(object), userInfo = \(userInfo)"
+        var description = "name = \(name.rawValue)"
+        if let obj = object { description += ", object = \(obj)" }
+        if let info = userInfo { description += ", userInfo = \(info)" }
+        return description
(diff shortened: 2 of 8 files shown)


parkera:
Why do we have to cast things to optional all over the place in the tests?

CodaFi:
We considered three ways to silence this warning:

1) Explicitly request the debug description of the optional to be printed with `String(describing:)`
2) Unwrap with a `guard` or `if let` statement (or `!`) - we recommend `?? <default>` in the fixit
3) Cast to `Any` or `Optional` to make it explicit

I think some of these could better be handled by providing defaults than by force unwrapping or casting.

Foundation/NSKeyedArchiver.swift
```diff
@@ -244,7 +244,7 @@ open class NSKeyedArchiver : NSCoder {
         if objv != nil &&
             self.requiresSecureCoding &&
             !NSKeyedArchiver._supportsSecureCoding(objv) {
-            fatalError("Secure coding required when encoding \(objv)")
```
CodaFi: You should change this to bind `obj` in an `if-let` and remove the force unwrap.

Foundation/Notification.swift
```diff
@@ -40,9 +40,12 @@ public struct Notification : ReferenceConvertible, Equatable, Hashable {
...
         return name.rawValue.hash
     }
-    
+
     public var description: String {
-        return "name = \(name.rawValue),  object = \(object), userInfo = \(userInfo)"
+        var description = "name = \(name.rawValue)"
+        if let obj = object { description += ", object = \(obj)" }
+        if let info = userInfo { description += ", userInfo = \(info)" }
+        return description
```
CodaFi: This changes the behavior of `description`.  Is it relevant to see `nil` in here?

e78l:
Hello Tony and Robert,
Thanks for taking a look!
In general, I was trying to change as little as necessary –
For a string like: "expected result was \(value)", I felt it briefest to cast the value to Optional: "expected result was \(value as Optional)"
The only real change, I think, is in Notification which (as Robert points out) may affect compatibility - should the code in my comment be used instead?

parkera:
I'm not too concerned about maintaining an exact match of the description method on `Notification` - as you point out, it's supposed to be for debugging only, anyway.

CodaFi:
Well then, LGTM unless there's a convention you want to enforce around optional values in interpolation segments.

CodaFi:
@swift-ci please test.

CodaFi:
⛵️

PR #686, opened 2016-10-13
Add Calendar fixes for Swift 3

alblue:
These fix a crashing bug with Calendar when more than four instances are created and destroyed, and adds support for ISO8601 Calendars and others.


pushkarnk:
@swift-ci please test

PR #687, opened 2016-10-14
Final implementation of Decimal

alblue:
This extends the normalisation routines to be able to handle
mismatched size elements, and division of decimals that are larger
than short values correctly. It adds tests that are the same coverage
as for NSDecimal on Darwin.

### DIFF
--- a/Foundation/NSDecimal.swift
+++ b/Foundation/NSDecimal.swift
@@ -33,6 +33,9 @@ public struct Decimal {
             return UInt32((__lengthAndFlags & 0b0000_1111))
         }
         set {
+            guard newValue <= maxMantissaLength else {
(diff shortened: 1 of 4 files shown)


Foundation/NSDecimal.swift
```diff
@@ -1479,26 +1566,6 @@ fileprivate struct WideDecimal : VariableLengthNumber {
             }
         }
     }
-    public func compare(to rhs: WideDecimal) -> ComparisonResult {
```
alblue: This was added as part of a transitional implementation in the previous commit, but is no longer necessary - so it has been removed.

pushkarnk:
@swift-ci please test

PR #688, opened 2016-10-14
Add tests for NSClassFromString and NSStringFromClass

JGiola:
I've add simple test coverage for NSStringFromClass and NSClassFromString


parkera:

JGiola:
It seems to fail tests on Linux, but I can't find what is failing, where I have to look?

JGiola:
Found it! It was the tests with the class from XCTest that was failing because on Linux the module is named SwiftXCTest and not only XCTest.

parkera:

PR #690, opened 2016-10-18
Implement FileHandle.nullDevice

kareman:
Not sure if the unit test is needed. I may be testing OS functionality instead of Foundation.

### DIFF
--- a/Foundation/NSFileHandle.swift
+++ b/Foundation/NSFileHandle.swift
@@ -214,11 +214,15 @@ extension FileHandle {
     open class var standardError: FileHandle {
         return _stderrFileHandle
     }
-    
+
+    internal static var _nulldeviceFileHandle: FileHandle = {
+        return FileHandle(forUpdatingAtPath: "/dev/null")!
+    }()
+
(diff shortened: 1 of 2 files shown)


Foundation/NSFileHandle.swift
```diff
@@ -214,11 +214,15 @@ extension FileHandle {
     open class var standardError: FileHandle {
         return _stderrFileHandle
     }
-    
+
+    internal static var _nulldeviceFileHandle: FileHandle = {
+        return FileHandle(forUpdatingAtPath: "/dev/null")!
```
parkera: What's the portability story around `/dev/null`? 

For what it's worth, the Darwin Foundation implementation here is a simple subclass of FileHandle that returns empty `Data` and values of `0` (except for file handle, which is `-1`).

kareman:
@parkera this should be more like the Darwin implementation.

kareman:
Any news on this?

parkera:
Oh, sorry! This looks good to me.

parkera:

PR #691, opened 2016-10-18
Add more NSRange tests

JGiola:
I've add more tests to cover NSRange implementation. They are pretty straight forward, but anyone can remember because the one for NSRangeFromString is commented out? On macOS it pass just fine, there is some issues on Ubuntu?


parkera:

parkera:
Let's try uncommenting it in a separate PR; we can run the tests and find out. It may be a holdover from something that's already fixed.

PR #692, opened 2016-10-19
Enable NSRangeFromString test

JGiola:
Uncommented the NSRangeFromString test, on macOS it works fine, on Linux it has to be seen if there will be issues.


pushkarnk:
It looks like we can enable this test now. Let's try.

@swift-ci please test

parkera:

PR #699, opened 2016-10-27
[NSJSONSerialization] SR-3013 Fix incorrect boolean coercion and enhance NSNumber encoding

swizzlr:
Please review!

This pull request resolves [SR-3013, "JSONSerialization coerces booleans to numbers on Linux"](https://bugs.swift.org/browse/SR-3013)., relates to [SR-72](https://bugs.swift.org/browse/SR-72)

The prevailing bug was due to a prior kludge: if the value could be coerced into an NSNumber, coerce to an NSNumber and encode. NSNumbers were (before this PR) type-opaque, so an `NSNumber(value: true)` -> `"1"`.

I resolved this by first rewriting the type checking code to use a pattern match so the intention was clearer, then added a special case for the primitive bool case. The switch statement should be a clear set of entry points for future maintainers, following the JSON spec.

Next, I implemented objCType for NSNumber. This allows us to correctly discover when an NSNumber is in use boxing a boolean and encode it properly.

As I write this, I realise that we need to:
- [x] tighten up the _NSObjCType assignments in the NSNumber initialiser to correctly reflect `Long` etc.

Future work, which could be included in this PR as it is relatively trivial but definitely out of scope of the bug is to:
1. Differentiate between Integers and Numbers as per the JSON spec (yes, they're different!) by inspecting NSNumber
2. Rewrite the pattern match code to stop using `_SwiftValue.store` and use something less obtuse e.g. type checking.

Longer term I think this code could do with some protocol oriented restructuring, which I would like to discuss with the code owner, whoever they are!

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -333,10 +333,12 @@ private struct JSONWriter {
             throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Number cannot be infinity or NaN"])
         }
         
-        // Cannot detect type information (e.g. bool) as there is no objCType property on NSNumber in Swift
-        // So, just print the number
-
-        writer(_serializationString(for: num))
+        switch num._objCType {
+        case .Bool:
+            serializeBool(num.boolValue)
+        default:
+            writer(_serializationString(for: num))
+        }
     }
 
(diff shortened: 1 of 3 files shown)


swizzlr:
Would anybody like to review this?

CodaFi:
👍 This LGTM, thank you for fixing this.  I can't speak for the Foundation team, but I can speak to Swift-ci

@swift-ci please test.

parkera:
This looks good to me too. Thanks!

parkera:

PR #700, opened 2016-10-28
SR-2301 - Fix XMLParser

KingOfBrian:
XMLParser was completely broken, even though the bridging was available and things are 99% complete. This PR adds a number of tests to show the problem, and fixes a number of issues:
- The contract was different from Foundation. Methods were not optional and NSError was still in use.
- There was an incorrect nullability annotation on the C Shim between libxml2. This was causing a crash when processing namespaces.
- The `bomChunk` parsing was broken. There was no way for the parser to actually be constructed.
- Some of the logic as to what values get reported to `startElement` and `endElement` I believe was wrong. I had trouble fixing the logic for the nullability changes and decided that it was easier to re-write things.

There's a lot of code in this file hasn't been spruced up in a while and doesn't follow many swift best practices. I'm not sure what the policy was on cleaning things up was, so I tried to just focus on my bugs. However I would be surprised if a cleanup didn't find a few more issues.

Brian King

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -171,7 +171,7 @@ struct _NSXMLParserBridge {
     void (*startElementNs)(_CFXMLInterface ctx,
                            const unsigned char *localname,
                            const unsigned char *_Nullable prefix,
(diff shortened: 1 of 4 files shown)


parkera:
@swift-ci please test

parkera:
Let's get this in.

parkera:

PR #702, opened 2016-10-28
[Notification] Implement CustomReflectable for Notifications

swizzlr:
Please review!

I've reverse engineered the reflectable behaviour to be effectively:
1. Translate NSNotification.Name types to Strings
2. Only include userInfo and object in children when the properties are non-nil
3. Provide them in the order of name, object, userInfo

This PR implements tests and the implementation to provide this.

For reference, this code can be dumped into a playground to see how Apple do it.

```
//: Playground - noun: a place where people can play

import UIKit
import Foundation

let thisObject = NSObject()
let anotherObject = NSObject()
let userInfo = ["something": 1, "anotherObject": anotherObject] as [String : Any]

let notif = Notification(name: .UIScreenDidConnect, object: thisObject, userInfo: userInfo)

let customMirror = notif.customMirror
let mirror = Mirror(reflecting: notif)

var a = Array(mirror.children).makeIterator()
a.next()
a.next()
a.next()


customMirror.subjectType
mirror.subjectType

customMirror.description
mirror.description

customMirror.displayStyle
mirror.displayStyle

customMirror.superclassMirror
mirror.superclassMirror

mirror.subjectType
```


parkera:

parkera:
see also:

https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/Notification.swift#L87

swizzlr:
Oh, so is there parallel work happening in both repos?

parkera:
Yah, unfortunately they are two different source files. One is compiled into the stdlib for Darwin, the other is part of the Foundation library for Linux. I'd love to figure out a way to keep them more in sync in the future, but some parts are fundamentally different (e.g., ObjC bridging logic).

PR #703, opened 2016-10-31
Return empty data when converting from empty string

rintaro:
`"".data(using: .utf8)` should return empty `Data` (instead of `nil`).

In Darwin, `"".data(using: .utf8) == Data()`


parkera:
Thanks!

parkera:

PR #704, opened 2016-11-01
Fix for SR-3101 - crash on `.trimmingCharacters(in:)`

naithar:
Replaced a forced optional unwrapping (`!`) of `UnicodeScalar` which resulted in crash if there was an emoji symbol at start or end of the string with `let` constraint.
Issue: https://bugs.swift.org/browse/SR-3101


parkera:

PR #706, opened 2016-11-02
SR-3052. Incorrect string result for String methods with non latin symbols

naithar:
Fixes Issue: https://bugs.swift.org/browse/SR-3052
Added tests that cover this issue. 
Implemented possible fix by providing correct UTF8-length for `_CFStringCreateByAddingPercentEncodingWithAllowedCharacters` and `_CFStringCreateByRemovingPercentEncoding`

Also switched to `CFStringGetLengthUTF8` method for string length where UTF-8 encoding is used.

### DIFF
--- a/Foundation/NSCFString.swift
+++ b/Foundation/NSCFString.swift
@@ -140,19 +140,18 @@ internal func _CFSwiftStringGetCharacters(_ str: AnyObject, range: CFRange, buff
 }
 
 internal func _CFSwiftStringGetBytes(_ str: AnyObject, encoding: CFStringEncoding, range: CFRange, buffer: UnsafeMutablePointer<UInt8>?, maxBufLen: CFIndex, usedBufLen: UnsafeMutablePointer<CFIndex>?) -> CFIndex {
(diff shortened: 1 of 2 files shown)


parkera:
Can we fix this issue without introducing a new length function for CFString? There is no "CFStringGetLengthUTF8" function on Darwin and I don't think we should introduce one. CFString has the `CFStringGetMaximumSizeForEncoding` function for dealing with differently-sized buffers when the string is encoded into something besides its native representation.

naithar:
I'll look into this more. Thanks for the tip on `CFStringGetMaximumSizeForEncoding`.

naithar:
@parkera updated fix. Tests is passing.

parkera:
Thanks!

parkera:
@swift-ci please test

parkera:
Ok, don't know why the tests are stuck. Trying again.

parkera:
@swift-ci please test

naithar:
Looks like everything is ok now

PR #708, opened 2016-11-08
Implementation of NSCoding for URLCredential

naithar:
Implemented a Keyed coding for URLCredentials. Non keyed is not supported in MacOS Foundation, so it's wasn't implemented.
Added test to cover `NSCoding` changes.
Overriden `isEqual` method to match MacOS Foundation.

### DIFF
--- a/Foundation/NSURLCredential.swift
+++ b/Foundation/NSURLCredential.swift
@@ -65,11 +65,32 @@ open class URLCredential : NSObject, NSSecureCoding, NSCopying {
      */
     
     public required init?(coder aDecoder: NSCoder) {
-        NSUnimplemented()
+        guard aDecoder.allowsKeyedCoding else {
+            preconditionFailure("Unkeyed coding is unsupported.")
+        }
+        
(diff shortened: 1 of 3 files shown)


naithar:
Is URLCredentials `NSCopying` actually implemented since it's not marked `NSUnimplemented()` and matches the MacOS behavior?

Foundation/NSURLCredential.swift
```diff
@@ -84,6 +105,20 @@ open class URLCredential : NSObject, NSSecureCoding, NSCopying {
         return self 
     }
     
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? URLCredential else {
+            return false
+        }
+        
+        guard other !== self else {
```
parkera: A style note, but this feels more like a traditional if/else than a guard to me. I'd like to preserve the use of guard for preconditions and such.

naithar:
@parkera updated

parkera:

parkera:
thanks!

PR #710, opened 2016-11-09
Implementation for NS(Mutable)URLRequest NSCoding.

naithar:
Implemented `NSCoding` for `NSURLRequest` and `NSMutableURLRequest`.
Overriden `isEqual(_:)` method to conform to macOS Foundation.
Also fixed `NSData.bytes` crash when empty data was decoded. `test_NSCoding_2()` tests this crash.

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -217,11 +217,12 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     }
     
     open var bytes: UnsafeRawPointer {
-        guard let bytes = CFDataGetBytePtr(_cfObject) else {
+        guard let bytePtr = CFDataGetBytePtr(_cfObject) else {
             //This could occure on empty data being encoded.
-            return UnsafeRawPointer([])
+            //TODO: switch with nil when signature is fixed
+            return UnsafeRawPointer(bitPattern: 0xf00deadb0c0)! //would not result in 'nil unwrapped optional'
         }
-        return UnsafeRawPointer(bytes)
+        return UnsafeRawPointer(bytePtr)
     }
 
     
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -217,7 +217,11 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     }
     
     open var bytes: UnsafeRawPointer {
-        return UnsafeRawPointer(CFDataGetBytePtr(_cfObject))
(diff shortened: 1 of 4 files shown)


naithar:
@parkera updated

Foundation/NSData.swift
```diff
@@ -217,7 +217,11 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     }
     
     open var bytes: UnsafeRawPointer {
-        return UnsafeRawPointer(CFDataGetBytePtr(_cfObject))
+        guard let bytes = CFDataGetBytePtr(_cfObject) else {
+            //This could occure on empty data being encoded.
+            return UnsafeRawPointer([])
+        }
+        return UnsafeRawPointer(bytes)
```
phausler: I think the return value here might not be safe since the array address is invalid after the return. In reality the "proper fix" is for us to change the signature of bytes

naithar: Wouldn't changing signature of `bytes` break conformance to MacOS Foundation?.
Also rechecked `NSData` in MacOS Foundation. `.bytes` there is also UnsafeRawPointer, but returns nil if `.length` is 0. Missed this at first.
Would changing this to something like this work:
```swift
guard self.length > 0 else {
    return nil
}
return UnsafeRawPointer(CFDataGetBytePtr(_cfObject))
```
Or an issue with invalid address will persist?

phausler: I am claiming that the objective-c version is not correctly annotated. Try this in objc:

```objc
assert([NSData data].bytes == NULL);
```

The header for NSData should probably be marked as 

```objc
@property (readonly) const void *_Nullable bytes NS_RETURNS_INNER_POINTER;
```

naithar: Finally understood the problem - swift is unable to create `UnsafeRawPointer(nil)` value.

naithar: Maybe temporarily implementing a static constant for `UnsafeRawPointer([])` would help in not invalidating a pointer value?

phausler: I would claim that a bit pattern would be more appropriate here than returning a pointer of deallocated memory: e.g. `UnsafeRawPointer(bitPattern: 0xdeadc0c0)` or something like that until we can fix the signature.

naithar:
@phausler would this work as temporarily workaround?

phausler:
👍

parkera:
@swift-ci please test

naithar:
@parkera seems like tests are stuck for this PR

parkera:

PR #715, opened 2016-11-14
suggestedFilename is by definition always nonnull

Coeur:


pushkarnk:
Looks reasonable to me.

pushkarnk:

pushkarnk:
However,though I appreciate that this will not affect existing uses of `suggestedFilename`, I feel a little uncomfortable. The API clearly defines this as an optional String. I just want to make sure we are adhering to it completely.

pushkarnk:
This is from the Swift reference: `An implicitly unwrapped optional is a normal optional behind the scenes, but can also be used like a nonoptional value, without the need to unwrap the optional value each time it is accessed.`

I am a little worried that this property being a part of the API could be used in two ways (as an optional or as a non-optional).

naithar:
Apple documentation suggests that `.suggestedFilename` is normal optional, not unwrapped one: 
[link](https://developer.apple.com/reference/foundation/urlresponse/1415924-suggestedfilename)
Shouldn't swift-foundation correspond to MacOS version?

parkera:
Yes, this is nullable in the Foundation API. We should leave it with the `?` for now, although if the Foundation API is wrong we can perhaps correct it there first.

Coeur:
@parkera documentation from the link also says "Accessing this property always returns a valid filename" and the implementation guaranties it's non-null from the initialiser itself:

https://github.com/Apple/swift-corelibs-foundation/blob/f8c7d7385e485b38a41d1309d5caa07c3dc11ad1/Foundation/NSURLResponse.swift#L55

Where can I pull request a fix for Foundation API?

parkera:
Please file a bug at bugreport.apple.com and let me know what the radar number is.

Coeur:
@parkera radar is 29525615

parkera:
Thanks, I've forwarded it to the right team to get a definitive answer. I'll let you know what we find out.

PR #717, opened 2016-11-16
[WIP] SR-936. Segmentation fault on NSMutableData().hash

naithar:
Issue: https://bugs.swift.org/browse/SR-936

Currently this code: `NSMutableData().hash` - produces a Segmentation fault.
The reason i found is that on `.hash` call, CoreFoundation uses `_CFIsSwift` to check if used class is Swift class or not. 
For `NSData` the result is`false` and runs normally without an error. 
For `NSMutableData` the result is `true` which causes it to call `.hash` again and creates a loop.

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -944,11 +944,6 @@ open class NSMutableData : NSData {
             CFDataReplaceBytes(_cfMutableObject, CFRangeMake(range.location, range.length), bytePtr, replacementLength)
         }
     }
-    
-    /// SR-936 workaround
-    open override var hash: Int {
-        return (self.copy() as! NSData).hash
-    }
 }
 
 extension NSData : _StructTypeBridgeable {
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -944,6 +944,11 @@ open class NSMutableData : NSData {
             CFDataReplaceBytes(_cfMutableObject, CFRangeMake(range.location, range.length), bytePtr, replacementLength)
         }
     }
+    
+    /// SR-936 workaround
+    open override var hash: Int {
+        return (self.copy() as! NSData).hash
+    }
 }
 
 extension NSData : _StructTypeBridgeable {
--- a/TestFoundation/TestNSData.swift
+++ b/TestFoundation/TestNSData.swift
@@ -90,6 +90,8 @@ class TestNSData: XCTestCase {
             ("test_initDataWithCapacity", test_initDataWithCapacity),
             ("test_initDataWithCount", test_initDataWithCount),
             ("test_emptyStringToData", test_emptyStringToData),
+            
+            ("test_hash", test_hash)
         ]
     }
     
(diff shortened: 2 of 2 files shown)


naithar:
This issue is also valid for `NSCharacterSet` and `NSMutableCharacterSet` as they return `true` in `_CFIsSwift` as they are not contained in `__CFRuntimeObjCClassTable` array.

naithar:
@parkera can you point in any direction for this one?

naithar:
Would this work as a fix? [Link](https://github.com/naithar/swift-corelibs-foundation/commit/d5cfbc0a019dfb7ce1c5b05c9271fb035aaaa86e)

parkera:
@phausler can you take a look at this one?

phausler:
changing the class tables would be a rather divergent change from how CF handles this in objc; I would consider that divergence a breaking change that would not be able to be merged back to the internal CF. 

There are few issues at hand here:
1) we need factory pattern initializers to actually return NSCF types from the init methods
2) we need to have dynamic dispatch callouts to the NS types that cover all of the cases like for example hashing
3) we need many more unit tests to verify all of this

naithar:
So what would be the best direction to take with this atm? 
Using a same workaround as in [workaround commit](https://github.com/apple/swift-corelibs-foundation/pull/717/commits/c7050b2bd91756b0548145de53815bb20b6ff4c1) for `NSMutableData` but for other classes that possibly used in `_CFIsSwift` check? 
Or is there better approach to this? I can't quite get it.

naithar:
When using any `NS(Mutable)CharacterSet` methods that calls to CoreFoundation methods, for example: `addCharacters(in: )`, `removeCharacters(in: )` - the use of`CF_SWIFT_FUNCDISPATCHV` leads to crash as there is no methods that are set to `__CFSwiftBridge` in `__CFInitializeSwift` for `NSMutableCharacterSet`. 
Implementing this methods leads to infinite loop, as `_CFIsSwift` returns true.
This also causes the crash of `CharacterSet` as described in: https://bugs.swift.org/browse/SR-3215.
But `CharacterSet` crahses can be fixed by using `_NSCFCharacterSet` in initializer instead of `NSCharacterSet`

naithar:
@phausler is it possible that #1035 can be a possible fix for this (segfault) issues?

I've tried using the factory initializer in [this commit](https://github.com/naithar/swift-corelibs-foundation/blob/ca886ee42f12fe2715a4746250df1d625c5199fe/Foundation/NSData.swift#L93-L118) and it seems to be working, but I'm not exactly sure if I was using it correctly.

phausler:
NSData should be using factory patterns but there are some edge cases I have not yet worked out - primarily how DI rules get enforced with swift. I need to do some more research to figure out how we can get that to work

naithar:
Thanks for the answer :) 
I think that factory initializer as it is right now is already a big step. Hopefully `NSCharacterSet`'s issues causing segmentation faults could be fixed more easily than `NSData`'s.

Also while I was trying to implement factory initializer for `NSData` I've faced some issues, like fields not allowing access. 
For example `_NSSwiftData`'s `_backing` did not allow access to setter or getter, that's why I've switch it with `NSData`. This is probably one of many edge cases that you've mentioned.

phausler:
yea NSCharacterSet should be dramatically easier to implement as a factory pattern since it funnels a bit cleaner

alblue:
This pull request seems to have gone stale, and has conflicting files. It's also not clear whether other requests have superseded it. In the case that it's still required, please re-open with a rebased version of the change.

PR #718, opened 2016-11-17
Improve JSON serialization performance on Linux

djones6:
This change improves the performance of serializing objects to a Data, by using String concatenation to build the result, and then convert to a Data once.  This change was developed by @shmuelk and myself.

We may want to revisit this in the future as the performance of appending to Data improves.

Note, we initially used `jsonStr.cString(using: .utf8)` to create the Data, but in testing this seemed to leak memory (perhaps this is a bug): https://github.com/IBM-Swift/SwiftyJSON/pull/22

I've provided a microbenchmark to demonstrate the effect of these changes:

Before:
```
TestFoundation/TestNSJSONSerialization.swift:760: Test Case 'TestNSJSONSerialization.test_serialization_perf' measured [Time, seconds] average: 0.463, relative standard deviation: 3.143%, values: [0.485122, 0.469292, 0.439064, 0.447310, 0.455177, 0.482358, 0.455306, 0.468886, 0.468181, 0.461310], performanceMetricID:org.swift.XCTPerformanceMetric_WallClockTime, maxPercentRelativeStandardDeviation: 10.000%, maxStandardDeviation: 0.100
Test Case 'TestNSJSONSerialization.test_serialization_perf' passed (4.632 seconds).
```
After:
```
TestFoundation/TestNSJSONSerialization.swift:760: Test Case 'TestNSJSONSerialization.test_serialization_perf' measured [Time, seconds] average: 0.238, relative standard deviation: 8.555%, values: [0.235258, 0.234057, 0.224006, 0.257510, 0.248202, 0.241663, 0.209058, 0.259959, 0.204592, 0.261825], performanceMetricID:org.swift.XCTPerformanceMetric_WallClockTime, maxPercentRelativeStandardDeviation: 10.000%, maxStandardDeviation: 0.100
Test Case 'TestNSJSONSerialization.test_serialization_perf' passed (2.376 seconds).
```
The above results were collected on Ubuntu 16.04.  I obtained similar results on OS X.

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -101,14 +101,13 @@ open class JSONSerialization : NSObject {
     /* Generate JSON data from a Foundation object. If the object will not produce valid JSON then an exception will be thrown. Setting the NSJSONWritingPrettyPrinted option will generate JSON with whitespace designed to make the output more readable. If that option is not set, the most compact possible JSON will be generated. If an error occurs, the error parameter will be set and the return value will be nil. The resulting data is a encoded in UTF-8.
      */
     internal class func _data(withJSONObject value: Any, options opt: WritingOptions, stream: Bool) throws -> Data {
-        var result = Data()
+        var jsonStr = String()
         
         var writer = JSONWriter(
             pretty: opt.contains(.prettyPrinted),
             writer: { (str: String?) in
                 if let str = str {
-                    let count = str.lengthOfBytes(using: .utf8)
-                    result.append(UnsafeRawPointer(str.cString(using: .utf8)!).bindMemory(to: UInt8.self, capacity: count), count: count)
+                    jsonStr.append(str)
                 }
             }
         )
@@ -131,6 +130,14 @@ open class JSONSerialization : NSObject {
             }
         }
         
+        let count = jsonStr.lengthOfBytes(using: .utf8)
+        let bufferLength = count+1 // Allow space for null terminator
+        var utf8: [CChar] = Array<CChar>(repeating: 0, count: bufferLength)
+        if !jsonStr.getCString(&utf8, maxLength: bufferLength, encoding: .utf8) {
+            fatalError("Failed to generate a CString from a String")
+        }
+        let rawBytes = UnsafeRawPointer(UnsafePointer(utf8))
+        let result = Data(bytes: rawBytes.bindMemory(to: UInt8.self, capacity: count), count: count)
         return result
     }
     open class func data(withJSONObject value: Any, options opt: WritingOptions = []) throws -> Data {
--- a/TestFoundation/TestNSJSONSerialization.swift
+++ b/TestFoundation/TestNSJSONSerialization.swift
@@ -15,7 +15,7 @@
     import SwiftFoundation
     import SwiftXCTest
 #endif
-import Dispatch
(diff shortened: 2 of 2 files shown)


TestFoundation/TestNSJSONSerialization.swift
```diff
@@ -743,6 +744,35 @@ extension TestNSJSONSerialization {
...
+                        } catch {
+                            XCTFail("Could not serialize to JSON: \(error)")
+                        }
+                    }
+                }
+            }
+            _ = group.wait(timeout: .distantFuture) // forever
+        }
+    }
+
```
parkera: Thanks for including a performance test.

I'd like to split it out into some other place. The unit test is to verify correctness in automation. As this lacks an assert, it wouldn't even really catch any regressions in reality.

Foundation/NSJSONSerialization.swift
```diff
@@ -131,6 +130,14 @@ open class JSONSerialization : NSObject {
             }
         }
         
+        let count = jsonStr.lengthOfBytes(using: .utf8)
+        let bufferLength = count+1 // Allow space for null terminator
+        var utf8: [CChar] = Array<CChar>(repeating: 0, count: bufferLength)
+        if !jsonStr.getCString(&utf8, maxLength: bufferLength, encoding: .utf8) {
+            fatalError("Failed to generate a CString from a String")
```
parkera: Why would getCString fail here? If it did, is there a way to recover instead of asserting?

djones6: Good question.  The API documentation for Swift String.getCString does not specify what the return value means, but I gather from the implementation that it calls through to NSString.getCString,  the documentation for which says that false would be returned if the buffer is too small, or an encoding error occurs.
In this case, we sized the buffer appropriately, and because UTF8 can represent all Unicode code points, we shouldn't encounter an encoding error.  Perhaps it would be safe to ignore the return value in this case?

TestFoundation/TestNSJSONSerialization.swift
djones6: I wasn't sure where Foundation performance tests belong.  Would the [Swift Benchmark Suite](https://github.com/apple/swift/tree/master/benchmark) be a better place?  I'm happy to remove it from this PR and submit a separate one.

Foundation/NSJSONSerialization.swift
parkera: If we're sure it should not happen, then asserting on that is the right answer instead of ignoring.

TestFoundation/TestNSJSONSerialization.swift
parkera: We don't really have the infrastructure yet, but we should set something up.

I think we should start a discussion on swift-dev mailing list and see if the benchmark suite is the right place or not. If not, we can create our own as part of the project.

djones6:
@parkera I've removed the performance test from this PR, and I posted to the swift-corelibs-dev mailing list re. where it belongs.

djones6:
@parkera Assuming you're happy with the changes, is there still time to get this (and #723) considered for inclusion in the 3.0.2 release?

parkera:

parkera:
Let's get this into master and then I'll check on the Swift 3.0.2 question.

PR #720, opened 2016-11-18
Use NSTemporaryDirectory helper function instead of hard coding /tmp

parkera:
NSTemporaryDirectory properly checks the environment variable TMP.


parkera:

PR #721, opened 2016-11-21
PLIST to Data alignment with Foundation

e78l:
- Going off:
https://developer.apple.com/reference/foundation/propertylistserialization/1418309-data
The `plist` in data(fromPropertyList plist: Any, format: PropertyListFormat, options opt: WriteOptions)  should be Any not AnyObject
- Also removed a seemingly unnecessary CFPropertyList cast


parkera:
We must have missed this one in our audit after the 'id as Any' changes landed. Thanks.

parkera:

PR #722, opened 2016-11-22
Use PKG_CONFIG for libicu foundation.CFLAGS

spevans:
- Mirrors use for foundation.LDFLAGS, allowing libicu to be
  used if installed in non-default include directory


ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
We've tested this on Ubuntu 14.04 and 16.04 and verified it allows Foundation to be built against a custom libicu, if wanted.  There is no change to current behaviour, by default we continue to link against the system libicu.

PR #723, opened 2016-11-23
Further improve performance of JSON serialization on Linux

djones6:
Following on from #718, this change further improves performance of JSON serialization:
- In `JSONWriter.serializeJSON`, move the expensive NSNumber case (which invokes `_SwiftValue.store`) below the cases for Array, Dictionary and primitive types.
- add fast paths for Int and UInt types, which can be converted directly to a String rather than going via NSNumber.
- Equivalent changes for `isValidJSONObject`

These changes were developed by @shmuelk and myself.

Additional performance can be gained by adding similar fast paths for Double and Float, however doing so fails the tests that expect a whole number to be serialized without decimal places (eg. `Double(1.0)` -> `1`).  For that reason, I have not included a fast path for Double or Float in this PR, but I included the performance numbers below for the sake of interest.

Update: I've removed the Int/UInt fastpath for now, until I can resolve the control over formatting issue raised in the review below.  This PR is now just contains the 1.2x improvement (from case reordering).  I'll submit the fastpaths in a future PR.

Using [the benchmark I developed for #718](https://github.com/djones6/swift-corelibs-foundation/commit/867109e253550e4d1bcb99cfc681fdb02ce0f566), increased to 10,000 repetitions, I get the following results on Ubuntu 16.04:

```
Payload type:       Int   (speedup)        Double   (speedup)

Baseline:           8.72                   8.78
With #718:          4.18  (2.1x)           3.97   (2.2x)
Reorder case:       3.43  (1.2x)           3.29   (1.2x)
Int/UInt fastpath:  2.33  (1.5x)           ----

(Double fastpath):  ----                   2.42   (1.35x)
```
(units are average execution time in seconds,  and the code changes are cumulative).

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -287,10 +287,6 @@ private struct JSONWriter {
             try serializeString(str)
         case let boolValue as Bool:
             serializeBool(boolValue)
-        case let num as Int:
-            writer(String(describing: num))
-        case let num as UInt:
-            writer(String(describing: num))
         case let array as Array<Any>:
             try serializeArray(array)
         case let dict as Dictionary<AnyHashable, Any>:
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -276,14 +276,18 @@ private struct JSONWriter {
             try serializeString(str)
         case let boolValue as Bool:
             serializeBool(boolValue)
-        case _ where _SwiftValue.store(obj) is NSNumber:
-            try serializeNumber(_SwiftValue.store(obj) as! NSNumber)
+        case let num as Int:
+            writer(String(describing: num))
+        case let num as UInt:
+            writer(String(describing: num))
         case let array as Array<Any>:
             try serializeArray(array)
         case let dict as Dictionary<AnyHashable, Any>:
             try serializeDictionary(dict)
         case let null as NSNull:
             try serializeNull(null)
+        case _ where _SwiftValue.store(obj) is NSNumber:
+            try serializeNumber(_SwiftValue.store(obj) as! NSNumber)
         default:
             throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Invalid object cannot be serialized"])
         }
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -62,13 +62,11 @@ open class JSONSerialization : NSObject {
 
             // object is a Double and is not NaN or infinity
             if let number = obj as? Double  {
-                let invalid = number.isInfinite || number.isNaN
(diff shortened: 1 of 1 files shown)


Foundation/NSJSONSerialization.swift
```diff
@@ -55,14 +55,19 @@ open class JSONSerialization : NSObject {
...
                 return true
             }
 
-            // object is NSNumber and is not NaN or infinity
-            if let number = _SwiftValue.store(obj) as? NSNumber {
-                let invalid = number.doubleValue.isInfinite || number.doubleValue.isNaN
+            // object is a Double and is not NaN or infinity
+            if let number = obj as? Double  {
+                let invalid = number.isInfinite || number.isNaN
+                return !invalid
```
e78l: Would `return number.isFinite` work?

Foundation/NSJSONSerialization.swift
```diff
@@ -271,19 +281,24 @@ private struct JSONWriter {
...
+        // For better performance, the most expensive conditions to evaluate should be last.
         switch (obj) {
         case let str as String:
             try serializeString(str)
         case let boolValue as Bool:
             serializeBool(boolValue)
-        case _ where _SwiftValue.store(obj) is NSNumber:
-            try serializeNumber(_SwiftValue.store(obj) as! NSNumber)
+        case let num as Int:
+            writer(String(describing: num))
```
parkera: Do we really have control over how String decides to write out this number? Wouldn't it be better to keep the exact logic inside the `JSONSerialization` class?

djones6: The current implementation represents the value as an `NSNumber` and then formats it using `CFNumberFormatterCreateStringWithNumber` using a `CFNumberFormatter` created by the `JSONSerialization` class.  It is the CF code (and format string that is passed to the formatter) determines how a number is represented.
However, this is significantly slower than using String.  Do you foresee a risk that the format of an Int / UInt produced by String may change?
An alternative could be to replicate what String is doing within `JSONWriter`. (is this what you had in mind?)

parkera: It's the `String(describing:)` output that I'm concerned about, yes. I would rather control the output in code more tightly coupled to the writer itself, since formatting that output exactly is basically the entire job of this class.

djones6: This isn't as straightforward as I'd expected, and I need to give it some more thought.  In the meantime, I'll revert the fast paths from this PR as I'd like to deliver the 20% improvement from case reordering.

djones6:
@parkera I've removed the Int/UInt fastpaths for now, to be submitted in a future PR, as I'd like to deliver the 20% improvement from case reordering.  Are you okay with this approach?

parkera:
@swift-ci please test

djones6:
@parkera Is there anything else needed for this to be merged?

parkera:
This should be good to go. We should also measure again once we can get #734 landed.

parkera:
@swift-ci test and merge

PR #725, opened 2016-11-28
Remove executable bit from swift and project files

alblue:
A number of files had executable bits set inadvertently. Remove the
executable bit for these files.


parkera:

PR #728, opened 2016-11-29
Fix tests when running in Darwin

alblue:
When running on Darwin, the /private/ is not stripped from the
URL and so the test fails. Change the logic so that this is only
run when running on Linux platforms.


parkera:

PR #731, opened 2016-12-01
Clean up `URLResourceValues`

JaSpa:
Analog to [831ad8050c22c1b1573ed610cda7fdc7137a9ee1](https://github.com/apple/swift/pull/5996/commits/831ad8050c22c1b1573ed610cda7fdc7137a9ee1) from apple/swift#5996

• The ivar `_keys` is only ever set, never queried, the private function
  `contains` isn't called from anywhere

• The bridging machinery can handle String?/[String]?/Bool?/... => Any?
  There is no need for specialized functions.

### DIFF
--- a/Foundation/URL.swift
+++ b/Foundation/URL.swift
@@ -19,22 +19,15 @@
  */
 public struct URLResourceValues {
     fileprivate var _values: [URLResourceKey: Any]
-    fileprivate var _keys: Set<URLResourceKey>
-    
+
     public init() {
         _values = [:]
-        _keys = []
     }
-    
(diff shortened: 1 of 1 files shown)


parkera:
Bridging works totally differently on Linux than Darwin. Are you sure this works the same way on Linux?

parkera:
For example, `SwiftValue.mm` is not even built on Linux, as it requires the Objective-C runtime.

Probably the NSNumber conversions are still required.

JaSpa:
Ah sorry, you're right. I just wanted to keep Foundation and corelibs-Foundation from diverging. I'll take another look!

alblue:

PR #735, opened 2016-12-08
Correct behavior for `NSMutableData`'s `replaceBytes(in:withBytes:length:)` method

Jumhyn:
# Explanation
Previously, `replaceBytes(in:withBytes:length: 0)` was a no-op when called as `data.replaceBytes(in: range, withBytes: nil length: 0)`. This PR corrects this, and calls the underlying `CFDataReplaceBytes` in all cases. When `replacementBytes` is `nil` and `length` is `0`, this deletes the bytes in `range`.

# Scope
The scope of this change is limited to the single method noted above, and should be fully source compatible--only the internal implementation is changing.

# SR Issue
[SR-3404](https://bugs.swift.org/browse/SR-3404)

# Risk
There should be no risk to taking this change. Any code which relied on this call being a no-op did so in direct contradiction with the documentation.

# Testing
A new test is also added to ensure that this method works as specified.

(Documentation taken from the [NSMutableData API Reference](https://developer.apple.com/reference/foundation/nsmutabledata/1412428-replacebytes))


Jumhyn:
@parkera Is this a reasonable change to include in Swift 3.1?

parkera:
I think so; I just need to review this. If we merge to master in the next few weeks it'll be in Swift 3.1.

johnno1962:
This was picked up in the end by https://github.com/apple/swift-corelibs-foundation/pull/905. I think you can close this.

alblue:
Yup, I concur.

PR #736, opened 2016-12-10
Change type of threadDictionary objects to Any

sclukey:
On macOS the threadDictionary is an NSMutableDictionary, so the objects are of type Any. For consistency, it seems this should also use Any instead of AnyObject.

I stumbled upon this when looking at making [this Regex library](https://github.com/sharplet/Regex) work on Linux. [This particular line](https://github.com/sharplet/Regex/blob/27b3aab37a6935ef7370dd06b4e47e932621e94e/Source/Regex.swift#L147) along with [this one](https://github.com/sharplet/Regex/blob/27b3aab37a6935ef7370dd06b4e47e932621e94e/Source/ThreadLocal.swift#L27) fails because `MatchResult` is a `struct` so it cannot be converted to `AnyObject`.


parkera:

PR #738, opened 2016-12-11
SR-3385 and SR-3391: NumberFormatter - update default attributes for decimalStyle

kweinmeister:
This PR addresses two issues that were recently raised while using the decimal style of NumberFormatter: [SR-3391](https://bugs.swift.org/browse/SR-3391) and [SR-3385](https://bugs.swift.org/browse/SR-3385). I found discrepancies in the behavior using Foundation vs what I observed in an Xcode Playground:
* The leading 0 in numbers was being dropped, as in .03 (vs 0.03 in the Playground).
* Large numbers like 987654321 were being formatted as 987,654,000 (vs 987,654,321 in the Playground).

I added a new case for `.decimal` that set attribute values to match what I observed in the NumberFormatter decimal style in the playground. I also added the above scenarios into existing test cases.

### DIFF
--- a/Foundation/NSNumberFormatter.swift
+++ b/Foundation/NSNumberFormatter.swift
@@ -192,6 +192,12 @@ open class NumberFormatter : Formatter {
                 _usesGroupingSeparator = true
                 _minimumFractionDigits = 2
                 
+            case .decimal:
+                _usesSignificantDigits = false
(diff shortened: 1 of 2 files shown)


parkera:
cc @lifeissweetgood

parkera:
@swift-ci test and merge

PR #739, opened 2016-12-11
Removing some NS prefixes from folders, files and documentation

hartbit:
Removing some NS prefixes from folders, files and documentation to fit with the new class names and make it easier to navigate the documentation and project structure.


hartbit:
@parkera Can I continue this work or if there a reason the NS prefixes were kept in the filenames and documentation?

parkera:
Thanks, I think this is a good idea to make things clearer.

One note: things like `class NSData` should stay in `NSData.swift`, because we have both the `class` and the `struct`.

parkera:
It would be good to merge this in fairly small chunks, so that we don't introduce too many merge conflicts. So if you want to merge this one then continue in other PRs, we can do that.

ianpartridge:
This PR has gone stale, unfortunately. Some of these renames have already happened in https://github.com/apple/swift-corelibs-foundation/pull/1058

@hartbit can you update this PR to fix any remaining issues? Or is it so stale you'd prefer to just close it?

hartbit:
@ianpartridge I forgot I had this PR :D I can have a look this evening!

hartbit:
I re-applied the PR on the latest master, co closing in favour of #1146

PR #742, opened 2016-12-12
Fix build error on s390x

vivkong:
HALT is only defined on i386 and x86_64 with recent changes.  This breaks the build on s390x.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -171,14 +171,16 @@ CF_PRIVATE CFIndex __CFActiveProcessorCount();
 #define __builtin_unreachable() do { } while (0)
 #endif
 
-#if defined(__i386__) || defined(__x86_64__)
-    #if defined(__GNUC__)
+#if defined(__GNUC__) 
+    #if defined(__i386__) || defined(__x86_64__)
         #define HALT do {asm __volatile__("int3"); kill(getpid(), 9); __builtin_unreachable(); } while (0)
-    #elif defined(_MSC_VER)
-        #define HALT do { DebugBreak(); abort(); __builtin_unreachable(); } while (0)
     #else
-        #error Compiler not supported
+        #define HALT do {__builtin_trap(); kill(getpid(), 9); __builtin_unreachable(); } while (0)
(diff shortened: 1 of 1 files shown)


CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -171,9 +171,13 @@ CF_PRIVATE CFIndex __CFActiveProcessorCount();
 #define __builtin_unreachable() do { } while (0)
 #endif
 
-#if defined(__i386__) || defined(__x86_64__)
-    #if defined(__GNUC__)
-        #define HALT do {asm __volatile__("int3"); kill(getpid(), 9); __builtin_unreachable(); } while (0)
+#if defined(__i386__) || defined(__x86_64__) || defined(__s390x__)
```
parkera: We have a few architecture checks across CF but we don't usually check for s390x there. Is neither i386 nor x86_64 defined on that platform?

vivkong: i386 and x86_64 are not defined on s390x.

Perhaps instead of doing a check on s390x, we can change the code to check like this?

```
#if defined(__GNUC__) 
        #if defined(__i386__) || defined(__x86_64__)
          #define HALT do {asm __volatile__("int3"); kill(getpid(), 9); __builtin_unreachable(); } while (0)
        #else
          #define HALT do {__builtin_trap(); kill(getpid(), 9); __builtin_unreachable(); } while (0)
        #endif
    #elif defined(_MSC_VER)
        #define HALT do { DebugBreak(); abort(); __builtin_unreachable(); } while (0)
    #else
        #error Compiler not supported
    #endif
```

parkera: Ok - my main question is, though: what happens on s390x in all of the other places we switch on architecture?

vivkong: It looks like there is always a fall back when the architecture is not i386/x86_64 so we can continue on s390x.  Need to take a look and see if there's work to be done for s390x in those places.  In the mean time, can we please fix this so our s390x build can continue?  Thanks.

parkera:

PR #743, opened 2016-12-12
downloadTask(withResumeData: ...) signature fixed

glessard:
It should have the same completion handler type as the other two `downloadTask` methods


pushkarnk:
Thanks!

@swift-ci please test

PR #744, opened 2016-12-13
[SR-3410] String.hasPrefix and String.hasSuffix returns false if string pattern…

sshamaia:
… is empty string

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -1432,9 +1432,12 @@ extension String {
         let range = CFRangeMake(0, CFStringGetLength(cfstring))
         let opts = CFStringCompareFlags(
             kCFCompareAnchored | kCFCompareNonliteral)
-        
-        return CFStringFindWithOptions(cfstring, prefix._cfObject,
+        if prefix.isEmpty {
+            return true
+        } else {
+            return CFStringFindWithOptions(cfstring, prefix._cfObject,
                                        range, opts, nil)
+        } 
     }
     
     public func hasSuffix(_ suffix: String) -> Bool {
@@ -1442,8 +1445,12 @@ extension String {
         let range = CFRangeMake(0, CFStringGetLength(cfstring))
         let opts = CFStringCompareFlags(
             kCFCompareAnchored | kCFCompareBackwards | kCFCompareNonliteral)
-        return CFStringFindWithOptions(cfstring, suffix._cfObject,
+        if suffix.isEmpty {
+            return true
+        } else {
+            return CFStringFindWithOptions(cfstring, suffix._cfObject,
                                        range, opts, nil)
+        }    
     }
 }
 #endif
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -90,6 +90,7 @@ class TestNSString : XCTestCase {
             ("test_deletingPathExtension", test_deletingPathExtension),
             ("test_ExternalRepresentation", test_ExternalRepresentation),
             ("test_mutableStringConstructor", test_mutableStringConstructor),
+            ("test_emptyStringPrefixAndSuffix",test_emptyStringPrefixAndSuffix),
             ("test_PrefixSuffix", test_PrefixSuffix),
             ("test_utf16StringRangeCount", test_StringUTF16ViewIndexStrideableRange),
             ("test_reflection", { _ in test_reflection }),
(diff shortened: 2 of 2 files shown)


parkera:
I don't think we want to change `CFString.c` here, but probably we need to change the definition of `hasPrefix` and `hasSuffix` in `NSString.swift`.

On Darwin, `NSString` says `NO` for `hasPrefix` with an empty string argument.

This is one unfortunate area where the Swift String type went a different direction than NSString, even though they are bridged.

TestFoundation/TestNSString.swift
```diff
@@ -1228,7 +1240,7 @@ func checkHasPrefixHasSuffix(_ lhs: String, _ rhs: String, _ stack: [UInt]) -> I
 
 extension TestNSString {
     func test_PrefixSuffix() {
-#if !_runtime(_ObjC)
+//#if !_runtime(_ObjC)
```
sshamaia: Through  this PR we are aligning the Linux behaviour with Darwin platform.Hence uncommenting  this as it is applicable to all platforms

TestFoundation/TestNSString.swift
```diff
@@ -1183,17 +1191,21 @@ enum Stack: Swift.Error {
 }
 
 func checkHasPrefixHasSuffix(_ lhs: String, _ rhs: String, _ stack: [UInt]) -> Int {
-    if lhs == "" {
+    if (lhs == "" && rhs == "") {
```
sshamaia: Modified method checkHasPrefixHasSuffix  to handle the case where lhs and rhs are both empty strings. Modified the conditions to report failures only if hasPrefix and hasSuffix returns incorrect value.

TestFoundation/TestNSString.swift
```diff
@@ -1086,7 +1093,8 @@ let comparisonTests = [
     // ASCII cases
     ComparisonTest("t", "tt"),
     ComparisonTest("t", "Tt"),
-    ComparisonTest("\u{0}", ""),
+    ComparisonTest("\u{0}", "",
```
sshamaia: Due to above change to checkHasPrefixHasSuffix ComparisonTest("\u{0}", "") fails on linux  because of https://bugs.swift.org/browse/SR-332 so modified this test with this reason.

sshamaia:
@parkera Thanks for the suggestion. I have made suggested changes , please review

Foundation/NSString.swift
```diff
@@ -1432,18 +1432,27 @@ extension String {
         let range = CFRangeMake(0, CFStringGetLength(cfstring))
         let opts = CFStringCompareFlags(
             kCFCompareAnchored | kCFCompareNonliteral)
-        
-        return CFStringFindWithOptions(cfstring, prefix._cfObject,
+        let prefix_len = CFStringGetLength(prefix._cfObject)
```
parkera: This would be far more efficient like this:

```
if prefix.isEmpty {
    return true
} else {
    return CFStringFindWithOptions(...)
```

Foundation/NSString.swift
```diff
@@ -1432,18 +1432,27 @@ extension String {
...
+        } 
     }
     
     public func hasSuffix(_ suffix: String) -> Bool {
         let cfstring = self._cfObject
         let range = CFRangeMake(0, CFStringGetLength(cfstring))
         let opts = CFStringCompareFlags(
             kCFCompareAnchored | kCFCompareBackwards | kCFCompareNonliteral)
-        return CFStringFindWithOptions(cfstring, suffix._cfObject,
+        let suffix_len = CFStringGetLength(suffix._cfObject)
```
parkera: Same comment here.

TestFoundation/TestNSString.swift
```diff
@@ -1228,7 +1240,7 @@ func checkHasPrefixHasSuffix(_ lhs: String, _ rhs: String, _ stack: [UInt]) -> I
 
 extension TestNSString {
     func test_PrefixSuffix() {
-#if !_runtime(_ObjC)
```
parkera: Let's just delete the guard instead of commenting it out.

Foundation/NSString.swift
```diff
@@ -1432,18 +1432,25 @@ extension String {
         let range = CFRangeMake(0, CFStringGetLength(cfstring))
         let opts = CFStringCompareFlags(
             kCFCompareAnchored | kCFCompareNonliteral)
-        
-        return CFStringFindWithOptions(cfstring, prefix._cfObject,
+        if prefix.isEmpty {
```
sshamaia: Addressed the review comments

Foundation/NSString.swift
```diff
@@ -1432,18 +1432,25 @@ extension String {
...
+        } 
     }
     
     public func hasSuffix(_ suffix: String) -> Bool {
         let cfstring = self._cfObject
         let range = CFRangeMake(0, CFStringGetLength(cfstring))
         let opts = CFStringCompareFlags(
             kCFCompareAnchored | kCFCompareBackwards | kCFCompareNonliteral)
-        return CFStringFindWithOptions(cfstring, suffix._cfObject,
+        if suffix.isEmpty {
```
sshamaia: Addressed the review comments

sshamaia: @parkera please review

parkera:

parkera:
Thanks!

pushkarnk:
Looks like the merge command didn't work. Retrying. 

PR #747, opened 2016-12-15
Cast a function argument in order to disambiguate the intended overload.

rudkx:
An upcoming change in the type checker will result in this call to
fetch() becoming ambiguous, so we need to explicitly cast the argument
to make it unambiguous.

The overloads for fetch are:
  fetch(_ object: AnyObject?) -> Any?
  fetch(_ object: AnyObject) -> Any

but the current call here is taking an AnyObject and returning it as an
Any? which means that both overload options would result in a
value-to-optional conversion. A current quick of the type checker
disambiguates this by chosing the version that returns an optional, but
that quirk will go away (at least under -swift-version 4).


rudkx:
@swift-ci Please test

rudkx:
@parkera I don't expect this to be at all controversial but just wanted to run it by you.

parkera:
LGTM

parkera:

PR #748, opened 2016-12-16
[NSThread] Implement main and isMainThread

antonmes:
All 4 tests pass on Linux but i didn't enable it because i don't know the reason they were excluded.

### DIFF
--- a/Foundation/NSThread.swift
+++ b/Foundation/NSThread.swift
@@ -8,10 +8,16 @@
 //
 
 
-#if os(OSX) || os(iOS)
+#if os(macOS) || os(iOS)
 import Darwin
-#elseif os(Linux) || CYGWIN
+#elseif os(Linux) || os(Android) || CYGWIN
 import Glibc
+private func pthread_main_thread_np() -> pthread_t {
+    return _CF_pthread_main_thread_np()
+}
+private func pthread_main_np() -> Int32 {
+    return _CFIsMainThread() ? 1 : 0
+}
 #endif
 
 import CoreFoundation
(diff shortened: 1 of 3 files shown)


Foundation/NSThread.swift
```diff
@@ -57,18 +57,30 @@ private func NSThreadStart(_ context: UnsafeMutableRawPointer?) -> UnsafeMutable
 }
 
 open class Thread : NSObject {
-    
+    static internal let _mainThread = Thread(thread: pthread_self())
```
parkera: When is this initializer actually called? Is it lazy, so the first to ask for it will be implicitly main?

antonmes: I have the same question, but test_mainThreadFirstAccess pass so i decided it's ok.
Is there a better way to guarantee initialization on the main thread?

parkera: On Darwin we have a pthread 'non-posix' API to get a handle on the main thread. I would be looking for something similar on other platforms pthread implementations.

antonmes: I didn't find any api for this, only suggestions to use pthread_self in the main function.
And i found that Linux [kernel tests](https://github.com/torvalds/linux/blob/5924bbecd0267d87c24110cbe2041b5075173a25/tools/testing/selftests/mqueue/mq_perf_tests.c#L545) use this technique as well. 

I checked current implementation with a complete toolchains build and this code
```swift
import Foundation
Thread {
    print(Thread.main == Thread.current)
}.start()
sleep(1)
```
prints `false`, so the expected behaviour is correct.

antonmes:
```swift
import Foundation
Thread {
    print(Thread.isMainThread)
    print(Thread.main == Thread.current)
    print(Thread.main.isMainThread)
}.start()

sleep(1)

print(Thread.isMainThread)
print(Thread.main == Thread.current)
print(Thread.main.isMainThread)
```
```bash
false
false
true
true
true
true
```
in both debug and release mode

parkera: Try dispatch_async and calling Thread.main in there.

antonmes:
```swift
DispatchQueue.global().async {
...
}
```
has the same output

parkera: Interesting, thanks.

phausler: perhaps for safety sake we could call this in __CFInitializeSwift to ensure the proper initialization? however that does make the assumption the library is loaded on the main thread (Android can break this rule)

antonmes: @phausler 👍  following your suggestion i found that CF already has pthread_main_np shim implemented [here](https://github.com/apple/swift-corelibs-foundation/blob/173e9ea16cf1a0ed9b21cc11ea8a2d15b3f8dca9/CoreFoundation/Base.subproj/CFRuntime.c#L982)

Foundation/NSThread.swift
```diff
@@ -186,10 +199,11 @@ open class Thread : NSObject {
 #endif
     }
     
-    open func main() {
-        _main()
+    // FIXME: should be a functiun but conflicts with the `class var main` propery
```
paulofaria: typo in "functiun".. shouldn't it be function?

ianpartridge:
This PR has gone stale and the functionality was merged under #1162, so I'm going to close it.

PR #751, opened 2016-12-19
[UUID] added unit tests 

Kirin-Chan:
Tested on macOS 10.12.2 / Xcode 8.2 and Ubuntu 16.10

### DIFF
--- a/TestFoundation/TestUUID.swift
+++ b/TestFoundation/TestUUID.swift
@@ -1,0 +1,302 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+
+#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
+    import Foundation
+    import XCTest
+#else
+    import SwiftFoundation
+    import SwiftXCTest
+#endif
+
+
+
+class TestUUID: XCTestCase {
+
+    static var allTests: [(String, (TestUUID) -> () throws -> Void)] {
+        return [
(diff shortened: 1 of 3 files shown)


TestFoundation/TestUUID.swift
```diff
@@ -0,0 +1,302 @@
...
+        XCTAssertEqual(children.count, 0)
+    }
+    
+    func test_bridgeToObjectiveC() {
+        let uuid    = UUID(uuid: test)
+        let bridged = uuid._bridgeToObjectiveC()
+        
+        XCTAssertNotNil(bridged)
+        XCTAssertTrue((bridged as Any) is NSUUID)
+        XCTAssertFalse((bridged as Any) is UUID)
```
phausler: this line fails when running this test on Darwin (not in swift-corelibs-foundation) so it probably means there is a bug somewhere.

TestFoundation/TestUUID.swift
```diff
@@ -0,0 +1,302 @@
...
+        let uuid    = NSUUID(uuidString: testString)
+        let bridged = UUID._unconditionallyBridgeFromObjectiveC(uuid)
+        var bridgedConditional : UUID? = nil
+        let _ = UUID._conditionallyBridgeFromObjectiveC(uuid!, result: &bridgedConditional)
+        
+        XCTAssertNotNil(bridged)
+        XCTAssertNotNil(bridgedConditional)
+        XCTAssertTrue((bridged as Any) is UUID)
+        XCTAssertTrue((bridgedConditional as Any) is UUID)
+        XCTAssertFalse((bridged as Any) is NSUUID)
```
phausler: This also fails on the objective-c foundation so perhaps there is a failure somehow in the bridge code for these? or is the assert perhaps bad?

TestFoundation/TestUUID.swift
```diff
@@ -0,0 +1,302 @@
...
+        XCTAssertFalse((bridgedConditional as Any) is NSUUID)
```
phausler: ditto here on the Darwin failure

Kirin-Chan: I will take a look at it.

alblue:

PR #752, opened 2016-12-19
More fixes for potentially ambiguous function calls.

rudkx:
This is the same issue that was fixed in one place in
d9cdc7ecdb91fdfd738b7347c69a6fa3b6d5d2be and another place in
15769584ce4b4e450e893f33a550e2c23f55db30.

Hopefully this is the last of these fixes.

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -226,7 +226,7 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
         for idx in 0..<count {
             let item = _SwiftValue.store(self[idx])
             if set.contains(item) {
-                return _SwiftValue.fetch(item)
+                return _SwiftValue.fetch(item as AnyObject?)
             }
         }
         return nil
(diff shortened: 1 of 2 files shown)


rudkx:
@swift-ci Please smoke test and merge

rudkx:

Foundation/NSArray.swift
```diff
@@ -226,7 +226,7 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
         for idx in 0..<count {
             let item = _SwiftValue.store(self[idx])
```
phausler: item here is a NSObject which is an object already; this should not bee need to be cast to `AnyObject?`

rudkx: The cast to `AnyObject?` is needed because this function (`firstObjectCommon`) returns an `Any?`, and there are two forms of `fetch`, one that takes an `AnyObject` and returns `Any`, and one that takes an `AnyObject?` and returns `Any?`. Without the cast, neither version of `fetch` is a better choice than the other because they both result in an implicit coercion to an optional (in one case for the parameter, in the other case for the return value).

This currently compiles without the cast due to a bug in the type checker, but I'm going to be landing a change to the type checker.

rudkx:
Closing out and looking at adding argument labels instead of casts to disambiguate.

PR #754, opened 2016-12-20
Disable TestURLSession pending SR-3464.

gparker42:


gparker42:
@swift-ci please test linux

parkera:
This is a very large set of tests to disable. Can we be more specific instead?

gparker42:
I only know that the first test fails; the others didn't run in CI after that. Narrowing this would take either a few hours watching CI or a local reproducer, and I didn't have either.

PR #756, opened 2016-12-22
Fix comment for parameters

uny:
Invalid syntax in `parameter` comment for quick help.
screen shot 2016-12-22 at 18 04 08


parkera:
Cool, thanks. We should fix up `Calendar.swift` in the swift overlay as well.

PR #757, opened 2016-12-28
[SE-0039] [SR-3504] Require file literal argument to be a static string

xwu:
A coordinate change for apple/swift#6503. See that PR for discussion.


xwu:
Closing this PR because an alternative solution is to be used to resolve SR-3504.

PR #758, opened 2017-01-02
Android SDK name changes.

gonzalolarralde:
* SDK name added for Android. The OSType remains as linux, but the platform specific files should be retrieved from the `android` directory instead of `linux`.
* conformance_begin/end was using a hardcoded `linux` directory. Now it's pulling `swift_being.o`/`swift_end.o` object files from the correct directory. 

This last point should probably be double checked in FreeBSD. I'll try to set up an environment for testing, but is there anyone I can reach to confirm this last point?


parkera:
@swift-ci please test

PR #649, opened 2016-09-20
String: prevent executable stacks on GNU targets

compnerd:
The use of the hand written assembly files could potentially introduce
executable stacks.  Since these files do not require the executable stacks,
ensure that we add the special marker to indicate that executable stacks are not
required.

PR #759, opened 2017-01-06
Fix libFoundation.so being flagged as requiring executable stack

yorickdowne:
libFoundation.so is flagged as requiring an executable stack on Linux, although it does not. This is due to GNU as behavior.
Please see discussion at https://bugs.swift.org/browse/SR-2700?jql=text%20~%20%22executable%20stack%22


yorickdowne:
On reflection, I should instead figure out why NO_EXEC_STACK_DIRECTIVE in CoreFoundation/Base.subproj/CFAsmMacros.h isn't doing what it's supposed to do. See https://github.com/apple/swift-corelibs-foundation/pull/649/files 

Closing this request as it's not the best way to solve the problem.

PR #760, opened 2017-01-06
[stdlib] Implement SE-147: Move initialize(from:) from Pointer to BufferPointer

airspeedswift:
This is a coordinated change with [this PR](https://github.com/apple/swift/pull/6601) on the Swift repo.

Since `UnsafeMutablePointer.initialize(from:)` is memory-unsafe with collections that underreport their count, it is being migrated to be on `UnsafeMutableBufferPointer` to avoid buffer overruns.

As part of this, `Sequence. _copyContents(initializing:)` is being altered to take a buffer not a pointer. This is an internal underscore protocol so not user-impacting, but corelibs-foundation does use it and needs to change at the same time.

### DIFF
--- a/Foundation/Data.swift
+++ b/Foundation/Data.swift
@@ -1436,7 +1436,11 @@ public struct Data : ReferenceConvertible, Equatable, Hashable, RandomAccessColl
                 }
                 
                 if replacementCount != 0 {
-                    newElements._copyContents(initializing: bytes + start)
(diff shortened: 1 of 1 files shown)


airspeedswift:
@parkera @phausler are you OK with making this change? (needs to be done simultaneously on both versions of Data for us to change it in the std lib)

parkera:
Looks good to me.

PR #762, opened 2017-01-06
[FileManager] Use Error to be consistent with `enumerator(at:includin…

ikesyo:
…gPropertiesForKeys:options:errorHandler:)`


parkera:

PR #763, opened 2017-01-06
Fix libFoundation.so being flagged as requiring executable stack

yorickdowne:
libFoundation.so is flagged as requiring an executable stack on Linux, although it does not. This is due to GNU as behavior.
Please see discussion at https://bugs.swift.org/browse/SR-2700?jql=text%20~%20%22executable%20stack%22

This was already intended to be solved by https://github.com/apple/swift-corelibs-foundation/pull/649/, but did not work as intended in Ubuntu. Checking for \_\_GNUC\_\_ in addition resolves the issue.


parkera:
Thanks!

parkera:

PR #764, opened 2017-01-06
Correct sequence initializers for Data when repeating:count: is called and add a memset fast-path

phausler:
Resolves SR-3566.


phausler:
@swift-ci Please test

phausler:
@parkera we could perhaps make a `private init(uninitializedCount: Int)` method to facilitate the memset so we are not paying the cost of a malloc + bzero + memset and just paying malloc + memset. Thoughts?

parkera:
Seems fine to me if it's private.

PR #765, opened 2017-01-07
[URLSessionStreamTask] Update `readData` and `write` methods to match the Darwin version

ikesyo:
- https://developer.apple.com/reference/foundation/urlsessionstreamtask/1411604-readdata
- https://developer.apple.com/reference/foundation/urlsessionstreamtask/1411602-write


parkera:

PR #766, opened 2017-01-07
[SR-3403] Rename a few URLSession related constants to match the Darwin version

ikesyo:
Fixes [SR-3403](https://bugs.swift.org/browse/SR-3403).

- Rename `URLSessionTransferSizeUnknown` to `NSURLSessionTransferSizeUnknown`
- Rename `URLSessionDownloadTaskResumeData` to `NSURLSessionDownloadTaskResumeData`
- Implement `NSURLSessionDownloadTaskResumeData`


parkera:
This doesn't seem right. I think those should probably have the NS dropped on Darwin. cc @phausler

ikesyo:
Okay, I withdraw this and hope that it will be fixed on Darwin in Swift 3.1.

PR #767, opened 2017-01-08
Enable building on powerpc

ruscur:
These changes enable building on the powerpc family of targets - powerpc (32bit BE), powerpc64 (64bit BE) and powerpc64le (64bit LE).  Only tested on powerpc64le since that's the only target properly supported in Swift overall.


parkera:

PR #769, opened 2017-01-09
URLSessionTaskPriority... constants should be class properties

ikesyo:
https://developer.apple.com/reference/foundation/urlsessiontask/1660699-url_session_task_priority


parkera:

PR #770, opened 2017-01-09
[URLSessionTask] Use `DispatchQueue.sync` overload which has a return value

ikesyo:


parkera:

PR #777, opened 2017-01-10
[NotificationQueue] Rename Coalescing constants to match the Darwin version

ikesyo:
`Coalescing` itself should be renamed as well in a separate pull request.


parkera:
Thanks for all these fixes @ikesyo

parkera:
@swift-ci test and merge

PR #780, opened 2017-01-10
[gardening] Remove unnecessary `break`s

ikesyo:


parkera:
Could we also apply the changes to the value type classes to the overlay in the Swift project as well?

parkera:

ikesyo:
> Could we also apply the changes to the value type classes to the overlay in the Swift project as well?

I will tackle it.

ikesyo:
https://github.com/apple/swift/pull/6735

PR #782, opened 2017-01-11
[NotificationQueue] Rename Coalescing to NotificationCoalescing to match the Darwin version

ikesyo:
(Renaming `NotificationCoalescing` to `Coalescing` on Darwin might be an option but I'm not sure)


parkera:

PR #788, opened 2017-01-12
[plutil][gardening] Use `switch` for type casting

ikesyo:
This would be more readable.


parkera:
@swift-ci test and merge

PR #794, opened 2017-01-14
[NSError] Use AnyHashable for userInfo consistently

ikesyo:


parkera:
Hm, there are certainly places where we intentionally wanted to restrict user info keys to `String`. cc @phausler

PR #796, opened 2017-01-14
Fix for [SR-3448] Crash in String.replacingOccurrences(of:with:) with…

abl:
… emoji on Linux

Foundation.NSMutableString.replaceCharacters assumed that the passed-in NSRange was character-based but it's UTF-16 based.

The fix was to use the UTF-16 view and then convert the indexes back to characters.

---

All credit to @allevato for coming up with the fix.

This also resolves [SR-3449].


parkera:
Thanks!

parkera:
@swift-ci test and merge

PR #799, opened 2017-01-15
[NotificationQueue] Add missing `none` to NotificationCoalescing

ikesyo:
https://developer.apple.com/reference/foundation/notificationqueue.notificationcoalescing/1415826-none


parkera:
I think this API probably goes against our usual conventions of not providing a 'none' option (just pass empty), but we should have this in the name of source compatibility.

parkera:
@swift-ci test and merge

PR #638, opened 2016-09-15
Implemented _resolveCurrent

amraboelela:
This is dependent on https://github.com/apple/swift/pull/4788

PR #803, opened 2017-01-18
Fix building NSHost.swift on macOS

ikesyo:
Related to #638.

- Import `<ifaddrs.h>` in `SwiftRuntime/CoreFoundation.h`
- Use `sa_family_t` which is `UInt8` on macOS instead of explicit `UInt16`


parkera:
@swift-ci test and merge

briancroom:
@parkera Would you be ok with pulling this fix onto `swift-3.1-branch`?

(See https://github.com/apple/swift-corelibs-xctest/pull/185#issuecomment-274221790)

cc @ikesyo

parkera:
Yup, that is fine with me (#818).

PR #807, opened 2017-01-18
[NSThread] Implement name property

TheCodez:


parkera:
On Darwin, this calls a pthread non-posix API (pthread_setname_np) to give the name to the pthread library for debugging purposes. Is there an equivalent we can use on other platforms?

TheCodez:
That was my initial plan for the implementation, but seeing Darwins `pthread_setname_np `and `pthread_getname_np `not having a `pthread_t `parameter I wasn't quite sure how to implement it correctly.

Linux equivalents are:
```
int pthread_setname_np(pthread_t thread, const char *name);
int pthread_getname_np(pthread_t thread, char *name, size_t len);
```

parkera:
On Darwin, the current thread is the implicit argument. We actually only set the name in pthread if the current thread is the same as the `NSThread` receiver.

TheCodez:
@parkera Implemented your suggestion.

parkera:
Cool! Is it possible to get a test case along with this?

TheCodez:
@parkera somehow `pthread_setname_np `and `pthread_getname_np `aren't found on Ubuntu? Do you have any ideas?

parkera:
I believe those are darwin-only extensions. Is there anything equivalent in the pthread library on used on ubuntu?

TheCodez:
@parkera I got it working on Ubuntu, the only problem was they weren't found in Swift. Calling the function form CoreFoundation fixed the problem.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1310,6 +1310,14 @@ _CFThreadRef _CFThreadCreate(const _CFThreadAttributes attrs, void *_Nullable (*
     return thread;
 }
 
+void _CFThreadSetName(const char *_Nullable name) {
```
parkera: Let's decorate this with `CF_SWIFT_EXPORT`

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -298,6 +298,8 @@ typedef pthread_t _CFThreadRef;
 
 CF_EXPORT _CFThreadRef _CFThreadCreate(const _CFThreadAttributes attrs, void *_Nullable (* _Nonnull startfn)(void *_Nullable), void *restrict _Nullable context);
 
+CF_EXPORT void _CFThreadSetName(const char *_Nullable name);
```
parkera: Here too

TheCodez:
Done

TheCodez:
Should I add a test to the test case, to see if the pthread name is set correctly?

parkera:
Sure, that'd be useful.

TheCodez:
Should I use `const char*` or `CFStringRef` in CF, e.g.

`CF_SWIFT_EXPORT const char* _CFThreadGetName();`
or:
`CF_SWIFT_EXPORT CFStringRef  _CFThreadGetName();`

parkera:
Generally we like to use the CF types for CF API. So `CFStringRef`

TheCodez:
Sorry, I couldn't quite get `CFStringRef `to work right so I used the other approach, hope thats fine. Also I have one question: Why should those function be decorated with `CF_SWIFT_EXPORT`, when all other `_CFThread `functions are marked with `CF_EXPORT`?

TheCodez:
@parkera could you test?

parkera:
Sure, sorry about that.

parkera:
@swift-ci please test

TheCodez:
@parkera I did some small improvements and cleanups to the test code. Could you review and test, please?

parkera:

PR #829, opened 2017-01-24
[gardening][RunLoopMode] Move operators into types

ikesyo:


Foundation/NSRunLoop.swift
```diff
@@ -33,14 +33,14 @@ public struct RunLoopMode : RawRepresentable, Equatable, Hashable, Comparable {
...
-public func ==(lhs: RunLoopMode, rhs: RunLoopMode) -> Bool {
-    return lhs.rawValue == rhs.rawValue
-}
+    public static func ==(_ lhs: RunLoopMode, _ rhs: RunLoopMode) -> Bool {
+        return lhs.rawValue == rhs.rawValue
+    }
 
-public func <(lhs: RunLoopMode, rhs: RunLoopMode) -> Bool {
-    return lhs.rawValue < rhs.rawValue
+    public static func <(_ lhs: RunLoopMode, _ rhs: RunLoopMode) -> Bool {
```
parkera: @phausler why does run loop mode have a `<` operator in the first place?

phausler: because NS_EXTENSIBLE_STRING_ENUM grants that behavior... I am not certain why that was done (iirc it is at the compiler level)

parkera: It seems like a bug, and unless all of our extensible string enums in scl-foundation have this operator I'd prefer to remove it here.

(if you really do want to compare the strings, then using rawValue with `<` seems reasonable to me)

rintaro: This behavior was introduced in https://github.com/apple/swift/commit/b4a58650dd95262010d5ae2a91be311a4ab3d37d, I believe.

parkera:
We can merge this for now since it's not really about the operator, but I'd like to follow up on that.

parkera:

PR #830, opened 2017-01-24
[SR-3708]PropertyListSerialization parses booleans as CFBoolean

nethraravindran:


TestFoundation/TestNSPropertyList.swift
```diff
@@ -90,4 +91,17 @@ class TestNSPropertyList : XCTestCase {
             XCTFail("value stored is not a string")
         }
     }
+    
+    func test_booleanProperty() {
+        let plistDocString = "<?xml version='1.0' encoding='utf-8'?><!DOCTYPE plist PUBLIC \"-//Apple//DTD PLIST 1.0//EN\" \"http://www.apple.com/DTDs/PropertyList-1.0.dtd\"> <plist version='1.0'><dict><key>state</key><true/></dict></plist>"
+        do {
+            let plistDoc = try XMLDocument(xmlString: plistDocString, options: [])
+            try plistDoc.validate()
+            let plist = try PropertyListSerialization.propertyList(from: plistDoc.xmlData, options: [], format: nil) as! [String: Any]
```
youming-lin: Should use `as? [String: Any]` here to avoid interrupted execution due to force unwrapping of `nil`,

TestFoundation/TestNSPropertyList.swift
```diff
@@ -90,4 +91,17 @@ class TestNSPropertyList : XCTestCase {
...
+            XCTAssertNotNil(plist)
+            XCTAssertEqual(plist["state"] as? NSNumber, 1)
```
youming-lin: and then use `plist?["state"] as? NSNumber` here.

CoreFoundation/Parsing.subproj/CFPropertyList.c
```diff
@@ -2178,7 +2178,8 @@ static Boolean parseXMLElement(_CFXMLPlistParseInfo *pInfo, Boolean *isKey, CFTy
             if (pInfo->skip) {
                 *out = NULL;
             } else {
-                *out = CFRetain(kCFBooleanTrue);
+                SInt32 val = 1;
+                *out = CFNumberCreate(kCFAllocatorSystemDefault, kCFNumberSInt8Type, (const void *)&val);
```
parkera: This change doesn't make a lot of sense to me. `CFPropertyList` is the same on both platforms, so if there is a behavior difference here it is due to bridging - not the creation of booleans in this file.

Also, returning the singleton for booleans is a huge performance advantage vs. creating a new one each time.

nethraravindran:
@parkera From your comments I presume that the change should be done  in Bridging.swift instead of CoreFoundation. So to fix this issue, one of the option would be to include  a check for CFBoolean type objects in ```fetch(nonOptional object: AnyObject) ``` [ Bridiging.swift ]
```
 if object is  __NSCFType{
// code for parsing the CFBoolean to Number  and return NSNumber
}
```
Can you please provide inputs on it.

parkera:
@phausler can you help out with this one?

parkera:
@nethraravindran not all `__NSCFType` are `kCFBoolean`.

nethraravindran:
PR https://github.com/apple/swift-corelibs-foundation/pull/909 fixes this issue. So closing this PR.

PR #846, opened 2017-01-28
Add ICU module map

dabrahams:
This should allow prototypes in the Swift project to import and use ICU facilities


parkera:
Why would swift-corelibs-foundation define a module map for ICU?

parkera:
n.b. that ICU is not public API on Darwin, so defining a module map only for the Linux version of Swift seems odd.

dabrahams:
It's for stdlib implementation, but I found a workaround, so if adding the module map seems like a bad idea to you, we can close this

parkera:
Ok.

PR #847, opened 2017-01-29
copyItem Implementation with Tests.

halfcharged:
Original pull request was #660.  This just adds the tests that were requested.

Don't know if creating a separate pull request was the thing to do here.  Is it possible to modify a pull request if you are not the owner/maintainer??

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -387,7 +387,7 @@ open class FileManager : NSObject {
             }
         } else {
             if createFile(atPath: dstPath, contents: contents(atPath: srcPath), attributes: nil) == false {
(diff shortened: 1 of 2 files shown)


parkera:
This is a good starting point, but it does seem inefficient to read the whole file into memory so we can write it out again. A more efficient implementation would read it in chunks. This one uses an arbitrarily large amount of memory...

parkera:
@swift-ci please test

amraboelela:
Instead of creating a new PR, wasn't it possible to push into the same PR of #660 ?

parkera:
Not sure if everyone has permission to do that; in any case this one has your commits in it still.

amraboelela:
Oh, cool, I see now that you guys merged my commits into this pull request. Thanks :)

PR #850, opened 2017-01-30
Make sure that os_log is turned off for both Darwin and Linux

parkera:
`os_log` calls back into CF on Darwin (for the ability to print `NSObject` types). However, this CF that we're testing in `swift-corelibs-foundation` is not the system one, and the types don't implement the expected forwarding protocols. Better to just disable this on all platforms.

Fixes SR-3703.


parkera:
@swift-ci test and merge

PR #851, opened 2017-01-30
Make sure that os_log is turned off for both Darwin and Linux

parkera:
See also #850. This puts the change in 3.1 branch.


parkera:
@swift-ci test and merge

PR #709, opened 2016-11-08
Import CoreFoundation changes from Sierra

phausler:

PR #855, opened 2017-01-31
Restore use of #include

kastiglione:
A change made in #663 was overwritten by #709. This change restores the switch from `#import` to `#include`.

The reason for this change is: Clang's MSVC compatibility does not handle `#import` statements outside of of objc.

> `#import` of type library is an unsupported Microsoft feature


parkera:
Under which circumstances would `DEPLOYMENT_RUNTIME_SWIFT` be false?

kastiglione:
Recently I merged and noticed this conflict. I put up this pull request to restore the original fix, assuming it was still useful. I see that's not the case, but would like to see it go in to prevent future changes in the conditionals from breaking compilation down the road.

parkera:
@swift-ci test and merge

PR #860, opened 2017-02-02
Implementing NSFileManager.homeDirectoryForCurrentUser and homeDirect…

saiHemak:
…ory(forUser userName: )

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -712,9 +712,15 @@ extension FileManager {
 }
 
 extension FileManager {
-    open var homeDirectoryForCurrentUser: URL { NSUnimplemented() }
+    open var homeDirectoryForCurrentUser: URL {
+        return homeDirectory(forUser: CFCopyUserName().takeRetainedValue()._swiftObject)!
+    }
     open var temporaryDirectory: URL { NSUnimplemented() }
-    open func homeDirectory(forUser userName: String) -> URL? { NSUnimplemented() }
+    open func homeDirectory(forUser userName: String) -> URL? {
+        guard !userName.isEmpty else { return nil }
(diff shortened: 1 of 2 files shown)


Foundation/NSFileManager.swift
```diff
@@ -712,9 +712,15 @@ extension FileManager {
 }
 
 extension FileManager {
-    open var homeDirectoryForCurrentUser: URL { NSUnimplemented() }
+    open var homeDirectoryForCurrentUser: URL {
+        return homeDirectory(forUser: CFCopyUserName().takeRetainedValue()._swiftObject)!
```
parkera: Hm, is a force unwrap the right answer here if we can't find a home directory? Maybe a fallback instead?

OscarSwanros: I wonder if the var declaration could be changed to `open var homeDirectoryForCurrentUser: URL?`. It would make sense for it to be `nil` if it couldn't be obtained by `homeDirectory(forUser:)`

parkera: Looking at CF, it seems the only way for this to totally fail is if even `$HOME` is unset. Perhaps that's an environment error and asserting is the right answer in that case.

parkera:
@swift-ci please test

Foundation/NSFileManager.swift
saiHemak: @parkera  When you say asserting is the right answer do you mean we abort the application like this 
``` 
        let currentUser = CFCopyUserName().takeRetainedValue()._swiftObject
        guard !currentUser.isEmpty  else { fatalError ("$HOME not set") }
        return homeDirectory(forUser: currentUser)!  
```
I agree with @OscarSwanros . I cannot return nil as the declaration says that I cannot return nil.

parkera: I'm saying it's ok as written.

PR #810, opened 2017-01-20
[Error] Implement `localizedDescription` using `_swift_Foundation_getErrorDefaultUserInfo`

ikesyo:
This should return an appropriate value instead of the fixed "The operation could not be completed".

PR #861, opened 2017-02-02
[3.1][Error] Implement `localizedDescription` using `_swift_Foundation_getErrorDefaultUserInfo`

ikesyo:
Cherry picks #810.

This should return an appropriate value instead of the fixed "The operation could not be completed".

### DIFF
--- a/Foundation/NSError.swift
+++ b/Foundation/NSError.swift
@@ -288,7 +288,8 @@ public extension Error where Self : CustomNSError {
 public extension Error {
     /// Retrieve the localized description for this error.
     var localizedDescription: String {
-        return NSError(domain: _domain, code: _code, userInfo: nil).localizedDescription
+        let defaultUserInfo = _swift_Foundation_getErrorDefaultUserInfo(self) as! [String : Any]
+        return NSError(domain: _domain, code: _code, userInfo: defaultUserInfo).localizedDescription
     }
 }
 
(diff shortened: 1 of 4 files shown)


ikesyo:
@swift-ci please test

Foundation/NSError.swift
```diff
@@ -288,7 +288,8 @@ public extension Error where Self : CustomNSError {
 public extension Error {
     /// Retrieve the localized description for this error.
     var localizedDescription: String {
-        return NSError(domain: _domain, code: _code, userInfo: nil).localizedDescription
+        let defaultUserInfo = _swift_Foundation_getErrorDefaultUserInfo(self) as! [String : Any]
```
alblue: I think this should use as? rather than as! because if it's not a [String:Any] then the effect of passing it in will be nil anyway, and that way, you aren't in danger of a further crash in the error handling.

parkera:
Let's take the fix as-is for 3.1 and apply Alex's feedback on master.

ikesyo:
@swift-ci test and merge

Foundation/NSError.swift
philpowers: I'm pretty sure I'm seeing a crash directly related to this. My process is crashing when I try to print an error, and the crash stack has swift_Foundation_getErrorDefaultUserInfo at the top...

parkera: What does your crash look like? Any reproducible steps?

PR #867, opened 2017-02-03
[gardening] Prefer `Array(seq)` over `seq.map { $0 }`

ikesyo:


ikesyo:
@swift-ci please test

pushkarnk:
@ikesyo I'm just curious if there a reason (performance or only elegance?) why we should prefer `Array(seq)` over `seq.map { $0 }`. Thanks.

ikesyo:
@pushkarnk I think that the former would/should be more efficient and the intent would be clearer (and also the latter looks a bit weird to me).

- [`Array.init`](https://github.com/apple/swift/blob/swift-DEVELOPMENT-SNAPSHOT-2017-02-01-a/stdlib/public/core/Arrays.swift.gyb#L1036-L1043)
- [`Sequence.map`](https://github.com/apple/swift/blob/swift-DEVELOPMENT-SNAPSHOT-2017-02-01-a/stdlib/public/core/Sequence.swift#L811-L829)
- [`Collection.map`](https://github.com/apple/swift/blob/swift-DEVELOPMENT-SNAPSHOT-2017-02-01-a/stdlib/public/core/Collection.swift#L1297-L1318)

See also an issue on SwiftLint project: https://github.com/realm/SwiftLint/issues/1271.

parkera:
For the files that are also part of the overlay, let's also raise a PR there to make the same changes.

ikesyo:
Submitted https://github.com/apple/swift/pull/7243.

parkera:
LGTM

ikesyo:
@swift-ci test and merge

PR #870, opened 2017-02-08
[NSRegularExpression] NSCoding implementation

naithar:


Foundation/NSRegularExpression.swift
```diff
@@ -39,11 +39,39 @@ open class NSRegularExpression: NSObject, NSCopying, NSCoding {
     }
     
     open func encode(with aCoder: NSCoder) {
-        NSUnimplemented()
+        guard aCoder.allowsKeyedCoding else {
+            preconditionFailure("Unkeyed coding is unsupported.")
+        }
+        
+        aCoder.encode(self.pattern._nsObject, forKey: "NS.pattern")
```
alblue: Shouldn't this be NSPattern, and not NS.pattern?

Foundation/NSRegularExpression.swift
```diff
@@ -39,11 +39,39 @@ open class NSRegularExpression: NSObject, NSCopying, NSCoding {
...
+        aCoder.encode(self.options.rawValue._bridgeToObjectiveC(), forKey: "NS.options")
```
alblue: I think this should be NSOptions as well ...

Foundation/NSRegularExpression.swift
```diff
@@ -39,11 +39,39 @@ open class NSRegularExpression: NSObject, NSCopying, NSCoding {
...
+        
+        do {
+            try self.init(pattern: pattern._swiftObject, options: Options(rawValue: options.uintValue))
+        } catch {
+            return nil
+        }
+    }
+    
+    open override func isEqual(_ object: Any?) -> Bool {
+        if let other = object as? NSRegularExpression {
```
alblue: Would this be more swifty if it was 'guard let other else { return false } and then have the return self == other at the end of the function?

naithar: That was the way i did the checks on #708 at first. If @parkera doesn't mind, i'll switch first `if` with `guard`.

parkera: Either way is fine; `guard` seems good to me.

naithar:
@parkera I've made requested changes, please review.

alblue:
@swift-ci please test

naithar:
@parkera can we merge this?

parkera:
@swift-ci test and merge

PR #874, opened 2017-02-08
re-enable building without libblocksruntime-dev (Blocks.h)

dgrove-oss:
Add -I./closure to foundation.CFLAGS so that Block.h
can be found from closure/Block.h instead of assuming
that it will be available in the system includes.
This simplifies building Foundation on platforms where
libblocksruntime-dev is not easily available as a
pre-built package.

### DIFF
--- a/CoreFoundation/Error.subproj/CFError.c
+++ b/CoreFoundation/Error.subproj/CFError.c
@@ -13,6 +13,9 @@
 #include "CFInternal.h"
 #include <CoreFoundation/CFPriv.h>
 #include <CoreFoundation/ForFoundationOnly.h>
+#if __BLOCKS__
+#include <Block.h>
+#endif
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED
 #include <mach/mach_error.h>
 #endif
--- a/build.py
+++ b/build.py
@@ -64,6 +64,7 @@ foundation.CFLAGS += " ".join([
 	'-I${SYSROOT}/usr/include/libxml2',
 	'-I${SYSROOT}/usr/include/curl',
 	'-I./',
+	'-I./closure',
 ])
 
 swift_cflags += [


parkera:
Can you build libdispatch without libblocksruntime being present?

My idea of the best long term solution to this mess was to simply put the support for blocks (runtime and the headers) into the swift runtime library. That way everyone is using the same implementation across the stack.

dgrove-oss:
libdispatch also has its own copy of the blocks runtime in src/BlocksRuntime and gets the include of Blocks.h from there (to avoid requiring libblocksruntime-dev to build).  

I agree that where we should end up is having 1 canonical copy of the thing in the swift runtime library and using it everywhere.   

On ubuntu x86_64, its not that important (because we can just apt-get libblocksruntime-dev).  However, I was helping a Kitura user who was trying to build swift for RHEL (no blocks runtime) and was getting blocked by this.  I was also working on getting swift building on ubuntu powerpc64le (also no blocks runtime).  So it seemed worth submitting a small change for the short term.

parkera:
If we do this, then shouldn't we explicitly make libblocksruntime *not* a dependency of CF - so that it builds using the built-in one in all cases?

dgrove-oss:
I think yes, but perhaps we are already doing that (at least as much as we can)?   I don't see anywhere where we explicitly add -lblocksruntime to a link argument.  We do invoke clang/swiftc with -fblocks (which we still need to do).   

I do observe on Linux that if you do ldd on libfoundation.so there isn't a runtime dependency on an external blocks runtime library (_Block_copy et al all get resolved to symbols from closure/runtime.c). 

If you think it is a better use of time, I could look at getting the blocks support into the swift runtime library instead.  Raise on swift-def list for feedback?   Or wait until post Swift 3.1?

spevans:
Given that Foundation links libdispatch, could we not simply remove the private copy in corelibs-foundation? I agree that there should only be one canonical version, but keeping it in libdispatch might be simpler since its already there and has tests setup.

parkera:
@spevans That may be possible. We started with the ability to build without dispatch, but that's no longer the case.

parkera:
Perhaps we just want to keep the header files for simplicity of building.

dgrove-oss:
Agreed it will be easier to have Block.h in swift-corelibs-foundation (until we can move the whole thing into swift runtime proper).  Block.h and Block_private.h in libdispatch are buried down in src/BlockRuntime which isn't part of the public API.  Seems like it is better to add -I./closure to the Foundation build than adding an include into the internals of dispatch's source tree.

ianpartridge:
@dgrove-oss is this PR still valid? I'm looking through old PRs and found it...

dgrove-oss:
I think this is still valid (but haven't tried to build with it).   I've lost track of the likelihood of getting the "right" fix done (which I still think is having a blocks runtime embedded in the Swift runtime that both foundation and dispatch would use; then we would include its Block.h here and in dispatch).

ianpartridge:
Thanks @dgrove-oss 

@parkera what are your current thoughts on this one? Seems like everyone is agreed that the blocks runtime should live in the Swift runtime eventually, but this PR helps people build Foundation on RHEL today.

parkera:
Ok, let's try this as long as it doesn't interfere with the normal path for other platforms.

ianpartridge:
@swift-ci please test

alblue:
Can this be rebased onto master and re-tested? There have been a number of build changes which may have invalidated this approach since it was initially suggested.

alblue:
This pull request has gone stale, and a number of changes have moved on since it was originally submitted. It's not clear whether the approach outlined is still valid.

A newer version of this patch would be welcomed; in particular, conditional inclusion of `Block.h` by being surrounded by `#if __BLOCK` would be a useful change, but in all places that `Block.h` is imported.

PR #876, opened 2017-02-09
A temporary work-around for the memory leak in SR-3536 

nethraravindran:


nethraravindran:
__NSRegularExpressionCache.object(forKey: key._nsObject) in NSString checks for key in Dictionary which holds the 'key reference pointer' as key and the 'NSCacheEntry' as value. Key reference pointer is the reference to NSString Object which can be different for each instance. This is because NSCache on Linux only works with static keys. And Hence we end up having duplicate entries for the same NSString, NSRegularExpression pair causing the memory to leak. So I have disabled the NSCache until a fix for [https://bugs.swift.org/browse/SR-3904] (NSCache working for dynamic keys) is provided.

alblue:
@swift-ci please test

alblue:
I agree; I think this change is necessary until the NSCache implementation can be fixed.

ianpartridge:
Let's get #877 reviewed before merging this? Would be great to just fix the cache.

pushkarnk:
I think this can be closed now.

PR #877, opened 2017-02-09
[NSCache] [SR-3904] Fix for object storage.

naithar:
An attempt to fix an issue with `NSCache` storing every key-value pair that's been set with `.setObject(...)` method. I've tried to implement an Darwin's Foundation behavior as much as possible.
Added some unit tests that check `NSCache` work.
Also changed default values for `.totalCostLimit` and `.countLimit` to match Darwin's Foundation.

Issue: [SR-3904](https://bugs.swift.org/browse/SR-3904)

### DIFF
--- a/Foundation/NSCache.swift
+++ b/Foundation/NSCache.swift
@@ -20,15 +20,45 @@ private class NSCacheEntry<KeyType : AnyObject, ObjectType : AnyObject> {
     }
 }
 
+fileprivate class NSCacheKey: NSObject {
+    
+    var value: AnyObject
+    
+    init(_ value: AnyObject) {
+        self.value = value
+        super.init()
+    }
+    
+    override var hashValue: Int {
+        switch self.value {
+        case let nsObject as NSObject:
+            return nsObject.hashValue
+        case let hashable as Hashable:
+            return hashable.hashValue
+        default: return 0
+        }
+    }
+    
+    override func isEqual(_ object: Any?) -> Bool {
+        guard let other = (object as? NSCacheKey) else { return false }
+        
+        guard let left = self.value as? NSObject,
+            let right = other.value as? NSObject else { return self.value === other.value }
+        
+        return left.isEqual(right)
+    }
+}
+
 open class NSCache<KeyType : AnyObject, ObjectType : AnyObject> : NSObject {
-    private var _entries = Dictionary<UnsafeRawPointer, NSCacheEntry<KeyType, ObjectType>>()
+    
+    private var _entries = Dictionary<NSCacheKey, NSCacheEntry<KeyType, ObjectType>>()
     private let _lock = NSLock()
     private var _totalCost = 0
     private var _byCost: NSCacheEntry<KeyType, ObjectType>?
     
     open var name: String = ""
-    open var totalCostLimit: Int = -1 // limits are imprecise/not strict
-    open var countLimit: Int = -1 // limits are imprecise/not strict
+    open var totalCostLimit: Int = 0 // limits are imprecise/not strict
+    open var countLimit: Int = 0 // limits are imprecise/not strict
     open var evictsObjectsWithDiscardedContent: Bool = false
 
     public override init() {}
@@ -38,10 +68,10 @@ open class NSCache<KeyType : AnyObject, ObjectType : AnyObject> : NSObject {
     open func object(forKey key: KeyType) -> ObjectType? {
         var object: ObjectType?
         
-        let keyRef = unsafeBitCast(key, to: UnsafeRawPointer.self)
+        let key = NSCacheKey(key)
         
         _lock.lock()
-        if let entry = _entries[keyRef] {
+        if let entry = _entries[key] {
             object = entry.value
         }
         _lock.unlock()
@@ -81,7 +111,7 @@ open class NSCache<KeyType : AnyObject, ObjectType : AnyObject> : NSObject {
     }
     
     open func setObject(_ obj: ObjectType, forKey key: KeyType, cost g: Int) {
-        let keyRef = unsafeBitCast(key, to: UnsafeRawPointer.self)
+        let keyRef = NSCacheKey(key)
         
         _lock.lock()
         _totalCost += g
@@ -149,13 +179,13 @@ open class NSCache<KeyType : AnyObject, ObjectType : AnyObject> : NSObject {
         
         _lock.lock()
         for entry in toRemove {
-            _entries.removeValue(forKey: unsafeBitCast(entry.key, to: UnsafeRawPointer.self)) // the cost list is already fixed up in the purge routines
+            _entries.removeValue(forKey: NSCacheKey(entry.key)) // the cost list is already fixed up in the purge routines
         }
         _lock.unlock()
     }
     
     open func removeObject(forKey key: KeyType) {
-        let keyRef = unsafeBitCast(key, to: UnsafeRawPointer.self)
+        let keyRef = NSCacheKey(key)
         
         _lock.lock()
         if let entry = _entries.removeValue(forKey: keyRef) {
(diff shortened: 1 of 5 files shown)


naithar:
@swift-ci please test

naithar:
@alblue @nethraravindran do you have any comments on this, since this changes could influence #876.

Foundation/NSCache.swift
```diff
@@ -20,15 +20,45 @@ private class NSCacheEntry<KeyType : AnyObject, ObjectType : AnyObject> {
...
+            return hashable.hashValue
+        default: return 0
+        }
+    }
+    
+    override func isEqual(_ object: Any?) -> Bool {
+        guard let other = (object as? NSCacheKey) else { return false }
+        
+        guard let left = self.value as? NSObject,
+            let right = other.value as? NSObject else { return self.value === other.value }
```
alblue: we are using 'hashable' as the test for the hashValue, but then we're using identical comparison for the value here. If they are Hashable then we should be using == here instead of ===, and only falling back to === if it's neither hashable nor NSObject.

Foundation/NSCache.swift
```diff
@@ -20,15 +20,45 @@ private class NSCacheEntry<KeyType : AnyObject, ObjectType : AnyObject> {
...
+    private var _entries = Dictionary<NSCacheKey, NSCacheEntry<KeyType, ObjectType>>()
     private let _lock = NSLock()
     private var _totalCost = 0
     private var _byCost: NSCacheEntry<KeyType, ObjectType>?
     
     open var name: String = ""
-    open var totalCostLimit: Int = -1 // limits are imprecise/not strict
-    open var countLimit: Int = -1 // limits are imprecise/not strict
+    open var totalCostLimit: Int = 0 // limits are imprecise/not strict
+    open var countLimit: Int = 0 // limits are imprecise/not strict
```
alblue: These seem like unrelated changes. Why move them from negative number to zero? A count limit of 0 would suggest that it wouldn't in fact cache anything.

naithar: As far as i can tell 0 is the value that is currently used in Darwin's Foundation. Shouldn't we try to match open source Foundation with that on Darwin?

Foundation/NSCache.swift
naithar: As i tested `NSCache` on Darwin's Foundation object corresponding to `Hashable` protocol does not determine the result of `NSCache`'s check for keys uniqueness.
Actually any non `NSObject`-child key is checked using only `===`.
If we want to match Darwin's Foundation that's the behavior we actually want.

alblue: Another thought -- you could do the identity comparison up front and return true if it matches, since if they are identical already then you could just compare pointers and leave early.

    if self.value === other.value {
      return true
    }

naithar: Using `guard self.value !== other.value` for this check would probably be a bad style, isn't it?

Foundation/NSCache.swift
alblue: OK, that makes sense - yes, that does appear to be the default on Darwin, so I agree, it should be the same. It would be worth amending the commit message for this change to indicate why the values have changed for the limits.

alblue: Re the above - I see you mentioned the default in the pull request, but the permanent record is in the git log -- so having the explanation of why you changed the limits in the commit message is a better permanent record of 'why' than in the pull request

pushkarnk:
Looks like the test didn't run.
@swift-ci please test

Foundation/NSCache.swift
naithar: I'll rename the commit after test job is completed.

naithar:
Changed the commit message to cover changes in more permanent way.
Also moved identity comparison before `NSObject` cast.

@swift-ci please test

naithar:
@parkera does master builds with current 3.1 features? Should i nest `NSCacheKey` in `NSCache` and rename it to something like `InnerCacheKey`?

parkera:
It's not critical to me what NSCacheKey is named right now since it's `fileprivate`. We may as well leave it outside so we don't require the newer Swift feature (yet) of nesting stuff inside generic types.

naithar:
Than lets leave it as it is atm.

parkera:
cc @phausler

Foundation/NSCache.swift
```diff
@@ -20,15 +20,49 @@ private class NSCacheEntry<KeyType : AnyObject, ObjectType : AnyObject> {
...
+            
+            return left.isEqual(right)
+        }
+    }
+}
+
 open class NSCache<KeyType : AnyObject, ObjectType : AnyObject> : NSObject {
-    private var _entries = Dictionary<UnsafeRawPointer, NSCacheEntry<KeyType, ObjectType>>()
+    
+    private var _entries = Dictionary<NSCacheKey, NSCacheEntry<KeyType, ObjectType>>()
```
phausler: The key here could have probably just been claimed as either NSObject or AnyHashable since the objc version can't really store non NSObject keys anyhow.

Foundation/NSCache.swift
phausler: Changing these is probably not what you want to do; having them as -1 mean not set whereas zero means that it is an actual set value. If you set the cost limit of a NSCache to zero it will end up purging any items that have a cost greater than zero.

Foundation/NSCache.swift
```diff
@@ -20,15 +20,49 @@ private class NSCacheEntry<KeyType : AnyObject, ObjectType : AnyObject> {
...
+        } else {
+            guard let left = self.value as? NSObject,
+                let right = other.value as? NSObject else { return false }
+            
+            return left.isEqual(right)
+        }
+    }
+}
+
 open class NSCache<KeyType : AnyObject, ObjectType : AnyObject> : NSObject {
```
phausler: I dont know if we could really pull this off but I think the "right" fix here would be to change the KeyType to just Hashable since the Any bridging allows all structural types to pass to objc, and the ObjectType should be Any as well. (But that would require compiler changes and overlay fixes for the Darwin version)

Foundation/NSCache.swift
alblue: @phausler I think the argument was that NSCache on Darwin defaults both of these to zero, rather than -1.

alblue:
@swift-ci please test

Foundation/NSCache.swift
naithar: @alblue is right, Darwin's NSCache defaults this values to 0, not -1. But if -1 default value is intentional, I'll change it back

Foundation/NSCache.swift
naithar: The Darwin's NSCache accepts exactly AnyObject as both KeyType and ObjectType. Also KeyType Is accepted even if it doesn't conform to Hashable protocol.
I guess there shoud be changes done in Darwins version before it's changed here.

Foundation/NSCache.swift
naithar: At first i did implement ditionary key as AnyHashable as NSObject do conform to Hashable protocol. But Darwin's version atm doest not implement this and accepts only AnyObject as KeyType without constraint on Hashable protocol

pushkarnk:
Trying again :)

@swift-ci please test

naithar:
@phausler here is the code that i used to test Darwin's `NSCache` behaviour: [link](https://gist.github.com/naithar/b8c9a55e8f42fe06a4e88ea096ce12d1)  

I've also included the resulting logs for Darwin and Open source Foundation with this PR.

alblue:
LGTM

ianpartridge:
Can we merge this? And cherry-pick for Swift 3.1?

parkera:
I'd like @phausler to sign off on this one

naithar:
ping?

naithar:
@phausler could you please give comments on the logs i've posted above?

parkera:

PR #882, opened 2017-02-15
Disable failing TestNSNumber cases on powerpc64

dgrove-oss:
The default C char on powerpc is unsigned (as on arm).
Add powerpc64 and powerpc64le as extra conditions to the
logic that already disabled the problematic test cases
for arm.


parkera:
@swift-ci test and merge

PR #791, opened 2017-01-12
[EnergyFormatter] Implementation and tests

kweinmeister:
Implemented all EnergyFormatter functions and created tests. Tested on macOS and Ubuntu 16.04.

PR #883, opened 2017-02-16
Initial implementation of MassFormatter.

jszumski:
Implemented MassFormatter and created tests.  

Tests pass 100% on macOS and in a sample iOS project against Objective-C Foundation.

### DIFF
--- a/Foundation/NSMassFormatter.swift
+++ b/Foundation/NSMassFormatter.swift
@@ -10,7 +10,6 @@
 
 extension MassFormatter {
     public enum Unit : Int {
-        
         case gram
         case kilogram
         case ounce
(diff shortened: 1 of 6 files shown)


jszumski:
Localization was not even attempted following the discussion from #791

Foundation/NSMassFormatter.swift
```diff
@@ -21,28 +20,238 @@ extension MassFormatter {
...
+            }
+        } else {
+            return unitString(fromValue: value, unit: unit)
+        }
+    }
+    
+
+    
+    /// The number of pounds in 1 stone
+    private static let poundsPerStone = 14.0
```
jszumski: My goal was to use `Measurement<UnitMass>(value:1, unit:.stones).converted(to:.pounds)` here, but that ends up being 0.3471688213196 lb instead of the expected 14 lb.

jszumski:
@lifeissweetgood Conflicts are resolved, have you had a chance to review?  If you have any pointers on the best way to start the localization work I can get started on that in another branch.

lifeissweetgood:
As an initial implementation, I think this is a pretty solid start.  It leaves room for us to make improvements once we have a better story for supporting localization backed with locale data.

jszumski:
Updated.  Thank you for the feedback!

Foundation/NSMassFormatter.swift
e78l: Not that I use stones regularly, but this seems like an issue :-)
Coefficient should be 6.35029? https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/Unit.swift#L1330

alblue:
@swift-ci please test

Foundation/NSMassFormatter.swift
jszumski: @e78l Those are kilograms not pounds.

e78l: Was referring to the 1 stone=0.347 lb conversion problem by 'this' - changing 0.157473 on L1330 of Unit.swift to 6.35029 should solve that problem.
Appears macOS 10.12.3's Foundation has the stone conversion issue too.
Hard-coding `poundsPerStone = 14.0` looks great (and perhaps is the better way)!

jszumski:
Conflicts are resolved, I think this is ready to merge if there isn't any more feedback.

alblue:
@swift-ci please test

alblue:
Looks good to me - @parkera any objections with merging in as is?

parkera:
Thanks everyone, let's merge it.

Foundation/NSMassFormatter.swift
```diff
@@ -21,28 +20,238 @@ extension MassFormatter {
...
+                                                            .kilogram: "kilograms",
+                                                            .ounce: "ounces",
+                                                            .pound: "pounds",
+                                                            .stone: "stones"]
 }
 
+internal extension Locale {
+    /// TODO: Replace calls to the below function to use Locale.usesMetricSystem
+    /// Temporary workaround due to unpopulated Locale attributes
+    /// See https://bugs.swift.org/browse/SR-3202
```
ianpartridge: SR-3202 is fixed by https://github.com/apple/swift-corelibs-foundation/pull/1020 - can this workaround be removed @jszumski ?

PR #885, opened 2017-02-17
Fixed file's groupOwnerName in attributesOfItem()

amosavian:
Fixed group owner name assigned to `.groupOwnerAccountID` instead of `.groupOwnerAccountName` in `FileManager.attributesOfItem(atPath:)`

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -305,7 +305,7 @@ open class FileManager : NSObject {
         let grd = getgrgid(s.st_gid)
         if grd != nil && grd!.pointee.gr_name != nil {
             let name = String(cString: grd!.pointee.gr_name)
-            result[.groupOwnerAccountID] = name
+            result[.groupOwnerAccountName] = name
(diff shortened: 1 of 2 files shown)


parkera:
Looks correct. Do we have a test case?

amosavian:
@parkera I didn't check, I just saw that incidentally. But this is test case:

https://github.com/apple/swift-corelibs-foundation/blob/master/TestFoundation/TestNSFileManager.swift#L117-L163

There is one for `.ownerAccountID`, but the rest are orphan.

parkera:
Could you add a line to that test case that verifies that the name is set?

I know this fix seems trivial but I want to keep expanding our test cases so we don't have a chance of re-introducing a regression.

amosavian:
@parkera Added. This checks mentioned attributes are `String` if they are returned (aka when file system supports), and checks not to be empty strings.

parkera:
Great, thanks.

parkera:

PR #887, opened 2017-02-20
Temporary workaround for Crash in Date.description[SR-2462]

saiHemak:


saiHemak:
@parkera  : Please find the answers inline 
Does it introduce a new warning though? Yes it introduces new  warning .
 And where else would we have to do this? I observed the crash in description of NSDate only.Other places am not sure of it .
Do you know if this problem is a regression in Swift 4? No.I have see my test case crashing with Swift 3.0.2 also

Foundation/NSDate.swift
```diff
@@ -114,7 +114,7 @@ open class NSDate : NSObject, NSCopying, NSSecureCoding, NSCoding {
      */
     open override var description: String {
         let dateFormatterRef = CFDateFormatterCreate(kCFAllocatorSystemDefault, nil, kCFDateFormatterFullStyle, kCFDateFormatterFullStyle)
-        let timeZone = CFTimeZoneCreateWithTimeIntervalFromGMT(kCFAllocatorSystemDefault, 0.0)
+        var timeZone = CFTimeZoneCreateWithTimeIntervalFromGMT(kCFAllocatorSystemDefault, 0.0)
```
pushkarnk: Can we have a comment here mentioning that this is a workaround? I guess this change will generate a compiler warning ...

alblue: It would be better to investigate further and understand where the crash is coming from; if we mask this here then it will simply go unnoticed thereafter

parkera: I agree with @alblue ; this bug is very concerning. We could be crashing all over the place.

saiHemak:
@parkera  modified workaround to avoid compiler warning.Please review

PR #888, opened 2017-02-20
Implemented FileManager.attributesOfFileSystem()

amosavian:
Also I'm not sure I should use `fileSystemRepresentation(withPath: path)` method to get proper variable for `statfs()` or not. Because `attributesOfItem(atPath:) didn't use that, I didn't either.

### DIFF
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -350,10 +350,11 @@ open class FileManager : NSObject {
         
         var result = [FileAttributeKey : Any]()
         let blockSize = UInt64(s.f_bsize)
-        result[.systemSize] = NSNumber(value: blockSize * s.f_blocks)
-        result[.systemFreeSize] = NSNumber(value: blockSize * s.f_bavail)
-        result[.systemNodes] = NSNumber(value: s.f_files)
-        result[.systemFreeNodes] = NSNumber(value: s.f_ffree)
+        result[.systemNumber] = NSNumber(value: UInt64(s.f_fsid.val.0))
+        result[.systemSize] = NSNumber(value: blockSize * UInt64(s.f_blocks))
+        result[.systemFreeSize] = NSNumber(value: blockSize * UInt64(s.f_bavail))
+        result[.systemNodes] = NSNumber(value: UInt64(s.f_files))
+        result[.systemFreeNodes] = NSNumber(value: UInt64(s.f_ffree))
         
         return result
     }
--- a/Foundation/NSFileManager.swift
+++ b/Foundation/NSFileManager.swift
@@ -343,7 +343,19 @@ open class FileManager : NSObject {
         This method replaces fileSystemAttributesAtPath:.
      */
     open func attributesOfFileSystem(forPath path: String) throws -> [FileAttributeKey : Any] {
-        NSUnimplemented()
+        var s = statfs()
+        guard statfs(path, &s) == 0 else {
+            throw _NSErrorWithErrno(errno, reading: true, path: path)
+        }
+        
+        var result = [FileAttributeKey : Any]()
+        let blockSize = UInt64(s.f_bsize)
+        result[.systemSize] = NSNumber(value: blockSize * s.f_blocks)
+        result[.systemFreeSize] = NSNumber(value: blockSize * s.f_bavail)
+        result[.systemNodes] = NSNumber(value: s.f_files)
+        result[.systemFreeNodes] = NSNumber(value: s.f_ffree)
+        
+        return result
     }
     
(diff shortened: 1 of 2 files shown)


alblue:
Can you squash these two commits? It doesn't really make sense having one commit per file, especially as this is adding the test for the one in the first commit.

amosavian:
@alblue Find this

TestFoundation/TestNSFileManager.swift
```diff
@@ -114,6 +115,32 @@ class TestNSFileManager : XCTestCase {
...
+        let fm = FileManager.default
+        let path = NSTemporaryDirectory()
+        
+        do {
+            let attrs = try fm.attributesOfFileSystem(forPath: path)
+            
+            XCTAssertTrue(attrs.count > 0)
+            
+            let systemFreeSize = attrs[.systemFreeSize] as? NSNumber
+            XCTAssertGreaterThan(systemFreeSize!.int64Value, 0)
```
phausler: might be good to XCTAssertNotNil on the result before using ! that way any failures (in any refactoring) would immediately be visible

Foundation/NSFileManager.swift
```diff
@@ -343,7 +343,19 @@ open class FileManager : NSObject {
...
+        guard statfs(path, &s) == 0 else {
+            throw _NSErrorWithErrno(errno, reading: true, path: path)
+        }
+        
+        var result = [FileAttributeKey : Any]()
+        let blockSize = UInt64(s.f_bsize)
+        result[.systemSize] = NSNumber(value: blockSize * s.f_blocks)
+        result[.systemFreeSize] = NSNumber(value: blockSize * s.f_bavail)
+        result[.systemNodes] = NSNumber(value: s.f_files)
+        result[.systemFreeNodes] = NSNumber(value: s.f_ffree)
```
phausler: `.systemNumber` should be the `s.f_fsid.val.0`
`.systemNodes` and `.systemFreeNodes` needs `quotactl` which I am not certain is portable.

Foundation/NSFileManager.swift
```diff
@@ -343,7 +343,19 @@ open class FileManager : NSObject {
         This method replaces fileSystemAttributesAtPath:.
      */
     open func attributesOfFileSystem(forPath path: String) throws -> [FileAttributeKey : Any] {
-        NSUnimplemented()
+        var s = statfs()
+        guard statfs(path, &s) == 0 else {
```
phausler: the path here does have a slight problem; non-ascii characters of file names probably will cause issues. In the objc version we use effectively the same idea as `fileSystemRepresentationWithPath(:String)`

However this shortcoming shouldn't stop this PR imho

Foundation/NSFileManager.swift
phausler: Also one other note: since swift-corelibs-foundation does not have bridging like the objc version- by storing the NSNumber here you are requiring callers to as cast to NSNumber and things like UInt64 (which they probably would want) will not work.

Foundation/NSFileManager.swift
amosavian: @phausler I mentioned this issue in conversation, however I considered `attributesOfItem()` method implementation to address. If we should correct something here, that will also effect `attributesOfItem()` method too.

Foundation/NSFileManager.swift
amosavian: There is little difference between Darwin and Linux version of `quotactl`

phausler:
@swift-ci test

amosavian:
I think quotactl implementation must be something like that:

```swift
let uid = Int32(getuid())
let memsize = MemoryLayout<dqblk>.size
var quotaresult = caddr_t.allocate(capacity: memsize)
let command = Q_GETQUOTA << SUBCMDSHIFT + USRQUOTA & SUBCMDMASK
let path = "/Users/amirabbas/"
let result: Int32
#if os(macOS) || os(iOS)
    result = quotactl(path, command, uid, quotaresult)
#elseif os(Linux)
    result = quotactl(command, path, uid, quotaresult)
#endif
if result == 0 {
    quotaresult.withMemoryRebound(to: dqblk.self, capacity: 1, {
        result[.systemNodes] = NSNumber(value: UInt64($0.pointee.dqb_ihardlimit))
        result[.systemFreeNodes] = NSNumber(value: UInt64($0.pointee.dqb_curinodes))
    })
}
quotaresult.deallocate(capacity: memsize)
```

alblue:
@swift-ci please test

alblue:
Can you rebase and we can retest? The cause of the test failure has now been cleaned by the CI system.

alblue:
This pull request has gone stale, closing. If you want to re-open this with a rebased version, please go ahead.

PR #672, opened 2016-10-06
Initial implementation of HTTPCookieStorage

pushkarnk:
I've created this pull request for review and comments. There are two important considerations here, around cookie persistence:

1. How do you store the cookies?
>> Writing an NSDictionary to a file using `NSDictionary(writeToFile:atomically:)`

2. Where do you store them? 
>> XDG_CONFIG_HOME. See #889 

TODOs:

- [x] setCookies with non-nil mainDocumentURL
- [x] Implementing `sharedCookieStorage(forGroupContainerIdentifier:)`
~~Implementing `sortedCookies(using sortOrder: [NSSortDescriptor])`~~
   NSSortDescriptor is not implemented (is it supported on Linux?)

PR #889, opened 2017-02-22
Implementation of XDG File specification and a basic implementation of HTTPCookieStorage

mamabusi:
This is an unpolished implementation of the XDG File specification. Have referred to related work started by Tony Parker in the repo: https://github.com/parkera/swift-corelibs-foundation/tree/parkera/xdg_implementation

Note that the tests for the API implementation have been done on my local build setup. The same are not available in the committed changes here as they involve adding API in the Foundation to access the CoreFoundation implementation that is done. 
Please suggest if there is an alternate way of adding tests.

### DIFF
--- a/CoreFoundation/Base.subproj/CFFileUtilities.c
+++ b/CoreFoundation/Base.subproj/CFFileUtilities.c
@@ -1221,14 +1221,21 @@ CFArrayRef _CFXDGCreateDataDirectoriesPaths(void) {
     // $XDG_DATA_DIRS defines the preference-ordered set of base directories to search for data files in addition to the $XDG_DATA_HOME base directory. The directories in $XDG_DATA_DIRS should be seperated with a colon ':'.
     // If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used.
     const char *dataDirectoriesPaths = __CFgetenv("XDG_DATA_DIRS");
+    CFStringRef defaultPath[2];
+    defaultPath[0] = CFSTR("/usr/local/share");
+    defaultPath[1] = CFSTR("/usr/share");
     if ((dataDirectoriesPaths == NULL) || (dataDirectoriesPaths[0] == '\0')) {
         // Environmental variable not set. Return default value.
-        CFStringRef defaultPath[2];
-        defaultPath[0] = CFStringCreateWithCString(kCFAllocatorSystemDefault, "/usr/local/share/", _kCFXDGStringEncoding);
-        defaultPath[1] = CFStringCreateWithCString(kCFAllocatorSystemDefault, "/usr/share/", _kCFXDGStringEncoding);
         return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)defaultPath, 2, &kCFTypeArrayCallBacks);
     }
-    return _CFTokenizeStringToCFArrayOfCFStrings(dataDirectoriesPaths, ':');
+    CFArrayRef dataDirPathsArray = _CFCreateCFArrayByTokenizingString(dataDirectoriesPaths, ':');
+    if(CFArrayGetCount(dataDirPathsArray) == 0) {
+        CFStringRef logMessage = CFSTR("Value set in XDG_DATA_DIRS variable not honoured. Returning the default.");
+        CFLog(kCFLogLevelWarning, CFSTR("%@"), logMessage);
+        CFRelease(dataDirPathsArray);
+        return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)defaultPath, 2, &kCFTypeArrayCallBacks);
+    }
+    return dataDirPathsArray;
 }
 
 
@@ -1238,13 +1245,20 @@ CFArrayRef _CFXDGCreateConfigDirectoriesPaths(void) {
     // $XDG_CONFIG_DIRS defines the preference-ordered set of base directories to search for configuration files in addition to the $XDG_CONFIG_HOME base directory. The directories in $XDG_CONFIG_DIRS should be seperated with a colon ':'.
     // If $XDG_CONFIG_DIRS is either not set or empty, a value equal to /etc/xdg should be used.
     const char *configDirectoriesPaths = __CFgetenv("XDG_CONFIG_DIRS");
+    CFStringRef defaultPath[1];
+    defaultPath[0] = CFSTR("/etc/xdg");
     if ((configDirectoriesPaths == NULL) || (configDirectoriesPaths[0] == '\0')) {
         //Environmental variable not set. Return default value.
-        CFStringRef defaultPath[1];
-        defaultPath[0] = CFStringCreateWithCString(kCFAllocatorSystemDefault, "/etc/xdg", _kCFXDGStringEncoding);
         return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)defaultPath, 1, &kCFTypeArrayCallBacks);
     }
-    return _CFTokenizeStringToCFArrayOfCFStrings(configDirectoriesPaths, ':');
+    CFArrayRef configDirPathsArray = _CFCreateCFArrayByTokenizingString(configDirectoriesPaths, ':');
+    if(CFArrayGetCount(configDirPathsArray) == 0) {
+        CFStringRef logMessage = CFSTR("Value set in XDG_CONFIG_DIRS variable not honoured. Returning the default.");
+        CFLog(kCFLogLevelWarning, CFSTR("%@"), logMessage);
+        CFRelease(configDirPathsArray);
+        return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)defaultPath, 1, &kCFTypeArrayCallBacks);
+    }
+    return configDirPathsArray;
 }
 
 /// a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable $XDG_CACHE_HOME.
@@ -1276,7 +1290,7 @@ CFStringRef _CFXDGCreateRuntimeDirectoryPath(void) {
 }
 
 
-CF_PRIVATE CFArrayRef _CFTokenizeStringToCFArrayOfCFStrings(const char *values, char delimiter) {
+CF_PRIVATE CFArrayRef _CFCreateCFArrayByTokenizingString(const char *values, char delimiter) {
     size_t pathCount = 0;
     char* tmpDirectoriesPaths = (char*)values;
     char* last_colon = 0;
@@ -1292,6 +1306,9 @@ CF_PRIVATE CFArrayRef _CFTokenizeStringToCFArrayOfCFStrings(const char *values, 
     }
     // Add count for trailing path unless ending with colon.
     pathCount += last_colon < (values + strlen(values) - 1);
+    if (pathCount > 64) {
+        return CFArrayCreate(kCFAllocatorSystemDefault, NULL, 0, &kCFTypeArrayCallBacks);
+    }
     if (pathCount > 0)
     {
         size_t validPathCount  = 0;
@@ -1304,16 +1321,25 @@ CF_PRIVATE CFArrayRef _CFTokenizeStringToCFArrayOfCFStrings(const char *values, 
         while (path)
         {
             assert(validPathCount < pathCount);
-            CFStringRef dirPath = CFStringCreateWithCString(kCFAllocatorSystemDefault, strdup(path), _kCFXDGStringEncoding);
+            char* pathString  = strdup(path);
+            CFStringRef dirPath = CFStringCreateWithCString(kCFAllocatorSystemDefault, pathString, _kCFXDGStringEncoding);
             CFStringRef slash = CFSTR("/");
             CFStringRef tilde = CFSTR("~");
             // Check for absolutePath, if not ignore.
             if (CFStringHasPrefix(dirPath, slash) || CFStringHasPrefix(dirPath, tilde) ) {
                 pathList[validPathCount++] = dirPath;
+            } else {
+                CFRelease(dirPath);
             }
-            path = strtok(0, delimiterStr);
+            path = strtok(NULL, delimiterStr);
+            free(pathString);
         }
-        return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)pathList, validPathCount, &kCFTypeArrayCallBacks);
+        free(copyDirPath);
+        CFArrayRef pathArray = CFArrayCreate(kCFAllocatorSystemDefault, (const void **)pathList  , validPathCount, &kCFTypeArrayCallBacks);
+        for(int i = 0; i < validPathCount; i ++) {
+            CFRelease(pathList[i]);
+        }
+        return pathArray;
     }
     return CFArrayCreate(kCFAllocatorSystemDefault, NULL, 0, &kCFTypeArrayCallBacks);
 }
--- a/CoreFoundation/Base.subproj/CFFileUtilities.c
+++ b/CoreFoundation/Base.subproj/CFFileUtilities.c
@@ -15,6 +15,7 @@
 #include <errno.h>
 #include <string.h>
 #include <stdio.h>
+#include <assert.h>
 
 #if DEPLOYMENT_TARGET_WINDOWS
 #include <io.h>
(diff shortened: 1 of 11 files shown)


parkera:
The testing is absolutely the hardest part of this; we also need a way to start new child processes with new environment variables to make sure everything is configured correctly.

mamabusi:
I have written to the community for ideas to get the tests done. Only way I could think of now is to use the existing APIs (one example being HTTPCookieStorage API) that use these environmental variables to get a part of the tests done. Need to check on where the remaining environmental variable usage would fit in. 

Regarding the  "start new child processes with new environment variables to make sure everything is configured correctly", does this mean that we would be setting the  environmental variables using these child processes and the test API is expected to reflect the values set? 
I am looking into the 'Process' API that would help in creating the child process and launch the tasks. Would that be a right approach? 
Thanks!

parkera:
Yes, probably we'll need to create a helper process that gets launched via `Process` with the specific environment variables set. Then we need that process to use some existing API that would be affected by the presence of the environment variables (preferences, maybe?). Probably then it would exit with a success or failure code, which the test harness would then interpret as either success or failure.

mamabusi:
@parkera Ok, will get onto it. Meanwhile, request you to review the implementation thus far.
Thanks!

parkera:
Yup, looks reasonable so far. Thanks!

mamabusi:
@parkera  By mentioning "Then we need that process to use some existing API that would be affected by the presence of the environment variables (preferences, maybe?)", did you mean to use the NSUserDefaults on Foundation?

parkera:
Yes, that would be a good candidate. `NSUserDefaults` has a search list that looks at several locations in a particular order to find a value for a user default. On Darwin this includes places like `~/Library`, `/Library`, etc. With this XDG stuff implemented, I would expect it to affect the places that `CFPreferences` (which is behind `NSUserDefaults`) looks for those values on Linux.

mamabusi:
@parkera  At the outlook, the NSUserDefaults code on Linux currently does not look complete. I see api code in it being commented stating compiler failures / code confined to only OS x/ iOS platforms / unimplemented api and also the test-cases on TestFoundation are commented out. 
There is this init that was of interest to me as it mentions about the search list:
```
/// nil suite means use the default search list that +standardUserDefaults uses
    public init?(suiteName suitename: String?) {
        suite = suitename
    }
```

On a search done in Foundation for 'standardUserDefaults' , the results do not yield the setting or use of this elsewhere. This implies that there is no default value set on Linux?
Any comments on this? 
Sorry if I am totally off-beat here. I would look into further (on the CFPreferences as well) to check if I can find something. 
Thanks!

e78l:
Hello @mamabusi 
Hoping this info could help:
`standardUserDefaults` was changed to `standard` in UserDefaults.swift

I found this to be a good read about suite names and such:
http://dscoder.com/defaults.html

CFXMLPreferencesDomain.c may be the simplest way to get things working; it seems to have been checked in[1], but some other changes in CF are needed[2]
[1] https://github.com/apple/swift-corelibs-foundation/blob/master/CoreFoundation/Preferences.subproj/CFXMLPreferencesDomain.c
[2] Delete CFPreferences.c line 500 + change the URL paths, perhaps in _CFPreferencesURLForStandardDomainWithSafetyLevel
https://github.com/apple/swift-corelibs-foundation/blob/master/CoreFoundation/Preferences.subproj/CFPreferences.c#L500

TestNSUserDefaults.swift has `test_getRegisteredDefaultItem` but that only works since there's a  volatile domain/dictionary in UserDefaults.swift. More tests are probably needed.



+ @parkera 
I've been thinking of some arguments to switch from XML settings to JSON(/binary PLIST/YAML?) on linux but don't have anything formal. My main arguments for the switch would be domains+hosts aren't necessary/common enough for Linux, and inter-OS file compatibility isn't a significant factor (few/no one would copy settings across platforms).
This might also be a good opportunity to rewrite UserDefaults.swift, perhaps with cleaner file writing/locking. I know there're reasons for not re-writing SCL in Swift, but I perceive the overhead of CF user defaults to be large.
Is this something worth discussing/pursuing?

parkera:
User defaults are simply property lists (which can be XML or binary), which is basically the lingua franca of CoreFoundation. I don't think there is much value in changing the format; but if we did it would be orthogonal to getting the basic API working. I do want to keep the CFPrefs concept of search lists in place, although some (like the network home) will not make sense on this platform.

Maybe instead of user defaults, which has its own set of challenges here, we should focus on the use case which got us here in the first place? I think that was where `URLSession` should store its cached data.

mamabusi:
@e78l @parkera Thanks for your time and info provided here. 

@parkera It was NSHTTPCookieStorage that got us to XDG implementation and as suggested I plan to use the same for adding in test-cases.

pushkarnk:
@mamabusi I will update the HTTPCookieStorage PR by early next week, its been a while

pushkarnk:
@mamabusi @parkera For #672 I plan to use `XDG_CONFIG_HOME`. Though `XDG_DATA_HOME` sounds better for cookies, `XDG_CONFIG_HOME` seems to be commonly used. We can change it later if needed.

e78l:
@parkera OK, and thanks for the response. There do seem to be some file size+performance penalties from writing a plist to XML (repeatedly), but as I also like readable settings, there isn't much to change (XML vs binary plist).
@mamabusi No problem!

mamabusi:
@parkera Could the commit of HTTPCookieStorage be combined with this? 
The reason being that it uses XDG_CONFIG_HOME and has its test-suite in place.
Thanks!

pushkarnk:
@mamabusi For #672 I am in the process of wrapping up with the tests. Though an initial implementation is in place, I still need to tweak it to use the `XDG_CONFIG_HOME`.

mamabusi:
@pushkarnk Cool.  Once the changes are in place for using the XDG variable, I'll verify the correctness of the path retrieval in HTTPCookieStorage. Thanks!

mamabusi:
@parkera 
Hi
Regarding the setting of the XDG environment variables, I had a couple of queries. 
 
I set the environment variable using the Foundation Api: 
       setenv("XDG_CONFIG_HOME", "/usr/test", 1)

Retrieved it using the api: 
        let rawValue = getenv("XDG_CONFIG_HOME")
        let sampleString = String(utf8String: rawValue!)
        print(sampleString!)

I could fetch the variable that is set. 

Whereas, on the CoreFoundation when I try to fetch this using the API __CFgetenv(), I see that the variable is not set.

Wondering what is the reason that it is not reflecting on the CoreFoundation side. Do they refer to different scopes of environment? Aren’t all of these running in the same process? 

Also, if I have to set the system-wide environment variables (in this case the XDG variables should be system-wide) after much reading I figured out that it is not possible programmatically. Hence, I would have to do it by launching a script file that would edit the system (read-only) files that require the root access of the machine where the tests are to be run. Is this a feasible option? Or is there a alternative that is already being used in TestFoundation? Am I missing something here? 
Your inputs would be helpful.

Thanks!

parkera:
`__CFgetenv` caches the result of `getenv` for performance reasons; so you will have to make sure the environment variable is set before the process runs.

lib/phases.py
```diff
@@ -417,6 +417,7 @@ class SwiftExecutable(BuildPhase):
     def __init__(self, executableName, sources):
         BuildAction.__init__(self, output=executableName)
         self.executableName = executableName
+        self.name = executableName
```
mamabusi: Adding a dependency of the executable creation for XDG  to foundation_tests (which is the SwiftExecutable: Test Foundation) in build.py caused a script failure here as the 'name' seemed to be 'nil'. Hence, the change.

mamabusi:
@pushkarnk Have cherry picked your HTTPCookieStorage implementation and modified the same to use the environment variable of XDG file specification. Please review. 

@parkera Have added test-case for XDG file specification implementation. Please review.

Thanks!

pushkarnk:
@swift please test

pushkarnk:
@swift please test

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci please test

pushkarnk:
Third time again!
@swift-ci please test

pushkarnk:
Looks like a swiftpm failure:
```
/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swiftpm/Tests/SourceControlTests/RepositoryManagerTests.swift:188: error: RepositoryManagerTests.testReset : XCTAssertEqual failed: ("1") is not equal to ("2") - 
```

pushkarnk:
Seems unrelated to me. I am kicking off a test again. 

@swift-ci please test

mamabusi:
@parkera Is this mergeable? Any comments?

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,155 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+CFArrayRef _CFXDGCreateDataDirectoriesPaths(void) {
+    // $XDG_DATA_DIRS defines the preference-ordered set of base directories to search for data files in addition to the $XDG_DATA_HOME base directory. The directories in $XDG_DATA_DIRS should be seperated with a colon ':'.
+    // If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used.
+    const char *dataDirectoriesPaths = __CFgetenv("XDG_DATA_DIRS");
+    if ((dataDirectoriesPaths == NULL) || (dataDirectoriesPaths[0] == '\0')) {
+        // Environmental variable not set. Return default value.
+        CFStringRef defaultPath[2];
+        defaultPath[0] = CFStringCreateWithCString(kCFAllocatorSystemDefault, "/usr/local/share/", _kCFXDGStringEncoding);
+        defaultPath[1] = CFStringCreateWithCString(kCFAllocatorSystemDefault, "/usr/share/", _kCFXDGStringEncoding);
+        return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)defaultPath, 2, &kCFTypeArrayCallBacks);
```
parkera: I believe we've leaked the two `defaultPath` entries here. I think we could just use `CFSTR("")` constant strings instead.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,155 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+/// a set of preference ordered base directories relative to which configuration files should be searched. This set of directories is defined by the environment variable $XDG_CONFIG_DIRS.
+CF_SWIFT_EXPORT
+CFArrayRef _CFXDGCreateConfigDirectoriesPaths(void) {
+    // $XDG_CONFIG_DIRS defines the preference-ordered set of base directories to search for configuration files in addition to the $XDG_CONFIG_HOME base directory. The directories in $XDG_CONFIG_DIRS should be seperated with a colon ':'.
+    // If $XDG_CONFIG_DIRS is either not set or empty, a value equal to /etc/xdg should be used.
+    const char *configDirectoriesPaths = __CFgetenv("XDG_CONFIG_DIRS");
+    if ((configDirectoriesPaths == NULL) || (configDirectoriesPaths[0] == '\0')) {
+        //Environmental variable not set. Return default value.
+        CFStringRef defaultPath[1];
+        defaultPath[0] = CFStringCreateWithCString(kCFAllocatorSystemDefault, "/etc/xdg", _kCFXDGStringEncoding);
```
parkera: Leak again, and the string is constant.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,155 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+    if (runtimeDir && strnlen(runtimeDir, CFMaxPathSize) > 1 && runtimeDir[0] == '/') {
+        return CFStringCreateWithCString(kCFAllocatorSystemDefault, runtimeDir, _kCFXDGStringEncoding);
+    } else {
+        // If $XDG_RUNTIME_DIR is not set applications should fall back to a replacement directory with similar capabilities and print a warning message.
+        return CFStringCreateWithCString(kCFAllocatorSystemDefault, "", _kCFXDGStringEncoding);
+    }
+}
+
+
+CF_PRIVATE CFArrayRef _CFTokenizeStringToCFArrayOfCFStrings(const char *values, char delimiter) {
```
parkera: This guy returns a +1 reference, so by convention needs to have `Create` in the name.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,155 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+        CFStringRef pathList[pathCount];
+        char* copyDirPath = strdup(values);
+        char delimiterStr[2];
+        delimiterStr[0] = delimiter;
+        delimiterStr[1] = '\0';
+        char* path = strtok(copyDirPath, delimiterStr);
+        while (path)
+        {
+            assert(validPathCount < pathCount);
+            CFStringRef dirPath = CFStringCreateWithCString(kCFAllocatorSystemDefault, strdup(path), _kCFXDGStringEncoding);
```
parkera: `strdup` requires a call to `free`

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,155 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+            CFStringRef slash = CFSTR("/");
+            CFStringRef tilde = CFSTR("~");
+            // Check for absolutePath, if not ignore.
+            if (CFStringHasPrefix(dirPath, slash) || CFStringHasPrefix(dirPath, tilde) ) {
+                pathList[validPathCount++] = dirPath;
+            }
+            path = strtok(0, delimiterStr);
```
parkera: Can we pass `NULL` instead of `0`? It's more semantically correct for a pointer value.

CoreFoundation/Base.subproj/CFFileUtilities.c
parkera: Also, after the array is created of these, each `dirPath` is leaked.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,155 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+        }
+        tmpDirectoriesPaths++;
+    }
+    // Add count for trailing path unless ending with colon.
+    pathCount += last_colon < (values + strlen(values) - 1);
+    if (pathCount > 0)
+    {
+        size_t validPathCount  = 0;
+        CFStringRef pathList[pathCount];
+        char* copyDirPath = strdup(values);
```
parkera: Another `free` needed for this.

TestFoundation/XDGTestHelper.swift
```diff
@@ -0,0 +1,19 @@
...
+            ])!
+let rawValue = getenv("XDG_CONFIG_HOME")        
+let xdg_config_home = String(utf8String: rawValue!)
+storage.setCookie(simpleCookie)
+XCTAssertEqual(storage.cookies!.count, 1)
+let fm = FileManager.default
+let destPath = xdg_config_home! + "/.cookies.shared"
+var isDir = false
+let exists = fm.fileExists(atPath: destPath, isDirectory: &isDir) 
+XCTAssertTrue(exists)
```
parkera: Double checking: XCTest does exit here with a -1 if this assert fails?

mamabusi:
@parkera Have addressed the review comments. Thanks for pointing them out. Please have a look.

TestFoundation/XDGTestHelper.swift
```diff
@@ -0,0 +1,20 @@
...
+            .domain: "example.com",
+            ])!
+let rawValue = getenv("XDG_CONFIG_HOME")        
+let xdg_config_home = String(utf8String: rawValue!)
+storage.setCookie(simpleCookie)
+let fm = FileManager.default
+let destPath = xdg_config_home! + "/.cookies.shared"
+var isDir = false
+let exists = fm.fileExists(atPath: destPath, isDirectory: &isDir) 
+if (!exists) {
```
mamabusi: An XCTAssert  here would not cascade the exit status as -1. Therefore, the explicit exit from process.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,162 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+            if (CFStringHasPrefix(dirPath, slash) || CFStringHasPrefix(dirPath, tilde) ) {
+                pathList[validPathCount++] = dirPath;
+            }
+            path = strtok(NULL, delimiterStr);
+            free(pathString);
+        }
+        free(copyDirPath);
+        CFArrayRef pathArray = CFArrayCreate(kCFAllocatorSystemDefault, (const void **)pathList  , validPathCount, &kCFTypeArrayCallBacks);
+        for(int i = 0; i < pathCount; i ++) {
+            CFRelease(pathList[i]);
```
mamabusi: Not sure if there is a better way to release the entries of CFStringRef here. But, this would ensure the dirPath entries are all wiped off.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1276,7 +1276,7 @@ CFStringRef _CFXDGCreateRuntimeDirectoryPath(void) {
 }
 
 
-CF_PRIVATE CFArrayRef _CFTokenizeStringToCFArrayOfCFStrings(const char *values, char delimiter) {
+CF_PRIVATE CFArrayRef _CFCreateCFArrayByTokenizingString(const char *values, char delimiter) {
     size_t pathCount = 0;
     char* tmpDirectoriesPaths = (char*)values;
     char* last_colon = 0;
```
parkera: Sorry to be picky here, but this is code that deals with untrusted input so we have to be extra careful of security concerns.

Can we put a cap on pathCount, as the value causes a stack allocation? Say, 64. If we exceed this count then return a sentinel value, like empty array or NULL (and deal with it one frame up).

TestFoundation/XDGTestHelper.swift
parkera: Ok, this makes more sense to me here.

parkera: We should remove the `import XCTest` from the top of this file too.

CoreFoundation/Base.subproj/CFFileUtilities.c
mamabusi: @parkera  Sure, I would do the needed changes and update. 
When NULL/empty array is returned, one way of dealing with it is to construct the default values for the XDG variable and return. Does it sound ok to you?

parkera: That seems reasonable.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,179 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+        if (*tmpDirectoriesPaths == delimiter)
+        {
+            pathCount++;
+            last_colon = tmpDirectoriesPaths;
+        }
+        tmpDirectoriesPaths++;
+    }
+    // Add count for trailing path unless ending with colon.
+    pathCount += last_colon < (values + strlen(values) - 1);
+    if (pathCount > 64) {
```
mamabusi: Limits the number of paths to 64 and returns an empty array if pathCount exceeds 64.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,179 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+    defaultPath[0] = CFSTR("/usr/local/share");
+    defaultPath[1] = CFSTR("/usr/share");
+    if ((dataDirectoriesPaths == NULL) || (dataDirectoriesPaths[0] == '\0')) {
+        // Environmental variable not set. Return default value.
+        return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)defaultPath, 2, &kCFTypeArrayCallBacks);
+    }
+    CFArrayRef dataDirPathsArray = _CFCreateCFArrayByTokenizingString(dataDirectoriesPaths, ':');
+    if(CFArrayGetCount(dataDirPathsArray) == 0) {
+        CFStringRef logMessage = CFSTR("Value set in XDG_DATA_DIRS variable not honoured. Returning the default.");
+        CFLog(kCFLogLevelWarning, CFSTR("%@"), logMessage);
```
mamabusi: In the case of the pathCount > 64 or all the paths in the variable turn out to be invalid, an empty array is returned. At such times the default path is returned although the variable was set. The log warning message gives a hint.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1166,3 +1167,179 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
...
+    CFStringRef defaultPath[1];
+    defaultPath[0] = CFSTR("/etc/xdg");
+    if ((configDirectoriesPaths == NULL) || (configDirectoriesPaths[0] == '\0')) {
+        //Environmental variable not set. Return default value.
+        return CFArrayCreate(kCFAllocatorSystemDefault, (const void **)defaultPath, 1, &kCFTypeArrayCallBacks);
+    }
+    CFArrayRef configDirPathsArray = _CFCreateCFArrayByTokenizingString(configDirectoriesPaths, ':');
+    if(CFArrayGetCount(configDirPathsArray) == 0) {
+        CFStringRef logMessage = CFSTR("Value set in XDG_CONFIG_DIRS variable not honoured. Returning the default.");
+        CFLog(kCFLogLevelWarning, CFSTR("%@"), logMessage);
```
mamabusi: In the case of the pathCount > 64 or all the paths in the variable turn out to be invalid, an empty array is returned. At such times the default path is returned although the variable was set. The log warning message gives a hint.

mamabusi:
@parkera Addressed the requested changes. Please take a look.

CoreFoundation/Base.subproj/CFFileUtilities.c
parkera: Ok; one more thing now - we need to release `configDirPathsArray` here before returning, otherwise we leak it.

mamabusi: @parkera Sorry I missed that. Done now.

alblue:
@swift-ci please test

pushkarnk:
Looks like the tests didn't run

@swift-ci please test

mamabusi:
@parkera All the review changes are completed. Please let me know if something is missed. 
Thanks!

parkera:
Thanks @mamabusi! Let's get this merged.

mamabusi:
Yay! Thanks a ton for all the guidance @parkera

parkera:
No problem, thank you for finishing this up.

ahti:
I know this has been merged already, but I don't think the current paths for cookie storage are appropriate.

Firstly, `$XDG_CONFIG_HOME` (`.config` by default) is supposed to be used for "user specific configuration files". I don't think cookies fit that description. `$XDG_DATA_HOME` (`.local/share` by default), used for "user specific data files", sounds like a better fit to me.

Secondly, none of my Linux machines (Arch Linux and Ubuntu) contain any files in either `.config` or `.local/share`. Every application, even ones with only one file, create their own subfolder. The specification is not clear on this matter, but it does state:
> Other specifications may reference this specification by specifying the location of a data file as $XDG_DATA_DIRS/subdir/filename.

This to me seems like reason enough to put any files Swift creates in a dedicated subfolder as well. This subfolder should probably be a separate folder per application, but see the following paragraphs for that.

Finally, the current implementation uses the same shared storage for all users of the api, while that behavior was removed from OSX Foundation. The doc comment of `HTTPCookieStorage.shared` says:
> Starting in OS X 10.11, each app has its own sharedHTTPCookieStorage singleton, which will not be shared with other applications.

I think it would be reasonable for the Linux implementation to have the same behaviour, although that change might be better to hold back until corelibs-foundation has a more fleshed-out concept of different directories on Linux including implementing `FileManager.url(for: ...)`, probably returning per-application subdirectories of xdg directories.

parkera:
Those comments do make sense to me. It would be good to get them into a bug on bugs.swift.org, and perhaps follow up with a discussion on swift-corelibs-dev mailing list.

ianpartridge:
@mamabusi can you take a look at this?

mamabusi:
@ ahti Thanks for the suggestions.  
The choice of XDG_DATA_HOME and XDG_CONFIG_HOME for cookie storage did come up during the implementation. We (@pushkarnk and me) had a discussion on the mailing list and based on the inputs we got, went with the XDG_CONFIG_HOME.  As the XDG_DATA_HOME seems a better fit here, I will make the necessary changes and we could discuss on the mailing list for the consensus.

Regarding "Starting in OS X 10.11, each app has its own sharedHTTPCookieStorage singleton, which will not be shared with other applications"  I will modify the current implementation on Linux to match that of Darwin.

Thanks!

ahti:
Sorry, I've been quite busy the past week.

@mamabusi Sounds good. Should I still file a bug in jira?

mamabusi:
@ahti  Yes, please go ahead and file a bug for this in JIRA. Thanks!

ahti:
I've split up my comment and filed [two](https://bugs.swift.org/browse/SR-4905) [bugs](https://bugs.swift.org/browse/SR-4906).

PR #891, opened 2017-02-22
Possible fix for a memory leak [SR-4036]

pushkarnk:
This is related to https://bugs.swift.org/browse/SR-4036

I've documented all my findings in the JIRA report. 

This may not be the best way to fix this problem. I'm keen to know what others think

### DIFF
--- a/Foundation/NSStringAPI.swift
+++ b/Foundation/NSStringAPI.swift
@@ -371,7 +371,10 @@ extension String {
   public func cString(using encoding: Encoding) -> [CChar]? {
     return withExtendedLifetime(_ns) {
       (s: NSString) -> [CChar]? in
-      _persistCString(s.cString(using: encoding.rawValue))
+      let cString = s.cString(using: encoding.rawValue)
+      let cchars = _persistCString(cString)
+      free(UnsafeMutableRawPointer(mutating: cString))
+      return cchars
     }
   }
 


parkera:
What does `_persistCString` do exactly?

On the face of it this seems like it will just result in a use-after-free. Getting the cString without having the ObjC autoreleasepool in place is probably not going to work well.

pushkarnk:
Hi @parkera :
`_persistCString` creates a new `[CChar]`, copies contents of the memory (which we `malloc`'d) pointed by the UnsafePointer passed to it into this array and then returns the array. So, I guess after we return from here, there's no need to hold on to the malloc'd memory. The `free` could also be called here but other callers of `_persistCString` may not expect this behaviour.

parkera:
Hm, I'm not really a fan of this approach for a couple of reasons:

* It makes this file different across darwin and Linux, because on darwin that C string is backed by an autoreleased NSData and it would be a serious error to free it.
* It leaves the leak in place at the source of the problem, which is `NSString`'s `cString` (and `utf8string`, which we do nothing about here).

It seems like we should elide `cString` and `utf8String` from `NSString` on Linux, because there is no choice but for them to leak, and implement this method in a way which is more canonical (perhaps `data:usingEncoding:` followed by a copy).

pushkarnk:
I agree @parkera . Closing this PR. Let's continue this discussion on https://bugs.swift.org/browse/SR-4036 
Thanks!

PR #893, opened 2017-02-23
Added basic implementation for NSString contentsOfURL:usedEncoding:

mohitathwani:
In NSString, for contentsOfURL:usedEncoding: checking the BOM of the received data and determining whether encoding to be used should be UTF 16 BE or UTF 16 LE


mohitathwani:
@swift-ci please test

Foundation/NSString.swift
```diff
@@ -1256,7 +1256,33 @@ extension NSString {
...
+        }
+        else if bytePtr[0] == 255 && bytePtr[1] == 254 {
+          enc?.pointee = String.Encoding.utf16LittleEndian.rawValue
+        }
+        else {
+          //Need to work on more conditions. This should be the default
+          enc?.pointee = String.Encoding.utf8.rawValue
+        }
+
+        guard let cf = CFStringCreateWithBytes(kCFAllocatorDefault, bytePtr, readResult.length, CFStringConvertNSStringEncodingToEncoding(enc!.pointee), true) else {
```
alblue: This has a guard here, but uses enc! in the dereferencing of the pointer. It would make sense to have let enc = enc, cf = ... enc, so that if enc isn't passed then we don't blow up the process.

TestFoundation/TestNSString.swift
```diff
@@ -66,7 +66,8 @@ class TestNSString : XCTestCase {
             ("test_rangeOfCharacterFromSet", test_rangeOfCharacterFromSet ),
             ("test_CFStringCreateMutableCopy", test_CFStringCreateMutableCopy),
             ("test_FromContentsOfURL",test_FromContentsOfURL),
-            ("test_FromContentOfFile",test_FromContentOfFile),
+            ("test_FromContentsOfURLUsedEncodingUTF16BE", test_FromContentsOfURLUsedEncodingUTF16BE),
+	    ("test_FromContentOfFile",test_FromContentOfFile),
```
alblue: Indentation looks off by one character on this line.

parkera:
@swift-ci please test

mohitathwani:
@parkera @alblue I'm wondering if any of you could help me here. The test on the CI server fails with error:
`TestFoundation/TestNSString.swift:305: error: TestNSString.test_FromContentsOfURLUsedEncodingUTF16BE : failed - URL for NSString-UTF16-BE-data.txt is nil`

I have added this file to TestFoundation/Resources/NSString-UTF16-BE-data.txt where all other resource files are for NSString test cases. Does anything else have to be changed somewhere else?

mohitathwani:
@swift-ci please test

mohitathwani:
@swift-ci please test

parkera:
@swift-ci please test

zmeyc:
```
 +        let bytePtr = readResult.bytes.bindMemory(to: UInt8.self, capacity:readResult.length)
 +        if bytePtr[0] == 254 && bytePtr[1] == 255 {
 +          enc?.pointee = String.Encoding.utf16BigEndian.rawValue
 +        }
```

If readResult.length is less than 2, unitialized memory will be referenced.

And a possible memory leak: cf is created by CFStringCreateWithBytes, but not released when exitting on errors. AFAIK ARC doesn't handle this?

mohitathwani:
@zmeyc I thought about that too but there is an else which sets the default encoding to UTF8 and proceed to try decoding the file which could be less than 2 bytes with UTF 8 encoding.

zmeyc:
@mohitathwani Referencing unitialized memory is undefined behavior, it will trigger tools like valgrind, unitialized memory can contain BOM with wrong byte order etc. I propose checking the length before analyzing these two bytes for BOM.

mohitathwani:
@zmeyc Updated the code with your suggestion.
@parkera could you please trigger a test?

mohitathwani:
I feel this is ready for a merge

mohitathwani:
Thanks @parkera ! Do you think this is merge worthy?

parkera:
Looks good, thanks!

PR #894, opened 2017-02-24
Updated NumberFormatter to match expected behavior

ddunn2:
Made some updates to the NumberFormatter class, now the usesSignificantFigures property will only be set if you want to set either minimum or maximum significant figures and will prevent setting numberStyle to .decimal overwriting the values you set for minimum or maximum significant figures. 

I ran the tests and they all passed, I also added a few new tests and made the necessary changes to the old tests to account for this change.

### DIFF
--- a/CoreFoundation/Collections.subproj/CFData.c
+++ b/CoreFoundation/Collections.subproj/CFData.c
@@ -312,10 +312,13 @@ CFTypeID CFDataGetTypeID(void) {
 void _CFDataInit(CFMutableDataRef memory, CFOptionFlags flags, CFIndex capacity, const uint8_t *bytes, CFIndex length, Boolean noCopy) {
     Boolean isMutable = ((flags & __kCFMutable) != 0);
     Boolean isGrowable = ((flags & __kCFGrowable) != 0);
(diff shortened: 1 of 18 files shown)


parkera:

ianpartridge:
TestFoundation/TestNSByteCountFormatter.swift:368: error: TestNSByteCountFormatter.test_isAdaptiveFalseZeroPadsFractionDigitsTrue : XCTAssertEqual failed: ("123 MB") is not equal to ("123.5 MB") -

ddunn2:
Thanks Ian, I imagine it's due to the ByteCountFormatter is currently configured to work with how the NumberFormatter use to be prior to this update, and just needs updating to work this updated version of the NumberFormatter but I'll look into it to confirm and update as needed.

ddunn2:
I've confirmed it's due to how the new NumberFormatter update works, I will update the ByteCountFormatter and add changes to a new commit. I'll also tidy up the ByteCountFormatter as a few bits can be removed with this update to NumberFormatter as well.

ddunn2:
@parkera @swift-ci  Changes have been added, I ran tests on mac and linux and both passed.

parkera:
Cool, let me kick this off again.

parkera:

PR #895, opened 2017-02-25
Enable URLSession tests in TestFoundation

pushkarnk:
URLSessions tests had been disabled because of a libkqueue failure seen in CI. See: https://bugs.swift.org/browse/SR-3464

Given that we've done away with libkqueue in Dispatch and SR-3464 is no longer relevant, can we re-enable the URLSession tests in TestFoundation? 

cc @gparker42

### DIFF
--- a/TestFoundation/main.swift
+++ b/TestFoundation/main.swift
@@ -77,8 +77,7 @@ XCTMain([
     testCase(TestURLRequest.allTests),
     testCase(TestNSURLResponse.allTests),
     testCase(TestNSHTTPURLResponse.allTests),
-    // FIXME: SR-3464 Assertion failure in TestURLSession.test_dataTaskWithURL
-    // testCase(TestURLSession.allTests),
+    testCase(TestURLSession.allTests),
     testCase(TestNSNull.allTests),
     testCase(TestNSUUID.allTests),
     testCase(TestNSValue.allTests),


parkera:
@swift-ci please test

parkera:
Yah, let's make sure this passes at least once then re-enable it for good. If it starts failing intermittently again we can re-disable it.

alblue:
The only annoying thing about this test is that when run on Darwin the triggering of running the HTTP server (that is fired up to test the client) invokes the macOS firewall asking if you want to allow/deny connections. Can we make it conditional so that it doesn't run when kicked off from a desktop build in some way?

pushkarnk:
Unrelated failure. Re-triggering test. 

@swift-ci please test

pushkarnk:
@alblue For that, we'd need to figure out how to disable the firewall in code. Btw, I came across a command that can enable/disable the firewall here: http://krypted.com/mac-os-x/command-line-alf-on-mac-os-x/

alblue:
Or we could just not run the test inside Xcode?

pushkarnk:
@alblue I am OK with that. What do you think @parkera ?

naithar:
I think that if tests will be disabled for XCode, there should be at least some notification that they are disabled, since `URLSession` is not fully implemented.

Also, using commands from the link didn't give any results on disabling/enabling firewall, but maybe it's just me.

pushkarnk:
I think we could continue the discussion around the proposal for disabling URLSession tests in Xcode over the mailing list.

pushkarnk:

PR #898, opened 2017-02-28
Temporary workaround for Crash in Date.description[SR-2462]

saiHemak:


parkera:
I think the fundamental questions I had about this patch are still there.

Where else in Foundation do we need to do this to avoid a crash? Or is it specific to time zone for some reason? Maybe it's a CFTimeZone bug if so, and we should fix it there.

spevans:
@saiHemak Have you been able to reproduce this crash lately? I filed the original bug SR-2462 but I have just tried to recreate the test cases in the original jira and I cant reproduce the crash anymore use the latest master or 3.1 snapshots.

@parkera If I recall correctly the original issue was not specific to CFTimeZone but it was due to a swift compiler bug in the generation of an extra release under some circumstances. I will have another look at the testcase when I get some time but it may well be that the issue has been fixed unless there is still a known issue.

saiHemak:
@spevans  yes its resolved with latest master .Thank you .

PR #905, opened 2017-03-04
NSMutableData compatiblity with macOS

johnno1962:
Came across this using NSMutableData on Linux. Replace range with nil pointer and length 0 does nothing when it should delete bytes in the range.


parkera:
Would you mind adding a small test case?

johnno1962:
Best I can do at the moment.. I don’t have a Linux build available to test the test.

parkera:

johnno1962:
Failed?

parkera:
Looks like we're seeing that error in a few places.

parkera:

PR #903, opened 2017-03-03
[gardening]print warning only if URLSessionDebug is set

pushkarnk:
There is only one place in URLSession where we are printing a warning message during normal execution. We've had a user complain about this message creating a `lot of noise`. I'm moving this warning message under the control of the `URLSessionDebug` env variable.

PR #910, opened 2017-03-07
[cherry-pick][urlsession]pull request #903 from pushkarnk/print-debug

pushkarnk:
@parkera Though this is not critical, we have complains about this message causing a lot of noise. Can we have it in 3.1 please?


parkera:

PR #911, opened 2017-03-07
[URLSession]Fix for a timeout bug(SR-2681)

pushkarnk:
Currently, we use the `URLSessionConfiguration.timeoutIntervalForRequest` value to trigger a timeout. https://bugs.swift.org/browse/SR-2681 reported a bug in which the timeout counter is not reset on every data transfer. 

This is from the API doc:
```
The request timeout interval controls how long (in seconds) a task should wait for additional data to
 arrive before giving up. The timer associated with this value is reset whenever new data arrives. 
When the request timer reaches the specified interval without receiving any new data, it triggers a timeout.
```

This patch tries to reuse the `_TimeoutSource` class which is currently used to trigger timeouts in MultiHandle, with some modifications. We reset the timer every time data is available.

### DIFF
--- a/Foundation/NSURLSession/NSURLSessionTask.swift
+++ b/Foundation/NSURLSession/NSURLSessionTask.swift
@@ -562,7 +562,14 @@ fileprivate extension URLSessionTask {
         //set the request timeout
         //TODO: the timeout value needs to be reset on every data transfer
         let s = session as! URLSession
-        easyHandle.set(timeout: Int(s.configuration.timeoutIntervalForRequest))
+        let timeoutInterval = Int(s.configuration.timeoutIntervalForRequest) * 1000
+        let timeoutHandler = DispatchWorkItem { [weak self] in
+            guard let currentTask = self else { fatalError("Timeout on a task that doesn't exist") } //this guard must always pass
+            currentTask.internalState = .transferFailed
+            let urlError = URLError(_nsError: NSError(domain: NSURLErrorDomain, code: NSURLErrorTimedOut, userInfo: nil))
+            currentTask.completeTask(withError: urlError)
+        }
+        easyHandle.timeoutTimer = _TimeoutSource(queue: workQueue, milliseconds: timeoutInterval, handler: timeoutHandler)
 
         easyHandle.set(automaticBodyDecompression: true)
         easyHandle.set(requestMethod: request.httpMethod ?? "GET")
@@ -823,6 +830,9 @@ extension URLSessionTask {
         }
         self.response = response
 
+        //We don't want a timeout to be triggered after this. The timeout timer needs to be cancelled.
+        easyHandle.timeoutTimer = nil
+
         //because we deregister the task with the session on internalState being set to taskCompleted
         //we need to do the latter after the delegate/handler was notified/invoked
         switch session.behaviour(for: self) {
@@ -874,6 +884,10 @@ extension URLSessionTask {
         guard case .transferFailed = internalState else {
             fatalError("Trying to complete the task, but its transfer isn't complete / failed.")
         }
+
+        //We don't want a timeout to be triggered after this. The timeout timer needs to be cancelled.
+        easyHandle.timeoutTimer = nil
+
         switch session.behaviour(for: self) {
         case .taskDelegate(let delegate):
             guard let s = session as? URLSession else { fatalError() }
--- a/Foundation/NSURLSession/EasyHandle.swift
+++ b/Foundation/NSURLSession/EasyHandle.swift
@@ -56,6 +56,8 @@ internal final class _EasyHandle {
     fileprivate var headerList: _CurlStringList?
     fileprivate var pauseState: _PauseState = []
     internal var fileLength: Int64 = 0
+    internal var timeoutTimer: _TimeoutSource!
+
     init(delegate: _EasyHandleDelegate) {
         self.delegate = delegate
         setupCallbacks()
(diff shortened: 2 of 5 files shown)


pushkarnk:
With the current URLSession testing infrastructure, a test case for this, is a challenge.

pushkarnk:
@parkera @naithar Can you please review and comment?

naithar:
I have no comments on implementation, except for adding ``guard `self` = self else { return }`` inside `DispatchWorkerItem` closure. But i'm not really sure if it's actually affects anything

As for testing `TestURLSessionServer` initializer could receive a delay interval for response, like `init(port: UInt16, delay: TimeInterval? = nil)` which could be used by `.readAndRespond()` method to delay actual response. 
Here's the `TestURLSessionServer` class with the changes that seem to correctly implement the timeout testing (I've changed `nil` default to `5` to test it): https://gist.github.com/naithar/7f26c24dd08c312858d13d4f5e97f728

naithar:
Oh, now I see that i didn't understand the problem with correctly.

pushkarnk:
Thanks @naithar I have added the guard

naithar:
@pushkarnk 
Again about testing. Seems like i've started digging in a right place actually.
If we can change a `func writeData(data: String)` method to split resulted string in chunks and send it with a delay this could emulate the request to `http://httpbin.org/drip?duration=10&numbytes=10&code=200` which is used in bug report. But currently i managed only to send multiple `data` parameter in a single response.

```swift
var bytes = Array(data.utf8)


for _ in 0..<5 {
    sleep(1)
    _  = try? attempt("write", valid: isNotNegative, CInt(write(connectionSocket, &bytes, bytes.count)))
}

```

Foundation/NSURLSession/NSURLSessionTask.swift
```diff
@@ -562,7 +562,14 @@ fileprivate extension URLSessionTask {
         //set the request timeout
         //TODO: the timeout value needs to be reset on every data transfer
         let s = session as! URLSession
-        easyHandle.set(timeout: Int(s.configuration.timeoutIntervalForRequest))
+        let timeoutInterval = Int(s.configuration.timeoutIntervalForRequest) * 1000
+        let timeoutHandler = DispatchWorkItem { [weak self] in
+                guard let currentTask = self else { fatalError("Timeout on a task that doesn't exist") } //this guard must always pass
```
naithar: Should it actually crash if we don't have `self`? Could it actually happen?

parkera:
The diff looks reasonable but we should be able to get some kind of unit test in place for this.

pushkarnk:
@naithar Thanks for your work. I was hoping that we can also delay the response as in:
```
    public func respond(with response: _HTTPResponse, delay: Int = 0) throws {
        let semaphore = DispatchSemaphore(value: 0)
        let deadlineTime = DispatchTime.now() + .seconds(delay)
        DispatchQueue.main.asyncAfter(deadline: deadlineTime) {
            do {
                try self.socket.writeData(data: response.description)
                semaphore.signal()
            } catch { }
        }
        semaphore.wait()
     }
```
I did this and I see a crash. Investigating it now.

naithar:
Does crash message anything? Maybe "Trying to access a behaviour for a task that in not in the registry."? I can reproduce this by just delaying response for long enough.

naithar:
Not sure what happened, but reapplying a PR to current `master` removed the crash.
[This commit](https://github.com/naithar/swift-corelibs-foundation/commit/9080ed8747b14262bf2b55e29a70b7566a928e9a) implements a chunked response, so this PR could be tested. It's a little ugly but it works.

pushkarnk:
@naithar Yes, I did see that fatal error. Nevertheless, your patch looks good and runs well. I am wondering how we go about. I'd suggest we have this PR merged first and then you create a PR with your patch.

pushkarnk:
I also realised that I need to cancel the timer when `completeTask()` is called.

naithar:
@pushkarnk i can also create a PR with test commented out. Actually i'm not agains you taking this implementation into this PR.

pushkarnk:
You creating the PR with calls to the new tests commented out is a better idea, in my opinion :)

naithar:
Then I'll do it now.

naithar:
Submitted #913

pushkarnk:
I can consistently see the "Trying to access a behaviour for a task that in not in the registry." failure with @naithar 's test case from #913

naithar:
That's strange. 
I can see same failure. But moving `test_taskTimeout` test to top actually solves the issue.

I'll look into it a little more.

pushkarnk:
Yes, looks like we need to cancel the timeoutTimer in `complete(withError:)` too.

naithar:
Actually `test_taskTimeout` causes failure only if it follows `test_cancelTask` test. Atm I'm not sure what exactly causing the failure.

naithar:
Running Cancel test actually causes this error for `TestURLSessionServer`:
```
bind failed: Unknown error: -1 (-1)
  - operation : "bind"
  - errno : -1
  - file : "/Volumes/MacintoshHDFiles/Documents/Projects/swift-core/swift-corelibs-foundation/TestFoundation/HTTPServer.swift"
  - line : 69
```

It's not displayed or handled because server is starting with this code (the problem is in `XCTAssertTrue(true)`), so it won't show anything:
```swift
globalDispatchQueue.async {
            do {
                try self.runServer(with: serverReady)
            } catch {
                XCTAssertTrue(true)
                return
            }
        }
```

Also not signaling semaphore in `catch` in `runServer(...)` method causes tests to hang sometimes.

naithar:
Switching 
```swift
} catch let e as ServerError {
    if e.operation != "bind" { continue }
```

with

```swift
} catch let e as ServerError {
    if e.operation == "bind" { continue }
```

in `runServer(...)` method seems like a workaround and removes the error and crashes.

pushkarnk:
Thanks @naithar That seems like a fix (not a workaround). We  retry on a bind failure and throw an error otherwise.

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci please test

pushkarnk:

pushkarnk:

PR #914, opened 2017-03-09
Fix for [SR-1250] NSJSONSerialization emits non-floating-point number…

saiHemak:
JSON serialization behaves differently on Linux.

eg : Serializing  
```let subject = "[12.1,10.0,0.0]" ```

on Darwin  gives - 12.1,10,0 
where as on Linux - 12.1,10.0,0.0  //.0 needs to be truncated

This fix aligns the ```JSONSerialization.jsonObject(with: data)```  behavior  to Darwin.

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -716,21 +716,24 @@ private struct JSONReader {
                 defer { intEndPointer.deallocate(capacity: 1) }
                 let doubleEndPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
                 defer { doubleEndPointer.deallocate(capacity: 1) }
-                
                 let intResult = strtol(startPointer, intEndPointer, 10)
                 let intDistance = startPointer.distance(to: intEndPointer[0]!)
                 let doubleResult = strtod(startPointer, doubleEndPointer)
                 let doubleDistance = startPointer.distance(to: doubleEndPointer[0]!)
-                
+
                 guard intDistance > 0 || doubleDistance > 0 else {
                     return nil
                 }
-                
+
                 if intDistance == doubleDistance {
                     return (intResult, intDistance)
                 }
                 guard doubleDistance > 0 else {
                     return nil
+                }
+
+                if doubleResult == doubleResult.rounded() {
+                    return (Int(doubleResult), doubleDistance)
                 }
                 return (doubleResult, doubleDistance)
             }
--- a/TestFoundation/TestNSJSONSerialization.swift
+++ b/TestFoundation/TestNSJSONSerialization.swift
@@ -468,7 +468,7 @@ extension TestNSJSONSerialization {
                 XCTAssertEqual(result?[1] as? Int,       -1)
                 XCTAssertEqual(result?[2] as? Double,   1.3)
                 XCTAssertEqual(result?[3] as? Double,  -1.3)
-                XCTAssertEqual(result?[4] as? Double,  1000)
+                XCTAssertEqual(result?[4] as? Int,     1000)
                 XCTAssertEqual(result?[5] as? Double, 0.001)
             }
         } catch {
@@ -871,6 +871,7 @@ extension TestNSJSONSerialization {
             ("test_jsonObjectToOutputStreamInsufficientBuffer", test_jsonObjectToOutputStreamInsufficientBuffer),
             ("test_booleanJSONObject", test_booleanJSONObject),
             ("test_serialize_dictionaryWithDecimal", test_serialize_dictionaryWithDecimal),
+            ("test_serializeDecimalNumberJSONObject", test_serializeDecimalNumberJSONObject),
         ]
     }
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by alblue** [comment]
@swift-ci please test

**Comment by alblue** [comment]
Wouldn't this change impact existing code, which would assume that such numbers are doubles? Does this behaviour also exist on Darwin platforms?

**Comment by saiHemak** [comment]
@alblue : Yeah ..That's what I observed on Darwin ..I updated the description with  details

**Comment by parkera** [comment]
This is all tied up in the same discussion about what the proper role of NSNumber is in Swift (cc @phausler), but in the meantime it looks reasonable to me.

File: Foundation/NSJSONSerialization.swift [hunk 1]
```diff
@@ -716,21 +716,16 @@ private struct JSONReader {
...
                 guard intDistance > 0 || doubleDistance > 0 else {
                     return nil
                 }
-                
-                if intDistance == doubleDistance {
-                    return (intResult, intDistance)
-                }
-                guard doubleDistance > 0 else {
-                    return nil
+                if doubleResult == roundedDouble || doubleResult == 0.0 {
```
xwu: Why is the second half of this `if` statement necessary? Shouldn't this be `if doubleResult == doubleResult.rounded()`?

Foundation/NSJSONSerialization.swift
```diff
@@ -716,21 +716,16 @@ private struct JSONReader {
...
                 let intDistance = startPointer.distance(to: intEndPointer[0]!)
                 let doubleResult = strtod(startPointer, doubleEndPointer)
                 let doubleDistance = startPointer.distance(to: doubleEndPointer[0]!)
-                
+                let roundedDouble = doubleResult.rounded()
                 guard intDistance > 0 || doubleDistance > 0 else {
                     return nil
                 }
-                
-                if intDistance == doubleDistance {
```
xwu: Removing these lines seems suspect. First, `intResult` is now unused. Second--more importantly--there are values of `Int` that cannot be exactly represented in `Double` (those above 2e53). Was there a reason why you removed these lines?

You will almost certainly want the second `guard` statement below before proceeding to use `doubleResult`.

TestFoundation/TestNSJSONSerialization.swift
```diff
@@ -468,7 +468,7 @@ extension TestNSJSONSerialization {
                 XCTAssertEqual(result?[1] as? Int,       -1)
                 XCTAssertEqual(result?[2] as? Double,   1.3)
                 XCTAssertEqual(result?[3] as? Double,  -1.3)
-                XCTAssertEqual(result?[4] as? Double,  1000)
+                XCTAssertEqual(result?[4] as? Int,  1000)
```
xwu: Nit: please follow file convention and align `1000` with the line above.

TestFoundation/TestNSJSONSerialization.swift
```diff
@@ -1214,7 +1215,22 @@ extension TestNSJSONSerialization {
         XCTAssertTrue(JSONSerialization.isValidJSONObject([true]))
     }
 
-    fileprivate func createTestFile(_ path: String,_contents: Data) -> String? {
+    func test_serializeDecimalNumberJSONObject() {
+        let decimalArray = "[12.1,10.0,0.0,0.0001,20]"
+        do {
+            let data = decimalArray.data(using: String.Encoding.utf8)
+            let result = try JSONSerialization.jsonObject(with: data!, options: []) as? [Any]
+            XCTAssertEqual(result?[0] as! Double,12.1)
```
xwu: Nit: can you put a space after the comma here and on subsequent lines?

TestFoundation/TestNSJSONSerialization.swift
```diff
@@ -1214,7 +1215,22 @@ extension TestNSJSONSerialization {
...
+            XCTAssertEqual(result?[1] as! Int,10)
+            XCTAssertEqual(result?[2] as! Int, 0)
+            XCTAssertEqual(result?[3] as! Double,0.0001)
+            XCTAssertEqual(result?[4] as! Int,20)
+        } catch {
+            XCTFail("Failed during serialization")
+        }
+    }
+
+   fileprivate func createTestFile(_ path: String,_contents: Data) -> String? {
```
xwu: Please restore the original indent of four spaces and not three here.

Foundation/NSJSONSerialization.swift
```diff
@@ -716,21 +716,16 @@ private struct JSONReader {
                 defer { intEndPointer.deallocate(capacity: 1) }
                 let doubleEndPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
                 defer { doubleEndPointer.deallocate(capacity: 1) }
-                
                 let intResult = strtol(startPointer, intEndPointer, 10)
                 let intDistance = startPointer.distance(to: intEndPointer[0]!)
                 let doubleResult = strtod(startPointer, doubleEndPointer)
                 let doubleDistance = startPointer.distance(to: doubleEndPointer[0]!)
-                
+                let roundedDouble = doubleResult.rounded()
```
xwu: At minimum, this is best moved after the `guard doubleDistance > 0` statement that should be restored below. As mentioned earlier, you may not need to have this line at all.

Foundation/NSJSONSerialization.swift
saiHemak: @xwu  Thanks for the review.  I agree with you .I have missed to test with huge integers and thought that it would be redundant .We need to retain the intResult too and added the check to handle  of rounded doubles . I have modified the fix now ..

pushkarnk:
@swift-ci please test

saiHemak:
@phausler Please review

phausler:
This seems reasonable, but for compatability it is worth noting that NSNumbers returned are a bit strange in Swift.

```swift
let n = NSNumber(value: 4.3)
if n is Int {
    // this is true as of the current build of Swift on Darwin... which imho seems a bit incorrect.
}
```

PR #917, opened 2017-03-10
Remove final from vars in error related protocol extensions

KingOfBrian:
This PR removes `final` from a few `var`s in the `__BridgedNSError` protocol. SR-1762 / apple/swift#8010 will generate a warning or error when `final` is present in a protocol extension. This PR will remove the warning, and will not modify dispatch behavior in any way.


parkera:
LGTM, test/merge when you are ready to coordinate with the Swift change.

KingOfBrian:
These changes are compatible with Swift 3.x so it doesn't need any coordination. Can you test and merge? The magic words don't work for me!

alblue:
@swift-ci please test

PR #918, opened 2017-03-14
[SR-4242] Add NSMutableOrderedSet subscript setter that is missing from non-Darwin

rxwei:
`NSMutableOrderedSet` has a subscript setter imported to Darwin platforms, but the setter doesn't exist on Linux. 

This PR adds a subscript setter to `NSMutableOrderedSet` as a shortcut to `replaceObject(at:with:)`


parkera:
@swift-ci test and merge

PR #921, opened 2017-03-17
[test] Temporarily disable TestNSNotificationQueue pending rdar://31106944

gparker42:
These tests are failing intermittently in CI.


gparker42:
@swift-ci please test

parkera:
Why did this start failing? I don't recall any changes in this area. Did something else change underneath us?

parkera:
Maybe the question is actually when this started failing, from which we can try to determine why.

gparker42:
@swift-ci please test

PR #922, opened 2017-03-23
Support for custom protocols in URLSession [WIP]

nethraravindran:
This is **work in progress**

The implementation of URLSession does not seem to be supporting custom protocols. Reference is added in the test case ` test_customProtocol `. For supporting custom URL protocols we need the following : 

- [ ] Implementing URLProtocol

- [ ] Implementing URLProtocolClient

- [ ] Adding support for custom protocols in URLSession

I have begun work on the above (with few methods implemented) and would keep this PR updated with the progress. I will highly appreciate any inputs or feedback.

### DIFF
--- a/Foundation/NSURLProtocol.swift
+++ b/Foundation/NSURLProtocol.swift
@@ -142,6 +142,25 @@ public protocol URLProtocolClient : NSObjectProtocol {
     func urlProtocol(_ protocol: URLProtocol, didCancel challenge: URLAuthenticationChallenge)
 }
 
+internal class ProtocolClient : NSObject, URLProtocolClient {
+
+    func urlProtocol(_ protocol: URLProtocol, didReceive response: URLResponse, cacheStoragePolicy policy: URLCache.StoragePolicy) { `protocol`.task?.setResponse = response }
+
+    func urlProtocolDidFinishLoading(_ protocol: URLProtocol) { `protocol`.task?.completeTask() }
+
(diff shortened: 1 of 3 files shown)


pushkarnk:
If I get it right the HTTP/HTTPS support that we currently have should also have a URLProtocol-based entry point (something like an `NSURLHTTPProtocol`). Currently, the URLSessionTask code is very tightly coupled with libcurl abstractions that are very specific to HTTP/HTTPS. In my opinion, URLSessionTask should have a protocol agnostic implementation. I think we'd need to refactor some of that code.

nethraravindran:
Provided a default implementation for the methods ` urlProtocol(_ protocol: URLProtocol, didReceive response: URLResponse , cacheStoragePolicy policy: URLCache.StoragePolicy?) ` and  `urlProtocolDidFinishLoading `

nethraravindran:
Here is the working prototype for URLProtocol https://github.com/apple/swift-corelibs-foundation/compare/master...nethraravindran:urlprotocol-prototype-branch 
With this prototype the test case seems to be successful.

URLSessionTask seems to be tightly coupled with EasyHandle. All the functions in URLSessionTask depend on the the state of ` internalState` which is tightly bound to EasyHandle. In order to decouple URLSession Task and EasyHandle, we will need to have a class that acts as an interface between URLSessionTask and EasyHandle.

nethraravindran:
Closing this PR for now since this requires a lot of refactoring in URLSessionTask and I'm working on it.

PR #923, opened 2017-03-23
Added fast path for Int and UInt

ddunn2:
As a follow on from #723 I've created fast path options for Int and UInt to further improve the performance of JSON Serialization. 

Below are some results I obtained by using a slightly modified version of the [benchmark](https://github.com/djones6/swift-corelibs-foundation/commit/867109e253550e4d1bcb99cfc681fdb02ce0f566) @djones6 developed, increased to 1000 repetitions and varying the payload size. I've included times for `String(describing:)` as that was the original proposed solution in #723 and felt it would offer a good comparative view.  

| Payload type:              |  Average Time (2 digit numbers) | Average Time (10 digit numbers)        | 
|------|:-------:|:------:|
| Baseline (current `master`)     | 1.932 | 2.631
| With String(describing:) (as rejected in #723 ) | 0.877 | 1.572 |
| Int/UInt fast Path (this PR) | 0.984 (2.0x faster than baseline) | 1.653 (1.6x faster than baseline) |

@djones6 and I also ran some tests against Kitura, these tests showed we are slightly better than `String(describing:)` for shorter numbers and on par with `String(describing:)` for longer numbers.

I've also added a few extra tests for serialization of the min/max values for Int and UInt as I felt these offered good values for testing the correct result is produced, also the current master implementation fails on max UInt. 

I haven't started work on Double/Floats yet but may do so if this fix is acceptable.

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -309,6 +309,10 @@ private struct JSONWriter {
             try serializeString(str)
         case let boolValue as Bool:
             serializeBool(boolValue)
+        case let num as Int:
+            try serializeInt(value: num)
+        case let num as UInt:
+            try serializeUInt(value: num)
         case let array as Array<Any>:
             try serializeArray(array)
         case let dict as Dictionary<AnyHashable, Any>:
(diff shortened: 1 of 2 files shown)


Foundation/NSJSONSerialization.swift
```diff
@@ -322,6 +326,93 @@ private struct JSONWriter {
...
 
+    private func serializeUInt(value: UInt) throws {
+        if value == 0 {
+            writer("0")
+            return
+        }
+        var array: [UInt] = []
+        var stringResult = ""
+        //Maximum length of an UInt
+        array.reserveCapacity(20)
```
parkera: This value of `20` seems magical. Can we formalize it somehow?

Foundation/NSJSONSerialization.swift
```diff
@@ -322,6 +326,93 @@ private struct JSONWriter {
...
+    
+    private func serializeInt(value: Int) throws {
+        if value == 0 {
+            writer("0")
+            return
+        }
+        var array: [Int] = []
+        var stringResult = ""
+        //Maximum length of an Int
+        array.reserveCapacity(19)
```
parkera: Same comment here.

TestFoundation/TestNSJSONSerialization.swift
```diff
@@ -1054,6 +1058,26 @@ extension TestNSJSONSerialization {
         XCTAssertEqual(try trySerialize(json), "[false,true]")
     }
     
+    func test_serialize_IntMax() {
+        let json: [Any] = [Int.max]
```
parkera: Is `Int.max` architecture specific?

Foundation/NSJSONSerialization.swift
```diff
@@ -322,6 +326,93 @@ private struct JSONWriter {
...
+            case 3, -3: stringResult.append("3")
+            case 4, -4: stringResult.append("4")
+            case 5, -5: stringResult.append("5")
+            case 6, -6: stringResult.append("6")
+            case 7, -7: stringResult.append("7")
+            case 8, -8: stringResult.append("8")
+            case 9, -9: stringResult.append("9")
+            default: fatalError()
+            }
+        }
```
pushkarnk: Though the loops where we walk backwards are not exactly identical in the two functions, I am wondering if we can still refactor them into a single function.

Foundation/NSJSONSerialization.swift
ianpartridge: I think we could make these constants like this:

```swift
let maxUIntLength = String(describing: UInt.max).characters.count
let maxIntLength = String(describing: Int.max).characters.count
```

We'd need to calculate these once (not every time we serialize), so make them private properties of `NSJSONSerialization`.

TestFoundation/TestNSJSONSerialization.swift
ianpartridge: Yes. [Documentation](https://developer.apple.com/reference/swift/int):
> On 32-bit platforms, Int is the same size as Int32, and on 64-bit platforms, Int is the same size as Int64.

So I think this will break on 32-bit platforms.  We should construct the string we're asserting against using the same constant, I think.  The same applies for the tests which use `Int.min`, `UInt.max` and `UInt.min`.

ddunn2: Thanks for finding that, I've updated the tests to compare against a string representation of the `Int.max` value so this should also work on 32-bit platforms now.

Foundation/NSJSONSerialization.swift
ddunn2: I've also made these changes, you can find them in the new commit. 
Is this an acceptable way of formalizing it @parkera ?

Foundation/NSJSONSerialization.swift
ddunn2: Thanks for the feedback, it is possible to refactor them into a single function however both Ian and myself felt the code looked more complicated and required a little more work in terms of casting and such, so I've omitted it for now.

ddunn2:
@parkera Following your comments I've made some changes, are these changes acceptable? Thanks.

parkera:
Thanks!

parkera:
@swift-ci test and merge

PR #924, opened 2017-03-23
NSRegularExpression: Add default value to init() `options` parameter

ianpartridge:
According to the [documentation](https://developer.apple.com/reference/foundation/nsregularexpression/1410900-init), the `options` parameter to the `NSRegularExpression` initializer should have a default value of `[]`.  This was not correct in swift-corelibs-foundation.

Correct the signature of the initializer and update one of the existing tests to cover this case.


pushkarnk:
Verified with the API documentation.

PR #925, opened 2017-03-28
Bugfix: newly assigned pausingHandler was called if the Progress is already cancelled

victor-pavlychko:
Looks like it should be called when we are already paused instead.


parkera:
@swift-ci test and merge

parkera:
thanks!

ianpartridge:
Build timed out after 30 mins.

alblue:

PR #927, opened 2017-03-29
[SR-3463][URLSession] config.httpAdditionalHeaders isn't picked up

enasser:
Fix for https://bugs.swift.org/browse/SR-3463

### DIFF
--- a/Foundation/NSURLSession/NSURLSessionTask.swift
+++ b/Foundation/NSURLSession/NSURLSessionTask.swift
@@ -554,9 +554,31 @@ fileprivate extension URLSessionTask {
         
         // HTTP Options:
         easyHandle.set(followLocation: false)
+        
+        // The httpAdditionalHeaders from session configuration has to be added to the request.
+        // The request.allHTTPHeaders can override the httpAdditionalHeaders elements. Add the
+        // httpAdditionalHeaders from session configuration first and then append/update the
+        // request.allHTTPHeaders so that request.allHTTPHeaders can override httpAdditionalHeaders.
+        
+        let httpSession = session as! URLSession
+        var httpHeaders: [AnyHashable : Any]?
+        
+        if let hh = httpSession.configuration.httpAdditionalHeaders {
+            httpHeaders = hh
+        }
+        
+        if let hh = currentRequest?.allHTTPHeaderFields {
+            if httpHeaders == nil {
+                httpHeaders = hh
+            } else {
+                hh.forEach {
+                    httpHeaders![$0] = $1
+                }
+            }
+        }
 
         let customHeaders: [String]
-        let headersForRequest = curlHeaders(for: request)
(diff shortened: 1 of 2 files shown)


pushkarnk:
LGTM
@naithar Can you also a take a look please?

Foundation/NSURLSession/NSURLSessionTask.swift
```diff
@@ -597,10 +618,11 @@ fileprivate extension URLSessionTask {
     /// expects.
     ///
     /// - SeeAlso: https://curl.haxx.se/libcurl/c/CURLOPT_HTTPHEADER.html
-    func curlHeaders(for request: URLRequest) -> [String] {
+    func curlHeaders(for httpHeaders: [AnyHashable : Any]?) -> [String] {
         var result: [String] = []
         var names = Set<String>()
-        if let hh = currentRequest?.allHTTPHeaderFields {
+        if httpHeaders != nil {
+            let hh = httpHeaders as! [String:String]
```
naithar: Wouldn't it be better to use an actual `[AnyHashable : Any]` dictionary and turn it's keys and values into strings using `String(describing:)`? 
This would also require `var names = Set<AnyHashable>()`, but `AnyHashable` can handle strings.

enasser: @naithar Thank you for reviewing the changes and providing your inputs. I will analyze that and get back to you. Sorry for the delay in response as I was on vacation.

enasser: @naithar I agree with you that String(describing:) is the preferred way to convert an instance of any type to a string as per the documentation. I think it makes sense if we need to convert only specific elements in the dictionary. Here, we need to process each element in the dictionary and hence converting the dictionary from any type to a string dictionary makes the processing simpler.

We cannot use .lowercased() and .isEmpty methods with the elements of the [AnyHashable : Any] dictionary. Hence, we need to convert them to string before applying those methods.

In order to process the httpHeaders as any type, we need to make following changes.
        var names = Set<String>()
        if httpHeaders != nil {
           // let hh = httpHeaders as! [String:String]
            httpHeaders?.forEach {
                let name = String(describing:$0.0).lowercased()
                guard !names.contains(name) else { return }
                names.insert(name)
                
                if String(describing:$0.1).isEmpty {
                    result.append(String(describing:$0.0) + ";")
                } else {
                    result.append(String(describing:$0.0) + ": " + String(describing:$0.1))
                }
            }
        }
As the element values are already converting to string, I did not change the names variable.

Please let me know your thoughts whether the existing changes are fine or you have any better way.

naithar: This changes looks ok to me

pushkarnk:
Thanks @enasser 

@swift-ci test and merge

PR #928, opened 2017-04-02
Added code to detect UTF32 files for NSString contentsOfURL:usedEncoding

mohitathwani:
This pull request is in continuation to PR #893


alblue:
Please can you squash the history of this, so that we don't get noisy messages like "Resolved merge conflicts"? and instead of having "Merge remote-tracking branch X" messages, you can do a `git pull --rebase` (which will replay your commits on top of the latest, instead of trying to do a merge commit). At the very least, please rebase this all on the top of master, and then squash all commits with 'Resolved' or 'Merge' in the title.

You can also configure the 'rebase' to be the default by doing `git config --global pull.rebase true` so that you don't have the same problems in future.

phausler:
@alblue usually any merges we accept are squash merged anyhow as a default rule so that isn't too bad.

build.py
```diff
@@ -462,6 +462,8 @@
     'TestFoundation/Resources/NSStringTestData.txt',
     'TestFoundation/Resources/NSString-UTF16-BE-data.txt',
     'TestFoundation/Resources/NSString-UTF16-LE-data.txt',
+    'TestFoundation/Resources/NSString-UTF32-BE-data.txt',
```
phausler: are these files missing from the commit? I don't see them currently on master

TestFoundation/TestNSString.swift
```diff
@@ -334,6 +336,38 @@ class TestNSString : XCTestCase {
...
+      do {
+         var encoding: UInt = 0
+         let string = try NSString(contentsOf: testFileURL, usedEncoding: &encoding)
+         XCTAssertEqual(string, "NSString fromURL usedEncoding test with UTF32 LE file", "Wrong result when reading UTF32LE file")
+         XCTAssertEqual(encoding, String.Encoding.utf32LittleEndian.rawValue, "Wrong encoding detected from UTF32LE file")
+      } catch {
+          XCTFail("Unable to init NSString from contentOf:usedEncoding:")
+      }
+    }
+
```
phausler: Do these pass on macOS using the system Foundation?

Foundation/NSString.swift
```diff
@@ -1301,12 +1301,18 @@ extension NSString {
         let readResult = try NSData(contentsOf: url, options:[])
 
         let bytePtr = readResult.bytes.bindMemory(to: UInt8.self, capacity:readResult.length)
-        if readResult.length >= 2 && bytePtr[0] == 254 && bytePtr[1] == 255 {
+        if readResult.length >= 4 && bytePtr[0] == 255 && bytePtr[1] == 254 && bytePtr[2] == 0 && bytePtr[3] == 0 {
```
phausler: might be good to use the hex values for these since most sources of the unicode standards refer to 0xFF and 0xFE etc. (this is just a style nit not a functionality commentary)

Foundation/NSString.swift
```diff
@@ -1301,12 +1301,18 @@ extension NSString {
...
+        }
+        else if readResult.length >= 2 && bytePtr[0] == 254 && bytePtr[1] == 255 {
           enc?.pointee = String.Encoding.utf16BigEndian.rawValue
         }
         else if readResult.length >= 2 && bytePtr[0] == 255 && bytePtr[1] == 254 {
           enc?.pointee = String.Encoding.utf16LittleEndian.rawValue
         }
+        else if readResult.length >= 4 && bytePtr[0] == 0 && bytePtr[1] == 0 && bytePtr[2] == 254 && bytePtr[3] == 255 {
+          enc?.pointee = String.Encoding.utf32BigEndian.rawValue
+        }
```
pushkarnk: We seem to be some `Int` literals. Will it make sense to replace them with constants so that their contextual meaning is clear?

build.py
mohitathwani: They're actually there. Because these files have BOM in them, Github is detecting them to be binary files. If you go ahead and click on "Files Changed" above, you would see them listed as binary files..

Foundation/NSString.swift
mohitathwani: What kind of naming conventions do you suggest for these literals?

Foundation/NSString.swift
mohitathwani: Updated my code with hex values

TestFoundation/TestNSString.swift
parkera: @mohitathwani can you double check on this part?

mohitathwani: @parkera @phausler I just tested this on MacOS... The tests are passing...

alblue:
@swift-ci please test

alblue:
This looks good to me - @phausler WDYT?

phausler:
looks good to me

PR #929, opened 2017-04-03
Address slice regions that incorrectly fetched from the wrong offset of the backing buffers

phausler:
Addresses the following issues:
https://bugs.swift.org/browse/SR-4462
https://bugs.swift.org/browse/SR-4473
rdar://problem/31361522

This also brings Data.swift on par with the implementation in the Foundation overlay. Making the slice of Data to be Data.


phausler:
@swift-ci test

PR #930, opened 2017-04-03
Correct Data slice iteration to be indexed by the startIndex to the endIndex

phausler:


phausler:
@swift-ci test and merge

PR #932, opened 2017-04-06
Update HTTPCookiePropertyKey.path comment

ddunn2:
This PR is relating to [SR-3821](https://bugs.swift.org/browse/SR-3821)

The comment seems to suggest that supplying a value for `path` is optional and if a value isn't provided it'll either infer from `HTTPCookiePropertyKey.originURL` or use a default of "/". None of those seem to be the case though, and `HTTPCookie([...])` will return `nil` if you haven't provided a value for `path`. 

Tracked down some [documentation](https://developer.apple.com/reference/foundation/httpcookie/1392975-init) to support this as well.

### DIFF
--- a/Foundation/NSHTTPCookie.swift
+++ b/Foundation/NSHTTPCookie.swift
@@ -184,10 +184,8 @@ open class HTTPCookie : NSObject {
     /// <tr>
     ///     <td>HTTPCookiePropertyKey.path</td>
     ///     <td>NSString</td>
-    ///     <td>NO</td>
-    ///     <td>Path for the cookie. Inferred from the value for
-    ///     HTTPCookiePropertyKey.originURL if not provided. Default is "/".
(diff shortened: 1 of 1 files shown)


pushkarnk:
The code comments that are the subject here don't seem to be consumed in any online documentation , neither in [HTTPCookie.init(properties:)](https://developer.apple.com/reference/foundation/httpcookie/1392975-init) nor [HTTPCookiePropertyKey](https://developer.apple.com/reference/foundation/httpcookiepropertykey). I wonder if we need those HTML tags anymore. May be we should do away with them too?

pushkarnk:
I'll merge this since it addresses the SR. But we should revisit the code comments in question to check if we could do more of a cleanup. Thanks!

@swift-ci test and merge

pushkarnk:
@swift-ci test and merge

PR #936, opened 2017-04-06
Fix bit shift math in CFCharacterSet equality code

mark1smi:
The code was originally shifting a 32-bit value by 32 bits in
some cases, which is undefined behavior. The intent was that
all bits would be shifted out, leaving a value of 0 to perform
computation with. We now treat the value as 64-bits for the shift
and re-cast to 32-bits for the computation.

### DIFF
--- a/CoreFoundation/String.subproj/CFCharacterSet.c
+++ b/CoreFoundation/String.subproj/CFCharacterSet.c
@@ -216,10 +216,10 @@ static Boolean __CFCSetIsBitmapEqualToRange(const UInt32 *bits, UniChar firstCha
         UInt32 lastCharMask;
 
         length = firstCharIndex % sizeof(UInt32);
-        firstCharMask = (((((UInt32)0xFF) << (firstChar & (BITSPERBYTE - 1))) & 0xFF) << (((sizeof(UInt32) - 1) - length) * BITSPERBYTE)) | (((UInt32)0xFFFFFFFF) >> ((length + 1) * BITSPERBYTE));
+        firstCharMask = (((((UInt32)0xFF) << (firstChar & (BITSPERBYTE - 1))) & 0xFF) << (((sizeof(UInt32) - 1) - length) * BITSPERBYTE)) | (UInt32)(((UInt64)0xFFFFFFFF) >> ((length + 1) * BITSPERBYTE));
         
         length = lastCharIndex % sizeof(UInt32);
-        lastCharMask = ((((UInt32)0xFF) >> ((BITSPERBYTE - 1) - (lastChar & (BITSPERBYTE - 1)))) << (((sizeof(UInt32) - 1) - length) * BITSPERBYTE)) | (((UInt32)0xFFFFFFFF) << ((sizeof(UInt32) - length) * BITSPERBYTE));
+        lastCharMask = ((((UInt32)0xFF) >> ((BITSPERBYTE - 1) - (lastChar & (BITSPERBYTE - 1)))) << (((sizeof(UInt32) - 1) - length) * BITSPERBYTE)) | (UInt32)(((UInt64)0xFFFFFFFF) << ((sizeof(UInt32) - length) * BITSPERBYTE));
 
         firstCharIndex = firstChar >> LOG_BPLW;
         lastCharIndex = lastChar >> LOG_BPLW;


mark1smi:
The code below exercises the code paths in fast path comparison `__CFCSetIsBitmapEqualToRange` between character sets with an internal representation of `__kCFCharSetClassRange` and `__kCFCharSetClassBitmap`. This code fails prior to the fix.

Interestingly, this bug also symptomizes in shipping OS X and iOS.

```
    CFCharacterSetRef rangeSet, bitmapSet;

    for (CFIndex startOffset = 0; startOffset < 32; startOffset++) {
        for (CFIndex endOffset = 0; endOffset < 32; endOffset++) {
            rangeSet = CFCharacterSetCreateWithCharactersInRange(NULL, CFRangeMake(1 + startOffset, 100 + endOffset));

            CFDataRef data = CFCharacterSetCreateBitmapRepresentation(NULL, rangeSet);
            bitmapSet = CFCharacterSetCreateWithBitmapRepresentation(NULL, data);
            CFRelease(data);

            XCTAssertTrue(CFEqual(rangeSet, bitmapSet));

            CFRelease(rangeSet);
            CFRelease(bitmapSet);
        }
    }

    for (CFIndex startOffset = 0; startOffset < 32; startOffset++) {
        for (CFIndex endOffset = 0; endOffset < 32; endOffset++) {
            rangeSet = CFCharacterSetCreateWithCharactersInRange(NULL, CFRangeMake(startOffset, 0 + endOffset));

            CFDataRef data = CFCharacterSetCreateBitmapRepresentation(NULL, rangeSet);
            bitmapSet = CFCharacterSetCreateWithBitmapRepresentation(NULL, data);
            CFRelease(data);

            XCTAssertTrue(CFEqual(rangeSet, bitmapSet));

            CFRelease(rangeSet);
            CFRelease(bitmapSet);
        }
    }
```

mark1smi:
cc: @modocache

pushkarnk:
@mark1smi Good catch. Will it be possible to include a unit test for this?

mark1smi:
@pushkarnk Thanks. I provided a unit test in the comment to exercise the C API, but didn't find anywhere in the project where C unit tests run. I spent a number of hours attempting to provide a test in NSCharacterSet.swift, but encountered two problems:

1) It was unclear to me how much the swift tests are used because they would not link (I submitted pull request https://github.com/apple/swift-corelibs-foundation/pull/937 to address that)
2) After fixing the link problem, it appears that something is fundamentally broken with NSCharacterSet in swift, such that you can't even compare two NSCharacterSets (infinite recursion with NSCharacterSet swift's isEqual sending its `_cfObject` down to `CFEqual`, and `CFEqual` sending back up via `CFTYPE_SWIFT_FUNCDISPATCH1(Boolean, cf1, NSObject.isEqual, (CFSwiftRef)cf2);` ). You'll notice that there are no enabled tests in TestNSCharacterSet.swift that compare two NSCharacterSets!

For reference, the swift code for the unit test would look like:

`
    func test_RangeCompareWithBitmap() {
        for startOffset: Int in 0..<32 {
            for endOffset: Int in 0..<32 {
                //let rangeSet = CharacterSet(Range:(1 + startOffset, 100 + endOffset))
                let rangeSet = CharacterSet(charactersIn: Range<UnicodeScalar> (uncheckedBounds:(UnicodeScalar(1 + startOffset)!, UnicodeScalar(100 + endOffset)!)));
                let bitmap = rangeSet.bitmapRepresentation
                let bitmapSet = CharacterSet(bitmapRepresentation: bitmap)
                XCTAssertEqual(rangeSet, bitmapSet)
            }
        }

        for startOffset: Int in 0..<32 {
            for endOffset: Int in 0..<32 {
                let rangeSet = CharacterSet(charactersIn: Range<UnicodeScalar> (uncheckedBounds:(UnicodeScalar(startOffset)!, UnicodeScalar(endOffset)!)));
                let bitmap = rangeSet.bitmapRepresentation
                let bitmapSet = CharacterSet(bitmapRepresentation: bitmap)
                XCTAssertEqual(rangeSet, bitmapSet)
            }
        }
    }
`
I could add that as a disabled test to the diff if you would like, but it felt to me a strange way of testing the C API.

parkera:
We know that NSCharacterSet is basically un-subclassable, which is why we switched the implementation over to has-a CFCharacterSet instead. That is likely the issue you ran into with it in Swift.

parkera:
I'll run our internal CF tests against your change to verify this.

parkera:
@swift-ci test and merge

PR #938, opened 2017-04-07
Remove blocks runtime when building with libdispatch

spevans:
- The blocks runtime is already contained in libdispatch
  so removing this 2nd copy from libFoundation removes duplicate
  code.

- The libdispatch version is also slightly more optimal as it
  doesnt compile functions only needed when HAVE_OBJC is defined.

This is also needed to avoid duplicate symbol errors when statically linking Foundation using https://github.com/apple/swift-corelibs-foundation/pull/873

Also see discussion about blocks runtime in https://github.com/apple/swift-corelibs-foundation/pull/874

An alternative, which I did test, is just to remove `data.c` and `runtime.c` although this means `libdispatch` always has to be linked in and Im not sure if it is guaranteed to always be a dependancy.

### DIFF
--- a/build.py
+++ b/build.py
@@ -210,8 +210,6 @@ project = [
 foundation.add_phase(headers)
 
 sources = CompileSources([
-    'closure/data.c',
-    'closure/runtime.c',
     'uuid/uuid.c',
 	# 'CoreFoundation/AppServices.subproj/CFUserNotification.c',
 	'CoreFoundation/Base.subproj/CFBase.c',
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci please test

spevans:
@parkera Can this be merged now?

parkera:
Sure; let me kick off another test since it's been a while.

parkera:
@swift-ci test and merge

PR #940, opened 2017-04-08
Fix building on RasPi

ffried:
Ubuntu on Raspberry Pi reports "arm-linux-gnueabihf". Since it still needs to link curl, I added the check for this triple ending.

### DIFF
--- a/build.py
+++ b/build.py
@@ -81,7 +81,7 @@ if "XCTEST_BUILD_DIR" in Configuration.current.variables:
 	]
 
 triple = Configuration.current.target.triple
-if triple.endswith("-linux-gnu") or triple == "armv7-none-linux-androideabi":
+if triple.find("linux") != -1:
 	foundation.LDFLAGS += '-lcurl '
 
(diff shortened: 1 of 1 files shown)


uraimo:
Hi, it could make sense to simply check for `if triple.endswith("linux") != -1:` to catch every Linux variant, since looking at the history it looks like that that line was added to support all Linux variants. The only thing I can't be 100% sure about is if that works for the IBM LinuxOne too, but I guess it should.

pushkarnk:
I agree with @uraimo . We should probably check if we can reduce the checks to a single check for "linux". If that isn't feasible, the current solution looks good to me. Thanks.

ffried:
Okay. I've changed the check to `if triple.find("linux") != -1:`. I've tested it with some common triples and it worked.

pushkarnk:
This looks good for now. Thanks.

@swift-ci test and merge

pushkarnk:
@swift-ci test and merge

PR #942, opened 2017-04-09
Refactoring in IndexPath

adamnemecek:


parkera:
Looks good; can we double check that this file matches what's in the swift overlay?

adamnemecek:
Is that something I should do?

parkera:
Yes please. I know it's annoying that we have two copies of the source in two repositories, but it's the best option we have at the moment.

adamnemecek:
I'm ok with doing it but I'm not sure what exactly I'm supposed to do. Or like I get what you are saying but idk how to accomplish that.

ianpartridge:
I think @parkera is asking for a similar PR to be opened, updating the Darwin overlay for IndexPath. You can find that here: https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/IndexPath.swift

alblue:
This pull request has gone stale. If you revisit it, please rebase against master and check the changes against the SDK/Foundation/IndexPath.swift as well.

PR #943, opened 2017-04-10
Swift 3.1 branch data fixes

phausler:
this cherry picks the Data fixes for indexing and slicing:

Address slice regions that incorrectly fetched from the wrong offset of the backing buffers (#929)

* Update Data to be in sync with Foundation overlay

* Account for slice starts in appending, replacing subranges and fetching unsafe byte pointers

Correct Data slice iteration to be indexed by the startIndex to the endIndex


phausler:
@swift-ci test

PR #944, opened 2017-04-11
Fix for SR-3634: Error while accessing Locale.availableIdentifiers

mamabusi:


Foundation/NSLocale.swift
```diff
@@ -97,7 +97,7 @@ extension NSLocale {
     open class var availableLocaleIdentifiers: [String] {
         var identifiers = Array<String>()
         for obj in CFLocaleCopyAvailableLocaleIdentifiers()._nsObject {
-            identifiers.append((obj as! NSString)._swiftObject)
+            identifiers.append(obj as! String)
```
mamabusi: At runtime type of 'obj' is received as String and the caste from Swift.String to Foundation.NSString fails.

parkera:
@swift-ci test and merge

PR #946, opened 2017-04-12
Fix a few small bugs in CFLocale

mark1smi:
* Adds missing mutex unlock in an error path
* Fixes use of an incorrect local variable
* Addresses a few areas where _prefs member was assumed to be non-NULL
  * Fix to __CFLocaleGetMeasurementSystemGuts and __CFLocaleCopyTemperatureUnit
    modelled after __CFLocaleCopyCollatorID
  * Fix to __CFLocaleEqual handles the case where one or both CFLocales
    have NULL _prefs.


mark1smi:
@parkera Please let me know if I should stop mentioning you on these CF patches, but it seems like you're the right person for these.

alblue:
@swift-ci please test

parkera:
@mark1smi That's fine, thanks.

PR #951, opened 2017-04-15
Upgrading to the new integer protocols

moiseev:


Foundation/CGFloat.swift
```diff
@@ -279,33 +292,28 @@ extension CGFloat : BinaryFloatingPoint {
...
-
-    @_transparent
     public mutating func negate() {
         native.negate()
     }
 
     @_transparent
-    public mutating func add(_ other: CGFloat) {
-        native.add(other.native)
+    public static func +=(_ lhs: inout CGFloat, _ rhs: CGFloat) {
```
parkera: Is this a source-compatible change?

moiseev: Technically not, but these methods were there only to implement operators on the protocol, which is no longer required. In any case these missing methods can now be added for source compatibility in the protocol. Thanks for looking at this!

moiseev: Added the missing methods here:  https://github.com/apple/swift/pull/3796/commits/192c11217fe0c1a1821004b34434ffe523875707

moiseev:
@parkera I've just merged the new integers into master. Is it OK for me to merge this one?

slavapestov:
This is blocking PR testing on Linux so I'm going to merge it.

PR #952, opened 2017-04-15
NSData: free() buffer before throw to prevent memory leak.

spevans:


parkera:
@swift-ci test and merge

PR #960, opened 2017-04-21
Add NSISO8601DateFormatter stub

ddunn2:
This is a class that was recently added to Darwin and would be great if it was added into swift-corelibs-foundation as well to keep Darwin and Linux more in sync.


parkera:
Thanks!

parkera:
@swift-ci test and merge

PR #965, opened 2017-04-25
Implementation of URLProtocol and refactoring URLSession

nethraravindran:
Currently URLSessionTask is tightly coupled with HTTP/HTTPS specific functions. This makes it difficult to introduce support for other native and custom protocols. Refactoring URLSession helps in adding support for other native protocols like ftp and data as well. Also, URLProtocol implementation provides support for custom protocols.
 
This document provides the proposal for restructuring URLSession
https://gist.github.com/pushkarnk/1bae3167d2e62b09ba0dd4b0dd285179

### DIFF
--- a/CoreFoundation/Base.subproj/CFFileUtilities.c
+++ b/CoreFoundation/Base.subproj/CFFileUtilities.c
@@ -15,6 +15,7 @@
 #include <errno.h>
 #include <string.h>
 #include <stdio.h>
+#include <assert.h>
 
 #if DEPLOYMENT_TARGET_WINDOWS
 #include <io.h>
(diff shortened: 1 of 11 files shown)


pushkarnk:
Can we move `libcurlHelpers.swift` to the `http` folder?

Foundation/NSURLProtocol.swift
```diff
@@ -323,19 +433,54 @@ open class URLProtocol : NSObject {
...
+            _lock.lock()
+            _registeredProtocolClasses.append(protocolClass)
+            _lock.unlock()
+            return true
+        }
+        return false
+    }
+
+    // TODO: Synchronize all the class funcs that use the registered protocol classes array
+    internal class func getProtocolClass(for urlRequest: URLRequest) -> AnyClass? {
```
pushkarnk: This also needs to use the lock

PR #967, opened 2017-04-26
Fix `(nsError as Error).localizedDescription` output

ikesyo:
Because `NSError` is not `LocalizedError` and `_swift_Foundation_getErrorDefaultUserInfo` does not handle `NSError` instances, `(nsError as Error).localizedDescription` didn't return expected output. So let's handle `NSError` case in `Error.localizedDescription` implementation.


ikesyo:
@swift-ci Please test

parkera:
Thanks!

PR #971, opened 2017-04-28
Drop NS prefix from some files to match the class names

ikesyo:

PR #973, opened 2017-05-01
[3.1] Drop NS prefix from some files to match the class names

ikesyo:
Cherry-picks #971 into `swift-3.1-branch`.


ikesyo:
@swift-ci Please test

PR #974, opened 2017-05-02
Update for AnyObject nukage

slavapestov:
AnyObject is now a special type and no longer a protocol, which changes the mangling as soon as the associated Swift PR is merged. There is no functional change here.


parkera:
Looks good, merge when needed.

PR #976, opened 2017-05-03
Revert "Update for AnyObject nukage"

slavapestov:
Swift changed the mangling for AnyObject in master, but not yet in swift-4.0-branch.

Foundation recently did not have a swift-4.0-branch. Now that it does, revert the change so that Foundation matches Swift across the two swift-4.0-branch branches.


slavapestov:
@parkera Sorry for the breakage!

PR #982, opened 2017-05-12
Working around intermittent URLSession Test Failures on async wait.

mamabusi:


mamabusi:
An intermittent test-failure is observed in TestURLSession.test_downloadTaskWithURLAndHandler where the test times out. JIRA bug: https://bugs.swift.org/browse/SR-4647

In URLSessionTask, the api ‘startNewTransfer’ https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURLSession/NSURLSessionTask.swift#L456 calls the setter for ‘currentRequest’ https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURLSession/NSURLSessionTask.swift#L142 which posts an async task of barrier type on the concurrent queue ‘taskAttributesIsolation’.

Further in the same api ‘startNewTransfer’, ‘configureEasyHandle’ api is called which calls the getter on ‘currentRequest’ https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURLSession/NSURLSessionTask.swift#L570 which posts a sync task https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURLSession/NSURLSessionTask.swift#L138 on the same concurrent queue ‘taskAttributesIsolation’.

Reading the dispatch_barrier_async api reference https://developer.apple.com/reference/dispatch/1452797-dispatch_barrier_async it says: “Calls to this function always return immediately after the block has been submitted and never wait for the block to be invoked. When the barrier block reaches the front of a private concurrent queue, it is not executed immediately. Instead, the queue waits until its currently executing blocks finish executing. At that point, the barrier block executes by itself. Any blocks submitted after the barrier block are not executed until the barrier block completes.”

In case of the test-failure, when the async task is posted no other tasks exist: Please refer to the stack-trace in the bug link: https://bugs.swift.org/browse/SR-4647

Instrumenting Foundation code, it is seen that the test is hung on the sync task and then fails with timeout.
Looks like the async barrier request task never completed on the dispatch queue which blocked the sync request task causing the timeout.

This fix uses the simpler dispatch semaphore to carry out the short critical sections in NSURLSessionTask to overcome the intermittent failure seen.

pushkarnk:
@swift-ci test

pushkarnk:
This work around looks good to me. Though parallel reads of the property will no longer be possible, it may not matter much because, as noted in the comment above, the critical sections are small - just one read/write. 

Another point to note here is that though this particular failure does show up locally, quite consistently, on a 2CPU box, we do not know if this was the actual (and only) reason for the CI failures. Nevertheless, I think we should enable the URLSession tests with this and monitor the CI failures for a few weeks. Thanks.

pushkarnk:
I am merging this one now. Since it is enabling the URLSession tests we will have to monitor any CI failures over the next few weeks. Thanks.

pushkarnk:
@swift-ci test and merge

PR #983, opened 2017-05-15
URLSession's delegateQueue should be serial as per the spec.

mamabusi:
Refer to the API document that mentions URLSession's delegateQueue should be serial:
https://developer.apple.com/reference/foundation/urlsession/1411597-init

### DIFF
--- a/Foundation/NSURLSession/NSURLSession.swift
+++ b/Foundation/NSURLSession/NSURLSession.swift
@@ -211,6 +211,7 @@ open class URLSession : NSObject {
         self.workQueue = DispatchQueue(label: "URLSession<\(identifier)>")
         self.taskAttributesIsolation = DispatchQueue(label: "URLSession<\(identifier)>.taskAttributes", attributes: DispatchQueue.Attributes.concurrent)
         self.delegateQueue = OperationQueue()
+        self.delegateQueue.maxConcurrentOperationCount = 1
         self.delegate = nil
         //TODO: Make sure this one can't be written to?
         // Could create a subclass of URLSessionConfiguration that wraps the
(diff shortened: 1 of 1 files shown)


Foundation/NSURLSession/NSURLSession.swift
```diff
@@ -226,6 +227,7 @@ open class URLSession : NSObject {
         self.workQueue = DispatchQueue(label: "URLSession<\(identifier)>")
         self.taskAttributesIsolation = DispatchQueue(label: "URLSession<\(identifier)>.taskAttributes", attributes: DispatchQueue.Attributes.concurrent)
         self.delegateQueue = queue ?? OperationQueue()
+        self.delegateQueue.maxConcurrentOperationCount = 1
```
mamabusi: The spec. says "The queue should be a serial queue, in order to ensure the correct ordering of callbacks". But yes  it is not mandatory and user can send in a concurrent queue as well. Tested with a small test-case on Darwin and I see that user's OperationQueue is not converted to serial if it isn't already. Thanks @ianpartridge for pointing this out. 
Modified the changes for the same.

pushkarnk:
@swift-ci test

pushkarnk:
@mamabusi I'd suggest we have a comment making it clear that we will NOT make a user supplied OperationQueue serial.

Foundation/NSURLSession/NSURLSession.swift
```diff
@@ -225,7 +226,9 @@ open class URLSession : NSObject {
         identifier = nextSessionIdentifier()
         self.workQueue = DispatchQueue(label: "URLSession<\(identifier)>")
         self.taskAttributesIsolation = DispatchQueue(label: "URLSession<\(identifier)>.taskAttributes", attributes: DispatchQueue.Attributes.concurrent)
-        self.delegateQueue = queue ?? OperationQueue()
+        let operationQueue = OperationQueue()
```
ianpartridge: We shouldn't create an `OperationQueue` unless we need to. This creates one in all cases.  How about this:

```swift
if queue != nil {
    self.delegateQueue = queue
} else {
    self.delegateQueue = OperationQueue()
    self.delegateQueue.maxConcurrentOperationCount = 1
}
```

mamabusi: Agreed!

mamabusi:
@pushkarnk @ianpartridge Review changes and suggestions are addressed. Thanks!

ianpartridge:
Thanks @mamabusi. @pushkarnk - can you re-run the CI?

pushkarnk:
Looks good to me. Thanks!

pushkarnk:
Just for the records, here's what @ianpartridge found:  https://www.malhal.com/2017/01/07/nsoperationqueue-maxconcurrentoperationcount-1-serial-queue-problem/  (though it is unlikely that a user will supply a delegate method that will block indefinitely)

pushkarnk:
@swift-ci test

pushkarnk:
@swift-ci test

pushkarnk:
@swift-ci test and merge

pushkarnk:
@swift-ci test and merge

PR #984, opened 2017-05-17
Small cleanup: Use () -> ... instead of (Void) -> ... for no-argument closures

slavapestov:
In Swift, (Void) -> Void is actually a function that takes
a single value of empty-tuple type, and not a function that
takes no arguments.

Swift 3 largely ignored the distinction because of the implicit
'tuple splat' behavior; in Swift 4, this behavior has been
eliminated as part of implementing SE-0110, so calling a function
of type (Void) -> () now requires passing in an empty tuple,
like foo(()).

I believe this is not intended behavior, so this patch changes
the functions in question to be written as () -> Void.

### DIFF
--- a/Foundation/Thread.swift
+++ b/Foundation/Thread.swift
@@ -24,7 +24,7 @@ private func _compiler_crash_fix(_ key: _CFThreadSpecificKey, _ value: AnyObject
 internal class NSThreadSpecific<T: NSObject> {
     private var key = _CFThreadSpecificKeyCreate()
     
-    internal func get(_ generator: (Void) -> T) -> T {
+    internal func get(_ generator: () -> T) -> T {
         if let specific = _CFThreadSpecificGet(key) {
             return specific as! T
         } else {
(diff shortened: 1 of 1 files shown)


slavapestov:
@parkera Does this look OK?

parkera:
Yup, thanks.

mattrajca:
@slavapestov how are you able to build `swift-corelibs-foundation` with Swift 4/master? I get hundreds of build errors due to renamed APIs and bit shifts not being available apparently.

e78l:
@mattrajca Could you confirm the platform you're building for? Ubuntu 16.10? I haven't seen NSGeometry errors you described in the email. Are you using the swift/util/build script - or maybe you're building against a snapshot somehow?

mattrajca:
@e78l I am using `utils/build-script` with `-r --foundation --xctest -t` and using Xcode 8.3.2 on macOS 10.12.4. I also tried using the standard toolchain as well as the latest Swift 4 snapshot.

pushkarnk:
@mattrajca #988 is expected to fix the build errors on macOS ([this thread](https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20170515/001197.html))

pushkarnk:
I think this is can be merged. It will eliminate a warning that we're seeing in the Foundation builds.

pushkarnk:

PR #956, opened 2017-04-20
Force use of Swift 3 compatibility mode to compile sources.

tkremenek:
This is a prerequisite for: https://github.com/apple/swift/pull/8861

PR #988, opened 2017-05-18
'Fix' macOS build errors, TestXDG folder cleanup

e78l:
- 'Fix' macOS build errors with swift compiler option `-swift-version 3`. Xcode seems to have a SWIFT_VERSION setting, but it doesn't have the same effect! Changes should be like those in https://github.com/apple/swift-corelibs-foundation/pull/956.
- Quiet some warnings with `-Wno-nullability-completeness-on-arrays` and `-Wno-format-security` for CoreFoundation - please review if `GCC_WARN_TYPECHECK_CALLS_TO_PRINTF` Xcode setting has the same effect/is better than `-Wno-format-security`
- Has mangling changed on Mac? I've have had issues with comparing CF/NS strings, and not sure if it /was just/ a mangling issue, but got warning and error with macOS build, so updating __CFSwiftGetBaseClass and __CFConstantStringClassReference.
@slavapestov  or @eeckstein - is there a way for a script to determine a mangled class/function name at build time?
The warning/error reported on macOS:
~~~~
ld: warning: undefined base symbol '__TMC15SwiftFoundation19_NSCFConstantString' for alias '___CFConstantStringClassReference'
Undefined symbols for architecture x86_64:
  "__T015SwiftFoundation21__CFSwiftGetBaseClassyXlXpyF", referenced from:
      ___CFInitialize in libCoreFoundation.a(CFRuntime.o)
ld: symbol(s) not found for architecture x86_64
~~~~
- Refactored XDG helper, and created target in Xcode so tests pass on Mac (well 2 failures still in NSString):
~~~~
TestNSString.test_FromContentsOfURLUsedEncodingUTF16BE : failed - URL for NSString-UTF16-BE-data.txt is nil
TestNSString.test_FromContentsOfURLUsedEncodingUTF16LE : failed - URL for NSString-UTF16-LE-data.txt is nil
~~~~
Target depends on SwiftFoundation being at same level in build directory (added rpath in lieu of framework copy phase). Perhaps copy can be disabled for TestFoundation too?
@mamabusi Are these changes okay with you?
I am thinking, perhaps, xdgTestHelper can be renamed `testHelper` or similar and used for any other 'checks' needing another process in the future.


mattrajca:
This fixed my recent macOS build failures. Thanks.

mamabusi:
@e78l XDG Test executable changes look good to me. I second your opinion of renaming xdgTestHelper to a more generic name if it helps other tests needing another process.

parkera:
I would really like to split up these fixes into a few different PRs; especially stuff about build settings vs warning fixes in CF.

Foundation.xcodeproj/project.pbxproj
```diff
@@ -2707,6 +2781,48 @@
...
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
+				CLANG_ANALYZER_NONNULL = YES;
+				CLANG_ANALYZER_NUMBER_OBJECT_CONVERSION = YES_AGGRESSIVE;
+				CLANG_ENABLE_MODULES = YES;
+				CLANG_ENABLE_OBJC_ARC = YES;
+				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
+				CODE_SIGN_IDENTITY = "";
+				COMBINE_HIDPI_IMAGES = YES;
+				INFOPLIST_FILE = TestFoundation/xdgTestHelper/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../../.. @loader_path/../../.. @executable_path/../Frameworks";
```
parkera: This path is wacky. Is that really the right value?

Foundation.xcodeproj/project.pbxproj
```diff
@@ -2652,6 +2722,8 @@
 					"-DCF_CHARACTERSET_UNICODE_DATA_B=\\\\\"CoreFoundation/CharacterSets/CFUnicodeData-B.mapping\\\\\"",
 					"-DCF_CHARACTERSET_UNICHAR_DB=\\\\\"CoreFoundation/CharacterSets/CFUniCharPropertyDatabase.data\\\\\"",
 					"-DCF_CHARACTERSET_BITMAP=\\\\\"CoreFoundation/CharacterSets/CFCharacterSetBitmaps.bitmap\\\\\"",
+					"-Wno-nullability-completeness-on-arrays",
```
parkera: Why did we need to turn these off, instead of maybe fixing the warnings?

TestFoundation/TestNSHTTPCookieStorage.swift
```diff
@@ -241,9 +241,14 @@ class TestNSHTTPCookieStorage: XCTestCase {
         let bundlePath = bundle.bundlePath
         let pathIndex = bundlePath.range(of: "/", options: .backwards)?.lowerBound
         let task = Process()
-        task.launchPath = bundlePath.substring(to: pathIndex!) + "/xdgTestHelper/xdgTestHelper"
+        #if os(macOS)
```
parkera: This should really just be using the Bundle API directly to find the correct path in the bundle, which would be platform independent.

Foundation.xcodeproj/project.pbxproj
e78l: Auditing CF for these warnings may be worthwhile, but I'm thinking it'll take quite a bit of time to fix+review. Meanwhile, there is a lot of noise; in particular, `no-format-security` flags CFSTRs which look fine.

Foundation.xcodeproj/project.pbxproj
e78l: I'm adding the SwiftFoundation in the build directory, so for the executable at:
Build/Products/Debug/xdgTestHelper.app/Contents/MacOS/xdgTestHelper 
look for framework at:
Build/Products/Debug/SwiftFoundation.framework 
Could also just copy the framework into `@executable_path/../Frameworks` but that seems unnecessary to me

TestFoundation/TestNSHTTPCookieStorage.swift
e78l: Removed for now (tests on mac will crash)
`xdgTestHelper` is not in TestFoundation, so I would create Bundle with URL at path `bundlePath.substring(to: pathIndex!)` and then get executable path?

e78l:
Hello @parkera,
Thanks for the review - I split some changes that were easy to do into #1006 and can split further+undo warnings silence if decided (would be a bit time consuming).
Removed Swift 4 changes which I think were covered by @spevans. After those changes, may be able to forget about `-swift-version 3`; in that case, may also want to update build.py.
@mamabusi Thanks also!

Foundation.xcodeproj/project.pbxproj
parkera: This xdgTestHelper was only being built on Linux before, but this ports it as a .app to macOS right?

e78l: Yes

parkera:
@swift-ci test

parkera:
For some reason status isn't being updated here after request (cc @shahmishal @erg) but the PR is testing here: https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/872/

parkera:
Looks like it passed. Thanks!

PR #990, opened 2017-05-18
[NSArray] Add check to `isEqual(to:)` whether values is `_ObjectBridgeable`

norio-nomura:
When `NSArray` are compared for equality, `NSArray` only checks if the element can be casted as `AnyHashable`. Therefore, the result was incorrect when the `NSArray` contained instances which are not `Hashable` but are subclasses of NSObject (eg `NSArray` and `NSDictionary`) .
This PR fixes that.

Fixes [SR-4903](https://bugs.swift.org/browse/SR-4903)

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -312,6 +312,13 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
                 if val1 != val2 {
                     return false
                 }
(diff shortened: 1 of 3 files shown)


norio-nomura:
Updated description.

alblue:
@swift-ci please test

aciidb0mb3r:
@parkera is this good to merge?

norio-nomura:
Thanks! 🙏

PR #992, opened 2017-05-19
Add tests for Scanner.atEnd.

mattrajca:
Finish up [SR-4398](https://bugs.swift.org/browse/SR-4398) by adding tests for Scanner's atEnd property.

### DIFF
--- a/TestFoundation/TestNSScanner.swift
+++ b/TestFoundation/TestNSScanner.swift
@@ -26,6 +26,7 @@ class TestNSScanner : XCTestCase {
             ("test_scanInteger", test_scanInteger),
             ("test_scanFloat", test_scanFloat),
             ("test_scanString", test_scanString),
+            ("test_charactersToBeSkipped", test_charactersToBeSkipped),
         ]
     }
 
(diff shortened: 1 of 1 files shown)


TestFoundation/TestNSScanner.swift
```diff
@@ -41,4 +43,19 @@ class TestNSScanner : XCTestCase {
...
+        guard let firstPart = scanner.scanString(string: "apple ") else {
+            XCTFail()
+            return
+        }
+
+        XCTAssertEqual(firstPart, "apple ")
+        XCTAssertFalse(scanner.atEnd)
+
+        let _ = scanner.scanString(string: "sauce")
+        XCTAssertTrue(scanner.atEnd)
```
kballard: `isAtEnd` is supposed to return `true` if the scanner hasn't actually consumed all the input, but the remaining input can all be skipped (due to `charactersToBeSkipped`). It would be great to actually test that behavior.

mattrajca: Added in a follow-up commit. Thanks for the note.

alblue:
@swift-ci please test

parkera:
Thanks!

PR #995, opened 2017-05-21
swift4: Fixes for closure arguments

spevans:
- Fix destructuring

- Use correct number of arguments in closures

Compatible with swift 3 & 4

### DIFF
--- a/Foundation/Dictionary.swift
+++ b/Foundation/Dictionary.swift
@@ -18,7 +18,8 @@ extension Dictionary : _ObjectTypeBridgeable {
         
         var idx = 0
         
-        self.forEach { (keyItem, valueItem) in
+        self.forEach { (arg) in
+            let (keyItem, valueItem) = arg
             let key = _SwiftValue.store(keyItem)
             let value = _SwiftValue.store(valueItem)
             keyBuffer.advanced(by: idx).initialize(to: key)
@@ -32,7 +33,7 @@ extension Dictionary : _ObjectTypeBridgeable {
         valueBuffer.deinitialize(count: count)
         keyBuffer.deallocate(capacity: count)
         valueBuffer.deallocate(capacity: count)
-        
+
         return dict
 
     }
--- a/Foundation/Dictionary.swift
+++ b/Foundation/Dictionary.swift
@@ -18,10 +18,9 @@ extension Dictionary : _ObjectTypeBridgeable {
         
         var idx = 0
         
-        self.forEach { (arg) in
-            let (keyItem, valueItem) = arg
-            let key = _SwiftValue.store(keyItem)
-            let value = _SwiftValue.store(valueItem)
+        self.forEach { entry in
+            let key = _SwiftValue.store(entry.key)
+            let value = _SwiftValue.store(entry.value)
             keyBuffer.advanced(by: idx).initialize(to: key)
             valueBuffer.advanced(by: idx).initialize(to: value)
             idx += 1
--- a/Foundation/NSIndexSet.swift
+++ b/Foundation/NSIndexSet.swift
@@ -92,7 +92,7 @@ open class NSIndexSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     
     open func mutableCopy(with zone: NSZone? = nil) -> Any {
         let set = NSMutableIndexSet()
-        enumerateRanges(options: []) { (range, stop) in
+        enumerateRanges(options: []) { (range, _) in
             set.add(in: range)
         }
         return set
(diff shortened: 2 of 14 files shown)


pushkarnk:
@swift-ci test

parkera:
I'm not sure I understand what was wrong with the previous syntax. Why did this become a problem?

spevans:
There were 2 sets of errors:

1. The closure argument count needs to match the signature, eg ` Contextual closure type '(UnsafeMutableRawPointer, Int) -> Void' expects 2 arguments, but 1 was used in closure body` and `Contextual type for closure argument list expects 1 argument, which cannot be implicitly ignored`

2. Some cases for { (key, value)  needed to be replaced with { (arg) ; let (key, value) = arg to solve the following: 
 `Closure tuple parameter '(Notification, [RunLoopMode])' does not support destructuring`

Most of the fixes were from fix-its in Xcode

damuellen:
Just yesterday, the error message "..does not support destructuring" occurred to me in some places in one of my projects. The Xcode fix-it has helped, but I would also like to know why Swift 4 is different in this case than Swift 3.

spevans:
Its due to SE-110 https://github.com/apple/swift-evolution/blob/master/proposals/0110-distingish-single-tuple-arg.md 
I think it made an incorrect argument count an  error in Swift4

parkera:
Is there an alternative to effectively re-declaring the argument all over the place? This can't be the pattern we really want people to use.

parkera:
I sent a message to the swift evolution mailing list about this. I think we should have a further discussion before we accept this; it seems like a dramatic reduction in usability of the language in general.

spevans:
Yeah thats a good idea. It does seem quite a source breaking change given how much closures are used in general

Foundation/Dictionary.swift
```diff
@@ -18,7 +18,8 @@ extension Dictionary : _ObjectTypeBridgeable {
         
         var idx = 0
         
-        self.forEach { (keyItem, valueItem) in
+        self.forEach { (arg) in
```
xwu: Stylistic point: this and all subsequent parameter lists in closures don't need the parentheses around `arg`.

Foundation/NSIndexSet.swift
```diff
@@ -92,8 +92,8 @@ open class NSIndexSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     
     open func mutableCopy(with zone: NSZone? = nil) -> Any {
         let set = NSMutableIndexSet()
-        enumerateRanges(options: []) {
-            set.add(in: $0.0)
+        enumerateRanges(options: []) { (range, stop) in
```
xwu: This could be `range, _`.

Foundation/NSXMLNode.swift
```diff
@@ -617,8 +617,8 @@ open class XMLNode: NSObject, NSCopying {
             if let parentNode = _CFXMLNodeGetParent(parent!) {
                 let grandparent = XMLNode._objectNodeForNode(parentNode)
                 let possibleParentNodes = grandparent.filter { $0.name == self.parent?.name }
-                let count = possibleParentNodes.reduce(0) {
-                    return $0.0 + 1
+                let count = possibleParentNodes.reduce(0) { (x, y) in
```
xwu: Here too, this could be `x, _`.

Foundation/Dictionary.swift
spevans: This change actually came from the migrator, it may be worth fixing that to not add parentheses as I was taking the fix-its as best practice. Although I'll see what the outcome of the discussion on swift-evolution is before making any changes to this PR

guidomb: Sorry if this is not the best place to ask this question but what about overloading functions like `forEach` or `filter` to do something like https://gist.github.com/guidomb/b313667616612fc3e0c74c02a066dd8b and avoid having to unpack tuples

igor-makarov:
It seems like most of those are derived from problems in they way functions in the Swift Standard Library are defined, for example `public func forEach(_ body: ((key: Key, value: Value)) throws -> Void) rethrows` expects a function that receives a tuple, whereas if it was written as `func forEach(_ body: (_ key: Key, _ value: Value) throws -> Void)`, it would expect a function that receives two params.

spevans:
Rebased. I added an extra commit with some tidy ups of the closure arguments although of course these are subject to personal taste more than anything.

@parkera Is this ok for merging now? I noticed most of the SE-110 discussion seems to have ended and any possible changes to the parser will be post swift4 now.

parkera:
I don't think that's actually the case, or at least I strongly hope it's not, so I still want to hold off on this.

spevans:
Closing for now, this is not needed as the SE-110 change is going to be backed out. See https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170619/037616.html

If any individual closures still generate warnings after the change a new PR can be opened.

PR #996, opened 2017-05-21
swift4: Fixes for integer initializers

spevans:
- Replace truncatingBitPattern: with extendingOrTruncating:
 - IntMax() -> Int(), UIntMax() -> UInt()

Compatible with swift 3 & 4


pushkarnk:
@swift-ci test

PR #997, opened 2017-05-21
swift4: Fixes for bit shifting

spevans:
Fixes bitshifts used with initializers  and converts some bitshifts into divides where the meaning makes more sense.

Compatible with swift 3 & 4

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -84,9 +84,9 @@ extension NSCalendar {
         public static let weekOfYear = Unit(rawValue: UInt(kCFCalendarUnitWeekOfYear))
         public static let yearForWeekOfYear = Unit(rawValue: UInt(kCFCalendarUnitYearForWeekOfYear))
 
-        public static let nanosecond = Unit(rawValue: UInt(1 << 15))
-        public static let calendar = Unit(rawValue: UInt(1 << 20))
-        public static let timeZone = Unit(rawValue: UInt(1 << 21))
+        public static let nanosecond = Unit(rawValue: UInt(1) << 15)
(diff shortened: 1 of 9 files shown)


pushkarnk:
@swift-ci test

pushkarnk:
@swift-ci test

Foundation/NSCalendar.swift
```diff
@@ -1309,7 +1309,7 @@ open class NSDateComponents : NSObject, NSCopying, NSSecureCoding {
         if NSDateComponentUndefined == yy {
             yy = 0
         }
-        return calHash + (32832013 * (y + yy) + 2678437 * m + 86413 * d + 3607 * h + 61 * mm + s) + (41 * weekOfYear + 11 * weekOfMonth + 7 * weekday + 3 * weekdayOrdinal + quarter) * (1 << 5)
+        return calHash + (32832013 * (y + yy) + 2678437 * m + 86413 * d + 3607 * h + 61 * mm + s) + (41 * weekOfYear + 11 * weekOfMonth + 7 * weekday + 3 * weekdayOrdinal + quarter) * 32
```
parkera: What's wrong now with writing 32 as 1 << 5?

spevans:
This PR is actually wrong, the underlying problems are caused by the `protocol _IntegerLike` conforming to `FixedWidthInteger` in
https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSScanner.swift#L223
since `<<` and `>>` were removed from `FixedWidthInteger`. So I think the correct fix is to understand what the code on lines 223-233 is doing and see if a better alternative can be found. It has a FIXME on it so its probably time to fix it!

PR #999, opened 2017-05-22
Ensure error is thrown when NSData is created from a non-existent file

mattrajca:
This patch adds a test for [SR-1547](https://bugs.swift.org/browse/SR-1547), which is now fixed on Linux, and ensures an error is thrown when an `NSData` instance is created from a non-existent file.


pushkarnk:
Thanks for the test!

@swift-ci test and merge

PR #1001, opened 2017-05-22
[SR-2679] Fix for URLSession delivers redirection delegate methods in wrong order

nethraravindran:


TestFoundation/TestNSURLSession.swift
```diff
@@ -608,3 +626,61 @@ class CustomProtocol : URLProtocol {
         return
     }
 }
+
+class HTTPRedirectionDataTask : NSObject {
```
pushkarnk: Can we inherit from DataTask here? That will possibly avoid some code duplication.

nethraravindran: Overriding delegate methods is not possible when inherited from `DataTask`

pushkarnk:
Thanks. The implementation looks good to me.

But it took me a while to understand the control flow related to the added test. Seems like the initial in-process HTTP server (`HTTPServer.swift`) is no longer good enough for the kind of tests that we are adding. There is some work happening on improving/refactoring the loopback server. Until that is available, this test is OK, but we'd need to improve upon it thereafter ;-)

pushkarnk:
@swift-ci please test

PR #1002, opened 2017-05-22
[Swift 3.1][NSArray] Add check to `isEqual(to:)` whether values is `_ObjectBridgeable`

norio-nomura:
When `NSArray` are compared for equality, `NSArray` only checks if the element can be casted as `AnyHashable`. Therefore, the result was incorrect when the `NSArray` contained instances which are not `Hashable` but are subclasses of NSObject (eg `NSArray` and `NSDictionary`) .
This PR fixes that.

Fixes [SR-4903](https://bugs.swift.org/browse/SR-4903)

Same with #990


parkera:
@swift-ci test and merge

norio-nomura:
Thanks! 🙏

PR #1003, opened 2017-05-22
Fix to evaluate the cookie storage path correctly

mamabusi:


Foundation/NSHTTPCookieStorage.swift
```diff
@@ -65,7 +65,7 @@ open class HTTPCookieStorage: NSObject {
         guard !FileManager.default.fileExists(atPath: path) else { return true }
 
         do {
-            try FileManager.default.createDirectory(atPath: path, withIntermediateDirectories: false, attributes: nil)
+            try FileManager.default.createDirectory(atPath: path, withIntermediateDirectories: true, attributes: nil)
```
mamabusi: Currently, if there was a non-existent parent directory in the 'path', this would return false and the cookie storage would happen in the current working directory. The expected behaviour is to create the parent directories as well.

Foundation/NSHTTPCookieStorage.swift
```diff
@@ -77,7 +77,7 @@ open class HTTPCookieStorage: NSObject {
             return path + fileName
         }
         //if we were unable to create the desired directory, create the cookie file in the `pwd`
-        return fileName
+        return FileManager.default.currentDirectoryPath + fileName
```
mamabusi: Before this fix, the path is considered to be the filename which looks like this: "/.cookies.shared". Instead of creating the cookie file in 'pwd', it tries to create the cookie file in the '/' (root) directory.

pushkarnk:
@swift-ci please test

pushkarnk:
Looks quite straightforward and. I'll merge this one.

PR #1004, opened 2017-05-22
swift4: Minor fixes

spevans:
- Fix 'Cannot call value of non-function type 'UnsafePointer<Int8>'
  by fully qualifying type(of:) in NSValue.swift

Compatible with swift 3 & 4

### DIFF
--- a/Foundation/NSURLResponse.swift
+++ b/Foundation/NSURLResponse.swift
@@ -396,8 +396,8 @@ private extension String {
                 type = string
             } else {
                 if let r = string.range(of: "=") {
-                    let name = string[string.startIndex..<r.lowerBound].trimmingCharacters(in: ws)
-                    let value = string[r.upperBound..<string.endIndex].trimmingCharacters(in: ws)
+                    let name = String(string[string.startIndex..<r.lowerBound]).trimmingCharacters(in: ws)
+                    let value = String(string[r.upperBound..<string.endIndex]).trimmingCharacters(in: ws)
                     parameters.append(ValueWithParameters.Parameter(attribute: name, value: value))
                 } else {
                     let name = string.trimmingCharacters(in: ws)
--- a/Foundation/NSValue.swift
+++ b/Foundation/NSValue.swift
@@ -98,7 +98,7 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
     
     public convenience required init(bytes value: UnsafeRawPointer, objCType type: UnsafePointer<Int8>) {
-        if type(of: self) == NSValue.self {
+        if Swift.type(of: self) == NSValue.self {
             self.init()
             if NSValue._isSpecialObjCType(type) {
                 self._concreteValue = NSSpecialValue(bytes: value.assumingMemoryBound(to: UInt8.self), objCType: type)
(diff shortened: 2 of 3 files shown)


aciidb0mb3r:
@swift-ci please test

Foundation/NSURLResponse.swift
```diff
@@ -396,8 +396,8 @@ private extension String {
                 type = string
             } else {
                 if let r = string.range(of: "=") {
-                    let name = string[string.startIndex..<r.lowerBound].trimmingCharacters(in: ws)
-                    let value = string[r.upperBound..<string.endIndex].trimmingCharacters(in: ws)
+                    let name = String(string[string.startIndex..<r.lowerBound]).trimmingCharacters(in: ws)
```
phausler: This is fairly unfortunate that we have to force a copy here, In the grand scheme of things it probably won't matter too much since the range is probably always small

Foundation/NSValue.swift
```diff
@@ -98,7 +98,7 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
     
     public convenience required init(bytes value: UnsafeRawPointer, objCType type: UnsafePointer<Int8>) {
-        if type(of: self) == NSValue.self {
+        if Swift.type(of: self) == NSValue.self {
```
phausler: imho this seems like a bug of type(of:) to require this! what is the causality here?

Foundation/Process.swift
```diff
@@ -241,7 +241,11 @@ open class Process: NSObject {
         if let env = environment {
             let nenv = env.count
             envp = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1 + nenv)
-            envp.initialize(from: env.map { strdup("\($0)=\($1)") })
+            let evars = env.map { (arg) -> UnsafeMutablePointer<Int8>? in
+                let (key, value) = arg
+                return UnsafeMutablePointer(strdup("\(key)=\(value)"))
+            }
+            envp.initialize(from: evars, count: 1 + nenv)
```
phausler: hmm there is a size mismatch, nenv is the count of elements in env, so the initialization will be reading past the mapped value.

The expectation here is the envp has a "null terminator"

Foundation/NSValue.swift
spevans: See https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20170515/005442.html

Its due to the clash of the 'type' argument and the 'type' function call

phausler: Hmmph that still seems like a bug to me... type as the variable seems quite clear that it is not invoked as a function especially since it has a named parameter. But ok, I guess chock it up to the same failure as print and NSView.print :/

Foundation/Process.swift
spevans: I was actually going to revert this change since there was some discussion on the closure variables that @parkera  raised (the let (key, value) = arg change).

Isnt the mapped size correct since its allocating 1 + env, but the actual issue is we need to ensure its set to NULL to terminate the **env array?

spevans:
@parkera Could you also review these 2 fixes required for swift4 please

alblue:
@swift-ci please test

parkera:
These seem good to me.

spevans:
I think the test got broken due to github's API issues yesterday could someone retest please

pushkarnk:
Seems ready for merge

pushkarnk:

PR #1005, opened 2017-05-22
Ensure FileManager.createDirectory does not crash if directory already exists on Linux.

mattrajca:
Add a test for [SR-4298](https://bugs.swift.org/browse/SR-4298) to ensure calling `FileManager.createDirectory` does not crash if the directory already exists.

The underlying issue is already fixed in tip-of-tree Swift.


parkera:
@swift-ci test and merge

PR #1006, opened 2017-05-22
Update Swift-mangled NSCFConstantString class name

e78l:
Fixing:
~~~~
ld: warning: undefined base symbol '__TMC15SwiftFoundation19_NSCFConstantString' for alias '___CFConstantStringClassReference'
Undefined symbols for architecture x86_64:
  "__T015SwiftFoundation21__CFSwiftGetBaseClassyXlXpyF", referenced from:
      ___CFInitialize in libCoreFoundation.a(CFRuntime.o)
ld: symbol(s) not found for architecture x86_64
~~~~
Split from https://github.com/apple/swift-corelibs-foundation/pull/988 (this has somewhat higher risk of breaking on some Macs, maybe)

### DIFF
--- a/CoreFoundation/Base.subproj/SymbolAliases
+++ b/CoreFoundation/Base.subproj/SymbolAliases
@@ -1,7 +1,7 @@
 
 
 # This is specific to Swift open source; Foundation's NSCFConstantString is used as our constant string class reference
-__TMC15SwiftFoundation19_NSCFConstantString ___CFConstantStringClassReference
+__T015SwiftFoundation19_NSCFConstantStringCN ___CFConstantStringClassReference
 
 _kCFCalendarIdentifierBuddhist		_kCFBuddhistCalendar
 _kCFCalendarIdentifierChinese		_kCFChineseCalendar


parkera:
Just to be clear on this: this is the mangled name produced by the top of tree Swift compiler right?

e78l:
Hello Tony,
Thanks for asking - I use snapshots on macOS, and when checking with [swift-DEVELOPMENT-SNAPSHOT-2017-05-23-a](https://swift.org/builds/development/xcode/swift-DEVELOPMENT-SNAPSHOT-2017-05-23-a/swift-DEVELOPMENT-SNAPSHOT-2017-05-23-a-osx.pkg), only the string warning appears:
~~~~
ld: warning: undefined base symbol '__TMC15SwiftFoundation19_NSCFConstantString' for alias '___CFConstantStringClassReference'
~~~~
Looks like the AnyObject part of the class name isn't in the Swift 4 branch.

So, I've dropped the CFRuntime init name change, but kept the SymbolAliases change.
As I mentioned in #988, I think it would be better to have a script determine these mangled names at build time.

parkera:
I think we're actually going to be ok for the long term once Swift 4 ships - changing mangling is a big deal and I'm hoping it calms down now.

PR #1007, opened 2017-05-23
URLRequest.timeoutInterval should override the URLSession.configurati…

saiHemak:
…on value[SR-2680]

### DIFF
--- a/TestFoundation/TestNSURLSession.swift
+++ b/TestFoundation/TestNSURLSession.swift
@@ -21,21 +21,23 @@ class TestURLSession : XCTestCase {
 
     static var allTests: [(String, (TestURLSession) -> () throws -> Void)] {
         return [
-            ("test_dataTaskWithURL", test_dataTaskWithURL),
-            ("test_dataTaskWithURLRequest", test_dataTaskWithURLRequest),
-            ("test_dataTaskWithURLCompletionHandler", test_dataTaskWithURLCompletionHandler),
-            ("test_dataTaskWithURLRequestCompletionHandler", test_dataTaskWithURLRequestCompletionHandler),
-            ("test_downloadTaskWithURL", test_downloadTaskWithURL),
-            ("test_downloadTaskWithURLRequest", test_downloadTaskWithURLRequest),
-            ("test_downloadTaskWithRequestAndHandler", test_downloadTaskWithRequestAndHandler),
-            ("test_downloadTaskWithURLAndHandler", test_downloadTaskWithURLAndHandler),
-            ("test_finishTaskAndInvalidate", test_finishTasksAndInvalidate),
-            ("test_taskError", test_taskError),
-            ("test_taskCopy", test_taskCopy),
-            ("test_cancelTask", test_cancelTask),
-            ("test_taskTimeout", test_taskTimeout),
-            ("test_verifyRequestHeaders", test_verifyRequestHeaders),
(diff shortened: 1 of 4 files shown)


TestFoundation/main.swift
```diff
@@ -80,7 +80,7 @@ XCTMain([
     testCase(TestNSURLResponse.allTests),
     testCase(TestNSHTTPURLResponse.allTests),
 //Disabling to avoid https://bugs.swift.org/browse/SR-4677 and a timeout failure
```
pushkarnk: We'd need to move this comment to TestNSURLSession.swift.

TestFoundation/TestNSURLSession.swift
```diff
@@ -409,6 +410,32 @@ class TestURLSession : XCTestCase {
...
+        serverReady.wait()
+        let config = URLSessionConfiguration.default
+        config.timeoutIntervalForRequest = 10
+        let session = URLSession(configuration: config, delegate: nil, delegateQueue: nil)
+        var expect = expectation(description: "download task with handler")
+        var req = URLRequest(url: URL(string: "http://127.0.0.1:\(serverPort)/Peru")!)
+        req.timeoutInterval = 1
+        var task = session.dataTask(with: req) { (data, _, error) -> Void in
+            defer { expect.fulfill() }
+            XCTAssertNotNil(error)
```
pushkarnk: Can we print the error info if this fails? Something like:
`XCTAssertNotNil(error as? URLError, "error = \(error as! URLError)")`

TestFoundation/TestNSURLSession.swift
```diff
@@ -21,7 +21,7 @@ class TestURLSession : XCTestCase {
 
     static var allTests: [(String, (TestURLSession) -> () throws -> Void)] {
         return [
-            ("test_dataTaskWithURL", test_dataTaskWithURL),
+/*            ("test_dataTaskWithURL", test_dataTaskWithURL),
```
mattrajca: I don't think you meant to comment out all these tests.

saiHemak: @mmatrajca  Currently ```URLSessionTests``` are disabled in main.swift due to the  failures  observed in CI environment. Through this PR I am enabling  my test alone  and disabled all other tests  as before . @pushkarnk Do you want me to enable all tests ?

pushkarnk: @saiHemak No, we shouldn't enable them until we get to the bottom of the CI failures. I'd suggest you have single line comments disabling each of the tests (instead of a multiline comment), with your test enabled. We should try enabling the other tests in small iterations (giving the CI enough time to reproduce the failures we've been seeing).

TestFoundation/TestNSURLSession.swift
saiHemak: I am expecting the error to be thrown in the passing scenario as the ```timeouInterval``` set for the request  is less than the ```sendDelay```  and hence am not printing the error.

pushkarnk:
Just mentioning a couple of things:
1. I realise that the addition of a new internal var to URLRequest , though undesirable,  is inevitable if we have to emulate Darwin behaviour related to timeout precedences. 
2. We are going to start enabling URLSession tests in iterations, beginning with this PR. I don't expect this test to fail but we'd need to monitor any CI failures starting today. 
Thanks!

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci merge

PR #1008, opened 2017-05-24
Add support for currentDirectoryPath in Process class

mattrajca:
This patch adds support for the `currentDirectoryPath` property on `Process`

Since `posix_spawn` doesn't make it easy to change the working directory, a
call to `chdir` is issued just before the process executes. It looks like
`lldb` follows a similar pattern in one of its tools
(`tools/darwin-debug/darwin-debug.cpp`).

This also addresses [SR-3524](https://bugs.swift.org/browse/SR-3524)

### DIFF
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -387,10 +387,29 @@ open class Process: NSObject {
             posix(posix_spawn_file_actions_addclose(&fileActions, fd))
         }
 
+        let fileManager = FileManager()
+        let previousDirectoryPath = fileManager.currentDirectoryPath
+        if !fileManager.changeCurrentDirectoryPath(currentDirectoryPath) {
+            // Foundation throws an NSException when changing the working directory fails,
+            // and unfortunately launch() is not marked `throws`, so we get away with a
+            // fatalError.
+            switch errno {
+            case ENOENT:
+                fatalError("Process: The specified working directory does not exist.")
+            case EACCES:
+                fatalError("Process: The specified working directory cannot be accessed.")
+            default:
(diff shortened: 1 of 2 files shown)


Foundation/Process.swift
```diff
@@ -387,6 +387,30 @@ open class Process: NSObject {
...
         }
 
+        let currentDirectoryURL = URL(fileURLWithPath: currentDirectoryPath)
+        let changeDirectorySucceeded = currentDirectoryURL.withUnsafeFileSystemRepresentation { rawPath -> Bool in
+            guard let rawPath = rawPath else {
+                return false
+            }
+
+            // posix_spawn doesn't support passing a custom working directory, so resort to chdir.
+            return chdir(rawPath) == 0
```
parkera: Am I missing something, or don't we have to reset this value after we're done with the launch?

mattrajca: @parkera Good catch, I updated the commit.

mattrajca:
@parkera please review the revised commit.

parkera:
Looks good, thanks. Can you just do one more thing for me and add a couple lines to the test to 1. get the CWD before/after and comparing, to verify that we did that correctly too.

mattrajca:
@parkera: Done! Please re-review the latest commit.

alblue:
@swift-ci please test

mattrajca:
@parkera Good to go?

parkera:
Yes, thanks for taking this one on!

parkera:
@swift-ci test and merge

PR #1012, opened 2017-05-26
Eliminate duplicate rules in generated build.ninja caused by PR #873

pushkarnk:
Changes of #873 produce these warnings in ninja for each target:
```
ninja: warning: multiple rules generate ../Ninja-ReleaseAssert/foundation-linux-x86_64/Foundation/Foundation/Date.swift.o. builds involving this target will not be correct; continuing anyway
```
The problem is that in the process of generating a static library and a dynamic library, we're generating compile rules for all the targets twice. The right thing to do is to generate them once (for the static library) and simply reuse the objects in the rule to generate the dynamic library.

### DIFF
--- a/lib/product.py
+++ b/lib/product.py
@@ -77,11 +77,12 @@ class Library(Product):
     def __init__(self, name):
         Product.__init__(self, name)
 
-    def generate(self, flags):
-        generated = Product.generate(self)
-        objects = []
-        for phase in self.phases:
-            objects += phase.objects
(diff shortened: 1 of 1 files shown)


pushkarnk:
@parkera can you review this please?

parkera:
@swift-ci test and merge

phausler:
looks good to me, so effectively there is only one build for a .o but two products that consume .o files?

PR #1014, opened 2017-05-26
Fix typo in renamed function SR-5015

alblue:


alblue:
@swift-ci please test

PR #1015, opened 2017-05-27
Add CFUserNotification.h to the CoreFoundation build process in Xcode to avoid importing the system version

phausler:
CFUserNotification.h could potentially be imported from the system CoreFoundation in certain build configurations. This avoids the mismatch of header imports so that the system CoreFoundation is never imported when building the swift-corelibs-foundation version of CoreFoundation.


phausler:
@swift-ci please smoke test and merge

phausler:
@swift-ci smoke test

PR #1016, opened 2017-05-27
&>> and &<< is the new >> and <<

phausler:
Swift 4 marked >> and << as obsolete. Since the intent of ALL usage cases were easily understood as masking shifts this was an automated find and replace of >> to &>> and << to &<<.


phausler:
@swift-ci please test

spevans:
Are you sure this isnt due to the protocol _IntegerLike in NSScanner.swift? I made the same change in a PR but then closed it as I thought the issue was something else? see my last comment in https://github.com/apple/swift-corelibs-foundation/pull/997

spevans:
FYI, using a swift playground with latest nightly snapshot, << seems to still be valid

phausler:
This is a preparation for Swift 4 which marks the bit shift operators as obsolete :(
I was attempting to prep all for all of the incoming naming changes etc.

phausler:
The offender is in Integers.swift.gyb

```swift
  @available(swift, obsoleted: 4) // <--- here
  @_transparent
  public static func ${op.nonMaskingOperator}(
    lhs: Self, rhs: Self
  ) -> Self {
```

spevans:
The issue here is due to the integer changes in https://github.com/apple/swift-evolution/blob/master/proposals/0104-improved-integers.md

To quote: "Shifts and other bitwise operations were moved from `FixedWidthInteger` to ` BinaryInteger`" so they were only marked as obsolete in `FixedWidthInteger`

However because of this in NSScanner.swift:
```
// FIXME(integers): replace this protocol with just a FixedWidthInteger
internal protocol _IntegerLike : FixedWidthInteger, _BitShiftable {
    init(_ value: Int)
    static var max: Self { get }
    static var min: Self { get }
}

extension Int : _IntegerLike { }
extension Int32 : _IntegerLike { }
extension Int64 : _IntegerLike { }
extension UInt32 : _IntegerLike { }
extension UInt64 : _IntegerLike { }
```

The Int, Int32 etc are now conforming  to `FixedWidthInteger` and so are picking up the obsoleted << and >> instead of getting the new ones in `BinaryInteger`.

I think the correct fix may be something along the lines of:

```
diff --git a/Foundation/NSScanner.swift b/Foundation/NSScanner.swift
index 3da53c9..8fddb81 100644
--- a/Foundation/NSScanner.swift
+++ b/Foundation/NSScanner.swift
@@ -225,13 +225,13 @@ internal protocol _IntegerLike : FixedWidthInteger, _BitShiftable {
     static var max: Self { get }
     static var min: Self { get }
 }
-
+/*
 extension Int : _IntegerLike { }
 extension Int32 : _IntegerLike { }
 extension Int64 : _IntegerLike { }
 extension UInt32 : _IntegerLike { }
 extension UInt64 : _IntegerLike { }
-
+*/
 private func numericValue(_ ch: unichar) -> Int {
     if (ch >= unichar(unicodeScalarLiteral: "0") && ch <= unichar(unicodeScalarLiteral: "9")) {
         return Int(ch) - Int(unichar(unicodeScalarLiteral: "0"))
@@ -264,7 +264,7 @@ private func decimalSep(_ locale: Locale?) -> String {
 }
 
 extension String {
-    internal func scan<T: _IntegerLike>(_ skipSet: CharacterSet?, locationToScanFrom: inout Int, to: (T) -> Void) -> Bool {
+    internal func scan<T: FixedWidthInteger>(_ skipSet: CharacterSet?, locationToScanFrom: inout Int, to: (T) -> Void) -> Bool {
         var buf = _NSStringBuffer(string: self, start: locationToScanFrom, end: length)
         buf.skip(skipSet)
         var neg = false
@@ -301,7 +301,7 @@ extension String {
         return true
     }
     
-    internal func scanHex<T: _IntegerLike>(_ skipSet: CharacterSet?, locationToScanFrom: inout Int, to: (T) -> Void) -> Bool {
+    internal func scanHex<T: FixedWidthInteger>(_ skipSet: CharacterSet?, locationToScanFrom: inout Int, to: (T) -> Void) -> Bool {
         var buf = _NSStringBuffer(string: self, start: locationToScanFrom, end: length)
         buf.skip(skipSet)
         var localResult: T = 0
```

ianpartridge:
I can confirm that @spevans 's patch fixes all the `<< is now obsolete` compilation failures with latest toolchains.

alblue:
There's a conflict in one of the files - can you rebase?

parkera:
Let's close this PR and open a new one. @spevans would you mind putting it up?

spevans:
No problem, I will sort out a PR in a bit

PR #1017, opened 2017-05-27
Make Foundation compile in Xcode again

spevans:
- The code is still using version 3 syntax so pass the
  -swift-version 3 flag


spevans:
@phausler This allowed foundation to compile in Xcode with latest toolchain

ianpartridge:
It seems like we're not too far away from being Swift 4 compliant, so maybe hold off on this?

parkera:
@swift-ci test and merge

parkera:
Actually let's just merge this since it's Mac only.

PR #1020, opened 2017-05-31
[SR-3202] Some of properties of Locale.current object are not filled …

saiHemak:
…on Linux

### DIFF
--- a/Foundation/NSLocale.swift
+++ b/Foundation/NSLocale.swift
@@ -28,7 +28,7 @@ open class NSLocale: NSObject, NSCopying, NSSecureCoding {
     }
     
     open func object(forKey key: NSLocale.Key) -> Any? {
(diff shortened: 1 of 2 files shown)


alblue:
@swift-ci please test

ianpartridge:
Failure caused by GitHub outage.

pushkarnk:
OK, I get what the problem was. 

`CFLocaleGetValue` called from `object(forKey:)` always returns any AnyObject. Subsequently for the different Locale properties, when we tried to conditionally cast it to a String, a nil was returned (this is what SR-3202 is). 

This fix first converts the AnyObject/NSString to String by using `_SwiftValue.fetch`, due to which a forced downcast in `localeIdentifier` fails . The second part of the fix removes the forced downcast.

pushkarnk:
LGTM

PR #1022, opened 2017-05-31
Update to Swift 4

spevans:
Changes only compatible with swift 4.

Depends on:
~~https://github.com/apple/swift-corelibs-foundation/pull/995~~ (but requires SE-110 to be reverted)
~~https://github.com/apple/swift-corelibs-foundation/pull/1023~~
~~https://github.com/apple/swift-corelibs-foundation/pull/1024~~

### DIFF
--- a/Foundation/NSNotificationQueue.swift
+++ b/Foundation/NSNotificationQueue.swift
@@ -109,16 +109,16 @@ open class NotificationQueue: NSObject {
         var predicate: (NSNotificationListEntry) -> Bool
         switch coalesceMask {
         case [.onName, .onSender]:
-            predicate = { (entryNotification, _) in
-                return _SwiftValue.store(notification.object) !== _SwiftValue.store(entryNotification.object) || notification.name != entryNotification.name
+            predicate = { entry in
+                return _SwiftValue.store(notification.object) !== _SwiftValue.store(entry.0.object) || notification.name != entry.0.name
             }
         case [.onName]:
-            predicate = { (entryNotification, _) in
-                return notification.name != entryNotification.name
+            predicate = { entry in
+                return notification.name != entry.0.name
             }
         case [.onSender]:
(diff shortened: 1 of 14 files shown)


spevans:
No there are still a few closure issues that popped up. I saw that some of SE-110 was reverted but I dont think it has been fully done. I will do a quick test and see what is still broken

spevans:
There were some changes still needed but mostly it was listing the exact number of arguments before `in`. Ive added a 2nd commit with the changes and it tested ok on Linux.

The PR for the rollback of SE110 is https://github.com/apple/swift/pull/10414

spevans:
@parkera If you are happy with the closure changes I think this is good to go

Foundation/NSNotificationQueue.swift
```diff
@@ -109,16 +109,16 @@ open class NotificationQueue: NSObject {
         var predicate: (NSNotificationListEntry) -> Bool
         switch coalesceMask {
         case [.onName, .onSender]:
-            predicate = { (entryNotification, _) in
-                return _SwiftValue.store(notification.object) !== _SwiftValue.store(entryNotification.object) || notification.name != entryNotification.name
+            predicate = { entry in
```
parkera: Could this one have remained as previous with the SE-110 revert?

parkera:
One small question but otherwise looks good

Foundation/NSNotificationQueue.swift
spevans: Unfortunately these three still gave the error 'does not support destructuring' even after I checked it with the latest compiler, but it was the only ones.

parkera: Maybe we should follow up on that separately but it shouldn't hold this up.

parkera:
@swift-ci test and merge

PR #1024, opened 2017-06-02
swift4: String fixes

spevans:
- Failable String initialiser has been obsoleted in swift4.

- Subscripts returning String were obsoleted in Swift 4,
  explicitly construct a String from subscripted result.

Tested on swift 3 and 4.


alblue:
@swift-ci please test

PR #1025, opened 2017-06-03
Fix `NSNumber.objCType` crashes, make it returns same values with darwin platform version

norio-nomura:
### Fixes [SR-4907](https://bugs.swift.org/browse/SR-4907)
- Add `TestNSNumber.test_objCType()` reproducing crash
  Expected results are sampled from macOS
- Fix crash on accessing `objCType` property of `NSNumber`  

### Change `NSNumber.objCType` to returning same values with darwin platform version
- Change `NSNumber.objCType` to returning same values with darwin
- Change `NSNumber.init(value: Bool)` to returning `__NSCFBoolean` by using `_Factory` protocol

Since `NSNumber.objCType` returns `c` with both of `Bool` and `Int8` on darwin platform, `NSNumber.init(value: Bool)` is changed to returns `__NSCFBoolean` by using `_Factory` protocol, and it changed to use `_cfTypeID` to distinguish between Bool and Int8.

- Change `JSONSerialization` to use `_cfTypeID` for checking whether `NSNumber` is Boolean or not
- Change `NSNumber.compare(_:)` to use `objCType`
- Add missing entry to `TestNumber.allTests`
- Fix `NSKeyedUnarchiver.decodeBool(forKey:)` and add test to `TestNSPredicate.test_NSCoding()`

### DIFF
--- a/Foundation/NSCFBoolean.swift
+++ b/Foundation/NSCFBoolean.swift
@@ -78,7 +78,7 @@ internal class __NSCFBoolean : NSNumber {
     override var objCType: UnsafePointer<Int8> {
         // This must never be fixed to be "B", although that would
         // cause correct old-style archiving when this is unarchived.
-        return UnsafePointer<Int8>(bitPattern: UInt(_NSSimpleObjCType.Char.rawValue.value))!
+        return String(_NSSimpleObjCType.Char.rawValue)._bridgeToObjectiveC().utf8String!
     }
     
     internal override func _cfNumberType() -> CFNumberType  {
--- a/Foundation/NSValue.swift
+++ b/Foundation/NSValue.swift
@@ -154,3 +154,14 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
 }
 
+extension NSValue : _Factory {}
+
+protocol _Factory {
+    init(factory: () -> Self)
+}
+
+extension _Factory {
+    init(factory: () -> Self) {
+        self = factory()
+    }
+}
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -458,8 +458,8 @@ private struct JSONWriter {
             throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Number cannot be infinity or NaN"])
         }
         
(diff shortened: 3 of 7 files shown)


norio-nomura:
Reflected differences of `NSNumber.objCType`'s result by value and architectures.

alblue:
@swift-ci please test

Foundation/NSCFBoolean.swift
```diff
@@ -78,7 +78,7 @@ internal class __NSCFBoolean : NSNumber {
     override var objCType: UnsafePointer<Int8> {
         // This must never be fixed to be "B", although that would
         // cause correct old-style archiving when this is unarchived.
-        return UnsafePointer<Int8>(bitPattern: UInt(_NSSimpleObjCType.Char.rawValue.value))!
+        return String(_NSSimpleObjCType.Char.rawValue)._bridgeToObjectiveC().utf8String!
```
phausler: this is going to cause leaks, can we have a static allocation somewhere stashed with the type to cstrings?

Foundation/NSValue.swift
```diff
@@ -154,3 +154,14 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
...
 
+extension NSValue : _Factory {}
+
+protocol _Factory {
+    init(factory: () -> Self)
+}
+
+extension _Factory {
+    init(factory: () -> Self) {
+        self = factory()
```
phausler: hmm this is rather strange, if this works perhaps we should investigate using this pattern further

Foundation/NSValue.swift
```diff
@@ -154,3 +154,14 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
 }
 
+extension NSValue : _Factory {}
+
+protocol _Factory {
```
phausler: just for maintainability; can we annotate this as internal?

Foundation/NSValue.swift
phausler: ok, just prototyped a test for this... wow! this will solve a ton of issues with swift-corelibs-foundation!

rintaro: Just FYI https://bugs.swift.org/browse/SR-5255

norio-nomura:
Superseded by #1093

norio-nomura:
@phausler Thanks for reviews. Those are addressed on #1093.

PR #1026, opened 2017-06-04
Fix build on ARM: Undefined NativeType

uraimo:
It looks like that `arch(powerpc)` is not available anymore (and it's never used in swift or any other repository).
Using it in the first of these two conditionals was breaking the build (swift-4.0-branch, this is not an issue when compiling swift-3.1.1-RELEASE) on ARM because NativeType was never being defined.

I've verified that *powerpc* is not available also with the Swift 3.1.1 build shipped with Xcode on macOS with this simple example:

```swift
#if arch(powerpc)
      print("test")
#endif
```
This is the output of swiftc:

```
ppc.swift:1:10: warning: unknown architecture for build configuration 'arch'
#if arch(powerpc)
         ^
ppc.swift:1:10: note: did you mean 'powerpc64'?
#if arch(powerpc)
         ^~~~~~~
         powerpc64
```

I wonder why this wasn't a problem for other platforms.

### DIFF
--- a/Foundation/CGFloat.swift
+++ b/Foundation/CGFloat.swift
@@ -9,7 +9,7 @@
 
 @_fixed_layout
 public struct CGFloat {
-#if arch(i386) || arch(arm) || arch(powerpc)
+#if arch(i386) || arch(arm)
     /// The native type used to store the CGFloat, which is Float on
     /// 32-bit architectures and Double on 64-bit architectures.
     public typealias NativeType = Float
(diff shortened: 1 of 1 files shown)


CodaFi:
Thanks!

CodaFi:
@ruscur If you want a `ppc32` target you'll have to change the compiler as well.

rintaro:
Ah, that breakage was my fault. Thanks Umberto!

uraimo:
No problem @rintaro , happy to help ;)

PR #1028, opened 2017-06-05
Reapply "Update for AnyObject nukage"

benlangmuir:
Reverts apple/swift-corelibs-foundation#976

After the rebranch of swift-4.0-branch from master, we need this update.

### DIFF
--- a/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/CoreFoundation/Base.subproj/CFRuntime.c
@@ -1008,9 +1008,9 @@ void __CFInitialize(void) {
         
 #ifndef __CFSwiftGetBaseClass
 #if TARGET_OS_LINUX
-#define __CFSwiftGetBaseClass _T010Foundation21__CFSwiftGetBaseClasss9AnyObject_pXpyF
+#define __CFSwiftGetBaseClass _T010Foundation21__CFSwiftGetBaseClassyXlXpyF
 #elif TARGET_OS_MAC
-#define __CFSwiftGetBaseClass _T015SwiftFoundation21__CFSwiftGetBaseClasss9AnyObject_pXpyF
+#define __CFSwiftGetBaseClass _T015SwiftFoundation21__CFSwiftGetBaseClassyXlXpyF
 #endif
 #endif
         extern uintptr_t __CFSwiftGetBaseClass();


parkera:
@benlangmuir we should be automatically merging into swift-4.0-branch from master still for this project.

benlangmuir:
@parkera because the original commit was on both branches and then reverted on swift-4, merging master into swift-4 wouldn't be enough; git would treat it as already being merged.

parkera:
Ok; we shouldn't be making any commits to this branch until after we stop merging.

benlangmuir:
I don't understand.  If we didn't make a commit to this branch it would have stayed broken.  I have no intention of making any further commits.  I was just fixing the build.  Merging master into this branch would not have fixed the build.

parkera:
I'm saying that from now until swift-4.0-branch does not auto merge, no commits should be made to this branch. Therefore, there will be no need to manually fix up the result of a revert on master. git will merge the revert correctly.

benlangmuir:
You have it backwards for this case.  Slava's original change was on both master and swift-4.  The revert was on swift-4 only.  There was no new commit on master that needed to be merged.  I'm effectively undoing a swift-4-specific change so that master and swift-4 are back in sync here.

benlangmuir:
To make sure we're understanding each other, this is the demonstration of the fact the merge does nothing here.  The state of swift-4.0-branch before this pull request was 167ce64d49cec34407b9efbedef0dda5d111d0e0.
```
$ git checkout -b tmp 167ce64d49cec34407b9efbedef0dda5d111d0e0
Switched to a new branch 'tmp'

$ git merge origin/master
Already up-to-date.
```

This is because the history is:
```
$ git log --oneline origin/master -- CoreFoundation/Base.subproj/CFRuntime.c | head -2
57331473 Update for AnyObject nukage
e4cfd27e Implementation of XDG File specification and a basic implementation of HTTPCookieStorage (#889)

$ git log --oneline origin/swift-4.0-branch  -- CoreFoundation/Base.subproj/CFRuntime.c | head -4
b38f9240 Revert "Revert "Update for AnyObject nukage""
5f5f27cd Revert "Update for AnyObject nukage"
57331473 Update for AnyObject nukage
e4cfd27e Implementation of XDG File specification and a basic implementation of HTTPCookieStorage (#889)
```

parkera:
Ben, you're missing my point: I understand why you made this commit. I'm saying that I don't want any more direct commits to this branch after now, from anyone. Therefore we don't have to manually keep track of how swift-4.0-branch and master differ until we actually want them to diverge.

I don't want to be continuously cherry picking changes to this branch for the next N months.

PR #1029, opened 2017-06-06
Make `totalDownloaded` to be specific to HTTPURLProtocol

nethraravindran:


alblue:
Why should this be specific to an HTTPURLDownloadTask? Other URL download types could potentially have a total download as well.

pushkarnk:
@swift-ci please test

nethraravindran:
@alblue  `totalDownloaded` is being used to track the progress of HTTP downloads. Other protocols may use their own mechanism to track downloads. Hence it makes more sense to make `totalDownloaded` specific to HTTPURLProtocol for now.

pushkarnk:
Looks good to me, for now. I also agree that, in future, we are going to have some common attributes between the different protocols and we have to figure out the best way to factor them out so that there is no redundancies - whether URLSessionTask is the best place to hold these attributes or we need a generic protocol class is something we'd need to look into.

pushkarnk:

pushkarnk:

PR #1031, opened 2017-06-07
Eliminate optional chaining in URLSession

nethraravindran:


alblue:
This looks like it's a change in behaviour - you're replacing optional uses with fatalErrors should they not be what you expect. If you're going to do that, you might as well have just used forced unwraps instead.

It's not clear to me that this is a sensible or desired change though. What was the thought behind this?

Foundation/NSURLProtocol.swift
```diff
@@ -148,7 +148,8 @@ public protocol URLProtocolClient : NSObjectProtocol {
 internal class _ProtocolClient : NSObject, URLProtocolClient {
 
     func urlProtocol(_ protocol: URLProtocol, didReceive response: URLResponse, cacheStoragePolicy policy: URLCache.StoragePolicy) {
-        `protocol`.task?.response = response
+	guard let task = `protocol`.task else { fatalError() }
```
alblue: This is a breaking change in behaviour. Previously it would be a no-operation, now it will crash call. It's not clear whether this is beneficial and certainly doesn't support the message that this PR was filed with.

nethraravindran:
@alblue Thanks for reviewing! These crashes are very unlikely to occur. Even if it crashes in very rare scenario, I believe fatal errors will give more information to the user about the crash than halting the application with no operation.

alblue:
They are implemented, and caught, in exactly the same way using force unwrapping:

```
fatalError()

fatal error: file tst.swift, line 1
Current stack trace:
0    libswiftCore.so                    0x00007fc137fda1c0 swift_reportError + 120
1    libswiftCore.so                    0x00007fc137ff4ad0 _swift_stdlib_reportFatalErrorInFile + 100
2    libswiftCore.so                    0x00007fc137def40c <unavailable> + 1188876
3    libswiftCore.so                    0x00007fc137f8737d <unavailable> + 2859901
4    libswiftCore.so                    0x00007fc137deebe6 <unavailable> + 1186790
5    libswiftCore.so                    0x00007fc137f8d290 <unavailable> + 2884240
6    libswiftCore.so                    0x00007fc137def01f <unavailable> + 1187871
7    libswiftCore.so                    0x00007fc137f4dfa9 <unavailable> + 2625449
8    libswiftCore.so                    0x00007fc137deebe6 <unavailable> + 1186790
9    libswiftCore.so                    0x00007fc137f0b260 specialized _assertionFailure(StaticString, String, file : StaticString, line : UInt, flags : UInt32) -> Never + 144
10   tst                                0x0000000000400a30 <unavailable> + 2608
11   libc.so.6                          0x00007fc137083740 __libc_start_main + 240
12   tst                                0x00000000004008b9 <unavailable> + 2233
Illegal instruction
```

```
let a:String? = nil
a!

fatal error: unexpectedly found nil while unwrapping an Optional value
Current stack trace:
0    libswiftCore.so                    0x00007f254974f1c0 swift_reportError + 120
1    libswiftCore.so                    0x00007f2549769be0 _swift_stdlib_reportFatalError + 62
2    libswiftCore.so                    0x00007f2549563be6 <unavailable> + 1186790
3    libswiftCore.so                    0x00007f25496c2ead <unavailable> + 2625197
4    libswiftCore.so                    0x00007f2549563be6 <unavailable> + 1186790
5    libswiftCore.so                    0x00007f254967e060 specialized _fatalErrorMessage(StaticString, StaticString, file : StaticString, line : UInt, flags : UInt32) -> Never + 96
6    tst                                0x0000000000400afa <unavailable> + 2810
7    libc.so.6                          0x00007f25487f8740 __libc_start_main + 240
8    tst                                0x00000000004008e9 <unavailable> + 2281
Illegal instruction
```

Arguably by hiding the fact that it was an unwrap of an optional value, you have resulted in less information being displayed about the error rather than increased it.

Can you tell me why you think this will halt without displaying information when using a force unwrap? Or why crashing is the preferred behaviour instead of just using optional chaining, which was designed to avoid crashing in this way?

nethraravindran:
Thanks @alblue. I agree with you.

PR #1037, opened 2017-06-12
Update the mangled name of CFSwiftGetBaseClass

S2dentik:
* Without this, CoreFoundation isn't properly linked.

### DIFF
--- a/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/CoreFoundation/Base.subproj/CFRuntime.c
@@ -1010,7 +1010,7 @@ void __CFInitialize(void) {
 #if TARGET_OS_LINUX
 #define __CFSwiftGetBaseClass _T010Foundation21__CFSwiftGetBaseClassyXlXpyF
 #elif TARGET_OS_MAC
-#define __CFSwiftGetBaseClass _T015SwiftFoundation21__CFSwiftGetBaseClassyXlXpyF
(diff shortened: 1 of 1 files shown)


S2dentik:
@parkera

spevans:
If you use the latest nightly snapshot I dont think this is needed anymore. At some point the name mangling must have changed as I just tested Foundation on 8.3.3 and 9.0beta with the 2017-06-11 nightly and this fix isnt required now.

parkera:
paging @jrose-apple to explain what to expect with mangling here...

parkera:
I did this exact same thing but then the XCTest project failed because the mangled name was wrong.

jrose-apple:
AnyObject is now magic instead of being a normal protocol, and so it's now mangled specially. @slavapestov?

S2dentik:
@parkera, @spevans  yeah, you are right, I've also downloaded the latest toolchain and it now works, thanks everyone. Closing this.

S2dentik:
With the latest toolchain however `TestFoundation` doesn't build because of `<unknown>:0: error: unknown argument: '-index-store-path'`. Is it something we can fix?

spevans:
Its an issue with the Xcode 9 beta and swift.org toolchains. This should fix it:

`defaults write com.apple.dt.Xcode IDEIndexEnableBoltIndex NO`

(see https://twitter.com/jckarter/status/872899857463205888)

S2dentik:
Thanks @spevans

PR #1038, opened 2017-06-12
build: Fix TestFoundation to build and run in Xcode

spevans:
- The paths to the built frameworks needed updating so that
  TestFoundation will now correctly build in Xcode and link
  with XCTest.

- Use the correct name to the xdgTestHelper which is different on
  Linux and macOS.

- The test HTTP server now listens on 127.0.0.1 instead of 0.0.0.0
  so that the firewall dialog doesnt popup on macOS if the firewall
  is enabled.

This PR works in conjunction with https://github.com/apple/swift-corelibs-xctest/pull/192 for swift-corelibs-xctest

This will fix building SwiftFoundation and then running TestFoundation in Xcode, correctly linking to SwiftXCTest. It requires the xctest PR listed above to work. Some of the tests are still broken on macOS but this should make it easier to fix them now. Also one bug was found by the malloc debugger so that fix is included here.

Tested on Xcode 8.3.3 and 9.0 beta with 2017-06-11 nightly toolchain.


parkera:
@swift-ci test and merge

PR #1041, opened 2017-06-13
Fix some more warnings

spevans:
Remove more uses of unsafeBitCast()

- NSXMLParser.swift: Remove all calls to strlen() except for a
  single use in UTF8String().

- Replace _colonSeparatedStringFromPrefixAndSuffix() with direct
  string construction.

- Replace a call to strncpy() with String._fromCodeUnitSequence()
  to simplify extracting a non-null terminated string from a
  buffer.

swift4: DispatchData(bytes:) is deprecated

- Use init(bytes: UnsafeRawBufferPointer) instead.

Tested on Linux and macOS

### DIFF
--- a/Foundation/NSXMLParser.swift
+++ b/Foundation/NSXMLParser.swift
@@ -237,11 +237,6 @@ internal func _NSXMLParserEndDocument(_ ctx: _CFXMLInterface) -> Void {
     }
 }
 
-internal func _colonSeparatedStringFromPrefixAndSuffix(_ prefix: UnsafePointer<UInt8>, _ prefixlen: UInt, _ suffix: UnsafePointer<UInt8>, _ suffixLen: UInt) -> String {
-    let prefixString = String._fromCodeUnitSequence(UTF8.self, input: UnsafeBufferPointer(start: prefix, count: Int(prefixlen)))
-    let suffixString = String._fromCodeUnitSequence(UTF8.self, input: UnsafeBufferPointer(start: suffix, count: Int(suffixLen)))
-    return "\(prefixString!):\(suffixString!)"
-}
 
 internal func _NSXMLParserStartElementNs(_ ctx: _CFXMLInterface, localname: UnsafePointer<UInt8>, prefix: UnsafePointer<UInt8>?, URI: UnsafePointer<UInt8>?, nb_namespaces: Int32, namespaces: UnsafeMutablePointer<UnsafePointer<UInt8>?>, nb_attributes: Int32, nb_defaulted: Int32, attributes: UnsafeMutablePointer<UnsafePointer<UInt8>?>) -> Void {
     let parser = ctx.parser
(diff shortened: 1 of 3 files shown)


Catfish-Man:
In general I'm somewhat concerned about the performance impact of these changes. I haven't profiled the Swift implementation much, but the ObjC NSXMLParser's performance is *heavily* dependent on avoiding C string -> NSString conversions (e.g. UTF8STRING calls) whenever possible.

It's possible Swift String operations are sufficiently different that that's not a concern here, but it would be good to check.

spevans:
@Catfish-Man This is a cleanup and shouldnt introduce any extra calls. `UTF8STRING()` and the removed `_colonSeparatedStringFromPrefixAndSuffix()` are both using`String._fromCodeUnitSequence()` so the underlying method of string creation hasnt changed however this change does eliminate an unnecessary extra call to `UTF8STRING(ns)` and also `String._fromCodeUnitSequence()` being called on a fixed string ("xmlns").

`String._fromCodeUnitSequence()` does appear to copy the string but Im not sure how that copy can be eliminated if the strings are required to exist after the initial XML buffer data has been freed.

Foundation/NSXMLParser.swift
```diff
@@ -254,10 +249,8 @@ internal func _NSXMLParserStartElementNs(_ ctx: _CFXMLInterface, localname: Unsa
             var namespaceNameString: String?
             var asAttrNamespaceNameString: String?
             if let ns = namespaces[idx] {
-                if reportNamespaces {
-                    namespaceNameString = UTF8STRING(ns)
-                }
-                asAttrNamespaceNameString = _colonSeparatedStringFromPrefixAndSuffix("xmlns", 5, ns, UInt(strlen(unsafeBitCast(ns, to: UnsafePointer<Int8>.self))))
+                namespaceNameString = UTF8STRING(ns)
```
Catfish-Man: This sort of thing is what I was commenting on. Is this not making the initialization of namespaceNameString unconditional, where before it would only happen if we were reporting namespaces?

spevans: `_colonSeparatedStringFromPrefixAndSuffix` actually unconditionally parses `ns` already as `_colonSeparatedStringFromPrefixAndSuffix` and `UTF8STRING` both call `String._fromCodeUnitSequence`, so if `reportNamespaces` was true it was actually being parsed twice, once into `namespaceNameString` and also into part of `asAttrNamespaceNameString`. This should actually fix it to only be parsed once.

Catfish-Man: Ah, got it. Thanks for clarifying :)

spevans: No worries, thanks for looking it over, it made me double check the changes. Could you run a swift-ci test on this PR please.

phausler:
@swift-ci please test

spevans:
@parkera Do these changes look ok?

parkera:
LGTM thanks

PR #1042, opened 2017-06-14
Enable URLSesssion tests

pushkarnk:
Enabling the remaining URLSession tests. These had been disabled following some CI failures that we couldn't reproduce locally. Late last month we enabled the tests with completion handlers with some extra diagnostic messages. It looks like the CI has been stable for a while (fingers crossed!). This PR enables the remaining tests that use delegates.

### DIFF
--- a/TestFoundation/TestNSURLSession.swift
+++ b/TestFoundation/TestNSURLSession.swift
@@ -21,20 +21,19 @@ class TestURLSession : XCTestCase {
 
     static var allTests: [(String, (TestURLSession) -> () throws -> Void)] {
         return [
-//Disabling to avoid https://bugs.swift.org/browse/SR-4677 and a timeout failure
-//            ("test_dataTaskWithURL", test_dataTaskWithURL),
-//            ("test_dataTaskWithURLRequest", test_dataTaskWithURLRequest),
+            ("test_dataTaskWithURL", test_dataTaskWithURL),
+            ("test_dataTaskWithURLRequest", test_dataTaskWithURLRequest),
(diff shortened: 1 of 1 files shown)


pushkarnk:
@swift-ci please test

ianpartridge:
`TestFoundation/TestNSURLSession.swift:264: error: TestURLSession.test_finishTaskAndInvalidate : Asynchronous wait failed - Exceeded timeout of 12.0 seconds, with unfulfilled expectations: URLSession wasn't invalidated`

pushkarnk:
I'm looking at the failure. Doesn't look straightforward.

pushkarnk:
Created #1049 for the failure

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci please test

pushkarnk:
One of those intermittent failures. The completion handler was called with a non-nil `error` which translated to `URLError(_nsError: The operation could not be completed)`. This seems to be a very general error type - we have no info on what actually went wrong.

pushkarnk:
I'm running the test one more time.

@swift-ci please test

pushkarnk:

pushkarnk:
There are two types of URLSession tests - ones that use the completion handler and others that use delegates. With the tests that use delegates disabled, we seem to be doing good. However, after enabling those, the completion handler tests seem to be intermittently failing. In a debugging session today I realised that there can be some inter-test dependencies because though we are creating a new server for each test, we aren't making sure the previous server has completed and shutdown. There is a need to comprehend things better here. I am putting this PR on hold until then.

PR #1043, opened 2017-06-14
Remove temporary workaround for SR-3202

skyline75489:
With the help of #1020 I believe we can safely remove the workaround here.

CC @jszumski

### DIFF
--- a/Foundation/NSLengthFormatter.swift
+++ b/Foundation/NSLengthFormatter.swift
@@ -69,7 +69,7 @@ open class LengthFormatter : Formatter {
         //Extract the number from the measurement
         let numberInUnit = unitMeasurement.value
         
-        if isForPersonHeightUse && !numberFormatter.locale.sr3202_fix_isMetricSystemLocale() {
+        if isForPersonHeightUse && !numberFormatter.locale.usesMetricSystem {
             let feet = numberInUnit.rounded(.towardZero)
             let feetString = string(fromValue: feet, unit: .foot)
             
@@ -123,7 +123,7 @@ open class LengthFormatter : Formatter {
     /// - Parameter numberInMeters: the magnitude in terms of meters
     /// - Returns: Returns the appropriate unit
     private func unit(fromMeters numberInMeters: Double) -> Unit {
-        if numberFormatter.locale.sr3202_fix_isMetricSystemLocale() {
+        if numberFormatter.locale.usesMetricSystem {
             //Person height is always returned in cm for metric system
             if isForPersonHeightUse { return .centimeter }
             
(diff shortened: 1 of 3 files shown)


jszumski:
Looks good, thank you!

alblue:
@swift-ci please test

skyline75489:
It seems that Locale("en_UK") are returning same value as Locale("en_US") and causing the test failure on linux. 

The workaround certainly returns different value for "en_UK" and "en_US". I'm a little confused here. What's the expected behavior?

alblue:
@swift-ci please test

skyline75489:
Sorry for breaking the tests. I didn't expect such a early merge. I'm not really familiar with the different metrics in the western culture myself. Will look into it.

alblue:
Do we need to revert? It seemed to pass the runtime tests when I ran them.

skyline75489:
Yeah if that's OK. I can come up with a new PR for this.

pushkarnk:
Yeah, reverting only the tests seems like a good idea for now

skyline75489:
The whole commit should be reverted. The  GB  UK thing doesn't really matter.

alblue:
It looks like the issue is due to some state leftover with the regional tests:

https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-16_04/3658/console

```
Test Case 'TestNSLengthFormatter.test_stringFromMetersMetric' started at 2017-06-21 10:16:31.688
TestFoundation/TestNSLengthFormatter.swift:71: error: TestNSLengthFormatter.test_stringFromMetersMetric : XCTAssertEqual failed: ("-6.214 mi") is not equal to ("-10 km") - 
TestFoundation/TestNSLengthFormatter.swift:72: error: TestNSLengthFormatter.test_stringFromMetersMetric : XCTAssertEqual failed: ("-0.001 mi") is not equal to ("-0.001 km") - 
TestFoundation/TestNSLengthFormatter.swift:73: error: TestNSLengthFormatter.test_stringFromMetersMetric : XCTAssertEqual failed: ("0 in") is not equal to ("0.01 mm") - 
```

skyline75489:
Yeah, I'll have a look and try to figure out what `mi` actually means.

alblue:
It is short for 'miles' - it's saying that 10 kilometres is 6.241 miles. I think because something is selecting the en_GB locale that it's reporting large distances in miles and kilometres, and small distances in inches instead of millimetres.

alblue:
The build of #1056 has succeeded, so master is green again:

https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-16_04/3661/console

PR #1045, opened 2017-06-17
[build.py] Change the option of rsync to the same one used in other projects

norio-nomura:
This may fix the bad permission of CoreFoundation on Linux.
https://bugs.swift.org/browse/SR-2783
https://bugs.swift.org/browse/SR-2929
https://bugs.swift.org/browse/SR-4423

Even if this won't fix, this is harmless.


parkera:
@swift-ci test and merge

PR #1047, opened 2017-06-18
[NSCache] Fix incorrect priority queue operation in NSCache

zetasq:
Fix and optimize the logic for maintaining the internal priority queue NSCache.

Currently when multiple objects are set and removed in an NSCache object, the internal priority queue is not maintained correctly. The logic for maintaining the queue is correct now.


parkera:
Do we have some new tests to go with the fix?

zetasq:
@parkera I just add the `test_objectCorrectlyReleased` test in `TestNSCache` suite. This test fails in the original implementation but passed in this fix. I also changed some other tests in the `TestNSCache` suite to make them pass.

parkera:
@swift-ci test and merge

PR #1051, opened 2017-06-19
Add fast paths for Data initialization for common sequences

phausler:
There were some unfortunate slower than expected initializers on Data that were inherited from Collection adoption. This allows Data initialization to take the fast path in a select group of initializations; previously it would create a Data and then iterate byte by byte appending. Albeit that was relatively quick, we have faster options available.


alblue:
@swift-ci please test

PR #1052, opened 2017-06-19
Better namespace support for XML

rothomp3:
So I wrote most of this to begin with, and when I went to actually try and use it for a little project (playing around with WebDAV), it crashed! Turns out I never went back and did namespace support. Well, here is more of (but still not all of) namespace support.

Fix one place was still expecting C String

XPath works with namespaces

implement XMLElement(xmlString:)

Add some tests, re-enable all tests

Implement more namespaces support

### DIFF
--- a/CoreFoundation/Parsing.subproj/CFXMLInterface.c
+++ b/CoreFoundation/Parsing.subproj/CFXMLInterface.c
@@ -20,6 +20,7 @@
 #include <libxml/xmlmemory.h>
 #include <libxml/xmlsave.h>
 #include <libxml/xpath.h>
+#include <libxml/xpathInternals.h>
 #include <libxml/dict.h>
 #include "CFInternal.h"
 
@@ -57,6 +58,7 @@ CFIndex _kCFXMLTypeElement = XML_ELEMENT_NODE;
 CFIndex _kCFXMLTypeAttribute = XML_ATTRIBUTE_NODE;
 CFIndex _kCFXMLTypeDTD = XML_DTD_NODE;
 CFIndex _kCFXMLDocTypeHTML = XML_DOC_HTML;
+CFIndex _kCFXMLTypeNamespace = 22; // libxml2 does not define namespaces as nodes, so we have to fake it
 
 CFIndex _kCFXMLDTDNodeTypeEntity = XML_ENTITY_DECL;
 CFIndex _kCFXMLDTDNodeTypeAttribute = XML_ATTRIBUTE_DECL;
@@ -360,10 +362,6 @@ _CFXMLNodePtr _CFXMLNewProperty(_CFXMLNodePtr node, const unsigned char* name, c
     return xmlNewProp(node, name, value);
 }
 
-_CFXMLNamespacePtr _CFXMLNewNamespace(_CFXMLNodePtr node, const unsigned char* uri, const unsigned char* prefix) {
-    return xmlNewNs(node, uri, prefix);
-}
-
 CF_RETURNS_RETAINED CFStringRef _CFXMLNodeURI(_CFXMLNodePtr node) {
     xmlNodePtr nodePtr = (xmlNodePtr)node;
     switch (nodePtr->type) {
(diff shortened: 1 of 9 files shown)


alblue:
@swift-ci please test

CoreFoundation/Parsing.subproj/CFXMLInterface.h
```diff
@@ -153,7 +153,7 @@ void _CFXMLNodeSetPrivateData(_CFXMLNodePtr node, void* data);
 void* _Nullable  _CFXMLNodeGetPrivateData(_CFXMLNodePtr node);
 _CFXMLNodePtr _Nullable _CFXMLNodeProperties(_CFXMLNodePtr node);
 CFIndex _CFXMLNodeGetType(_CFXMLNodePtr node);
-const char* _Nullable _CFXMLNodeGetName(_CFXMLNodePtr node);
+CF_RETURNS_RETAINED CFStringRef _Nullable _CFXMLNodeGetName(_CFXMLNodePtr node);
```
parkera: For this and others - can we avoid the returns retained macro if we just follow the 'copy' convention, e.g.: `_CFXMLNodeCopyName`?

Foundation.xcodeproj/xcshareddata/xcschemes/TestFoundation.xcscheme
```diff
@@ -85,8 +85,8 @@
    </TestAction>
    <LaunchAction
       buildConfiguration = "Debug"
-      selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
-      selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      selectedDebuggerIdentifier = ""
+      selectedLauncherIdentifier = "Xcode.IDEFoundation.Launcher.PosixSpawn"
```
parkera: I think these two lines should be omitted from the PR

Foundation/NSXMLElement.swift
```diff
@@ -49,8 +49,15 @@ open class XMLElement: XMLNode {
         @method initWithXMLString:error:
         @abstract Returns an element created from a string. Parse errors are collected in <tt>error</tt>.
     */
-    public init(xmlString string: String) throws {
-        NSUnimplemented()
+    public convenience init(xmlString string: String) throws {
+        // If we prepend the XML line to the string
+        let docString = """
+        <?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\"?>\(string)
```
parkera: Mostly for my own education on the new triple-quote syntax: do we still need to backslash escape the quotes?

rothomp3: Nope, sure don't, I'll fix that.

alblue:
@swift-ci please test

CoreFoundation/Parsing.subproj/CFXMLInterface.h
rothomp3: Probably? I was mostly concerned with making extremely sure that Swift knew the correct retain count, because memory leaks are my personal crusade haha.

alblue:
@swift-ci please test

CoreFoundation/Parsing.subproj/CFXMLInterface.c
```diff
@@ -1179,6 +1236,131 @@ void _CFXMLDTDNodeSetPublicID(_CFXMLDTDNodePtr node, const unsigned char* public
     }
 }
 
+// Namespaces
+_CFXMLNodePtr _Nonnull * _Nullable _CFXMLNamespaces(_CFXMLNodePtr node, CFIndex* count) {
```
phausler: we might wanna use a CFArrayRef here because this requires freeing the return value

CoreFoundation/Parsing.subproj/CFXMLInterface.h
```diff
@@ -227,11 +228,22 @@ CFIndex _CFXMLDTDElementNodeGetType(_CFXMLDTDNodePtr node);
...
 
-CF_RETURNS_RETAINED CFStringRef _Nullable _CFXMLDTDNodeGetSystemID(_CFXMLDTDNodePtr node);
+CFStringRef _Nullable _CFXMLDTDNodeCopySystemID(_CFXMLDTDNodePtr node);
 void _CFXMLDTDNodeSetSystemID(_CFXMLDTDNodePtr node, const unsigned char* _Nullable systemID);
-CF_RETURNS_RETAINED CFStringRef _Nullable _CFXMLDTDNodeGetPublicID(_CFXMLDTDNodePtr node);
+CFStringRef _Nullable _CFXMLDTDNodeCopyPublicID(_CFXMLDTDNodePtr node);
 void _CFXMLDTDNodeSetPublicID(_CFXMLDTDNodePtr node, const unsigned char* _Nullable publicID);
 
+// Namespaces
+CF_RETURNS_RETAINED _CFXMLNodePtr _Nonnull * _Nullable _CFXMLNamespaces(_CFXMLNodePtr node, CFIndex* count);
```
phausler: this is not retained, this is just returning allocated

CoreFoundation/Parsing.subproj/CFXMLInterface.c
rothomp3: I actually tried that first, because it seemed like the right way to do it, but it crashed immediately on adding a _CFXMLNodePtr to the array. I wasn't able to figure out why, exactly, but I also didn't spend a long time trying.

CoreFoundation/Parsing.subproj/CFXMLInterface.h
rothomp3: This is a relic of how I started off using CFArrayRef here. I would prefer to go back to that, I will see if the crash adding _CFXMLNodePtr still happens now.

phausler: You would need to use custom retain and release methods on the structure pointer to use CFArrayRef but for managing the buffer it is probably safer I would say.

phausler: eh, after thinking about it for a bit, perhaps the malloc'd buffer is better here because creating proper callbacks would mean that you would need to copy the backing pointers appropriately for the beyond the lifespan of the tree. tbh it might not be worth it if this is the only place that it is used.

rothomp3: This is the only place it's used at the moment, and honestly anyone else should just call the Swift method that calls this one, so it should remain the only place.

parkera:
@swift-ci test and merge

PR #1057, opened 2017-06-22
URLProtocol doesn't deliver response to URLSession delegates

nethraravindran:
URLProtocol should delegate the received response to URLSession delegate methods.

### DIFF
--- a/TestFoundation/TestNSTimer.swift
+++ b/TestFoundation/TestNSTimer.swift
@@ -59,7 +59,7 @@ class TestNSTimer : XCTestCase {
             XCTAssertEqual(timer.timeInterval, interval)
 
             let currentInterval = Date().timeIntervalSince1970
-            XCTAssertEqual(currentInterval, previousInterval + interval, accuracy: 0.01)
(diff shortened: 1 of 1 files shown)


Foundation/NSURLProtocol.swift
```diff
@@ -148,7 +148,26 @@ public protocol URLProtocolClient : NSObjectProtocol {
...
 
     func urlProtocol(_ protocol: URLProtocol, didReceive response: URLResponse, cacheStoragePolicy policy: URLCache.StoragePolicy) {
-        `protocol`.task?.response = response
+        guard let task = `protocol`.task else { fatalError("Received response, but there's no task.") }
+        task.response = response
+        let session = task.session as! URLSession
+        guard let dt = task as? URLSessionDataTask else { return }
+        switch session.behaviour(for: task) {
+        case .taskDelegate(let delegate as URLSessionDataDelegate):
+            //TODO: There's a problem with libcurl / with how we're using it.
```
pushkarnk: Looks like we are reusing code from http. Are these comments relevant here?

PR #1059, opened 2017-06-22
Added implementation of NSString init(contentsOfFile:usedEncoding:)

andybest:
I have added an implementation for init(contentsOfFile:usedEncoding:)

After discussion on the mailing list, it was determined that the original implementation:

1. Checks extended file attributes for "com.apple.TextEncoding", and uses the encoding there.
2. If that doesn't exist, it checks the first few bytes of the data for a Byte Order Mark (BOM).
3. If there is no BOM present, just try to decode as UTF8.

I also noticed that the init(contentsOf:usedEncoding:) initializer only looked for UTF16 BOMs, so I have updated the implementation for that too.

I have added some tests for the BOM detection, and updated the status document accordingly.


Foundation/NSString.swift
```diff
@@ -1297,38 +1297,137 @@ extension NSString {
...
+                            "NSDebugDescription" : "The file \"\(path)\" couldn't be opened because the text encoding of its contents can't be determined.",
+                            "NSFilePath": path
+                ])
+        }
+        
+        var str: String?
+        if String._conditionallyBridgeFromObjectiveC(cf._nsObject, result: &str) {
+            self.init(str!)
+        } else {
+            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.fileReadUnknownStringEncoding.rawValue, userInfo: [
```
parkera: This seems more like a fatal error than an NSError here.

parkera:
@swift-ci please test

parkera:
Looks pretty good, thanks for taking this one.

Foundation/NSString.swift
andybest: Which one, the bridging? I was going by other code that handled the same thing- I can convert to a fatal error if that's preferred.

pushkarnk:
```
Foundation/NSString.swift:1396:24: error: use of unresolved identifier 'getxattr'
        let bufCount = getxattr(path, attrName, nil, 0, 0, 0)
                       ^~~~~~~~
```

andybest:
I forgot to add `include GLibc` to my commit. I'll add it and fix the merge conflicts

andybest:
Ok, looks like the module map for Glibc doesn't include sys/xattr.h, so I've enabled the xattr check only for MacOS for now. I will add the include to the modulemap in the Swift project and do a PR over there to get it included, then I will make a separate PR here to re-enable this functionality under Linux.

I believe this PR should be ready to go here, in any case :)

ianpartridge:
Hi @andybest did you open a PR against the Glibc modulemap? Just wondering where we're up to with this PR 😃

alblue:
@swift-ci please test

andybest:
@ianpartridge Apologies, I've been away and haven't had the chance- I will sort this out today :)

andybest:
By the way, this PR can be merged as-is, since there is an '#if os(' block around the missing functionality which can removed later

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

alblue:
There were test failures on this run:

```
Test Case 'TestNSString.test_FromContentsOfFileUsedEncodingUTF32LEBOM' started at 2017-07-28 22:48:00.244
TestFoundation/TestNSString.swift:381: error: TestNSString.test_FromContentsOfFileUsedEncodingUTF32LEBOM : failed - URL for NSString-UTF32-LE-data.txt is nil
Test Case 'TestNSString.test_FromContentsOfFileUsedEncodingUTF32LEBOM' failed (0.0 seconds)
Test Case 'TestNSString.test_FromContentsOfFileUsedEncodingUTF32BEBOM' started at 2017-07-28 22:48:00.244
TestFoundation/TestNSString.swift:400: error: TestNSString.test_FromContentsOfFileUsedEncodingUTF32BEBOM : failed - URL for NSString-UTF32-BE-data.txt is nil
```

Did those files get added correctly?

spevans:
Do the tests in the PR pass in Xcode? I noticed the tests for the UTF16 LE/BE were failing the other day and I was wondering if this fixes them as well?

alblue:
This pull request has gone stale, closing. Please rebase and reopen if desired. See also #928

PR #1060, opened 2017-06-22
Updated NotificationCenter API to reflect current Apple's syntax

ygini:
Current documented NotificationCenter's syntax does not match the Open Source version.

To enforce code compatibility between Linux and Mac, this PR fix the API and keep the old syntax to avoid compatibility issue with existing code.


parkera:
In general we haven't been worrying much about source compatibility with previous Linux versions in these cases - it's better to match Darwin. We could mark the old version with the right availability annotation (unavailable and replaced with the new one in swift 4) instead.

ygini:
@parkera what's the availability annotation syntax? I could update the PR if you want.

parkera:
There are some examples in the Swift standard library; look for "obsoleted"

ianpartridge:
I've merged https://github.com/apple/swift-corelibs-foundation/pull/1172 which fixes this. Thanks for opening your PR.

PR #1061, opened 2017-06-22
Validate indexes and ranges passed into Data so that bounding conditions are respected

phausler:


phausler:
@swift-ci please test

PR #1062, opened 2017-06-23
Refactor URLSession tests to only run with a single instance server

ddunn2:
Previously any test that needed a server would create it's own. While not an issue at the moment it's clear this wouldn't scale very well. This change creates a single server at the start of the tests, and then all the tests use this server. 
This PR addresses [SR-4676](https://bugs.swift.org/browse/SR-4676).

I've run the tests on both Mac and Linux to confirm they all pass.


pushkarnk:
Thanks for getting this done!

PR #1063, opened 2017-06-23
Fix for URLSession redirect does not inherit request timeout value[SR…

saiHemak:
…-2682]

### DIFF
--- a/Foundation/NSURLSession/http/EasyHandle.swift
+++ b/Foundation/NSURLSession/http/EasyHandle.swift
@@ -240,9 +240,16 @@ extension _EasyHandle {
         try! CFURLSession_easy_setopt_int64(rawHandle, CFURLSessionOptionINFILESIZE_LARGE, length).asError()
     }
 
-   func set(timeout value: Int) {
+    func set(timeout value: Int) {
        try! CFURLSession_easy_setopt_long(rawHandle, CFURLSessionOptionTIMEOUT, value).asError()
-   }
+    }
+
+    func getTimeoutIntervalSpent() -> Double {
+        var timeSpent = Double()
+        CFURLSession_easy_getinfo_double(rawHandle, CFURLSessionInfoTOTAL_TIME, &timeSpent)
+        return timeSpent / 1000
+    }
+
 }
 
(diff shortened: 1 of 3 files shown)


bubski:
@saiHemak does this maybe fix problems with `TestURLSession.test_httpRedirection`?

pushkarnk:
@saiHemak #1062 changed TestURLSession to using a single server. We'd need to change the test here accordingly.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -240,9 +240,18 @@ extension _EasyHandle {
         try! CFURLSession_easy_setopt_int64(rawHandle, CFURLSessionOptionINFILESIZE_LARGE, length).asError()
     }
 
-   func set(timeout value: Int) {
+    func set(timeout value: Int) {
        try! CFURLSession_easy_setopt_long(rawHandle, CFURLSessionOptionTIMEOUT, value).asError()
-   }
+    }
+
+    func getTimetaken() -> Double {
```
xwu: Spelling: should be `getTimeTaken()`.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -240,9 +240,18 @@ extension _EasyHandle {
...
+        let timeTaken = UnsafeMutablePointer<Double>.allocate(capacity: 8)
```
xwu: This is not correct. You are allocating 8 instances of type `Double`. The body of this function should be:

```swift
var timeSpent = Double()
CFURLSession_easy_getinfo_double(rawHandle, CFURLSessionInfoTOTAL_TIME,  &timeSpent)
return timeSpent
```

Since this is used only for subtracting from the timeout interval, it might be worthwhile to return `timeTaken / 1000` and to rename this function `getTimeoutIntervalSpent()`.

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci please test

saiHemak:
@ bubski  I could see the test case``` TestURLSession.test_httpRedirection``` is passing successfully  even without this change. .Can you please elaborate about the issue you are facing  ?
@pushkarnk ,@xwu : Thanks for the review.I have addressed your review comments.

bubski:
@saiHemak I pulled newest changes and the problem went away.

pushkarnk:
Thanks @saiHemak 

@swift-ci test and merge

pushkarnk:
@swift-ci test and merge

PR #1064, opened 2017-06-23
Import NSRange from the Darwin overlay

phausler:
This has two factors that are updated;
NSRange.swift had a number of changes to handle String range conversions in the Darwin overlay. This commit imports the latest changes to NSRange in but also makes it mergeable back to the Darwin overlay so that we can keep the same source between the two repositories.

This change requires NSScanner for key functions to be updated to be in parity with their Darwin counterparts. Note: not all of NSScanner was updated (and seems to have more work to be done to make the naming updates be in sync)


phausler:
@swift-ci please test

PR #1065, opened 2017-06-23
NotificationCenter corrections for observers

phausler:
NotificationCenter should not reference the object and the observer returned is the opaque token for storage


phausler:
@swift-ci please test

PR #1066, opened 2017-06-23
Implement concurrent enumeration for NSArray, NSDictionary and NSIndexSet

phausler:
This implements concurrent enumeration options so that they no longer claim un-implemented.
Ideally we should use a more efficient locking mechanism like stdatomic.h style but the best we have in swift is NSLock/allocated mutex


phausler:
@swift-ci please test

PR #1067, opened 2017-06-23
Update IndexPath to be merged in from the Darwin overlay and bring across unit tests from the Darwin overlay for IndexPath

phausler:


phausler:
@swift-ci please test

PR #1068, opened 2017-06-23
[SR-90] Fixed NSNumber with large unsigned integers

bubski:
This is to fix https://bugs.swift.org/browse/SR-90?jql=text%20~%20%22unsigned%22

## unsigned CFNumber
**I'm not sure if my reasoning is correct**, but as far as I understand, `CFNumber` operates only on signed integers, which caused problems for large unsigned values.  
The solution here is to promote big unsigned values to wider signed types.

## boolValue
There was also an issue with `NSNumber(value: Int64.min).boolValue` returning `true` while on Darwin it returns `false`.
This seems to be a bug on Darwin, but I added this exception for consistency.

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -461,9 +461,12 @@ open class NSNumber : NSValue {
     }
     
     open var boolValue: Bool {
-        return int64Value != 0
+        // Darwin Foundation NSNumber appears to have a bug and return false for NSNumber(value: Int64.min).boolValue,
+        // even though the documentation says:
+        // "A 0 value always means false, and any nonzero value is interpreted as true."
+        return (int64Value != 0) && (int64Value != Int64.min)
     }
-    
+
     open var intValue: Int {
         var val: Int = 0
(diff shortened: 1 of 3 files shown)


bubski:
@swift-ci please test

Foundation/NSNumber.swift
```diff
@@ -461,9 +461,12 @@ open class NSNumber : NSValue {
     }
     
     open var boolValue: Bool {
-        return int64Value != 0
+        // Darwin Foundation NSNumber appears to have a bug and return false for NSNumber(value: Int64.min).boolValue,
+        // even though the documentation says:
+        // "A 0 value always means false, and any nonzero value is interpreted as true."
+        return (int64Value != 0) && (int64Value != Int64.min)
```
phausler: yea... I've tried to fix this a number of times... sadly it is a 32 bit truncation that occurs. Unfortunately there are a number of apps that rely on that behavior. The actual code here is effectively `int32Value != 0` but lets keep the way it should be and use the previous value

phausler:

PR #1069, opened 2017-06-23
Import IndexSet from the overlay, add a number of unit tests and bring NSIndexSet a bit closer to pairty to the Darwin counterpart

phausler:


phausler:
@swift-ci please test

PR #1070, opened 2017-06-24
Remove temporary workaround for SR-3202

skyline75489:
This succeeds #1043 which was merged and later reverted. This PR has more love for Linux.

Turns out on Linux `Swift.Bool` can not be converted to `Foundation.NSNumber`. On Darwin the conversion works fine. That's why the tests failed on Linux.

For the test cases, `en_UK` isn't really a valid locale identifier. For `en_UK` libicu actually returns a fallback value for measurement system(`UMS_SI`) instead of `UMS_UK`. To prevent further confusion I'd say `en_GB` is a better option.

### DIFF
--- a/Foundation/LengthFormatter.swift
+++ b/Foundation/LengthFormatter.swift
@@ -69,7 +69,7 @@ open class LengthFormatter : Formatter {
         //Extract the number from the measurement
         let numberInUnit = unitMeasurement.value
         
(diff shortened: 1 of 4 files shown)


skyline75489:
CC  @jszumski  @alblue  @saiHemak

skyline75489:
Also, is it expected that `Swift.Bool` can not be converted to `Foundation.NSNumber` on Linux? I think it seems to make sense for this conversion to work as on Darwin.

alblue:
@swift-ci please test

PR #1071, opened 2017-06-24
Fix warnings about possibly wrong inferred type in NSURLSession

skyline75489:


pushkarnk:
Thanks!

PR #1072, opened 2017-06-24
Add a localized Info.plist for testing

skyline75489:


parkera:
localizedInfoPlist should return non-nil even if there were no `lproj` directories in the project, so I'm a bit unclear on how this fixes anything.

skyline75489:
Oh that's the expected behavior? I thought localizedInfoPlist *should* return nil if there were no `lproj` in the project as the test case is writtent that way.

skyline75489:
I'll close this for now.

PR #1077, opened 2017-06-27
Split up a couple expressions to work around slow type checking.

rudkx:
The type checker is particularly slow on these expressions. I've opened
`https://bugs.swift.org/browse/SR-5322` to investigate the type checker
issue, but in the meantime it would be nice to speed up builds. This
cuts more than a minute from my unoptimized asserts-enabled build.

### DIFF
--- a/Foundation/NSCFString.swift
+++ b/Foundation/NSCFString.swift
@@ -59,13 +59,19 @@ internal class _NSCFString : NSMutableString {
 
 internal final class _NSCFConstantString : _NSCFString {
     internal var _ptr : UnsafePointer<UInt8> {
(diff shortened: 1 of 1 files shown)


rudkx:
@swift-ci Please smoke test

rudkx:
@parkera Is this okay to merge?

parkera:
Can we put a SR or radar number in the comment so we know when it's ok to remove this hack?

rudkx:
Sure - I've added a commit that mentions the SR number, and updated the SR to point back to this PR so that we know to revert it once the type checker performance issue is resolved.

rudkx:
@swift-ci Please smoke test

phausler:

PR #1081, opened 2017-06-27
Avoid cases that inflame tuple destructuring

phausler:
This is perhaps a more clear approach for a compatible avoidance of tuple destructuring so that we can hopefully build no matter the status of SE-110

### DIFF
--- a/Foundation/NSJSONSerialization.swift
+++ b/Foundation/NSJSONSerialization.swift
@@ -684,13 +684,22 @@ private struct JSONReader {
         }
         return index
     }
-
-    func takeMatching(_ match: @escaping (UInt8) -> Bool) -> ([Character], Index) -> ([Character], Index)? {
-        return { input, index in
-            guard let (byte, index) = self.source.takeASCII(index), match(byte) else {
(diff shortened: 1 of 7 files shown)


Foundation/NSJSONSerialization.swift
```diff
@@ -785,11 +794,11 @@ private struct JSONReader {
     }
     func parseCodeUnit(_ input: Index) -> (UTF16.CodeUnit, Index)? {
         let hexParser = takeMatching(isHexChr)
-        guard let (result, index) = hexParser([], input).flatMap(hexParser).flatMap(hexParser).flatMap(hexParser),
-            let value = Int(String(result), radix: 16) else {
+        guard let result = hexParser(CharactersAtIndex([], input)).flatMap(hexParser).flatMap(hexParser).flatMap(hexParser),
```
parkera: This one I'm honestly not sure is any clearer than the previous approach.

phausler: not by much, that code is a bit hard to read as it is

PR #1083, opened 2017-06-27
Relax the accuracy for timer validation to account for heavier CI server loads

phausler:


phausler:

PR #1087, opened 2017-06-28
Optionality updates to Codable API

itaiferber:
The swift-corelibs-foundation side of apple/swift/pull/10667.


itaiferber:
This needs to be tested in conjunction with apple/swift#10667

itaiferber:
Looks like this is good to merge simultaneously with apple/swift#10667. Waiting for review on that and I'll merge both.

PR #1088, opened 2017-06-29
Uncomment NumberFormatter roundingBehavior

skyline75489:
I think it's safe to uncomment this shiny new Swift 4 API now. Tested on macOS & Ubuntu.


alblue:
@swift-ci please test

PR #1092, opened 2017-06-30
Fixed infinite recursion in NSCharacterSet.isEqual

bubski:
`CFEqual` for `CFCharacterSet` would call `NSCharacteSet.isEqual` which would again call `CFEqual` and so on.

### DIFF
--- a/Foundation/NSCharacterSet.swift
+++ b/Foundation/NSCharacterSet.swift
@@ -50,13 +50,18 @@ open class NSCharacterSet : NSObject, NSCopying, NSMutableCopying, NSCoding {
     }
     
     open override func isEqual(_ value: Any?) -> Bool {
+        guard let runtimeClass = _CFRuntimeGetClassWithTypeID(CFCharacterSetGetTypeID()) else {
+            fatalError("Could not obtain CFRuntimeClass of CFCharacterSet")
+        }
+
+        guard let equalFunction = runtimeClass.pointee.equal else {
+            fatalError("Could not obtain equal function from CFRuntimeClass of CFCharacterSet")
+        }
+
         switch value {
-        case let other as CharacterSet:
-            return CFEqual(_cfObject, other._cfObject)
-        case let other as NSCharacterSet:
-            return CFEqual(_cfObject, other._cfObject)
-        default:
-            return false
+        case let other as CharacterSet: return equalFunction(self._cfObject, other._cfObject) == true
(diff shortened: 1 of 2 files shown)


bubski:
@swift-ci please test

Foundation/NSCharacterSet.swift
```diff
@@ -50,13 +50,18 @@ open class NSCharacterSet : NSObject, NSCopying, NSMutableCopying, NSCoding {
     }
     
     open override func isEqual(_ value: Any?) -> Bool {
+        guard let runtimeClass = _CFRuntimeGetClassWithTypeID(CFCharacterSetGetTypeID()) else {
```
phausler: this is rather sub-optimal of a way to approach the type testing

bubski: @phausler Could you please explain a bit more? Isn't this what `CFEqual` does under the hood?

I'm open for suggestions.

phausler: Yea, but the problem is that it is reaching into CF internals pretty heavily - I have a patch in the works that will resolve this (and passes your unit test) without needing to reach into CF guts so much. This is going the opposite direction of the factory version of NSCharacterSet (which generalizes this method without needing to reach into CF internals)

bubski: @phausler I assumed it was good enough because I didn't need to expose anything from CF.
I also noticed you were working around this area and expected your changed would address this as well, but was not sure if and when.

When do you expect to have those in?
And do you suggest dropping this PR?

bubski: @phausler oh, and by the way, I'm not sure what to think of:
https://github.com/apple/swift-corelibs-foundation/pull/1092/files#diff-c6672fda2d6b50ba301f8b945cacbd8aR298
Any thoughts?

phausler: In this branch I am working in I have those tests enabled as well (and they are passing). I have a few more things to wrap up before pushing that as an addendum to the factory_generalizations. (it isn't ready for pushing yet since I am working on getting a few more goodies published as well)

phausler:
hmm it looks like my pr is going to take a bit to get all the new bits approved - so lets go ahead and merge this for now since it is blocking testing of Codable to be ported.

My new stuff will refactor the implementation but I will make certain your tests are included.

phausler:

PR #1093, opened 2017-07-02
Use `CFNumberCreate()` as factory initializer on `NSNumber`

norio-nomura:
This replaces #1025 by further utilizing the Factory Initializer Method ([SR-5255](https://bugs.swift.org/browse/SR-5255)).

---
### Use `CFNumberCreate()` as factory initializer on `NSNumber.init(value:)`
- Change `NSNumber.init(value:)` to returning instance created by `CFNumberCreate()`
- Remove `NSNumber._objCType`

### Fixes [SR-4907](https://bugs.swift.org/browse/SR-4907)
- Add `TestNSNumber.test_objCType()` reproducing crash
  Expected results are sampled from macOS
- Fix crash on accessing `objCType` property of `NSNumber`  

### Change `NSNumber.objCType` to returning same values with darwin platform version
- Change `NSNumber.objCType` to returning same values with darwin
- Change `NSNumber.init(value: Bool)` to returning `__NSCFBoolean` by using `_Factory` protocol

Since `NSNumber.objCType` returns `c` with both of `Bool` and `Int8` on darwin platform, `NSNumber.init(value: Bool)` is changed to returns `__NSCFBoolean` by using `_Factory` protocol, and it changed to use `_cfTypeID` to distinguish between Bool and Int8.

- Change `JSONSerialization` to use `_cfTypeID` for checking whether `NSNumber` is Boolean or not
- Change `_JSONDecoder` to use `_cfTypeID` for checking whether `NSNumber` is Boolean or not
- Change `NSNumber.compare(_:)` to use `objCType`
- Add missing entry to `TestNumber.allTests`
- Fix `NSKeyedUnarchiver.decodeBool(forKey:)` and add test to `TestNSPredicate.test_NSCoding()`

### DIFF
--- a/Foundation/NSCFBoolean.swift
+++ b/Foundation/NSCFBoolean.swift
@@ -78,7 +78,10 @@ internal class __NSCFBoolean : NSNumber {
     override var objCType: UnsafePointer<Int8> {
         // This must never be fixed to be "B", although that would
         // cause correct old-style archiving when this is unarchived.
-        return UnsafePointer<Int8>(bitPattern: UInt(_NSSimpleObjCType.Char.rawValue.value))!
+        func _objCType(_ staticString: StaticString) -> UnsafePointer<Int8> {
+            return UnsafeRawPointer(staticString.utf8Start).assumingMemoryBound(to: Int8.self)
+        }
+        return _objCType("c")
     }
     
     internal override func _cfNumberType() -> CFNumberType  {
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -205,9 +205,6 @@ private struct CFSInt128Struct {
 
 open class NSNumber : NSValue {
     typealias CFType = CFNumber
-    // This layout MUST be the same as CFNumber so that they are bridgeable
-    private var _base = _CFInfo(typeID: CFNumberGetTypeID())
-    private var _pad: UInt64 = 0
 
     internal var _cfObject: CFType {
         return unsafeBitCast(self, to: CFType.self)
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -309,12 +309,6 @@ open class NSNumber : NSValue {
         _CFNumberInitDouble(_cfObject, value)
     }
     
-    public init(value: Bool) {
-        _objCType = .Bool
-        super.init()
-        _CFNumberInitBool(_cfObject, value)
-    }
-
     override internal init() {
         _objCType = .Undef
         super.init()
@@ -614,6 +608,12 @@ open class NSNumber : NSValue {
     }
     
     open override var classForCoder: AnyClass { return NSNumber.self }
+}
+
+extension NSNumber {
+    public convenience init(value: Bool) {
+        self.init(factory: value._bridgeToObjectiveC)
+    }
 }
 
 extension CFNumber : _NSBridgeable {
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -247,26 +247,6 @@ open class NSNumber : NSValue {
             return _objCType("f")
         case kCFNumberFloat64Type:
             return _objCType("d")
-        case kCFNumberCharType:
-            return _objCType("c")
-        case kCFNumberShortType:
-            return _objCType("s")
-        case kCFNumberIntType:
-            return _objCType("i")
-        case kCFNumberLongType:
-            return _objCType("q")
-        case kCFNumberLongLongType:
-            return _objCType("q")
-        case kCFNumberFloatType:
-            return _objCType("f")
-        case kCFNumberDoubleType:
-            return _objCType("d")
-        case kCFNumberCFIndexType:
-            return _objCType("q")
-        case kCFNumberNSIntegerType:
-            return _objCType("q")
-        case kCFNumberCGFloatType:
-            return _objCType("d")
         case kCFNumberSInt128Type:
             return _objCType("Q")
         default:
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -205,6 +205,9 @@ private struct CFSInt128Struct {
 
 open class NSNumber : NSValue {
     typealias CFType = CFNumber
+    // This layout MUST be the same as CFNumber so that they are bridgeable
+    private var _base = _CFInfo(typeID: CFNumberGetTypeID())
+    private var _pad: UInt64 = 0
 
     internal var _cfObject: CFType {
         return unsafeBitCast(self, to: CFType.self)
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -198,13 +198,19 @@ extension NSNumber : ExpressibleByFloatLiteral, ExpressibleByIntegerLiteral, Exp
 
 }
 
+@_silgen_name("_CFNumberGetType2")
+func _CFNumberGetType2(_ number: CFNumber!) -> CFNumberType
+
+private struct CFSInt128Struct {
+    var high: Int64
+    var low: UInt64
+}
+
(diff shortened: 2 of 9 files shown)


Foundation/NSNumber.swift
```diff
@@ -198,14 +198,20 @@ extension NSNumber : ExpressibleByFloatLiteral, ExpressibleByIntegerLiteral, Exp
 
 }
 
+@_silgen_name("_CFNumberGetType2")
```
phausler: We really should not use silgen_name. Instead can you just put this in ForSwiftFoundationOnly.h?

Foundation/NSNumber.swift
```diff
@@ -198,14 +198,20 @@ extension NSNumber : ExpressibleByFloatLiteral, ExpressibleByIntegerLiteral, Exp
...
+
+private struct CFSInt128Struct {
+    var high: Int64
+    var low: UInt64
+}
+
 open class NSNumber : NSValue {
     typealias CFType = CFNumber
     // This layout MUST be the same as CFNumber so that they are bridgeable
     private var _base = _CFInfo(typeID: CFNumberGetTypeID())
```
phausler: If we are using factories for all numbers then we can delete this!

Foundation/NSCFBoolean.swift
```diff
@@ -78,7 +78,10 @@ internal class __NSCFBoolean : NSNumber {
     override var objCType: UnsafePointer<Int8> {
         // This must never be fixed to be "B", although that would
         // cause correct old-style archiving when this is unarchived.
-        return UnsafePointer<Int8>(bitPattern: UInt(_NSSimpleObjCType.Char.rawValue.value))!
+        func _objCType(_ staticString: StaticString) -> UnsafePointer<Int8> {
```
phausler: I like this trick better than the way I was thinking, however I wonder if we have a guarantee this is safe?

Foundation/NSNumber.swift
```diff
@@ -230,93 +236,143 @@ open class NSNumber : NSValue {
     }
 
     open override var objCType: UnsafePointer<Int8> {
-        return UnsafePointer<Int8>(bitPattern: UInt(_objCType.rawValue.value))!
+        func _objCType(_ staticString: StaticString) -> UnsafePointer<Int8> {
+            return UnsafeRawPointer(staticString.utf8Start).assumingMemoryBound(to: Int8.self)
+        }
+        let numberType = _CFNumberGetType2(_cfObject)
+        switch numberType {
```
phausler: Iirc gettype2 will only emit SInt* and Float* types

phausler:
The sil_gen part is the only absolute thing that is required

Foundation/NSCFBoolean.swift
xwu: Wouldn't the static string need to be a static property to guarantee its lifetime?

Foundation/NSNumber.swift
norio-nomura: What's the difference between `ForSwiftFoundationOnly.h` and `ForFoundationOnly.h`?
It seems `_CFNumberInit*()` functions are placed in `ForFoundationOnly.h`.
Should `_CFNumberGetType2` also be placed in `ForFoundationOnly.h`?

norio-nomura:
Can we remove `_CFNumberInit*()` functions?

Foundation/NSNumber.swift
norio-nomura: I didn't know. It looks `__CFNumberTypeTable.canonicalType` contains only those types. 👍

Foundation/NSNumber.swift
phausler: ForFoundationOnly.h is used in both the CoreFoundation shipped on Darwin as well as swift-corlibs-foundation, whereas ForSwiftFoundationOnly.h is used for swift-corelibs-foundation exclusively.

Putting it in ForSwitFoundationOnly.h means that I can merge the closed source version and the open source version easier ;)

Foundation/NSCFBoolean.swift
phausler: I am not fully certain which way this falls, constant strings should emit static references in the binary so it is probably safe-ish? 

We have a few alternatives; we could thunk out to C and use the real `@encode` since we always have clang available. We could thunk out to C and reference C strings as opaque memory blobs. This approach. Or the other thought I had was something like this https://gist.github.com/phausler/2630092f1b391d6a5db0a562288514bf

phausler:
lets hold off on removing _CFNumberInit just yet, I need to verify we are not using that in the objc version (which I was considering)

Foundation/NSCFBoolean.swift
xwu: I was thinking along the same lines, but simpler:

```swift
fileprivate let boolEncoding: StaticString = "B"
fileprivate let boolEncodingPtr: UnsafePointer<Int8> = {
  precondition(boolEncoding.isASCII)
  return UnsafeRawPointer(boolEncoding.utf8Start).assumingMemoryBound(to: Int8.self)
}()

// etc.
```

phausler: yea, I think there is also value in abstracting it out to an isolated file to do that work in and provide a shared internal (minimal) version of `@encode`

norio-nomura: I discarded following codes that rewrote `_NSSimpleObjCType` in `NSObjCRuntime.swift`, since memory layout is not compatible with current implementation.
https://gist.github.com/norio-nomura/a1383f258c38bb1e081597a538f46e8f

Should we create another static values for `@encode`?
I feel we should do re-writing `_NSSimpleObjCType` instead of adding them, and doing that with this PR will make it too large.

norio-nomura: As I tested, `utf8Start` points directly cstring in `__TEXT,__cstring,cstring_literals` whether `StaticString` is instantiated as static variable or function parameter.

norio-nomura:
Ah, sorry, it seems I have to fix `NSDecimalNumber` as well.
It will take some time yet.

Foundation/NSNumber.swift
norio-nomura: These properties are necessary to maintain memory layout compatibility of `NSDecimalNumber`.

norio-nomura:
> Ah, sorry, it seems I have to fix `NSDecimalNumber` as well.

Removing stored instance properties caused issues on `NSDecimalNumber`. I reverted that.

phausler:
I can follow up with the removals of _CFNumberInit* and the ivars.

phausler:

norio-nomura:
Thanks! 🙏

norio-nomura:
I forgot that I have to ask a question.

The results of [test](https://gist.github.com/norio-nomura/fe151549addaed10a9a365b183d0f9ed) are different between `arch(arm64)` and `arch(arm)`.
On `arch(arm64)` (iPhone SE w/ iOS 10.3.3):
```
sInt8Type: c
sInt16Type: s
sInt32Type: i
sInt64Type: q
float32Type: f
float64Type: d
charType: c
shortType: s
intType: i
longType: q
longLongType: q
floatType: f
doubleType: d
cfIndexType: q
nsIntegerType: q
cgFloatType: d
sInt128Type: Q
```
On `arch(arm)` (iPad 3rd w/ iOS 9.3.5):
```
sInt8Type: i
sInt16Type: i
sInt32Type: i
sInt64Type: q
float32Type: f
float64Type: d
charType: i
shortType: i
intType: i
longType: i
longLongType: q
floatType: f
doubleType: d
cfIndexType: i
nsIntegerType: i
cgFloatType: f
sInt128Type: Q
```

Is not this a problem?

phausler:
For now that is probably fine, my changes will hopefully bring our strange conditionals per 64 vs 32 bit in more pairity

PR #1094, opened 2017-07-02
OperationQueue.main dangling pointer fix

bubski:
Calling `OperationQueue.main` would
instantiate an `OperationQueue`,
set an `Unmanaged` reference to it as **main queue specific**
and return `OperationQueue`.

Subsequent calls to `OperationQueue.main` would
get the `Unmanaged` reference to `OperationQueue` from **main queue specific**
 and return `OperationQueue` wrapped in `Unmanaged`.

However, there's nothing stopping the `OperationQueue` from being deallocated, so the `Unmanaged` reference from subsequent call would in fact return garbage.

### DIFF
--- a/Foundation/Operation.swift
+++ b/Foundation/Operation.swift
@@ -577,12 +577,13 @@ open class OperationQueue: NSObject {
 #endif
     }
     
+#if DEPLOYMENT_ENABLE_LIBDISPATCH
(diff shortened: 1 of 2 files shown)


bubski:
@swift-ci please test

parkera:
@swift-ci test and merge

PR #1097, opened 2017-07-06
fix URLSession crashing on HTTP/0.9 simple-responses

weissi:
also add tests for misbehaving HTTP servers.

we discovered that if the server is just returning random bytes, the whole app would crash with `fatalError("Received body data, but the header is not complete, yet.")` . The reason we can get into that state is that CURL support HTTP/0.9 [simple-responses](https://www.w3.org/Protocols/HTTP/1.0/spec.html#Message-Types) which are just the body bytes...

The issue is [also discussed on CURL's github](https://github.com/curl/curl/issues/467).

To make sure we don't regress and to document the current behaviour, I also wrote tests for a few other cases when the HTTP server is misbehaving or behaving oddly.

### DIFF
--- a/Foundation/NSURLSession/http/HTTPURLProtocol.swift
+++ b/Foundation/NSURLSession/http/HTTPURLProtocol.swift
@@ -439,8 +439,14 @@ extension _HTTPURLProtocol {
 extension _HTTPURLProtocol: _EasyHandleDelegate {
     
     func didReceive(data: Data) -> _EasyHandle._Action {
-        guard case .transferInProgress(let ts) = internalState else { fatalError("Received body data, but no transfer in progress.") }
-        guard ts.isHeaderComplete else { fatalError("Received body data, but the header is not complete, yet.") }
+        guard case .transferInProgress(var ts) = internalState else { fatalError("Received body data, but no transfer in progress.") }
+        if !ts.isHeaderComplete {
+            ts.response = HTTPURLResponse(url: ts.url, statusCode: 200, httpVersion: "HTTP/0.9", headerFields: [:])
+            /* we received body data before CURL tells us that the headers are complete, that happens for HTTP/0.9 simple responses, see
+               - https://www.w3.org/Protocols/HTTP/1.0/spec.html#Message-Types
+               - https://github.com/curl/curl/issues/467
+            */
+        }
         notifyDelegate(aboutReceivedData: data)
         internalState = .transferInProgress(ts.byAppending(bodyData: data))
         return .proceed
--- a/TestFoundation/TestNSURLSession.swift
+++ b/TestFoundation/TestNSURLSession.swift
@@ -41,6 +41,10 @@ class TestURLSession : XCTestCase {
 	    ("test_customProtocolResponseWithDelegate", test_customProtocolResponseWithDelegate),
             ("test_httpRedirection", test_httpRedirection),
             ("test_httpRedirectionTimeout", test_httpRedirectionTimeout),
+            ("test_http0_9SimpleResponses", test_http0_9SimpleResponses),
+            ("test_outOfRangeButCorrectlyFormattedHTTPCode", test_outOfRangeButCorrectlyFormattedHTTPCode),
+            ("test_missingContentLengthButStillABody", test_missingContentLengthButStillABody),
+            ("test_illegalHTTPServerResponses", test_illegalHTTPServerResponses),
         ]
     }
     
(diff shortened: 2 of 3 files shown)


weissi:
@swift-ci please test

TestFoundation/TestNSURLSession.swift
```diff
@@ -41,6 +41,10 @@ class TestURLSession : XCTestCase {
 	    ("test_customProtocolResponseWithDelegate", test_customProtocolResponseWithDelegate),
             ("test_httpRedirection", test_httpRedirection),
             ("test_httpRedirectionTimeout", test_httpRedirectionTimeout),
+            ("test_illegalHTTPServerResponses", test_illegalHTTPServerResponses),
```
alblue: If we're keeping the order of the tests the same as in the test list, this line should be moved to the end.

weissi: thanks @alblue , addressed

weissi:
@swift-ci please test

weissi:
@swift-ci please test

pushkarnk:
This is interesting. The change and the tests look good to me. 

Does Foundation on Darwin also gracefully handle such responses?

weissi:
@pushkarnk two intentions in this pull request

1. most importantly don't just crash the program on unexpected server response as that's terrible for the user
2. document the current behaviour in unit tests

I haven't tried how regular Foundation behaves, but it certainly won't just crash ;). Do we want to match regular Foundation exactly or what is the intention here? With my patch we handle these not so common server responses just the way CURL handles them which is what we seem to do usually in Swift Foundation's URLSession.

weissi:
@pushkarnk @parkera @phausler My suggestion would be to get this patch in as soon as possible to be sure it'll land in Swift 4 as it's just terrible if the program crashes if the server misbehaves. Worst is that if you connect to HTTP2-only servers you might hit that behaviour (as a HTTP2 only response looks like a HTTP/0.9 simple-response to CURL versions without HTTP2 support).

After we addressed the crasher issues we should talk about what we want to achieve here and how we want to handle all the special cases. See [here](https://github.com/apple/swift-corelibs-foundation/pull/1097/files#diff-e06c282e627962e05b9b7b7936ec63b6R186) for the list of odd server behaviours I could quickly come up with. We should then go through that list one by one and decide what the appropriate behaviour should be.

But IMHO more importantly we should get something in to fix the crashes. We can then always decide later that we want to throw away HTTP/0.9 simple responses even though CURL handles them.

pushkarnk:
Specifically, if the server is using HTTP/1.1 but the header is malformed (as you've demonstrated in the tests) does Darwin also return an HTTPURLResponse with httpVersion set to 0.9?

pushkarnk:
>> Do we want to match regular Foundation exactly or what is the intention here? 
Yes, we'd ideally want to. Unless we are sure that the Darwin Foundation isn't doing the right thing

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -439,8 +439,14 @@ extension _HTTPURLProtocol {
 extension _HTTPURLProtocol: _EasyHandleDelegate {
     
     func didReceive(data: Data) -> _EasyHandle._Action {
-        guard case .transferInProgress(let ts) = internalState else { fatalError("Received body data, but no transfer in progress.") }
-        guard ts.isHeaderComplete else { fatalError("Received body data, but the header is not complete, yet.") }
+        guard case .transferInProgress(var ts) = internalState else { fatalError("Received body data, but no transfer in progress.") }
+        if !ts.isHeaderComplete {
+            ts.response = HTTPURLResponse(url: ts.url, statusCode: 200, httpVersion: "0.9", headerFields: [:])
```
pushkarnk: Is this valid in the three different scenarios that you have pointed out?
1. When the server sends a HTTP/0.9 response aka "only the body"
2. When the server is running HTTP/1.1 but sends a malformed response
3. When the server running HTTP/2 sends a response to client HTTP/1.1 
... I'm specifically concerned about  the `httpVersion: "0.9"`

pushkarnk:
In short, I seem to have only two questions:
1. Is the behaviour consistent with Darwin? (Do we desire that? I'd say yes!)
2. Is the HTTPURLResponse being returned valid in all the failure scenarios pointed out?
Having said that, I agree that we need to fix these crashes first. 
@parkera @phausler What would you suggest?

weissi:
@pushkarnk so in Darwin Foundation, it's really a `NSURLHTTPResponse` that's created using a `CFURLHTTPResponse` that's created with a `CFHTTPMessageCreateResponse` and that is initialised with the following parameters:

```
* thread #2, queue = 'com.apple.CFNetwork.LoaderQ', stop reason = breakpoint 1.5
    frame #0: 0x00007fff2d1be705 CFNetwork`CFHTTPMessageCreateResponse
CFNetwork`CFHTTPMessageCreateResponse:
->  0x7fff2d1be705 <+0>: pushq  %rbp
    0x7fff2d1be706 <+1>: movq   %rsp, %rbp
    0x7fff2d1be709 <+4>: pushq  %r15
    0x7fff2d1be70b <+6>: pushq  %r14
Target 0: (URLSessionRepro) stopped.
(lldb) po $rdi
\<CFAllocator 0x7fff579b1f10 [0x7fff579b1f10]>{info = 0x7fff5def2000}

(lldb) po $rsi
200

(lldb) po $rdx
<nil>

(lldb) po $rcx
HTTP/0.9

(lldb) 
```

so it does indeed treat it as HTTP/0.9 too but the string seems to be `HTTP/0.9` not just `0.9`. I will change my patch accordingly.

weissi:
@pushkarnk also the `NSURLHTTPResponse` doesn't actually seem to have a property so it seems like you won't be able to query the HTTP version anyway but I guess we want it compatible.

weissi:
@swift-ci please test

pushkarnk:
@swift-ci please test

weissi:
@pushkarnk tests seemed to have passed. Any further comments/questions? @phausler happy with this?

phausler:
The change looks sensible to me

PR #1100, opened 2017-07-06
Change type of the file attributes dictionary key to match the Darwin version

weichsel:
`FileManager` already uses `FileAttributeKey` instead of `String` as key type for attribute dictionaries for all methods except `createFile`. I changed it there to make the method compatible with the Darwin version.
This is a small but API breaking change. How are such cases handled usually?

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -624,7 +624,7 @@ open class FileManager : NSObject {
         }
     }
     
-    open func createFile(atPath path: String, contents data: Data?, attributes attr: [String : Any]? = [:]) -> Bool {
+    open func createFile(atPath path: String, contents data: Data?, attributes attr: [FileAttributeKey : Any]? = nil) -> Bool {
         do {
             try (data ?? Data()).write(to: URL(fileURLWithPath: path), options: .atomic)
             return true


phausler:
We don't really have a migration strategy since there is no migrator on linux per-se... (other than the manual fix and replace).

The advantage of this version of Foundation being open source and not a system framework is that users of it can check-out to an old version and just use that for what its worth. The more important goal is to keep the swift-corelibs-foundation interface up to date with the latest and greatest version of Foundation imported from objective-c. So the change looks great to me. (as a side note it would be interesting to collect a set of project users and build a compatibility test suite for integration to other popular projects)

weichsel:
I recently released a [ZIP framework](https://github.com/weichsel/ZIPFoundation) that has a comprehensive test suite. It also works on Linux. 
While the framework itself already works fine on Linux, the test suite currently has some compile issues (including one that should be fixed as soon as this PR gets merged). 
My goal is to get ZIP Foundation to a) compile on Linux and b) to make the test suite run through. So ZIP Foundation might be a good candidate for Darwin Foundation/Swift Foundation compatibility tests.

parkera:
@shahmishal would be the person to talk to (I think) about adding a project to the source compatibility suite - although we don't run that against swift-corelibs-foundation yet, I think.

shahmishal:
We don’t currently support Linux, however we are planning on supporting it in the future.

weichsel:
I am currently using a custom Docker file to test Darwin/corelibs-Swift compatibility. (A similar approach as outlined [here](https://oleb.net/blog/2017/03/testing-swift-packages-on-linux/)).
An automated solution similar to the source compatibility test suite, but for Darwin/corelibs-Swift compatibility would be awesome.  
Let me know if I can contribute something.

PR #1104, opened 2017-07-09
NSNotification observing on specific queue implemented.

bubski:
`NSNotification` observing on specific queue implemented.


phausler:
@swift-ci please test

PR #1106, opened 2017-07-10
Several compilation warnings resolved

bubski:
Several compilation warnings resolved

### DIFF
--- a/TestFoundation/TestNSIndexPath.swift
+++ b/TestFoundation/TestNSIndexPath.swift
@@ -277,8 +277,10 @@ class TestNSIndexPath: XCTestCase {
         let ip2: IndexPath = [1, 1, 1]
         
         XCTAssertNotEqual(ip1.hashValue, ip2.hashValue)
-        
-        IndexPath(indexes: [Int.max >> 8, 2, Int.max >> 36]).hashValue // this should not cause an overflow crash
+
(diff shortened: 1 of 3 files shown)


TestFoundation/TestNSIndexPath.swift
```diff
@@ -277,8 +277,10 @@ class TestNSIndexPath: XCTestCase {
         let ip2: IndexPath = [1, 1, 1]
         
         XCTAssertNotEqual(ip1.hashValue, ip2.hashValue)
-        
-        IndexPath(indexes: [Int.max >> 8, 2, Int.max >> 36]).hashValue // this should not cause an overflow crash
+
+        // this should not cause an overflow crash
+        let hash: Int? = IndexPath(indexes: [Int.max >> 8, 2, Int.max >> 36]).hashValue
```
bubski: Is this reasonable? Or is guaranteed that compiler will not optimize out the below to a no-op?
```swift
_ = IndexPath(indexes: [Int.max >> 8, 2, Int.max >> 36]).hashValue
```

phausler: eh the assert below validates the non-crash case; and it is a test of an overflow so I think it should be fine as you have it

phausler:

PR #1107, opened 2017-07-10
NSPersonNameComponents.isEqual implemented

bubski:
`NSPersonNameComponents.isEqual` implemented


phausler:

PR #1108, opened 2017-07-11
Codable conformance to some Foundation types

bubski:
This PR mirrors _some_ changes apple/swift#9715

### Add conformances + unit tests for

- PersonNameComponents
- UUID
- URL
- NSRange
- Locale
- IndexSet
- IndexPath
- AffineTransform
- Decimal
- DateInterval
- CGFloat (unit tested indirectly)

### Not included in this PR, because of some problems that need to be solved

- Calendar
- CharacterSet
- DateComponents
- Measurement
- TimeZone

### DIFF
--- a/Foundation/CGFloat.swift
+++ b/Foundation/CGFloat.swift
@@ -962,9 +962,9 @@ extension CGFloat : Codable {
             // other type.
             do {
                 if NativeType.self == Float.self {
-                    self.native = NativeType(try container.decode(Float.self))
-                } else {
                     self.native = NativeType(try container.decode(Double.self))
+                } else {
+                    self.native = NativeType(try container.decode(Float.self))
                 }
             } catch {
                 // Failed to decode as the other type, too. This is neither a Float nor
--- a/Foundation/CGFloat.swift
+++ b/Foundation/CGFloat.swift
@@ -949,3 +949,35 @@ extension CGFloat : _CVarArgPassedAsDouble, _CVarArgAligned {
         return native._cVarArgAlignment
     }
 }
+
+extension CGFloat : Codable {
+    @_transparent
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        do {
+            self.native = try container.decode(NativeType.self)
+        } catch DecodingError.typeMismatch(let type, let context) {
+            // We may have encoded as a different type on a different platform. A
+            // strict fixed-format decoder may disallow a conversion, so let's try the
+            // other type.
+            do {
+                if NativeType.self == Float.self {
+                    self.native = NativeType(try container.decode(Float.self))
+                } else {
+                    self.native = NativeType(try container.decode(Double.self))
+                }
+            } catch {
+                // Failed to decode as the other type, too. This is neither a Float nor
+                // a Double. Throw the old error; we don't want to clobber the original
+                // info.
+                throw DecodingError.typeMismatch(type, context)
+            }
+        }
+    }
+    
+    @_transparent
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self.native)
+    }
+}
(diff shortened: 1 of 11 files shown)


Foundation/CGFloat.swift
```diff
@@ -949,3 +949,35 @@ extension CGFloat : _CVarArgPassedAsDouble, _CVarArgAligned {
...
+        let container = try decoder.singleValueContainer()
+        do {
+            self.native = try container.decode(NativeType.self)
+        } catch DecodingError.typeMismatch(let type, let context) {
+            // We may have encoded as a different type on a different platform. A
+            // strict fixed-format decoder may disallow a conversion, so let's try the
+            // other type.
+            do {
+                if NativeType.self == Float.self {
+                    self.native = NativeType(try container.decode(Float.self))
```
bubski: @itaiferber Could you double check this for me please?
over at https://github.com/apple/swift/blob/master/stdlib/public/SDK/CoreGraphics/CGFloat.swift.gyb#L823 it's
```swift

if NativeType.self == Float.self {
    self.native = NativeType(try container.decode(Double.self))
} else {
    self.native = NativeType(try container.decode(Float.self))
}
```
But shouldn't `Double.self` and `Float.self` be the other way around?

phausler: That is behaving correctly; it is in the fallback case where it tries to decode in the reversed order because the primary type failed.

phausler:
@swift-ci please test

Foundation/CGFloat.swift
itaiferber: Philippe is correct. This is a fallback case for encoders which differentiate between `Double`s and `Float`s in a way that makes them incompatible. On a platform where `CGFloat.NativeType == Double.self`, we will encode it as a `Double`. If we try to read it back on a platform where `CGFloat.NativeType == Float.self`, however, the decode could fail since the `Double` value would not fit in a `Float`. So this is that fallback case — if our native type is a `Float` but we failed to read it as a `Float`, try to read it as a `Double` since the decode may have failed.

The code is correct — you'll want to have the types the same way here as in the Foundation overlay.

bubski:
@itaiferber @phausler Oh I didn't realize that. Thanks for the explanation!

bubski:
@itaiferber fix pushed 👌

itaiferber:
@bubski Thanks!

itaiferber:
@swift-ci Please test

Foundation/CGFloat.swift
bubski: @phausler fix pushed

ianpartridge:
Hi @bubski this is really great. Please could you explain what the problems are with `Calendar`, `CharacterSet`, `DateComponents`, `Measurement` and `TimeZone`. Are these problems tracked somewhere and do we know how to resolve them?

bubski:
@ianpartridge Sure, I don't have the full picture but I can share what I've seen so far.

### Calendar, DateComponents, TimeZone
`Calendar`, `DateComponents` and `TimeZone` I believe share the same issue.

`CFTimeZone.c` includes `<tzfile.h>`
https://github.com/apple/swift-corelibs-foundation/blob/master/CoreFoundation/NumberDate.subproj/CFTimeZone.c#L35
Which defines some paths.
The one we're interested in is `TZDIR`.

`tzfile.h` in **macOS 10.12 SDK** defines
```c
#define TZDIR	"/usr/share/zoneinfo" /* Time zone object file directory */
```
`tzfile.h` in **macOS 10.13 SDK** defines
```
#define TZDIR	"/var/db/timezone/zoneinfo" /* Time zone object file directory */
```

I'm running **macOS 10.12**.
I can't build `swift-corelibs-foundation` against **macOS 10.12 SDK**. I'm not sure why.
So I'm running **Xcode 9** and build against **macOS 10.13 SDK**.
But my `zoneinfo` doesn't exist at `/var/db/timezone/zoneinfo` but at `/usr/share/zoneinfo` since I'm on **10.12**.

I don't know where to go from here and didn't have time to look into it.
Maybe I'm fundamentally wrong here and missing something.

### CharacterSet
https://github.com/apple/swift-corelibs-foundation/blob/master/TestFoundation/TestNSCharacterSet.swift#L298

This discrepancy makes it hard for me to write reasonable tests.
I'm not sure if this was the only problem though.

### Measurement
Sorry, I don't remember :(

ianpartridge:
Thanks for the comprehensive update!

The `TZDIR` issue is known and there has been some discussion on the `swift-corelibs-dev` mailing list. As you discovered, the location of the timezone database on macOS has changed in High Sierra. For now you can either symlink `/var/db/timezone/zoneinfo` to `/usr/share/zoneinfo` or add `#define TZDIR "/usr/share/zoneinfo"` in just after the `#include` line [here](https://github.com/apple/swift-corelibs-foundation/blob/master/CoreFoundation/NumberDate.subproj/CFTimeZone.c#L35). Hopefully one of those workarounds can help move the `Codable` conformances forward?

bubski:
@ianpartridge Thanks for the info! I appreciate it and it's very helpful.

I don't want to make a PR with anything that's breaking on macOS, so for some things I might have to hold off until they underlying reasons are fixed.

ianpartridge:
Thanks @bubski :)

Can we merge this PR?  The team in IBM are waiting to start experimenting with these new conformances :)

itaiferber:

PR #1109, opened 2017-07-11
Locale: rename systemLocale() to system

ianpartridge:
As documented at https://developer.apple.com/documentation/foundation/nslocale/1407691-system


pushkarnk:
@swift-ci test and merge

PR #1112, opened 2017-07-11
Use the file attribute dictionary passed to createFile

weichsel:
Currently the optional `attributes` parameter is ignored in `createFile`.  
I added a call to `setAttributes` (if applicable) to the implementation and some test code to the existing `test_createFile` case.


phausler:
@swift-ci please test

PR #1113, opened 2017-07-11
Rewrite of PR #1103 + addition Android specific patches

johnno1962:
Hi, This PR is a rewrite of https://github.com/apple/swift-corelibs-foundation/pull/1103 after discussion that started just after it was merged. This is a restructuring of the Android specific api to provide the path to a certificates authorities file required for libcurl to work. It also includes a few other patches required for foundation to build for Android.


Foundation/Host.swift
```diff
@@ -138,7 +140,12 @@ open class Host: NSObject {
                 }
                 let sa_len: socklen_t = socklen_t((family == AF_INET6) ? MemoryLayout<sockaddr_in6>.size : MemoryLayout<sockaddr_in>.size)
                 let lookupInfo = { (content: inout [String], flags: Int32) in
-                    if getnameinfo(info.ai_addr, sa_len, host, socklen_t(NI_MAXHOST), nil, 0, flags) == 0 {
+                    #if os(Android)
+                    typealias maxhostlen_t = Int
+                    #else
+                    typealias maxhostlen_t = socklen_t
+                    #endif
+                    if getnameinfo(info.ai_addr, sa_len, host, maxhostlen_t(NI_MAXHOST), nil, 0, flags) == 0 {
```
xwu: What is going on here? The fourth argument should be of type `size_t` on all platforms, should it not? I'm not aware of any type known as `maxhostlen_t`. If necessary, you can simply use `numericCast(NI_MAXHOST)` on all platforms.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -172,18 +169,13 @@ extension _EasyHandle {
...
-            // Downloadable here: https://curl.haxx.se/ca/cacert.pem
-            if let caInfo = _EasyHandle._CAInfoFile  {
-                if String(cString: caInfo) == "UNSAFE_SSL_NOVERIFY" {
-                    try! CFURLSession_easy_setopt_int(rawHandle, CFURLSessionOptionSSL_VERIFYPEER, 0).asError()
-                }
-                else {
-                    try! CFURLSession_easy_setopt_ptr(rawHandle, CFURLSessionOptionCAINFO, caInfo).asError()
-                }
-            }
+        if URLSession.sslUnsafeBypassPeerVerify {
```
xwu: `Unsafe` should not be part of this name. In Swift, "safe" and "unsafe" refer to memory safety and there is nothing unsafe about this particular property.

Furthermore, as this is a Boolean property, it should be named according to Swift guidelines for Boolean properties (i.e., it should begin with a verb such as "is," "has," or in this case "verify"). Probably best to align with underlying cURL usage and phrase this in the positive, as in: `URLSession.verifyPeerSSLCertificate` (default: true).

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -172,18 +169,13 @@ extension _EasyHandle {
...
+            try! CFURLSession_easy_setopt_int(rawHandle, CFURLSessionOptionSSL_VERIFYPEER, 0).asError()
+        }
+        else if let caInfo = URLSession._sslCertificateAuthorityFile {
+            try! CFURLSession_easy_setopt_int(rawHandle, CFURLSessionOptionSSL_VERIFYPEER, 1).asError()
+            try! CFURLSession_easy_setopt_ptr(rawHandle, CFURLSessionOptionCAINFO, caInfo).asError()
+        }
```
xwu: Else...? That is, what happens if the user wants verification but has a `nil` value for `_sslCertificateAuthorityFile`?

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,38 @@ internal extension CFURLSessionEasyCode {
         throw self
     }
 }
+
+#if os(Android)
+extension URLSession {
+    /// See https://curl.haxx.se/docs/sslcerts.html
+    /// For SSL to work you need a "cacert.pem" to be accessible at the
+    /// by setting URLSession.sslCertificateAuthorityFile to a file URL.
```
xwu: This sentence is ungrammatical--"to be accessible at the <what???> by setting"...

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,38 @@ internal extension CFURLSessionEasyCode {
...
+    /// Downloadable here: https://curl.haxx.se/docs/caextract.html
+    /// Alternatively you can bypass SSL peer verification altogether:
+    /// URLSession.sslUnsafeBypassPeerVerify = true (not advised)
```
xwu: These doc comments should be affixed to the corresponding properties. Here, they are documenting a `fileprivate static var`, which is not useful in autogenerated documentation.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,38 @@ internal extension CFURLSessionEasyCode {
...
+    fileprivate static var _sslCertificateAuthorityFile: UnsafeMutablePointer<Int8>?
+    public static var sslUnsafeBypassPeerVerify = false
+    public static var sslCertificateAuthorityFile: URL {
+        set(value) {
+            sslUnsafeBypassPeerVerify = false
+            free(_sslCertificateAuthorityFile)
+            guard value.isFileURL && FileManager.default.fileExists(atPath: value.path) else {
+                NSLog("*** sslCertificateAuthorityFile not FileURL or does not exist ***")
```
xwu: * Put single quotes around properties (i.e. 'sslCertificateAuthorityFile').
* "FileURL" is neither a real type name nor an English word.

Reword:
> 'sslCertificateAuthorityFile' does not represent a file that exists

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,38 @@ internal extension CFURLSessionEasyCode {
...
+
+#if os(Android)
+extension URLSession {
+    /// See https://curl.haxx.se/docs/sslcerts.html
+    /// For SSL to work you need a "cacert.pem" to be accessible at the
+    /// by setting URLSession.sslCertificateAuthorityFile to a file URL.
+    /// Downloadable here: https://curl.haxx.se/docs/caextract.html
+    /// Alternatively you can bypass SSL peer verification altogether:
+    /// URLSession.sslUnsafeBypassPeerVerify = true (not advised)
+    fileprivate static var _sslCertificateAuthorityFile: UnsafeMutablePointer<Int8>?
```
xwu: Shouldn't this be `_sslCertificateAuthorityFilePathPtr` or something like that? It's not simply the backing variable for `sslCertificateAuthorityFile`.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,38 @@ internal extension CFURLSessionEasyCode {
...
+            value.path.withCString {
+                _sslCertificateAuthorityFile = strdup($0)
+            }
+        }
+        get {
+            if let value = _sslCertificateAuthorityFile {
+                return URL(fileURLWithPath: String(cString: value))
+            }
+            else {
+                return URL(fileURLWithPath: "NOVALUE")
```
xwu: Huh? There is a standard way to indicate the absence of something in Swift, and that's `nil`. This computed property should be of type `URL?`, and here the return value should be `nil`.

Foundation/NSURLSession/http/EasyHandle.swift
phausler: this should probably be fatal

Foundation/Host.swift
```diff
@@ -138,7 +140,12 @@ open class Host: NSObject {
                 }
                 let sa_len: socklen_t = socklen_t((family == AF_INET6) ? MemoryLayout<sockaddr_in6>.size : MemoryLayout<sockaddr_in>.size)
                 let lookupInfo = { (content: inout [String], flags: Int32) in
-                    if getnameinfo(info.ai_addr, sa_len, host, socklen_t(NI_MAXHOST), nil, 0, flags) == 0 {
+                    #if os(Android)
```
phausler: our other guards in Foundation are outdented to the first column

Foundation/NSURLSession/http/EasyHandle.swift
phausler: again also should be fatal

Foundation/Host.swift
johnno1962: Thanks, top tip. Didin’t know about numericCast()!

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: Now verifyPeerSSLCertificate.

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: No bypass + no authority file == SSL not working.

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: ok, better than a nil I figure.

johnno1962:
Ready to roll?

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -172,18 +169,13 @@ extension _EasyHandle {
...
-                    try! CFURLSession_easy_setopt_ptr(rawHandle, CFURLSessionOptionCAINFO, caInfo).asError()
-                }
-            }
+        if !URLSession.verifyPeerSSLCertificate {
+            try! CFURLSession_easy_setopt_int(rawHandle, CFURLSessionOptionSSL_VERIFYPEER, 0).asError()
+        }
+        else if let caInfo = URLSession._sslCertificateAuthorityFilePathPtr {
+            try! CFURLSession_easy_setopt_int(rawHandle, CFURLSessionOptionSSL_VERIFYPEER, 1).asError()
+            try! CFURLSession_easy_setopt_ptr(rawHandle, CFURLSessionOptionCAINFO, caInfo).asError()
+        }
```
xwu: Again, what happens if neither of these cases is true? It should be a fatal error, probably.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,35 @@ internal extension CFURLSessionEasyCode {
...
+    fileprivate static var _sslCertificateAuthorityFilePathPtr: UnsafeMutablePointer<Int8>?
+    /// You can bypass SSL peer verification altogether (not advised)
+    public static var verifyPeerSSLCertificate = true
+    /// See https://curl.haxx.se/docs/sslcerts.html
+    /// For SSL to work you need a "cacert.pem" to be accessible at the path
+    /// provided by setting URLSession.sslCertificateAuthorityFile to a URL.
+    /// Downloadable here: https://curl.haxx.se/docs/caextract.html
+    public static var sslCertificateAuthorityFile: URL {
+        set(value) {
+            verifyPeerSSLCertificate = true
```
xwu: Setting `sslCertificateAuthorityFile` should not automatically flip the switch on the second, independent setting.

If the user bothered to turn off certificate verification, then it should stay off until it's explicitly flipped on again. Nothing about setting a file path for CA information implies that you'd automatically want to verify SSL certificates. That is, after all, the intuition that led you to separate these two things into two properties instead of one.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,35 @@ internal extension CFURLSessionEasyCode {
...
+            free(_sslCertificateAuthorityFilePathPtr)
+            guard value.isFileURL && FileManager.default.fileExists(atPath: value.path) else {
+                fatalError("'sslCertificateAuthorityFile' set: not file URL or does not exist ***")
```
xwu: How about let's use a grammatical error message: "'sslCertificateAuthorityFile' does not reference a file that exists"

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,35 @@ internal extension CFURLSessionEasyCode {
...
+            value.path.withCString {
+                _sslCertificateAuthorityFilePathPtr = strdup($0)
+            }
+        }
+        get {
+            if let value = _sslCertificateAuthorityFilePathPtr {
+                return URL(fileURLWithPath: String(cString: value))
+            }
+            else {
+                fatalError("'sslCertificateAuthorityFile' get: no value available")
```
xwu: Likewise here: "'sslCertificateAuthorityFile' has not been set"

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: ok

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: Altered to return nil

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,35 @@ internal extension CFURLSessionEasyCode {
         throw self
     }
 }
+
+#if os(Android)
+extension URLSession {
+    fileprivate static var _sslCertificateAuthorityFilePathPtr: UnsafeMutablePointer<Int8>?
+    /// You can bypass SSL peer verification altogether (not advised)
```
xwu: Doc comments have a particular style in Swift. Here, let's try to mimic that:

```
/// A Boolean that is `true` if peer SSL certificates are to be verified.
///
/// You can bypass verification by setting this property to `false` (not recommended).
```

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: I’ll push back on this one. There is no point in setting the CA file is verifyPeerSSLCertificate isn’t true. This is for safety.

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: It may work one day if libcurl starts to support SSL. Also, At this point we don’t know if the user is making an SSL request.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,35 @@ internal extension CFURLSessionEasyCode {
...
+    public static var verifyPeerSSLCertificate = true
+    /// See https://curl.haxx.se/docs/sslcerts.html
+    /// For SSL to work you need a "cacert.pem" to be accessible at the path
+    /// provided by setting URLSession.sslCertificateAuthorityFile to a URL.
+    /// Downloadable here: https://curl.haxx.se/docs/caextract.html
+    public static var sslCertificateAuthorityFile: URL {
```
xwu: /// The URL representing a PEM-encoded certificate authority file to be used for peer SSL certificate verification.
///
/// See:
/// - https://curl.haxx.se/docs/sslcerts.html
/// - https://curl.haxx.se/docs/caextract.html

Foundation/NSURLSession/http/EasyHandle.swift
xwu: Then there should not be two distinct properties.

Does cURL automatically discard the one setting when the other is set? If not, then users would have no expectation that it's the case here.

Moreover, your implementation does nothing in particular to provide such "safety" if a user sets this property and then the Boolean. Doing one thing vs. another when two independent properties are set in a different order is difficult to reason about.

Foundation/NSURLSession/http/EasyHandle.swift
xwu: I see.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,35 @@ internal extension CFURLSessionEasyCode {
...
+
+#if os(Android)
+extension URLSession {
+    fileprivate static var _sslCertificateAuthorityFilePathPtr: UnsafeMutablePointer<Int8>?
+    /// You can bypass SSL peer verification altogether (not advised)
+    public static var verifyPeerSSLCertificate = true
+    /// See https://curl.haxx.se/docs/sslcerts.html
+    /// For SSL to work you need a "cacert.pem" to be accessible at the path
+    /// provided by setting URLSession.sslCertificateAuthorityFile to a URL.
+    /// Downloadable here: https://curl.haxx.se/docs/caextract.html
```
xwu:
```
/// The URL representing a PEM-encoded certificate authority file to be used for peer SSL certificate verification.
///
/// See:
/// - https://curl.haxx.se/docs/sslcerts.html
/// - https://curl.haxx.se/docs/caextract.html
```

johnno1962: Thanks. Updated

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: Originally there was only one property a string which could be null, a path or a dummy value. In some ways it was better but the code got to be a bit of a mess. The only alternative would be some kind of magic enum that looked after the ownership of the file path pointer. While it’s interesting to meditate on making it airtight I think we need to draw a guillotine down on this. Its a very niche api. I think the api is at least simple which is a good start.

xwu: Yes, I recall. I'm not arguing for a more complicated API. What I'm saying is, since you've chosen the path of presenting two properties, they should be independent of each other. This is consistent with user expectation both in Foundation and in cURL. Precisely _because_ this is a niche API that will not be amply documented, it's important that these expectations are met. It is very surprising that...

```swift
NSSession.verifyPeerSSLCertificate = false
NSSession.sslCertificateAuthorityFile = url
```

...gives a different result than...

```swift
NSSession.sslCertificateAuthorityFile = url
NSSession.verifyPeerSSLCertificate = false
```

...keeping in mind that the default for `verifyPeerSSLCertificate` is `true`, so if a user wants that setting to be `false`, he or she will necessarily have to set that value somewhere explicitly.

johnno1962: I've updated the Doc comment to make explicit that it does this and I’d like to leave it that way if possible. I hear you about the inconsistency but in this practical context it’s important to not have SSL unprotected when you specifiy a file URL just because you have previously happened to have set verify to false somewhere else in your code base.

johnno1962: OK, final offer. I’ve updated the code to ignore the value of verifyPeerSSLCertificate if a CA cert file has been set. This removes the dependency on the order of calls with the presence of a cert file taking preceedence between the two properties which is simpler to reason.

```swift

#if os(Android)
extension URLSession {
    fileprivate static var _sslCertificateAuthorityFilePathPtr: UnsafeMutablePointer<Int8>?
    /// A Boolean that is `true` if peer SSL certificates are to be verified.
    ///
    /// You can bypass verification by setting this property to `false` (not recommended).
    public static var verifyPeerSSLCertificate = true
    /// The URL representing a PEM-encoded certificate authority file to be used for
    /// peer SSL certificate verification. Setting a certificate authority file takes
    /// precedence over tha value of verifyPeerSSLCertificate.
    ///
    /// See:
    /// - https://curl.haxx.se/docs/sslcerts.html
    /// - https://curl.haxx.se/docs/caextract.html
    public static var sslCertificateAuthorityFile: URL? {
        set(value) {
            free(_sslCertificateAuthorityFilePathPtr)
            guard let value = value else {
                _sslCertificateAuthorityFilePathPtr = nil
                return
            }
            guard value.isFileURL && FileManager.default.fileExists(atPath: value.path) else {
                fatalError("'sslCertificateAuthorityFile' does not reference a file that exists")
            }
            value.path.withCString {
                _sslCertificateAuthorityFilePathPtr = strdup($0)
            }
        }
        get {
            if let value = _sslCertificateAuthorityFilePathPtr {
                return URL(fileURLWithPath: String(cString: value))
            }
            else {
                return nil
            }
        }
    }
}
#endif
```

alblue:
@swift-ci please test

johnno1962:
Sorry, fix of typo..

Foundation/NSURLSession/http/EasyHandle.swift
xwu: You’ll need to correct the typos in that comment and use consistent terminology (i.e., “certificate authority file”).

However, this design is not the right way to go, IMO. You’ve got to commit to your convictions. If you believe that there is some configuration that is insecure, then it should be prevented and not merely made unlikely. That is, setting a non-nil CA file should be a fatal error when the other setting is false, and changing the other setting to false should be a fatal error too. On the other hand, you can say that the configuration is unwise but legitimate; in that case, there should not be some arbitrary sequence of setting properties that prevents it but others that don’t.

xwu: (This comment was in reference to the previous design. Let me think about this one. I like it better...)

johnno1962: While I enjoy our online jousts this one is getting a bit OTT :) An app dev has two choices when they develop an app: Set verifyPeerSSLCertificate to false (not advised but can be useful if you self-sign during development) or set sslCertificateAuthorityFile to a valid URL of a pem fille. If they do both the more secure option prevails. I’d realy like to keep this simple rather than create a minefiled of interlocks that need to be implemented and documented.

xwu: Well, thanks for bearing with me. I do like this design better. No need for the `fatalError` guard since you've documented the behavior. In fact, I think I like it better without the additional guard, as the guard itself is triggered or not depending on the order in which the properties are set :P

johnno1962: Cool. Guard removed. I think we’re done here. Now, about elided newlines i nmultiline strings...

xwu: :D

johnno1962:
@phausler I think this PR is ready now and I’ve just done a final test run. Merge away!

johnno1962:
For reference, the android CACerts are there but are not prepared correctly for openSSL to use http://www.javacms.tech/questions/1728116/how-to-make-ssl-peer-verify-work-on-android

phausler:
@swift-ci please test

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,43 @@ internal extension CFURLSessionEasyCode {
...
+    /// - https://curl.haxx.se/docs/caextract.html
+    public static var sslCertificateAuthorityFile: URL? {
+        set(value) {
+            free(_sslCertificateAuthorityFilePathPtr)
+            guard let value = value else {
+                _sslCertificateAuthorityFilePathPtr = nil
+                return
+            }
+            guard value.isFileURL && FileManager.default.fileExists(atPath: value.path) else {
+                fatalError("'sslCertificateAuthorityFile' does not reference a file that exists: \(value.path)")
```
phausler: So if the path is a directory this won't hit btw

johnno1962: D'oh

johnno1962: I’ve got a fix do you want me to push it through?

```swift
            var isDir: ObjCBool = false
            guard value.isFileURL && FileManager.default.fileExists(atPath: value.path,
                                                                    isDirectory: &isDir) && !isDir else {
                fatalError("'sslCertificateAuthorityFile' does not reference a file that exists: \(value.path)")
            }
```

johnno1962: pushed

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -661,7 +661,9 @@ extension URLSession {
                 _sslCertificateAuthorityFilePathPtr = nil
                 return
             }
-            guard value.isFileURL && FileManager.default.fileExists(atPath: value.path) else {
+            var isDir: ObjCBool = false
+            guard value.isFileURL && FileManager.default.fileExists(atPath: value.path,
+                                                                    isDirectory: &isDir) && !isDir else {
```
phausler: Looks good to me

phausler:
@swift-ci please test

johnno1962:
Did that second please test go through?

xwu:
@johnno1962 ...Uh, did you mean to invalidate testing for this PR by pushing unrelated changes?

Foundation/NSPathUtilities.swift
```diff
@@ -30,7 +30,11 @@ public func NSTemporaryDirectory() -> String {
             return tmpdir
         }
     }
+    #if os(Android)
```
xwu: This should be outdented, if I recall, according to Philippe.

johnno1962: Just trying to follow the existing guards in the file.. Should I change those too?

xwu: Nah, you're right. I didn't see the other ones.

johnno1962:
Testing didn’y go through anyway got some reason...

Foundation/NSPathUtilities.swift
johnno1962: The following would also need to be changed
	modified:   Foundation/Data.swift
	modified:   Foundation/DateFormatter.swift
	modified:   Foundation/NSData.swift
	modified:   Foundation/NSObjCRuntime.swift
	modified:   Foundation/NSPathUtilities.swift
	modified:   Foundation/NSUserDefaults.swift
	modified:   Foundation/NumberFormatter.swift
	modified:   Foundation/Process.swift

johnno1962: /data/local/tmp isn’t writable by an app anyway so I’ve reverted this change. You have to set TMPDIR env var to the Content.getCacheDir() for various things to work.

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci please test

pushkarnk:
👆You need to request twice, sometimes!

johnno1962:
Thanks @pushkarnk, looks good to me now.

johnno1962:
@phausler, would you like me to delete the “android" directory. It’s obsolete now.

parkera:
Sure, that'd be good.

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -660,3 +639,45 @@ internal extension CFURLSessionEasyCode {
...
+    /// You can bypass verification by setting this property to `false` (not recommended).
+    public static var verifyPeerSSLCertificate = true
+    /// The URL representing a PEM-encoded certificate authority file to be used for
+    /// peer SSL certificate verification. Setting a certificate authority file takes
+    /// precedence over the value of verifyPeerSSLCertificate.
+    ///
+    /// See:
+    /// - https://curl.haxx.se/docs/sslcerts.html
+    /// - https://curl.haxx.se/docs/caextract.html
+    public static var sslCertificateAuthorityFile: URL? {
```
parkera: Wait a minute here - are we attempting to add new Android-specific API?

johnno1962: It’s Linux specific really where libcurl is being used but only essential on Android where the CA root certificates don't work with OpenSSL 1.0. Otherwise, no https: requests work. http://www.javacms.tech/questions/1728116/how-to-make-ssl-peer-verify-work-on-android

parkera: Where does the developer get the value to pass to this API? When do they set it? Why can't we just do the right thing for them automatically so they don't have platform-specific API to use?

johnno1962: My god, it’s a long story on Android…. very difficult to replicate in C. The simplest way is to add a raw asset for the cacerts.pem file then copy that to a place where you can get the actual file path in Java and pass that to Swift to set using this api. https://github.com/SwiftJava/swift-android-samples/blob/master/swifthello/src/main/java/net/zhuoweizhang/swifthello/SwiftHello.java#L35

johnno1962: This way we don’t fill up corefoundation with quite a bit of android specific NDK code.

johnno1962: A bit more info here.. https://stackoverflow.com/questions/13317387/how-to-get-file-in-assets-from-android-ndk might not be insumountable...

johnno1962: OK, so it looks like an NDK solution is insurmountable. There simply isn’t enough api to find resources on an app that starts out as Java as far as I can see. The original implementatoin was an enivronment variable avoiding an API but that was not considered secure which is how we arrived here.

johnno1962:
I thnk this is as ready to merge as it will ever be.

johnno1962:
@parkera, I can roll this API back to being an environment variable if it will help get it merged. I’ve had no luck looking for an NDK based solution. I’ve also added code to pass the error message from curl through to the NSError thrown when there is a network error.

parkera:
Yah, I would prefer the environment variable for now. We're really cautious about adding API.

johnno1962:
Thanks for getting back. I’ve reverted to an env var URLSessionCAInfo.

johnno1962:
You again :)... I agree for an api for for an environment variable I think “UNSAFE_SSL_NOVERIFY” encapsulates that is is a bad idea more than just “NOVERIFY”. If you feel the word UNSAFE already has a conitation in Swift, I could settle for “UNSECURE_SSL_NOVERIFY”. I want to make sure there is a clear disinsentive for people who might be tempted to avoid the problems wit pem files etc.

xwu:
_Insecure_, not _unsecure_; but yes, that's much better IMO. Let's throw in an underscore between "NO" and "VERIFY", a la "GIT_SSL_NO_VERIFY".

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -90,8 +87,8 @@ extension _EasyHandle {
 }
 
 internal extension _EasyHandle {
-    func completedTransfer(withErrorCode errorCode: Int?) {
-        delegate?.transferCompleted(withErrorCode: errorCode)
+    func completedTransfer(withErrorCode errorCode: Int?, errorDesc: String?) {
```
xwu: Here and elsewhere, this should be spelled out: `errorDescription`, as it is in the rest of the project.

johnno1962:
Cool, “INSECURE_SSL_NO_VERIFY” it is then.

Foundation/NSURLSession/http/EasyHandle.swift
johnno1962: fixed.

johnno1962:
Good to go my end

johnno1962:
Conflicts resolved, https://github.com/apple/swift-corelibs-foundation/pull/1122 may also be merged soon

johnno1962:
@xwu, I’m minded to change the environment variable name from “URLSessionCAInfo” to “URLSessionCertificateAuthorityInfoFile”. Thoughts?

johnno1962:
Hi @phausler, @parkera, @pushkarnk any chance we could get moving on this PR again please? I’ve updated swift to master my end and made one further change that seemed to be required and renamed the environment variable URLSessionCertificateAuthorityInfoFile so it is self documenting. This PR also deletes the unsightly “android” directory files which are way out of date and provides a localized description from curl in the NSError thrown when there is a networking failure. Thanks!

xwu:
@johnno1962 Can we leave the TimeZone changes out of this PR?

It’s quite another kettle of fish and the changes here aren’t clearly correct (for example: is GMT guaranteed to be equal to UTC for all time going forward? why does “GMT” need to be rewritten to “GMT+00”? should any of this be done in the Swift portion of Foundation or in CF? is simply logging the error acceptable, or should this fatalError on Android? ultimately, a complete solution would be to figure out how to hook up CF to Android’s tzinfo—this should not be insurmountable and may not even be more involved than the workaround.)

johnno1962:
No problem. It was mainly to get the tests working. I’ll check again if there is a solution in the NDK.

johnno1962:
@xwu I’ve resubmitted the time zone fix under https://github.com/apple/swift-corelibs-foundation/pull/1189 in the finish. I’ve looked into the NDK options and they are very limited like they are for any UNIX. On my device there is no /usr/share/zoneinfo/ database but a single file in a proprietary format known only to Java. Answering your reservations one by one UTC and GMT are synonnyms  for all intents an purposes in this context and GMT needs to be rewritten to GMT+0000 as this is the only format CFTimeZone.c understands without acces to a database by name. I’d rather not try to alter CFTimeZone.c as it is sourced differently from the swift and swift is a higher level language. An invalid Timezone should be fatal when you are debugging but a foundation library should not crash peoples apps out in the field IMO so I’ve opted for the latter. I’ve introduced a pseudo timezone “System” which people can use to format dates to the timezone of the device.

xwu:
@johnno1962 Let's discuss the time zone issue on #1189, then.

amraboelela:
Please don't delete the files in android directory as I am working on fixing them. Thanks.

alblue:
@johnno1962 this PR has become somewhat stale with everything else that's been going on; can we close this one down and then later open a new PR for when the remaining changes can be viewed? I presume you've got the changes in a branch and can call them back when needed. Keeping it open won't help get it reviewed any faster, and rebasing/amending after the work on other PRs isn't likely to be of benefit.

johnno1962:
@alblue, I’m still of the opinion this PR can be rehabilitated once the other PR’s have landed. It’s actually quite minor and I’d like to see it through on principle. Once https://github.com/apple/swift-corelibs-foundation/pull/1198 has landed I’ll rebase and you’ll see what I mean.

alblue:
@johnno1962 the problem is (from my reviewer's perspective) is that this thread has had 90+ comments on it and covered several different versions of the changes that you want to make, some of which are no longer relevant to the ultimate change set and with commentary that has diverged. So even if this change can be rehabilitated on this PR, it's still going to be difficult to go through and figure out what is still relevant and what isn't.

On the other hand, a new PR rebased onto master as smaller patches is much easier to get through, because they can be reviewed individually. It also helps from a punctuality perspective because more recently created PRs are towards the top of the list of reviews, and this one is languishing down at the bottom (so I have to remember to go and look for it specifically).

Finally, for this particular change set, it was a rebase and split from a prior PR with a number and 'android specific patches' as the title - so it's not really clear to me what this is actually trying to achieve.

So once the other patch set you're working on is changed, then rebasing and taking whatever is outstanding on this PR and creating a new one will make it much easier for me (and others) to help review and move through, especially if the resulting PR is relatively small. More frequent smaller PRs that are targeted to fix a specific issue are better than one large one with a lot of different changes.

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -27,7 +27,9 @@
 #include <CoreFoundation/ForFoundationOnly.h>
 #include <fts.h>
 #include <pthread.h>
+#ifndef __ANDROID__
```
amraboelela: This is going to have conflict with https://github.com/apple/swift-corelibs-foundation/pull/1249

Foundation/URLSession/http/HTTPURLProtocol.swift
```diff
@@ -528,16 +532,16 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
         }
     }
 
-    func transferCompleted(withErrorCode errorCode: Int?) {
+    func transferCompleted(withError error: NSError?) {
```
amraboelela: How is this related to Android only?

johnno1962:
Hi @alblue, chipping away at this rather ill-fated PR I’ve created https://github.com/apple/swift-corelibs-foundation/pull/1264 which contains the original nub of the original PR. Hopefully we can get this merged without too much drama.

Foundation/URLSession/http/HTTPURLProtocol.swift
johnno1962: Wait for the rebase

johnno1962:
HI @alblue et all. This PR has been rebased and tested and be merged as is. It contains three changes:

1. The original change to Foundation/NSURLSession/http/EasyHandle.swift that started all this off.
2. Deleting the out of date scripts in the android directory.
3. The remainder are the minor changes to have Foundation build for Android.

As item 3. will clash with @amraboelela’s versions there is also PR #1264 which will contains just items 1. & 2. which you could also merge - Take your pick. Note though, this PR has been built on OSX and Android, and the foundation tests run to competition (with the last commit applied for android.)

	 Executed 1161 tests, with 25 failures (0 unexpected) in 16.021 (16.021) seconds

You can ignore almost all of the above comments as in the end the PR was presented separately and rebased to pretty much just the original changes to Foundation/NSURLSession/http/EasyHandle.swift

Thanks for the drive on merging outstanding PRs this week. It’s been a big help to me remaining sane!

alblue:
Since #1264 has been put through test-and-merge, I'm going to close this down and leave any outstanding fixes to other/newly filed PRs. Thanks for splitting it up and making the process easier for us to review!

PR #1114, opened 2017-07-11
[Linux] Remove swift_begin/swift_end from libFoundation.a

spevans:
- swift_begin.o and swift_end.o should only be added to dynamic
  libraries and to the executable, which covers any static libraries
  that are linked in.


parkera:
@swift-ci test and merge

PR #1117, opened 2017-07-14
Fix negation of 0 length Decimal

mortenbekditlevsen:
A Decimal value with _length 0 and _isNegative set to 1 is interpreted as a NaN. The 'negate()' function however, flipped the _isNegative flag without regard for the _length 0 case. This meant that -0 would become NaN. The fix checks for the _length 0 special case. In NSDecimalSubtract() the same check was performed. Since this now happens in negate(), it is removed from the NSDecimalSubtract() function.

This pull request fixes rdar://33311786


ianpartridge:
When I run

```swift
import Foundation
var e = Decimal(0)
print(e)
e.negate()
print(e)
```

in an Xcode playground I see

```
0
NaN
```

so it looks like we are already compatible with Darwin here?

mortenbekditlevsen:
Ok, I fixed the wrong error then. 
In Xcode 9 beta 3 the infix operator - started behaving differently. For:

import Foundation
let e = Decimal(0)
print(e)
print(-e)

In Xcode 8.3.3 you get:

0
0

While in Xcode 9 beta 3 you get
0
NaN

So the error appears to be in the infix operator overload instead.
I would love to see if I can find the correct place to fix it.

mortenbekditlevsen:
Ok, now I have compared the file:
stdlib/public/core/Integers.swift.gyb in swift-4.0-branch of the swift repo
and
stdlib/public/core/IntegerArithmetic.swift.gyb in swift-3.0-branch 

In swift 3 we had:
public prefix func - <T : SignedNumber>(x: T) -> T {
  return 0 - x
}

While in swift 4 the SignedNumeric defines:
extension SignedNumeric {
  @_transparent
  public static prefix func - (_ operand: Self) -> Self {
    var result = operand
    result.negate()
    return result
  }

  @_transparent
  public mutating func negate() {
    self = 0 - self
  }
}

Since the Decimal type implements it's own 'negate' then I guess this is why we get the NaN instead of the 0 when using the minus prefix operator.

It is not 100% clear to me how this should be fixed - but it is probably in the swift core instead of in corelibs-foundation.

Would you happen to have some insight into how this could be fixed?

ianpartridge:
The difference in behaviour between Xcode 8 and 9 is interesting. It might be worth filing a Radar about that - what do you think @phausler?

mortenbekditlevsen:
I filed a radar before beginning to look into the issue: rdar://33311786 (http://www.openradar.me/radar?id=5031952226713600)

I must admit that I was getting a bit confused about what lies in Foundation and what is a swift overlay. As far as I can tell the 'negate' behavior is not Foundation, but part of stdlib/public/SDK/Foundation/Decimal.swift.

As such, I believe that 'negate' has been broken since the Decimal extensions were introduced. But if you used the prefix operator instead of calling negate directly, you would never see this since SignedNumber called the infix minus operator (on 0 - value) - thus hitting the NSDecimalSubtract function.

So the issue is that in Swift 3, 'a.negate(); return a' behaved differently than '-a' - and in Swift 4 they behave the same since '-a' just calls 'a.negate(); return a'.

So it does not seem possible to maintain backwards compatibility here (without beginning to redefine the behavior of the SignedNumeric protocol in Swift 4).

I would opt for fixing the logical error here. This would basically merge both this pull request and a similar one in the Swift repo version of stdlib/public/SDK/Foundation/Decimal.swift

As a funny side-effect to the existing issue, calling negate on NaN gives 0. It does not make a whole lot of sense. :-)

mortenbekditlevsen:
I've now made similar changes to the swift repo. Let me know if you think that I should create a pull request for the similar change. @ianpartridge @phausler

phausler:
Yea this should be done in both places: thanks for taking this on

phausler:
@swift-ci please test

mortenbekditlevsen:
@phausler my pleasure - happy to be able to contribute. :-)

Here's my similar pull request for the swift repo:
https://github.com/apple/swift/pull/10967

PR #1119, opened 2017-07-15
Match types when calculating remaining in FileHandle's _readDataOfLength

gonzalolarralde:
This was causing an error when trying to get the remaining amount of data.

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -76,7 +76,7 @@ open class FileHandle : NSObject, NSSecureCoding {
                 fatalError("Unable to fetch current file offset")
             }
             if off_t(statbuf.st_size) > offset {
-                var remaining = size_t(statbuf.st_size - offset)
+                var remaining = size_t(off_t(statbuf.st_size) - offset)
                 remaining = min(remaining, size_t(length))
                 
                 dynamicBuffer = malloc(remaining)


mattrajca:
Can you add a test for this?

pushkarnk:
Per [this](https://linux.die.net/man/2/stat) the `st_size` member is also of type `off_t`. Am I missing something here?

gonzalolarralde:
@mattrajca @pushkarnk sorry for the lack of context in the original message.

@pushkarnk it is true for linux, it is not for 32-bits arm or at least in [bionic](https://github.com/android/platform_bionic/blob/master/libc/include/sys/stat.h#L107). This is causing a compilation error when trying to combine a `long long` with an `off_t` without casting.

@mattrajca this wasn't causing a runtime error but a compilation error under some particular circumstances (`-target armv7-linux-androideabi` in my case)

parkera:
@swift-ci please test

parkera:
Let's see if this causes a compilation error on other platforms...

pushkarnk:
I think we can merge this. It seems to be blocking compilation of Foundation on the said platform. On platforms where `st_size` is of type `off_t` (where a cast isn't necessary), I don't think this explicit cast should cause a problem.

pushkarnk:

PR #1120, opened 2017-07-20
Stub out new KeyPath APIs in NSSortDescriptor

ddunn2:
There are new KeyPath APIs in `NSSortDescriptor` which needed stubbing out in swift-corelibs-foundation. This PR adds those stubs using the declarations found [here](https://developer.apple.com/documentation/foundation/nssortdescriptor?changes=latest_minor)

### DIFF
--- a/Foundation/NSSortDescriptor.swift
+++ b/Foundation/NSSortDescriptor.swift
@@ -35,10 +35,13 @@ open class NSSortDescriptor: NSObject, NSSecureCoding, NSCopying {
     
     open var key: String? { NSUnimplemented() }
     open var ascending: Bool { NSUnimplemented() }
+    var keyPath: AnyKeyPath? { NSUnimplemented() }
     
     open func allowEvaluation() { NSUnimplemented() } // Force a sort descriptor which was securely decoded to allow evaluation
     
     public init(key: String?, ascending: Bool, comparator cmptr: Comparator) { NSUnimplemented() }
+    convenience init<Root, Value>(keyPath: KeyPath<Root, Value>, ascending: Bool) { NSUnimplemented() }
+    convenience init<Root, Value>(keyPath: KeyPath<Root, Value>, ascending: Bool, comparator cmptr: @escaping Comparator) { NSUnimplemented() }
     
     open var comparator: Comparator { NSUnimplemented() }
     


pushkarnk:
@swift-ci test and merge

Foundation/NSSortDescriptor.swift
```diff
@@ -35,10 +35,13 @@ open class NSSortDescriptor: NSObject, NSSecureCoding, NSCopying {
     
     open var key: String? { NSUnimplemented() }
     open var ascending: Bool { NSUnimplemented() }
+    var keyPath: AnyKeyPath? { NSUnimplemented() }
```
ikesyo: This should be `open var`.

Foundation/NSSortDescriptor.swift
```diff
@@ -35,10 +35,13 @@ open class NSSortDescriptor: NSObject, NSSecureCoding, NSCopying {
...
     
     open func allowEvaluation() { NSUnimplemented() } // Force a sort descriptor which was securely decoded to allow evaluation
     
     public init(key: String?, ascending: Bool, comparator cmptr: Comparator) { NSUnimplemented() }
+    convenience init<Root, Value>(keyPath: KeyPath<Root, Value>, ascending: Bool) { NSUnimplemented() }
+    convenience init<Root, Value>(keyPath: KeyPath<Root, Value>, ascending: Bool, comparator cmptr: @escaping Comparator) { NSUnimplemented() }
```
ikesyo: These should be `public`.

ddunn2: I took the declaration directly from [here](https://developer.apple.com/documentation/foundation/nssortdescriptor/2908856-init?changes=latest_minor)
Are we saying this is incorrect or that it's common practice to replace convenience with public?

ikesyo: Symblos in the API documents are all `public` (or `open`). If you don't specify any access control it defaluts to `internal` visibility so consumers of the framework can't see that symbols.

You should check here: https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AccessControl.html.

ddunn2: Perfect, thanks for taking the time to explain that :) 

I'll update shortly with the correct access controls, I'll also have a read through the documentation you've provided. 

Thanks again! :)

PR #1121, opened 2017-07-20
[URLSession] Fix for SR-5516

pushkarnk:
https://bugs.swift.org/browse/SR-5516

Upload tasks must invoke the `urlSession(_:task:didSendBodyData: totalBytesSent:totalBytesExpectedToSend)` delegate method every time data is uploaded.

### DIFF
--- a/Foundation/NSURLSession/http/HTTPURLProtocol.swift
+++ b/Foundation/NSURLSession/http/HTTPURLProtocol.swift
@@ -14,6 +14,8 @@ internal class _HTTPURLProtocol: URLProtocol {
 
     fileprivate var easyHandle: _EasyHandle!
     fileprivate var totalDownloaded = 0
+    fileprivate var totalUploaded: Int64 = 0
+    fileprivate var requestBodyLength: Int64 = 0
     fileprivate var tempFileURL: URL
 
     public required init(task: URLSessionTask, cachedResponse: CachedURLResponse?, client: URLProtocolClient?) {
@@ -127,6 +129,7 @@ fileprivate extension _HTTPURLProtocol {
                 set(requestBodyLength: .noBody)
             case (_, .some(let length)):
                 set(requestBodyLength: .length(length))
+                requestBodyLength = Int64(length)
             case (_, .none):
                 set(requestBodyLength: .unknown)
             }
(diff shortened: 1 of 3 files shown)


alblue:
@swift-ci please test

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -495,6 +498,16 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
         }
     }
 
+    fileprivate func notifyDelegate(aboutSentData count: Int64) {
```
ianpartridge: Would it be better for this function's parameter to be `Data`?

ianpartridge: As the function is specific to upload tasks I wonder if this should be mentioned in the name too.

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -495,6 +498,16 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
...
+        guard let session = self.task?.session as? URLSession else { fatalError() }
```
weissi: maybe a better error message in the `fatalError()`? Otherwise just `as! URLSession`

TestFoundation/HTTPServer.swift
```diff
@@ -339,6 +339,12 @@ public class TestURLSessionServer {
 
     func getResponse(request: _HTTPRequest) -> _HTTPResponse {
         let uri = request.uri
+
+        if uri == "/upload" {
+            let text = "Upload completed!"
+            return _HTTPResponse(response: .OK, headers: "Content-Length: \(text.characters.count)", body: text)
```
weissi: hmm, technically `text.characters.count` is not the length in bytes. It happens to be for this string in `.ascii.` but maybe we want to be more correct? `text.data(using: .ascii)...`

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -127,6 +129,7 @@ fileprivate extension _HTTPURLProtocol {
                 set(requestBodyLength: .noBody)
             case (_, .some(let length)):
                 set(requestBodyLength: .length(length))
+                bodyLength = Int64(length)
```
weissi: what happens if an attacker sends a content-length that exceeds `Int64` but not `UInt64`? Could we have a test case for that? I already added a few test cases for a misbehaving server, this one should be really easy to add.

pushkarnk: @weissi I hope I am getting your comment right here. This length isn't something sent by the server. It corresponds to the length of the data being uploaded - it is either the count of a Data in memory or the length of a file being uploaded (see the definitions of `_HTTPURLProtocol._Body` and `_Body.getBodyLength()`). A size of more that `Int64.max` doesn't seem possible then. Even if it were possible, we'd have an OOM before we get here.

Foundation/NSURLSession/http/HTTPURLProtocol.swift
pushkarnk: @ianpartridge I agree that the name should be better. However, it may not make sense to have Data as a parameter since the delegate call only needs the length of the data that was uploaded. Also, the caller of this method has the data in the form of`DispatchData`. I reckon there'll be some unnecessary cost of converting DispatchData to Data.

Foundation/NSURLSession/http/HTTPURLProtocol.swift
weissi: Sorry, ignore me, didn't quite realise this is the upload length rather than the Content-Length header sent by the server. Ignore me please!

pushkarnk:
Thanks for the review comments @ianpartridge and @weissi Can you please take a look at the changes/responses?

ianpartridge:
@swift-ci please test

weissi:
@pushkarnk 👍, thanks

ianpartridge:

PR #1123, opened 2017-07-20
Make public to doubleValue property SR-4396

giginet:
closes #SR-4396

[\[SR\-4396\] Decimal\.doubleValue should be public \- Swift](https://bugs.swift.org/browse/SR-4396?filter=10813&jql=issuetype%20%3D%20Bug%20AND%20status%20%3D%20Open%20AND%20component%20%3D%20Foundation%20AND%20assignee%20in%20(EMPTY)%20ORDER%20BY%20createdDate%20DESC)

`doubleValue` property of `Decimal` should be `public`. However, it is `internal`.

### DIFF
--- a/Foundation/NSDecimal.swift
+++ b/Foundation/NSDecimal.swift
@@ -200,7 +200,7 @@ extension Decimal {
 }
 
 extension Decimal : Hashable, Comparable {
-    internal var doubleValue: Double {
+    public var doubleValue: Double {
         var d = 0.0
         if _length == 0 && _isNegative == 1 {
             return Double.nan


pushkarnk:
Are we sure this is a part of the API? I can't find it [here](https://developer.apple.com/documentation/foundation/decimal). Or am I missing something?

giginet:
Is this issue not correct?

xwu:
@giginet The issue expresses an opinion as to what the API should be. Changes to the API need to go through Swift Evolution. Since `doubleValue` is not in fact a part of the API as designed, it is not correct to make it `public` here.

ianpartridge:
@xwu Changes to Foundation APIs are in general not handled through the Swift Evolution process. Exceptions have been made for certain APIs that are closely tied to features being developed via that process (most notably `Codable` and the corresponding Foundation classes `JSONEncoder` and `PlistEncoder`).

If `doubleValue` becomes `public` on Darwin then we will mirror that change here in `swift-corelibs-foundation`.

I think SR-4396 would be better filed as a Radar.

PR #1124, opened 2017-07-21
[test] Simplify expression in a test

moiseev:
This expression became too complex due to the changes introduced in
https://github.com/apple/swift/pull/11044

### DIFF
--- a/TestFoundation/TestNSDecimal.swift
+++ b/TestFoundation/TestNSDecimal.swift
@@ -185,7 +185,7 @@ class TestNSDecimal: XCTestCase {
             _length: 0xff,
             _isNegative: 3,
             _isCompact: 4,
-            _reserved: UInt32(1<<18 + 1<<17 + 1),
+            _reserved: 1<<18 + 1<<17 + 1,
             _mantissa: (6, 7, 8, 9, 10, 11, 12, 13)
         )
         XCTAssertEqual(0x7f, explicit._exponent)


moiseev:
@swift-ci Please test
Will it work?...

alblue:
Since this removes the clarity of where the parenthesis lies (is this 1 as Uint or express Uint?) this actually makes the initialiser worse in my opinion. 

Can you clarify why you think this is an improvement?

moiseev:
> On Jul 21, 2017, at 2:00 PM, Alex Blewitt <notifications@github.com> wrote:
> 
> Since this removes the clarity of where the parenthesis lies (is this 1 as Uint or express Uint?) this actually makes the initialiser worse in my opinion. 
> 
> Can you clarify why you think this is an improvement?
Because it makes test compile.
> 
> Sent from my iPhone 📱
>

moiseev:
@alblue I am more than willing to reinstate the clarity, but to me expression without parens is the best it can be. Adding parens like this `1<<18 + 1<<17 + (1 as UInt32)` only makes it worse, without changing the behavior.

alblue:
It's not clear why you claim that it makes the test compile since the test is compiling previously?

moiseev:
_test__simplify_expression_in_a_test_by_moiseev_ _pull_request__1124_ _apple_swift-corelibs-foundation

moiseev:
@alblue Turns out completely removing the `UInt32` hint works as well, without affecting the clarity.

parkera:
I guess this is ok; seems unfortunate to have to hack around compiler limitations like this but if we don't have a choice then we're ok to merge.

PR #1125, opened 2017-07-22
Drop RandomAccessCollection conformance for String.UTF16View

dabrahams:
This is consistent with what we've done in the foundation overlay for
OSX.


dabrahams:
@swift-ci Please test

PR #1126, opened 2017-07-22
Determine the Timezone directory to use at runtime on macOS.

spevans:
- The timezone directory changed on macOS 10.13 so get the os version
  at runtime and set the timezone directory dynamically. This allows
  the tests to be on 10.12 and 10.13

- Added __InitTZStrings() for both Linux and macOS to work the same way as Windows.

Tested on Linux and 10.12 but not 10.13

### DIFF
--- a/CoreFoundation/NumberDate.subproj/CFTimeZone.c
+++ b/CoreFoundation/NumberDate.subproj/CFTimeZone.c
@@ -33,9 +33,12 @@
 #endif
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED
 #include <tzfile.h>
+#define MACOS_TZDIR1 "/usr/share/zoneinfo/"          // 10.12 and earlier
+#define MACOS_TZDIR2 "/var/db/timezone/zoneinfo/"    // 10.13 onwards
+
 #elif DEPLOYMENT_TARGET_LINUX || DEPLOYMENT_TARGET_FREEBSD
 #ifndef TZDIR
-#define TZDIR	"/usr/share/zoneinfo" /* Time zone object file directory */
+#define TZDIR	"/usr/share/zoneinfo/" /* Time zone object file directory */
 #endif /* !defined TZDIR */
 
 #ifndef TZDEFAULT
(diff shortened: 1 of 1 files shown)


CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -704,6 +696,50 @@ void __InitTZStrings(void) {
...
+                        minor = 0,
+                        patch = 0;
+                    }
+                }
+            }
+            CFRelease(dict);
+        }
+
+        // Timezone files moved in High Sierra(10.13)
+        if (major == 10 && minor >= 13) {
```
alblue: If major ever changes, this is going to provide the wrong behaviour here, surely? Wouldn't it be better to use `major == 10 && minor < 13` as the fallback condition, and then everything else as the new condition?

spevans: Yeah good catch, I will swap it round. Of course it will also break if the TZDIR is changed again - I was assuming any changes of that magnitude would be caught in the beta period :)

spevans:
@alblue I swapped the test around.

ianpartridge:
I confirm this PR works for me, and the tests now run on Sierra.  I don't have a High Sierra machine to test on, unfortunately.

The patch touches a fair amount of `CFTimeZone.c` including Windows-specific code so will need a careful review by @parkera.

alblue:
@swift-ci please test

spevans:
is there anyway to get the swift-ci bot to also test on macOS? Am I correct in thinking only Linux is tested?

ianpartridge:
AFAIK only Linux is tested by the CI.  I agree it would be nice to run the tests on macOS too, I don't know what would be needed to make that happen.

parkera:
We're talking to @shahmishal about getting macOS tests running here. I think it's important to get that started, probably after we wrap up this release.

ianpartridge:
Thanks @parkera. How does this PR look to you?

parkera:
Looks good to me.

ianpartridge:

ianpartridge:

ianpartridge:
Thanks for working on this @spevans 👍

parkera:
@spevans can we re-apply this to master after the re-apply of the import?

spevans:
There was also a 2nd PR to fix an off-by-one error that this one introduced. I will create a new PR that covers both and test it on 10.12 then that should solve it

spevans:
@swift-ci please test

PR #1127, opened 2017-07-22
CharacterSet : Codable implementation

bubski:
Continuation of #1108
This PR mirrors _some_ changes from apple/swift#9715

---

- CharacterSet
`CharacterSet : Equatable` was not working correctly in some cases and was fixed to make `Codable` unit tests for `CharacterSet` work.

### Not included in this PR, because of some problems that need to be solved

- Calendar
- DateComponents
- Measurement
- TimeZone

### Already implemented

- PersonNameComponents
- UUID
- URL
- NSRange
- Locale
- IndexSet
- IndexPath
- AffineTransform
- Decimal
- DateInterval
- CGFloat (unit tested indirectly)

### DIFF
--- a/Foundation/CharacterSet.swift
+++ b/Foundation/CharacterSet.swift
@@ -506,3 +506,20 @@ extension CharacterSet : _ObjectTypeBridgeable {
     }
     
 }
+
+extension CharacterSet : Codable {
+    private enum CodingKeys : Int, CodingKey {
+        case bitmap
+    }
+    
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let bitmap = try container.decode(Data.self, forKey: .bitmap)
+        self.init(bitmapRepresentation: bitmap)
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(self.bitmapRepresentation, forKey: .bitmap)
+    }
+}
--- a/Foundation/CharacterSet.swift
+++ b/Foundation/CharacterSet.swift
@@ -475,7 +475,7 @@ public struct CharacterSet : ReferenceConvertible, Equatable, Hashable, SetAlgeb
     
     /// Returns true if the two `CharacterSet`s are equal.
     public static func ==(lhs : CharacterSet, rhs: CharacterSet) -> Bool {
-        return lhs._mapUnmanaged { l in rhs._mapUnmanaged { r in l.isEqual(r) } }
+        return lhs._mapUnmanaged { $0.isEqual(rhs) }
     }
 }
 
--- a/Foundation/CharacterSet.swift
+++ b/Foundation/CharacterSet.swift
@@ -94,6 +94,12 @@ internal final class _SwiftNSCharacterSet : NSCharacterSet, _SwiftNativeFoundati
     override func isSuperset(of other: CharacterSet) -> Bool {
         return _mapUnmanaged { $0.isSuperset(of: other) }
     }
+
+    override var _cfObject: CFType {
+        // We cannot inherit super's unsafeBitCast(self, to: CFType.self) here, because layout of _SwiftNSCharacterSet
+        // is not compatible with CFCharacterSet. We need to bitcast the underlying NSCharacterSet instead.
+        return _mapUnmanaged { unsafeBitCast($0, to: CFType.self) }
+    }
 }
 
 /**
@@ -486,10 +492,7 @@ extension CharacterSet : _ObjectTypeBridgeable {
     
     @_semantics("convertToObjectiveC")
     public func _bridgeToObjectiveC() -> NSCharacterSet {
-        switch _wrapped.__wrapped {
-        case .Mutable(let wrapped): return wrapped.takeUnretainedValue()
-        case .Immutable(let wrapped): return wrapped.takeUnretainedValue()
-        }
+        return _wrapped
     }
     
     public static func _forceBridgeFromObjectiveC(_ input: NSCharacterSet, result: inout CharacterSet?) {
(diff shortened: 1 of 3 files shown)


Foundation/CharacterSet.swift
```diff
@@ -469,7 +469,7 @@ public struct CharacterSet : ReferenceConvertible, Equatable, Hashable, SetAlgeb
     
     /// Returns true if the two `CharacterSet`s are equal.
     public static func ==(lhs : CharacterSet, rhs: CharacterSet) -> Bool {
-        return lhs._wrapped.isEqual(rhs._bridgeToObjectiveC()) // TODO: mlehew - as  NSCharacterSet
+        return lhs._mapUnmanaged { l in rhs._mapUnmanaged { r in l.isEqual(r) } }
```
itaiferber: What changed here? Is there an effective difference? A comment would be helpful.

TestFoundation/TestNSCharacterSet.swift
```diff
@@ -17,7 +15,93 @@ import SwiftFoundation
...
+    
+    if equal {
+        XCTAssertEqual(lhs, rhs, message, file: file, line: line)
+    }
+    else {
+        XCTAssertNotEqual(lhs, rhs, message, file: file, line: line)
+    }
+}
 
+private func assert(equal: Bool,
```
itaiferber: If keeping the `Box` is helpful/necessary, then the following comment applies: All of these new asserts seem a bit out of place. Is it not sufficient to make the `Box` `Equatable` and perform these checks in `==`? (The unit tests can then continue using `XCTAssertEqual`.)

TestFoundation/TestNSCharacterSet.swift
```diff
@@ -17,7 +15,93 @@ import SwiftFoundation
 import SwiftXCTest
 #endif
 
+private struct Box {
```
itaiferber: Is this `Box` necessary? This doesn't seem to be significantly less work than just replicating some tests:

```swift
for (lhs, rhs) in pairs {
    XCTAssertEqual(NSCharacterSet(charactersIn: lhs), NSCharacterSet(charactersIn: rhs))
    XCTAssertEqual(CharacterSet(charactersIn: lhs), CharacterSet(charactersIn: rhs))
}
```

Foundation/CharacterSet.swift
bubski: I'll add a comment.

The problem was `rhs._bridgeToObjectiveC()` would always return an invalid (empty) CharacterSet.
`._mapUnmanaged` behaved correctly.

Now that I think about it, it can actually be changed to
```swift
return lhs._mapUnmanaged { $0.isEqual(rhs) }
```

PS
I actually fixed `._bridgeToObjectiveC()` in this PR too, so 
`lhs._wrapped.isEqual(rhs._bridgeToObjectiveC())` would start working correctly,
but I think `lhs._mapUnmanaged { $0.isEqual(rhs) }` follows better the pattern seen in other methods of `_SwiftNSCharacterSet`, like
```swift
    override func characterIsMember(_ member: unichar) -> Bool {
        return _mapUnmanaged { $0.characterIsMember(member) }
    }
    
    override func longCharacterIsMember(_ member: UInt32) -> Bool {
        return _mapUnmanaged { $0.longCharacterIsMember(member) }
    }
    
    override func isSuperset(of other: CharacterSet) -> Bool {
        return _mapUnmanaged { $0.isSuperset(of: other) }
    }
```

Does that make sense @itaiferber ?

TestFoundation/TestNSCharacterSet.swift
bubski: @itaiferber These are very good suggestions, thanks. I'll revisit this. And update this PR.

I wanted to make sure various different equality check combinations are covered, because I was running into inconsistent results in unexpected scenarios.

Foundation/CharacterSet.swift
itaiferber: Yep, reasonable, and makes sense. Comments both in here and in `_bridgeToObjectiveC()` explaining what makes this work (vs. doesn't) would be nice. Thank you! 🙂

Foundation/CharacterSet.swift
```diff
@@ -486,7 +486,10 @@ extension CharacterSet : _ObjectTypeBridgeable {
     
     @_semantics("convertToObjectiveC")
     public func _bridgeToObjectiveC() -> NSCharacterSet {
-        return _wrapped
+        switch _wrapped.__wrapped {
```
bubski: @phausler I'd like to run this against you and ask for advice.

I'm not sure if this solutionis ok.

Perhaps it makes more sense to keep `return _wrapped` here, and add
```swift
override var _cfObject: CFType {
    return _mapUnmanaged { unsafeBitCast($0, to: CFType.self) }
}
```
to `_SwiftNSCharacterSet`

The problem I'm trying to solve is that we return `_wrapped` here, which is `_SwiftNSCharacterSet`.
Somewhere later, we access `._cfObject` in this object, which does `unsafeBitCast(self, to: CFType.self)`.
But this doesn't work correctly, because `_SwiftNSCharacterSet` has different structure than `CFCharacterSet`.

ianpartridge: @parkera ^^^

itaiferber: @bubski The new solution sounds reasonable to me. Have you tested locally to confirm the expected behavior?

bubski: @itaiferber yes, I have 👌 seems to be working ok.

But I'm not that comfortable with the way copy-on-write is set up yet, so would like someone to double-check this please.

TestFoundation/TestNSCharacterSet.swift
bubski: @itaiferber I've done the tests refactor. Much cleaner 👍 
Please let me know if it looks ok to you now.

itaiferber:
@swift-ci Please test

ianpartridge:

PR #1128, opened 2017-07-23
Codable conformance added to some CG types

bubski:
This PR mirrors some changes from https://github.com/apple/swift/pull/10404

---

- CGPoint
- CGSize
- CGRect

### Not mirrored from https://github.com/apple/swift/pull/10404

- CGVector 📝
- CGAffineTransform 📝

📝 I've stumbled across [this](https://github.com/apple/swift-corelibs-foundation/pull/135), which leads me to believe that `CGVector` is not supposed to be implemented in `swift-corelibs-foundation`.
I suppose same goes for `CGAffineTransform`, since we already have `AffineTransform`/`NSAffineTransform` implemented.

### DIFF
--- a/TestFoundation/TestCodable.swift
+++ b/TestFoundation/TestCodable.swift
@@ -235,6 +235,21 @@ class TestCodable : XCTestCase {
             expectRoundTripEqualityThroughJSON(for: point)
         }
     }
+    
+    // MARK: - CGSize
+    lazy var cgsizeValues: [CGSize] = [
+        CGSize(),
+        CGSize(width: 30, height: 40),
+        CGSize(width: -30, height: -40),
+        // Disabled due to limit on magnitude in JSON. See SR-5346
+        // CGSize(width: .greatestFiniteMagnitude, height: .greatestFiniteMagnitude),
+    ]
+    
+    func test_CGSize_JSON() {
+        for size in cgsizeValues {
+            expectRoundTripEqualityThroughJSON(for: size)
+        }
+    }
 
 }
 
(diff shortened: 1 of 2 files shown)


TestFoundation/TestCodable.swift
```diff
@@ -220,6 +220,54 @@ class TestCodable : XCTestCase {
...
+    ]
+    
+    func test_CGSize_JSON() {
+        for size in cgsizeValues {
+            expectRoundTripEqualityThroughJSON(for: size)
+        }
+    }
+    
+    // MARK: - CGRect
+    lazy var cgrectValues: [CGRect] = [
```
ianpartridge: Can we also test `CGRect.zero` and `CGRect.null` like in https://github.com/apple/swift/pull/10404?

bubski: @ianpartridge These are not implemented in `swift-corelibs-foundation` yet, and I intend to add that in a separate PR.

ianpartridge: Ah I see, thanks. It looks like the CG types are missing quite a few properties etc. I didn't notice that because usually we stub them out with `NSUnimplemented()`.

bubski: @ianpartridge btw, let me ask something.
Do we expect to have
```swift
CGRect.init(x: Double, y: Double, width: Double, height: Double)
CGRect.init(x: Int, y: Int, width: Int, height: Int)
```
implemented too in `swift-corelibs-foundation`?

ianpartridge: I would say yes, but defer to @parkera?

parkera: We decided to provide some ultra-basic CG geometry type primitives in swift-corelibs-foundation just to help people with things like cross-platform archiving. So it seems appropriate to add the inits if we have to.

bubski: @parkera Wasn't fact that `Foundation` has `typealias NSRect = CGRect` and others the key reason?

PS. I'll add those inits in a seperate PR

itaiferber:
@swift-ci Please test

PR #1130, opened 2017-07-24
Simplify substring(with:), provide resilience

dabrahams:
This is in support of apple/swift#11088


dabrahams:

dabrahams:
@parkera @phausler Once apple/swift#11088 lands, you should consider using [this init](https://github.com/apple/swift/pull/11088/files#diff-87b54f29daf73bb22c88494da6b13e63R21) directly, which may be more efficient when there are no unpaired surrogates, at least on longer strings.

PR #1135, opened 2017-07-26
Special values and initializers for some CG types

bubski:
- [x] `CGPoint.zero`
- [x] `CGPoint.init(x: Int, y: Int)`
- [x] `CGPoint.init(x: Double, y: Double)`
- [x] `CGSize.zero`
- [x] `CGSize.init(width: Int, height: Int)`
- [x] `CGSize.init(width: Double, height: Double)`
- [x] `CGRect.zero`
- [x] `CGRect.init(x: CGFloat, y: CGFloat, width: CGFloat, height: CGFloat)`
- [x] `CGRect.init(x: Double, y: Double, width: Double, height: Double)`
- [x] `CGRect.init(x: Int, y: Int, width: Int, height: Int)`
- [x] `CGRect.null`
- [x] `CGRect.infinite`

### DIFF
--- a/Foundation/NSGeometry.swift
+++ b/Foundation/NSGeometry.swift
@@ -117,6 +117,20 @@ public struct CGSize {
     }
 }
 
+extension CGSize {
+    public static var zero: CGSize {
+        return CGSize(width: CGFloat(0), height: CGFloat(0))
+    }
+    
+    public init(width: Int, height: Int) {
+        self.init(width: CGFloat(width), height: CGFloat(height))
+    }
+    
+    public init(width: Double, height: Double) {
+        self.init(width: CGFloat(width), height: CGFloat(height))
+    }
+}
+
 extension CGSize: Equatable {
     public static func ==(lhs: CGSize, rhs: CGSize) -> Bool {
         return lhs.width == rhs.width && lhs.height == rhs.height
--- a/Foundation/NSGeometry.swift
+++ b/Foundation/NSGeometry.swift
@@ -209,6 +209,24 @@ public struct CGRect {
     }
 }
 
+extension CGRect {
+    public static var zero: CGRect {
+        return CGRect(origin: CGPoint(), size: CGSize())
+    }
+    
+    public init(x: CGFloat, y: CGFloat, width: CGFloat, height: CGFloat) {
+        self.init(origin: CGPoint(x: x, y: y), size: CGSize(width: width, height: height))
+    }
+    
+    public init(x: Double, y: Double, width: Double, height: Double) {
+        self.init(origin: CGPoint(x: x, y: y), size: CGSize(width: width, height: height))
+    }
+    
+    public init(x: Int, y: Int, width: Int, height: Int) {
+        self.init(origin: CGPoint(x: x, y: y), size: CGSize(width: width, height: height))
+    }
+}
+
 extension CGRect: Equatable {
     public static func ==(lhs: CGRect, rhs: CGRect) -> Bool {
         return lhs.origin == rhs.origin && lhs.size == rhs.size
--- a/Foundation/NSGeometry.swift
+++ b/Foundation/NSGeometry.swift
@@ -25,6 +25,20 @@ public struct CGPoint {
     }
 }
 
+extension CGPoint {
+    public static var zero: CGPoint {
+        return CGPoint(x: CGFloat(0), y: CGFloat(0))
+    }
+    
+    public init(x: Int, y: Int) {
+        self.init(x: CGFloat(x), y: CGFloat(y))
+    }
+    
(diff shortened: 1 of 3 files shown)


Foundation/NSGeometry.swift
```diff
@@ -25,6 +25,20 @@ public struct CGPoint {
     }
 }
 
+extension CGPoint {
```
ianpartridge: Why is this an extension? Is this a convention I'm not aware of?  I would have added to the `struct CGPoint {}` definition just above.

TestFoundation/TestCodable.swift
```diff
@@ -51,15 +51,17 @@ func expectRoundTripEquality<T : Codable>(of value: T, encode: (T) throws -> Dat
     XCTAssertEqual(value, decoded, "Decoded \(T.self) <\(decoded)> not equal to original <\(value)>")
 }
 
-func expectRoundTripEqualityThroughJSON<T : Codable>(for value: T) where T : Equatable {
+func expectRoundTripEqualityThroughJSON<T : Codable>(for value: T,
```
ianpartridge: This diverges the code in `swift-corelibs-foundation` from https://github.com/apple/swift/blob/b8dc826d41a25d856d6e827eb67cbfef9c4f9705/test/stdlib/CodableTests.swift#L60

Can we keep things as closely in sync as possible?

TestFoundation/TestCodable.swift
```diff
@@ -254,19 +258,31 @@ class TestCodable : XCTestCase {
...
-        //                     height: .greatestFiniteMagnitude)),
+        // CGRect.infinite
     ]
     
     func test_CGRect_JSON() {
         for rect in cgrectValues {
             expectRoundTripEqualityThroughJSON(for: rect)
         }
+        
+        do {
```
ianpartridge: Is the `do {}` block necessary? I can't see anything inside it which can throw.

bubski: `do {}` is not necessary and misleading. I'll remove it. 👍 

As for `CGRect.null` in `cgrectValues`, it requires `JSONEncoder` and `JSONDecoder` configured differently than for the other test values.
Also I don't think the other test values should share the configuration that is required by `CGRect.null`
So I test it separately from the rest.

Is this approach unreasonable? Would like to hear your opinion on that.

TestFoundation/TestCodable.swift
bubski: Because of reasons mentioned in https://github.com/apple/swift-corelibs-foundation/pull/1135#discussion_r129784775 I don't think it makes sense putting 
```swift
encoder.nonConformingFloatEncodingStrategy = .convertToString(positiveInfinity: inf,
                                                              negativeInfinity: negInf,
                                                              nan: nan)
decoder.nonConformingFloatDecodingStrategy = .convertFromString(positiveInfinity: inf,
                                                                negativeInfinity: negInf,
                                                                nan: nan)
```

inside `expectRoundTripEqualityThroughJSON`, forcing all checks, even the ones that don't require it, to use it.

wdyt?

Foundation/NSGeometry.swift
bubski: Ah I was just mimicing the way it's structured in `Foundation` on **macOS**.
It also made sense to me to keep this syntax-sugar-like initializers in an `extension`, just for the sake of visual cue.

These are the reasons I did it.

ianpartridge: OK, it's no big deal - I wondered if I could learn something here 🙂

TestFoundation/TestCodable.swift
ianpartridge: That is what the [stdlib tests](https://github.com/apple/swift/blob/b8dc826d41a25d856d6e827eb67cbfef9c4f9705/test/stdlib/CodableTests.swift#L60) do though? I think we should follow the pattern from those. As far as I'm aware there's no extra cost involved in doing this.

Keeping the tests in sync will make it easier in future, if/when new tests are added we can port them over here more easily.

TestFoundation/TestCodable.swift
ianpartridge: See my comment above about this - I'm just thinking from a future maintenance point of view.

bubski: That's a very valid point 👍

bubski:
@ianpartridge Thanks for pointing these out. I'll push the updates later 👍

bubski:
@ianpartridge PR updated ⬆️

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:

ianpartridge:

PR #1138, opened 2017-07-27
Use URLSessionTask.count* properties to track upload/download progress 

pushkarnk:
This is a kind of an extension to #1121 

We had defined three private properties in URLSessionTask to track upload & download progress and make delegate calls. It turns out that the URLSessionTask API defines [four properties](https://developer.apple.com/documentation/foundation/urlsessiontask/1410663-countofbytesexpectedtoreceive) to do the same. This PR replaces the uses of the private properties with these.

Also, the constant `URLSessionTransferSizeUnknown` is changed to `NSURLSessionTransferSizeUnknown` to match Darwin.

### DIFF
--- a/Foundation/NSURLSession/http/EasyHandle.swift
+++ b/Foundation/NSURLSession/http/EasyHandle.swift
@@ -55,7 +55,6 @@ internal final class _EasyHandle {
     weak var delegate: _EasyHandleDelegate?
     fileprivate var headerList: _CurlStringList?
     fileprivate var pauseState: _PauseState = []
-    internal var fileLength: Int64 = 0
     internal var timeoutTimer: _TimeoutSource!
     #if os(Android)
     static fileprivate var _CAInfoFile: UnsafeMutablePointer<Int8>?
@@ -100,7 +99,7 @@ internal protocol _EasyHandleDelegate: class {
     func didReceive(data: Data) -> _EasyHandle._Action
     /// Handle header data read from the network.
     /// - returns: the action to be taken: abort, proceed, or pause.
-    func didReceive(headerData data: Data) -> _EasyHandle._Action
+    func didReceive(headerData data: Data, contentLength: Int64) -> _EasyHandle._Action
     /// Fill a buffer with data to be sent.
     ///
     /// - parameter data: The buffer to fill
@@ -454,7 +453,7 @@ fileprivate extension _EasyHandle {
             }
             var length = Double()
             try! CFURLSession_easy_getinfo_double(handle.rawHandle, CFURLSessionInfoCONTENT_LENGTH_DOWNLOAD, &length).asError()
-            return handle.didReceive(headerData: data, size: size, nmemb: nmemb, fileLength: length)
+            return handle.didReceive(headerData: data, size: size, nmemb: nmemb, contentLength: length)
         }.asError()
 
         // socket options
@@ -513,11 +512,10 @@ fileprivate extension _EasyHandle {
     /// data.
     ///
     /// - SeeAlso: <https://curl.haxx.se/libcurl/c/CURLOPT_HEADERFUNCTION.html>
-    func didReceive(headerData data: UnsafeMutablePointer<Int8>, size: Int, nmemb: Int, fileLength: Double) -> Int {
-        self.fileLength = Int64(fileLength)
+    func didReceive(headerData data: UnsafeMutablePointer<Int8>, size: Int, nmemb: Int, contentLength: Double) -> Int {
         let d: Int = {
             let buffer = Data(bytes: data, count: size*nmemb)
-            switch delegate?.didReceive(headerData: buffer) {
+            switch delegate?.didReceive(headerData: buffer, contentLength: Int64(contentLength)) {
             case .some(.proceed): return size * nmemb
             case .some(.abort): return 0
             case .some(.pause):
--- a/Foundation/NSURLSession/http/HTTPURLProtocol.swift
+++ b/Foundation/NSURLSession/http/HTTPURLProtocol.swift
@@ -13,9 +13,6 @@ import Dispatch
 internal class _HTTPURLProtocol: URLProtocol {
 
     fileprivate var easyHandle: _EasyHandle!
-    fileprivate var totalDownloaded = 0
-    fileprivate var totalUploaded: Int64 = 0
-    fileprivate var requestBodyLength: Int64 = 0
     fileprivate var tempFileURL: URL
 
     public required init(task: URLSessionTask, cachedResponse: CachedURLResponse?, client: URLProtocolClient?) {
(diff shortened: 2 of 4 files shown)


pushkarnk:
cc @saihemak

Foundation/NSURLSession/http/EasyHandle.swift
```diff
@@ -454,7 +453,7 @@ fileprivate extension _EasyHandle {
             }
             var length = Double()
             try! CFURLSession_easy_getinfo_double(handle.rawHandle, CFURLSessionInfoCONTENT_LENGTH_DOWNLOAD, &length).asError()
```
ianpartridge: Just an aside... This API returns the ContentLength as a `Double`, which seems crazy. There is a [newer API](https://curl.haxx.se/libcurl/c/CURLINFO_CONTENT_LENGTH_DOWNLOAD_T.html) which has a better return type. It was introduced in libcurl 7.55.0; could we switch to it instead?

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -469,21 +466,22 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
...
                 fileHandle.closeFile()
                 self.properties[.temporaryFileURL] = self.tempFileURL
             }
         }
     }
 
-    func didReceive(headerData data: Data) -> _EasyHandle._Action {
+    func didReceive(headerData data: Data, contentLength: Int64) -> _EasyHandle._Action {
         guard case .transferInProgress(let ts) = internalState else { fatalError("Received body data, but no transfer in progress.") }
+        task!.countOfBytesExpectedToReceive = contentLength > 0 ? contentLength : NSURLSessionTransferSizeUnknown
```
ianpartridge: I would be more defensive:

```swift
guard let task = task else { fatalError("No task") }
```

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -469,21 +466,22 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
             let fileHandle = try! FileHandle(forWritingTo: self.tempFileURL)
             _ = fileHandle.seekToEndOfFile()
             fileHandle.write(data)
-            self.totalDownloaded += data.count
+            task.countOfBytesReceived += Int64(data.count)
             
             s.delegateQueue.addOperation {
-                downloadDelegate.urlSession(s, downloadTask: task, didWriteData: Int64(data.count), totalBytesWritten: Int64(self.totalDownloaded),
-                                            totalBytesExpectedToWrite: Int64(self.easyHandle.fileLength))
+                downloadDelegate.urlSession(s, downloadTask: task, didWriteData: Int64(data.count), totalBytesWritten: Int64(task.countOfBytesReceived),
```
ianpartridge: Isn't `task.countOfBytesReceived` an `Int64` already?

Foundation/NSURLSession/http/EasyHandle.swift
pushkarnk: I guess Ubuntu 16.04  comes with a [default libcurl 7.47](https://launchpad.net/ubuntu/xenial/+source/curl). I am not sure if we should be using an API that is newer than 7.47.

Foundation/NSURLSession/http/HTTPURLProtocol.swift
pushkarnk: Yes! Thanks for catching that.

Foundation/NSURLSession/http/EasyHandle.swift
ianpartridge: I didn't realise that we rely on the system libcurl. Ubuntu 14.04 LTS ships 7.35 which is even older 😢

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -469,21 +466,23 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
...
                 fileHandle.closeFile()
                 self.properties[.temporaryFileURL] = self.tempFileURL
             }
         }
     }
 
-    func didReceive(headerData data: Data) -> _EasyHandle._Action {
+    func didReceive(headerData data: Data, contentLength: Int64) -> _EasyHandle._Action {
         guard case .transferInProgress(let ts) = internalState else { fatalError("Received body data, but no transfer in progress.") }
+        guard let task = task else { fatalError("Receieved header data but no task available") }
```
ianpartridge: Typo: Received.

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -469,21 +466,23 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
...
+            if task.countOfBytesExpectedToReceive == task.countOfBytesReceived {
                 fileHandle.closeFile()
                 self.properties[.temporaryFileURL] = self.tempFileURL
             }
         }
     }
 
-    func didReceive(headerData data: Data) -> _EasyHandle._Action {
+    func didReceive(headerData data: Data, contentLength: Int64) -> _EasyHandle._Action {
         guard case .transferInProgress(let ts) = internalState else { fatalError("Received body data, but no transfer in progress.") }
```
ianpartridge: Should this say "Received header data"?

Foundation/NSURLSession/http/HTTPURLProtocol.swift
ianpartridge: Also, add a full stop?

Foundation/NSURLSession/http/HTTPURLProtocol.swift
```diff
@@ -501,10 +500,10 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
     fileprivate func notifyDelegate(aboutUploadedData count: Int64) {
         let session = self.task?.session as! URLSession
```
ianpartridge: How about:

```swift
fileprivate func notifyDelegate(aboutUploadedData count: Int64) {
    guard let task = self.task as? URLSessionUploadTask,
          let session = task.session as? URLSession,
          case .taskDelegate(let delegate) = session.behaviour(for: task) else { return }
    task.countOfBytesSent += count
    session.delegateQueue.addOperation {
        delegate.urlSession(session, task: task, didSendBodyData: count,
                            totalBytesSent: task.countOfBytesSent,
                            totalBytesExpectedToSend: task.countOfBytesExpectedToSend)
    }
}

(not compiled)

pushkarnk:
@ianpartridge Agree with your suggestions. I've included them. Thanks!

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

PR #1140, opened 2017-07-28
Add availability indicator to XMLElement to indicate correct function

alblue:
The implementation of `setAttributesAs` was deprecated bfeore the API was
copied over to the Linux implementation, and won't be implemented. By
adding an availability marker coupled with a new name, we allow for a
fix-it to run and use the new name in future compiled code.


alblue:
@swift-ci please test

alblue:
See also https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20170724/001254.html

PR #1143, opened 2017-07-31
Replace precondition if with guard in NSNumber.encode.

oc243:
Previously NSNumber.encode would call NSUnimplemented when you try to
do a keyed coding. Whilst it is true that this wasn't implemented it was
due to not being possible rather than not yet implemented.

With this patch we call preconditionFailure which is more consistent
with the rest of the project. We also use a guard so can remove a level of
indentation.

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -609,35 +609,34 @@ open class NSNumber : NSValue {
     }
     
     open override func encode(with aCoder: NSCoder) {
-        if !aCoder.allowsKeyedCoding {
-            NSUnimplemented()
+        guard aCoder.allowsKeyedCoding else {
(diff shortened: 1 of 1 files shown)


alblue:
@swift-ci please test

alblue:
LGTM. What do you think @ianpartridge @phausler ?

ianpartridge:
Yes I think this is fine. Thanks @oc243

ianpartridge:

alblue:
Looks like an unrelated segfault:

`/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift/utils/build-script-impl: line 261:  9831 Segmentation fault      (core dumped) "$@"`

PR #1145, opened 2017-07-31
Create StringMap.swift

omarshariffathi:
This is a fast way to search and match strings from a String Array

This implementation is mainly improving the time taken to match a string from a string array. This isn't a complex or long implementation like other implementations, yet drastically reduces the time taken to do such searching and matching tasks. 

This is a faster implementation for String `hasSuffix` & `hasPrefix` & == but for large String arrays.
This is a short yet efficient and to the point implementation. No concurrency.

### DIFF
--- a/Foundation/StringMap.swift
+++ b/Foundation/StringMap.swift
@@ -1,0 +1,527 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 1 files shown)


parkera:
Hi @omarshariffathi. In general, we are not adding new API to swift-corelibs-foundation but instead focusing on finishing up the implementation of the APIs we already have. We can discuss more on the mailing list if you like, but for now I'll go ahead and close this. Thanks.

parkera:
This functionality seems like a great candidate of something to ship via the Swift package manager, actually.

omarshariffathi:
Yea so I should first ship it via SPM, I have a copy live on Cocoapods now but I will make it available for SPM as well, thanks @parkera

o-micron:
This is a quick update, I have create a SPM for StringMap here https://github.com/o-micron/StringMap,
and here is a quick gist showing the huge performance gain using StringMap for full matching Strings compared to google's RE2.

PR #1147, opened 2017-07-31
SR-5591: Update broken DateFormatter tests

spevans:
- The tests were broken for timestyle .full and failed on macOS.
  Fix the tests so that they work correctly on macOS.

- DateFormatter is still broken on Linux so disable the tests for
  now until the Linux implementation can be fixed.


phausler:

phausler:
btw I think this will fix some of the failures I was having on my branch I was working on for Linux

PR #1151, opened 2017-08-01
[SR-1702] NSJSONSerialization.data(...) fatalErrors on bad input

glenna:
[SR-1702](https://bugs.swift.org/browse/SR-1702) NSJSONSerialization.data(withJSONObject:options:) should fatalError on bad input not throw

-removed the `throw` in the case of using a stream: in both cases in objc it is an `NSInvalidArgumentException`

-removed the test checking for the `throw` in the case of the stream (and, AFAIK, there's no good way to test for fatalErrors, but happy to be corrected on that fact)

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -139,13 +139,7 @@ open class JSONSerialization : NSObject {
         } else if let container = value as? Dictionary<AnyHashable, Any> {
             try writer.serializeJSON(container)
         } else {
-            if stream {
-                throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: [
-                    "NSDebugDescription" : "Top-level object was not NSArray or NSDictionary"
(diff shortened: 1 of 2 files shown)


hartbit:
Its a pity that we loose the unit-test :-/ Is there anything we can do about it?

glenna:
@phausler can you take a look? also any guidance about the tests?

phausler:
Sadly we do not yet have a way to validate fatal errors in XCTest. Shy of perhaps launching a NSTask to spawn the test in a sub-process.

Overall this looks like a matching behavior to the objc counterpart.

phausler:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

glenna:
seems just the build timed out @ianpartridge

phausler:

alblue:

ianpartridge:

PR #1152, opened 2017-08-01
Use pkg-config to find libxml2 and libcurl on Linux [SR-5603]

kevints:
This allows linking swift-corelibs-foundation to packages with flags
defined by `pkg-config`, rather than hardcoded ones. There are a couple
of structural changes as well, namely that `pkg-config` will now be
executed during `configure` rather than later by `ninja`. This seems
acceptable as it's how cmake and autoconf work, but means a reconfigure
will be required if any transitive dependency of a package we depend on changes.

The other structural change is that `-lcurl` will be passed
unconditionally on platforms without pkg-config support (technically the
`pkg_config_required` variable set). This makes curl a harder dependency
on those platforms, but that seems unavoidable given that it's the only
implementation of `URLSession`.

### DIFF
--- a/build.py
+++ b/build.py
@@ -15,8 +15,8 @@ foundation.GCC_PREFIX_HEADER = 'CoreFoundation/Base.subproj/CoreFoundation_Prefi
 
 swift_cflags = ['-DDEPLOYMENT_RUNTIME_SWIFT']
 if Configuration.current.target.sdk == OSType.Linux:
-	foundation.CFLAGS = '`${PKG_CONFIG} icu-uc icu-i18n --cflags-only-I` -DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE -DCF_CHARACTERSET_DATA_DIR="CoreFoundation/CharacterSets"'
-	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc `${PKG_CONFIG} icu-uc icu-i18n --libs` -Wl,-Bsymbolic '
+	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE -DCF_CHARACTERSET_DATA_DIR="CoreFoundation/CharacterSets"'
+	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc -Wl,-Bsymbolic '
 	Configuration.current.requires_pkg_config = True
 elif Configuration.current.target.sdk == OSType.FreeBSD:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_FREEBSD -I/usr/local/include -I/usr/local/include/libxml2 -I/usr/local/include/curl '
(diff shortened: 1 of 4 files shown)


kevints:
@parkera can you kick off CI for this change?

build.py
```diff
@@ -61,35 +61,55 @@
...
-		'-I${SYSROOT}/usr/include/libxml2',
-		'-I${SYSROOT}/usr/include/curl'
 	]
 
-triple = Configuration.current.target.triple
-if triple.find("linux") != -1:
-	foundation.LDFLAGS += '-lcurl '
+if Configuration.current.requires_pkg_config:
+    pkg_config_dependencies = [
+        'icu-i18n',
```
kevints: These should probably specify a minimum package version (probably from Ubuntu 14.04 as it's the oldest supported target), WDYT?

build.py
```diff
@@ -61,35 +61,55 @@
...
+        try:
+            package = PkgConfig(package_name)
+        except PkgConfig.Error as e:
+            sys.exit("pkg-config error for package {}: {}".format(package_name, e))
+        foundation.CFLAGS += ' {} '.format(' '.join(package.cflags))
+        foundation.LDFLAGS += ' {} '.format(' '.join(package.ldflags))
+        swift_cflags += package.swiftc_flags
+else:
+	foundation.CFLAGS += ''.join([
+		'-I${SYSROOT}/usr/include/curl ',
```
kevints: I mechanically moved this from the previous unconditional block; however I'm not sure it ever did anything given that CF headers include `<curl/curl.h>` and this directory appears to contain `curl.h` and friends, without subdirectories. It also seems to have been repeated at least 4 times - I've condensed 3 of those into this line but I'm unsure about the 4th in the FreeBSD config (the [pkg-plist|https://github.com/freebsd/freebsd-ports/blob/master/ftp/curl/pkg-plist#L3-L12] seems to suggest that it wouldn't be needed there either but I don't have a system to test on).

build.py
```diff
@@ -61,35 +61,55 @@
...
+    for package_name in pkg_config_dependencies:
+        try:
+            package = PkgConfig(package_name)
+        except PkgConfig.Error as e:
+            sys.exit("pkg-config error for package {}: {}".format(package_name, e))
+        foundation.CFLAGS += ' {} '.format(' '.join(package.cflags))
+        foundation.LDFLAGS += ' {} '.format(' '.join(package.ldflags))
+        swift_cflags += package.swiftc_flags
+else:
+	foundation.CFLAGS += ''.join([
```
ianpartridge: Why don't these get added via `package.cflags` on line 89 above?

alblue:
@swift-ci please test

build.py
ianpartridge: There's no harm in doing that, I guess. Currently we link against the system libicu which has plusses and minuses.

ianpartridge:
https://bugs.swift.org/browse/SR-5608 doesn't exist?

kevints:
@ianpartridge looks like the issue key is actually SR-5603, updated the title

ianpartridge:
Thanks. I really like the look of this PR - thanks for working on it 👍

build.py
kevints: This is the else block of `Configuration.current.requires_pkg_config`, so these are hardcoded guess paths (moved from being unconditionally included on line 64-65 and other places). When we have `pkg-config` there's no reason to include them as they'll be included automatically by `pkg-config <package_name> --libs` and `pkg-config <package_name> --cflags` respectively. More importantly, if the `PKG_CONFIG_PATH` environment variable points to somewhere other than the system libraries and headers for these packages we won't consider them and will instead use the intended paths.

ianpartridge: Got it, thanks.

ianpartridge:
@spevans how does this look to you?

spevans:
Overall this looks good. Being able to specify pkg-config for the other libraries will be useful if we need to build new versions of them.

The only thing to check is that an external ICU still builds correctly.

@kevints Could you run a local test build of swift and swift-corelibs-foundation with the `--libicu --install-libicu` options? 

You will need to download ICU57.1 http://site.icu-project.org/download/57#TOC-ICU4C-Download  (I think 59.1 has some issues)

Then untar it so that you have an `icu` directory at the same level as the other `swift-corelibs-*` directories. Then build swift with the icu options and check the tests work.

kevints:
@spevans hmm, tests don't seem to pass against my local build with `--libicu` and `--install-libicu` due to 

```
...swiftpm-linux-x86_64/x86_64-unknown-linux/release/swift-build-stage1: error while loading shared libraries: libicui18n.so.57: cannot open shared object file: No such file or directory
```

having a look at ldd it looks like both my system libicu (version 52) and the built icu (version 57) are both getting linked in but the 57 one is not found (since it isn't copied to the library runpath which is $ORIGIN/../lib/swift/linux, this might be a race between install and test invocation).

```
	libicuuc.so.52 => /usr/lib/x86_64-linux-gnu/libicuuc.so.52 (0x00007f3c9ca35000)
	libicui18n.so.52 => /usr/lib/x86_64-linux-gnu/libicui18n.so.52 (0x00007f3c9c62d000)
	/lib64/ld-linux-x86-64.so.2 (0x0000562e03598000)
	libicui18n.so.57 => not found
	libicuuc.so.57 => not found
	libicudata.so.57 => not found
```

This seems in line with what I'd expect having read through the foundation build - it unconditionally invokes pkg-config for icu, which will find the system one. This change doesn't change anything about that so I suspect this may have been broken for a while, if it ever worked at all (looking through the buildbot presets there doesn't appear to be CI for building on Linux with `--libicu --install-libicu`). I'll file a bug but I don't think the breakage is related to this change.

build.py
```diff
@@ -61,35 +61,55 @@
...
+	])
 
+triple = Configuration.current.target.triple
 if triple == "armv7-none-linux-androideabi":
 	foundation.LDFLAGS += '-llog '
 else:
 	foundation.LDFLAGS += '-lpthread '
 
-foundation.LDFLAGS += '-ldl -lm -lswiftCore -lxml2 '
+foundation.LDFLAGS += '-ldl -lm -lswiftCore '
```
kevints: These are the flags that get passed to the build for the CoreFoundation C library, which includes `<math.h>` and `<dlfcn.h>` so I believe `-lm` and `-ldl` should remain here in the spirit of include-what-you-use. I'm not sure about `-lswiftCore` as CoreFoundation doesn't appear to be importing any swift headers and swiftc should indeed be passing that flag, but maybe there is some magic I don't understand. I'll file a bug to follow up as I'd like to keep this change focused on the PR title.

kevints:
@spevans filed:

https://bugs.swift.org/browse/SR-5620
https://bugs.swift.org/browse/SR-5619
https://bugs.swift.org/browse/SR-5618

spevans:
@kevints  No worries, I'll take a look at the ICU stuff as I did it originally. What version of linux are you using?

build.py
spevans: `-lswiftCore` is actually added by the compiler for any swift program that is compiled. `-ldl` is a dependancy of `libswiftCore.so` so it gets picked up there. Math functions are also used by the stdlib and I think `-lm` is auto added by the C compiler (the same a `-lc`) since it is dependant on how a given C library is structured, ie if certain functions are broken out into other libraries

kevints:
@spevans saw those issues on Ubuntu 14.04

kevints:
@spevans modulo those ICU issues does this change look good to you otherwise?

spevans:
@kevints Yes this change looks ok to me. Im taking a look at the ICU issues but I'll deal with that separately - I think its because libxml2 is also dependant on ICU hence you seeing 2 versions.

I'll also look at cleaning up the linker flags, again that can be done as a separate PR.

spevans:
@kevints I tested this PR whilst also building an external ICU and it was ok. I think the problem you saw may have been due to old config in the build directory - I did a clean build/test and it was ok.

The 2 versions is because ICU is built with versioned symbols and libxml2 uses the system ICU, that will need to be solved in the future.

ianpartridge:

ianpartridge:

PR #1154, opened 2017-08-02
Use String.decodeCString() instead of bitcasting.

ianpartridge:
Use String.decodeCString() instead of bitcasting. Fixes a warning.


ianpartridge:
@swift-ci please test

ianpartridge:

PR #1158, opened 2017-08-03
Mark Bundle.unload as unsupported

alblue:
Since NSBundle's unload mechanism is a Mach-specific concept, which
is not relevant on non-Darwin platforms, mark it as deprecated
and change the unavailability macro so that they are distinguishable
from implementations that are not yet available.

### DIFF
--- a/Foundation/Bundle.swift
+++ b/Foundation/Bundle.swift
@@ -77,7 +77,8 @@ open class Bundle: NSObject {
     open var isLoaded: Bool {
         return CFBundleIsExecutableLoaded(_bundle)
     }
-    open func unload() -> Bool { NSUnimplemented() }
+    @available(*,deprecated,message:"Not available on non-Darwin platforms")
+    open func unload() -> Bool { NSUnsupported() }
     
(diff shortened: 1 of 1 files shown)


alblue:
In pull 857, @parkera mentioned that `Bundle.unload` was never going to be implementable on non-Darwin platforms. We can use this marker and replace the existing `NSUnimplemented` to reduce the noise of which parts of foundation are awaiting implementation vs. those which won't be.

It also has a deprecated availability marker to result in a warning (not a compile error) if this is called.

alblue:
@swift-ci please test

ianpartridge:

alblue:

alblue:
Erm ... @shahmishal @ianpartridge should this have been merged by the CI bot if the build failed?

shahmishal:
The build passed but status did not get updated https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/11/

alblue:
OK thanks for confirming 👍🏻

PR #1157, opened 2017-08-03
Fix C function prototypes for functions that take no arguments.

spevans:
- Update prototype() to prototype(void).

Xcode9/Clang includes new warnings for C function prototypes which are declared with an unspecified number of arguments `()` instead of with `(void)`. Fixing these warnings will allow updating the Xcode project to the latest recommended settings.

I broke this into 2 PRs for easier review, a 2nd PR will follow to fixup the rest.

PR #1159, opened 2017-08-04
Fix more C warnings

spevans:
- Merge function prototypes with the declaration that follows
  on the next line to simplify declaration when the arguments
  differ due to preprocessor arguments.

- Fix use of comma in while statement.

This is the second PR of C fixes along with https://github.com/apple/swift-corelibs-foundation/pull/1157 after which the Xcode project can be updated to recommended settings to include the latest Clang warnings.

The `while` statement definitely needs checking, I went with `&&` instead of `||` but a 2nd opinion would be good.

### DIFF
--- a/CoreFoundation/String.subproj/CFString.c
+++ b/CoreFoundation/String.subproj/CFString.c
@@ -3233,7 +3233,7 @@ Boolean CFStringFindWithOptionsAndLocale(CFStringRef string, CFStringRef stringT
 
                                     do {
                                         str1Char = CFStringGetCharacterFromInlineBuffer(&inlineBuf1, --index);
-                                    } while (CFUniCharIsMemberOfBitmap(str1Char, graphemeBMP), (rangeToSearch.location < index));
+                                    } while (CFUniCharIsMemberOfBitmap(str1Char, graphemeBMP) && (rangeToSearch.location < index));
 
                                     if (str1Char < 0x0510) {
                                         while (++str1Index < maxStr1Index) if (!CFUniCharIsMemberOfBitmap(CFStringGetCharacterFromInlineBuffer(&inlineBuf1, str1Index), graphemeBMP)) break;
--- a/CoreFoundation/Base.subproj/CFUtilities.c
+++ b/CoreFoundation/Base.subproj/CFUtilities.c
@@ -1519,7 +1519,7 @@ CFDictionaryRef __CFGetEnvironment() {
     static CFMutableDictionaryRef envDict = NULL;
     dispatch_once(&once, ^{
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED
-        extern char ***_NSGetEnviron();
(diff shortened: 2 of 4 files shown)


ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

parkera:
We're in the process of a giant merge of CF stuff right now, so let's hold off on this for a bit.

CoreFoundation/String.subproj/CFString.c
```diff
@@ -3233,7 +3233,7 @@ Boolean CFStringFindWithOptionsAndLocale(CFStringRef string, CFStringRef stringT
 
                                     do {
                                         str1Char = CFStringGetCharacterFromInlineBuffer(&inlineBuf1, --index);
-                                    } while (CFUniCharIsMemberOfBitmap(str1Char, graphemeBMP), (rangeToSearch.location < index));
+                                    } while (CFUniCharIsMemberOfBitmap(str1Char, graphemeBMP) && (rangeToSearch.location < index));
```
phausler: oof! how was the previous even valid C?

spevans: I had to look it up! The first call is always executed then the result of the function after the , is evaluated for the while loop. I assume && is the correct operator here.

phausler: doing some testing here; this is actually a bug btw... the first function is called but the result is discarded, if you mark the CFUniCharIsMemberOfBitmap as `__attribute__((pure))` or better yet `__attribute__((const))` (which it is most appropriately delineated as) it ends up emitting a warning that the result is not used.

As Tony mentioned we are in the process of merging a bunch of CF things - I will keep an eye on this particular portion; file a bug on the owner and get back to you on this. We should probably fix it somehow but I need to validate exactly the right fix.

weissi: @spevans / @phausler I think the original code is correct. The comma operator in C evaluates the first expression, discards the result and then evaluates the second and the overall expression evaluates to that (the value of the second expr).

weissi: oh, sorry, might be wrong. If `CFUniCharIsMemberOfBitmap(str1Char, graphemeBMP)` doesn't have any side-effects, the original code looks wrong.

spevans: FYI All I did in Xcode9 was 'Update to recommended settings'  which enabled the latest clang warnings so I was just fixing those up.

alblue:
@parkera the merging has happened now, can @spevans rebase and try this again?

spevans:
@alblue I don't think the merge has happened yet, I thought it was going to be the public bits of 10.13 but I don't see anything in the commit log.

alblue:
I vote we merge these changes today - they're just minor cleanups. @parkera @phausler any objections?

alblue:

PR #1162, opened 2017-08-07
Thread: Implement more functionality

spevans:
- Eliminate _compiler_crash_fix() as no crashes are observed anymore.

- Re-enable TestThread.swift test cases.

I exposed `_CFMainPThread` since `pthread_main_thread_np()` is a `#define` and not a real function on linux and so doesn't export to swift correctly.

I used `_CFIsMainThread` instead of `pthread_main_np` for consistency and to avoid the local definition of `pthread_main_np()` in `Operation.swift` not having the same semantics as `pthread_main_np()` on macOS where it can return -1.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -180,6 +180,12 @@ const char *_CFProcessPath(void) {
     }
     return __CFProcessPath;
 }
+
+#else
+
+Boolean _CFIsMainThread(void) {
+    return pthread_main_np() == 1;
+}
 #endif
 
 CF_PRIVATE CFStringRef _CFProcessNameString(void) {
--- a/TestFoundation/TestThread.swift
+++ b/TestFoundation/TestThread.swift
@@ -24,6 +24,9 @@ class TestThread : XCTestCase {
             ("test_currentThread", test_currentThread ),
             ("test_threadStart", test_threadStart),
             ("test_threadName", test_threadName),
+            ("test_mainThread", test_mainThread),
+            ("test_callStackSymbols", test_callStackSymbols),
+            ("test_callStackReurnAddresses", test_callStackReturnAddresses),
         ]
     }
 
@@ -33,6 +36,7 @@ class TestThread : XCTestCase {
         XCTAssertNotNil(thread1)
         XCTAssertNotNil(thread2)
         XCTAssertEqual(thread1, thread2)
+        XCTAssertEqual(thread1, Thread.mainThread)
     }
     
     func test_threadStart() {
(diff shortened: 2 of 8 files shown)


ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

spevans:
Looks like ssh to github.com failed

ianpartridge:
Yeah, timeout.  I'll give it another try.

ianpartridge:
@swift-ci please test

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -180,6 +180,12 @@ const char *_CFProcessPath(void) {
     }
     return __CFProcessPath;
 }
+
+#else
+
+Boolean _CFIsMainThread(void) {
+    return pthread_main_np() == 1;
+}
```
parkera: This is meant to be a Darwin implementation then?

parkera: I'm wondering if we should make that `#else` more platform specific. Other than that, this looks great to me.

spevans: I wasnt sure what other platforms support it, I thought FreeBSD might have it as well. Thought it would be easier to leave it open and then a version of `_CFIsMainThread` can be added to platforms as they need it.

spevans: It also works for Windows since there is a `#define pthread_main_np _NS_pthread_main_np` wrapped inside a `#if DEPLOYMENT_TARGET_WINDOWS`

parkera: Ok, I guess we'll hear back from people if this breaks their compile.

ianpartridge:
Can you update the `Thread` status in https://github.com/apple/swift-corelibs-foundation/blob/master/Docs/Status.md too?

spevans:
Done, although I think some more test coverage will be needed at some point.

TestFoundation/TestThread.swift
```diff
@@ -84,4 +88,44 @@ class TestThread : XCTestCase {
...
+            started = true
+            XCTAssertFalse(Thread.isMainThread)
+            XCTAssertFalse(Thread.mainThread == Thread.current)
+            condition.broadcast()
+            condition.unlock()
+        }
+        thread.start()
+
+        condition.lock()
+        if !started {
```
phausler: isn't this a race on started? what if the thread takes a while to start up and the condition lock here is locked, then started will be false and the condition wont wait.

phausler: also this could potentially be waiting forever (if there was a problem with Thread), perhaps a semaphore with a timeout might be a good idea?

spevans: I actually just copied the `test_threadStart()` function. You are correct that with a delay starting up the thread, `!started` might be `false` and the `.wait()` will never happen.

There is a `pthread_cond_timedwait()` function to timeout a `.wait()` so I will see if I can find a way to incorporate that to handle the thread never calling `.broadcast()`.

I notice that `NSLock` also has 2 time bounded lock functions that need implementing so I might see if I can implement and incorporate them.

spevans: Looks like `NSCondition wait(until limit: Date)` (using `pthread_cond_timedwait()`) has already been implemented so I will update to use that.

spevans: I removed the `started` flag and used `wait(until:)` (including fixing a bug with the implementation) and then tried some tests including not starting the thread or making the thread take too long, and the timeout worked fine under those conditions.

I gave it 10seconds timeout which should cover all cases except incredibly high server load and if thats the case -CI will have other problems anyway

spevans:
@phausler Does this look ok now?

phausler:
@swift-ci please test

spevans:
@swift-ci please test

ianpartridge:
@swift-ci please test

PR #1164, opened 2017-08-09
SR-5557: Fix NSRegularExpression errors with invalid pattern

spevans:
- The error dictionary in CFRegularExpression was failing with the
  error key being a CFSTR inside an array so initialiase it outside
  of the array.

I removed the `_nsObject` as I think throwing the error dictionary is more useful as it gave a better error message when stringified.

Should a description be added to the error dictionary to eliminate the '(null)'?

### DIFF
--- a/CoreFoundation/String.subproj/CFRegularExpression.c
+++ b/CoreFoundation/String.subproj/CFRegularExpression.c
@@ -133,8 +133,9 @@ _CFRegularExpressionRef _CFRegularExpressionCreate(CFAllocatorRef allocator, CFS
     if (regex == NULL || U_FAILURE(errorCode)) {
         // ??? do we need more detailed errors here?
         if (errorPtr) {
-            CFStringRef keys[] = {
-                CFSTR("NSInvalidValue")
+            CFStringRef key = CFSTR("NSInvalidValue");
+            CFTypeRef keys[] = {
+                key
             };
             CFTypeRef values[] = {
                 pattern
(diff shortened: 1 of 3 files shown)


ianpartridge:
I probably haven't had enough coffee but I don't quite follow this one... Why does it make a difference whether it's an array of `CFStringRef` or a `CFStringRef` inside an array of `CFTypeRef`?

spevans:
When `CFSTR("NSInvalidValue")` was declared directly inside the array it would crash `CFHash`, which is the actual fix.

I don't think the `CFStringRef` v `CFTypeRef` makes much difference here except that `CFTypeRef` seems to be the correct type for a key based on other code I looked at.

I think we will need either @parkera or @phausler to validate this fix.

ianpartridge:
Yes you're right. I imagine this code is unchanged from Darwin so maybe there's some other factors involved.

spevans:
Yes, although it did actually crash on macOS as well, which makes even less sense!

phausler:
The problem previously with the key was because CFSTR in swift-corelibs does not use the builtin but instead uses an inline address of. That hack won't work with c arrays.

Overall this change looks reasonable to me.

spevans:
Thats good to know. I did a quick audit of CSTR in CoreFoundation and noticed a few places where it is used in arrays so I'll add a PR soon to fix those cases

ianpartridge:

ianpartridge:

spevans:
@ianpartridge This looks to have failed due to an unrelated swift test failing, can you retry?

ianpartridge:

PR #1166, opened 2017-08-10
SR-4993: Read zero length files on Linux

spevans:
- Some files (eg in /proc) report a filesize of zero and need to
  be read into a buffer instead of using mmap().

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -146,6 +146,7 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     public convenience init(bytesNoCopy bytes: UnsafeMutableRawPointer, length: Int, deallocator: ((UnsafeMutableRawPointer, Int) -> Void)? = nil) {
         self.init(bytes: bytes, length: length, copy: false, deallocator: deallocator)
     }
+
     public convenience init(contentsOfFile path: String, options readOptionsMask: ReadingOptions = []) throws {
         let readResult = try NSData.readBytesFromFileWithExtendedAttributes(path, options: readOptionsMask)
         self.init(bytes: readResult.bytes, length: readResult.length, copy: false, deallocator: readResult.deallocator)
@@ -380,7 +381,10 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
         }
         
         let length = Int(info.st_size)
-        
+        if length == 0 && (info.st_mode & S_IFMT == S_IFREG) {
+            return try readZeroSizeFile(fd)
+        }
+
         if options.contains(.alwaysMapped) {
             let data = mmap(nil, length, PROT_READ, MAP_PRIVATE, fd, 0)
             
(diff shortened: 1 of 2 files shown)


TestFoundation/TestNSData.swift
```diff
@@ -908,6 +910,45 @@ extension TestNSData {
...
+            let ptr =  UnsafeMutableRawPointer(mutating: contents.bytes)
+            let str = String(bytesNoCopy: ptr, length: contents.length,
+                         encoding: .ascii, freeWhenDone: false)
+            XCTAssertEqual(str, "swift-corelibs-foundation")
+        }
+    }
+
+    func test_contentsOfZeroFile() {
+#if os(Linux)
+        let contents = NSData(contentsOfFile: "/proc/self/cmdline")
```
alblue: It's possible that the `/proc/` filesystem isn't mounted in some situations, so I would guard this with a check to see if `/proc/self/cmdline` exists and silently abort if it doesn't - because it's possible someone will run this with a minimal container where procfs isn't present.

alblue:
@swift-ci please test

TestFoundation/TestNSData.swift
spevans: Good point, I will wrap something around it

spevans:
Rebased, as I was printing the contents of `/proc/self/maps` to the logs. I added a check for `/proc/self` before using it.

TestFoundation/TestNSData.swift
```diff
@@ -908,6 +910,47 @@ extension TestNSData {
...
+            if let str = str {
+                XCTAssertTrue(str.hasSuffix("TestFoundation"))
+            }
+        }
+
+        do {
+            let maps = try String(contentsOfFile: "/proc/self/maps", encoding: .utf8)
+            XCTAssertTrue(maps.characters.count > 0)
+        } catch {
+            XCTFail()
```
alblue: Might be worth having a failure reason here inside the XCTFail?

Foundation/NSData.swift
```diff
@@ -414,6 +421,39 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
...
+
+#if os(Linux)
+    internal static func readZeroSizeFile(_ fd: Int32) throws -> NSDataReadResult {
+        let blockSize = 1024 * 1024 // 1MB
+        var data: UnsafeMutableRawPointer? = nil
+        var bytesRead = 0
+        var amt = 0
+
+        repeat {
+            data = realloc(data, bytesRead + blockSize)
```
alblue: This may return NULL, in which case the unwrap subsequently will fail. Is there any reason why we can't use a single byte array for reading and then have an NSMutableData to append the bytes into? That would solve the realloc calls.

spevans: I was duplicating the usage of `malloc()` in the `readBytesFromFileWithExtendedAttributes` function above. Generally the pattern in Foundation seems to be not checking the return value of `malloc()/realloc()` and letting it eventually abort on a NULL deref. The same is true if `bytesRead` exceeds `Int.max`

alblue:
@swift-ci please test

spevans:
@alblue could you give this another test please, I think ci missed the last one

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
It seems a shame to wrap all this in `#if os(Linux)`. There are other operating systems which use a `procfs` and might find this code useful in future. Is there anything we can do about that?

spevans:
Its not a problem to remove them, although I will leave them in the test case for now, I dont think macOS has any files that can be used for tests.

spevans:
I removed the `#if os(Linux)` and also squashed & rebased to master. Tested OK in xcode9 with swift-DEVELOPMENT-SNAPSHOT-2017-08-14

ianpartridge:
What do you think @alblue?

spevans:
@parkera Any thoughts on this PR?

parkera:
We've been having a discussion about this on the team, trying to decide if it's non-portable behavior or not. In the end we decided that if you're accessing `/proc` then you're already platform-specific.

spevans:
@parkera Is this ok to merge now?

spevans:
@swift-ci please test

parkera:
Yup, we can merge when ready.

spevans:
@parkera Could you test and merge please, my swift-ci access isnt currently working

ianpartridge:

ianpartridge:

PR #1167, opened 2017-08-10
[RFC] Sync NSStringAPI.swift from overlay

ianpartridge:
Over time, `NSStringAPI.swift` has diverged from the [version in the overlay](https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/NSStringAPI.swift).

This PR brings them back into sync, by copying the overlay version into `swift-corelibs-foundation`, then fixing the resulting breakages so it builds and passes TestFoundation on Linux and in Xcode.

Hopefully this will make it easier to port changes between the two in future.

Comments please - is this worth doing and have I made a mess of it? 🙂

### DIFF
--- a/Foundation/NSStringAPI.swift
+++ b/Foundation/NSStringAPI.swift
@@ -2,11 +2,11 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
-// See http://swift.org/LICENSE.txt for license information
-// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
 //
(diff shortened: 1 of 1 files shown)


ianpartridge:
@swift-ci please test

ianpartridge:
Go to https://www.diffchecker.com/ysVPYFqr to see the diff of `NSStringAPI.swift` between the overlay and `swift-corelibs-foundation` after this PR.

alblue:
@parkera @phausler Given the upcoming merge of future changes does this make sense to get in for the Swift 4 branch?

spevans:
As a general question, shouldnt all of the APIs in the Foundation SDK overlay be kept in sync with the corelibs-foundation ones (excepting macOS/iOS specific ones) or should we expect some differences?
Is it currently just kept updated in an ad-hoc manner?

ianpartridge:
Yes they should be in sync. Complications arise because the overlay is missing some public API (because part of the Darwin implementation is closed source Objective-C), and also has extra API that's deliberately not in `swift-corelibs-foundation` because it doesn't make sense on Linux.

ianpartridge:
Rebased and fixed merge conflicts.

phausler:
Can you ifdef the deletions/changes from the overlay? That way we can copy/paste back and forth

ianpartridge:
Yes. What should the `#if` be?  `os(Linux)` or `DEPLOYMENT_RUNTIME_SWIFT`?

phausler:
Probably DEPLOYMENT_RUNTIME_SWIFT is appropriate in most if not all of the cases. We want the Xcode build of swift-corelibs-foundation to behave as close to the linux one as possible. Also we want the Darwin overlay to share the same code; any differential should be a objc differential. So it might even be good to remove `as?` casts and replace them with the compiler runtime call that turns into such as `_bridgeToObjectiveC()` etc

ianpartridge:
Sounds good. Do you know what that compiler runtime call is? Or where I should look to find it?

phausler:
`str as NSString` will emit `str._bridgeToObjectiveC()`

`nsString as? String` will emit `String._conditionallyBridgeFromObjectiveC(nsString, &res)`

ianpartridge:
OK I'll have a crack and see how I get on. Thanks.

ianpartridge:
Closing in favour of #1178

PR #1170, opened 2017-08-12
NSURL: Dont leak the result of dataRepresentation

spevans:
Validated with valgrind.


ianpartridge:
Hmm there were a bunch of these `deallocator: .none` initializations added in https://github.com/apple/swift-corelibs-foundation/commit/188a4e17036f9b7c8f1012a175a0f85f648cd31c

Worth auditing them all?

ianpartridge:
@swift-ci please test

spevans:
I audited all of the `malloc()` and `bytesNoCopy:` calls, I will look at that commit and check the ` deallocator: .none`s

spevans:
There was only one other use, in `XMLParser.swift`, and that has a `defer { free() }` after the `malloc` so I think it is all ok.

PR #1172, opened 2017-08-12
Notification: fix addObserver(forName:object:queue:using:) signature

ianpartridge:
This is a pickup of https://github.com/apple/swift-corelibs-foundation/pull/1060 which has gone stale.

Correct the API signature of 

### DIFF
--- a/Foundation/NSNotification.swift
+++ b/Foundation/NSNotification.swift
@@ -187,8 +187,13 @@ open class NotificationCenter: NSObject {
             self._observers = _observers.filterOutObserver(observer, name: aName, object: object)
         })
     }
-    
+
+    @available(swift, obsoleted: 4.0)
     open func addObserver(forName name: Notification.Name?, object obj: Any?, queue: OperationQueue?, usingBlock block: @escaping (Notification) -> Void) -> NSObjectProtocol {
+        return addObserver(forName: name, object: obj, queue: queue, using: block)
+    }
+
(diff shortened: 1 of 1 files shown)


Foundation/NSNotification.swift
```diff
@@ -187,8 +187,13 @@ open class NotificationCenter: NSObject {
             self._observers = _observers.filterOutObserver(observer, name: aName, object: object)
         })
     }
-    
+
+    @available(swift, obsoleted: 4.0)
```
alblue: We can get client-side fixes if we annotate it with a 'renamed' attribute

`@available(*,unavailable,renamed:"addObserver(forName:object:queue:using:)")`

ianpartridge: Nice. So two `@available` attributes on the same method?

alblue: Although you can have two `@available` attributes, you can also combine them for different platforms:

https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Attributes.html

So `@available(*,obsoleted:4.0,renamed:"addObserver(forName:object:queue:using:)")` will work in this case.

ianpartridge:
@swift-ci please test

alblue:
@phausler Is this the right way to indicate a renamed method, or should there be some kind of alternate way of representing changes due to the way methods are imported?

phausler:
This is probably the closest we have to a solution that does not break third party code

ianpartridge:

PR #1173, opened 2017-08-13
SR-5640:  JSONEncoder misrepresents UInt.max

spevans:
This is a workaround rather than a full fix. However it should make it easier to create the correct fix.

- NSNumber(value: UInt.max).stringValue was returning "-1"
  instead of "18446744073709551615" because NSNumber holds any
  value > Int64.max as a 128Bit quantity using a high Int and
  low UInt. It marks this type as an 'SInt128Type'.

- CFNumberFormatterCreateStringWithNumber() uses CFNumberGetType()
  to get this type however CoreFoundation wants to hide SInt128Type
  (probably for backwards compatibilty) and instead returns it
  as an SInt64Type. Thus only the low part of the NSNumber is used
  for the number and this is interpreted as an Int64.

- The workaround is simply in NSNumber.description(withLocale:)
  to test to see if the value is of type SInt128Type and if so
  use String(format: "%@") to convert it to a string instead of
  using CFNumberFormatterCreateStringWithNumber(). This should be
  ok for most situations since it is only used for positive
  integers and there are no issues with formatting leading zeros
  or decimal points.

- For JSONWriter._serializationString(for: NSNumber) the value
  is tested to see if it is an SInt128Type and if so the .stringValue
  method is used to create a string.

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -577,9 +577,12 @@ private struct JSONWriter {
             writer(" ")
         }
     }
-    
+
     //[SR-2151] https://bugs.swift.org/browse/SR-2151
     private mutating func _serializationString(for number: NSNumber) -> String {
+        if !CFNumberIsFloatType(number._cfObject) {
+            return number.stringValue
+        }
         return CFNumberFormatterCreateStringWithNumber(nil, _numberformatter, number._cfObject)._swiftObject
     }
 }
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -256,7 +256,7 @@ open class NSNumber : NSValue {
             fatalError("unsupported CFNumberType: '\(numberType)'")
         }
     }
-    
+
     deinit {
         _CFDeinit(self)
     }
@@ -536,13 +536,25 @@ open class NSNumber : NSValue {
         }
     }
 
+    private static let _numberFormatterForNilLocale: CFNumberFormatter = {
+        let formatter: CFNumberFormatter
+        formatter = CFNumberFormatterCreate(nil, CFLocaleCopyCurrent(), kCFNumberFormatterNoStyle)
+        CFNumberFormatterSetProperty(formatter, kCFNumberFormatterMaxFractionDigits, 15._bridgeToObjectiveC())
+        return formatter
+    }()
+
     open func description(withLocale locale: Locale?) -> String {
+        // CFNumberFormatterCreateStringWithNumber() does not like numbers of type
+        // SInt128Type, as it loses the type when looking it up and treats it as
+        // an SInt64Type, so special case them.
+        if _CFNumberGetType2(_cfObject) == kCFNumberSInt128Type {
+            return String(format: "%@", unsafeBitCast(_cfObject, to: UnsafePointer<CFNumber>.self))
+        }
+
         let aLocale = locale
         let formatter: CFNumberFormatter
         if (aLocale == nil) {
-            formatter = CFNumberFormatterCreate(nil, CFLocaleCopyCurrent(), kCFNumberFormatterNoStyle)
-            CFNumberFormatterSetProperty(formatter, kCFNumberFormatterMaxFractionDigits, 15._bridgeToObjectiveC())
-
+            formatter = NSNumber._numberFormatterForNilLocale
         } else {
             formatter = CFNumberFormatterCreate(nil, aLocale?._cfObject, kCFNumberFormatterDecimalStyle)
         }
--- a/TestFoundation/TestJSONEncoder.swift
+++ b/TestFoundation/TestJSONEncoder.swift
@@ -401,6 +401,68 @@ class TestJSONEncoder : XCTestCase {
         test_codingOf(value: URL(string: "https://swift.org")!, toAndFrom: "\"https://swift.org\"")
     }
 
+
+    // UInt and Int
+    func test_codingOfUIntMinMax() {
+
+        let encoder = JSONEncoder()
+
+        struct MyValue: Codable {
+            let intMin:Int = Int.min
+            let intMax:Int = Int.max
+            let uintMin:UInt = UInt.min
+            let uintMax:UInt = UInt.max
+        }
+
+        let myValue = MyValue()
+        let myDictI: [String:Any] = ["intMin": myValue.intMin, "intMax": myValue.intMax]
+        let myDictU: [String:Any] = ["uintMin": myValue.uintMin, "uintMax": myValue.uintMax]
+        let myDict1: [String:Any] = ["intMin": myValue.intMin]
+        let myDict2: [String:Any] = ["intMax": myValue.intMax]
+        let myDict3: [String:Any] = ["uintMin": myValue.uintMin]
+        let myDict4: [String:Any] = ["uintMax": myValue.uintMax]
+
+        func compareJSON(_ s1: String, _ s2: String) {
+            let ss1 = s1.trimmingCharacters(in: CharacterSet(charactersIn: "{}")).split(separator: Character(",")).sorted()
+            let ss2 = s2.trimmingCharacters(in: CharacterSet(charactersIn: "{}")).split(separator: Character(",")).sorted()
+            XCTAssertEqual(ss1, ss2)
+        }
+
+        do {
+            let result = try encoder.encode(myValue)
+            let r = String(data: result, encoding: .utf8) ?? "nil"
+            compareJSON(r, "{\"uintMin\":0,\"uintMax\":18446744073709551615,\"intMin\":-9223372036854775808,\"intMax\":9223372036854775807}")
+
+            let resultI = try JSONSerialization.data(withJSONObject: myDictI)
+            let rI = String(data: resultI, encoding: .utf8) ?? "nil"
+            compareJSON(rI, "{\"intMin\":-9223372036854775808,\"intMax\":9223372036854775807}")
+
+            let resultU = try JSONSerialization.data(withJSONObject: myDictU)
+            let rU = String(data: resultU, encoding: .utf8) ?? "nil"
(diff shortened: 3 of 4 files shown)


Foundation/NSNumber.swift
```diff
@@ -256,6 +256,10 @@ open class NSNumber : NSValue {
             fatalError("unsupported CFNumberType: '\(numberType)'")
         }
     }
+
+    public var isSInt128Type: Bool {
```
ianpartridge: We can't add new public API unfortunately.

spevans:
Ah, didn't realise that. Ive removed the function and inlined the test instead

TestFoundation/TestNSNumber.swift
```diff
@@ -1019,4 +1020,56 @@ class TestNSNumber : XCTestCase {
         XCTAssertEqual("f" /* 0x66 */, objCType(NSNumber(value: Float.greatestFiniteMagnitude)))
         XCTAssertEqual("d" /* 0x64 */, objCType(NSNumber(value: Double.greatestFiniteMagnitude)))
     }
+
+    func test_stringValue() {
+        XCTAssertEqual(NSNumber(value: UInt.min).stringValue, "0")
+        XCTAssertEqual(NSNumber(value: UInt.min + 1).stringValue, "1")
+        XCTAssertEqual(NSNumber(value: UInt.max).stringValue, "18446744073709551615")
```
ianpartridge: On a 32-bit platform `UInt.max` is 4294967295 so this will break.

TestFoundation/TestNSNumber.swift
```diff
@@ -1019,4 +1020,56 @@ class TestNSNumber : XCTestCase {
...
+        XCTAssertEqual(NSNumber(value: UInt32.min + 1).stringValue, "1")
+        XCTAssertEqual(NSNumber(value: UInt32.max).stringValue, "4294967295")
+        XCTAssertEqual(NSNumber(value: UInt32.max - 1).stringValue, "4294967294")
+
+        XCTAssertEqual(NSNumber(value: UInt64.min).stringValue, "0")
+        XCTAssertEqual(NSNumber(value: UInt64.min + 1).stringValue, "1")
+        XCTAssertEqual(NSNumber(value: UInt64.max).stringValue, "18446744073709551615")
+        XCTAssertEqual(NSNumber(value: UInt64.max - 1).stringValue, "18446744073709551614")
+
+        XCTAssertEqual(NSNumber(value: Int.min).stringValue, "-9223372036854775808")
```
ianpartridge: Same comment here - `Int.min` and `Int.max` are different on 32-bit platforms.

spevans: is there any easy test for platform wordsize apart from Int.max == Int32.max etc ?

TestFoundation/TestJSONEncoder.swift
```diff
@@ -401,6 +401,68 @@ class TestJSONEncoder : XCTestCase {
         test_codingOf(value: URL(string: "https://swift.org")!, toAndFrom: "\"https://swift.org\"")
     }
 
+
+    // UInt and Int
+    func test_uintMinMax() {
```
ianpartridge: Can the naming of this be consistent with the other tests, so `test_codingOfUIntMinMax()`?

Also we should add this testcase on the Darwin side too: https://github.com/apple/swift/blob/master/test/stdlib/CodableTests.swift

TestFoundation/TestNSNumber.swift
ianpartridge: I'm only aware of testing `MemoryLayout<Int>.size == 32`, which isn't much better.

ianpartridge:
I'm wondering if the upcoming merge of code from Apple might have a better fix for this...

ianpartridge:
@swift-ci please test

spevans:
I fixed up the test names. I left the CodableTests.swift out for now as I will do a separate JSON PR for some other minor fixes I want to do.

I suspect you are correct about there being a better fix in the 10.13 code dump but won't that happen after 4.0 has been released, i.e. when Xcode9 and High Sierra and publicly available?

spevans:
Could you please retest, I added another commit after you kicked off the last test, thanks.

ianpartridge:
@swift-ci please test

ianpartridge:
I think the code drop is due pretty soon: after we stop auto-merging from `master` to `swift-4.0-branch` but before Swift 4 ships. It's possible we might be able to extract the fix into a PR against `swift-4.0-branch`, I don't know.

ianpartridge:
@swift-ci please test

Foundation/JSONSerialization.swift
```diff
@@ -577,9 +577,12 @@ private struct JSONWriter {
             writer(" ")
         }
     }
-    
+
     //[SR-2151] https://bugs.swift.org/browse/SR-2151
     private mutating func _serializationString(for number: NSNumber) -> String {
+        if _CFNumberGetType2(number._cfObject) == kCFNumberSInt128Type {
```
parkera: It seems like an encapsulation violation to have JSONSerialization calling this function.

itaiferber: @parkera I agree, but this seems like a reasonable fix for the moment until changes come in from the CF side, no? Or is it possible to get CF changes in for this?

spevans: Since the `stringValue` can be used for any integer value, I could replace it with: 
```
        if !CFNumberIsFloatType(number._cfObject) {
            return number.stringValue
        }
```

parkera: The problem I have is the way this ties the implementation detail of how NSNumber stores its value to JSONSerialization - the two shouldn't really be coupled to each other that tightly if we can help it. Maybe the formatter is a better place to encapsulate that...

spevans: @parkera So ideally you want all references to _cfObject removed from JSONSerialization?

parkera: No - it's the `_CFNumberGetType2` that is the problem. `_cfObject` is just our way to get toll-free bridging to work here.

spevans: so is ` !CFNumberIsFloatType(number._cfObject)` ok to use instead?

spevans:
I replaced the ` _CFNumberGetType2(number._cfObject) == kCFNumberSInt128Type`

ianpartridge:
@swift-ci please test

parkera:
Yes, using the public CFNumber API is much better. Thanks.

spevans:
Looks like the test failed due to a build timeout, Ive just tested it against swift-DEVELOPMENT-SNAPSHOT-2017-08-14 on Xcode and it worked ok

ianpartridge:
@swift-ci please test

djones6:
I think there's a performance problem with the latest change - by routing the stringification of all int types through `NSNumber.description`, with the exception of the special case for SInt128, every conversion will create a new CFNumberFormatter.

A quick benchmark of JSONEncoder (on Linux) with a struct containing 10 Ints ran 2x slower.
(benchmark: https://github.com/djones6/swift-benchmarks/blob/master/jsonEncoderTest.swift)

ianpartridge:
I guess we could change `NSNumber.description(withLocale:)` so it caches a static `CFNumberFormatter` and only creates a new one if the locale changes?

spevans:
It could just just a fixed one when locale is `nil` as that is the case we care about

spevans:
@djones6 I added a cache for the CFNumberFormatter when locale is nil, can you see if that helps with the performance regression?

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

Foundation/NSNumber.swift
```diff
@@ -536,13 +536,26 @@ open class NSNumber : NSValue {
...
 
+    private static let _numberFormatterForNilLocale: CFNumberFormatter = {
+        let formatter: CFNumberFormatter
+        formatter = CFNumberFormatterCreate(nil, CFLocaleCopyCurrent(), kCFNumberFormatterNoStyle)
+        CFNumberFormatterSetProperty(formatter, kCFNumberFormatterMaxFractionDigits, 15._bridgeToObjectiveC())
+        return formatter
+    }()
+
     open func description(withLocale locale: Locale?) -> String {
+        // CFNumberFormatterCreateStringWithNumber() doesnt like SInt128Type as
```
xwu: Drive-by nit: can you correct the spelling of "doesn't" and throw in a few punctuation marks (before the first "as" and before "so")?

Foundation/NSNumber.swift
```diff
@@ -536,13 +536,26 @@ open class NSNumber : NSValue {
...
+        // it loses the type when looking it up and treats it as an SInt64Type
+        // so special case them.
+        if _CFNumberGetType2(_cfObject) == kCFNumberSInt128Type {
+            return String(format: "%@", unsafeBitCast(_cfObject,
+                                                      to: UnsafePointer<CFNumber>.self))
```
xwu: Drive-by nit no. 2: can you put this on the same line?

djones6:
@spevans Thanks - this does address the performance regression.  In fact, it seems to perform slightly better than the original code.

spevans:
Thanks for running those tests @djones6.

I have squashed and rebased and tidied up the comment so I think this may be ready to go.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
I think this is good for merge - @spevans @djones6 @itaiferber @parkera?

spevans:
I think it is. Its fixes the NSNumber bug and I will have a follow up PR soon to hopefully speed up some of the integer serialisation.

PR #1176, opened 2017-08-14
Fix memory leaks by using more specialized xml*Free functions.

mattrajca:
Fixes https://bugs.swift.org/browse/SR-5256.


parkera:
Does our current test case hit these lines?

spevans:
TestXMLDocument.test_xpath() covers them.

djones6:
I can confirm this fixes the leak. Thanks @mattrajca !

ianpartridge:

ianpartridge:
FYI @svanimpe this fixes the XML memory leak you reported on the Kitura slack 🙂

PR #1177, opened 2017-08-15
Fix String.substring() deprecation warnings

ianpartridge:
Fix these warnings:

`Warning: 'substring(from:)' is deprecated: Please use String slicing subscript with a 'partial range from' operator.`

`Warning: 'substring(to:)' is deprecated: Please use String slicing subscript with a 'partial range upto' operator.`

Personally I don't think this looks like an improvement, but it's what https://github.com/apple/swift/commit/fc0cf33bdc3cdf02513cb058f086603ad5c1fdcb implements.

I think a lot of people are going to reimplement `substring(from:)` and `substring(to:)` in their own extensions.


ianpartridge:
@swift-ci please test

alblue:
Looks reasonable to fix the compiler warnings - what do you think @phausler ?

phausler:
The conversion looks fine but I wonder if we should perhaps test the build directory with a ü in the path or something like that. But that might just be an existing bug that we should address separately.

ianpartridge:

PR #1181, opened 2017-08-17
Whole module optimization

phausler:
NSClassFromString fails in whole module optimization so this corrects that to at least work-around the failures of Foundation classes. This changes the `-O` optimization to `-Owholemodule`


phausler:
@swift-ci please test

phausler:
@swift-ci please test

Foundation/NSObjCRuntime.swift
```diff
@@ -266,22 +266,143 @@ public func NSStringFromClass(_ aClass: AnyClass) -> String {
...
  */
 public func NSClassFromString(_ aClassName: String) -> AnyClass? {
-    let aClassNameWithPrefix : String
-    let components = aClassName._bridgeToObjectiveC().components(separatedBy: ".")
-    
-    switch components.count {
-    case 1:
-        guard !aClassName.hasPrefix("_Tt") else {
-            NSLog("*** NSClassFromString(\(aClassName)): cannot yet decode mangled class names")
+    switch aClassName {
```
parkera: Isn't this going to be a complete maintenance nightmare? There has to be some other way to handle this.

phausler: the only other option here is to use a dictionary but this is a work-around for a bug that seems to be in `swift_getTypeByName` in the runtime library for swift or somehow in the metadata records that gets stripped out in release builds.

It seems that the failure occurs only with a two level deep inheritance:

```
// assuming module named test
class A { }
class B: A { }
class C: B { }

assert(_typeByName("test.A") != nil)
assert(_typeByName("test.B") != nil)
assert(_typeByName("test.C") != nil) // fails here
```

I need to understand a bit more to file the appropriate bugs for this.

spevans: I have an alternate solution, based on a workaround in this comment: 
https://bugs.swift.org/browse/SR-381?focusedCommentId=11290&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-11290


The `NSClassFromString()` seems to fail for any Class that doesn't directly conform to a protocol, so I added a dummy protocol and made the Classes that fail the tests conform to it.
See https://github.com/spevans/swift-corelibs-foundation/commit/99fa8153c830b5799eee10cf4bbeecd77eb444d7

Note that there are probably some other classes we need to fix, I just fixed up the ones that made `TestFoundation` fail when built for release mode running under Instruments.


https://bugs.swift.org/browse/SR-3333 is tracking this issue.

phausler: The other problem here is that if a class had the NS dropped the name should still have the NS.

E.g.

`NSTask` got renamed to `Process`. If you ask the NSStringFromClass(Process.self) on the objc backed swift it will return "NSTask" whereas if you ask for it in the swift-corelibs-foundation version you can "Process" which is wrong.

phausler: This change fixes the reverse of that; we should follow up with the other direction.

spevans: Since we have a default case can't we remove the specific cases where the String and class match, e.g. `NSAffineTransform`, that should reduce the number of entries

TestFoundation/TestNSClassFromString.swift
```diff
@@ -0,0 +1,269 @@
...
+        XCTAssertEqual("NSURLSessionStreamTask", NSStringFromClass(URLSessionStreamTask.self))
+        XCTAssertEqual("NSURLSessionTask", NSStringFromClass(URLSessionTask.self))
+        XCTAssertEqual("NSURLSessionUploadTask", NSStringFromClass(URLSessionUploadTask.self))
+        XCTAssertEqual("NSXMLDocument", NSStringFromClass(XMLDocument.self))
+        XCTAssertEqual("NSXMLDTD", NSStringFromClass(XMLDTD.self))
+        XCTAssertEqual("NSXMLParser", NSStringFromClass(XMLParser.self))
+    }
+    
+    func test_classes() {
+        func XCTAssertEqualClasses(_ expected: AnyClass?, _ actual: AnyClass?, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) {
```
phausler: XCTAssertEqual does not work on `AnyClass?`

Foundation/NSObjCRuntime.swift
```diff
@@ -234,6 +234,146 @@ private let _SwiftFoundationModuleName = "SwiftFoundation"
...
+    "NSURLSessionDownloadTask" : URLSessionDownloadTask.self,
+    "NSURLSessionStreamTask" : URLSessionStreamTask.self,
+    "NSURLSessionTask" : URLSessionTask.self,
+    "NSURLSessionUploadTask" : URLSessionUploadTask.self,
+    "NSXMLDocument" : XMLDocument.self,
+    "NSXMLDTD" : XMLDTD.self,
+    "NSXMLParser" : XMLParser.self,
+]
+
+internal struct _NSClassWrapper : Hashable {
```
phausler: classes are not directly Hashable so cannot be used as keys to a dictionary.

phausler:
@parkera thoughts on this? should I re-take it back to the drawing boards?

parkera:
I'm still concerned about the maintenance burden of this. The fact that we need to hard code our list of classes feels like a bug - or we should figure out a way to stop supporting NSClassFromString as API.

phausler:
NSClassFromString is required for NSKeyedArchiving so I don't think we can shirk that. The table is just a work-around until we can get the bug resolved.

The problem is that a class that does not adopt a protocol is not serialized with it's name metadata in the binary. But 90% of the time if a consumer of Foundation is making a class I that would adopt NSCoding I would guess they are not subclassing that. So this just keeps us in the clear for the few cases that are failing and memoizes the access to a table. I was easily able to generate this with a script that validates the proper name bindings. 

The other issue is that we have renamed classes - this will NEVER pick up on their NS prefix'd name no matter what the compiler does so we will always have to have a work-around in some shape or form.

spevans:
Could we not break this into 2 fixes:

1. Use a dictionary to map names to classes where the class name has been renamed. This mapping may also be useful elsewhere at some point and I do not see any other way of managing it. The dictionary should hopefully be quite small.

2. For the classes that fail to be looked up because of the lack of direct protocol conformance, add a dummy protocol for them to conform to. If a proper fix is ever added to Swift it can then be removed.

spevans:
@phausler Whats the status of this PR? For linux, the `NSClassFromString` isn't required for `-whole-module-optimization` to be applied, its only required for Release mode to work in Xcode for macOS. Can the `-whole-module-optimization` single change be broken out into its own PR to enable it for Linux?

alblue:
@phausler ping ...

alblue:
I believe this change request has gone stale, so closing. Please re-open if you bring the change back up with master, and/or break itno separate patches for the whole-module vs NSClassFromString changes (which may conflict with the recent update from HighSierra)

PR #1183, opened 2017-08-22
Fix AsyncOperation test by adding locking around flags.

spevans:
- Use an NSRecursiveLock() as didChangeValue() will call isFinished
  when the key is 'isFinished'.

This test was failing under Xcode when running in Release mode.

### DIFF
--- a/TestFoundation/TestOperationQueue.swift
+++ b/TestFoundation/TestOperationQueue.swift
@@ -175,12 +175,17 @@ class AsyncOperation: Operation {
 
     override internal(set) var isExecuting: Bool {
         get {
-            return _executing
+            lock.lock()
+            let wasExecuting = _executing
+            lock.unlock()
+            return wasExecuting
         }
         set {
-            if _executing != newValue {
+            if isExecuting != newValue {
                 willChangeValue(forKey: "isExecuting")
+                lock.lock()
                 _executing = newValue
+                lock.unlock()
                 didChangeValue(forKey: "isExecuting")
             }
         }
@@ -188,12 +193,17 @@ class AsyncOperation: Operation {
 
     override internal(set) var isFinished: Bool {
         get {
-            return _finished
+            lock.lock()
+            let wasFinished = _finished
+            lock.unlock()
+            return wasFinished
         }
         set {
-            if _finished != newValue {
+            if isFinished != newValue {
                 willChangeValue(forKey: "isFinished")
+                lock.lock()
                 _finished = newValue
+                lock.unlock()
                 didChangeValue(forKey: "isFinished")
             }
         }
@@ -213,10 +223,8 @@ class AsyncOperation: Operation {
 
         queue.async {
             sleep(1)
-            self.lock.lock()
             self.isExecuting = false
             self.isFinished = true
-            self.lock.unlock()
         }
     }
 


spevans:
@weissi Since you are looking at locking and threading could you give an opinion on this PR please?

TestFoundation/TestOperationQueue.swift
```diff
@@ -168,34 +168,42 @@ class TestOperationQueue : XCTestCase {
 class AsyncOperation: Operation {
 
     private let queue = DispatchQueue(label: "async.operation.queue")
-    private let lock = NSLock()
+    private let lock = NSRecursiveLock()
```
weissi: - why a lock and not a `DispatchQueue`?
 - and why a recursive lock?

spevans: I changed the lock to `NSRecursiveLock()` because when `isFinished.set` is called and it takes the lock, the call to `didChangeValue` https://github.com/apple/swift-corelibs-foundation/blob/84ee193f9181838a366fe5f5e02d6345c7338c3a/Foundation/Operation.swift#L179  calls `isFinished.get` which would reaquire the lock. Making it recursive was the simplest fix. I didnt want to make any invasive changes in `Operation.swift`

What would be the benefit of the `DispatchQueue`? Would it not be too heavy an operation?

weissi: Right, it's a test and therefore I'd be happy with a recursive lock, generally speaking they lead to dead-lock prone code (as establishing a lock order isn't as easy anymore).

A `DispatchQueue` is very cheap and in ObjC at least a synchronous dispatch on a dispatch queue is basically the same speed as taking a mutex. They should be both > 5 ns and < 50 ns for a non-contended case. Said that, in Swift, there's currently a [performance issue with synchronous dispatches](https://bugs.swift.org/browse/SR-4937). This  shouldn't matter here as an Asynchronous Operation is a class anyway meaning we'll get memory allocations and reference counting. In my benchmarks it should be 20 ns (nano seconds) per synchronous dispatch but in Swift it's 140 ns. I'd be really surprised if you'd see this difference when using Operation/OperationQueue.

The advantage of a dispatch queue is that the API is easier to use (can't forget to unlock) and it's the preferred way of synchronising.

spevans: TBH I don't actually like the way `didChangeValue` is written, but I was just doing the simplest fix to get `TestFoundation` to run under Release mode in Xcode, so that it can be run under Instruments, and this was one of the failing tests. Is that 140ns in `-Onone` or `-O`?

If you have an alternate solution I have no problem with using it, this is just test code anyway.

weissi: 140ns for an optimised build :(. I expect that to be fixed soon though.

Yeah, go with it, as you say test code and there's certainly bigger issues :)

phausler: KVO callouts should NEVER be done inside locks btw you never know who is watching.
the lock should only be done for the critical value change of the actual assignment of the new value.

We should show good design patterns especially for long standing mysterious classes like NSOperation or NSOperationQueue. The atomicity of the values is appropriate (unfortunately we have to hit it with a sledgehammer of pthread_mutex_t instead of using the appropriate atomic operations)

TestFoundation/TestOperationQueue.swift
```diff
@@ -168,34 +168,42 @@ class TestOperationQueue : XCTestCase {
...
 
     private var _executing = false
     private var _finished = false
 
     override internal(set) var isExecuting: Bool {
         get {
+            self.lock.lock()
```
phausler: you really don't need the `self.` here; it should be expressible as lock.lock()

TestFoundation/TestOperationQueue.swift
```diff
@@ -168,34 +168,42 @@ class TestOperationQueue : XCTestCase {
...
+            defer { self.lock.unlock() }
```
phausler: using the defer here does miss out on the thread safety consideration that consumers should think about:
```
var isExecuting: Bool {
    get {
        lock.lock()
        let wasExecuting = _executing
        lock.unlock()
        return wasExecuting
    }
}
```

it is a bit cheeky but it gets the point across that the return value should only be a perspective of the potential past.

TestFoundation/TestOperationQueue.swift
weissi: @phausler ha, you remind me that I wanted to ask a question about that. Obviously it's really bad to call out whilst holding a lock. I was wondering what the semantics of KVO are though. What happens in this case:

current value: 1

T1: wants to write 2, takes lock, reads the current value (1), releases lock, calls out `willChange: 2`
T2: wants to write 3, takes lock, reads the current value (still 1), releases lock, calls out `willChange: 3`
T1: KVO callout has returned, takes lock, writes value (2), releases lock
T2: KVO callout has returned, takes lock, writes value (3), releases lock
T1: KVO callout `didChange: 2` (technically incorrect as it's already 3)
T2: KVO callout `didChange: 3`

is this legal? Because if we don't hold the lock for long, we can get KVO sequences like `willChange: X`, `willChange: Y`, `didChange: X`, `didChange: Y`, is that expected?

phausler: For KVO in objc, the last wins and the will change is practically not used. This is of course a non-issue since the result should only transition to a valid state and any invalid transition e.g. from finished back to not finished are non-sense in the first place.

The objective-c implementation of NSOperationQueue assumes that the state machine transition of operations is always forward progressing. Any backwards movement is considered either a) invalid by nature and will halt or b) ignorable since it is nonsense.

TestFoundation/TestOperationQueue.swift
spevans: So are you suggesting something like:
```
set {
    if isFinished != newValue {  // get() in a lock
        willChangeValue(forKey: "isFinished")
        lock.lock()
        _finished = newValue
        lock.unlock()
        didChangeValue(forKey: "isFinished")
   }
}
```

phausler: yup exactly

spevans:
@swift-ci please test

spevans:
@phausler How does this look now?

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

spevans:
FYI, this allows `-whole-module-optimization` to be used on Linux in release mode (it was the only test that was failing).

ianpartridge:

ianpartridge:

PR #1186, opened 2017-08-22
improve threading correctness in URLSession

weissi:
this is really only the beginning to prevent everything from breaking with the simplest test that does use more than one thread. Thread-safely seems to be large absent in this part of foundation 😢. 
Unfortunately, I seem to be going down a rabbit hole here but it's a start.

### DIFF
--- a/Foundation/URLSession/URLSession.swift
+++ b/Foundation/URLSession/URLSession.swift
@@ -179,11 +179,13 @@ extension URLSession {
     }
 }
 
+fileprivate let globalVarSyncQ = DispatchQueue(label: "org.swift.Foundation.URLSession.GlobalVarSyncQ")
 fileprivate var sessionCounter = Int32(0)
 fileprivate func nextSessionIdentifier() -> Int32 {
-    //TODO: find an alternative for OSAtomicIncrement32Barrier() on Linux
-    sessionCounter += 1
-    return sessionCounter
+    return globalVarSyncQ.sync {
+        sessionCounter += 1
+        return sessionCounter
+    }
 }
 public let NSURLSessionTransferSizeUnknown: Int64 = -1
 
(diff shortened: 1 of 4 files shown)


Foundation/URLSession/URLSessionTask.swift
```diff
@@ -555,7 +517,9 @@ extension _ProtocolClient : URLProtocolClient {
             session.delegateQueue.addOperation {
                 delegate.urlSession(session, task: task, didCompleteWithError: nil)
                 task.state = .completed
-                session.taskRegistry.remove(task)
+                task.workQueue.async {
```
pushkarnk: I think this will fix an extremely intermittent crash a few of us have seen.

pushkarnk:
@swift-ci please test

Foundation/URLSession/URLSessionTask.swift
```diff
@@ -62,7 +59,8 @@ open class URLSessionTask : NSObject, NSCopying {
     }
     internal init(session: URLSession, request: URLRequest, taskIdentifier: Int, body: _Body) {
         self.session = session
-        self.workQueue = session.workQueue
+        /* make sure we're actually having a serial queue as it's used for synchronization */
+        self.workQueue = DispatchQueue.init(label: "org.swift.URLSessionTask.WorkQueue", target: session.workQueue)
```
pushkarnk: Makes sense. However, I recall we had replaced the use of DispatchQueues with semaphores in attempts to work around a weird and intermittent hang in `DispatchQueue.async` - something we could reproduce only in a couple of local environments.

Foundation/URLSession/URLSession.swift
```diff
@@ -171,12 +171,13 @@
 import CoreFoundation
 import Dispatch
 
-
+fileprivate let globalVarSyncQ = DispatchQueue(label: "org.swift.Foundation.URLSession.GlobalVarSyncQ")
```
pushkarnk: Thanks! This was long pending and ignored :(

weissi:
@swift-ci please test

weissi:
@swift-ci please test

Foundation/URLSession/URLSessionTask.swift
weissi: @pushkarnk well, if that's the case I want to know and I'll fix the issue.

weissi:
tests all passed and approved, so merging.

PR #1188, opened 2017-08-24
Revert "Fix for a race condition in URLSession (#949)"

weissi:
This reverts commit 015ded4d70a7cf4b2741d7206fc60a3d1ceffe03.

The reason to revert this is that it this pretty certainly works around another bug ([SR-5759](https://bugs.swift.org/browse/SR-5759)) and IMHO introduces two new bugs instead.

Dispatch Sources should be cancellable independently. If you have two DispatchSources watching for events on one and the same PR, the should be independent. That was seemingly the problem that was seen here, see the [original bug](https://bugs.swift.org/browse/SR-4567) and the discussion in #949.

The two new bugs introduced:

 - a blocking wait `self.cancelInProgress?.wait()` is introduced when creating a `_MultiHandle`. Creating a large number of `_MultiHandle`s concurrently could lead to a lot of threads being blocked in the wait. That might exhaust libdispatch's thread pool which might then lead to the cancellation handlers never been fully run
 - `CFURLSessionMultiHandleAssign` is run on the wrong thread (`commandQueue.async `)

this should only be merged when SR-5759 is done.

### DIFF
--- a/Foundation/URLSession/http/MultiHandle.swift
+++ b/Foundation/URLSession/http/MultiHandle.swift
@@ -39,11 +39,7 @@ extension URLSession {
         let group = DispatchGroup()
         fileprivate var easyHandles: [_EasyHandle] = []
         fileprivate var timeoutSource: _TimeoutSource? = nil
-
-        //SR-4567: we need to synchronize the register/unregister commands to the epoll machinery in libdispatch
-        fileprivate let commandQueue: DispatchQueue = DispatchQueue(label: "Register-unregister synchronization")
-        fileprivate var cancelInProgress: DispatchSemaphore? = nil
-
+        
         init(configuration: URLSession._Configuration, workQueue: DispatchQueue) {
             queue = DispatchQueue(label: "MultiHandle.isolation", target: workQueue)
             setupCallbacks()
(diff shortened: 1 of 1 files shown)


pushkarnk:
Looks like https://github.com/apple/swift-corelibs-libdispatch/pull/295 was merged. Let me edit the title first

pushkarnk:
I ran the test related to SR-4567 and it seems to pass.

pushkarnk:
@MadCoder I took a look at https://github.com/apple/swift-corelibs-libdispatch/pull/295 and  have a fundamental question in the context of #949 

A DispatchSource.cancel() call returns immediately, the actual `EPOLL_CTL_DEL` happens asynchronously, and the cancel handler invocation indicates the completion of the `cancel` operation. The problem in #949 was that after the `cancel` returned but before the `EPOLL_CTL_DEL` was executed, if we get `EPOLL_CTL_ADD` or `EPOLL_CTL_MOD` requests on the same fd, the latter are missed and we see a hang. Does your PR address this problem?

I did run the test case for #949 only with your fix and it goes through, but just wanted to be sure the problem was actually addressed.

MadCoder:
Yes, I think my fix addresses this problem

pushkarnk:
Thanks @weissi for this great piece of work. As you proved, #949 was vulnerable, because, IMO, fixing a race condition with added synchronisation often creates more race conditions :) I'll go ahead commit this revert.

pushkarnk:

PR #1190, opened 2017-08-26
(Not for checkin) Fix factory init hack for convenience init 'self' type change

slavapestov:


slavapestov:
If we implement https://github.com/apple/swift-evolution/blob/master/proposals/0068-universal-self.md, then the generic hack can go away, and instead we would do `unsafeBitCast(cfnumber, to: Self)`.

Or of course we could implement `factory` initializers. My current thinking is the following:

- We should allow 'self.init()' delegation as well as 'self =' assignment in *both* `convenience` and `factory` inits.
- `factory` inits would never be inherited, unlike `convenience` inits.

This would mean that `self = ...` in convenience inits would be severely restricted to a right hand side that is known at compile-time to have the same dynamic type as 'self'. But a common case would be static Self-returning methods, eg

```
class C {
  static func factory(_: Int) -> Self { ... }

  convenience init(factory: Int) {
    self = type(of: self).factory()
  }
}
```

From an implementation standpoint the above would be trivial to add because the current restriction on self-assignment in class initializers is largely artificial. The self value is already stored in an lvalue box to handle cases where self.init delegation to a protocol extension or @objc method replaces the self value.

@parkera @phausler Curious to hear your thoughts on what we want to do with this long-term...

slavapestov:
@parkera @phausler Here is an even better hack: https://github.com/apple/swift-corelibs-foundation/pull/1191

If we merge that PR, I could remove the `-swift-version 5` conditional. I think this would be desirable, since the source compatibility suite passed, and the new behavior fixes miscompiles.

PR #1191, opened 2017-08-26
Different hack for factory inits

slavapestov:
This will allow swift-corelibs-foundation to build if the change in https://github.com/apple/swift/pull/11637 is enabled unconditionally.

### DIFF
--- a/Foundation/NSValue.swift
+++ b/Foundation/NSValue.swift
@@ -157,11 +157,12 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
 extension NSValue : _Factory {}
 
 internal protocol _Factory {
-    init(factory: () -> Self)
+    associatedtype StaticSelf = Self
+    init(factory: () -> StaticSelf)
 }
 
 extension _Factory {
-    init(factory: () -> Self) {
-        self = factory()
+    init(factory: () -> StaticSelf) {
+        self = factory() as! Self
     }
 }


phausler:
The cast is better, but perhaps we have a different definition of factory patterns? So the init methods that are implemented are often conditional methods that in the abstract case do self replacement in init (emulating the replacement in alloc and separate return in the init) and in the subclass case call to funnel init methods.

Is that behavior on par with the swift definition? I know these methods are missing the conditional part.

Foundation/NSValue.swift
```diff
@@ -157,11 +157,12 @@ open class NSValue : NSObject, NSCopying, NSSecureCoding, NSCoding {
...
-    init(factory: () -> Self)
+    associatedtype StaticSelf = Self
+    init(factory: () -> StaticSelf)
 }
 
 extension _Factory {
-    init(factory: () -> Self) {
-        self = factory()
+    init(factory: () -> StaticSelf) {
+        self = factory() as! Self
```
phausler: The as! might be problematic with the CF changes I have for generalizing the factory patterns. I know I had a problem with unsafeReferenceCast which is approx the same as as!

slavapestov: What kind of problems were you having?

slavapestov:
> Is that behavior on par with the swift definition? I know these methods are missing the conditional part.

We have the same definition, the part that Swift cannot currently express is defining initializers which are never inherited by subclasses.

For a self-replacing constructor to be safe, it cannot be applied to a subclass (or we just enforce that it is overridden in a subclass).

Right now we have no way to enforce this statically, so it was possible to write a factory initializer that would just crash with a subclass at runtime. I made this a type check error -- this PR which adds the 'as!' cast basically allows the old idiom to type check even though it's not safe with subclassing.

Does this clarify the situation?

Foundation/NSValue.swift
slavapestov: Can you try this patch with your in-progress changes and let me know if you run into problems? It should work with -swift-version 4 as well.

alblue:
This change is a couple of months old and may be outdated. Can you confirm if you're still working on it, or whether we should close this PR?

alblue:
@slavapestov I'm assuming this has gone stale, or there's another way of solving it. In the meantime, I'm going to close this PR - but if you pick up the work again, can you rebase and resubmit?

PR #1148, opened 2017-07-31
Codable conformance to some Foundation types

bubski:
Continuation of #1108 & #1127
This PR mirrors _some_ changes from apple/swift#9715

---

### Added conformances + unit tests for

- Calendar
- DateComponents
- TimeZone

📝 I had to disable or work around some issues on Linux caused by https://bugs.swift.org/browse/SR-5598.
Broken `TimeZone.current == TimeZone(identifier: TimeZone.current.identifier)` equality causes **encode -> decode -> compare** roundtrip tests to fail.

### Not included in this PR

- Measurement

### Already implemented

- PersonNameComponents
- UUID
- URL
- NSRange
- Locale
- IndexSet
- IndexPath
- AffineTransform
- Decimal
- DateInterval
- CGFloat (unit tested indirectly)
- CharacterSet

PR #1192, opened 2017-08-27
Codable conformance for Measurement

bubski:
Continuation of #1108, #1127 & #1148 
This PR mirrors _final missing_ changes from apple/swift#9715

---

- Measurement

### DIFF
--- a/Foundation/Unit.swift
+++ b/Foundation/Unit.swift
@@ -152,6 +152,18 @@ open class Unit : NSObject, NSCopying, NSSecureCoding {
     }
 
     public static var supportsSecureCoding: Bool { return true }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? Unit else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return self.symbol == other.symbol
+    }
 }
 
 open class Dimension : Unit {
@@ -192,6 +204,18 @@ open class Dimension : Unit {
             preconditionFailure("Unkeyed coding is unsupported.")
         }
         aCoder.encode(self.converter, forKey:"converter")
+    }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? Dimension else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return super.isEqual(object) && self.converter == other.converter
     }
 }
 
@@ -234,6 +258,18 @@ open class UnitAcceleration : Dimension {
     
     public required init?(coder aDecoder: NSCoder) { super.init(coder: aDecoder) }
     open override func encode(with aCoder: NSCoder) { super.encode(with: aCoder) }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? UnitAcceleration else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return super.isEqual(object)
+    }
 }
 
 open class UnitAngle : Dimension {
@@ -308,6 +344,18 @@ open class UnitAngle : Dimension {
     
     public required init?(coder aDecoder: NSCoder) { super.init(coder: aDecoder) }
     open override func encode(with aCoder: NSCoder) { super.encode(with: aCoder) }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? UnitAngle else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return super.isEqual(object)
+    }
 }
 
 open class UnitArea : Dimension {
@@ -446,6 +494,18 @@ open class UnitArea : Dimension {
     
     public required init?(coder aDecoder: NSCoder) { super.init(coder: aDecoder) }
     open override func encode(with aCoder: NSCoder) { super.encode(with: aCoder) }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? UnitArea else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return super.isEqual(object)
+    }
 }
 
 open class UnitConcentrationMass : Dimension {
@@ -494,6 +554,18 @@ open class UnitConcentrationMass : Dimension {
     
     public required init?(coder aDecoder: NSCoder) { super.init(coder: aDecoder) }
     open override func encode(with aCoder: NSCoder) { super.encode(with: aCoder) }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? UnitConcentrationMass else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return super.isEqual(object)
+    }
 }
 
 open class UnitDispersion : Dimension {
@@ -528,6 +600,18 @@ open class UnitDispersion : Dimension {
     
     public required init?(coder aDecoder: NSCoder) { super.init(coder: aDecoder) }
     open override func encode(with aCoder: NSCoder) { super.encode(with: aCoder) }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? UnitDispersion else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return super.isEqual(object)
+    }
 }
 
 open class UnitDuration : Dimension {
@@ -577,6 +661,18 @@ open class UnitDuration : Dimension {
     
     public required init?(coder aDecoder: NSCoder) { super.init(coder: aDecoder) }
     open override func encode(with aCoder: NSCoder) { super.encode(with: aCoder) }
+
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? UnitDuration else {
+            return false
+        }
+
+        if self === other {
+            return true
+        }
+
+        return super.isEqual(object)
+    }
 }
 
 open class UnitElectricCharge : Dimension {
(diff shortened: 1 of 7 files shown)


bubski:
@itaiferber could you please have a look at it?
To mimic your approach I had to add `required init` in some types.

Also, I'm not really sure how this loop can work
https://github.com/apple/swift/blob/master/test/stdlib/CodableTests.swift#L495
since it specializes to
```swift
expectRoundTripEqualityThroughJSON<Measurement<Dimension>>
```
thus losing the real type of associated `Unit` when decoding.
Interestingly, I ran this in playgrounds and it causes and exception.
🤔 Maybe I'm missing something.

And lastly, do you plan revisiting `Measurement : Codable`, it feels like there should be a better solution to that.
- no support for `UnitConverterReciprocal`
- no support for custom `Unit` subclasses

itaiferber:
@bubski Taking a look now. `Measurement` conformance to `Codable` is... complicated, due to the class structure we've got, and the generics involved. In fact, I would say that `Measurement` conformance in the overlay is incomplete (and if you'll notice, the unit test you mention is actually commented out at the bottom of the file).

I'll give further comments soon as I reacquaint myself with the structure of this.

itaiferber:
To give a high-level overview of why the Foundation overlay does what it does:

* The primary issue is that while `Measurement` is the Swift value type equivalent for `NSMeasurement`, `Unit` and `Dimension` are actually imported from ObjC on Darwin (they're actually `NSUnit` and `NSDimension`)
* `Unit` is the abstract-_ish_ superclass of all units; it's relatively rare to need to use it directly, and more so, relatively rare to need to subclass it for something which `Dimension` does not already cover. It's relatively rare to need to subclass `Dimension`, too, as you can create `Dimension` instances directly
* However, since `Unit` and `Dimension` are imported from ObjC, new methods on them can only be added in extensions; Swift does not allow you to override methods which come from extensions, so adopting `Codable` directly on `Unit` is dangerous because _all_ `Unit`s appear to be `Codable`, even though that behavior cannot be overridden. `Unit` could maybe handle the case for `Dimension` as well, but it wouldn't be able to handle custom unit types
* `UnitConverter` & co. have the same issue — since they are imported directly from ObjC, you can only make them `Codable` by adopting conformance in an extension, making it non-overridable
* Hence, `Measurement` does the nasty work of manually encoding these types itself, unfortunately breaking OOP encapsulation...
* As for `NSUnitConverterReciprocal` — it's not public API on Darwin, and is not exposed in Swift at all. Luckily, only one unit actually uses it, but supporting it would require awkwardly exposing this class via the runtime.

So all of this comes down to ObjC importing and how these classes are exposed in Swift. While none of these issues are present in swift-corelibs-foundation as all of the classes and types are local, I don't think we should expose `Codable` conformance on some types through swift-corelibs-foundation, while not being able to do the same through the Foundation overlay.

This implementation is unfortunately messy, but about as good as we can do at the moment. Going to go through a more specific review of this right now.

Foundation/Unit.swift
```diff
@@ -185,21 +221,45 @@ open class Dimension : Unit {
         self.converter = converter
         super.init(symbol: symbol)
     }
-    
+
+    public required init(symbol: String) {
+        fatalError("You must use the designated initializer.")
```
itaiferber: I think this error message will read better as something like `"\(type(of: self) must be initialized with designated initializer \(type(of: self)).init(symbol: String, converter: UnitConverter)"`, because... (further comment below in overrides of this method)

Foundation/Unit.swift
```diff
@@ -185,21 +221,45 @@ open class Dimension : Unit {
...
 }
 
 open class UnitAcceleration : Dimension {
     
     /*
      Base unit - metersPerSecondSquared
      */
+
+    public required init(symbol: String) {
+        fatalError("init(symbol:) has not been implemented")
```
itaiferber: ... it would now be sufficient to simply override this with `super.init(symbol: symbol)` and the error message will still be readable.

Foundation/Unit.swift
```diff
@@ -234,13 +294,33 @@ open class UnitAcceleration : Dimension {
...
+    }
 }
 
 open class UnitAngle : Dimension {
     
     /*
      Base unit - degrees
      */
+
+    public required init(symbol: String) {
```
itaiferber: If both of these methods call into super, neither needs to be overridden and can just be inherited (defining the convenience initializer below won't affect anything).

Foundation/Unit.swift
```diff
@@ -308,13 +388,33 @@ open class UnitAngle : Dimension {
...
 }
 
 open class UnitArea : Dimension {
     
     /*
      Base unit - squareMeters
      */
+
+    public required init(symbol: String) {
+        fatalError("init(symbol:) has not been implemented")
```
itaiferber: Same here (and in all repetitions below)

Foundation/Unit.swift
bubski: That's a very good suggestion. Thanks 👍

Foundation/Unit.swift
bubski: Of course! I have no idea why I assumed otherwise 😶

itaiferber: Hehe, that's alright, happens to the best of us 🙂

bubski:
Thanks for a very detailed explanation @itaiferber 🙇
I was afraid there wouldn't be an obvious better way to do this with limitations coming from deriving `NSUnit` from **ObjC**/**Foundation.**

I fully agree that we should be as aligned with **Foundation** as possible, and not adding conformance to `Codable` where not expected.
When I asked
>Are we not able to make Unit conform to Codable?

I was really asking if it would be possible in **Foundation** at all (by some other trickery than plain ObjC-Swift interop).

Then again, I'm not sure if it would be fair to force `Unit` to conform to `Codable` by design.

---

On a side note, Obj-C interop introduces lots of other weird stuff with `Measurement` in `Foundation`,
e.g.
```swift
let d = Dimension(symbol: "a") // ok
d.converter // EXC_BAD_ACCESS
```
So these `fatalError`s in "illegal" inits are not 100% aligned with **Foundation**'s behavior

or
```swift
let m1 = Measurement<UnitLength>(value: 1, unit: .miles)
let m2 = Measurement<Dimension>(value: 1, unit: Dimension(symbol: "mi", converter: UnitLength.miles.converter))
m2 == m1 // ok
m1 == m2 // EXC_BAD_ACCESS
```

Foundation/Unit.swift
bubski: I'm afraig I can't simply use `"\(type(of: self)) must be …` though, unless I fully initialize `self`.

So I think our two options are

```swift
public required init(symbol: String) {
    // instance initialized only to provide a more readable error message with class type.
    self.converter = UnitConverter()
    super.init(symbol: symbol)
    fatalError("\(type(of: self)) must be initialized with designated initializer \(type(of: self)).init(symbol: String, converter: UnitConverter)")
}
```
```swift
public required init(symbol: String) {
    fatalError("Dimension subclass must be initialized with designated initializer init(symbol: String, converter: UnitConverter)")
}
```

Which one would you recommend?

Unless there is a third way I'm missing 🤔

itaiferber: @bubski The diagnostic here is misleading; this is actually happening because string interpolation actually puts arguments in capturing autoclosures (hence the complaint).

You should be able to check the type up-front:
```swift
let T = type(of: self)
fatalError("\(T) must be initialized with designated initializer \(T).init(symbol: String, converter: UnitConverter)")
```

bubski: Ha! That's a surprise. Thanks again @itaiferber :)

itaiferber:
@bubski Regarding `Unit`: conceptually, I don't think there's any reason for it to not conform to `Codable`, short of the technical limitations. We've got other places where this is really inconvenient; in the future, it's possible that we'll work around this by introducing `class Unit` and `class Dimension` in pure Swift and having those be 1-to-1 compatible with `NSUnit` and `NSDimension` (and having the `Codable` implementation actually be inheritable), but for now, I think we're just going to have to settle for things as-is.

Re: the `fatalError`s — I would actually consider it a bug in the ObjC implementation that we don't enforce DI rules there; ObjC (like Swift) can't enforce that `Dimension(symbol: "a")` be uncallable statically, but we should certainly halt there since the result is nonsensical. There's a difference between being 100% in line with Foundation's _interface_ (swift-corelibs-foundation shouldn't offer protocol conformances/API that Foundation itself does not), and with its implementation, and I think implementation discrepancies like this are understandable. In this case, I'd say it's Foundation itself that's at fault, and the `fatalError`s are in line with how things _should_ be. So, leave them in for now.

Foundation/Unit.swift
itaiferber: Sure thing! It surprised me too — @phausler just showed this to me, so I'm just passing on the knowledge. 😉

itaiferber:
@swift-ci Please test

bubski:
@itaiferber Thanks for all the help and especially for sharing your take on the whole `Measurement` stuff so extensively. It's very interesting and helpful 👍

itaiferber:
@bubski Happy to help — thanks for taking the time and porting over all of these implementations!

itaiferber:
@swift-ci Please test

ianpartridge:
@bubski thanks for working on this! I think the only remaining one is `URLComponents` - see https://github.com/apple/swift/commit/bf4a12568aabeb30da13811aea1200d720905567

Are you willing to do this one too?

bubski:
@ianpartridge oh, I totally missed it.
Sure, I'll port it 👍

PR #1193, opened 2017-08-28
NSLock: Implement lock(before:)

spevans:
- Implemented for NSLock and NSRecursiveLock.

- Use pthread_mutex_timedlock() where available (eg Linux).

- For macOS / iOS, emulate a timed lock using a condition
  variable/mutex pair and use pthread_cond_timedwait().

- Add timeSpecFrom(date:) helper function and update
  NSCondition.wait(until:) to use it, eliminating two calls
  to gettimeofday().

The testing just consists of creating 1000 threads and then have each thread trying to get the lock in a loop along with some other contrived lock activity. Some more testing will probably need to be added to try and force some errors and also stress the thread/lock system.

`getOrAllocateTimedLockData()` needs the most attention as that uses `_stdlib_atomicCompareExchangeStrongPtr` to atomically set the shared mutex/cond pair used on macOS and the use of allocated pointers etc could be incorrect.

### DIFF
--- a/Foundation/NSLock.swift
+++ b/Foundation/NSLock.swift
@@ -1,6 +1,6 @@
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See http://swift.org/LICENSE.txt for license information
@@ -21,39 +21,71 @@ public protocol NSLocking {
     func unlock()
 }
 
-open class NSLock: NSObject, NSLocking {
 #if CYGWIN
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t?>.allocate(capacity: 1)
+private typealias _PthreadMutexPointer = UnsafeMutablePointer<pthread_mutex_t?>
+private typealias _PthreadCondPointer = UnsafeMutablePointer<pthread_cond_t?>
 #else
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t>.allocate(capacity: 1)
+private typealias _PthreadMutexPointer = UnsafeMutablePointer<pthread_mutex_t>
+private typealias _PthreadCondPointer = UnsafeMutablePointer<pthread_cond_t>
 #endif
-    
+
+open class NSLock: NSObject, NSLocking {
+    internal var mutex = _PthreadMutexPointer.allocate(capacity: 1)
+#if os(OSX) || os(iOS)
+    private var timeoutCond = _PthreadCondPointer.allocate(capacity: 1)
+    private var timeoutMutex = _PthreadMutexPointer.allocate(capacity: 1)
+#endif
+
     public override init() {
         pthread_mutex_init(mutex, nil)
+#if os(OSX) || os(iOS)
+        pthread_cond_init(timeoutCond, nil)
+        pthread_mutex_init(timeoutMutex, nil)
+#endif
     }
     
     deinit {
         pthread_mutex_destroy(mutex)
         mutex.deinitialize()
         mutex.deallocate(capacity: 1)
+#if os(OSX) || os(iOS)
+        deallocateTimedLockData(cond: timeoutCond, mutex: timeoutMutex)
+#endif
     }
     
     open func lock() {
         pthread_mutex_lock(mutex)
     }
-    
+
     open func unlock() {
         pthread_mutex_unlock(mutex)
+#if os(OSX) || os(iOS)
+        // Wakeup any threads waiting in lock(before:)
+        pthread_mutex_lock(timeoutMutex)
+        pthread_cond_broadcast(timeoutCond)
+        pthread_mutex_unlock(timeoutMutex)
+#endif
     }
-    
+
     open func `try`() -> Bool {
         return pthread_mutex_trylock(mutex) == 0
     }
     
-    open func lock(before limit: Date) {
-        NSUnimplemented()
+    open func lock(before limit: Date) -> Bool {
+        if pthread_mutex_trylock(mutex) == 0 {
+            return true
+        }
+
+#if os(OSX) || os(iOS)
+        return timedLock(mutex: mutex, endTime: limit, using: timeoutCond, with: timeoutMutex)
+#else
+        guard var endTime = timeSpecFrom(date: limit) else {
+            return false
+        }
+        return pthread_mutex_timedlock(mutex, &endTime) == 0
+#endif
     }
-    
+
     open var name: String?
 }
 
@@ -143,12 +175,12 @@ open class NSConditionLock : NSObject, NSLocking {
 }
 
 open class NSRecursiveLock: NSObject, NSLocking {
-#if CYGWIN
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t?>.allocate(capacity: 1)
-#else
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t>.allocate(capacity: 1)
+    internal var mutex = _PthreadMutexPointer.allocate(capacity: 1)
+#if os(OSX) || os(iOS)
+    private var timeoutCond = _PthreadCondPointer.allocate(capacity: 1)
+    private var timeoutMutex = _PthreadMutexPointer.allocate(capacity: 1)
 #endif
-    
+
     public override init() {
         super.init()
 #if CYGWIN
@@ -166,6 +198,9 @@ open class NSRecursiveLock: NSObject, NSLocking {
         pthread_mutex_destroy(mutex)
         mutex.deinitialize()
         mutex.deallocate(capacity: 1)
+#if os(OSX) || os(iOS)
+        deallocateTimedLockData(cond: timeoutCond, mutex: timeoutMutex)
+#endif
     }
     
     open func lock() {
@@ -174,28 +209,40 @@ open class NSRecursiveLock: NSObject, NSLocking {
     
     open func unlock() {
         pthread_mutex_unlock(mutex)
+#if os(OSX) || os(iOS)
+        // Wakeup any threads waiting in lock(before:)
+        pthread_mutex_lock(timeoutMutex)
+        pthread_cond_broadcast(timeoutCond)
+        pthread_mutex_unlock(timeoutMutex)
+#endif
     }
     
     open func `try`() -> Bool {
         return pthread_mutex_trylock(mutex) == 0
     }
     
-    open func lock(before limit: Date) {
-        NSUnimplemented()
+    open func lock(before limit: Date) -> Bool {
+        if pthread_mutex_trylock(mutex) == 0 {
+            return true
+        }
+
+#if os(OSX) || os(iOS)
+        return timedLock(mutex: mutex, endTime: limit, using: timeoutCond, with: timeoutMutex)
+#else
+        guard var endTime = timeSpecFrom(date: limit) else {
+            return false
+        }
+        return pthread_mutex_timedlock(mutex, &endTime) == 0
+#endif
     }
 
     open var name: String?
 }
 
 open class NSCondition: NSObject, NSLocking {
-#if CYGWIN
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t?>.allocate(capacity: 1)
-    internal var cond = UnsafeMutablePointer<pthread_cond_t?>.allocate(capacity: 1)
-#else
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t>.allocate(capacity: 1)
-    internal var cond = UnsafeMutablePointer<pthread_cond_t>.allocate(capacity: 1)
-#endif
-    
+    internal var mutex = _PthreadMutexPointer.allocate(capacity: 1)
+    internal var cond = _PthreadCondPointer.allocate(capacity: 1)
+
     public override init() {
         pthread_mutex_init(mutex, nil)
         pthread_cond_init(cond, nil)
@@ -221,31 +268,12 @@ open class NSCondition: NSObject, NSLocking {
     open func wait() {
         pthread_cond_wait(cond, mutex)
     }
-    
+
     open func wait(until limit: Date) -> Bool {
-        let lim = limit.timeIntervalSinceReferenceDate
-        let ti = lim - CFAbsoluteTimeGetCurrent()
-        if ti < 0.0 {
+        guard var timeout = timeSpecFrom(date: limit) else {
             return false
         }
-        var ts = timespec()
-        ts.tv_sec = Int(floor(ti))
-        ts.tv_nsec = Int((ti - Double(ts.tv_sec)) * 1_000_000_000.0)
-        var tv = timeval()
-        withUnsafeMutablePointer(to: &tv) { t in
-            gettimeofday(t, nil)
-            ts.tv_sec += t.pointee.tv_sec
-            ts.tv_nsec += Int(t.pointee.tv_usec) * 1000
-            if ts.tv_nsec >= 1_000_000_000 {
-                ts.tv_sec += ts.tv_nsec / 1_000_000_000
-                ts.tv_nsec = ts.tv_nsec % 1_000_000_000
-            }
-        }
-        let retVal: Int32 = withUnsafePointer(to: &ts) { t in
-            return pthread_cond_timedwait(cond, mutex, t)
-        }
-
-        return retVal == 0
+        return pthread_cond_timedwait(cond, mutex, &timeout) == 0
     }
     
     open func signal() {
@@ -258,3 +286,48 @@ open class NSCondition: NSObject, NSLocking {
     
     open var name: String?
 }
+
+private func timeSpecFrom(date: Date) -> timespec? {
+    guard date.timeIntervalSinceNow > 0 else {
+        return nil
+    }
+    let nsecPerSec: Int64 = 1_000_000_000
+    let interval = date.timeIntervalSince1970
+    let intervalNS = Int64(interval * Double(nsecPerSec))
+
+    return timespec(tv_sec: Int(intervalNS / nsecPerSec),
+                    tv_nsec: Int(intervalNS % nsecPerSec))
+}
+
+#if os(OSX) || os(iOS)
+
+private func deallocateTimedLockData(cond: _PthreadCondPointer, mutex: _PthreadMutexPointer) {
+    pthread_cond_destroy(cond)
+    cond.deinitialize()
+    cond.deallocate(capacity: 1)
+
+    pthread_mutex_destroy(mutex)
+    mutex.deinitialize()
+    mutex.deallocate(capacity: 1)
+}
+
+// Emulate pthread_mutex_timedlock using pthread_cond_timedwait.
+// lock(before:) passes a condition variable/mutex pair to use.
+// unlock() will use pthread_cond_broadcast() to wake any waits in progress.
+private func timedLock(mutex: _PthreadMutexPointer, endTime: Date,
+                       using timeoutCond: _PthreadCondPointer,
+                       with timeoutMutex: _PthreadMutexPointer) -> Bool {
+
+    var timeSpec = timeSpecFrom(date: endTime)
+    while var ts = timeSpec {
+        pthread_mutex_lock(timeoutMutex)
+        pthread_cond_timedwait(timeoutCond, timeoutMutex, &ts)
+        pthread_mutex_unlock(timeoutMutex)
+        if pthread_mutex_trylock(mutex) == 0 {
+            return true
+        }
+        timeSpec = timeSpecFrom(date: endTime)
+    }
+    return false
+}
+#endif
--- a/Foundation/NSLock.swift
+++ b/Foundation/NSLock.swift
@@ -320,12 +320,22 @@ private func timedLock(mutex: _PthreadMutexPointer, endTime: Date,
 
     var timeSpec = timeSpecFrom(date: endTime)
     while var ts = timeSpec {
-        pthread_mutex_lock(timeoutMutex)
-        pthread_cond_timedwait(timeoutCond, timeoutMutex, &ts)
-        pthread_mutex_unlock(timeoutMutex)
-        if pthread_mutex_trylock(mutex) == 0 {
+        let lockval = pthread_mutex_lock(timeoutMutex)
+        precondition(lockval == 0)
+        let waitval = pthread_cond_timedwait(timeoutCond, timeoutMutex, &ts)
+        precondition(waitval == 0 || waitval == ETIMEDOUT)
+        let unlockval = pthread_mutex_unlock(timeoutMutex)
+        precondition(unlockval == 0)
+
+        if waitval == ETIMEDOUT {
+            return false
+        }
+        let tryval = pthread_mutex_trylock(mutex)
+        precondition(tryval == 0 || tryval == EBUSY)
+        if tryval == 0 { // The lock was obtained.
             return true
         }
+        // pthread_cond_timedwait didnt timeout so wait some more.
         timeSpec = timeSpecFrom(date: endTime)
     }
     return false
--- a/TestFoundation/TestNSLock.swift
+++ b/TestFoundation/TestNSLock.swift
@@ -1,0 +1,125 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
+import Foundation
+import XCTest
+#else
+import SwiftFoundation
+import SwiftXCTest
+#endif
+
+import CoreFoundation
+
+class TestNSLock: XCTestCase {
+    static var allTests: [(String, (TestNSLock) -> () throws -> Void)] {
+        return [
+
+            ("test_lockWait", test_lockWait),
+            ("test_threadsAndLocks", test_threadsAndLocks),
+
+        ]
+    }
+
+
+    func test_lockWait() {
+        let condition = NSCondition()
+        let lock = NSLock()
+
+        func test(waitTime: CFTimeInterval, shouldLock: Bool) -> Bool {
+            let locked = lock.lock(before: Date.init(timeIntervalSinceNow: waitTime))
+            if locked {
+                lock.unlock()
+            }
+            return locked == shouldLock
+        }
+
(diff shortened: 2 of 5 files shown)


TestFoundation/TestNSLock.swift
```diff
@@ -0,0 +1,152 @@
...
+        ]
+    }
+
+
+    private func debugLog(_ msg: String) {
+#if false
+        NSLog(msg)
+#endif
+    }
+
```
parkera: Nit pick, but let's just leave out this kind of temporary debug stuff.

TestFoundation/TestNSLock.swift
```diff
@@ -0,0 +1,152 @@
...
+        } while totalThreads < threadCount
+        XCTAssertEqual(totalThreads, threadCount)
+
+        let gotLock = lock.try()
+        XCTAssertTrue(gotLock)
+        if gotLock {
+            lock.unlock()
+        }
+    }
+}
```
parkera: Out of curiosity, how long do these tests take to run? Is there going to be a problem with creating so many threads on some platforms?

spevans: I just ran the tests on my 2014 Macbook Pro and they took 8.008 and 10.003 seconds so about 18seconds for both tests.

1000 threads were created in .15secs. I was originally creating 10,000 but the limit looks to be 2048 total (on macOS)  and I could only create 2044. It didn't error out but I think `Thread.start` could be modified to get the error result from `pthread_create` and set the status appropriately.

I guess some low end devices will have a problem but the `threadCount` can just be lowered, the idea is really to try and stress the system to force some locking errors.

I have some more tests to add at some point for the `NSConditionLock` so I'll tidy up the debug then.

ianpartridge:
Nice. I didn't know about the stdlib intrinsics for cmpxchg.

ianpartridge:
@swift-ci please test

Foundation/NSLock.swift
```diff
@@ -36,24 +43,47 @@ open class NSLock: NSObject, NSLocking {
...
     open func lock() {
         pthread_mutex_lock(mutex)
     }
-    
+
     open func unlock() {
         pthread_mutex_unlock(mutex)
+#if os(OSX) || os(iOS)
+        if let tld = timedLockData {
+            pthread_mutex_lock(&tld.pointee.mutex)
```
phausler: I am not fully certain this is safe since swift can potentially write into the mutex memory when it takes the address of the mutex as a stack based copy

Foundation/NSLock.swift
```diff
@@ -258,3 +286,80 @@ open class NSCondition: NSObject, NSLocking {
...
+private func getOrAllocateTimedLockData(timedLockData: inout UnsafeMutablePointer<_timed_lock_data>?) -> UnsafeMutablePointer<_timed_lock_data> {
+    if timedLockData == nil, let tld = allocateTimedLockData() {
+        var expected: UnsafeMutablePointer<_timed_lock_data>? = nil
+
+        // Use an atomic compare and exchange to set the pointer to the just allocated
+        // data. On X86 this is a LOCK CMPXCHG.
+        // `object` is the target to update, timedLockData, shared by all threads.
+        // `expected` is the value the target should still be (nil).
+        // `desired` is the new value to set it to.
+        if !_stdlib_atomicCompareExchangeStrongPtr(object: &timedLockData,
```
phausler: oof! are we certain that taking the address here is safe? is it actually atomic?

using _ prefix methods from the standard library should always be done with their consent imho since we are effectively making it API.

@airspeedswift What is your take on using this?

Foundation/NSLock.swift
```diff
@@ -258,3 +286,80 @@ open class NSCondition: NSObject, NSLocking {
...
+
+// Emulate pthread_mutex_timedlock using pthread_cond_timedwait.
+// lock(before:) has obtained a shared pointer to a _time_lock_data structure
+// and the condition variable to be waited on.
+// unlock() will check for the existance of this structure and if its pointer is
+// non-nil, it will use pthread_cond_broadcast() to stop any waits in progress.
+private func timedLock(mutex: _PthreadMutexPointer,
+                       endTime: inout timespec,
+                       using tld: UnsafeMutablePointer<_timed_lock_data>) -> Bool {
+    pthread_mutex_lock(&tld.pointee.mutex)
```
phausler: again I am not certain this is safe

Foundation/NSLock.swift
spevans: I checked the disassembly (on x86_64) and it looked correct. My biggest concern was actually that the documentation seems to imply that the `expect` pointer should be heap allocated but I wasn't sure if that was correct.

I understand your point about _ prefix methods, however we use `@_` function decorators, is that not the same violation? Getting a definitive answer from the Swift core team would be good on both of these points.

I guess as an alternative this could always be done in C in CoreFoundation instead, bypassing these issues.

phausler: for the specific cases we use `@_` either I have been chastised for using that feature and eventually have been forced to remove it or we have been given special dispensation to utilize it usually with the caveat that when they change things we must change as well.

airspeedswift: We should definitely not be calling `_stdlib_atomicCompareExchangeStrongPtr` from std lib code.

airspeedswift: Checking the disassembly is only guaranteeing that it's correct _now_. It could easily change and be incorrect later if it's relying on undefined behavior. But I think (assuming the _ method use is removed) that this usage is correct so long as it is using allocated memory not a property for the `pointee`. cc @jckarter

airspeedswift: Oops that should read "We should definitely not be calling `_stdlib_atomicCompareExchangeStrongPtr` from *outside* the std lib."

spevans: @airspeedswift What would you recommend as an alternative?

spevans: I changed the `_stdlib_atomicCompareExchangeStrongPtr` to using a `lazy var`

Foundation/NSLock.swift
spevans: @phausler Who would be able to give a definitive answer as to whether it is safe or not?

jckarter: Formally it isn't guaranteed to be safe, but I can't think of anything better today.

parkera: I think we'll need to implement it in C and call out from Swift.

parkera: The performance may not be great, but that is a secondary concern to correctness.

spevans: At least doing it in C saves adding a 2nd ptr to every `Lock` allocation which will mostly be wasted (except for the locks using the `lock(before:)` method. I will look into it.

jckarter: A `static inline` C function to project the pointer to the mutex field from a pointer to the struct could also still be inlined away.

spevans:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

spevans:
@parkera I think this is ready to merge now

spevans:
@swift-ci please test

xwu:
@spevans Drive-by commentary: lazy var isn't thread-safe; did you change the use of _stdlib_atomicCompareExchangeStrongPtr to lazy var on the assumption that it is?

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -307,6 +307,21 @@ CF_EXPORT _CFThreadSpecificKey _CFThreadSpecificKeyCreate(void);
...
 typedef pthread_t _CFThreadRef;
 
+// _timed_lock_data is a condition variable and mutex pair used to wait for
+// a lock to be unlocked by another thread. See NSLock lock(before:).
+struct _timed_lock_data {
+    pthread_mutex_t mutex;
+    pthread_cond_t cond;
+};
+
+CF_INLINE pthread_mutex_t *_Nonnull _CFGetTimedLockMutex(struct _timed_lock_data *_Nonnull data) {
```
weissi: why are these needed? Isn't `&timedLockData.mutex` all you need?

Foundation/NSLock.swift
```diff
@@ -36,24 +47,51 @@ open class NSLock: NSObject, NSLocking {
...
     }
-    
+
     open func unlock() {
         pthread_mutex_unlock(mutex)
+#if os(OSX) || os(iOS)
+        // Only use _timedLockData if it was initialized by lock(before:)
+        if let tld = _timedLockData {
+            pthread_mutex_lock(_CFGetTimedLockMutex(tld))
+            pthread_cond_broadcast(_CFGetTimedLockCond(tld))
```
weissi: - why do we use `pthread_cond_broadcast` instead of `pthread_cond_signal`
 - all the return values aren't checked (it's fine to `precondition(ret == 0)`)

Foundation/NSLock.swift
```diff
@@ -143,12 +181,15 @@ open class NSConditionLock : NSObject, NSLocking {
...
 open class NSRecursiveLock: NSObject, NSLocking {
-#if CYGWIN
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t?>.allocate(capacity: 1)
-#else
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t>.allocate(capacity: 1)
+    internal var mutex = _PthreadMutexPointer.allocate(capacity: 1)
+#if os(OSX) || os(iOS)
+    private var _timedLockData: UnsafeMutablePointer<_timed_lock_data>? = nil
+    private lazy var timedLockData: UnsafeMutablePointer<_timed_lock_data>? = {
+        _timedLockData = allocateTimedLockData()
```
weissi: not thread safe

Foundation/NSLock.swift
```diff
@@ -174,28 +218,47 @@ open class NSRecursiveLock: NSObject, NSLocking {
...
     
-    open func lock(before limit: Date) {
-        NSUnimplemented()
+    open func lock(before limit: Date) -> Bool {
+        if pthread_mutex_trylock(mutex) == 0 {
+            return true
+        }
+
+        guard var endTime = timeSpecFrom(date: limit) else {
+            return false
```
weissi: why do we return `false` if `timeSpecFrom(date:)` fails? It should crash here, `false` means the timeout expired.

Foundation/NSLock.swift
```diff
@@ -174,28 +218,47 @@ open class NSRecursiveLock: NSObject, NSLocking {
...
+        }
+
+#if os(OSX) || os(iOS)
+        if let tld = timedLockData {
+            return timedLock(mutex: mutex, endTime: &endTime, using: tld)
+        } else {
+            return false
+        }
+#else
+        return pthread_mutex_timedlock(mutex, &endTime) == 0
```
weissi: that puts all errors in the same category. Should be

```
            switch pthread_mutex_timedlock(mutex, &endTime) {
            case 0:
                return true
            case ETIMEDOUT:
                return false
            case let e:
                fatalError("caught error \(e) when calling pthread_cond_timedwait")
            }
```

Foundation/NSLock.swift
```diff
@@ -221,31 +284,12 @@ open class NSCondition: NSObject, NSLocking {
...
-                ts.tv_sec += ts.tv_nsec / 1_000_000_000
-                ts.tv_nsec = ts.tv_nsec % 1_000_000_000
-            }
-        }
-        let retVal: Int32 = withUnsafePointer(to: &ts) { t in
-            return pthread_cond_timedwait(cond, mutex, t)
-        }
-
-        return retVal == 0
+        return pthread_cond_timedwait(cond, mutex, &timeout) == 0
```
weissi: same here, errors besides `ETIMEDOUT` should be handled separately.

Foundation/NSLock.swift
```diff
@@ -258,3 +302,59 @@ open class NSCondition: NSObject, NSLocking {
...
+
+private func timeSpecFrom(date: Date) -> timespec? {
+    let interval = date.timeIntervalSince1970
+    guard interval >= 0 else {
+        return nil
+    }
+
+    var ts = timespec()
+    ts.tv_sec = Int(interval.rounded(.down))
+    ts.tv_nsec = Int((interval - Double(ts.tv_sec)) * 1_000_000_000.0)
```
weissi: this looks rather dangerous. If for floating point reasons `ts.tv_nsec` will be `1_000_000_000` or larger, all the timed wait functions will just return `EINVAL` IIRC.

Why not something similar to

```
private func timeSpecFrom(date: Date) -> timespec {
        guard interval >= 0 else { fatalError ("current time negative") }
        let nsecPerSec: Int64 = 1000000000
        let timeoutNS = Int64(interval * Double(nsecPerSec))

        var timeoutAbs = timespec(tv_sec: Int((timeoutNS / nsecPerSec)),
                                  tv_nsec: Int(timeoutNS % nsecPerSec))
        assert(timeoutAbs.tv_nsec >= 0 && timeoutAbs.tv_nsec < Int(nsecPerSec))
        assert(timeoutAbs.tv_sec >= curTime.tv_sec)
        return timeoutAbs
}
```

Foundation/NSLock.swift
```diff
@@ -258,3 +302,59 @@ open class NSCondition: NSObject, NSLocking {
...
+
+#if os(OSX) || os(iOS)
+
+private func allocateTimedLockData() -> UnsafeMutablePointer<_timed_lock_data>? {
+    let tld = UnsafeMutablePointer<_timed_lock_data>.allocate(capacity: 1)
+    if pthread_mutex_init(_CFGetTimedLockMutex(tld), nil) == 0 {
+        if pthread_cond_init(_CFGetTimedLockCond(tld), nil) == 0 {
+            return tld
+        } else {
+            pthread_mutex_destroy(_CFGetTimedLockMutex(tld))
```
weissi: need to check the error code

Foundation/NSLock.swift
```diff
@@ -258,3 +302,59 @@ open class NSCondition: NSObject, NSLocking {
...
+
+    var ts = timespec()
+    ts.tv_sec = Int(interval.rounded(.down))
+    ts.tv_nsec = Int((interval - Double(ts.tv_sec)) * 1_000_000_000.0)
+    return ts
+}
+
+#if os(OSX) || os(iOS)
+
+private func allocateTimedLockData() -> UnsafeMutablePointer<_timed_lock_data>? {
```
weissi: in Swift, failure to allocate memory usually results in a crash. Why does this return an Optional instead? What's the caller supposed to do if this fails?

Foundation/NSLock.swift
```diff
@@ -258,3 +302,59 @@ open class NSCondition: NSObject, NSLocking {
...
+            return tld
+        } else {
+            pthread_mutex_destroy(_CFGetTimedLockMutex(tld))
+        }
+    }
+    tld.deallocate(capacity: 1)
+    return nil
+}
+
+private func deallocateTimedLockData(_ tld: UnsafeMutablePointer<_timed_lock_data>?) {
```
weissi: shouldn't take an optional

Foundation/NSLock.swift
```diff
@@ -258,3 +302,59 @@ open class NSCondition: NSObject, NSLocking {
...
+
+// Emulate pthread_mutex_timedlock using pthread_cond_timedwait.
+// lock(before:) has obtained a shared pointer to a _time_lock_data structure
+// and the condition variable to be waited on.
+// unlock() will check for the existance of this structure and if its pointer is
+// non-nil, it will use pthread_cond_broadcast() to stop any waits in progress.
+private func timedLock(mutex: _PthreadMutexPointer,
+                       endTime: inout timespec,
+                       using tld: UnsafeMutablePointer<_timed_lock_data>) -> Bool {
+    pthread_mutex_lock(_CFGetTimedLockMutex(tld))
```
weissi: errors not checked

Foundation/NSLock.swift
```diff
@@ -258,3 +302,59 @@ open class NSCondition: NSObject, NSLocking {
...
+    let status = pthread_cond_timedwait(_CFGetTimedLockCond(tld),
+                                        _CFGetTimedLockMutex(tld),
+                                        &endTime)
+    pthread_mutex_unlock(_CFGetTimedLockMutex(tld))
```
weissi: wait, why are we unlocking here?

Foundation/NSLock.swift
```diff
@@ -258,3 +302,59 @@ open class NSCondition: NSObject, NSLocking {
...
+
+    if status == 0 {
+        return pthread_mutex_trylock(mutex) == 0
```
weissi: why are we trying to re-lock here?

Foundation/NSLock.swift
spevans: Crashing is a little extreme because a timeout was exceeded. The documentation at https://developer.apple.com/documentation/foundation/nslock/1411133-lock states that the return value is `true if the lock is acquired before limit, otherwise false.`
The `lock(before:)` could have been called with a perfectly valid date in the future but could have expired before it is processed (e.g. if it was something in the near future like 3ns)

Foundation/NSLock.swift
spevans: This takes an optional to simplify the `deinit`. Its a `private` function so its not exposing any sort of api.

Foundation/NSLock.swift
spevans: This was copied from the `-[NSLock lock(before:)]` function but you are correct, it should be stricter with the checking, I will update it to call `fatalError()` if anything fails.

Foundation/NSLock.swift
spevans: `pthread_cond_timedwait` requires the lock so it needs to be unlocked.

Foundation/NSLock.swift
phausler: if pthread_mutex_lock fails I would consider the host OS to potentially be defective; The errors are EINVAL - which means the lock is corrupt, or EDEADLK which only happens when the lock is initialized with the parameters PTHREAD_MUTEX_ERRORCHECK

Foundation/NSLock.swift
phausler: not needed; the failure cases are EINVAL for a corrupted lock, or EBUSY which only happens in the cases of PTHREAD_MUTEX_ERRORCHECK

Foundation/NSLock.swift
phausler: there may be multiple waiters on a given NSLock so broadcast is the correct idiom

Foundation/NSLock.swift
spevans: This is how the lock is actually acquired. The `pthread_mutex_lock` / `pthread_cond_timedwait` / `pthread_mutex_unlock` merely handles the waiting/timeout, the `pthread_mutex_trylock` is the actual (non-waiting) lock.

Foundation/NSLock.swift
```diff
@@ -21,13 +21,24 @@ public protocol NSLocking {
...
+private typealias _PthreadMutexPointer = UnsafeMutablePointer<pthread_mutex_t>
+private typealias _PthreadCondPointer = UnsafeMutablePointer<pthread_cond_t>
 #endif
-    
+
+open class NSLock: NSObject, NSLocking {
+    internal var mutex = _PthreadMutexPointer.allocate(capacity: 1)
+#if os(OSX) || os(iOS)
+    private var _timedLockData: UnsafeMutablePointer<_timed_lock_data>? = nil
+    private lazy var timedLockData: UnsafeMutablePointer<_timed_lock_data>? = {
```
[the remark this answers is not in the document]
spevans: What is the correct way of doing an implicit `swift_once()` with `var`s? Or should I use an explicit `dispatch_once()`?

Foundation/NSLock.swift
weissi: @phausler please correct me if I'm wrong but AFAIK _any_ of the waiters will be able to make progress and also after every unlock we run that code. So why do we need to wake up _all_ and not just one?

Foundation/NSLock.swift
weissi: but we want to know this, right? What if we have memory corruption or something. We should just `preconditon()` or at the very least `assert()` that the return value is 0.

Foundation/NSLock.swift
weissi: so why not just `precondition` or `assert`? Just to be sure.

Foundation/NSLock.swift
weissi: @spevans sorry I think I'm slow. So I agree that `pthread_cond_timedwait` needs the lock to be locked which is the case. But the function is called `timedLock` so when the function returns, the lock should still be locked, right? Otherwise the function returns and the lock isn't actually locked.

Foundation/NSLock.swift
weissi: @spevans sorry, now I'm completely lost, what is a non-waiting lock? And why do we need two locks for a timed lock?

Foundation/NSLock.swift
weissi: @spevans sorry, my comment is totally bogus. `timeSpecFrom(date:)` doesn't return `false` if the timeout expired but if it couldn't convert the timeout. That should not be possible. The right fix is obvs to make `timeSpecFrom(date:)` not even return a `Bool` as it should always succeed. Then we also don't need this `return` here.

Foundation/NSLock.swift
weissi: fair enough

Foundation/NSLock.swift
spevans: On macOS there are only two ways to lock: `pthread_mutex_lock` which waits indefinitely and `pthread_mutex_trylock` which tries to lock and returns immediately whether it locked or not (i.e. non-waiting).

`pthread_cond_timewait` uses a `pthread_mutex_t`/`pthread_cond_t` pair to perform its actions. So there are actually 2 mutexes and one condition variable involved here.

`mutex` is the actual lock that is trying to be acquired. The `_timed_lock_data` struct holds the `pthread_mutex_t`/`pthread_cond_t` pair for the `pthread_cond_timewait`.
Does that make sense?

weissi: @spevans apologies, I didn't quite read the code correctly, thought this is all the same mutex :P.

Ok, in that case I still think it's not quite right as the `pthread_mutex_unlock(_CFGetTimedLockMutex(tld))` is done too early. If I'm not mistaken what could happen is the following: 

- thread 1 progresses until line  354 with `status == 0`
- thread 1 is now scheduled off CPU
- thread 2 arrives and also progresses until line 354 with `status == 0`
- thread 2 runs the function to completion and gets the lock `mutex` (with `pthread_mutex_trylock(mutex) == 0`)
- now thread 1 gets scheduled again and also tries to get the lock (with `pthread_mutex_trylock(mutex) == 0`) but because it's already taken by thread 2, it doesn't get it

now, in thread 1 `timedLock(...)` returns `false` even though the timeout might not have expired.


The same problem can happen if `pthread_cond_timedwait` spuriously wakes up. It's allowed to do that. Basically conditional variables are hardly ever used correctly if they're not in a loop. See [here](https://linux.die.net/man/3/pthread_cond_wait)

> Spurious wakeups from the pthread_cond_timedwait() or pthread_cond_wait() functions may occur.

please let me  know if I missed anything again and sorry agaiin

Foundation/NSLock.swift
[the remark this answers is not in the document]
jckarter: The `once` APIs only work with global variables. You'll need some other form of synchronization, like a lock.

Foundation/NSLock.swift
spevans: @weissi Yes you are correct in that I have made a small error assuming that as soon as one thread gets the lock the others should return `false`, when they should actually retry the loop until the `limit: Date` is reached. Adding a loop would fix this and also work around any spurious wake ups.

I will add a fix for the above, however note that this solution is only used on macOS and the manual page for `pthread_cond_timedwait` doesn't actually mention spurious wakeups, it will only return on an invalid argument or a timeout.

On Linux, all of this is replaced by a single call of `pthread_mutex_timedlock`.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
I updated this to simplify the cond variable/mutex pair to always be allocated. Its makes the `unlock` slightly less efficient on macOS since the `pthread_cond_broadcast` always happens now but it should fix any thread safety issues due to the lazy allocation. A more better solution can alway be used in a follow up PR.

@weissi I added a loop around the `pthread_cond_timedwait` to check that the end time is reached which should solve any issues of early wakeup. I also modified the `timeSpecFrom(date:)` to use integers as I also saw a potential issue with the use of `Double` overflowing the `tv_nsec` field.

Foundation/NSLock.swift
```diff
@@ -258,3 +286,48 @@ open class NSCondition: NSObject, NSLocking {
...
+// lock(before:) passes a condition variable/mutex pair to use.
+// unlock() will use pthread_cond_broadcast() to wake any waits in progress.
+private func timedLock(mutex: _PthreadMutexPointer, endTime: Date,
+                       using timeoutCond: _PthreadCondPointer,
+                       with timeoutMutex: _PthreadMutexPointer) -> Bool {
+
+    var timeSpec = timeSpecFrom(date: endTime)
+    while var ts = timeSpec {
+        pthread_mutex_lock(timeoutMutex)
+        pthread_cond_timedwait(timeoutCond, timeoutMutex, &ts)
```
weissi: please, can we `precondition` that the return value of `pthread_cond_timedwait` is either `0` or `ETIMEDOUT` and nothing else.

Also we should `precondition` (at least `assert`) that `pthread_mutex_(un)lock` return `0` in every case.

Foundation/NSLock.swift
```diff
@@ -21,39 +21,71 @@ public protocol NSLocking {
...
-    internal var mutex = UnsafeMutablePointer<pthread_mutex_t>.allocate(capacity: 1)
+private typealias _PthreadMutexPointer = UnsafeMutablePointer<pthread_mutex_t>
+private typealias _PthreadCondPointer = UnsafeMutablePointer<pthread_cond_t>
 #endif
-    
+
+open class NSLock: NSObject, NSLocking {
+    internal var mutex = _PthreadMutexPointer.allocate(capacity: 1)
+#if os(OSX) || os(iOS)
+    private var timeoutCond = _PthreadCondPointer.allocate(capacity: 1)
```
weissi: wouldn't it make sense to unify this code and use the more portable BSD/Darwin code for all platforms?

Foundation/NSLock.swift
```diff
@@ -21,39 +21,71 @@ public protocol NSLocking {
...
+    private var timeoutMutex = _PthreadMutexPointer.allocate(capacity: 1)
+#endif
+
     public override init() {
         pthread_mutex_init(mutex, nil)
+#if os(OSX) || os(iOS)
+        pthread_cond_init(timeoutCond, nil)
```
weissi: why aren't we checking the return values of `pthread_*` anywhere really?

spevans:
@swift-ci please test

alblue:

PR #1197, opened 2017-09-01
JSONSerialization: Add missing numeric types

spevans:
- Add support for Int8/Int16/Int32/Int64, UInt8/UInt16/UInt32/UInt64,
  Decimal and NSDecimalNumber.
- Improve serializeInteger performance to avoid overall speed regression.
- Move JSONSerialization tests into correct testfile and use Int64 to stop tests
   breaking on 32bit platforms.
- Add tests for isValidJSONObject() to test the support of new types.
- Add tests for Float, Double, Decimal and NSDecimalNumber.

This PR does not address the output format of `Float` and `Double`, which are not currently the same as `Foundation`. That will be addressed in a follow-up PR

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -63,7 +63,9 @@ open class JSONSerialization : NSObject {
             }
             
             // object is Swift.String, NSNull, Int, Bool, or UInt
-            if obj is String || obj is NSNull || obj is Int || obj is Bool || obj is UInt {
+            if obj is String || obj is NSNull || obj is Int || obj is Bool || obj is UInt ||
+                obj is Int8 || obj is Int16 || obj is Int32 || obj is Int64 ||
+                obj is UInt8 || obj is UInt16 || obj is UInt32 || obj is UInt64 {
                 return true
             }
 
@@ -73,6 +75,10 @@ open class JSONSerialization : NSObject {
             }
             // object is a Float and is not NaN or infinity
             if let number = obj as? Float  {
+                return number.isFinite
+            }
+
+            if let number = obj as? Decimal {
                 return number.isFinite
             }
 
@@ -99,8 +105,11 @@ open class JSONSerialization : NSObject {
             // object is NSNumber and is not NaN or infinity
             // For better performance, this (most expensive) test should be last.
             if let number = _SwiftValue.store(obj) as? NSNumber {
-                let invalid = number.doubleValue.isInfinite || number.doubleValue.isNaN
-                return !invalid
+                if CFNumberIsFloatType(number._cfObject) {
+                    return number.doubleValue.isFinite
+                } else {
+                    return true
+                }
             }
 
             // invalid object
@@ -285,9 +294,7 @@ internal extension JSONSerialization {
 
 //MARK: - JSONSerializer
 private struct JSONWriter {
-    
-    private let maxUIntLength = String(describing: UInt.max).count
-    private let maxIntLength = String(describing: Int.max).count
+
     var indent = 0
     let pretty: Bool
     let sortedKeys: Bool
@@ -320,13 +327,37 @@ private struct JSONWriter {
         case let boolValue as Bool:
             serializeBool(boolValue)
         case let num as Int:
-            try serializeInt(value: num)
+            serializeInteger(value: num)
+        case let num as Int8:
+            serializeInteger(value: num)
+        case let num as Int16:
+            serializeInteger(value: num)
+        case let num as Int32:
+            serializeInteger(value: num)
+        case let num as Int64:
+            serializeInteger(value: num)
         case let num as UInt:
-            try serializeUInt(value: num)
+            serializeInteger(value: num)
+        case let num as UInt8:
+            serializeInteger(value: num)
+        case let num as UInt16:
+            serializeInteger(value: num)
+        case let num as UInt32:
+            serializeInteger(value: num)
+        case let num as UInt64:
+            serializeInteger(value: num)
         case let array as Array<Any?>:
             try serializeArray(array)
         case let dict as Dictionary<AnyHashable, Any?>:
             try serializeDictionary(dict)
+        case let num as Float:
+            try serializeNumber(NSNumber(value: num))
+        case let num as Double:
+            try serializeNumber(NSNumber(value: num))
+        case let num as Decimal:
+            writer(num.description)
+        case let num as NSDecimalNumber:
+            writer(num.description)
         case is NSNull:
             try serializeNull()
         case _ where _SwiftValue.store(obj) is NSNumber:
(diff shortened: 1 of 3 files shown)


Foundation/JSONSerialization.swift
```diff
@@ -63,7 +63,9 @@ open class JSONSerialization : NSObject {
             }
             
             // object is Swift.String, NSNull, Int, Bool, or UInt
```
ianpartridge: Delete this comment, which is now wrong (and was pointless anyway)?

Foundation/JSONSerialization.swift
```diff
@@ -336,92 +367,33 @@ private struct JSONWriter {
...
-            case 6: stringResult.append("6")
-            case 7: stringResult.append("7")
-            case 8: stringResult.append("8")
-            case 9: stringResult.append("9")
-            default: fatalError()
-            }
-        }
-        
-        writer(stringResult)
+        let output = String(cString: Array(buffer.suffix(from: pos)))
```
ianpartridge: Won't this do copies? One to create a new `Array` and another to create the `String`. Can we avoid them?

Foundation/JSONSerialization.swift
```diff
@@ -462,15 +434,20 @@ private struct JSONWriter {
...
-        }
-        
-        switch num._cfTypeID {
-        case CFBooleanGetTypeID():
-            serializeBool(num.boolValue)
-        default:
-            writer(_serializationString(for: num))
+        if CFNumberIsFloatType(num._cfObject) {
+            if !num.doubleValue.isFinite {
+                throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Number cannot be infinity or NaN"])
```
ianpartridge: Perhaps this should say "Number must be finite" now. Or is the string exactly matching Darwin?

ianpartridge:
@swift-ci please test

Foundation/JSONSerialization.swift
spevans: Originally I was allocating a buffer once in the `JSONWriter()` and using `String(bytesNoCopy:)` but even that does a copy and then there is another one when the string is appended to the JSON output.
So I decided not to do anything too complicated in this PR and just copied this from `Decimal`.

In a followup PR I will try converting `JSONWriter` to allocate the output buffer at the start and just have the serialise methods write directly into it and hopefully avoid any copying.

@djones6 Ran some tests against this change and confirmed that it is faster than the current method.

Foundation/JSONSerialization.swift
spevans: I updated it to emit the same message as Darwin

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

Foundation/JSONSerialization.swift
ianpartridge: Writing directly into the output buffer sounds much better - thanks.

TestFoundation/TestJSONEncoder.swift
```diff
@@ -405,64 +405,30 @@ class TestJSONEncoder : XCTestCase {
...
     func test_codingOfUIntMinMax() {
 
-        let encoder = JSONEncoder()
-
         struct MyValue: Codable {
-            let intMin:Int = Int.min
-            let intMax:Int = Int.max
-            let uintMin:UInt = UInt.min
-            let uintMax:UInt = UInt.max
+            let intMin = Int64.min
```
ianpartridge: Perhaps this would be better called `int64Min` now. No big deal though.

Foundation/JSONSerialization.swift
```diff
@@ -336,92 +366,33 @@ private struct JSONWriter {
...
-        //Maximum length of an UInt
-        array.reserveCapacity(maxUIntLength)
-        stringResult.reserveCapacity(maxUIntLength)
-        var number = value
-        
-        while number != 0 {
-            array.append(number % 10)
+    private func serializeInteger<T: UnsignedInteger>(value: T, isNegative: Bool = false) {
+        let maxIntLength = 22   // 20 digits in UInt64 + optional sign + trailing '\0'
+        let ZERO: CChar = 0x30  // ASCII '0' == 0x30
```
itaiferber: From a purely stylistic standpoint, I don't think we really use all caps anywhere to indicate a constant. I think this would read more clearly as just `asciiZero` (and `MINUS` would read a bit better as `asciiMinus`).

Foundation/JSONSerialization.swift
```diff
@@ -462,15 +433,30 @@ private struct JSONWriter {
...
-        case CFBooleanGetTypeID():
-            serializeBool(num.boolValue)
-        default:
-            writer(_serializationString(for: num))
+        if CFNumberIsFloatType(num._cfObject) {
+            let dv = num.doubleValue
+            if !dv.isFinite {
+                let value: String
+                if dv.isNaN {
+                    value = "NaN"
```
itaiferber: This is inconsistent with `NSJSONSerialization`, which does not support `NaN` or infinite values. This should arguably `preconditionFailure`/`fatalError`.

TestFoundation/TestJSONSerialization.swift
```diff
@@ -1172,7 +1207,80 @@ extension TestJSONSerialization {
...
+    func test_serialize_64BitSizes() {
+        let json1 = [Int64.min, Int64(-1), Int64(0), Int64(1), Int64.max]
+        XCTAssertEqual(try trySerialize(json1), "[-9223372036854775808,-1,0,1,9223372036854775807]")
+        let json2 = [UInt64.min, UInt64(0), UInt64(1), UInt64.max]
+        XCTAssertEqual(try trySerialize(json2), "[0,0,1,18446744073709551615]")
+    }
+
+    func test_serialize_Float() {
+        XCTAssertEqual(try trySerialize([-Float.leastNonzeroMagnitude, Float.leastNonzeroMagnitude]), "[-0,0]")
+        XCTAssertEqual(try trySerialize([-Float.greatestFiniteMagnitude]), "[-340282346638529000000000000000000000000]")
```
itaiferber: This value (and all floating-point values below, actually) is inconsistent with `NSJSONSerialization`, which outputs floating-point numbers in scientific notation (via `snprintf("%*.g", DBL_DECIMAL_DIG, ...)`). We should likely bridge that gap.

Foundation/JSONSerialization.swift
```diff
@@ -99,8 +104,11 @@ open class JSONSerialization : NSObject {
             // object is NSNumber and is not NaN or infinity
             // For better performance, this (most expensive) test should be last.
             if let number = _SwiftValue.store(obj) as? NSNumber {
-                let invalid = number.doubleValue.isInfinite || number.doubleValue.isNaN
-                return !invalid
+                if CFNumberIsFloatType(number._cfObject) {
+                    return number.doubleValue.isFinite
```
itaiferber: This line made me do a double-take for a moment because at first glance, it's hard to visually distinguish between `number.doubleValue.isFinite` and `number.doubleValue.isInfinite`. Do you mind just calling this out with a comment saying something like "All floating-point numbers except for NaN and infinity are valid." (or similar)?

Foundation/JSONSerialization.swift
itaiferber: (Seeing also as we claim in `isValidJSONValue` that infinite numbers/`NaN` are not valid, we shouldn't support them here.)

TestFoundation/TestJSONSerialization.swift
```diff
@@ -924,6 +925,32 @@ extension TestJSONSerialization {
...
+
+        XCTAssertTrue(JSONSerialization.isValidJSONObject([NSNumber(value: true), NSNumber(value: Float.greatestFiniteMagnitude), NSNumber(value: Double.greatestFiniteMagnitude)]))
+        XCTAssertTrue(JSONSerialization.isValidJSONObject([NSNumber(value: Int.max), NSNumber(value: Int8.max), NSNumber(value: Int16.max), NSNumber(value: Int32.max), NSNumber(value: Int64.max)]))
+        XCTAssertTrue(JSONSerialization.isValidJSONObject([NSNumber(value: UInt.max), NSNumber(value: UInt8.max), NSNumber(value: UInt16.max), NSNumber(value: UInt32.max), NSNumber(value: UInt64.max)]))
+        XCTAssertTrue(JSONSerialization.isValidJSONObject([NSDecimalNumber(booleanLiteral: true), NSDecimalNumber(decimal: Decimal.greatestFiniteMagnitude), NSDecimalNumber(floatLiteral: Double.greatestFiniteMagnitude), NSDecimalNumber(integerLiteral: Int.min)]))
+        XCTAssertTrue(JSONSerialization.isValidJSONObject([Decimal(123), Decimal(Double.leastNonzeroMagnitude)]))
+
+        XCTAssertFalse(JSONSerialization.isValidJSONObject(Float.nan))
+        XCTAssertFalse(JSONSerialization.isValidJSONObject(Float.infinity))
```
itaiferber: For completeness' sake, it's worth also asserting this on negative infinity (below as well).

Foundation/JSONSerialization.swift
spevans: Sure, I will actually put it back to how it was but store the `doubleValue` in a local since getting the `doubleValue` twice involved into 2 calls into `CFNumber` so was a significant speedup.

Foundation/JSONSerialization.swift
spevans: This is actually to exactly match the error messaging on Darwin, not to serialise the value. `value` is used in the `NSError` below.

itaiferber: You're totally right, of course — I'd misread this. 😬

TestFoundation/TestJSONSerialization.swift
spevans: I did try and play around with the `Float` and `Double` output formats but it broke some other tests. I will see if I can find a better solution - I was actually intending to do a follow-up PR to fix up the output formatting.

Foundation/JSONSerialization.swift
spevans: I actually copied this from `Decimal` but I will tidy it up.

spevans:
I have fixed all of the outstanding comments expect for the `Float`/`Double` output format which I will fix in a followup PR.

itaiferber:
@swift-ci Please test

spevans:
@itaiferber Looks like the test worked on CI, is this ok to merge now?

itaiferber:
@spevans Which builder did this build on? I don't see any GitHub CI state for it

PR #1198, opened 2017-09-02
NSLocalizedDescription for http errors taken from libcurl

johnno1962:
Hi Apple,

At the moment any network error using a URLSession contains only the default "The operation could not be completed” as its NSLocalizedDescription. This PR conveys the value of curl_easy_strerror through the implementation so something more specific is reported such as "Foundation.URLError(_nsError: Peer certificate cannot be authenticated with given CA certificates)"

Thanks.

### DIFF
--- a/CoreFoundation/URL.subproj/CFURLSessionInterface.c
+++ b/CoreFoundation/URL.subproj/CFURLSessionInterface.c
@@ -19,6 +19,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "CFURLSessionInterface.h"
+#include <CoreFoundation/CFString.h>
 #include <curl/curl.h>
 
 FILE* aa = NULL;
@@ -31,6 +32,11 @@ static CFURLSessionMultiCode MakeMultiCode(CURLMcode value) {
     return (CFURLSessionMultiCode) { value };
 }
 
+CFStringRef CFURLSessionCreateErrorDescription(int value) {
+    const char *description = curl_easy_strerror(value);
+    return CFStringCreateWithBytes(kCFAllocatorSystemDefault,
+        (const uint8_t *)description, strlen(description), kCFStringEncodingUTF8, NO);
+}
 
 CFURLSessionEasyHandle _Nonnull CFURLSessionEasyHandleInit() {
     return curl_easy_init();
@@ -135,6 +141,7 @@ CFURLSessionEasyCode CFURLSessionInit(void) {
     return MakeEasyCode(curl_global_init(CURL_GLOBAL_SSL));
 }
 
+int const CFURLSessionEasyErrorSize = { CURL_ERROR_SIZE + 1 };
 
 CFURLSessionEasyCode const CFURLSessionEasyCodeOK = { CURLE_OK };
 CFURLSessionEasyCode const CFURLSessionEasyCodeUNSUPPORTED_PROTOCOL = { CURLE_UNSUPPORTED_PROTOCOL };
--- a/CoreFoundation/URL.subproj/CFURLSessionInterface.h
+++ b/CoreFoundation/URL.subproj/CFURLSessionInterface.h
@@ -48,6 +48,10 @@ typedef struct CFURLSessionEasyCode {
     int value;
 } CFURLSessionEasyCode;
 
+CF_EXPORT CFStringRef _Nonnull CFURLSessionCreateErrorDescription(int value);
+
+CF_EXPORT int const CFURLSessionEasyErrorSize;
+
 /// CURLcode
 CF_EXPORT CFURLSessionEasyCode const CFURLSessionEasyCodeOK; // CURLE_OK
 CF_EXPORT CFURLSessionEasyCode const CFURLSessionEasyCodeUNSUPPORTED_PROTOCOL; // CURLE_UNSUPPORTED_PROTOCOL
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -20,7 +20,7 @@ extension unichar : ExpressibleByUnicodeScalarLiteral {
     }
 }
 
-// Placeholder for a future implementation
+/// Returns a localized string, using the main bundle if one is not specified.
 public
 func NSLocalizedString(_ key: String,
                        tableName: String? = nil,
--- a/Foundation/URLSession/http/EasyHandle.swift
+++ b/Foundation/URLSession/http/EasyHandle.swift
@@ -56,6 +56,7 @@ internal final class _EasyHandle {
     fileprivate var headerList: _CurlStringList?
     fileprivate var pauseState: _PauseState = []
     internal var timeoutTimer: _TimeoutSource!
+    internal lazy var errorBuffer = [UInt8](repeating: 0, count: Int(CFURLSessionEasyErrorSize))
     #if os(Android)
     static fileprivate var _CAInfoFile: UnsafeMutablePointer<Int8>?
     #endif
@@ -89,8 +90,8 @@ extension _EasyHandle {
 }
 
 internal extension _EasyHandle {
-    func completedTransfer(withErrorCode errorCode: Int?) {
-        delegate?.transferCompleted(withErrorCode: errorCode)
+    func completedTransfer(withError error: NSError?) {
+        delegate?.transferCompleted(withError: error)
     }
 }
 internal protocol _EasyHandleDelegate: class {
@@ -107,7 +108,7 @@ internal protocol _EasyHandleDelegate: class {
     func fill(writeBuffer buffer: UnsafeMutableBufferPointer<Int8>) -> _EasyHandle._WriteBufferResult
     /// The transfer for this handle completed.
     /// - parameter errorCode: An NSURLError code, or `nil` if no error occured.
-    func transferCompleted(withErrorCode errorCode: Int?)
+    func transferCompleted(withError error: NSError?)
     /// Seek the input stream to the given position
     func seekInputStream(to position: UInt64) throws
     /// Gets called during the transfer to update progress.
@@ -146,7 +147,8 @@ extension _EasyHandle {
     /// Set error buffer for error messages
     /// - SeeAlso: https://curl.haxx.se/libcurl/c/CURLOPT_ERRORBUFFER.html
     func set(errorBuffer buffer: UnsafeMutableBufferPointer<UInt8>?) {
-        try! CFURLSession_easy_setopt_ptr(rawHandle, CFURLSessionOptionERRORBUFFER, buffer?.baseAddress ?? nil).asError()
+        let buffer = buffer ?? errorBuffer.withUnsafeMutableBufferPointer { $0 }
+        try! CFURLSession_easy_setopt_ptr(rawHandle, CFURLSessionOptionERRORBUFFER, buffer.baseAddress).asError()
     }
     /// Request failure on HTTP response >= 400
     func set(failOnHTTPErrorCode flag: Bool) {
--- a/Foundation/URLSession/http/HTTPURLProtocol.swift
+++ b/Foundation/URLSession/http/HTTPURLProtocol.swift
@@ -134,7 +134,9 @@ fileprivate extension _HTTPURLProtocol {
             //     NSURLErrorNoPermissionsToReadFile
             //     NSURLErrorFileDoesNotExist
             self.internalState = .transferFailed
-            failWith(errorCode: errorCode(fileSystemError: e), request: request)
+            let error = NSError(domain: NSURLErrorDomain, code: errorCode(fileSystemError: e),
+                                userInfo: [NSLocalizedDescriptionKey: "File system error"])
+            failWith(error: error, request: request)
             return
         }
 
@@ -328,17 +330,19 @@ internal extension _HTTPURLProtocol {
         case stream(InputStream)
     }
 
-    func failWith(errorCode: Int, request: URLRequest) {
+    func failWith(error: NSError, request: URLRequest) {
         //TODO: Error handling
         let userInfo: [String : Any]? = request.url.map {
             [
+                NSUnderlyingErrorKey: error,
                 NSURLErrorFailingURLErrorKey: $0,
                 NSURLErrorFailingURLStringErrorKey: $0.absoluteString,
+                NSLocalizedDescriptionKey: NSLocalizedString(error.localizedDescription, comment: "N/A")
                 ]
         }
-        let error = URLError(_nsError: NSError(domain: NSURLErrorDomain, code: errorCode, userInfo: userInfo))
-        completeTask(withError: error)
-        self.client?.urlProtocol(self, didFailWithError: error)
+        let urlError = URLError(_nsError: NSError(domain: NSURLErrorDomain, code: error.code, userInfo: userInfo))
+        completeTask(withError: urlError)
+        self.client?.urlProtocol(self, didFailWithError: urlError)
     }
 }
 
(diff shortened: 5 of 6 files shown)


Foundation/URLSession/http/EasyHandle.swift
```diff
@@ -170,6 +171,9 @@ extension _EasyHandle {
         let protocols = (CFURLSessionProtocolHTTP | CFURLSessionProtocolHTTPS)
         try! CFURLSession_easy_setopt_long(rawHandle, CFURLSessionOptionPROTOCOLS, protocols).asError()
         try! CFURLSession_easy_setopt_long(rawHandle, CFURLSessionOptionREDIR_PROTOCOLS, protocols).asError()
+        errorBuffer.withUnsafeMutableBufferPointer {
```
ianpartridge: `setAllowedProtocolsToHTTPAndHTTPS()` doesn't seem to be an accurate name for this function any more.

Foundation/URLSession/http/HTTPURLProtocol.swift
```diff
@@ -575,6 +576,10 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
     }
 }
 
+internal struct URLErrorInfo {
+    let code: Int, description: String?
```
ianpartridge: Put these on separate lines.

Foundation/URLSession/http/HTTPURLProtocol.swift
```diff
@@ -330,12 +330,13 @@ internal extension _HTTPURLProtocol {
         case stream(InputStream)
     }
 
-    func failWith(errorCode: Int, request: URLRequest) {
+    func failWith(errorCode: Int, errorDescription: String?, request: URLRequest) {
```
ianpartridge: Would it be better if this function took a `URLErrorInfo` instead?

Foundation/URLSession/http/EasyHandle.swift
johnno1962: I’m just using that function for some other initialisation as a convenience rather than create a new one. I think the name if probably good as it is as it relates to it’s primary purpose.

Foundation/URLSession/http/HTTPURLProtocol.swift
johnno1962: This doesn’t realy pan out as you would be having to create a URLErrrorInfo at the last minute in two out of three of the places where this function is called. Better to leave it as an extra argument. URLErrorInfo is mostly to covery the into through the various callbacks.

Foundation/URLSession/http/HTTPURLProtocol.swift
johnno1962: Fair enough ;) Just being lazy.

Foundation/URLSession/http/EasyHandle.swift
johnno1962: I could move this extra code into another of the setup functions.

ianpartridge: What do you think @pushkarnk ?

johnno1962: I see there is a function set(errorBuffer:) which is called from _HTTPURLProtocolconfigureEasyHandle(forRequest:) I could set the buffer there but it wouldn’t guarantee the buffer still exists when you call curl_easy_strerror() on the handle.

johnno1962: I’ve pushed a new revision introducing a default error buffer so hte messages are always as descriptive as possible. See what you think.

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -31,6 +31,10 @@ static CFURLSessionMultiCode MakeMultiCode(CURLMcode value) {
     return (CFURLSessionMultiCode) { value };
 }
 
+const char *CFURLSessionErrorDescription(int value) {
+    return curl_easy_strerror(value);
+}
```
parkera: What's the memory management here? Does the result need to be freed?

Foundation/URLSession/http/MultiHandle.swift
```diff
@@ -214,12 +214,14 @@ fileprivate extension URLSession._MultiHandle {
...
+            URLErrorInfo(code: $0, description: String(cString: CFURLSessionErrorDescription(easyCode.value)))
+        }
+        completedTransfer(forEasyHandle: easyHandle, errorInfo: errorInfo)
     }
     /// Transfer completed.
-    func completedTransfer(forEasyHandle handle: _EasyHandle, errorCode: Int?) {
-        handle.completedTransfer(withErrorCode: errorCode)
+    func completedTransfer(forEasyHandle handle: _EasyHandle, errorInfo: URLErrorInfo?) {
+        handle.completedTransfer(withErrorInfo: errorInfo)
     }
```
parkera: This should all use Foundation's existing abstraction for errors: `NSError`, instead of a custom struct.

Foundation/URLSession/http/HTTPURLProtocol.swift
```diff
@@ -557,7 +558,7 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
             completeTask()
         case .failWithError(let errorCode):
             internalState = .transferFailed
-            failWith(errorCode: errorCode, request: request)
+            failWith(errorCode: errorCode, errorDescription: "Completion failure", request: request)
```
parkera: We're in a somewhat unfortunate place here, because localization doesn't quite work yet, but we still should find a way to make this use the usual localized string lookup API (on bundle) instead of hard coding English strings.

CoreFoundation/URL.subproj/CFURLSessionInterface.h
```diff
@@ -48,6 +48,10 @@ typedef struct CFURLSessionEasyCode {
     int value;
 } CFURLSessionEasyCode;
 
+CF_EXPORT const char * _Nonnull CFURLSessionErrorDescription(int value);
```
parkera: CF API really should use a `CFStringRef` as a result instead of C strings. That resolves the memory management question of the result for callers.

CoreFoundation/URL.subproj/CFURLSessionInterface.c
johnno1962: buffer is either static string or buffer set/owned by EasyHandle.swift. It doesn’t need to be freed.

Foundation/URLSession/http/MultiHandle.swift
johnno1962: We can’t construct an NSError before we pass it around as the code doen’t initially know the URL that is later wrapped in the userInfo.

Foundation/URLSession/http/HTTPURLProtocol.swift
johnno1962: I’ve created a placeholder for a future NSLocalizedString(_, _) implementation and used it.

CoreFoundation/URL.subproj/CFURLSessionInterface.h
johnno1962: Changed to CFString. Unsure of the allocator I should be using if you want to check it.

parkera: CF (almost) always uses `kCFAllocatorSystemDefault` internally.

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -19,6 +19,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "CFURLSessionInterface.h"
+#include "CFString.h"
```
parkera: `#include <CoreFoundation/CFString.h>`

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -31,6 +32,10 @@ static CFURLSessionMultiCode MakeMultiCode(CURLMcode value) {
     return (CFURLSessionMultiCode) { value };
 }
 
+CFStringRef CFURLSessionErrorDescription(int value) {
```
parkera: `CFURLSessionCreateErrorDescription`

CoreFoundation/URL.subproj/CFURLSessionInterface.h
```diff
@@ -48,6 +48,10 @@ typedef struct CFURLSessionEasyCode {
     int value;
 } CFURLSessionEasyCode;
 
+CF_EXPORT CFStringRef _Nonnull CFURLSessionErrorDescription(int value);
+
+CF_EXPORT int const CFURLSessionEasyErrorSize;
```
parkera: This feels like an abstraction leak, but I'm not sure what to do about it exactly.

Foundation/URLSession/http/MultiHandle.swift
parkera: That can be set later by creating another `NSError`, or using the concept of underlying error.

It's important that we don't create several different kinds of abstractions for the same concept in one source base.

CoreFoundation/URL.subproj/CFURLSessionInterface.h
johnno1962: yeah, the buffer thing is a bit untidy.

CoreFoundation/URL.subproj/CFURLSessionInterface.c
johnno1962: New version with this change and your other suggestions.

johnno1962:
I’ve switched to using a static error buffer as it is unclear what happens when multiple threads are using different easy handles https://curl.haxx.se/libcurl/c/curl_easy_strerror.html.

johnno1962:
I was using the errorBuffer incoreectly. In the event of an error it can be written to with a more specific error and you have to check for or fall back to curl_easy_strerror() as shown in the example further down https://curl.haxx.se/libcurl/c/CURLOPT_ERRORBUFFER.html. I’m now seeing complete error messages like: Optional(Foundation.URLError(_nsError: Couldn't resolve host 'wsssww.bbc.co.uk'))

johnno1962:
This has stabilized and I’ve squashed it unless you have any further comments.

Foundation/URLSession/http/HTTPURLProtocol.swift
```diff
@@ -330,17 +332,19 @@ internal extension _HTTPURLProtocol {
...
+                NSUnderlyingErrorKey: error,
                 NSURLErrorFailingURLErrorKey: $0,
                 NSURLErrorFailingURLStringErrorKey: $0.absoluteString,
+                NSLocalizedDescriptionKey: NSLocalizedString(error.localizedDescription, nil)
                 ]
         }
-        let error = URLError(_nsError: NSError(domain: NSURLErrorDomain, code: errorCode, userInfo: userInfo))
-        completeTask(withError: error)
-        self.client?.urlProtocol(self, didFailWithError: error)
+        let urlErrror = URLError(_nsError: NSError(domain: NSURLErrorDomain, code: error.code, userInfo: userInfo))
```
xwu: Tyyypo ;)

Foundation/NSString.swift
```diff
@@ -20,6 +20,11 @@ extension unichar : ExpressibleByUnicodeScalarLiteral {
     }
 }
 
+// Placeholder for a future implementation
+public func NSLocalizedString(_ key: String, _ comment: String?) -> String {
```
xwu: This is, I believe, not the correct signature for NSLocalizedString in Swift. For one, “comment” should be an argument label (i.e., `comment:` and not `_ comment:`), but there are also other parameters with default values that are not reflected here. I understand it is a placeholder, but the signature should match Darwin Foundation.

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -135,6 +141,7 @@ CFURLSessionEasyCode CFURLSessionInit(void) {
     return MakeEasyCode(curl_global_init(CURL_GLOBAL_SSL));
 }
 
+int const CFURLSessionEasyErrorSize = { CURL_ERROR_SIZE+1 };
```
xwu: Nit: spaces around `+`

Foundation/URLSession/http/HTTPURLProtocol.swift
```diff
@@ -557,7 +561,9 @@ extension _HTTPURLProtocol: _EasyHandleDelegate {
             completeTask()
         case .failWithError(let errorCode):
             internalState = .transferFailed
-            failWith(errorCode: errorCode, request: request)
+            let error = NSError(domain: NSURLErrorDomain, code: errorCode,
+                                userInfo: [NSLocalizedDescriptionKey: "Completion failure"])
```
parkera: We should be calling that new `NSLocalizedString` function here (once its signature is fixed).

Foundation/NSString.swift
parkera: Also, if we're going to add it we may as well put in the placeholder as it's done in Darwin:

```
/// Returns a localized string, using the main bundle if one is not specified.
public func NSLocalizedString(_ key: String, tableName: String? = default, bundle: Bundle = default, value: String = default, comment: String) -> String
```

Foundation/URLSession/http/HTTPURLProtocol.swift
johnno1962: Signature fixed. NSLocalizedString() is called indirectly in the creation of the final URLError struct in failWith(error:….

Foundation/NSString.swift
johnno1962: Done. Seems to work

pushkarnk:
fyi @saiHemak

johnno1962:
Hi @parkera, anything else you need from me to merge this? I could then rebase and we could focus on landing https://github.com/apple/swift-corelibs-foundation/pull/1113 and I’ll be out of your hair for another year! The Android toolchain is essentially complete.

alblue:
@swift-ci please test

pushkarnk:
@johnno1962 Sorry for the delay in reviewing. I went through the review comments and the changes, and I think we should go ahead and merge this. 

Does this look good to you @parkera ?

pushkarnk:
@swift-ci please test

pushkarnk:
@johnno1962 That the Android toolchain is complete is fantastic news indeed! Thank you :)

alblue:
@parkera has the above met your requirements now? If so, can you complete the review?

Foundation/NSString.swift
```diff
@@ -20,6 +20,16 @@ extension unichar : ExpressibleByUnicodeScalarLiteral {
...
 }
 
+// Placeholder for a future implementation
+public
+func NSLocalizedString(_ key: String,
+                       tableName: String? = nil,
+                       bundle: Bundle = Bundle.main,
+                       value: String = "",
+                       comment: String) -> String {
+    return bundle.localizedString(forKey: key, value:value, table:tableName)
```
ianpartridge: Nit: spaces after colons.

johnno1962: Thanks Ian. Amended.

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

alblue:
@parkera can you revisit this please; in particular the signature changes in d809b59 which may or may not be desirable?

johnno1962:
Signature changes were to bring it into line with the documented Darwin signature. It was me that made it optional.

alblue:
Can we split this into two PRs?

The first can add the `NSLocalizedDescription` with the right syntax. I think that's the main one that @parkera wanted to check. Once we've go that agreed and merged, we can rebase the remainder of this change onto that one. Sound good?

johnno1962:
No.

johnno1962:
OK, I’ve created https://github.com/apple/swift-corelibs-foundation/pull/1263 but main main concern as it has always been if is to have this PR merged before https://github.com/apple/swift-corelibs-foundation/pull/1216 as it will break in way that I don’t have the time to fix. Hopefully we can agree the signature of  NSLocalizedDescription quickly!

johnno1962:
Right, rebased and merged. You know I’ll never never get that hour of my life back 😀!

alblue:
OK, I've gone through all of @parkera @xwu and @ianpartridge comments and I'm happy that they've been addressed. I'm going to kick off a test run, and provided that passes and there's no objections from @parkera et al will do a test'n'merge later on tonight.

@swift-ci please test

alblue:
@swift-ci please test

alblue:

johnno1962:
fireworks

PR #1199, opened 2017-09-04
Mark NSPort and NSPortMessage as deprecated

alblue:
Since NSPort and NSPortMessage are Mach-specific concepts, which
are not relevant on non-Darwin platforms, mark these as deprecated
and change the unavailability macro so that they are distinguishable
from implementations that are not yet available.

### DIFF
--- a/Foundation/Port.swift
+++ b/Foundation/Port.swift
@@ -7,13 +7,15 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-
+@available(*,deprecated,message:"Not available on non-Darwin platforms")
 public typealias SocketNativeHandle = Int32
 
+@available(*,deprecated,message:"Not available on non-Darwin platforms")
 extension Port {
     public static let didBecomeInvalidNotification  = NSNotification.Name(rawValue:  "NSPortDidBecomeInvalidNotification")
 }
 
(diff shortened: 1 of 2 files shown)


alblue:
@swift-ci please test

ianpartridge:
Looks good to me. @parkera ?

parkera:
These actually are not Mach-specific - `NSMessagePort` is explicitly designed to be cross-platform.

parkera:
That is why `NSMachPort` is not present but `NSPort` and subclass `NSMessagePort` were included in swift-corelibs-foundation in the first place. `NSSocketPort` is also portable.

ianpartridge:
But https://developer.apple.com/documentation/foundation/nsmessageport says "Avoid NSMessagePort. There's little reason to use NSMessagePort rather than NSMachPort or NSSocketPort. There's no particular performance or functionality advantage. It is recommended avoiding its use."

Do we want API here which is actively discouraged on Darwin?

parkera:
Interesting. We didn't actually mark it as deprecated despite the documentation.

Message port is supposed to be a platform-specific implementation of a platform-neutral API. For example, on Windows we could implement it using named pipes.

If it looks like we won't ever get a great implementation of this then maybe we could just remove it (like we discussed a while ago about getting rid of things that won't ever lose their NSUnimplemented). However, it does seem plausible to use it for IPC along with NSRunLoop, still.

PR #1201, opened 2017-09-05
SR-5803: NSNumber no longer equates 1 and true

spevans:
- Fix isEqual to work with NSNumbers initialised with different types
  (eg, float and UInt).

- Match the behaviour of Darwin where true can only equal an NSNumber
  with a value of exactly 1.

I essentially added tests that worked correctly on Darwin and then fixed the current code until all of the tests passed.

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -226,40 +226,9 @@ open class NSNumber : NSValue {
         case let other as Bool:
             return boolValue == other
         case let other as NSNumber:
-            // CFEqual() requires both values to be the same type.
-            if _CFNumberGetType2(_cfObject) == _CFNumberGetType2(other._cfObject) {
-                return CFEqual(_cfObject, other._cfObject)
-            }
-            // If either NSNumber is a bool, use a specific test.
-            if CFGetTypeID(self) == CFBooleanGetTypeID() {
-                return testBool(self.boolValue, other)
-            } else if CFGetTypeID(other) == CFBooleanGetTypeID() {
-                return testBool(other.boolValue, self)
-            }
             return compare(other) == .orderedSame
         default:
             return false
-        }
-    }
-
-    private func testBool(_ boolean: Bool, _ nonBoolean: NSNumber) -> Bool {
-        let isOne: Bool
-        let isZero: Bool
-
-        // false is only equal to 0,
-        // true is only equal to exactly 1, not any non-zero value.
-        // isOne and isZero can both be false.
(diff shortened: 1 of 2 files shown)


Foundation/NSNumber.swift
```diff
@@ -226,12 +226,43 @@ open class NSNumber : NSValue {
         case let other as Bool:
             return boolValue == other
         case let other as NSNumber:
-            return CFEqual(_cfObject, other._cfObject)
+            // CFEqual() requires both values to be the same type.
```
parkera: I took a look at the Darwin implementation here. It defers to compare for all of this. A few notes for comparison:

For doubles d1/d2:
1. if d1 && d2 are NaN -> ordered same
2. If d2 is negative (including -0), we put nan after it; otherwise before it
3. If d1 is negative (including -0), we put nan after it; otherwise before it

For 'Q' type: call unsignedLongLongValue and compare that

For everything else: call longLongValue and compare that result.

phausler: `@YES.longLongValue == 1`
`@NO.longLongValue == 0`

`NSNumber(value: 1.1) != NSNumber(value: Int64(1))`

parkera: The NSNumber code certainly handles float/double differently - it could be a consequence of bridging which is causing `Float(1.1)` to compare equal to `true`. What's your exact test case so I can check?

spevans: I must have done something wrong in my initial testing as `compare(other) == .orderedSame` seems to work ok now!

spevans:
@parkera Does this look ok now? It mirrors the Darwin results using the same method.

parkera:
Yup

parkera:
@swift-ci test and merge

spevans:
Is there anyway to get a clean build done? I tested on my Linux host (merging into master) and got the same segmentation fault as seen on the CI but when I deleted my `foundation-linux-x86_64` build dir it tested successfully.

parkera:
I think there was something... let me try

parkera:
@swift-ci clean test

parkera:
@swift-ci clean

parkera:
Ok I give up. cc @shahmishal  :)

shahmishal:
@swift-ci clean test

shahmishal:
@swift-ci test

shahmishal:
Clean is not supported with with swift-corelibs-foundation PR testing.

ianpartridge:
@swift-ci please test

PR #1203, opened 2017-09-06
URLSessionTask retain cycle fix

johnno1962:
This PR fixes a retain cycle in URLSessionTask which is causing memory to leak by the amount of data returned by the request. See: https://github.com/apple/swift-corelibs-foundation/pull/1195


pushkarnk:

johnno1962:
Excellent. Thanks @pushkarnk, nice to see this patch in.

PR #1204, opened 2017-09-07
NSNumber: update API for SE-0170

ianpartridge:
`Int(truncating: NSNumber)` is missing in swift-corelibs-foundation. This function was added as part of SE-0170.

This PR implements the `init(truncating:)` and `init?(exactly:)` APIs from SE-0170, and also deprecates `init(_:)` appropriately.

It seems we also lacked implementations of `init(_:)` for many of the primitive number types (e.g. `Int8`) to begin with so this PR also adds those.

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -32,158 +32,523 @@ internal let kCFNumberSInt128Type = CFNumberType(rawValue: 17)!
 internal let kCFNumberSInt128Type: CFNumberType = 17
 #endif
 
+extension Int8 : _ObjectTypeBridgeable {
+    @available(swift, deprecated: 4, renamed: "init(truncating:)")
+    public init(_ number: NSNumber) {
+        self = number.int8Value
+    }
+
+    public init(truncating number: NSNumber) {
+        self = number.int8Value
+    }
+
+    public init?(exactly number: NSNumber) {
+        let value = number.int8Value
+        guard NSNumber(value: value) == number else { return nil }
+        self = value
+    }
+
+    public func _bridgeToObjectiveC() -> NSNumber {
+        return NSNumber(value: self)
+    }
+
+    public static func _forceBridgeFromObjectiveC(_ x: NSNumber, result: inout Int8?) {
+        result = _unconditionallyBridgeFromObjectiveC(x)
+    }
+
+    public static func _conditionallyBridgeFromObjectiveC(_ x: NSNumber, result: inout Int8?) -> Bool {
+        guard let value = Int8(exactly: x) else { return false }
+        result = value
+        return true
+    }
+
+    public static func _unconditionallyBridgeFromObjectiveC(_ source: NSNumber?) -> Int8 {
+        var result: Int8?
+        guard let src = source else { return Int8(0) }
+        guard _conditionallyBridgeFromObjectiveC(src, result: &result) else { return Int8(0) }
+        return result!
+    }
+}
+
+extension UInt8 : _ObjectTypeBridgeable {
+    @available(swift, deprecated: 4, renamed: "init(truncating:)")
+    public init(_ number: NSNumber) {
+        self = number.uint8Value
+    }
+
+    public init(truncating number: NSNumber) {
+        self = number.uint8Value
+    }
+
+    public init?(exactly number: NSNumber) {
+        let value = number.uint8Value
+        guard NSNumber(value: value) == number else { return nil }
+        self = value
+    }
+
+    public func _bridgeToObjectiveC() -> NSNumber {
+        return NSNumber(value: self)
+    }
+
+    public static func _forceBridgeFromObjectiveC(_ x: NSNumber, result: inout UInt8?) {
+        result = _unconditionallyBridgeFromObjectiveC(x)
+    }
+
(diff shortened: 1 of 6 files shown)


ianpartridge:
@swift-ci please test

Foundation/NSNumber.swift
```diff
@@ -32,158 +32,523 @@ internal let kCFNumberSInt128Type = CFNumberType(rawValue: 17)!
...
+        self = number.int8Value
+    }
+
+    public init?(exactly number: NSNumber) {
+        let value = number.int8Value
+        guard NSNumber(value: value) == number else { return nil }
+        self = value
+    }
+
+    public func _bridgeToObjectiveC() -> NSNumber {
```
ianpartridge: Question: should this (and other `_bridgeToObjectiveC()` functions) be marked `@_semantics("convertToObjectiveC")`?

Foundation/NSNumber.swift
```diff
@@ -32,158 +32,523 @@ internal let kCFNumberSInt128Type = CFNumberType(rawValue: 17)!
...
-    static public func _forceBridgeFromObjectiveC(_ source: _ObjectType, result: inout Int?) {
-        result = _unconditionallyBridgeFromObjectiveC(source)
+    public static func _forceBridgeFromObjectiveC(_ x: NSNumber, result: inout Int?) {
+        result = _unconditionallyBridgeFromObjectiveC(x)
     }
     
-    @discardableResult
-    static public func _conditionallyBridgeFromObjectiveC(_ source: _ObjectType, result: inout Int?) -> Bool {
-        result = source.intValue
+    public static func _conditionallyBridgeFromObjectiveC(_ x: NSNumber, result: inout Int?) -> Bool {
```
ianpartridge: I removed the `@discardableResult` annotation as this is not present in the overlay. Is that ok?

Foundation/NSNumber.swift
```diff
@@ -32,158 +32,523 @@ internal let kCFNumberSInt128Type = CFNumberType(rawValue: 17)!
...
     }
-    
-    public typealias _ObjectType = NSNumber
-    public func _bridgeToObjectiveC() -> _ObjectType {
+
+    public init(truncating number: NSNumber) {
+        self = number.boolValue
+    }
+
+    public init?(exactly number: NSNumber) {
```
ianpartridge: I'd welcome a close review of this function.

Foundation/Stream.swift
```diff
@@ -119,7 +119,7 @@ open class InputStream: Stream {
 
     // reads up to length bytes into the supplied buffer, which must be at least of size len. Returns the actual number of bytes read.
     open func read(_ buffer: UnsafeMutablePointer<UInt8>, maxLength len: Int) -> Int {
-        return CFReadStreamRead(_stream, buffer, CFIndex(len._bridgeToObjectiveC()))
+        return CFReadStreamRead(_stream, buffer, CFIndex(truncating: len._bridgeToObjectiveC()))
```
ianpartridge: This is produced by the Xcode fixit. Could it just be `return CFReadStreamRead(_stream, buffer, len)` though?

spevans: If `len` is an `Int` and `CFIndex` is defined as `public typealias CFIndex = Int` why is there any bridging at all?

ianpartridge: That's what I wondered!

phausler: Yea let's drop the extra round trip bridge here

Foundation/NSNumber.swift
phausler: Looks good to me

Foundation/NSNumber.swift
phausler: I am not sure what that does for compiler methods

Foundation/NSNumber.swift
phausler: For swift corelibs it is fine to omit, iirc that is just a hint not a strict enforcement

Foundation/NSNumber.swift
```diff
@@ -32,158 +32,523 @@ internal let kCFNumberSInt128Type = CFNumberType(rawValue: 17)!
 internal let kCFNumberSInt128Type: CFNumberType = 17
 #endif
 
+extension Int8 : _ObjectTypeBridgeable {
```
phausler: This looks like a faithful copy of the overlay, how much effort do you think will be required to make this shared tween the overlay and SCLF?

Foundation/NSNumber.swift
ianpartridge: I guess it generates a warning, I couldn't see why ignoring that would ever be a good idea though.

Foundation/NSNumber.swift
ianpartridge: The protocol is called `_ObjectiveCBridgeable` in the overlay - giving it the same name here would be a start. The files will never be identical because SCLF has the `class NSNumber` definition which isn't in the overlay. But I think we can get them closer.

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
This should be merged after #1213.

ianpartridge:
@swift-ci please test

spevans:

PR #1206, opened 2017-09-07
Implicit import of Dispatch when importing Foundation?

johnno1962:
Hi Apple,

With apologies for the rash of controversial PRs this week, I’m getting to the end the preparation of a release candidate for the Android/Swift toolchain that I’ve been working on and testing with a wide range of Open Source packages (Alamofire, Moya, RxSwift etc.) for portability. One assumption that commonly comes up in these projects is that importing Foundation is enough to have Dispatch classes available to the Swift source i.e it is imported implicitly when you import Foundation.

You must have considered this, and for compatibility with Darwin I would have thought this was the way to go so I’ve added an @_exported import of Dispatch to NSObject.swift in the toolchain for want of a better place to put it. I had also experimented with exporting CoreFoundation but functions like bind() and accept() seem to leak from CoreFoundation.h and are defined in Glibc causing ambiguity errors.

What are your thoughts?

### DIFF
--- a/Foundation/NSObject.swift
+++ b/Foundation/NSObject.swift
@@ -8,6 +8,7 @@
 //
 
 import CoreFoundation
+@_exported import Dispatch
 
 /// The `NSObjectProtocol` groups methods that are fundamental to all Foundation objects.
 ///
--- a/Foundation/NSObject.swift
+++ b/Foundation/NSObject.swift
@@ -7,8 +7,12 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-import CoreFoundation
+// @_exported import of Dispatch here makes it available to all
+// classes in Foundation and all sources that import Foundation.
+// This brings it into line with Darwin usage for compatbility.
 @_exported import Dispatch
+
+import CoreFoundation
 
 /// The `NSObjectProtocol` groups methods that are fundamental to all Foundation objects.
 ///


ianpartridge:
I support the general idea: that as `import Foundation` on Darwin implies `import Dispatch` we should behave the same on Linux.

I'll defer to @parkera etc. on whether this is the best way to achieve that.

parkera:
This gets us the nice Swift interface of Dispatch, right? Not an exported C API?

johnno1962:
I’m pretty sure that’s the case. The C dispatch API would be an import of “CDispatch".

alblue:
@swift-ci please test

johnno1962:
Interesting failure. Implicit import of Dispatch means XCTest must now know how to link against it when building. You can fix it with this small tweak: https://github.com/SwiftJava/swift-corelibs-xctest/commit/1b73c3e4d5b8807c541e680d8b006ab9c02f9aae

johnno1962:
I’ve created a PR on XCTest to reoslve this: https://github.com/apple/swift-corelibs-xctest/pull/201

parkera:
Can you create a simple app that is outside the Swift build system and imports Foundation with this change? Does this change what linker flags they need to use as well?

johnno1962:
Sure, I’ll try to create a test. It shouldn’t be a problem as something using a completed toolchain will find libdispatch.so and libFoundation.so in the same directory. It’s only during the build system build they are in different directories: `Ninja-ReleaseAssert/foundation-linux-x86_64/Foundation` & `Ninja-ReleaseAssert/libdispatch-linux-x86_64/src/.libs`

johnno1962:
You should be able to try the test again now the tweak to swift-corelibs-xctest was merged.

briancroom:
@swift-ci please test

johnno1962:
Thanks again @briancroom, @parkera re: creating a test app I’ve been using this to patch run apps for weeks as well as running the foundation tests so I think this won't have an impact. Foundation itself is dependent on Dispatch anyway.

parkera:
Ok

xwu:
IMO, this should really be its own file, which IIRC in the same way that `swift-package-manager` organizes its code that provides for libc shims, etc. There is no reason why `NSObject` is the correct place to put this important declaration.

ianpartridge:
A whole new file just for one line seems a bit unnecessary - I'm fine with using `NSObject.swift`. @parkera - what do you think? Is this OK for merge?

parkera:
I'm ok with doing it in NSObject - but can we put a comment or something with a bit of explanation of why it's there?

parkera:
@swift-ci test and merge

alblue:

johnno1962:
Thanks @alblue, @parkera just can’t bring himself to merge my PRs 😀, one down three to go...

PR #1207, opened 2017-09-08
Android: Fixed prepare.sh to download the toolchain instead from http…

amraboelela:
Android: Fixed prepare.sh to download the toolchain instead from http://johnholdsworth.com/android_toolchain.tgz as the old link is not working anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ Build
 *.swp
 *.orig
 .arcconfig
+
+*.tgz
--- a/android/prepare.sh
+++ b/android/prepare.sh
@@ -6,12 +6,17 @@
 BUILD_DIR="build/Ninja-ReleaseAssert/swift-linux-x86_64/lib/swift"
 
 cd "$(dirname $0)" &&
-TAR_FILE="$(pwd)/android_dispatch.tgz" &&
+TAR_FILE="$(pwd)/android_toolchain.tgz" &&
 
 if [[ ! -f "${TAR_FILE}" ]]; then
-    echo "Fetching binaries and headers for libxml2, libcurl and libdispatch"
-    curl "https://raw.githubusercontent.com/SwiftJava/SwiftJava/master/android_dispatch.tgz" > "${TAR_FILE}"
+echo "Fetching binaries and headers for libxml2, libcurl and libdispatch"
+curl http://johnholdsworth.com/android_toolchain.tgz > "${TAR_FILE}"
 fi
 
-cd "../.." && mkdir -p "${BUILD_DIR}" &&
-cd "${BUILD_DIR}" && tar xfvz "${TAR_FILE}"
+tar xfvz "${TAR_FILE}"
+
+mkdir -p "../../${BUILD_DIR}/android/armv7" &&
+
+cp -r swift-install/usr/lib/swift/android/{libxml2.so,libcurl.so,libdispatch.so} ../../${BUILD_DIR}/android
+cp -r swift-install/usr/lib/swift/android/armv7/Dispatch.swiftmodule ../../${BUILD_DIR}/android/armv7
+rm -rf swift-install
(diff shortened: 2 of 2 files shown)


amraboelela:
Hi @jckarter please check this one, thanks.

johnno1962:
Hi @amraboelela, all these files can be deleted now thanks. I have an open PR to do just that :)

amraboelela:
@johnno1962 which PR is that?

johnno1962:
https://github.com/apple/swift-corelibs-foundation/pull/1113

amraboelela:
I don't understand, how the changes in #1113 will prevent the need for all the files in android directory?

johnno1962:
They are a long way out of date and intended for people who are porting to android themselves. It’s not really the place for such documentation.

amraboelela:
@johnno1962 how can I build Foundation for Android using your changes in #1113 ? I can test it and let you know.

johnno1962:
Thanks. I had hoped to be able to support people building their own foundation using the toolchain as all the headers and libs are there but it is missing clang. The easiest way to replicate the process is follow the swift-robot-environment scripts https://github.com/SwiftJava/swifty-robot-environment. You shouldn’t need to test it though as all those changes are already in the toolchain of last weekend.

amraboelela:
I prefer to keep the manual script, as the robot script runs too many subscripts like cloning, which I already did. The manual scripts, everyone can customize them later anyway they like.

amraboelela:
Hi @milseman could you find the right people to check this PR?

Thanks :)

milseman:
@johnno1962 is probably the best for this change. CC-ing @parkera as it's in Foundation.

amraboelela:
@milseman is @johnno1962 the one you mentioned to me to contact, when I met you in WWDC17 ?

milseman:
Sorry, I don't remember. @modocache also has some experience in Swift for Android.

johnno1962:
HI @amraboelela, I’ll have a new toolchain for you tomorrow which allows people to rebuild foundation independently on macOS. There is no point in trying to fix these scripts. They are miles out of date.

amraboelela:
@johnno1962 I am trying to use the scripts in your robot repo to fix the existing scripts, so this should be feasible. Also the robot scripts are tied to your own swift fork, and tied to older revisions to the other swift repositories. We need to have scripts inside apple repositories which does the build and be up to date.

johnno1962:
Hi @amraboelela, I’ve uploaded a new version of the toolchain containing the script `"swift-install/rebuild_foundation.sh” which should do what you are trying to do without any external dependencies. It’s very difficult setting up a build environment otherwise. I’d still rather these scripts were deleted as they are of low quality (I should know - I wrote them!) and not relevant to foundation for most people. If you want to fix these scripts rebuild_foundation.sh would be your starting point.

amraboelela:
I don't mind deleting them if we have an alternative script which is accessable from within the apple foundation repository. Thanks.

johnno1962:
I think it's unrealistic to expect "We need to have scripts inside apple repositories which does the build and be up to date.” at this time. The process, though documented at the [swifty-robot-environment](https://github.com/SwiftJava/swifty-robot-environment) repo (for anybody who wants to replicate it) is too informal to ever make it into an Apple repo and Android is not a supported platform anyway. I can’t think of anything useful that could be left in the place of these scripts other than a README.md referring people to this repo and the pre-built toolchain for now.

amraboelela:
Hi @johnno1962 I already created swift/utils/android/build-toolchain script and would like to continue improving it. Everybody welcomed that script which means they have no objection having such scripts inside Apple repo. You did a good job in the robot scripts, so let me work into improving the build-toolchain script that I started.

@milseman do you have input on this?

milseman:
I think @johnno1962's points are compelling, though there will always be some Android build configuration left in the Swift's build system (e.g. CMake support). I have no horses in this race otherwise. @parkera, what do you think?

amraboelela:
Let's close this PR for now.

PR #1208, opened 2017-09-09
NotificationCenter should us NSNotification.Name per docs

glenna:
Per [Apple docs](https://developer.apple.com/documentation/foundation/notificationcenter), `addObserver` and `post` use `NSNotification.Name`

### DIFF
--- a/Foundation/NSNotification.swift
+++ b/Foundation/NSNotification.swift
@@ -178,7 +178,7 @@ open class NotificationCenter: NSObject {
         removeObserver(observer, name: nil, object: nil)
     }
 
-    open func removeObserver(_ observer: Any, name aName: NSNotification.Name?, object: Any?) {
+    open func removeObserver(_ observer: Any, name aName: Notification.Name?, object: Any?) {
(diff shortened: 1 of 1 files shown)


ianpartridge:
According to https://developer.apple.com/documentation/foundation/notificationcenter/1407263-removeobserver it should be `NSNotification.Name`?

glenna:
As does addObserver(...), but in source it’s using `Notification.Name` (not that that makes this PR correct)

ianpartridge:
It would be great to fix up these incompatibilities. Do you want to do that in this PR? Or should we close it?

glenna:
Just to clarify, you mean to change them both back to `NSNotificaiton.Name` to match the docs?

ianpartridge:
Yes. It looks like `post()` is wrong too.

spevans:
@swift-ci please test

glenna:
@ianpartridge got a bit busy, but it's done now. i'll edit the PR title/description so the merge commit will have a relevant message

ianpartridge:
Thanks!

PR #1209, opened 2017-09-10
Correct a typo for thread-specific set [NFC]

xwu:
I was reading through `Thread.swift` and found the name of an internal function to be confusing. After some head scratching, it's pretty clear that `_CThreadSpecificSet` should be `_CFThreadSpecificSet`. This PR updates the name so as to improve clarity.


alblue:
@swift-ci please test

parkera:
Yup, thanks!

PR #1210, opened 2017-09-11
Thread safety for URLSession

johnno1962:
Hi Apple,

I’ve been seeing assertion failures using a URLSession from multiple threads on [this line](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/URLSession/TaskRegistry.swift#L63). This PR makes a small change so allocation of the taskIdentifier in URLSession is thread safe.

Thanks.

### DIFF
--- a/Foundation/URLSession/URLSession.swift
+++ b/Foundation/URLSession/URLSession.swift
@@ -192,7 +192,6 @@ public let NSURLSessionTransferSizeUnknown: Int64 = -1
 open class URLSession : NSObject {
     fileprivate let _configuration: _Configuration
     fileprivate let multiHandle: _MultiHandle
-    fileprivate let taskIdentifierLock = NSLock()
     fileprivate var nextTaskIdentifier = 1
     internal let workQueue: DispatchQueue 
     /// This queue is used to make public attributes on `URLSessionTask` instances thread safe.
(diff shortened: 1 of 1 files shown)


alblue:
@swift-ci please test

Foundation/URLSession/URLSession.swift
```diff
@@ -192,6 +192,7 @@ public let NSURLSessionTransferSizeUnknown: Int64 = -1
 open class URLSession : NSObject {
     fileprivate let _configuration: _Configuration
     fileprivate let multiHandle: _MultiHandle
+    fileprivate let taskIdentifierLock = NSLock()
```
weissi: my only question is what is the preferred way, `NSLock` or a dispatch queue? I think we should be consistent. @phausler / @parkera do we have a preferred option?

weissi: also: the `sessionDescription` and `delegate` properties are also mutable and accessible from the outside. I suppose the should be locked too.

johnno1962: I’ve switched to using the workQueue for consistency

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

pushkarnk:

pushkarnk:
cc @saiHemak

johnno1962:
Thanks @pushkarnk, Are we any closer to merging https://github.com/apple/swift-corelibs-foundation/pull/1198? Is there anything more I can do?

PR #1212, opened 2017-09-12
AffineTransform: implement hashValue and improve tests

ianpartridge:
1. Implement `NSAffineTransform.hashValue` and enable two tests which were previously disabled.

2. Add `test_rotation_compose()` to test composition of rotation. Test taken from [here](https://github.com/apple/swift/blob/d2926cfb05177d4db0ec89e8609cad1aee14fba1/test/stdlib/TestAffineTransform.swift#L376).


ianpartridge:
@swift-ci please test

alblue:
Looks reasonable to me.

PR #1217, opened 2017-09-14
Stub out new properties/enum in URLSessionConfiguration

pushkarnk:
New properties per [the API](https://developer.apple.com/documentation/foundation/urlsessionconfiguration?changes=latest_minor) doc. Though they may be irrelevant in the Linux server context, they may make sense in the Android context.

### DIFF
--- a/Foundation/URLSession/URLSessionConfiguration.swift
+++ b/Foundation/URLSession/URLSessionConfiguration.swift
@@ -209,4 +209,22 @@ open class URLSessionConfiguration : NSObject, NSCopying {
      */
      open var protocolClasses: [AnyClass]?
 
+     /* A Boolean value that indicates whether the session should wait for connectivity to become available, or fail immediately */
+     @available(*, unavailable, message: "Not available on non-Darwin platforms")
+     open var waitsForConnectivity: Bool { NSUnsupported() }
+
+     /* A service type that specifies the Multipath TCP connection policy for transmitting data over Wi-Fi and cellular interfaces*/
+     @available(*, unavailable, message: "Not available on non-Darwin platforms")
+     open var multipathServiceType: URLSessionConfiguration.MultipathServiceType { NSUnsupported() }
+
+}
+
+@available(*, unavailable, message: "Not available on non-Darwin platforms")
+extension URLSessionConfiguration {
+    public enum MultipathServiceType {
+        case none
+        case handover
+        case interactive
+        case aggregate
+    }
 }


pushkarnk:
@johnno1962 The two new properties in URLSessionConfiguration `waitForConnectivity` and `multipathServiceType` are possibly irrelevant to the Linux server context. Do you see them being implemented in the near future for Swift Foundation on Android though?

johnno1962:
Hi, this might be useful in future but for now I can’t think of a way to implement this from the NDK.

pushkarnk:
@parkera Can you please let us know your opinion on this?

parkera:
@salgernon could also chime in here, but -- it seems that if we have no way to reasonably implement this, we may be better off either leaving these out or marking them unavailable to warn developers that there is no implementation.

parkera:
This one falls into the category of "leafy" API that doesn't need to be implemented with `NSUnimplemented`.

alblue:
If we think these aren't going to be implemented any time soon, we could use `NSUnsupported()` and an availability macro to indicate that it's deprecated/unavailable on non-Darwin platforms, as we did in:

https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/Bundle.swift#L80

That way we allow those coding against it to know that these aren't available and will generate a compile error/warning, rather than relying on the behaviour and wondering why it doesn't work.

Of course if we do think this will be implemented in the near future then `NSUnimplemented` may be better.

ianpartridge:
I agree that `NSUnsupported()` seems the best choice here.

Foundation/URLSession/URLSessionConfiguration.swift
```diff
@@ -209,4 +209,19 @@ open class URLSessionConfiguration : NSObject, NSCopying {
      */
      open var protocolClasses: [AnyClass]?
 
+     /* A Boolean value that indicates whether the session should wait for connectivity to become available, or fail immediately */
+     open var waitsForConnectivity: Bool = false
```
alblue: These should have a warning:

`@available(*,deprecated,message:"Not available on non-Darwin platforms")`

Foundation/URLSession/URLSessionConfiguration.swift
```diff
@@ -209,4 +209,19 @@ open class URLSessionConfiguration : NSObject, NSCopying {
...
+
+     /* A service type that specifies the Multipath TCP connection policy for transmitting data over Wi-Fi and cellular interfaces*/
+     open var multipathServiceType: URLSessionConfiguration.MultipathServiceType = .none
+
+}
+
```
alblue: This should have a warning:

`@available(*,deprecated,message:"Not available on non-Darwin platforms")`

parkera: We can also use `unavailable` to make it an error to attempt to use it.

pushkarnk:
Thanks @parkera @alblue and @ianpartridge for reviewing! I agree with all the points made above and have made changes suitably.

ianpartridge:

ianpartridge:

PR #1218, opened 2017-09-14
SR-5871: TestThread.test_threadStart failing sporadically on Ubuntu 14.04

spevans:
- Lock the NSCondition variable before creating the thread,
  otherwise the thread could broadcast on the condition variable
  before the parent has waited on it.

### DIFF
--- a/TestFoundation/TestThread.swift
+++ b/TestFoundation/TestThread.swift
@@ -41,14 +41,15 @@ class TestThread : XCTestCase {
     
     func test_threadStart() {
         let condition = NSCondition()
+        condition.lock()
+
         let thread = Thread() {
             condition.lock()
             condition.broadcast()
             condition.unlock()
         }
         thread.start()
-        
-        condition.lock()
+
         let ok = condition.wait(until: Date(timeIntervalSinceNow: 10))
         condition.unlock()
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

spevans:
@ianpartridge I reread the man page for condition variables so I think these are being used in the correct order now. Worth a merge to see if it fixes the issue?

TestFoundation/TestThread.swift
```diff
@@ -41,14 +41,15 @@ class TestThread : XCTestCase {
     
     func test_threadStart() {
         let condition = NSCondition()
+        condition.lock()
```
alblue: Doesn't this end up locking on the condition twice? Is that supported?

spevans: The first lock happens in the parent thread - it acquires the lock.
The second lock happens in the newly created lock - it blocks waiting to acquire it.
When the parent thread calls `condition.wait()` the lock is unlocked and then the 2nd thread acquires it.

alblue: Ah, yes - the change makes sense in that regard.

ianpartridge:

spevans:

PR #1220, opened 2017-09-17
Fix TestFoundation Imports

spevans:
As part of getting the swift-corelibs-foundation tests running against the system Foundation and XCTest, some issues have shown up with module imports.

- Fix `Glibc` and `Darwin` to only be imported on the appropriate OS.
- Remove unneeded `import CoreFoundation`.


spevans:
@swift-ci please test

ianpartridge:

PR #1221, opened 2017-09-17
NSString: Remove incorrect init?(CString:encoding:) method

spevans:
- This is a misspelt duplicate of init?(cString:encoding:)


spevans:
@swift-ci please test

ianpartridge:
How odd, looks like this has been here from the beginning, but yes it should be `cString`: https://developer.apple.com/documentation/foundation/nsstring/1411950-init

PR #1222, opened 2017-09-17
NSKeyedUnarchiver: Rename init(forReadingWithData:) -> init(forReadingWith:)

spevans:
This makes the init function match up to the Darwin version, found as part of getting the swift-corelibs-foundation tests running against the system Foundation and XCTest.

https://developer.apple.com/documentation/foundation/nskeyedunarchiver/1410862-init


spevans:
@swift-ci please test

ianpartridge:
Verified against https://developer.apple.com/documentation/foundation/nskeyedunarchiver/1410862-init

PR #1224, opened 2017-09-18
Fixed "Expression too complex" build failures.

dplanitzer:
Fixed multiple occurrences of "Expression was too complex to be solved in reasonable time; consider breaking up the expression into distinct sub-expressions" errors which the Xcode 9 (9A235) tool chain generates while trying to compile the unit tests on a 2013 MacBook Pro.

### DIFF
--- a/TestFoundation/TestNSNumber.swift
+++ b/TestFoundation/TestNSNumber.swift
@@ -1091,8 +1091,8 @@ class TestNSNumber : XCTestCase {
 
     func test_Equals() {
         // Booleans: false only equals 0, true only equals 1
-        XCTAssertTrue(NSNumber(value: true) == NSNumber(value: true))
-        XCTAssertTrue(NSNumber(value: true) == NSNumber(value: 1))
+        XCTAssertTrue(NSNumber(value: true) == NSNumber(value: Bool(true)))
+        XCTAssertTrue(NSNumber(value: true) == NSNumber(value: Int(1)))
         XCTAssertTrue(NSNumber(value: true) == NSNumber(value: Float(1)))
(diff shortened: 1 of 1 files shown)


ianpartridge:
Confirmed I see this error when building `master` with `swift-4.0-DEVELOPMENT-SNAPSHOT-2017-09-18-a` and `swift-DEVELOPMENT-SNAPSHOT-2017-09-18-a-osx`.

It looks like a regression in the compiler though - we shouldn't have to work around like this, should we?  Have you raised an issue at `bugs.swift.org`?

spevans:
This does seem wrong to me and I don't see this issues with the nightly master branch. If `1` and `true` is too complex then there must be a regression somewhere.

dplanitzer:
I filed a radar today: 34520799

We see this build failure on the MacBook Pro from 2013. We haven't had time to test on other Mac hardware. We don't see this compile failure when building in a Amazon AWS instance. But that one is much faster than the MacBook.

The failures started happening for us after commit cffdab33bd108f95840ccabcd4e0dfd8d2618022 was applied. Applying this pull request or just commenting those lines out makes the build work again.

spevans:
On a 2014 MacBook Pro I just compiled master with Swift 4.0 Snapshot 2017-09-18 and saw this issue.
With Swift Development snapshot 2017-09-18 I didn't see this issue.

spevans:
Ive tested this with various swift-4.0 tool chains and it still shows up. It is also broken on Xcode 9 Release using the `Xcode9` toolchain

phausler:

phausler:
lets try that again the failure was not from this change

PR #1226, opened 2017-09-19
compatibility with ObjCBool

weissi:
we need to be compatible with https://developer.apple.com/documentation/objectivec/objcbool

### DIFF
--- a/Foundation/NSSwiftRuntime.swift
+++ b/Foundation/NSSwiftRuntime.swift
@@ -26,6 +26,12 @@ import CoreFoundation
 
 public typealias ObjCBool = Bool
 
+public extension ObjCBool {
+    public var boolValue: Bool {
+        return self
+    }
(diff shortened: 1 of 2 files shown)


weissi:
CC @lukasa

ianpartridge:
https://github.com/apple/swift-corelibs-foundation/pull/1223 🙂

spevans:
I think this is actually a better solution than mine in https://github.com/apple/swift-corelibs-foundation/pull/1223, see my comment about using the struct from the stdlib.

ianpartridge:
Hi @weissi - I think the merging of https://github.com/apple/swift-corelibs-foundation/pull/1223 supercedes this. Please reopen this PR if you disagree for any reason!

weissi:
@ianpartridge indeed, thanks 👍

PR #1228, opened 2017-09-19
Align the UserDefaults implementation with Darwin Foundation

dplanitzer:
- the register(defaults:) function now ensures that values are stored in the internal registration dictionary as NSObject objects which allows object(forKey:) and friends to actually find those values
- the bool(forKey:), integer(forKey:), double(forKey:) and float(forKey:) functions now implicitly convert string represetations to the requested type
- the set(_:forKey:) function now ensures that values of type String, Int and Double are first converted to the proper CFType before they are stored in the CFPreferences object
- register(defaults:) and set(_:forKey:) now cause a fatalError() if an attempt is made to pass an unsupported type. This more closely resembles what the ObjC Foundation does and makes bugs in the code obvious
- added unit tests to cover the above problems
- added the CFXMLPreferencesDomain.c file to the macOS project and enabled the UserDefaults unit tests on macOS
- enabled support for CFXMLPreferencesDomain on Linux, added it to the Linux build process and enabled the UserDefaults unit tests on Linux

### DIFF
--- a/CoreFoundation/Preferences.subproj/CFApplicationPreferences.c
+++ b/CoreFoundation/Preferences.subproj/CFApplicationPreferences.c
@@ -17,7 +17,7 @@
 #include <CoreFoundation/CFNumberFormatter.h>
 #include <CoreFoundation/CFDateFormatter.h>
 #include <sys/types.h>
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED
+#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_LINUX
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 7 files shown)


dplanitzer:
There is hack in UserDefaults:47 that I want to specifically call out and that revolves around a failing forced cast on Linux:

`
            return (anObj as! CFString)._nsObject
`

This cast compiles and executes fine on macOS. It compiles fine on Linux but triggers a crash on Linux. It's not clear to me why this is so because "anObj" is definitely a CFString at that point. I confirmed that by printing out the CFTypeID of "anObj" on Linux.

I replaced the CFString casts with NSString casts for now. This code compiles and runs fine on macOS and Linux. However, I'd like to understand why the cast fails on Linux and what the proper solution is. Note that if write this code:

`
if let theObj = anObj as? CFString {
`

and try to compile it on Linux, then the compiler complains that the conditional cast will always succeed at runtime. Well that then begs the question though why the forced cast crashes the app...

@phausler : I guess that you might know why the forced CFString cast doesn't work on Linux while it does on macOS?

dplanitzer:
@swift-ci please test

phausler:
if the object is an NSString you cannot use `as?` cast to convert to CFString you must use `unsafeBitCast` because on Linux hosts it does not know that the CFTypeRef of CFStringRef is actually a swift object, it just thinks it is an opaque ref counted type. So my guess is that on linux that line of as? CFString always fails.

dplanitzer:
Thanks for the info. I've updated the pull request such that we now do `unsafeBitCast` calls instead of the `as! NSType` force casts.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

ianpartridge:
Hmm, failure in `clang_builtins`. Let's try again and see what happens.

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

PR #1229, opened 2017-09-19
Aligned Operation cancel behavior with Darwin Foundation

dplanitzer:
- calling cancel() or operationQueue.cancelAllOperations() from inside Operation.main() no longer crashes the app when the operation returns from main()
- start() now skips the call to main() if the Operation is already in cancelled state


alblue:

PR #1231, opened 2017-09-20
NSTextCheckingResult: Rename method to match Darwin

spevans:
- resultByAdjustingRangesWithOffset() -> adjustingRanges(offset:)


spevans:
@swift-ci please test

ianpartridge:
Verified against https://developer.apple.com/documentation/foundation/nstextcheckingresult/1417597-adjustingranges

PR #1232, opened 2017-09-20
Adding String->Data fastpath for utf8

alexwishart:
The swift-server [http project](https://github.com/swift-server/http) makes heavy use of the String and Data APIs for converting between representations of the same utf8 encoded data. Profiling it using microbenchmarks showed a lot of time being spent in this area. I have implemented the utf8 fastpath, based on @phausler's work in this [PR](https://github.com/apple/swift/pull/10953). The changes were run against this [microbenchmark](https://github.com/djones6/swift-benchmarks/blob/master/StringToData.swift) and showed noticeable performance improvements on Linux:
Performance without change:

> Concurrency 1: completed 911 loops (4555.00k ops) in 5.00 seconds, 910.19k ops/sec
> Concurrency 2: completed 1167 loops (5835.00k ops) in 5.00 seconds, 1166.57k ops/sec
> Concurrency 3: completed 1181 loops (5905.00k ops) in 5.01 seconds, 1179.74k ops/sec
> Concurrency 4: completed 1470 loops (7350.00k ops) in 5.01 seconds, 1468.04k ops/sec
> Concurrency 5: completed 1432 loops (7160.00k ops) in 5.01 seconds, 1428.46k ops/sec
> Concurrency 6: completed 1588 loops (7940.00k ops) in 5.01 seconds, 1584.07k ops/sec
> Concurrency 7: completed 1588 loops (7940.00k ops) in 5.01 seconds, 1584.37k ops/sec
> Concurrency 8: completed 1571 loops (7855.00k ops) in 5.02 seconds, 1565.02k ops/sec
> Concurrency 9: completed 1605 loops (8025.00k ops) in 5.01 seconds, 1600.37k ops/sec
> Concurrency 10: completed 1602 loops (8010.00k ops) in 5.02 seconds, 1596.28k ops/sec

Performance with change:

> Concurrency 1: completed 5192 loops (25960.00k ops) in 5.00 seconds, 5191.14k ops/sec
> Concurrency 2: completed 9674 loops (48370.00k ops) in 5.00 seconds, 9672.96k ops/sec
> Concurrency 3: completed 13663 loops (68315.00k ops) in 5.00 seconds, 13660.55k ops/sec
> Concurrency 4: completed 17825 loops (89125.00k ops) in 5.00 seconds, 17821.83k ops/sec
> Concurrency 5: completed 21388 loops (106.94m ops) in 5.00 seconds, 21.38m ops/sec
> Concurrency 6: completed 25791 loops (128.96m ops) in 5.00 seconds, 25.79m ops/sec
> Concurrency 7: completed 29370 loops (146.85m ops) in 5.00 seconds, 29.36m ops/sec
> Concurrency 8: completed 33349 loops (166.75m ops) in 5.00 seconds, 33.34m ops/sec
> Concurrency 9: completed 36188 loops (180.94m ops) in 5.00 seconds, 36.18m ops/sec
> Concurrency 10: completed 37536 loops (187.68m ops) in 5.00 seconds, 37.53m ops/sec

### DIFF
--- a/Foundation/NSStringAPI.swift
+++ b/Foundation/NSStringAPI.swift
@@ -759,6 +759,16 @@ extension StringProtocol where Index == String.Index {
     using encoding: String.Encoding,
     allowLossyConversion: Bool = false
   ) -> Data? {
+    let ephemeralString = _ephemeralString
+    if ephemeralString._core.isASCII {
+        switch encoding {
+                case .utf8: return Data(
+                    bytes: UnsafeRawPointer(ephemeralString._core.startASCII), count: ephemeralString._core.count)
+                default:  return _ns.data(
+                    using: encoding.rawValue,
+                    allowLossyConversion: allowLossyConversion)
+        }
+    }
     return _ns.data(
       using: encoding.rawValue,
       allowLossyConversion: allowLossyConversion)
--- a/Foundation/NSStringAPI.swift
+++ b/Foundation/NSStringAPI.swift
@@ -762,6 +762,8 @@ extension StringProtocol where Index == String.Index {
     let ephemeralString = _ephemeralString
     if ephemeralString._core.isASCII {
         switch encoding {
+                case .ascii: fallthrough
+                case .nonLossyASCII: fallthrough
                 case .utf8: return Data(
                     bytes: UnsafeRawPointer(ephemeralString._core.startASCII), count: ephemeralString._core.count)
                 default:  return _ns.data(
--- a/TestFoundation/TestNSData.swift
+++ b/TestFoundation/TestNSData.swift
@@ -34,6 +34,10 @@ class TestNSData: XCTestCase {
     static var allTests: [(String, (TestNSData) -> () throws -> Void)] {
         return [
             ("testBasicConstruction", testBasicConstruction),
+            ("test_asciiString_utf8Encoding", test_asciiString_utf8Encoding),
+            ("test_asciiString_asciiEncoding", test_asciiString_asciiEncoding),
+            ("test_asciiSubstring_utf8Encoding", test_asciiSubstring_utf8Encoding),
+            ("test_asciiSubstring_asciiEncoding", test_asciiSubstring_asciiEncoding),
             ("test_base64Data_medium", test_base64Data_medium),
             ("test_base64Data_small", test_base64Data_small),
             ("test_openingNonExistentFile", test_openingNonExistentFile),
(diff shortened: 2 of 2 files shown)


phausler:
we should consider this for the overlay as well if it is not already there

Foundation/NSStringAPI.swift
```diff
@@ -759,6 +759,16 @@ extension StringProtocol where Index == String.Index {
     using encoding: String.Encoding,
     allowLossyConversion: Bool = false
   ) -> Data? {
+    let ephemeralString = _ephemeralString
+    if ephemeralString._core.isASCII {
+        switch encoding {
+                case .utf8: return Data(
```
phausler: you can also use any ASCII encoding here too in the switch

Foundation/NSStringAPI.swift
```diff
@@ -759,6 +759,16 @@ extension StringProtocol where Index == String.Index {
...
+                    bytes: UnsafeRawPointer(ephemeralString._core.startASCII), count: ephemeralString._core.count)
```
phausler: you should check that the core before doing this; try strings like `"ünîcødé"`

Foundation/NSStringAPI.swift
```diff
@@ -759,6 +759,16 @@ extension StringProtocol where Index == String.Index {
...
+                default:  return _ns.data(
```
phausler: if we are asking for unicode encoding and the backing is unicode we can take another fast path too

Foundation/NSStringAPI.swift
```diff
@@ -759,6 +759,16 @@ extension StringProtocol where Index == String.Index {
...
+                    using: encoding.rawValue,
+                    allowLossyConversion: allowLossyConversion)
+        }
+    }
     return _ns.data(
       using: encoding.rawValue,
       allowLossyConversion: allowLossyConversion)
```
spevans: Since the `default` and  non-ASCII  paths result in the same value, can the if be rewritten as
 ```
if ephemeralString._core.isASCII && encoding == .utf8 {
 return Data(...)
``` ?

**Comment by alexwishart** [comment]
@phausler have added in the extra cases and some tests, could you review again please? 🙂

**Comment by ianpartridge** [comment]
@phausler what do you think about this?

**Comment by alblue** [comment]
@phausler can you take another look at this, please?

**Comment by alblue** [comment]
@phausler please take a look at this review

File: Foundation/NSStringAPI.swift [hunk 5]
```diff
@@ -759,6 +759,18 @@ extension StringProtocol where Index == String.Index {
...
+        switch encoding {
+                case .ascii: fallthrough
+                case .nonLossyASCII: fallthrough
+                case .utf8: return Data(
+                    bytes: UnsafeRawPointer(ephemeralString._core.startASCII), count: ephemeralString._core.count)
+                default:  return _ns.data(
+                    using: encoding.rawValue,
+                    allowLossyConversion: allowLossyConversion)
+        }
+    }
```
phausler: This variant looks good to me; however we could do a touch better an get the unicode variation too.

```swift
public func data(using encoding: String.Encoding, allowLossyConversion: Bool = false) -> Data? {
        let ephemeralString = _ephemeralString
        if ephemeralString._core.isASCII {
            switch encoding {
            case .ascii: fallthrough
            case .nonLossyASCII: fallthrough
            case .utf8: return Data(
                bytes: UnsafeRawPointer(ephemeralString._core.startASCII), count: ephemeralString._core.count)
            default: break
            }
        } else {
            switch encoding {
            case .utf16: fallthrough
            case .unicode:
                return Data(bytes: UnsafeRawPointer(ephemeralString._core.startUTF16), count: ephemeralString._core.count * MemoryLayout<UInt16>.size)
            default: break
            }
        }
        
        return _ns.data(using: encoding.rawValue, allowLossyConversion: allowLossyConversion)
    }
```

TestFoundation/TestNSData.swift
```diff
@@ -897,6 +901,35 @@ extension TestNSData {
...
     }
+    
+    func test_asciiString_utf8Encoding() {
```
phausler: can we test a Unicode codepath? or some other encodings?

djones6:
I agree we should cover the Unicode path.

I think it would be good to get this initial implementation for `utf8` and `ascii` merged as it stands, and @alexwishart to submit another PR to cover additional encodings.

Foundation/NSStringAPI.swift
alexwishart: @phausler have tried implementing the code you suggested, but it doesn't seem to work. It looks like the conversion might be missing the byte order mark based on examining the data produced by the two code paths. Will have a look today into how the existing code does this to try and fix it.

alexwishart: @phausler I've managed to get it to work using the following code, but it's not particularly neat! It basically creates the Byte Order Mark and prepends it to the front of the data.
```swift
public func data(
    using encoding: String.Encoding,
    allowLossyConversion: Bool = false
  ) -> Data? {
    let ephemeralString = _ephemeralString
    if ephemeralString._core.isASCII {
        switch encoding {
                case .ascii: fallthrough
                case .nonLossyASCII: fallthrough
                case .utf8: return Data(
                    bytes: UnsafeRawPointer(ephemeralString._core.startASCII), count: ephemeralString._core.count)
                default: break
        }
    }
    else {
        let str = ""
        var BOMData = str._ns.data(
           using: encoding.rawValue,
            allowLossyConversion: allowLossyConversion)
        
        switch encoding {
            case .utf16: fallthrough
            case .unicode:
                let bytes = UnsafeRawPointer(ephemeralString._core.startUTF16)
                let data = Data(
                    bytes: bytes, count: ephemeralString._core.count * MemoryLayout<UInt16>.size)
                BOMData!.append(data)
                let result = BOMData
                return result
            default: break
        }
    }
    return _ns.data(
      using: encoding.rawValue,
      allowLossyConversion: allowLossyConversion)
  }
```
Have pushed it to a [different branch](https://github.com/alexwishart/swift-corelibs-foundation/tree/unicode_fastpath) with the added tests and would appreciate some advice on how to potentially tidy it up. /cc @djones6 @ianpartridge

parkera:
@milseman @airspeedswift is this use of the internals of `Swift.String` ok with you?

airspeedswift:
@parkera It's emphatically not ok to be accessing the internals of `String` like this. It's a problem that the standard library itself does it too liberally, and we're trying to unwind it – we definitely don't want it happening outside the std lib.

The alternative to this is to use the technique of implementing `_copyContents` to allow for fast copying between unsafe buffer-based containers. Not sure if this particular case requires `Data` to implement it, or `String`.

We'd like `_copyContents` to drop its underscore at some point and become an official thing. If implementing this needs some enhancements to it to make it work, that'd be interesting as progress towards understanding what we'd want to change before we do that.

alexwishart:
@airspeedswift Sounds like a better solution, could you give me a couple of pointers on how to get started?

ianpartridge:
@airspeedswift the method being modified in this PR is [`StringProtocol.data(using:allowLossyConversion:)`](https://developer.apple.com/documentation/swift/stringprotocol/2923420-data) - as such I would have thought it is reasonable for this method to access `_core`?

alblue:
@airspeedswift can you give some feedback as above?

alblue:
@airspeedswift if you have no further objections to this, I'll dismiss your review objections since they've been approved by @phausler

phausler:
nah, @airspeedswift is the one in charge of those private SPIs so that should be the final word; I filed an enhancement request to provide us something faster.

airspeedswift:
Sorry I missed the pings on this earlier.

Essentially, corelibs should be using only public APIs available to any other library. No methods implemented inside corelibs should access underscored APIs as these are not considered public. 

The one option I can see that is public is for `Data` to use `UnsafeMutableBufferPointer.initialize(from:)` to implement an initializer from a sequence. This will then bounce through private methods internal to the standard library that attempt to do a memcpy (via the `_copyContents` method on `Sequence` that `UnsafeMutableBufferPointer.initialize(from:)` calls. `String` doesn't currently implement `_copyContents` though, so that will need to be added (I have a start of a PR to do that, but that needs picking back up, here: https://github.com/apple/swift/pull/8881). That would in turn do a raw copy of the string's buffer if it happens to be ascii already.

If that doesn't quite work, what will be needed is a design to take to evolution for other public APIs. One interesting avenue is looking at how `_copyContents` works and coming up with a proposal for how that could become official API (i.e. drop the underscore) that anyone could use, possibly with some modifications to how it works today..

Adding @milseman who is working on string internals currently, and may be able to offer some suggestions on what we could do.

phausler:
The `initialize(from:)` is perhaps not sufficient due to the conversions. Btw this is the same code as is used in the overlay; so this is kind-of part of the NSString.. just so happens that it is re-implemented in swift-corelibs-foundation. Albeit that I understand your reluctance to allow direct access to internals I think we should have a consistent code-base between the two platforms: I filed https://bugs.swift.org/browse/SR-6475 for an enhancement request outlining a few of the nitty gritties.

milseman:
Nothing will be ready in time for Swift 4.1, but for 5.0 we probably want to be able to expose access to the raw backing storage. Something akin to a `withUnsafeASCIIString`, `withUnsafeUTF16String`, etc., that's essentially pointer+length+flags.

phausler:
That sounds like it would fit the bill, as long as we have a way to look before we leap into the withUnsafe* method or those methods are failable somehow.

alblue:
Thanks @airspeedswift and @milseman for the comments. Should we close this request for now pending further observations for Swift 5?

phausler:
Yea, for now. I think we have a game plan of sorts: unfortunately the perf gain here will have to wait until we have better answers.

airspeedswift:
@phausler you can switch inside your `init` and use `initialize(from:)` on `String.utf8` when appropriate. And in turn `String.utf8._copyContents(from:)` should do a memcpy when the string is holding ASCII.

milseman:
"should" as in it does so today, or "should" as in that's a TODO for the standard library?

airspeedswift:
Sorry, should as in it's a TODO, not ought to work already.

PR #1233, opened 2017-09-20
Import changes from the overlay for Data mutation of slices

phausler:
Mutation of Data slices had issues when it comes to re-indexing to the base slice region. Previously when a slice of Data was mutated the re-calculation of the slice region was incorrectly assigned to a relative region to the previous backing. Now the slice region will copy the specific region out creating a new data upon CoW semantics and offset the index region upon access. Mutating a non CoW slice will retain the previous region backing but adjust the remaining allocation buffer accordingly. To validate this I have added tests for combinations that should approximate all possible combinations of mutations, backing storages and CoW scenarios.

This fixes the following issues:
rdar://problem/34206043
SR-5887
SR-5873
SR-5810

tl;dr – Slicing a CoW should result in fillet mignon not ground chuck.

NOTE: Data has a failure set for custom backing mutation because NSData does not properly delegate to CFData because of missing factory pattern initializers. This prevents testing of a number of mutation cases and should be re-enabled once we have a suitable solution for factory initializers.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -29,6 +29,10 @@
 #include <pthread.h>
 #include <execinfo.h>
 
+#if __has_include(<malloc/malloc.h>)
+#include <malloc/malloc.h>
+#endif
+
 _CF_EXPORT_SCOPE_BEGIN
 
 struct __CFSwiftObject {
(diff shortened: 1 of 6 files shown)


spevans:
@phausler you may want to use ` _CFIsMainThread()` instead of `pthread_main_np()`

phausler:
hmm good point

phausler:
@swift-ci please test

phausler:
@swift-ci please test

PR #1236, opened 2017-09-21
Import changes from the overlay for Data mutation of slices (#1233)

phausler:
* Import changes from the overlay for Data mutation of slices

* Use _CFIsMainThread instead of pthread APIs

* Assert on null pointers for linux targets

### DIFF
--- a/Foundation/JSONEncoder.swift
+++ b/Foundation/JSONEncoder.swift
@@ -64,6 +64,9 @@ open class JSONEncoder {
 
     /// The strategy to use for encoding `Data` values.
     public enum DataEncodingStrategy {
+        /// Defer to `Data` for choosing an encoding.
+        case deferredToData
+        
         /// Encoded the `Data` as a Base64-encoded string. This is the default strategy.
         case base64Encode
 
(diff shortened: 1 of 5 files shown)


phausler:
This is a correction of the branch merge previously made - it is a application of the changes for modification to struct Data that allow for mutations on slices.

Foundation/JSONEncoder.swift
```diff
@@ -64,6 +64,9 @@ open class JSONEncoder {
 
     /// The strategy to use for encoding `Data` values.
     public enum DataEncodingStrategy {
+        /// Defer to `Data` for choosing an encoding.
+        case deferredToData
```
alblue: Isn't adding the deferred case here both a change in ABI and also alphabetically unsorted?

Foundation/JSONEncoder.swift
```diff
@@ -826,6 +834,9 @@ open class JSONDecoder {
 
     /// The strategy to use for decoding `Data` values.
     public enum DataDecodingStrategy {
+        /// Defer to `Data` for decoding.
+        case deferredToData
```
alblue: Same comment a above

ianpartridge:
@phausler what's the status of this? We really want this in the 4.0 branch :)

PR #1239, opened 2017-09-26
Stub out value(forKey key: String)

ddunn2:
I noticed this method doesn't exist on Linux however apparently this was deliberately omitted from sclf as it relies on KVC which is not supported on Linux.

This PR stubs out the method as NSUnsupported().


pushkarnk:

PR #1240, opened 2017-09-27
Fix to NSKeyedArchiver when object not AnyHashable

johnno1962:
Hi Apple,

a minor fix to NSKeyedArchiver for when objects being archived are not “AnyHashable”. Without the fix the following code does not work:
```Swift
        let session = URLSession( configuration: .default )
        let url = URL(string: "https://httpbin.org/get")!
        _ = session.dataTask( with: URLRequest( url: url ) ) {
            (data, response, error) in
            if let data = data {
                do {
                    let json = try JSONSerialization.jsonObject(with: data)
                    let archive = NSKeyedArchiver.archivedData(withRootObject: json)
                    let object = NSKeyedUnarchiver.unarchiveObject(with: archive)
                    let text = try JSONSerialization.data(withJSONObject: object)
                    NSLog( String(describing: String( data: text, encoding: .utf8 )) )
                }
                catch let e {
                    NSLog( "\(e)" )
                }
            }
            else {
                NSLog( "\(String(describing: error))" )
            }
            }.resume()
```
It crashes during the archiving with the error:
```
Could not cast value of type 'Swift.Dictionary<Swift.String, Any>' (0x6000248c0038) to 'Swift.AnyHashable' (0x10262c040).
```

### DIFF
--- a/TestFoundation/TestNSKeyedArchiver.swift
+++ b/TestFoundation/TestNSKeyedArchiver.swift
@@ -106,6 +106,7 @@ class TestNSKeyedArchiver : XCTestCase {
             ("test_archive_user_class", test_archive_user_class),
             ("test_archive_uuid_bvref", test_archive_uuid_byref),
             ("test_archive_uuid_byvalue", test_archive_uuid_byvalue),
+            ("test_archive_unhashable", test_archive_unhashable),
         ]
     }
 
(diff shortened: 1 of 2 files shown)


ianpartridge:
Looks good, can we add a test too?

TestFoundation/TestNSKeyedArchiver.swift
```diff
@@ -106,6 +106,7 @@ class TestNSKeyedArchiver : XCTestCase {
             ("test_archive_user_class", test_archive_user_class),
             ("test_archive_uuid_bvref", test_archive_uuid_byref),
             ("test_archive_uuid_byvalue", test_archive_uuid_byvalue),
+            ("test_archive_unhasable", test_archive_unhashable),
```
ianpartridge: Typo: `test_archive_unhasable`

TestFoundation/TestNSKeyedArchiver.swift
```diff
@@ -312,4 +313,25 @@ class TestNSKeyedArchiver : XCTestCase {
...
+                "Accept-Language": "en",
+                "Connection": "close",
+                "Host": "httpbin.org",
+                "User-Agent": "TestFoundation (unknown version) curl/7.54.0"
+              },
+              "origin": "0.0.0.0",
+              "url": "https://httpbin.org/get"
+            }
+            """.data(using: .utf8)!
+        let json = try! JSONSerialization.jsonObject(with: data)
```
ianpartridge: I prefer to `do { }` this, then `XCTFail()` in the catch. That way a failure doesn't blow up the whole test suite.  What do you think?

spevans: @ianpartridge I agree its definitely better to try and catch the error and fail the test then to prevent all of the other tests from even running.

TestFoundation/TestNSKeyedArchiver.swift
johnno1962: Thanks, fixed.

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

alblue:
Looks good to me - @johnno1962 are you OK with this being merged now?

johnno1962:
Sure. it’s a very minor change.

PR #1241, opened 2017-09-27
Added build-android script

amraboelela:


johnno1962:
Hi @amraboelela, see if you can slave a few tips from https://github.com/apple/swift-corelibs-foundation/pull/1113/files which has some different takes on thread names, NI_MAXHOST etc.

.gitignore
```diff
@@ -20,3 +20,5 @@ Build
 *.swp
 *.orig
 .arcconfig
+
+build.ninjae
```
spevans: Is this meant to be `build.ninja`? If so it is already covered by a wildcard

spevans:
@amraboelela Could you rebase this onto master just to eliminate the merge commits please.

.gitignore
johnno1962: This is probably coming from the "sed -ie” commands in the script where “e” is taken as the name for the temp file.

alblue:
Would also help to squash the commits and improve the commit messages for future readability.

amraboelela:
Created new PR https://github.com/apple/swift-corelibs-foundation/pull/1249 

Closing this one

PR #1244, opened 2017-09-29
Fix Foundation extensions to Substring

moiseev:
When a substring gets bridged to NSString, it loses the initial offset,
therefore APIs that accept or return StringIndex ranges should handle
this case explicitly by adding/subtracting the substring start offset.

Similar to https://github.com/apple/swift/pull/12151


moiseev:
@swift-ci Please test

PR #1245, opened 2017-10-01
Process: Uncomment disabled tests and make compatible with Darwin

spevans:
- Handle differences in output from /bin/cat on Linux/macOS when
  printing an error.

-  test_pipe_stdout_and_stderr_same_pipe(): Dont pass malloc debug
  flags to subprocesses when debugging under Xcode.

- runTask(): return stdout and stderr as separate values so the malloc debug is captured separately.

### DIFF
--- a/TestFoundation/TestProcess.swift
+++ b/TestFoundation/TestProcess.swift
@@ -28,13 +28,11 @@ class TestProcess : XCTestCase {
                    ("test_pipe_stdout", test_pipe_stdout),
                    ("test_pipe_stderr", test_pipe_stderr),
                    ("test_current_working_directory", test_current_working_directory),
-                   // disabled for now
-                   // ("test_pipe_stdout_and_stderr_same_pipe", test_pipe_stdout_and_stderr_same_pipe),
+                   ("test_pipe_stdout_and_stderr_same_pipe", test_pipe_stdout_and_stderr_same_pipe),
                    ("test_file_stdout", test_file_stdout),
-                   // disabled for now
-                   // ("test_passthrough_environment", test_passthrough_environment),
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

TestFoundation/TestProcess.swift
```diff
@@ -206,7 +207,12 @@ class TestProcess : XCTestCase {
             XCTFail("Could not read stdout")
             return
         }
-        XCTAssertEqual(string, "/bin/cat: invalid_file_name: No such file or directory\n")
+
+        // Remove the leading '/bin/' since on macOS '/bin/cat' just outputs 'cat:'
+        let searchStr = "/bin/"
+        let errMsg = string.replacingOccurrences(of: searchStr, with: "", options: [.literal, .anchored],
+                                              range: searchStr.startIndex..<searchStr.endIndex)
+        XCTAssertEqual(errMsg, "cat: invalid_file_name: No such file or directory\n")
```
alblue: This smells a bit fragile to me - we're expecting a specific formatting of the error message and text. Why not just assert that the string contains "No such file or directory" instead?

alblue:
This looks good to me (with only one minor concern regarding exact error messages). @phausler can you approve the review when you get a chance?

TestFoundation/TestProcess.swift
spevans: Reasonable point, however I would probably wait until it breaks on a specific platform before loosening it. Also, that would also match if the executable didn't exist:

```
$ /bin/foo
-bash: /bin/foo: No such file or directory
```

alblue: Fair enough :)

PR #1247, opened 2017-10-02
Fix against se-0184a

kelvin13:
Patch for [SE 0184a](https://github.com/apple/swift/pull/12200)

### DIFF
--- a/Foundation/Host.swift
+++ b/Foundation/Host.swift
@@ -42,8 +42,7 @@ open class Host: NSObject {
     static internal func currentHostName() -> String {
         let hname = UnsafeMutablePointer<Int8>.allocate(capacity: Int(NI_MAXHOST))
         defer {
-            hname.deinitialize()
-            hname.deallocate(capacity: Int(NI_MAXHOST))
+            hname.deallocate()
         }
         let r = gethostname(hname, Int(NI_MAXHOST))
         if r < 0 || hname[0] == 0 {
@@ -80,8 +79,7 @@ open class Host: NSObject {
         let address = UnsafeMutablePointer<Int8>.allocate(capacity: Int(NI_MAXHOST))
         defer {
             freeifaddrs(ifaddr)
-            address.deinitialize()
-            address.deallocate(capacity: Int(NI_MAXHOST))
+            address.deallocate()
         }
         while let ifaValue = ifa?.pointee {
             if let ifa_addr = ifaValue.ifa_addr, ifaValue.ifa_flags & UInt32(IFF_LOOPBACK) == 0 {
@@ -137,8 +135,7 @@ open class Host: NSObject {
             var res: UnsafeMutablePointer<addrinfo>? = res0
             let host = UnsafeMutablePointer<Int8>.allocate(capacity: Int(NI_MAXHOST))
             defer {
-                host.deinitialize()
-                host.deallocate(capacity: Int(NI_MAXHOST))
+                host.deallocate()
             }
             while res != nil {
                 let info = res!.pointee
@@ -184,4 +181,3 @@ open class Host: NSObject {
         return nil
     }
 }
-
--- a/Foundation/NSConcreteValue.swift
+++ b/Foundation/NSConcreteValue.swift
@@ -91,17 +91,17 @@ internal class NSConcreteValue : NSValue {
 
         self._typeInfo = typeInfo!
 
-        self._storage = UnsafeMutableRawPointer.allocate(bytes: self._typeInfo.size, alignedTo: 1)
-        self._storage.copyBytes(from: value, count: self._typeInfo.size)
+        self._storage = UnsafeMutableRawPointer.allocate(byteCount: self._typeInfo.size, alignment: 1)
+        self._storage.copyMemory(from: value, byteCount: self._typeInfo.size)
     }
 
     deinit {
         // Cannot deinitialize raw memory.
-        self._storage.deallocate(bytes: self._size, alignedTo: 1)
+        self._storage.deallocate()
     }
     
     override func getValue(_ value: UnsafeMutableRawPointer) {
-        value.copyBytes(from: self._storage, count: self._size)
+        value.copyMemory(from: self._storage, byteCount: self._size)
     }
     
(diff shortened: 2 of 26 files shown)


Foundation/Host.swift
```diff
@@ -38,8 +38,8 @@ open class Host: NSObject {
     static internal func currentHostName() -> String {
         let hname = UnsafeMutablePointer<Int8>.allocate(capacity: Int(NI_MAXHOST))
         defer {
-            hname.deinitialize()
-            hname.deallocate(capacity: Int(NI_MAXHOST))
+            hname.deinitialize(count: 1)
```
ianpartridge: I've always found this method confusing.  What does it actually _do_?

https://developer.apple.com/documentation/swift/unsafemutablepointer/1641720-deinitialize is no help, it just says that `deinitialize()` deinitializes!  Does it zero the memory, for example?  I assume not otherwise the documentation would say so.

If it is just changing the "state" of the memory then I question the meaning/utility of the "states".

Foundation/NSConcreteValue.swift
```diff
@@ -91,17 +91,17 @@ internal class NSConcreteValue : NSValue {
 
         self._typeInfo = typeInfo!
 
-        self._storage = UnsafeMutableRawPointer.allocate(bytes: self._typeInfo.size, alignedTo: 1)
-        self._storage.copyBytes(from: value, count: self._typeInfo.size)
+        self._storage = UnsafeMutableRawPointer.allocate(byteCount: self._typeInfo.size, alignment: 1)
+        self._storage.copyMemory(from: value, byteCount: self._typeInfo.size)
```
ianpartridge: This one doesn't read as an improvement to me - the previous API seemed quite clear as to what it would do. Maybe you are fixing some overloading of the meaning of `count` though? I'm not sure.

Foundation/Host.swift
spevans: @ianpartridge Isn't `deinitialize` for objects where the `deinit` method is called on them? If Im correct its not needed for arrays of `Int8`

ianpartridge: No I think this is unrelated to `deinit` which only applies to classes. `UnsafeMutablePointer` et al are structs.

spevans: I was referring to the type stored in the allocated memory not the `UnsafeMutablePointer` itself.

ianpartridge: Oh I see what you mean, sorry :) If that's correct then the doc doesn't mention it!

kelvin13: deinitialize does nothing for trivial types, it doesn’t actually need to be there

Foundation/NSConcreteValue.swift
kelvin13: this was discussed heavily during the review of SE-0184. The core team felt strongly that it should be `copyMemory` and `byteCount`.

kelvin13:
[SE 184](https://github.com/kelvin13/swift-evolution/blob/improved-pointers/proposals/0184-unsafe-pointers-add-missing.md)
[SE 184a](https://github.com/kelvin13/swift-evolution/blob/improved-pointers/proposals/0184a-unsafe-pointers-part-1.md)

Foundation/Host.swift
ianpartridge: Oh OK! Could you update this PR to remove it then?

kelvin13: I’ve removed those calls in `Host.swift` though I would hold off removing them elsewhere because it could inadvertently introduce memory bugs if the type isn’t absolutely a trivial type and will never be anything but a trivial type in the future

kelvin13: I’ve removed those calls in `Host.swift` though I would hold off removing them elsewhere because it could inadvertently introduce memory bugs if the type isn’t absolutely a trivial type and will never be anything but a trivial type in the future

alblue:
@swift-ci please test

kelvin13:
you have to test against the `swift` PR

alblue:
@swift-ci please test

alblue:
@swift-ci please test

alblue:
The errors in the build are:

```
Foundation/AffineTransform.swift:317:55: error: incorrect argument labels in call (have 'byteCount:alignment:', expected 'bytes:alignedTo:')
        let pointer = UnsafeMutableRawPointer.allocate(byteCount: MemoryLayout<Float>.stride * 6, alignment: 1)
                                                      ^~~~~~~~~~                                  ~~~~~~~~~
                                                       bytes                                      alignedTo
```

kelvin13:
@alblue This PR will only build if you build it with the actual compiler comifications in @12200

alblue:
OK, so when apple/swift#12200 is merged then we can revisit this request

ianpartridge:
Although [SE-184](https://github.com/apple/swift-evolution/blob/master/proposals/0184-unsafe-pointers-add-missing.md) has been accepted, this PR seems to be for SE-184a - I'm not sure how that relates?

@kelvin13 will SE-184a require a separate Swift Evolution review cycle? Is this PR dependent on another Swift Evolution proposal beyond SE-184.

If so, I don't think we should keep this PR open for potentially months while that process proceeds. We should close it and reopen it when it's ready. The evolution proposal can link to the closed PR in the meantime.

Thoughts?

kelvin13:
@ianpartridge SE-184a *is* SE-184, SE-184b will likely get a completely different Swift evolution number. It’s confusing I know. Originally it was all one proposal, it got split up in review.

ianpartridge:
Oh great, thanks for clarifying! In that case, as SE-184 is Accepted we should keep this open.

Do you have an estimate of when you hope to land the implementation of SE-184 into `master`?

ianpartridge:
Great. By the way, this needs rebasing as it has a conflict now.

kelvin13:
done

alblue:
@kelvin13 has SE-184 landed in Swift yet, and if so, can this be merged? If not, do you have a link for the PR in swift that is blocking this PR?

atrick:
This is being blocked by https://github.com/apple/swift/pull/12200

PR #1248, opened 2017-10-04
NSMutableArray: Rename methods to match Darwin

spevans:
- addObjectsFromArray() -> addObjects(from:)

- replaceObjects(in:withObjectsFromArray) ->
  replaceObjects(in:withObjectsFrom:)

- sort() -> sort(comparator:)

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -27,18 +27,33 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
         return _SwiftValue.fetch(nonOptional: _storage[index])
     }
     
-    public convenience override init() {
-        self.init(objects: [], count:0)
+    public override init() {
+        _storage.reserveCapacity(0)
     }
-    
-    public required init(objects: UnsafePointer<AnyObject>!, count cnt: Int) {
+
+    public required init(objects: UnsafePointer<AnyObject>?, count cnt: Int) {
+        precondition(cnt >= 0)
+        precondition(cnt == 0 || objects != nil)
+
+        _storage.reserveCapacity(cnt)
+        for idx in 0..<cnt {
+            _storage.append(objects![idx])
+        }
+    }
+
+    public convenience init(objects: UnsafePointer<AnyObject>, count cnt: Int) {
+        self.init()
         _storage.reserveCapacity(cnt)
         for idx in 0..<cnt {
             _storage.append(objects[idx])
         }
     }
-    
-    required public convenience init(arrayLiteral elements: Any...) {
+
+    convenience init(objects elements: AnyObject...) {
+        self.init(objects: elements, count: elements.count)
+    }
+
+    public required convenience init(arrayLiteral elements: Any...) {
         self.init(array: elements)
     }
     
@@ -168,7 +183,7 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
     open func addingObjects(from otherArray: [Any]) -> [Any] {
         return allObjects + otherArray
     }
-    
+
     open func componentsJoined(by separator: String) -> String {
         // make certain to call NSObject's description rather than asking the string interpolator for the swift description
         return allObjects.map { "\($0)" }.joined(separator: separator)
@@ -188,8 +203,15 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
         }
         return false
     }
-    
-    open func description(withLocale locale: Locale?) -> String { return description(withLocale: locale, indent: 0) }
+
+    override open var description: String {
+        return description(withLocale: nil)
+    }
+
+    open func description(withLocale locale: Locale?) -> String {
+        return description(withLocale: locale, indent: 0)
+    }
+
     open func description(withLocale locale: Locale?, indent level: Int) -> String {
         var descriptions = [String]()
         let cnt = count
@@ -406,12 +428,19 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
         getObjects(&objects, range: range)
         return objects
     }
-    
+
+    open func write(to url: URL) throws {
+        let pListData = try PropertyListSerialization.data(fromPropertyList: self, format: .xml, options: 0)
+        try pListData.write(to: url, options: .atomic)
+    }
+
+    @available(*, deprecated)
     open func write(toFile path: String, atomically useAuxiliaryFile: Bool) -> Bool {
         return write(to: URL(fileURLWithPath: path), atomically: useAuxiliaryFile)
     }
-    
+
     // the atomically flag is ignored if url of a type that cannot be written atomically.
+    @available(*, deprecated)
     open func write(to url: URL, atomically: Bool) -> Bool {
         do {
             let pListData = try PropertyListSerialization.data(fromPropertyList: self, format: .xml, options: 0)
@@ -421,7 +450,7 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
             return false
         }
     }
-    
+
     open func objects(at indexes: IndexSet) -> [Any] {
         var objs = [Any]()
         indexes.rangeView.forEach {
@@ -430,7 +459,7 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
         
         return objs
     }
-    
+
     open subscript (idx: Int) -> Any {
         guard idx < count && idx >= 0 else {
             fatalError("\(self): Index out of bounds")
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci please test

Foundation/NSArray.swift
```diff
@@ -853,8 +853,8 @@ open class NSMutableArray : NSArray {
         self.setArray(self.sortedArray(compare, context: context))
     }
 
-    open func sort(_ cmptr: Comparator) {
-        self.sort(options: [], usingComparator: cmptr)
+    open func sort(comparator: Comparator) {
```
phausler: wow is this really how it came out? it really ought to be `sort(using:)` imho by the naming rules...

spevans: https://developer.apple.com/documentation/foundation/nsmutablearray list all sorts of variants:

`func sort(using: [NSSortDescriptor])`
`func sort(comparator: (Any, Any) -> ComparisonResult)`
`func sort(options: NSSortOptions = [], usingComparator: (Any, Any) -> ComparisonResult)`
`func sort((Any, Any, UnsafeMutableRawPointer?) -> Int, context: UnsafeMutableRawPointer?)`
`func sort(using: Selector)`

Looks like no one could make up their mind :)

phausler: This is partially from a breakdown in the importer most likely

spevans: Which importer is that? 

The changes I made to the tests were to make them compile under Darwin Foundation in Xcode, so then I just changed the sclf code to match.

ianpartridge:
There are other differences in `NSMutableArray` too - for example `removeObject(identicalTo: inRange)` should be `removeObject(identicalTo: in)`.  But this looks like a good start 👍

spevans:
Ah, that one didn't get done as there was no unit test covering it so it didn't break when I compiled the tests under Darwin :). I will audit the others and add some tests as well.

ianpartridge:
Great - thanks.

spevans:
@ianpartridge I went through the documentation at https://developer.apple.com/documentation/foundation/nsarray and updated `NSArray.swift` with all the differences I could find, which seemed to be quite extensive for the `init()` methods.
Before I write some unit tests could you give the commit a quick check to make sure Im not making a big incorrect change :)

spevans:
@swift-ci please test

spevans:
@swift-ci please test

Foundation/NSArray.swift
```diff
@@ -27,18 +27,37 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
         }
     }
-    
-    required public convenience init(arrayLiteral elements: Any...) {
+
+    convenience init(objects elements: AnyObject...) {
+        self.init(objects: elements, count: elements.count)
+    }
+
+    convenience init(object: AnyObject) {
```
ianpartridge: This one is [documented](https://developer.apple.com/documentation/foundation/nsarray/1411981-init) as `convenience init(object anObject: Any)`. Not a big deal though.

Foundation/NSArray.swift
```diff
@@ -589,22 +628,42 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
+        do {
+            try self.init(contentsOf: URL(fileURLWithPath: path), error: ())
+        } catch {
+            return nil
+        }
     }
-    
+
+    @available(*, deprecated)
     public convenience init?(contentsOfURL url: URL) {
```
ianpartridge: This one isn't right: https://developer.apple.com/documentation/foundation/nsarray/1410518-init

spevans: Are you sure about this one? That url has the word deprecated in red letters inside a box 5 times in the top half of the page :)

Foundation/NSArray.swift
spevans: I think I saw this one and wasn't sure if NSArray was only supposed to hold references (`AnyObject`) and not all types. I will fix that.

Foundation/NSArray.swift
ianpartridge: Just because it's deprecated doesn't mean it's OK for it to be wrong, does it?

spevans: Ha, hadn't see the mis-spelling, yes I will fix that one s well

Foundation/NSArray.swift
spevans: I removed this one, there was an already existing `convenience init(object anObject: Any)`

spevans:
@swift-ci please test

spevans:
@swift-ci please test

Foundation/NSArray.swift
```diff
@@ -589,22 +624,42 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
...
             return nil
         }
     }
-    
+
+    open func pathsMatchingExtensions(_ filterTypes: [String]) -> [String] {
+        NSUnimplemented()
+    }
+
+    open func shuffled() -> [Any] {
```
phausler: hmm where does this come from?

spevans: https://developer.apple.com/documentation/foundation/nsarray/1640855-shuffled

phausler: That comes from GameplayKit; which I am certain is not under the realms of what we are going to be porting to Linux ;)

spevans: Surely only `func shuffled(using: GKRandomSource)` requires GameKit? A shuffled function might be useful although I can always make it unavailable if not. Im sure there are lots of example interview answers that can be used for an implementation :)

phausler: There are a number of issues with claiming Foundation provides a shuffle - one case is performance, another case is cryptographic security, etc. I think we should not be adding new API to Foundation here and if a developer needs a shuffle they could probably be better serviced by adding a category on Collection and tailor it to their needs performance, cryptographic needs etc.

For now lets just omit this part of the change and address any additional APIs separate from this bug fix

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

TestFoundation/TestNSArray.swift
```diff
@@ -53,6 +56,45 @@ class TestNSArray : XCTestCase {
...
+
+        let array1 = NSArray(array: [foo], copyItems: false)
+        let array2 = NSArray(array: [foo], copyItems: true)
+
+        XCTAssertEqual(array1[0] as! String, "foo")
+        XCTAssertEqual(array2[0] as! String, "foo")
+
+        foo.append("1")
+        XCTAssertEqual(array1[0] as! String, "foo1")
+        XCTAssertEqual(array2[0] as! String, "foo")
```
spevans: @phausler Is this test (which fails on CI) testing the `copyItems` correctly? The test passed against Darwin Foundation but Im not sure if the issue is the test or the `init(array:copyItems:)` constructor.

spevans:
@swift-ci please test

alblue:
@swift-ci please test

ianpartridge:
```
TestFoundation/TestNSArray.swift:96: error: TestNSArray.test_constructorWithCopyItems : XCTAssertEqual failed: ("foo1") is not equal to ("foo") - 
```

spevans:
@ianpartridge Im not sure its that is a bug in my test or a bug in the existing implementation of `init(array:copyItems:)`. The test  worked ok on Darwin with native foundation

spevans:
@swift-ci please test

alblue:
I'm not sure commenting out the test is a good idea. The code works fine on Darwin, for example.

Perhaps it's an issue with the way the strings are converted back and forth with the runtime? The goal for the test is that if you pass in a mutable string into an array, you're allowed to subsequently mutate it.

The question is - is this something that was missing in the first place, or has it started to fail recently?

spevans:
Its a test I added as part of this PR as I was adding a least one call per constructor, although this constructor was already there.

Does the test look valid to you? If there is an easy fix for `init(copyItems:)` then Im happy to add it but given the number of changes in this PR I don't want it to languish for too long and would rather add a fix in a followup if it is not a simple change.

alblue:
OK - can you add the test as is (with the commenting), and file a bug on bugs.swift.org so we can track if this is actually an issue?

spevans:
I filed https://bugs.swift.org/browse/SR-6139

Foundation/NSArray.swift
```diff
@@ -687,22 +742,26 @@ open class NSMutableArray : NSArray {
...
+        super.init(objects: nil, count: 0)
 
         if type(of: self) === NSMutableArray.self {
             _storage.reserveCapacity(numItems)
         }
     }
     
-    public required convenience init(objects: UnsafePointer<AnyObject>!, count cnt: Int) {
-        self.init(capacity: cnt)
+    public required init(objects: UnsafePointer<AnyObject>?, count cnt: Int) {
```
alblue: Small nit: why have `count cnt`? I would just replace `cnt` with `count` throughout the body.

spevans: Its what the documentation has, but maybe thats autogenerated from some other code?
https://developer.apple.com/documentation/foundation/nsarray/1460096-init

alblue: Yeah, it's probably autogenerated from header files where the original implementation used cnt. But count is a far better name and is the name of the parameter in swift, and it doesn't change the API, so using count: Int is better for the Swift version IMHO.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@ianpartridge @alblue Are there any outstanding changes I need to make to this one or is it good to go?

ianpartridge:
LGTM.

spevans:

PR #1189, opened 2017-08-25
For testing Foundation on Android

johnno1962:
Changes I had to make to be able to test Foundation on Android. Results are good considering the differrences of the platform to vanilla Linux:

	 Executed 956 tests, with 15 failures (0 unexpected) in 20.926 (20.926) seconds

There seems to be some problem with TestNSAttributedString.test_enumerateAttributes. An invalid dictionary is returned from CFAttributedStringGetAttributesAndLongestEffectiveRange. I couldn’t puzzle out more than that. I’ve also disabled TestURLSession.test_cancelTask as it seems to result in intermmittent crashes though I’m running Foundation from a week ago and I see related commits. Other changes (to TestNSNumber.swift for example) might be useful for other 32 bit platforms.

Thanks!

PR #1249, opened 2017-10-05
Added build-android script

amraboelela:


.gitignore
```diff
@@ -20,3 +20,6 @@ Build
 *.swp
 *.orig
 .arcconfig
+
+*.tgz
+build.ninjae
```
alblue: Instead of doing this, why not change the sed script so that you're using an already known extension like ~ which will be ignored? Then you don't need to add this.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1325,11 +1325,13 @@ CF_SWIFT_EXPORT void _CFThreadSetName(const char *_Nullable name) {
 }
 
 CF_SWIFT_EXPORT int _CFThreadGetName(char *buf, int length) {
+#if !TARGET_OS_ANDROID
```
alblue: Rather than wrapping this in an `if !` construct, why not add an `#elif TARGET_OS_ANDROID` above the `#elif DEPLOYMENT_TARGET_LINUX` line? Then there's a sensible place to add implementations in the future should it be needed.

Also, do you know why the variable is TARGET_OS_ANDROID rather than DEPLOYMENT_TARGET_ANDROID?

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -27,7 +27,10 @@
 #include <CoreFoundation/ForFoundationOnly.h>
 #include <fts.h>
 #include <pthread.h>
+
+#if !TARGET_OS_ANDROID
```
alblue: Would using the other syntax of `#if __has_include(<execinfo.h>)` work here, to keep it in line with the other tests?

build-android
```diff
@@ -0,0 +1,139 @@
...
+                    ./configure \
+                        Release \
+                        --target=armv7-none-linux-androideabi \
+                        --sysroot=$ANDROID_STANDALONE_SYSROOT \
+                        -DXCTEST_BUILD_DIR=$SWIFT_ANDROID_BUILD_PATH/xctest-linux-x86_64 \
+                        -DLIBDISPATCH_SOURCE_DIR=$SWIFT_PATH/swift-corelibs-libdispatch \
+                        -DLIBDISPATCH_BUILD_DIR=$SWIFT_PATH/swift-corelibs-libdispatch &&
+
+                cp -r /usr/include/uuid $ANDROID_STANDALONE_SYSROOT/usr/include 
+                sed -ie "s/-I.\/ -I\/usr\/include\/x86_64-linux-gnu  -I\/usr\/include\/x86_64-linux-gnu    -I\/usr\/include\/libxml2//" build.ninja
```
alblue: (From above) if this was `sed -i~` instead, you would generate `build.ninja~` as the backup file, which would automatically be excluded.

.gitignore
```diff
@@ -20,3 +20,6 @@ Build
 *.swp
 *.orig
 .arcconfig
+
+*.tgz
```
alblue: It's not clear what this is blocking here, and would prevent people adding .tgz examples in the future (for example). Can you explain why this is needed?

build-android
amraboelela: How about if I just do "sed -i" will it edit without backup? I don't need to generate backup.

spevans: @amraboelela Yes it should do, from the manage:
```
 -i[SUFFIX], --in-place[=SUFFIX]
             edit files in place (makes backup if extension supplied)
```

alblue: I've seen some versions of using `sed -i` when you don't explicitly specify an extension, but it then takes the following word from the command line and uses that as the extension itself. I would recommend using `sed -i~` so that the backups aren't committed to git (the `*~` pattern is already in the `.gitattributes` file) and it avoids any versions of `sed` where running `sed -i something` doesn't get translated as "use something as the backup extension".

.gitignore
alblue: Still need to know why this *.tgz is here. Can you remove this from the commit or justify why it's necessary?

build-android
alblue: This still looks outstanding from GitHub's diffs?

.gitignore
```diff
@@ -20,3 +20,4 @@ Build
 *.swp
 *.orig
 .arcconfig
+
```
alblue: This file doesn't need to be changed, and the addition of a newline doesn't add any value. Can you revert the file to what it was beforehand?

amraboelela:
I am now using -i~ but git status gives:

(use "git add <file>..." to include in what will be committed)

	build.ninja~

Also I am not able to rebase the branch correctly.

alblue:
@swift-ci please test

alblue:
@swift-ci please test

TestFoundation/TestNSLock.swift
```diff
@@ -41,14 +41,11 @@ class TestNSLock: XCTestCase {
         }
 
         let thread = Thread() {
-            lock.lock()
```
alblue: This appears to revert #1261 presumably accidentally. Can you remove this file from this PR?

Foundation/Host.swift
```diff
@@ -28,6 +28,10 @@ open class Host: NSObject {
     internal var _names = [String]()
     internal var _addresses = [String]()
     
+    #if os(Android)
```
xwu: Nit: `#if` and `#endif` should be de-indented (no spaces at the beginning of the line before `#`) and the enclosed statements should be indented as though it isn't surrounded by a compile-time condition (that is, `static internal let NI_MAXHOST` here should align with `static internal let _current`).

Foundation/Host.swift
```diff
@@ -36,7 +40,7 @@ open class Host: NSObject {
     }
     
     static internal func currentHostName() -> String {
-        let hname = UnsafeMutablePointer<Int8>.allocate(capacity: Int(NI_MAXHOST))
+       let hname = UnsafeMutablePointer<Int8>.allocate(capacity: Int(NI_MAXHOST))
```
xwu: Please restore the original indentation.

Foundation/Host.swift
```diff
@@ -65,6 +69,9 @@ open class Host: NSObject {
     }
     
     internal func _resolveCurrent() {
+        #if os(Android)
```
xwu: Same nit as above about indentation of `#if` and `#endif`.

Foundation/Host.swift
```diff
@@ -65,6 +69,9 @@ open class Host: NSObject {
...
+        return
+        #else
```
xwu: Likewise with indentation of `#else`.

Foundation/Thread.swift
```diff
@@ -256,7 +256,11 @@ open class Thread : NSObject {
         let maxSupportedStackDepth = 128;
         let addrs = UnsafeMutablePointer<UnsafeMutableRawPointer?>.allocate(capacity: maxSupportedStackDepth)
         defer { addrs.deallocate(capacity: maxSupportedStackDepth) }
+        #if os(Android)
```
xwu: Same nit here about indentation of `#if`, `#else` and `#endif`.

Foundation/Thread.swift
```diff
@@ -270,6 +274,9 @@ open class Thread : NSObject {
     }
 
     open class var callStackSymbols: [String] {
+        #if os(Android)
```
xwu: Same nit here about `#if`, `#else` and `#endif`; please also de-indent `return []` by four spaces so that it aligns with `return backtraceAddresses`.

alblue:
@amraboelela There was another change recently from @johnno1962 that was merged yesterday regarding Android support: #1189

Can you rebase the changes on top of these and test that your changes still make sense? I'm not sure if there's overlap (the lack of conflicts suggests not) but it would be worth verifying it for Android support.

Separately, if you're interested, you can configure your git repository with `git config pull.rebase true` which means whenever you do a pull, it'll automatically rebase it instead of creating merge nodes. This makes it much easier to keep up-to-date with the branches :)

If you're doing local changes, you might also want to use `git config rebase.autostash true` - this allows you to a pull in place while not losing your local changes.

The above will apply to the single repository that you run in; if you want them to apply to all repositories, run `git config --global` with the above. Hope that's of interest!

Finally, keep up the good work. I know there's been a lot of back-and-forth on this, as well as delays, for which I apologise. I am working to clear the backlog and want to merge these changes in so keep going👍🏻

amraboelela:
All good now?

amraboelela:
@alblue did Apple assign you the task of "cleaning out" all the PR in Foundation? :)

what about swift  repo, I have a related PR there needs to get moving: https://github.com/apple/swift/pull/11870

and in libdispatch: https://github.com/apple/swift-corelibs-libdispatch/pull/302

Foundation/Host.swift
```diff
@@ -88,9 +95,13 @@ open class Host: NSObject {
             }
             ifa = ifaValue.ifa_next
         }
+#endif
     }
     
     internal func _resolve() {
+#if os(Android)
+    return
```
alblue: It looks like the indentation of this return doesn't line up - it should be one indent to the right, I think. @xwu can you confirm?

.gitignore
```diff
@@ -10,6 +10,7 @@ project.xcworkspace
 
 # build files generated by the configure script
 *.ninja
+*.ninja~
```
alblue: Since we're making other changes to this script, I'd suggest putting *~ at the top of the file and removing this line. That way any backup file won't get committed. It's a fairly standard thing to have in Linux repos, but tends to be forgotten on Mac ones.

See e.g. https://github.com/github/gitignore/blob/master/Global/Linux.gitignore#L1

Foundation/Host.swift
```diff
@@ -28,6 +28,10 @@ open class Host: NSObject {
     internal var _names = [String]()
     internal var _addresses = [String]()
     
+#if os(Android)
+        static internal let NI_MAXHOST = 1025
```
alblue: This line looks like it doesn't line up with the `static internal let` below - can you indent it to the left?

Foundation/Thread.swift
```diff
@@ -270,6 +274,9 @@ open class Thread : NSObject {
     }
 
     open class var callStackSymbols: [String] {
+#if os(Android)
+            return []
```
alblue: I think the return here should line up with the `return` of the `return backtraceAddresses` below. It's not rendering as that in the GitHub UI. Maybe a mix of tabs and spaces causing problems?

alblue:
> @alblue did Apple assign you the task of "cleaning out" all the PRs in Foundation? :)

No, I'm just trying to help clear out the queue :)

alblue:
One other thing might help - there's two sorts of changes here; stubbing out APIs that don't work on Android, and the build script to help build it. Can we split those into two PRs? The build-script and the .gitignore can go in one and the fixes to help it build in another.

In swift's apple/swift#11870 repo there seems to be a proposal of moving towards a CMake instead of autotools for building the code. Is that an approach we can think of using here?

amraboelela:
If we don’t include those changes, the build would give errors, so they should go together.

amraboelela:
The cmake proposal is for libdispatch I think. But the cmake script is not ready yet, so I suggest to keep the autoconf build script for now.

alblue:
> If we don’t include those changes, the build would give errors, so they should go together.

I was thinking, maybe we commit the code changes first, then have the build script as a dependent change.

> The cmake proposal is for libdispatch I think. But the cmake script is not ready yet, so I suggest to keep the autoconf build script for now.

OK. We don't have a CMake list for now in foundation, so that makes sense. It would be interesting to know how much it would be to convert it to CMake in the future. Maybe that's something that @compnerd could advise with.

alblue:
@swift-ci please test

Foundation/Host.swift
xwu: Yeah, `return` needs four additional preceding spaces.

Foundation/Thread.swift
xwu: Looks like all spaces here, but agree that `return` needs to be de-indented four spaces.

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -27,10 +27,13 @@
...
 #include <execinfo.h>
+#endif
 
-#if __has_include(<malloc/malloc.h>)
-#include <malloc/malloc.h>
+#if __has_include(<execinfo.h>)
```
xwu: What is going on here now? Why is execinfo.h being conditionally included twice?

Foundation/Host.swift
xwu: Please don't forget to address this comment

Foundation/Host.swift
xwu: This one as well.

Foundation/FileManager.swift
```diff
@@ -343,6 +343,9 @@ open class FileManager : NSObject {
         This method replaces fileSystemAttributesAtPath:.
      */
     open func attributesOfFileSystem(forPath path: String) throws -> [FileAttributeKey : Any] {
+#if os(Android)
+        NSUnimplemented()
```
alblue: In this line, `NSUnimplemented()` means 'This hasn't been implemented yet, but it will be in the future' - whereas `NSUnsupported()` means that 'This will not be implemented on this platform'. Can you confirm this is intended to be implemented in the future? If not, using `NSUnsupported()` would be preferable.

amraboelela: I think we should keep it NSUnimplremented, as someone later should be able to implement it in Android.

alblue: OK

Foundation/Thread.swift
```diff
@@ -256,7 +256,11 @@ open class Thread : NSObject {
         let maxSupportedStackDepth = 128;
         let addrs = UnsafeMutablePointer<UnsafeMutableRawPointer?>.allocate(capacity: maxSupportedStackDepth)
         defer { addrs.deallocate(capacity: maxSupportedStackDepth) }
+#if os(Android)
+        let count = maxSupportedStackDepth
```
alblue: The count here is the number of successful frames returned by `backtrace`, and that they've been filled in by the caller. Since they won't have been any returned, isn't 0 the correct answer in the face of not being implemented? Admittedly this is being stubbed out from the caller below, but if it gets added elsewhere later then it would be better if this didn't contain a known bad number.

build-android
```diff
@@ -0,0 +1,142 @@
...
+#
+# Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+
+set -e
+
+cd "$(dirname $0)/.." || exit 1
```
alblue: Small observation; this changes the directory you're currently in. If you wanted to just find out what directory you are in, you can do it in parenthesis e.g.

https://github.com/apple/swift/blob/cb3fcd4b83410f93558bb449695e1f1c2e69cfb9/utils/swift_build_support/test_swift_build_support.sh#L5

alblue:
@swift-ci please test

alblue:
@swift-ci please test

alblue:
@swift-ci please test

alblue:
@swift-ci please test

alblue:

amraboelela:
Yeaaah :) 
Thank you @alblue

PR #1262, opened 2017-10-12
TestNSLock: Lower thread test from 1000 to 10 threads and 10secs to 2…

spevans:
…secs


spevans:
@swift-ci please test

PR #1263, opened 2017-10-13
NSLocalizedString stub

johnno1962:
For separate review

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -20,6 +20,16 @@ extension unichar : ExpressibleByUnicodeScalarLiteral {
     }
 }
 
+// Placeholder for a future implementation
+public
+func NSLocalizedString(_ key: String,
+                       tableName: String? = nil,
+                       bundle: Bundle = Bundle.main,
+                       value: String = "",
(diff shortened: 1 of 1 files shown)


alblue:
Looks good, same implementation as in https://github.com/apple/swift/blob/485ebefba7ae044c85bcbd4ebf48bfd18c807c2f/stdlib/public/SDK/Foundation/Foundation.swift#L34-L40

Foundation/NSString.swift
```diff
@@ -20,6 +20,16 @@ extension unichar : ExpressibleByUnicodeScalarLiteral {
     }
 }
 
+// Placeholder for a future implementation
```
ianpartridge: Is this comment correct? Haven't you provided the implementation?

johnno1962: I’ll remove it from the rebase.

alblue: The comment from the SDK implementation reads:

/// Returns a localized string, using the main bundle if one is not specified.

Maybe we can use that instead?

johnno1962: Sure, added to the rebase

PR #1274, opened 2017-10-17
Update NSNumberBridging test for init(exactly:)

natecook1000:
A fix in the behavior of FloatingPoint.init(exactly:) means that the NSNumber bridging tests need to be updated. That fix is in apple/swift#11311, this PR replaces #1273.

PR #1275, opened 2017-10-17
Fix NSNumber bridging test for Double.nan

natecook1000:
This catches a test that I didn't update in https://github.com/apple/swift-corelibs-foundation/pull/1274.


natecook1000:
@swift-ci Please test

PR #1280, opened 2017-10-21
Upgrade xcode project to latest Recommended Settings

spevans:
- This enables more clang warnings which finds some issues
  in the CoreFoundation code. Fix some of these.

### DIFF
--- a/Foundation.xcodeproj/project.pbxproj
+++ b/Foundation.xcodeproj/project.pbxproj
@@ -2028,7 +2028,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0830;
-				LastUpgradeCheck = 0830;
+				LastUpgradeCheck = 0910;
 				ORGANIZATIONNAME = Apple;
 				TargetAttributes = {
 					5B5D885C1BBC938800234F36 = {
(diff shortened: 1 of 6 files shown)


spevans:
@swift-ci please test

CoreFoundation/RunLoop.subproj/CFMessagePort.c
```diff
@@ -496,8 +496,8 @@ static CFMessagePortRef __CFMessagePortCreateRemote(CFAllocatorRef allocator, CF
     ctx.retain = NULL;
     ctx.release = NULL;
     ctx.copyDescription = NULL;
-    task_get_bootstrap_port(mach_task_self(), &bp);
-    native = (KERN_SUCCESS == ret) ? CFMachPortCreateWithPort(allocator, port, __CFMessagePortDummyCallback, &ctx, NULL) : NULL;
+    ret = task_get_bootstrap_port(mach_task_self(), &bp);
```
alblue: These couple of lines may need a quick eyeball by @phausler -- the direct context in the GitHub view doesn't necessarily determine that these changes are valid. What was the warning that you were trying to fix here? Unused return or local variable?

alblue: So looking at the changes they look sensible in the context of this change, but it would be good to see if the fixes can be upstreamed efficiently.

spevans: I think they were both reads before being initialised, but the `port` -> `bp` fix is more of an educated guess  :)

parkera: I don't think this is what you want actually. This function really shouldn't exist at all in this environment.

The core idea here is to look up the port by name but that part is missing. We should probably just replace the entire thing with `return NULL`.

parkera: If we want to keep the rest of the code for some reason (documentation purposes?) then we should not bother with `task_get_bootstrap_port` and just make sure `port == NULL` for the rest of the function.

spevans: If the function can be replaced with a `return NULL` then it would be better to just do that. The old body will still be in git for the curious.

spevans:
I removed the `CFMessagePort.c` changes for now, they can be done in a future PR if its still warranted.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

Foundation.xcodeproj/project.pbxproj
```diff
@@ -2763,7 +2775,6 @@
 					"-DCF_CHARACTERSET_UNICODE_DATA_B=\\\\\"CoreFoundation/CharacterSets/CFUnicodeData-B.mapping\\\\\"",
 					"-DCF_CHARACTERSET_UNICHAR_DB=\\\\\"CoreFoundation/CharacterSets/CFUniCharPropertyDatabase.data\\\\\"",
 					"-DCF_CHARACTERSET_BITMAP=\\\\\"CoreFoundation/CharacterSets/CFCharacterSetBitmaps.bitmap\\\\\"",
-					"-Wno-nullability-completeness-on-arrays",
```
alblue: Was this removed by the upgrade process? Is it implied by default now?

spevans: No, I removed it manually. It was added in https://github.com/apple/swift-corelibs-foundation/pull/988 but 
a) I think its a useful warning to have,
b) The `build.ninja` doesnt have this flag so these warnings currently show up on Linux anyway and so this keeps the two builds in sync.

alblue:
Given that Xcode 9.1 has been released, can you see if there are other changes which get added as part of that, and resubmit a newer version?

Foundation.xcodeproj/project.pbxproj
alblue: OK, makes sense.

spevans:
@alblue Apart from bumping the LastUpgradeCheck from `0900` to `0910` there were no other changes.

spevans:
@alblue Any other issues with this?

ianpartridge:
Should we change the `LastUpgradeVersion` to `0910`?

P.S. 👍 to this PR.

alblue:
Sounds good.

spevans:
@alblue I bumped the `LastUpgradeCheck` to `0910`. Is this good to go now?

alblue:
Works for me.

PR #1282, opened 2017-10-24
Synchronize Data.swift from the overlay

phausler:
This pulls in the fixes for:
Guard Data access to NSData._isCompact under availability guards
Ensure that hashing data with zero bytes avoids empty allocations and fix bridged empty data hashes from de-referencing null values (#12509)


phausler:
@swift-ci please test

PR #1285, opened 2017-10-27
TestCodable: Replace fatalError() with XCTFail

spevans:
Having tests abort via `fatalError()` is not as useful as just failing the tests in situations where further tests can still be run.

### DIFF
--- a/TestFoundation/TestCodable.swift
+++ b/TestFoundation/TestCodable.swift
@@ -33,28 +33,17 @@ private func makePersonNameComponents(namePrefix: String? = nil,
     return result
 }
 
-func expectRoundTripEquality<T : Codable>(of value: T, encode: (T) throws -> Data, decode: (Data) throws -> T) -> Bool where T : Equatable  {
-    let data: Data
+func expectRoundTripEquality<T : Codable>(of value: T, encode: (T) throws -> Data, decode: (Data) throws -> T) throws where T : Equatable  {
     do {
-        data = try encode(value)
-    } catch {
-        XCTFail("Unable to encode \(T.self) <\(value)>: \(error)")
-        return false
+        let data = try encode(value)
+        let decoded: T = try decode(data)
+        if value != decoded {
+            throw NSError(domain: "Decode mismatch", code: 0, userInfo: ["msg": "Decoded \(T.self) <\(decoded)> not equal to original <\(value)>"])
+        }
     }
-
-    let decoded: T
-    do {
-        decoded = try decode(data)
-    } catch {
-        XCTFail("Unable to decode \(T.self) <\(value)>: \(error)")
-        return false
-    }
-
-    XCTAssertEqual(value, decoded, "Decoded \(T.self) <\(decoded)> not equal to original <\(value)>")
-    return value == decoded
 }
 
-func expectRoundTripEqualityThroughJSON<T : Codable>(for value: T) -> Bool where T : Equatable  {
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

TestFoundation/TestCodable.swift
```diff
@@ -33,25 +33,28 @@ private func makePersonNameComponents(namePrefix: String? = nil,
...
+        XCTFail("Unable to encode \(T.self) <\(value)>: \(error)")
+        return false
     }
 
     let decoded: T
     do {
         decoded = try decode(data)
     } catch {
-        fatalError("Unable to decode \(T.self) <\(value)>: \(error)")
+        XCTFail("Unable to decode \(T.self) <\(value)>: \(error)")
```
parkera: Should this just be returning NO and leaving the XCTFail for the higher level call?

spevans: I left it in because it acts as a handy debug message in the failure case and isn't any noiseier if the `decode` works.

TestFoundation/TestCodable.swift
```diff
@@ -33,25 +33,28 @@ private func makePersonNameComponents(namePrefix: String? = nil,
...
 
-func expectRoundTripEquality<T : Codable>(of value: T, encode: (T) throws -> Data, decode: (Data) throws -> T) where T : Equatable {
+func expectRoundTripEquality<T : Codable>(of value: T, encode: (T) throws -> Data, decode: (Data) throws -> T) -> Bool where T : Equatable  {
     let data: Data
     do {
         data = try encode(value)
     } catch {
-        fatalError("Unable to encode \(T.self) <\(value)>: \(error)")
+        XCTFail("Unable to encode \(T.self) <\(value)>: \(error)")
+        return false
```
alblue: Personally I would prefer not to have a return value of false which may forget to be called by the calling site. Instead, you should pass in the message into the `expect()` call and then there's no need to have the `XCTAssertTrue` sprinkled throughout the code base.

In other words, I like the idea of fatalError -> XCTFail change, but I don't like the modification of return type to Bool or the other changes in this file.

spevans: If the return value isn't acted upon then a compiler warning is emitted.

The advantage of having the `XCTAssert..` in the `test_()` function is that failing tests show up more easily when running the tests under Xcode, when run as normal unit tests and not via `TestFoundation`. This makes it quicker to know what is failing as Xcode will highlight the failing test rather than having to step though the code.

alblue: If we're going to do something which requires the caller take action (as opposed to merely a warning when it doesn't work) then we should change this from returning a Bool to making it a throw, which will explicitly force unadorned calls to be handled. What do you think?

spevans: Yes, agreed a `throw` is better since it causes and error rather than a warning if not handled. I will add a commit to see what it looks like.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

alblue:
Looks good to me. WDYT @itaiferber ?

itaiferber:
Sorry for the delay; will review today when I get a moment.

itaiferber:
Looks reasonable to me! The stdlib test runner runs every test in a separate process so failing via fatalError is okay over there; this is much better for XCTest infrastructure though.

ianpartridge:

ianpartridge:

PR #1286, opened 2017-10-27
DarwinComptibilityTests: Add Xcode Project

spevans:
This PR adds a new Xcode Project `DarwinCompatibiltyTests` that compiles the current `swift-corelibs-foundation` tests against the native Foundation and XCTest. This allows tests to be validated against the native Foundation and is useful for checking edge cases especially for types such as `NSNumber`, `Decimal` etc. The tests show up as unit tests in Xcode so can be run individually.

All of the tests currently compile, with the help of some compatibility shims, however some of the tests fail. Tests that hang or crash are commented out using the `#if !DARWIN_COMPATIBILTY_TESTS/#endif` pair which is defined in the project's compiler flags.

On 10.12 about 100 tests (185 individual `XCTAsserts`) currently fail for various reasons. Also some methods require 10.13 etc (currently using `if #available()` but this isn't a permanent solution). 

The `xdgTestHelper` is also not currently compiled and is disabled in the tests as I could not find the cookie storage location, but this can be fixed in the future.

This has been tested on 10.12 (Sierra) but not 10.13 (High Sierra). I won't fix all of the tests before this is merged as there will be too many to fix upfront and some may also pass on 10.13 but not 10.12, however it is envisaged that tests will be fixed on an ongoing basis.

Any comments, suggestions welcome.

- Add Xcode project `DarwinCompatibilityTests`
- Add shims and minor test fixes
- Comment out tests that don't compile/run against native Foundation using `DARWIN_COMPATIBILITY_TESTS` flag.


parkera:
Great! This should really help with compatibility.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
I think this is ready for comments/merging now. If someone could test it locally on 10.13 that would be useful as I currently only run 10.12

phausler:
What else is remaining for this to be integrated into CI?

ianpartridge:
@swift-ci please test

spevans:
Fixing the tests will be an ongoing effort and shouldn't prevent this being used for CI in some form. Simply compiling the tests against the native Foundation finds issues where the methods and therefore the tests have been implemented with the incorrect function signatures. I have fixed up all of the current issues in previous PRs so we should be able to proceed with a CI job that invokes an Xcode build of the tests but doesn't run them.

Im not actually sure how to use `xcodebuild` from the command line to just compile the tests, hopefully someone with more knowledge of Xcode can step in. There is also the issue that the Xcode will need the latest nightly toolchain installed although we might be able to get away with installing the last one that was built as the tests rarely use the latest swift features as soon as they are introduced.

I would suggest this being triggered as a manual job for now until we have confidence in how it works e.g., using `please test darwin_compatibility` or some such command.

As to the broken tests they mainly fall into the following groups:

1. Differences between ObjC Foundation and SCLF that will always be different. These can just be disabled.

2. Differences due to floating point representations and other cases where neither is wrong and we just need to decide on how compatible we want to be.

3. Bugs in the SCLF implementation and therefore incorrect tests that need  to be fixed.

4. Bugs that are found in the native foundation. Again these will need to be disabled until they are fixed (if they are fixed) in a newer version of macOS.

spevans:
@alblue Do you have any thoughts on this?

alblue:
I see a couple of crashes when running it against a newer version, including `test_validateMutation_slice_mutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound` and `test_validateMutation_slice_customMutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound` along with a slightly different one in `test_URLComponents_JSON`.

But I think we should migrate this in and then figure out how to fix these things incrementally to get to a clean bill of health.

The only comment I have is that the `DarwinShims.swift` at the top level could be misleading. Could we put that into the TestFoundation or DarwinCompatiblityTests directories instead?

Also, in the sections where you have defined the macro `DARWIN_COMPATIBILITY_TESTS` to stub out explicit frees that are handled by the autorelease pool (e.g. `TestNSKeyedArchiver.swift`) it might be better to use the `DEPLOYMENT_RUNTIME_OBJC` macro instead, because that's really what needs to gate the availability.

spevans:
@alblue I fixed the issues you identified. `TestNSKeyedArchiver.swift` was also the only place I stubbed out the explicit free.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

alblue:

PR #1288, opened 2017-10-30
Fix CFSwiftStringGetBytes on big endian

vivkong:
This will allow `TestNSString.test_swiftStringUTF16` and a few TimeZone related test cases to pass.

### DIFF
--- a/Foundation/NSCFString.swift
+++ b/Foundation/NSCFString.swift
@@ -168,8 +168,13 @@ internal func _CFSwiftStringGetBytes(_ str: AnyObject, encoding: CFStringEncodin
             for idx in 0..<range.length {
                 // Since character is 2 bytes but the buffer is in term of 1 byte values, we have to split it up
                 let character = encodingView[start.advanced(by: idx + range.location)]
+#if _endian(big)
+                let byte0 = UInt8((character >> 8) & 0x00ff)
+                let byte1 = UInt8(character & 0x00ff)
+#else
                 let byte0 = UInt8(character & 0x00ff)
                 let byte1 = UInt8((character >> 8) & 0x00ff)
+#endif
                 buffer.advanced(by: idx * 2).initialize(to: byte0)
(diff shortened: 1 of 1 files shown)


parkera:
So to be clear, this is with data that is UTF16LE on a big endian host?

vivkong:
Thanks for your review.   The previous 2 lines
```
                 let byte0 = UInt8(character & 0x00ff)
                 let byte1 = UInt8((character >> 8) & 0x00ff)
```
 turned the data into LE format.  Maybe a better fix is to modify those 2 lines instead.  I'll rework this.

alblue:
@swift-ci please test

vivkong:
Doesn't seem like the tests were kicked off.  Can we please try again?  Thanks.

spevans:
@swift-ci please test

alblue:
@parkera this looks good to me - WDYT?

vivkong:
Is it possible to have this backported to the 4.0 branch?  Is it via PR?

alblue:
Yeah, you'll have to create a new PR cherry-picking this against the swift-4.0 branch (you can cross-reference this pull request when you do). @parkera will have to explicitly approve it though.

Also, I'm not sure when/if the swift-4.1 branch will pick up changes for master - if there's no further merges happening you might have to cherry-pick there as well.

I'm not sure if the 4.0 branch will build another release or not though, so even if it ends up there you may find we don't build a new release with it in. However, @parkera is the one who can tell you.

I'd suggest getting the PR in today, as next week is Thanksgiving week in the USA and there may be people on vacation during that week.

vivkong:
Thanks @alblue.  I'll try to get one in today.  I saw that this commit is already in swift-4.1-branch.

parkera:
I think we should focus on the Swift 4.1 branch at this time.

parkera:
We should still be auto-merging the `master` branch to `swift-4.1` though, so it'll pick this up automatically.

PR #1290, opened 2017-11-01
[SR-5989] Added missing call to _CFDeinit for NSLocale

ogres:
Resolves  [SR-5989](https://bugs.swift.org/browse/SR-5989)

Added `deinit` to NSLocale that calls to `_CFDeinit` to release identifier and other resources.

I also wanted to add test cases for this, but could not figure it out how to do it.

I wanted to create a `weak` reference to an identifier of the locale and check if it becomes nil after `Locale` goes out of scope, but ...

`Locale` / `NSLocale` returns `Swift String` and as its a `struct`, we won't be able to test with it.
Another idea was to get this directly from `CFLocaleGetValue` but as CoreFoundation objects are automatically memory managed in Swift, I am not sure if this would work. 

Any ideas ?

### DIFF
--- a/Foundation/NSLocale.swift
+++ b/Foundation/NSLocale.swift
@@ -50,6 +50,10 @@ open class NSLocale: NSObject, NSCopying, NSSecureCoding {
         self.init(localeIdentifier: String._unconditionallyBridgeFromObjectiveC(identifier))
     }
     
+    deinit {
+        _CFDeinit(self)
+    }
+    
     open override func copy() -> Any {
         return copy(with: nil)
     }


spevans:
@swift-ci please test

spevans:
@ogres I can't think of a specific test case at the moment, the only thing I can think of at the moment is manually running a valgrind, pre and post patch, to see if the leak has gone.

phausler:
What you have is appropriate, unfortunately we don’t have a facility via XCTest to validate leaks

PR #1291, opened 2017-11-01
[SR-6139] Implement NSString.copy() method

sashabelonogov:
A potential fix for this issue: 
https://bugs.swift.org/browse/SR-6139

Before, the method that is responsible to create the immutable copy of NSString was simply
returning self, that’s why copying didn’t happen and this test was failing: 

```
func test_constructorWithCopyItems() {
        let foo = "foo" as NSMutableString

        let array1 = NSArray(array: [foo], copyItems: false)
        let array2 = NSArray(array: [foo], copyItems: true)

        XCTAssertEqual(array1[0] as! String, "foo")
        XCTAssertEqual(array2[0] as! String, "foo")

        foo.append("1")
        XCTAssertEqual(array1[0] as! String, "foo1")
        // The below test fails on sclf
        XCTAssertEqual(array2[0] as! String, "foo")
    }

```

Please, let me know what you think.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -231,7 +231,15 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
     }
     
     open func copy(with zone: NSZone? = nil) -> Any {
-        return self
+        if type(of: self) === NSString.self {
+            return self
+        }
+        let characters = UnsafeMutablePointer<unichar>.allocate(capacity: length)
+        getCharacters(characters, range: NSMakeRange(0, length))
+        let result = NSString(characters: characters, length: length)
+        characters.deinitialize()
+        characters.deallocate(capacity: length)
+        return result
     }
     
     open override func mutableCopy() -> Any {
--- a/TestFoundation/TestNSArray.swift
+++ b/TestFoundation/TestNSArray.swift
@@ -90,12 +90,9 @@ class TestNSArray : XCTestCase {
 
         XCTAssertEqual(array1[0] as! String, "foo")
         XCTAssertEqual(array2[0] as! String, "foo")
-
-        // Disable this test for now as it fails, althought it works against Darwin Foundation.
-        // The test may not acutally be correct anyway.
-        //foo.append("1")
-        //XCTAssertEqual(array1[0] as! String, "foo1")
-        //XCTAssertEqual(array2[0] as! String, "foo")
+        foo.append("1")
+        XCTAssertEqual(array1[0] as! String, "foo1")
+        XCTAssertEqual(array2[0] as! String, "foo")
     }
     
     func test_enumeration() {
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -99,6 +99,7 @@ class TestNSString : XCTestCase {
             ("test_replacingOccurrences", test_replacingOccurrences),
             ("test_getLineStart", test_getLineStart),
             ("test_substringWithRange", test_substringWithRange),
+            ("test_createCopy", test_createCopy),
         ]
     }
 
@@ -1168,6 +1169,16 @@ class TestNSString : XCTestCase {
         // SR-3363
         let s6 = NSString(string: "Beyonce\u{301} and Tay")
         XCTAssertEqual(s6.substring(with: NSMakeRange(7, 9)), "\u{301} and Tay")
+    }
+    
+    func test_createCopy() {
+        let string: NSMutableString = "foo"
+        let stringCopy = string.copy() as! NSString
+        XCTAssertEqual(string, stringCopy)
+        string.append("bar")
+        XCTAssertNotEqual(string, stringCopy)
+        XCTAssertEqual(string, "foobar")
+        XCTAssertEqual(stringCopy, "foo")
     }
 }
 


spevans:
@swift-ci please test

sashabelonogov:
I'm sorry for a failed build, I rushed with solving merge conflicts in a test and missed one "}" in the end. I fixed it in the same commit. Everything should be fine now.

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

spevans:
@sashabelonogov Could you also uncomment the following lines in `TestFoundation/TestNSArray.swift`

```
//foo.append("1")
//XCTAssertEqual(array1[0] as! String, "foo1")
//XCTAssertEqual(array2[0] as! String, "foo")
```
As this was the original failing test case so checking it now works would be good, thanks.

sashabelonogov:
@spevans sure! I also deleted my test because it's identical to the existing one.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

Foundation/NSString.swift
```diff
@@ -1292,9 +1297,11 @@ extension NSString {
         // manually skip the BOM header.
         guard let cf = CFStringCreateWithBytes(kCFAllocatorDefault, bytePtr + offset, readResult.length - offset,
                                                CFStringConvertNSStringEncodingToEncoding(encoding), true) else {
-            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.fileReadInapplicableStringEncoding.rawValue, userInfo: [
-                "NSDebugDescription" : "Unable to create a string using the specified encoding."
-                ])
+                                                throw NSError(domain: NSCocoaErrorDomain,
```
alblue: Can you revert this reformatting, since it isn't really relevant or part of this change?

TestFoundation/TestNSArray.swift
```diff
@@ -46,7 +46,7 @@ class TestNSArray : XCTestCase {
             ("test_mutableCopying", test_mutableCopying),
             ("test_writeToFile", test_writeToFile),
             ("test_initWithContentsOfFile", test_initWithContentsOfFile),
-            ("test_readWriteURL", test_readWriteURL)
+            ("test_readWriteURL", test_readWriteURL),
```
alblue: Although in principle having a trailing comma makes for less invasive changes in the future, since you aren't actually adding anything in this line then it's not relevant for this commit. Can you revert this line change and then squash the commits?

TestFoundation/TestNSString.swift
```diff
@@ -1169,6 +1170,17 @@ class TestNSString : XCTestCase {
         let s6 = NSString(string: "Beyonce\u{301} and Tay")
         XCTAssertEqual(s6.substring(with: NSMakeRange(7, 9)), "\u{301} and Tay")
     }
+    
+    func test_createCopy() {
+        //SR-6139
```
alblue: This comment isn't particularly relevant - as long as the SR-6139 is mentioned in the commit message, people will be able to identify that it came from this bug in the first place. Can you remove this comment?

sashabelonogov:
@alblue, makes sense, sure.
 The code is updated, rebased and squashed. Thank you for pointing it out.

Foundation/NSString.swift
```diff
@@ -231,7 +231,13 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
...
     open func copy(with zone: NSZone? = nil) -> Any {
-        return self
+        if let contents = _fastContents,
+            type(of: self) === NSString.self || type(of: self) === NSMutableString.self {
+            return NSString(characters: contents, length: length)
+        }
+        let result = NSString()
+        result._storage = self._storage
+        return result
     }
```
parkera: This doesn't make a lot of sense to me. Why shouldn't an immutable `NSString` be allowed to return self for a copy?

sashabelonogov: @parkera you're absolutely right, I didn't think about it. I updated the commit adding the case of the immutable NSString copy, what do you think about it?

sashabelonogov: even though this method will not do what the name of it says, it makes sense to return self for performance reasons

sashabelonogov:
I updated the code handling the case of the immutable NString.copy()

alblue:
Seems to be OK for me now. @parkera is the proposed approach with the `copy` done correctly, or do we have to worry about other immutable string implementations like CF implementations?

Foundation/NSString.swift
parkera: I think the best answer here is going to depend on if we can get a factory pattern proposal through swift-evolution... the way it works in the Darwin Foundation today is that when you use any of NSString's init methods, you're most likely getting a subclass of NSString, which implements copy by returning self. 

For now, I think we should do this:

1. if `type(of: self) === NSString.self` then return self
2. else, use one of NSString's other public initializers to create a new instance with the contents.
3. Don't assume `result._storage` exists, because a subclass would not have access to that and may not use it to hold the contents

sashabelonogov:
@parkera thanks for updates, that makes total sense, I didn't think about subclasses. I updated it to this:
```
    open func copy(with zone: NSZone? = nil) -> Any {
        if type(of: self) === NSString.self {
            return self
        }
        if type(of: self) === NSMutableString.self {
            if let contents = _fastContents {
                return NSString(characters: contents, length: length)
            }
            let result = NSString()
            result._storage = _storage
            return result
        }
        let characters = UnsafeMutablePointer<unichar>.allocate(capacity: length)
        getCharacters(characters, range: NSMakeRange(0, length))
        let result = NSString(characters: characters, length: length)
        characters.deinitialize()
        characters.deallocate(capacity: length)
        return result
    }
```
What do you think? Or should I rather remove the _storage usage at all? I can also replace it with NSString(string: self._bridgeToSwift()) that will do pretty much the same.

parkera:
The part about `NSMutableString` should be in `NSMutableString` itself (subclasses need to override this to provide behavior that works for them, until we have factory patterns).

The rest looks fine, I think.

sashabelonogov:
The NSMutableString part is deleted. I can't disagree with "subclasses need to override this to provide behavior that works for them". It's especially interesting to notice that all the similar methods for other classes are implemented for both cases: mutable and immutable. Thanks for the feedback!

alblue:
@swift-ci please test

sashabelonogov:
@parkera I'm sorry for disturbing you but what is the status with this PR? I updated it with all the changes requested by you.

parkera:
Thanks for the ping. We should merge this.

PR #1292, opened 2017-11-02
Use precondition instead of fatalError in DateInterval initializer

ogres:
Comment of the function says:

> /// - precondition: `end >= start`

but it actually uses `if -> fatalError` instead of `precondition`,

Other functions of `DateInterval` use `precondition`-s, only this one uses 'fatalError', probably makes sense to change it as well.

### DIFF
--- a/Foundation/DateInterval.swift
+++ b/Foundation/DateInterval.swift
@@ -52,10 +52,7 @@ public struct DateInterval : ReferenceConvertible, Comparable, Hashable {
     ///
     /// - precondition: `end >= start`
     public init(start: Date, end: Date) {
-        if end < start {
-            fatalError("Reverse intervals are not allowed")
-        }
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

alblue:
LGTM

parkera:
This is shared with overlay code; we should attempt to keep them in sync.

ogres:
@parkera Should I make the same change here https://github.com/apple/swift/blob/4654a230fb67fc8ba34348c1f5f3758292223260/stdlib/public/SDK/Foundation/DateInterval.swift#L58  ?

parkera:
Yup, that would be great.

ogres:
@parkera here you are https://github.com/apple/swift/pull/12743

PR #1293, opened 2017-11-03
Fix NSNumber bridging test for FloatingPoint(exactly:) fix

natecook1000:
This test change goes with the fix in https://github.com/apple/swift/pull/12739.


natecook1000:
Merging to sync with https://github.com/apple/swift/pull/12739.

PR #1294, opened 2017-11-03
[SR-5582] Changed NSString.appendingPathComponent implementation on Lin…

ogres:
Resolves [SR-5582](https://bugs.swift.org/browse/SR-5582)

Changed NSString.appendingPathComponent implementation on Linux to match [Darwin's](https://developer.apple.com/documentation/foundation/nsstring/1417069-appendingpathcomponent)

Fixed following issues where path.appendingPathComponent(component) would produce incorrect result:

path was empty and component did not start with /
or
path ended with / and component started with /

Before: 

	"/tmp/".appendingPathComponent("scratch.tiff") == "/tmp//scratch.tiff"
 	"".appendingPathComponent("scratch.tiff") == "/scratch.tiff"

After:  

	"/tmp/".appendingPathComponent("scratch.tiff") == "/tmp/scratch.tiff"
	"".appendingPathComponent("scratch.tiff") == "scratch.tiff"

Added `test_appendingPathComponent`.
Moved `test_deletingLastPathComponent` close to related tests ( `test_stringByAppendingPathExtension`, `test_deletingPathExtension`)

Before this commit, NSString and String had their own implementation and code was duplicated.
In this change, `NSString._stringByAppendingPathComponent` directly calls `String`'s internal function ( `_swiftObject._stringByAppendingPathComponent` ). Is this okay? @phausler

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -82,13 +82,13 @@ internal extension String {
     }
     
     internal func _stringByAppendingPathComponent(_ str: String, doneAppending : Bool = true) -> String {
-        if str.length == 0 {
+        if str.isEmpty {
             return self
         }
-        if self == "" {
-            return "/" + str
+        if isEmpty {
+            return str
         }
-        if self == "/" {
+        if hasSuffix("/") {
             return self + str
         }
         return self + "/" + str
@@ -237,16 +237,7 @@ public extension NSString {
     }
     
     internal func _stringByAppendingPathComponent(_ str: String, doneAppending : Bool = true) -> String {
-        if str.length == 0 {
-            return _swiftObject
-        }
-        if self == "" {
-            return "/" + str
-        }
-        if self == "/" {
-            return _swiftObject + str
-        }
-        return _swiftObject + "/" + str
+        return _swiftObject._stringByAppendingPathComponent(str, doneAppending: doneAppending)
     }
     
     public func appendingPathComponent(_ str: String) -> String {
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -77,6 +77,7 @@ class TestNSString : XCTestCase {
             ("test_initializeWithFormat", test_initializeWithFormat),
             ("test_initializeWithFormat2", test_initializeWithFormat2),
             ("test_initializeWithFormat3", test_initializeWithFormat3),
+            ("test_appendingPathComponent", test_appendingPathComponent),
             ("test_deletingLastPathComponent", test_deletingLastPathComponent),
             ("test_getCString_simple", test_getCString_simple),
             ("test_getCString_nonASCII_withASCIIAccessor", test_getCString_nonASCII_withASCIIAccessor),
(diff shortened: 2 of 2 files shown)


Foundation/NSPathUtilities.swift
```diff
@@ -82,15 +82,15 @@ internal extension String {
...
     
     internal func _stringByAppendingPathComponent(_ str: String, doneAppending : Bool = true) -> String {
-        if str.length == 0 {
-            return self
-        }
-        if self == "" {
-            return "/" + str
-        }
-        if self == "/" {
+        if hasSuffix("/") {
```
alblue: I think the isEmpty check should become before the self check, as it was before. Is there a reason why they were reversed?

The change to use hasSuffix is a good one, though, and will work in the self == "/" case.

TestFoundation/TestNSString.swift
```diff
@@ -77,7 +77,6 @@ class TestNSString : XCTestCase {
             ("test_initializeWithFormat", test_initializeWithFormat),
             ("test_initializeWithFormat2", test_initializeWithFormat2),
             ("test_initializeWithFormat3", test_initializeWithFormat3),
-            ("test_deletingLastPathComponent", test_deletingLastPathComponent),
```
alblue: Since this is just moving the contents rather than changing, please leave it here. If necessary, you can add the `test_appendingPathComponent` above the `test_deletingLastPathComponent` to minimise the diffs.

TestFoundation/TestNSString.swift
```diff
@@ -753,50 +754,6 @@ class TestNSString : XCTestCase {
         }
     }
     
-    func test_deletingLastPathComponent() {
```
alblue: As above, don't delete this method and move it elsewhere in the file. If you want to keep the new test with this one, insert it above here.

Foundation/NSPathUtilities.swift
ogres: > I think the isEmpty check should become before the self check, as it was before. Is there a reason why they were reversed?

I thought about it and I think appending a path component to an empty string is less common than appending to actual path, therefore, cases where `isEmpty` evaluates true should be fewer. Most cases will be handled in `hasSuffix` and last line of the function ( `self + "/" + str` )

What do you think about it?

ogres:
Added information about fixes to commit message;
Changed TestNSString.swift to minimise the diffs.

Foundation/NSPathUtilities.swift
alblue: That may be the case, but we now have different short-circuit behaviour on both platforms, and it's really unrelated to this change (which is to fix the bug where appending something onto a string that ends with / results in two // I suspect). I would prefer it to keep it as is (and minimise diffs) in order to fix this bug, and then if it is decided that the re-ordering of the test cases makes sense then it can be done. Several bugs have made it through to production where code was rewritten as:

```
a->something();
if (a==NULL) { abort(); }
```

because a compiler could 'prove' that dereferencing something implied that the variable could never be null, since otherwise would be undefined behaviour and thus the if check could be compiled out completely from the code.

While Swift doesn't suffer from this, the `isEmtpy` check could easily be handling such special-cases (now or in the future) and the re-ordering based on 'a thought' probably isn't the right way to tackle this issue.

Can you commit without the re-ordering on the checks so that you're just fixing the bug, and then squash all the commits so that we don't have interim mess getting in the way? We can then propose via separate PR/bug report that reordering the cases is desirable, with data to back up that particular part of the claim.

ogres: @alblue  ok, what about calling Swift.String's internal function from NSString.append... ?

alblue: I think it makes sense for the calls to be de-duplicated as you have done now; but @parkera or @phausler are the ones to definitively make that call. For me, the less duplicated code we have, the better :)

ogres:
Moved `isEmpty` checks at the beginning of the function

ogres:
@alblue @parkera could you invoke swift-ci to merge? I do not have access for it

ianpartridge:

PR #1295, opened 2017-11-03
FIX re-indent #if OLD_CRAP_TOO codes in CFNumber.c

boraseoksoon:
When analyzing source in CFNumber,
these lines of source codes in #if OLD_CRAP_TOO #endif were not aligned, giving me bad code reading experience.


phausler:
Lets not do this just yet, I have some incoming changes here soon

parkera:
Yes, this code is going to be deleted soon.

PR #1296, opened 2017-11-03
FIX variable naming change in HTTPMessage.swift

boraseoksoon:
var "l" can have ambiguous meaning.
while this function(func byAppending) scope is not broad now, it may be okay.
However, changing this ambiguous variable name to more obvious meaning such as "var lines" would be better in the long term purpose.


alblue:
Closing pull request

PR #1300, opened 2017-11-06
Fix cast that is disallowed by SE-0054.

rudkx:
SE-0054 (Abolish ImplicitlyUnwrappedOptionalType) substantially reduced
the number of places that '!' can be used, but the implementation missed
certain cases like this one. That implementation is being overhauled to
catch all the illegal cases.


rudkx:
@swift-ci Please smoke test

rudkx:
@swift-ci Please test Linux platform

PR #1304, opened 2017-11-07
Better test failure messages in TestCodable

bubski:
### Replaced `"msg"` with `NSLocalizedDescriptionKey`
This makes log messages associated with `NSError` be actually printed, instead of the generic
```
"The operation could not be completed"
```

### Better failure line number

Following the approach from https://github.com/apple/swift/blob/master/test/stdlib/CodableTests.swift
Failing tests will report the line where the failing data was created. This makes tracking down which test cases fail easy.

### DIFF
--- a/TestFoundation/TestCodable.swift
+++ b/TestFoundation/TestCodable.swift
@@ -33,35 +33,42 @@ private func makePersonNameComponents(namePrefix: String? = nil,
     return result
 }
 
-func expectRoundTripEquality<T : Codable>(of value: T, encode: (T) throws -> Data, decode: (Data) throws -> T) throws where T : Equatable  {
+func expectRoundTripEquality<T : Codable>(of value: T,
+                                          encode: (T) throws -> Data,
+                                          decode: (Data) throws -> T) throws where T : Equatable  {
     do {
         let data = try encode(value)
         let decoded: T = try decode(data)
         if value != decoded {
-            throw NSError(domain: "Decode mismatch", code: 0, userInfo: ["msg": "Decoded \(T.self) <\(decoded)> not equal to original <\(value)>"])
+            let errorMessage = "Decoded \(T.self) <\(decoded)> not equal to original <\(value)>"
+            throw NSError(domain: "Decode mismatch", code: 0, userInfo: [NSLocalizedDescriptionKey: errorMessage])
         }
     }
 }
 
-func expectRoundTripEqualityThroughJSON<T : Codable>(for value: T) throws where T : Equatable {
-    let inf = "INF", negInf = "-INF", nan = "NaN"
-    let encode = { (_ value: T) throws -> Data in
-        let encoder = JSONEncoder()
-        encoder.nonConformingFloatEncodingStrategy = .convertToString(positiveInfinity: inf,
-                                                                      negativeInfinity: negInf,
-                                                                      nan: nan)
-        return try encoder.encode(value)
-    }
+func expectRoundTripEqualityThroughJSON<T : Codable>(for value: T, lineNumber: UInt) where T : Equatable {
+    do {
+        let inf = "INF", negInf = "-INF", nan = "NaN"
+        let encode = { (_ value: T) throws -> Data in
+            let encoder = JSONEncoder()
+            encoder.nonConformingFloatEncodingStrategy = .convertToString(positiveInfinity: inf,
+                                                                          negativeInfinity: negInf,
+                                                                          nan: nan)
+            return try encoder.encode(value)
+        }
 
-    let decode = { (_ data: Data) throws -> T in
-        let decoder = JSONDecoder()
-        decoder.nonConformingFloatDecodingStrategy = .convertFromString(positiveInfinity: inf,
-                                                                        negativeInfinity: negInf,
-                                                                        nan: nan)
-        return try decoder.decode(T.self, from: data)
-    }
+        let decode = { (_ data: Data) throws -> T in
+            let decoder = JSONDecoder()
+            decoder.nonConformingFloatDecodingStrategy = .convertFromString(positiveInfinity: inf,
+                                                                            negativeInfinity: negInf,
+                                                                            nan: nan)
+            return try decoder.decode(T.self, from: data)
+        }
 
-    try expectRoundTripEquality(of: value, encode: encode, decode: decode)
+        try expectRoundTripEquality(of: value, encode: encode, decode: decode)
+    } catch {
+        XCTFail("\(error)", line: lineNumber)
+    }
 }
 
 // MARK: - Helper Types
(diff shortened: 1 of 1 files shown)


ianpartridge:
@swift-ci please test

TestFoundation/TestCodable.swift
```diff
@@ -33,35 +33,42 @@ private func makePersonNameComponents(namePrefix: String? = nil,
     return result
 }
 
-func expectRoundTripEquality<T : Codable>(of value: T, encode: (T) throws -> Data, decode: (Data) throws -> T) throws where T : Equatable  {
+func expectRoundTripEquality<T : Codable>(of value: T,
+                                          encode: (T) throws -> Data,
+                                          decode: (Data) throws -> T) throws where T : Equatable  {
```
parkera: Let's not reformat all of these functions.

TestFoundation/TestCodable.swift
```diff
@@ -33,35 +33,42 @@ private func makePersonNameComponents(namePrefix: String? = nil,
...
+                                                                            negativeInfinity: negInf,
+                                                                            nan: nan)
+            return try decoder.decode(T.self, from: data)
+        }
 
-    try expectRoundTripEquality(of: value, encode: encode, decode: decode)
+        try expectRoundTripEquality(of: value, encode: encode, decode: decode)
+    } catch {
+        XCTFail("\(error)", line: lineNumber)
+    }
```
parkera: Why is using XCTFail here directly instead of just propagating the error out a better approach?

TestFoundation/TestCodable.swift
bubski: Do we have any guides for line lengths?

TestFoundation/TestCodable.swift
bubski: To avoid repetitive `do { … } catch { XCTFail("\(error)", line: lineNumber) }` in each unit test.
With this `expectRoundTripEqualityThroughJSON` behaves like an `XCTAssert`.

Unless I misunderstood you question.

bubski:
@parkera I agree with the boilerplate, but can't see a better way of approaching this.

As for helpfulness, I think it is helpful.
These tests have been failing quite often, due to various reasons like
`NSNumber` casting issues
`Calendar` and `TimeZone` misalignment between **Linux** and **Darwin**
floating point handling bugs in `JSONSerialization`
and such.
And it's been always difficult (at least for me) to quickly figure out, based on the error message, what went wrong.

Printing out the object that failed is not often helpful, because descriptions are sometimes not detailed enough to figure out which one it is.

Also, I think that's what you expect from a test failure message. To be able to easily identify point of failure, usually by the line in the test file.

I'd like to hear your opinion on that.

TestFoundation/TestCodable.swift
parkera: In Foundation code, we avoid explicit wraps except under very specific circumstances (preferring to leave it up to the editor to wrap as appropriate).

bubski: Oh, good to know. I'll stick to this rule, thanks.

bubski:
@parkera also, I suppose we _could_ cut down on boilerplate if we called `expectRoundTripEqualityThroughJSON` inline on each tested object,
instead of doing those `for loops`,
and worry about that if we ever add tests for `expectRoundTripEqualityThroughPlist`.

parkera:
Adding `#line` everywhere just adds so much visual noise and potential for confusion... I'd like to explore factoring out that into something common, if we can.

bubski:
@parkera I'll explore further options 👌

ianpartridge:
I'm not a massive fan of the `#line` noise either, but it does make it easier to see exactly where the failure occurred.  Plus, it's [what the overlay tests do](https://github.com/apple/swift/blob/37078133735109937bc630176f1ea8f8914b8f31/test/stdlib/CodableTests.swift#L121-L132) and we should make it as easy as possible to move tests between repos.

bubski:
I definitely agree we should keep those aligned.
But I also don't think we should follow overlay only because it's the way it is, but try and see if it's in fact the best way to go.

Obviously, the best way would be to implement those tests like so:
```swift
func test_NSRange_JSON() {
    expectRoundTripEqualityThroughJSON(NSRange())
    expectRoundTripEqualityThroughJSON(NSRange(location: 0, length: Int.max))
    expectRoundTripEqualityThroughJSON(NSRange(location: NSNotFound, length: 0))
}
```
This would work just fine in **Foundation** at this moment.
In **overlay** however we test both `JSONEncoder` and `PropertyListEncoder` so it makes sense to put the tested objects in `nsrangeValues` and iterate through them in `test_NSRange_JSON` and `test_NSRange_Plist`.

Doing the same in **Foundation** makes sense because we keep the two repos more aligned
and it'll be trivial to add tests for `PropertyListEncoder` when it's implemented, or any other encoder for that matter.

---

I guess what we're left with, is figuring out if passing the line number using `[#line: testObject]` pattern is the best way to have convenient error messages.
I believe line numbers are essential.

The only alternative that comes to my mind is doing something like:
```swift
private struct TestCase<T> {
    let object: T
    let line: UInt

    init(_ object: T, line: UInt = #line) {
        self.object = object
        self.line = line
    }
}

…

private let nsrangeTestCases: [TestCase<NSRange>] = [
    TestCase(NSRange()),
    TestCase(NSRange(location: 0, length: Int.max)),
    TestCase(NSRange(location: NSNotFound, length: 0))
]
    
func test_NSRange_JSON() {
    for testCase in nsrangeTestCases {
        assertRoundTripEqualityThroughJSON(for: testCase.object, line: testCase.line)
    }
}
```
But I can't tell if this is better.
It's just as noisy and not sure if less confusing.

alblue:
I think if you're using this approach, you should capture the file and function as well. The name `TestCase` is likely to get confused with test cases, but a `TestDataPoint` or another name might make sense. And we probably want to avoid labels like `for` since that can conflict with the `for` loop.

Personally, I like the approach of having a data capture like this, as opposed to the `#line` everywhere, because it hides some of the implementation details and permits future work. For example, you could have an extension on that structure such that the `assertRountTripEqualityThroughJSON` is defined on the `TestCase`/`TestDataPoint` itself, in which case you can just do nsrangeTestCase.map($0.assertRoundTrip) to iterate through all of them.

bubski:
`TestDataPoint` sounds better to me too.

>I think if you're using this approach, you should capture the file and function as well.

I'm not sure about `function`. I do not want `nsrangeTestCases` to be logged, but `test_NSRange_JSON`.

>And we probably want to avoid labels like for since that can conflict with the for loop.

I suppose `assertRoundTripEqualityThroughJSON(with: …` would be more consistent with others.
I don't see how using `for` could be a problem though. Could you elaborate or give an example @alblue?

alblue:
I meant conflict as in 'people might get confused with' as opposed to a syntax related issue :) I'm not sure, but I think 'with' has generally been withdrawn from use as well.

parkera:
@itaiferber what do you think?

itaiferber:
As the person who introduced the `#line`s on the overlay side — I don't feel too strongly one way or another about hiding `#line`, though I don't know if there's that strong a benefit either. I found that collecting `#function` and `#file` ended up just being noise and not any more useful than the line number in figuring out where something went wrong. As @bubski points out, the test cases are repeated among multiple encoders and decoders in the overlay as well.

I'm personally in favor of keeping the tests simple and obvious and not wasting too much time on trying to build up functionality that may not end up being useful. `TestCase<T>` above has no more info in it than `(Int, T)` anyway, so I'd say until there's a clear benefit to refactoring like that, this change is sufficient.

itaiferber:
FWIW, too, although we don't need to keep these tests in sync with the overlay just for the sake of consistency, it does make it significantly easier to bring over any new tests we add to the overlay.

itaiferber:
The rest looks good though. Using `NSLocalizedDescriptionKey` is reasonable, and using `XCTFail` in `expectRoundTripEquality` makes it behave more like it does in the overlay, too.

parkera:
If @itaiferber is ok with the `#line` then we can keep it.

alblue:

alblue:

alblue:
@bubski can you rebase and fix the conflicts please?

alblue:
Closing, as this has unresolved conflicts. Please re-open as necessary if the changes get rebased appropriately.

PR #1305, opened 2017-11-07
Import CoreFoundation sources from High Sierra

phausler:
This is the latest changes from CoreFoundation, round tripped to include the latest additions in High Sierra. Careful review should be taken for non x86_64 linux platforms since we have not yet tested other targets. Some header paths were simplified to use __has_include instead of relying on `DEPLOYMENT_RUNTIME_SWIFT` macros. Other areas were unfortunately reverted back to the Darwin implementation since the open source versions caused problems when integrating back into the Darwin version.

Major areas of consideration include the way retain counts are handled and CF type bit fields work. 

The Xcode project now removes the index store disable flag which should vastly improve build times (since that option is no longer needed)

PR #1308, opened 2017-11-09
Fix CFCharacterSet equality detection

sashabelonogov:
I was investigating why the test checking the CharacterSet equality was failing.

Firstly, I found out that the _hashValue of the CFCharacterSet is always 0. The method that checks equality(__CFCharacterSetEqual) is checking the _hashValue, which is correct and the method calculating hash(__CFCharacterSetHash) seemed to do it properly but it seemed that it was simply never called to set the _hashValue field. 
So, as the solution for it, I decided to add this method call at the step of CFCharacterSet initialization when _hashValue should not be 0. After that the tests started to execute correctly, but I'm not sure if it's a correct solution, maybe there is a better way to do it.

Nevertheless, __CFCharacterSetEqual shouldn't rely only on hash but it also should check the contents, which it did, but most likely not correctly. I found out that here: 
```
const UniChar *buf1 = __CFCSetStringBuffer((CFCharacterSetRef)cf1);
const UniChar *buf1End = buf1 + __CFCSetStringLength((CFCharacterSetRef)cf1);
const UniChar *buf2 = __CFCSetStringBuffer((CFCharacterSetRef)cf2);
const UniChar *buf2End = buf2 + __CFCSetStringLength((CFCharacterSetRef)cf2);

 while ((buf1 < buf1End) && (buf2 < buf2End)) {
                        UniChar char1 = *buf1;
                        UniChar char2 = *buf2;

                        if (char1 != char2) return false;

                        do { ++buf1; } while ((buf1 < buf1End) && (char1 == *buf1));
                        do { ++buf2; } while ((buf2 < buf2End) && (char2 == *buf2));
                    }
```
because of `&&` operator in `while ((buf1 < buf1End) && (buf2 < buf2End)) `
 it was not comparing all the characters in both string buffers and replacing it with `||` fixes the issue. 
Also, for some cases we can make the comparison here easier simply by checking the length of the strings beforehand, which is also implemented in this PR.

Please, let me know what you think

### DIFF
--- a/CoreFoundation/String.subproj/CFCharacterSet.c
+++ b/CoreFoundation/String.subproj/CFCharacterSet.c
@@ -896,12 +896,16 @@ static Boolean __CFCharacterSetEqual(CFTypeRef cf1, CFTypeRef cf2) {
 
             case __kCFCharSetClassString:
                 if (isInvertStateIdentical) {
+                    const CFIndex len1 = __CFCSetStringLength((CFCharacterSetRef)cf1);
+                    if (len1 != __CFCSetStringLength((CFCharacterSetRef)cf2)) {
+                        return false;
+                    }
                     const UniChar *buf1 = __CFCSetStringBuffer((CFCharacterSetRef)cf1);
-                    const UniChar *buf1End = buf1 + __CFCSetStringLength((CFCharacterSetRef)cf1);
+                    const UniChar *buf1End = buf1 + len1;
                     const UniChar *buf2 = __CFCSetStringBuffer((CFCharacterSetRef)cf2);
-                    const UniChar *buf2End = buf2 + __CFCSetStringLength((CFCharacterSetRef)cf2);
+                    const UniChar *buf2End = buf2 + len1; // lengths are equal
 
-                    while ((buf1 < buf1End) && (buf2 < buf2End)) {
+                    while ((buf1 < buf1End) || (buf2 < buf2End)) {
                         UniChar char1 = *buf1;
                         UniChar char2 = *buf2;
 
@@ -1320,6 +1324,7 @@ Boolean _CFCharacterSetInitWithCharactersInRange(CFMutableCharacterSetRef cset, 
         if (!__CFCSetGenericInit(cset, __kCFCharSetClassRange)) return false;
         __CFCSetPutRangeFirstChar(cset, theRange.location);
         __CFCSetPutRangeLength(cset, theRange.length);
+        __CFCharacterSetHash(cset);
     } else {
         if (!__CFCSetGenericInit(cset, __kCFCharSetClassBitmap)) return false;
         __CFCSetPutBitmapBits(cset, NULL);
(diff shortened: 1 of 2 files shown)


CoreFoundation/String.subproj/CFCharacterSet.c
```diff
@@ -1365,7 +1371,9 @@ Boolean _CFCharacterSetInitWithCharactersInString(CFMutableCharacterSetRef cset,
         
         if (0 == length) {
             __CFCSetPutHasHashValue(cset, true); // _hashValue is 0
-        } else if (length > 1) { // Check for surrogate
```
alblue: Please don't move the comments as it increases noise in the delta. Can you move that back here, and then it will be clearly seen you're just adding a new line in.

CoreFoundation/String.subproj/CFCharacterSet.c
```diff
@@ -1406,7 +1414,9 @@ CFCharacterSetRef CFCharacterSetCreateWithCharactersInString(CFAllocatorRef allo
 
         if (0 == length) {
 	    __CFCSetPutHasHashValue(cset, true); // _hashValue is 0
-	} else if (length > 1) { // Check for surrogate
+	} else if (length > 1) {
+        __CFCharacterSetHash(cset);
```
alblue: As above, plus the indentation of the `_CFCharacterSetHash` doesn't appear to be indented correctly

CoreFoundation/String.subproj/CFCharacterSet.c
```diff
@@ -1547,7 +1558,7 @@ CFCharacterSetRef CFCharacterSetCreateInvertedSet(CFAllocatorRef alloc, CFCharac
 Boolean _CFCharacterSetInitMutable(CFMutableCharacterSetRef cset) {
     if (!__CFCSetGenericInit(cset, __kCFCharSetClassBitmap| __kCFCharSetIsMutable)) return false;
     __CFCSetPutBitmapBits(cset, NULL);
-    __CFCSetPutHasHashValue(cset, true);
+    __CFCSetPutHasHashValue(cset, true); // Hash value is 0
```
alblue: Was this comment copied and pasted from above? I thought the comment was regarding the condition rather than the action of the line, so I'm not sure it makes sense to add it here - but @phausler would know more, I suspect.

CoreFoundation/String.subproj/CFCharacterSet.c
sashabelonogov: Yes, this comment is not related to the line that I was adding, so I decided to move it below, if you think it's still will be clear where it will "Check for surrogate" - I can leave it where it was.

CoreFoundation/String.subproj/CFCharacterSet.c
sashabelonogov: Will be fixed, thanks for noticing

CoreFoundation/String.subproj/CFCharacterSet.c
sashabelonogov: I simply added it for the consistency, it all the similar places the same comment exist. I actually tend to have "no comment" policy, so I would also go for removing it.

parkera:
I think that https://github.com/apple/swift-corelibs-foundation/pull/1305#issuecomment-343540282 finally has the other character set fixes we needed. We should check the behavior of this bug vs that once it lands too.

sashabelonogov:
Yes, the tests will execute successfully with the changes done in [#1305](https://github.com/apple/swift-corelibs-foundation/pull/1305), surely the comparison of buffers with memcmp idea is better than what is implemented in this PR. The simple comparison of lengths is also implemented in the other PR. I will check the situation with _hashValue later.

But anyways, it means that most (if not all of them) of the changes implemented in this PR is not relevant anymore.

alblue:
Since #1305 has been merged (then reverted in #1328 and then re-reverted in #1338) I don't think this PR is necessary as is. However, it might be worth testing whether or not the hashValue change can be applied as a subsequent PR if needed.

PR #1309, opened 2017-11-09
Fixed _CFThreadSetName line for Android

amraboelela:


alblue:
Does Android only allow the current thread's native name to be changed?

spevans:
If might be better to modify `_CFThreadSetName` as that already checks the thread is `pthread_self()` on Darwin platforms, so adding Android in there will keep the checks in one place.

amraboelela:
The error happens in this line **if let thread = _thread {**
Do you want to change  **_CFThreadSetName(thread, name ?? "" )** 
to make the thread parameter optional?

alblue:
Leave the Swift code as is, do the fix in the c function instead was @spevans suggestion, I believe.

amraboelela:
any suggestion of how to fix the c code?

spevans:
@amraboelela I mis-read your original patch and thought the thread name could only be set for the current thread hence my suggestion of fixing `_CFThreadSetName`.

However on re-reading it I think the issue might be because on android the `_thread` is not an optional it is defined as:

```
#if os(Android)
    private var _thread = pthread_t()
#else
    private var _thread: pthread_t? = nil
#endif
```

I wonder if the special case for Android could be removed so it just uses the `pthread_t?` that other platforms use? That may be sufficient to fix it.

amraboelela:
But there must be a reason why they did this code:

```
#if os(Android)
    private var _thread = pthread_t()
#else
    private var _thread: pthread_t? = nil
#endif
```

spevans:
I don't think its correct. It used to be like that for Linux as well however `_thread = pthread_t()` sets _thread to a pthread_t with value 0 which is a valid ID that should not be used for an uninitialised thread ID. I changed it for Linux but not for Android as I could not test it for Android. So I would just try changing it and seeing if the tests still work.

amraboelela:
Ok let me check it later

amraboelela:
@spevans  I made the update and was able to build it successfully for Android

spevans:
@swift-ci please test

alblue:
Can we squash these two commits please?

amraboelela:
Done

alblue:

PR #1312, opened 2017-11-10
Change NSURL.description to match Darwin

jaysk1:
Consider this program
```swift
import Foundation

let nsurl = NSURL(string: "http://google.com")!
let url = URL(string: "http://amazon.in")!

print(url)
print(nsurl)
```

Output on Darwin
```
http://amazon.in
http://google.com
```

Output on Linux
```
<CFURL 0x8eb3ed0 [0x7f7f63d54c38]>{string = http://amazon.in, encoding = 134217984, base = (null)}
<CFURL 0x8e41620 [0x7f7f63d54c38]>{string = http://google.com, encoding = 134217984, base = (null)}
```

This PR makes the output on Linux same as that on Darwin.

### DIFF
--- a/Foundation/NSURL.swift
+++ b/Foundation/NSURL.swift
@@ -254,7 +254,7 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
     }
     
     open override var description: String {
-        return CFCopyDescription(_cfObject)._swiftObject
+        return self.absoluteString
     }
 
     deinit {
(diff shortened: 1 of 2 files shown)


ianpartridge:
Thanks! Can we add a test too?

jaysk1:
@ianpartridge : Added Test case to test URL.description under TestURL.swift.

alblue:

pushkarnk:

PR #1313, opened 2017-11-10
Implementing HTTPCookieStorage.description

vipinmenon:
Implementation of the description property for HTTPCookieStorage, to remain as close as possible to its behavior on darwin.
For this program:
```
import Foundation

print(HTTPCookieStorage.shared)
print(HTTPCookieStorage.sharedCookieStorage(forGroupContainerIdentifier: "test"))
```
... the output on Darwin is:
```
<NSHTTPCookieStorage cookies count:12>
<NSHTTPCookieStorage cookies count:1>
```
With this PR the output on Linux will be
```
<HTTPCookieStorage cookies count:12>
<HTTPCookieStorage cookies count:1>
```

### DIFF
--- a/TestFoundation/TestHTTPCookieStorage.swift
+++ b/TestFoundation/TestHTTPCookieStorage.swift
@@ -109,7 +109,7 @@ class TestHTTPCookieStorage: XCTestCase {
 
         storage.setCookie(simpleCookie)
         XCTAssertEqual(storage.cookies!.count, 0)
-        XCTAssertEqual(storage.description, "<HTTPCookieStorage cookies count:0>")
+        XCTAssertEqual(storage.description, "<NSHTTPCookieStorage cookies count:0>")
 
         let simpleCookie0 = HTTPCookie(properties: [   //no expiry date
            .name: "TestCookie1",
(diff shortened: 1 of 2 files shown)


ianpartridge:
Thanks! I think we should include the `NS` prefix to match Darwin.

TestFoundation/TestHTTPCookieStorage.swift
```diff
@@ -109,6 +109,7 @@ class TestHTTPCookieStorage: XCTestCase {
 
         storage.setCookie(simpleCookie)
         XCTAssertEqual(storage.cookies!.count, 0)
+        XCTAssertEqual(storage.description, "<HTTPCookieStorage cookies count:0>")
```
ianpartridge: Can we do the new testing in a `test_description()` test instead?

ianpartridge:
@vipinmenon are you able to make the suggested changes?

ianpartridge:
I've made the changes to match Darwin myself, so we can get this merged. The test can be split out later.

@swift-ci please test

ianpartridge:
@swift-ci please test

PR #1314, opened 2017-11-11
[SR-6361] Fix Data.withUnsafeMutableBytes() for slices with length < range.lowerBound

sashabelonogov:
Solves [SR-6361](https://bugs.swift.org/browse/SR-6361) 
This code snippet:
```
var data = Data(bytes: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])[4..<6]
data.withUnsafeMutableBytes { (ptr: UnsafeMutablePointer<UInt8>) in
    ptr.advanced(by: 1).pointee = 0xFF
}
```
crashes with:
`fatal error: UnsafeMutableRawBufferPointer with negative count`

because `count` was not calculated correctly (`_offset` was not taken into account)

The existing tests were not covering this case because it works correctly for data slices when _length > range.lowerBound.

I created new tests for this exact situation but I can also update the existing ones by adding one more check to them.

Please, let me know what you think.

### DIFF
--- a/TestFoundation/TestNSData.swift
+++ b/TestFoundation/TestNSData.swift
@@ -509,6 +509,12 @@ class TestNSData: XCTestCase {
 //            ("test_sliceEnumeration", test_sliceEnumeration),
             ("test_sliceInsertion", test_sliceInsertion),
             ("test_sliceDeletion", test_sliceDeletion),
+            ("test_validateMutation_slice_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_immutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_immutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_mutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_mutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_customBacking_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_customBacking_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_customMutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound",
+             test_validateMutation_slice_customMutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound),
         ]
     }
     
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

my0114:
Just curious, maybe Data.withUnsafeBytes() followed the same logic and should be fixed too?

TestFoundation/TestNSData.swift
```diff
@@ -509,6 +509,12 @@ class TestNSData: XCTestCase {
 //            ("test_sliceEnumeration", test_sliceEnumeration),
             ("test_sliceInsertion", test_sliceInsertion),
             ("test_sliceDeletion", test_sliceDeletion),
+            ("test_validateMutation_slice_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_immutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_immutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_mutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_mutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_customBacking_withUnsafeMutableBytes_lengthLessThanLowerBound", test_validateMutation_slice_customBacking_withUnsafeMutableBytes_lengthLessThanLowerBound),
+            ("test_validateMutation_slice_customMutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound",
+             test_validateMutation_slice_customMutableBacking_withUnsafeMutableBytes_lengthLessThanLowerBound)
```
alblue: If you finish this line with a comma (like in like 511) then others won't need to change line 517 when they add more tests in the future.

alblue:
@phausler the tests pass on this; do you think the change is reasonable?

TestFoundation/TestNSData.swift
sashabelonogov: Thanks @alblue, I updated the commit

sashabelonogov:
@my0114 thanks for your feedback! You might be right, I will check it in a couple of hours to make sure.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

sashabelonogov:
thanks @phausler, sure, this is the PR: https://github.com/apple/swift/pull/12898

sashabelonogov:
@phausler I'm sorry for changing this PR after review but I tried to investigate the issue further and realized that simple `count: Swift.min(range.count, _length)))` is simpler and will do the job, so I updated both PRs.
I hope I understood everything correctly and it makes sense. Thanks.

PR #1319, opened 2017-11-14
Remove macOS @available version restrictions.

spevans:
Im not sure if this is valid, but restricting some methods to certain versions of macOS doesn't seem necessary here.

At the same time, some of there `if #available` that rely on Foundation don't seem to be needed given that they are contained in this version of Foundation.

Tested on linux and macOS 10.12

@parkera @phausler Is this ok or am I on the wrong track here?

### DIFF
--- a/Foundation/Calendar.swift
+++ b/Foundation/Calendar.swift
@@ -1051,83 +1051,45 @@ public struct Calendar : Hashable, Equatable, ReferenceConvertible, _MutableBoxi
     }
     
     internal static func _toNSCalendarIdentifier(_ identifier: Identifier) -> NSCalendar.Identifier {
-        if #available(OSX 10.10, iOS 8.0, *) {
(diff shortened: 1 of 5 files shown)


phausler:
This will likely break our capability of copy/paste to and from the overlay

parkera:
I agree with @phausler on trying to keep the source code in sync; that's the best option we have while we need to have these things in separate repositories.

I assume that generated interfaces would show these APIs as available since Linux and other platforms fall under the `*`.

alblue:
I concur - although these are no-ops on Linux, they will introduce problems in future sync updates, so leaving them as is will be more appropriate.

spevans:
I had forgotten about the code-sync issue and the only real value of this PR was it slightly simplified some of the tests in the Darwin compatibiltity testing (https://github.com/apple/swift-corelibs-foundation/pull/1286) so its not really that important.

PR #1321, opened 2017-11-16
Change JSONSerialization to escape solidus like macOS does

norio-nomura:
Fixes [SR-6401](https://bugs.swift.org/browse/SR-6401)


ianpartridge:

norio-nomura:
Thanks!

PR #1315, opened 2017-11-11
Update API names for NSMutableArray

bubski:

PR #1323, opened 2017-11-17
[SR-6419] Use `@available(*, unavailable)` instead of inappropriate `@available(*, obsoleted: 4.0)`

ikesyo:
Fixes [SR-6419](https://bugs.swift.org/browse/SR-6419).

### DIFF
--- a/Foundation/NSNotification.swift
+++ b/Foundation/NSNotification.swift
@@ -188,7 +188,7 @@ open class NotificationCenter: NSObject {
         })
     }
 
-    @available(*,obsoleted:4.0,renamed:"addObserver(forName:object:queue:using:)")
+    @available(*, unavailable, renamed: "addObserver(forName:object:queue:using:)")
     open func addObserver(forName name: NSNotification.Name?, object obj: Any?, queue: OperationQueue?, usingBlock block: @escaping (Notification) -> Void) -> NSObjectProtocol {
         return addObserver(forName: name, object: obj, queue: queue, using: block)
     }


alblue:
One of the reasons for having the obsoleted API is that it should allow users (who are not using the escaping closure) to perform a rename of the call. However, for trailing closures whose API is not being called this probably doesn't make sense.

I'm not sure if the right fix is to allow the obsoleted calls to be filtered from the swift compiler if it can't find an exact match, or to remove this overload (added in 7eefc15a08fb3e894e8c47e9656c7f93f1cf3997 as part of a different change).

@ianpartridge what do you think, since you reviewed the addition of the obsoleted call?

ianpartridge:
This is an interesting issue. I didn't consider the case of someone using trailing closure syntax and hence not providing the parameter name.

I'm curious how this works on Darwin, surely the same issue would be seen there? It obviously isn't seen there though, so I must be missing something...

ikesyo:
I think we usually don't keep obsoleted APIs (e.g. https://github.com/apple/swift-corelibs-foundation/pull/1315) so

> One of the reasons for having the obsoleted API is that it should allow users (who are not using the escaping closure) to perform a rename of the call

would not be a rationale.

alblue:
As is trivial to search in the codebase, this isn't necessarily always the case:

https://github.com/apple/swift-corelibs-foundation/search?utf8=✓&q=renamed&type=

The hints are here to allow the fix-its to move forward. Though I wonder, @ianpartridge, whether we should have used 'unavailable' rather than 'obsoleted' here? Would that solve the problem?

Foundation/NSNotification.swift
```diff
@@ -188,11 +188,6 @@ open class NotificationCenter: NSObject {
         })
     }
 
-    @available(*,obsoleted:4.0,renamed:"addObserver(forName:object:queue:using:)")
```
alblue: Instead of deleting this, does changing 'obsoleted' to 'unavailable' fix the problem?

parkera:
We used `obsolete` to attempt to help people migrate from Swift 2/3 and 3/4. The migrator keyed off them. The idea was, though, that as soon as that release was in the past, we could remove the obsolete API.

alblue:
You get the same effect with `unavailable` as well, except that new code cannot refer to `unavailable` (but in this case, it can, which is causing the error).

ianpartridge:
@ikesyo @parkera do you think `@available(*, unavailable, renamed: "addObserver(forName:object:queue:using:)")` would be a better solution? Or should we just merge this PR?

alblue:
Let’s check if the unavailable option works and doesn’t cause problems with the original issue, and that it allows the fixme transition to work if not using a trailing closure. Only if we find an issue with that approach should we consider deleting it.

ikesyo:
Or maybe we should explicitly use `swift` instead of `*`.

https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Attributes.html

> An `available` attribute specifying a Swift version availability can’t use the asterisk.

ikesyo:
Updated to use `@available(swift, ...)` as I said above.

ikesyo:
@alblue @ianpartridge Please re-check this.

ikesyo:
cc @parkera

alblue:
Please update this to use unavailable as requested, thanks.

ikesyo:
Hm I’m not sure why `obsoleted` was used here if `unavailable` is preferable but I will do it anyway.

ikesyo:
@alblue Done

ikesyo:
@swift-ci Please test

alblue:
@swift-ci please test

alblue:

spevans:

alblue:

PR #1325, opened 2017-11-17
[SR-6420] Ensure that the pthread_key is initialised once

alblue:


alblue:
@swift-ci please test

parkera:
How are we calling `_CFTSDInitialize` more than once in the first place?

alblue:
The table is a thread-local, so it is called once per thread. If you submit many async dispatch blocks that trigger a CF code path, this table gets assigned as a thread local per dispatch worker thread.

alblue:
Need to check under the load test that found this problem that it resolves the issue. Please hold off merging until then.

parkera:
If we're calling `__CFInitialize` more than once we have serious problems. That is once per process start.

parkera:
And that is the only place we call the `_CFTSDInitialize` function...

alblue:
You’ll note that there is a call to this inside the same file on line 666 which is the one causing the duplicate calls.

parkera:
You're right; that's different than the Darwin Foundation.

I think the right solution here is actually just to delete the call from CFPlatform.c.

parkera:
That is unless somehow this call happens before __CFInitialize, but I hope not. That's supposed to be first.

alblue:
Will have a look next week to see why the call is there. It may have been added because the CFInitialize wasn’t being called first, and so this was added as a workaround. Or it may be unnecessary. We have a test case that can verify either solution so will get to that on Monday.

alblue:
@parkera the change (adding the initialize) was dropped into a mega-drop in a455cde38e06052385560d0a7c5aeb5a4991f844 by @phausler last year. Given that this patch allows the call to happen at most once per process, I'm going to merge this as is, and then maybe @phausler knows why the call to adding the initialize was required subsequently.

alblue:

parkera:
I'm not happy at all about merging this without getting to the root cause.

PR #1330, opened 2017-11-20
build: migrate to the new ELF registration scheme

compnerd:
The new ELF registration scheme uses linker tables and only requires a
single object to be added to the link.

### DIFF
--- a/lib/product.py
+++ b/lib/product.py
@@ -71,8 +71,7 @@ class Product(BuildAction):
         return Path.path(Configuration.current.build_directory.path_by_appending(self.name).absolute() + "/" + self.PROJECT_HEADERS_FOLDER_PATH)
 
 class Library(Product):
-    conformance_begin = ""
-    conformance_end = ""
+    runtime_object = ''
     rule = None
     def __init__(self, name):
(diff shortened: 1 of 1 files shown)


alblue:
Does this require synchronous testing across all repositories with this change? Do you know when it's planned to land?

compnerd:
@alblue yeah, it requires synchronous testing :-(.  I was looking into a final failure with the ELF migration.  There is a single failure in a single test case now (associated type reflection for a remote image which is not loaded but rather mapped into memory).  It is part of the same ELF metadata loading change for now at least.

compnerd:
This was tested as part of apple/swift#12758

aschwaighofer:
Hi @compnerd this got auto merged to swift-4.1-branch of corelibs. And is breaking there (https://ci.swift.org/job/oss-swift-4.1-incremental-RA-linux-ubuntu-16_04-long-test/258/)

I have reverted it on that branch: https://github.com/apple/swift-corelibs-foundation/commits/swift-4.1-branch

Because I believe if we want it there we also need to port apple/swift#12758 to swift's swift-4.1-branch.

PR #1331, opened 2017-11-21
Fix CFNumber for Float on s390x

vivkong:
Needs to set value on the most significant bits.

It was causing test cases with NSNumber(value: Float(1)) to fail.


alblue:
Does the BITSFORFLOATONE not have the correct value on s390x? Also, since the value is technically a 32 bit number, doesn't <<32 just end up assigning it to zero instead?

Can we add some additional tests that capture the exact bits that you see so that we know what's going on, and how we can avoid it in future?

vivkong:
Thanks for the review @alblue.

In Swift 3.x, to initialize a NSNumber for a `Float`, ` _CFNumberInit() `is called and the `CFNumberRef` is initialized by `case kCFNumberFloat32Type: memmove((void *)&result->_pad, valuePtr, 4); `
for a `Float` which sets the most significant bits of `result` on big endian.  

In 4.x, `CFNumberCreate()` is called instead.  For special case of `Float(1)`, `kCFNumberFloat32One` is assigned to the resulting `CFNumberRef`.

Yes `BITSFORFLOATONE` is 32-bit but `__CFNumber._pad` is `uint64_t`so when we assign the special 32-bit value for `Float(1)` and later retrieve it in 64-bit, it led to the problem we see here.  This is a special case and test cases that deals with ` NSNumber(value: Float(1))` will verify everything is correct.  I've added more test cases for that.

I didn't see `CFNumberCreate()` being called with `CFNumberType` of `kCFNumberFloat32Type` anywhere else in Foundation (for `Float(1)`) in 3.x.  I'm guessing this code wasn't executed before on s390x and therefore it wasn't caught until now.

alblue:
Seems reasonable. We've got the tests covering it as well so we'll catch it if there's any issues subsequently.

@swift-ci please test

vivkong:
Doesn't look like the test was triggered.  Can we try again?  Thanks.

ianpartridge:
@swift-ci please test

alblue:

PR #1332, opened 2017-11-21
copying libFoundation.so to android directory and not to armv7

a6a8:
copying libFoundation.so to armv7 was not recognized by the ld tool


alblue:

PR #1335, opened 2017-11-24
Fix typo in DateFormatter method dateFormat:fromTemplate

njuri:


ianpartridge:
Surprisingly this is actually wrong on Darwin too: https://developer.apple.com/documentation/foundation/dateformatter/1408112-dateformat

As we try to match Darwin as closely as possible I don't think we should change this.

spevans:
Might require @parkera to do another documentation fix first.

parkera:
This is the internal name so it should be fine to change it. I'll just fix it up in the Foundation header for a future release.

parkera:
Ah, now I remember why it is spelled this way. I believe it was an Objective-C++ compatibility hack, as `template` is a reserved keyword.

alblue:
@parkera so do we want to close this, or would changing to a different name (e.g. `tmplt`) make sense?

parkera:
Let's just leave this as-is.

PR #1337, opened 2017-11-27
Fix warning: overlapping access to 'self'

ianpartridge:
> Warning: Overlapping accesses to 'self', but modification requires exclusive access; consider copying to a local variable

There are more of these in the codebase, but here's a fix for a few.

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -583,8 +583,9 @@ open class NSCalendar : NSObject, NSCopying, NSSecureCoding {
         var at: CFAbsoluteTime = date.timeIntervalSinceReferenceDate
         
         let res: Bool = withUnsafeMutablePointer(to: &at) { t in
+            let copy = vector
             return vector.withUnsafeMutableBufferPointer { (vectorBuffer: inout UnsafeMutableBufferPointer<Int32>) in
-                return _CFCalendarAddComponentsV(_cfObject, t, CFOptionFlags(opts.rawValue), compDesc, vectorBuffer.baseAddress!, Int32(vector.count))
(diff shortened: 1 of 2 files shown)


Foundation/NSCalendar.swift
```diff
@@ -583,8 +583,9 @@ open class NSCalendar : NSObject, NSCopying, NSSecureCoding {
         var at: CFAbsoluteTime = date.timeIntervalSinceReferenceDate
         
         let res: Bool = withUnsafeMutablePointer(to: &at) { t in
+            let copy = vector
             return vector.withUnsafeMutableBufferPointer { (vectorBuffer: inout UnsafeMutableBufferPointer<Int32>) in
```
spevans: Would `let count = Int32(vector.count)` be better to reduce the size of the copy?

TestFoundation/TestNSData.swift
```diff
@@ -924,13 +924,14 @@ extension TestNSData {
...
         // Get the underlying data for hello2
+        let copy = mutatingHello
         mutatingHello.withUnsafeMutableBytes { (bytes : UnsafeMutablePointer<UInt8>) in
             XCTAssertEqual(bytes.pointee, 0x68, "First byte should be 0x68")
             
             // Mutate it
             bytes.pointee = 0x67
             XCTAssertEqual(bytes.pointee, 0x67, "First byte should be 0x67")
-            XCTAssertEqual(mutatingHello[0], 0x67, "First byte accessed via other method should still be 0x67")
+            XCTAssertEqual(copy[0], 0x67, "First byte accessed via other method should still be 0x67")
```
spevans: Isn't this copy making the test redundant?
Would moving the `XCTAssertEqual(mutatingHello[0] ... ` outside of the closure give the same affect?

ianpartridge:
@swift-ci please test

spevans:
Those fixes look good.

ianpartridge:

ianpartridge:

PR #1338, opened 2017-11-27
Re apply high sierra import

phausler:
This reverts the revert of the High Sierra import and applies a guard against UMS_UK usage which only exists on ICU versions 55 and greater (which does not exist on Ubuntu 14)


phausler:
@swift-ci please test

phausler:
@slavapestov is there a way we can actually trigger ubuntu 14 to be tested so that we dont have to revert this again?

spevans:
The changes to `CFTimeZone.c` undo https://github.com/apple/swift-corelibs-foundation/pull/1126 which made the timezone code work on pre 10.13 macOS. Can the CFTimeZone.c changes be reverted if the  High Sierra changes don't actually add any new code?

phausler:
@spevans I can re-apply that in a subsequent PR

phausler:
@shahmishal do you know if there is a way to request a Ubuntu 14 test?

parkera:
We don't have anything at the moment other than merging, so we need to just put this in.

PR #1339, opened 2017-11-27
SR-6405: URLRequest: Implement normalization for httpMethod

ianpartridge:
Darwin performs normalization of `httpMethod` for 6 specific HTTP methods. Match this behaviour.

### DIFF
--- a/Foundation/NSURLRequest.swift
+++ b/Foundation/NSURLRequest.swift
@@ -159,6 +159,8 @@ open class NSURLRequest : NSObject, NSSecureCoding, NSCopying, NSMutableCopying 
         guard aDecoder.allowsKeyedCoding else {
             preconditionFailure("Unkeyed coding is unsupported.")
         }
+
+        super.init()
         
         if let encodedURL = aDecoder.decodeObject(forKey: "NS.url") as? NSURL {
             self.url = encodedURL._swiftObject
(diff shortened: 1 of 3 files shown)


ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

ianpartridge:
@swift-ci please test

alblue:
I'm not particularly happy with this change, as it seems to be based on an implementation assumption that it outside of the specification of this method. Would like @phausler to opine in case there's some other assumption about behaviour here, such as whether there's a normalised cache of httpMethods.

alblue:
NB I'm not disagreeing that this is the behaviour on Darwin - and in fact, there are several places in the code where it only works if the httpMethod is e.g. GET instead of gEt - but I'd like to try and understand what the correct behaviour should be implemented as.

ianpartridge:
This PR matches what I reverse engineered from Darwin. I'm sure there are other aspects of normalization that aren't covered here (e.g. HTTP header names) but those can wait for another day...

alblue:
OK, having looked into the way it works on Darwin, I agree that it does do a resolution based on a routine in CFNetwork, and then uses those in place of the supplied one if it finds them. So I withdraw my objections :)

alblue:

PR #1340, opened 2017-11-27
Revert "[SR-6420] Ensure that the pthread_key is initialised once"

parkera:
Reverts apple/swift-corelibs-foundation#1325

I am leaving this open temporarily, as a placeholder for figuring out the root cause of the original regression.


parkera:
Alright, file the JIRA please.

alblue:
Filed SR-6491

PR #1163, opened 2017-08-08
SR-5596: TimeZone.current returns wrong value

spevans:
- The length passed to CFStringCreateWithBytes() included an extra
  random byte after the NUL. This meant the string was sometimes
  interpreted as non-ASCII (if the byte was >= 0x80) and the
  timezone would come back as NULL thus defaulting to GMT.

This was broken in https://github.com/apple/swift-corelibs-foundation/pull/1126 which dealt with the macOS timezone directory

PR #1341, opened 2017-11-27
Determine the Timezone directory to use at runtime on macOS.

spevans:
- The timezone directory changed on macOS 10.13 so get the os version
  at runtime and set the timezone directory dynamically. This allows
  the tests to be run on 10.12 and 10.13

- Added __InitTZStrings() for both Linux and macOS to work the same
  way as Windows.

Reapplication of #1126 / #1163 after revert in #1338


spevans:
@swift-ci please test

alblue:

PR #1342, opened 2017-11-28
DateFormatter: disable testing of medium date style

ianpartridge:
Currently, some of the `DateFormatter` tests fail on High Sierra because the format of the `.medium` date style has changed.  This is likely due to changes in the underlying platform ICU.

@parkera expressed the opinion to me that we shouldn't be testing the appearance of the `DateFormatter` presets because they are subject to change and not precisely documented.

I don't entirely agree with that point of view as if the format does change on Darwin platforms then we want to reflect those changes in SCLF, and having our tests start failing is the best way to detect that.

However, there is a whole separate discussion about whether Swift on Linux should use the platform ICU at all, or whether we should ship Apple's fork of ICU (which is used on Darwin) with Swift on Linux to sidestep all these problems.

As fixing the behaviour of SCLF on Linux in this regard is not practical in the short term (we don't want to diverge our CF code further from Darwin's), this PR just disables the currently failing tests.  We could go further and remove/disable all the date style tests but I leave that ~~argument~~discussion for another day.

### DIFF
--- a/TestFoundation/TestDateFormatter.swift
+++ b/TestFoundation/TestDateFormatter.swift
@@ -24,7 +24,7 @@ class TestDateFormatter: XCTestCase {
         return [
             ("test_BasicConstruction", test_BasicConstruction),
             ("test_dateStyleShort",    test_dateStyleShort),
-            ("test_dateStyleMedium",   test_dateStyleMedium),
+            //("test_dateStyleMedium",   test_dateStyleMedium),
             ("test_dateStyleLong",     test_dateStyleLong),
             ("test_dateStyleFull",     test_dateStyleFull),
             ("test_customDateFormat", test_customDateFormat),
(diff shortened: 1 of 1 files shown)


ianpartridge:
@swift-ci please test

alblue:
If the tests fail only on Darwin, can we have an `if os` that allows us to conditionally execute these changes solely for the Linux platform, but not have them cause additional noise when they run on Darwin?

ianpartridge:
I'm not sure what the purpose would be of running them on Linux. In an ideal world they would be failing there too, but they don't because of the ICU differences I mentioned.

parkera:
Regarding noticing if they are different on different platforms: I agree that is interesting, but perhaps we should also consider that they're even different between different versions of iOS or macOS. So it's not necessarily the end of the world if they are not exactly the same.

That said - I would like consistent behavior. But, the noise caused by failing tests when attempting to get real stuff done (bigger upgrades) is costly.

ianpartridge:
Thanks @parkera - I'll merge this as a step along the road...

PR #1345, opened 2017-11-29
DateComponents: Fix Overlapping accesses to 'self' warnings.

spevans:


spevans:
@swift-ci please test

Foundation/DateComponents.swift
```diff
@@ -217,8 +259,9 @@ public struct DateComponents : ReferenceConvertible, Hashable, Equatable, _Mutab
     ///
     /// The calendar and timeZone and isLeapMonth properties cannot be set by this method.
     public mutating func setValue(_ value: Int?, for component: Calendar.Component) {
+        let _value = _setter(value)
```
ianpartridge: Was it intentional that this is `_value` instead of `value`? All the others were `value`.

parkera:
I believe this code is shared with the overlay.

ianpartridge:
@spevans can you open a PR against https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/DateComponents.swift too?

PR #1347, opened 2017-11-29
Port JSON key strategy to swift-corelibs-foundation

parkera:
This is a work in progress branch.

First step is to bring JSONEncoder.swift up to date with changes to the file in the overlay before the key strategy changes landed.

### DIFF
--- a/Foundation/JSONEncoder.swift
+++ b/Foundation/JSONEncoder.swift
@@ -66,7 +66,7 @@ open class JSONEncoder {
     public enum DataEncodingStrategy {
         /// Defer to `Data` for choosing an encoding.
         case deferredToData
-        
+
         /// Encoded the `Data` as a Base64-encoded string. This is the default strategy.
         case base64
 
@@ -131,13 +131,11 @@ open class JSONEncoder {
     /// - throws: An error if any value throws an error during encoding.
     open func encode<T : Encodable>(_ value: T) throws -> Data {
         let encoder = _JSONEncoder(options: self.options)
-        try value.encode(to: encoder)
 
-        guard encoder.storage.count > 0 else {
+        guard let topLevel = try encoder.box_(value) else {
             throw EncodingError.invalidValue(value, EncodingError.Context(codingPath: [], debugDescription: "Top-level \(T.self) did not encode any values."))
         }
 
-        let topLevel = encoder.storage.popContainer()
         if topLevel is NSNull {
             throw EncodingError.invalidValue(value, EncodingError.Context(codingPath: [], debugDescription: "Top-level \(T.self) encoded as null JSON fragment."))
         } else if topLevel is NSNumber {
@@ -147,7 +145,11 @@ open class JSONEncoder {
         }
 
         let writingOptions = JSONSerialization.WritingOptions(rawValue: self.outputFormatting.rawValue)
-        return try JSONSerialization.data(withJSONObject: topLevel, options: writingOptions)
+        do {
+           return try JSONSerialization.data(withJSONObject: topLevel, options: writingOptions)
+        } catch {
+            throw EncodingError.invalidValue(value, EncodingError.Context(codingPath: [], debugDescription: "Unable to encode the given top-level value to JSON.", underlyingError: error))
+        }
     }
 }
 
@@ -177,19 +179,6 @@ fileprivate class _JSONEncoder : Encoder {
         self.options = options
         self.storage = _JSONEncodingStorage()
         self.codingPath = codingPath
-    }
-
-    // MARK: - Coding Path Operations
-
-    /// Performs the given closure with the given key pushed onto the end of the current coding path.
-    ///
-    /// - parameter key: The key to push. May be nil for unkeyed containers.
-    /// - parameter work: The work to perform with the key in the path.
-    fileprivate func with<T>(pushedKey key: CodingKey, _ work: () throws -> T) rethrows -> T {
-        self.codingPath.append(key)
-        let ret: T = try work()
-        self.codingPath.removeLast()
-        return ret
     }
 
     /// Returns whether a new element can be encoded at this coding path.
@@ -313,19 +302,6 @@ fileprivate struct _JSONKeyedEncodingContainer<K : CodingKey> : KeyedEncodingCon
         self.container = container
     }
 
-    // MARK: - Coding Path Operations
-
-    /// Performs the given closure with the given key pushed onto the end of the current coding path.
-    ///
-    /// - parameter key: The key to push. May be nil for unkeyed containers.
-    /// - parameter work: The work to perform with the key in the path.
-    fileprivate mutating func with<T>(pushedKey key: CodingKey, _ work: () throws -> T) rethrows -> T {
-        self.codingPath.append(key)
-        let ret: T = try work()
-        self.codingPath.removeLast()
-        return ret
-    }
-
     // MARK: - KeyedEncodingContainerProtocol Methods
 
     public mutating func encodeNil(forKey key: Key)               throws { self.container[key.stringValue._bridgeToObjectiveC()] = NSNull() }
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci please test

parkera:
@itaiferber I put two sections of code in an `#if` block that look to me like they are intentionally different, at least for the time being.

ianpartridge:
Nice. I wonder if we could use the `DEPLOYMENT_RUNTIME_SWIFT` define for the differences, then share an identical file in the overlay?

parkera:
Yah that may be possible. @itaiferber and I believe we have a few errors lurking in here since there are some `as` casts. Maybe we can wrap those.

Foundation/JSONEncoder.swift
```diff
@@ -344,41 +320,42 @@ fileprivate struct _JSONKeyedEncodingContainer<K : CodingKey> : KeyedEncodingCon
...
-            self.container[key.stringValue._bridgeToObjectiveC()] = try self.encoder.box(value)
-        }
+        self.encoder.codingPath.append(key)
+        defer { self.encoder.codingPath.removeLast() }
+        self.container[key.stringValue] = try self.encoder.box(value)
     }
 
     public mutating func nestedContainer<NestedKey>(keyedBy keyType: NestedKey.Type, forKey key: Key) -> KeyedEncodingContainer<NestedKey> {
         let dictionary = NSMutableDictionary()
-        self.container[key.stringValue._bridgeToObjectiveC()] = dictionary
```
itaiferber: We should keep `_bridgeToObjectiveC()` here

Foundation/JSONEncoder.swift
```diff
@@ -344,41 +320,42 @@ fileprivate struct _JSONKeyedEncodingContainer<K : CodingKey> : KeyedEncodingCon
...
+        self.codingPath.append(key)
+        defer { self.codingPath.removeLast() }
+
+        let container = _JSONKeyedEncodingContainer<NestedKey>(referencing: self.encoder, codingPath: self.codingPath, wrapping: dictionary)
+        return KeyedEncodingContainer(container)
     }
 
     public mutating func nestedUnkeyedContainer(forKey key: Key) -> UnkeyedEncodingContainer {
         let array = NSMutableArray()
-        self.container[key.stringValue._bridgeToObjectiveC()] = array
```
itaiferber: Same

Foundation/JSONEncoder.swift
```diff
@@ -698,13 +661,18 @@ extension _JSONEncoder {
     }
 
     fileprivate func box<T : Encodable>(_ value: T) throws -> NSObject {
-        if T.self == Date.self {
+        return try self.box_(value) ?? NSDictionary()
+    }
+
+    // This method is called "box_" instead of "box" to disambiguate it from the overloads. Because the return type here is different from all of the "box" overloads (and is more general), any "box" calls in here would call back into "box" recursively instead of calling the appropriate overload, which is not what we want.
+    fileprivate func box_<T : Encodable>(_ value: T) throws -> NSObject? {
+        if T.self == Date.self || T.self == NSDate.self {
```
itaiferber: All of these casts will fail on Linux because of bridging. We can keep the behavior by turning all of the NS cases into actual conversions

Foundation/JSONEncoder.swift
```diff
@@ -1750,6 +1750,25 @@ extension _JSONDecoder {
     fileprivate func unbox(_ value: Any, as type: Bool.Type) throws -> Bool? {
         guard !(value is NSNull) else { return nil }
 
+#if false
```
itaiferber: The `#if false`es are good candidates for converting to `DEPLOYMENT_TARGET_SWIFT`

Foundation/JSONEncoder.swift
```diff
@@ -2081,21 +2101,32 @@ extension _JSONDecoder {
...
         // FIXME: Attempt to grab a Decimal value if JSONSerialization on Linux produces one.
         let doubleValue = try self.unbox(value, as: Double.self)!
         return Decimal(doubleValue)
+#endif
     }
 
     fileprivate func unbox<T : Decodable>(_ value: Any, as type: T.Type) throws -> T? {
         let decoded: T
-        if T.self == Date.self {
+        if T.self == Date.self || T.self == NSDate.self {
```
itaiferber: Same here — all of the `|| T.self == ...` lead to `as!`-casts which are not valid on Linux

parkera:
@swift-ci smoke test

parkera:
@swift-ci please test

parkera:
@swift-ci please test

itaiferber:
@swift-ci Please test

parkera:
@swift-ci please test

alblue:
Looks like this failed with a core dump:

```
Test Case 'TestJSONEncoder.test_encodingTopLevelSingleValueEnum' started at 2017-11-30 00:49:44.323
/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift/utils/build-script-impl: line 264: 26232 Illegal instruction     (core dumped) "$@"
/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift/utils/build-script: fatal error: command terminated with a non-zero exit status 132, aborting
```

parkera:
Fixed the core dump (we needed Codable.swift changes too), but have some bridging failures to figure now now.

alblue:
@swift-ci please test

parkera:
I already know this fails. Just need some time to come back to this and sort out the bridging issues.

Andrewangeta:
@parkera Any updates on this? I'm hoping this makes into 4.1 keyDecodingStragety is a really nice API and cleans up so much code.

parkera:
Unfortunately I got stuck on this one, having trouble figuring out why the tests don't pass after merging in the prerequisite changes from the Swift overlay version.

Lukasa:
So in case it helps, this seems to be a problem initially isolated to the decoder. Here's basically the smallest repro I can give you:

```swift
import Foundation

struct Test: Codable {
    var field: Int
}

let json = "{\"field\": 5}".data(using: .utf8)!
let decoder = JSONDecoder()
let result = try decoder.decode(Test.self, from: json)
print(result.field)
```

In Swift 4.0, this works fine and prints 5, as expected. In Swift 4.1, it explodes with

```
typeMismatch(Swift.Int, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "field", intValue: nil)], debugDescription: "Expected to decode Int but found Int instead.", underlyingError: nil))
```

This seems to be related to the underlying parsing producing an `Int` instead of an `NSNumber`. As identified before, this is likely a bridging issue, and of the two of us I suspect you're vastly better equipped to go further with this than I am @parkera.

parkera:
@itaiferber I know we are still talking about the returned types from `JSONSerialization` in a few other PRs, but I don't recall any changes here for 4.1 specifically. Do you?

itaiferber:
None that I've seen; I can't find any PRs that changed `JSONSerialization` in Swift 4.1 on here either (but I'll keep looking). @Lukasa What platform are you on? Which toolchain are you running there?

Lukasa:
@itaiferber Sorry, I could really have made that clearer. That's Ubuntu 16.04, using a hand-compiled Swift toolchain from the `swift-4.1-DEVELOPMENT-SNAPSHOT-2018-03-21-a` tag, with a Foundation branched from that tag with *this patch* merged on top of it.

If I just use the `swift-4.1-DEVELOPMENT-SNAPSHOT-2018-03-21-a` tag directly **without this patch**, the above test passes just fine. So it's this patch that is affecting the return type.

parkera:
Ah, ok.

Andrewangeta:
Any updates on this issue? @parkera 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gperdomor:
Any update on this?

millenomi:
We are dealing with some bridging landing fallout and WWDC. My plan is to look at this starting late this week, unless someone else can.

gperdomor:
We can expect this merged in swift 4.2? :(

alblue:
@parkera can you rebase this so that we can run the tests again, or close it if it has gone stale?

parkera:
I'm going to close this, since I think #1561 should cover it.

PR #1348, opened 2017-11-29
[WIP] Improve consistency in NSNumber bridging

natecook1000:
Resolves [SR-4634](https://bugs.swift.org/browse/SR-4634) and [SR-6322](https://bugs.swift.org/browse/SR-6322).

### DIFF
--- a/Foundation/NSNumber.swift
+++ b/Foundation/NSNumber.swift
@@ -423,9 +423,17 @@ extension Float : _ObjectTypeBridgeable {
     }
 
     public init?(exactly number: NSNumber) {
-        guard let value = Double(exactly: number) else { return nil }
-        guard let result = Float(exactly: value) else { return nil }
-        self = result
+        let type = number.objCType.pointee
+        if type == 0x49 || type == 0x4c || type == 0x51 {
+            guard let result = Float(exactly: number.uint64Value) else { return nil }
+            self = result
+        } else if type == 0x69 || type == 0x6c || type == 0x71 {
+            guard let result = Float(exactly: number.int64Value) else { return nil }
+            self = result
+        } else {
+            guard let result = Float(exactly: number.doubleValue) else { return nil }
+            self = result
+        }
     }
 
     public func _bridgeToObjectiveC() -> NSNumber {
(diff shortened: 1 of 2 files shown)


natecook1000:
@swift-ci Please test

TestFoundation/TestNSNumberBridging.swift
```diff
@@ -515,15 +483,22 @@ class TestNSNumberBridging : XCTestCase {
...
+                    expectNil(expectedFloat)
+                } else {
+                    testFloat(expectedFloat, float)
+                }
+              
+                let double = Double(exactly: number!)
+                let expectedDouble = Double(exactly: interestingValue)
+                if interestingValue.isNaN {
+                    expectTrue(double?.isNaN == true)
+                    expectNil(expectedDouble)
```
parkera: We have the XCTest functions here instead of the standard library ones.

ianpartridge:
Thank you for working on this! cc: @spevans who I think has also worked in this area.

Foundation/NSNumber.swift
```diff
@@ -423,9 +423,17 @@ extension Float : _ObjectTypeBridgeable {
     }
 
     public init?(exactly number: NSNumber) {
-        guard let value = Double(exactly: number) else { return nil }
-        guard let result = Float(exactly: value) else { return nil }
-        self = result
+        let type = number.objCType.pointee
+        if type == 0x49 || type == 0x4c || type == 0x51 {
+            guard let result = Float(exactly: number.uint64Value) else { return nil }
```
spevans: Can `_cfNumberType()` be used to use names instead of magic numbers?

TestFoundation/TestNSNumberBridging.swift
natecook1000: Thanks—had tried to type-check this in my head 😳

Foundation/NSNumber.swift
natecook1000: It looks like `_cfNumberType()` doesn't distinguish between signed and unsigned integer types, which could lead to overflow with large unsigned integers.

natecook1000:
@swift-ci Please test

ianpartridge:
@swift-ci please test

natecook1000:
@swift-ci Please test

Foundation/NSNumber.swift
phausler: The cf storage always stores signed types so using the base constants are fine

phausler: With the caveat that you need to up-cast appropriately: but those values are fine as-is

natecook1000: How do you handle values that are out of range for `Int64`?

```swift
let x = (UInt64(Int64.max) + 1) as NSNumber
// x.uint64Value == 9223372036854775808
// x.int64Value == -9223372036854775808
```

alblue:
@swift-ci please test

Foundation/NSNumber.swift
phausler: the out of range values are handled via kCFNumberSInt128Type (using CFSInt128Struct)

natecook1000: Wow, TIL… I think that would even bitcast to a `DoubleWidth<Int64>`.

PR #1349, opened 2017-11-30
[gardening][Measurement] Move operators into types

ikesyo:
https://github.com/apple/swift-corelibs-foundation/pull/1349/files?w=1

### DIFF
--- a/Foundation/Measurement.swift
+++ b/Foundation/Measurement.swift
@@ -72,176 +72,204 @@ extension Measurement where UnitType : Dimension {
     
 }
 
-/// Add two measurements of the same Unit.
-/// - precondition: The `unit` of `lhs` and `rhs` must be `isEqual`.
-/// - returns: A measurement of value `lhs.value + rhs.value` and unit `lhs.unit`.
-public func +<UnitType>(lhs: Measurement<UnitType>, rhs: Measurement<UnitType>) -> Measurement<UnitType> {
-    if lhs.unit.isEqual(rhs.unit) {
-        return Measurement(value: lhs.value + rhs.value, unit: lhs.unit)
-    } else {
-        fatalError("Attempt to add measurements with non-equal units")
+extension Measurement {
+    /// Add two measurements of the same Unit.
+    /// - precondition: The `unit` of `lhs` and `rhs` must be `isEqual`.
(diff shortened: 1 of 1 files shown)


ikesyo:
@swift-ci Please test

Foundation/Measurement.swift
```diff
@@ -72,176 +72,204 @@ extension Measurement where UnitType : Dimension {
...
-/// If the `unit` of the `lhs` and `rhs` are `==`, then this returns the result of subtracting the `value` of each `Measurement`. If they are not equal, then this will convert both to the base unit of the `Dimension` and return the result as a `Measurement` of that base unit.
-/// - returns: The result of adding the two measurements.
-public func -<UnitType : Dimension>(lhs: Measurement<UnitType>, rhs: Measurement<UnitType>) -> Measurement<UnitType> {
-    if lhs.unit == rhs.unit {
-        return Measurement(value: lhs.value - rhs.value, unit: lhs.unit)
-    } else {
-        let lhsValueInTermsOfBase = lhs.unit.converter.baseUnitValue(fromValue: lhs.value)
-        let rhsValueInTermsOfBase = rhs.unit.converter.baseUnitValue(fromValue: rhs.value)
-        return Measurement(value: lhsValueInTermsOfBase - rhsValueInTermsOfBase, unit: type(of: lhs.unit).baseUnit())
+extension Measurement where UnitType : Dimension {
```
ianpartridge: I'm not sure that having a lot of new `extension`s is ideal. Could we move this inside the existing extension on line 47?

spevans: Every single method seems to be in its own extension although some are using a conditional conformance so this is consistent. However maybe a followup PR could merge them so there are only two extensions.

ikesyo: > Every single method seems to be in its own extension although some are using a conditional conformance so this is consistent. However maybe a followup PR could merge them so there are only two extensions.

Yep I'm going to do that.

ianpartridge: OK as long as this PR isn't the end then 👍

ianpartridge:

ianpartridge:

ikesyo:
Thanks @ianpartridge and @spevans!

parkera:
I believe this code is also shared with the overlay.

PR #1351, opened 2017-12-01
[SR-6398] Ensure CFSTR() is CFRetained because it is not permanent on Linux.

parkera:


parkera:
@swift-ci please test

jrose-apple:
Is there a reason not to put that into the CFSTR macro itself?

parkera:
CFSTR has no ability to determine how many times the result will be released.

parkera:
The real fix for this is the ability to mark Swift objects as immortal and have swift_release short-circuit.

PR #1352, opened 2017-12-03
Reduce test compile times.

parkera:
I noticed that the tests were taking quite a while to compile.

Using the "-warn-long-expression-type-checking" flag, I measured the compile time of the test project. The total of types over 100ms was around 27 seconds before this change and around 25 seconds after (on my build configuration), so it shaves off a few seconds there.

I also took the opportunity to reduce the number of attempted threads in the URL test from 640 to 10. There isn't much value add in going above the number of cores we're going to find on most machines, and that dramatically reduces the run time of that test.

The tests still take quite a while to build, so I'd like to continue this investigation. But this is a reasonable start.

### DIFF
--- a/TestFoundation/TestDecimal.swift
+++ b/TestFoundation/TestDecimal.swift
@@ -180,13 +180,15 @@ class TestDecimal: XCTestCase {
     }
 
     func test_ExplicitConstruction() {
+        let reserved: UInt32 = (1<<18 as UInt32) + (1<<17 as UInt32) + 1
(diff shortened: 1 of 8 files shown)


parkera:
@swift-ci please test

spevans:
@parkera Does using `-whole-module-optimization` make any difference?  I don't think it is used for either the library or the tests at the moment.

parkera:
Do you expect whole module optimization to reduce compile times? I've always thought it was mostly related to run time performance optimization.

ianpartridge:
Do you know whether the type-checker folks are aware of these long compile times?  SCLF might be an interested testcase for them to investigate.

alblue:

PR #1355, opened 2017-12-05
[URLSession] Refactor common protocol code into a new class

saiHemak:
The new class called `NativeProtocol` now contains a protocol-agnostic implementation which is currently used by HTTPURLProtocol, and in future will be used by other native protocols -  FTP, Data etc.

TransferState has 1-on-1 relationship with EasyHandle and is independent of protocol.Hence it has been moved out of the ```http``` folder.

`HTTPCharacters` has been renamed to ```Delimiters```  which has  characters or delimiters  like CRLF which is common across the protocols.
eg: FTP headers and HTTP headers end with CRLF characters.

Common implementation  of header parsing has been moved to  new file ```Message.swift``` and the code specific to ```HTTP``` has been retained in HTTPMessage.swift.


pushkarnk:
Looks like there's some problem here?

alblue:
They were going to take the systems down to install new versions of Xcode - there was a mail out about it. We should try kicking it off again later.

alblue:
@swift-ci please test

saiHemak:
Splitting the commits to improve readability of each commit.

pushkarnk:
@swift-ci please test

pushkarnk:
@swift-ci please test

ianpartridge:
I agree with Al.

saiHemak:
@alblue  Thanks for the review .The code under TransferState.swift and Message.swift are extensions of _NativeProtocol class.So to compile TransferState.swift and Message.swift individually _NativeProtocol class should be in place.Hence I had to raise the PR with movement and refactoring together.
However, to address your comments I am trying to compile with a skeleton of _NativeProtocol class in place and perform the movement under one PR ..

PR #1359, opened 2017-12-07
Add a testcase for HTTPCookieStorage.description

vipinmenon:
Per the guidance in https://github.com/apple/swift-corelibs-foundation/pull/1313 this PR splits out the test for `HTTPCookieStorage.description`

cc @ianpartridge

### DIFF
--- a/TestFoundation/TestHTTPCookieStorage.swift
+++ b/TestFoundation/TestHTTPCookieStorage.swift
@@ -32,6 +32,7 @@ class TestHTTPCookieStorage: XCTestCase {
             ("test_cookiesForURL", test_cookiesForURL),
             ("test_cookiesForURLWithMainDocumentURL", test_cookiesForURLWithMainDocumentURL),
             ("test_cookieInXDGSpecPath", test_cookieInXDGSpecPath),
+            ("test_descriptionCookie", test_descriptionCookie),
         ]
     }
 
@@ -90,6 +91,11 @@ class TestHTTPCookieStorage: XCTestCase {
         setCookiesForURLWithMainDocumentURL(with: .groupContainer("test"))
     }
 
+    func test_descriptionCookie() {
+        descriptionCookie(with: .shared)
+        descriptionCookie(with: .groupContainer("test"))
+    }
+
     func getStorage(for type: _StorageType) -> HTTPCookieStorage {
         switch type {
         case .shared:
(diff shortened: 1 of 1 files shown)


TestFoundation/TestHTTPCookieStorage.swift
```diff
@@ -32,6 +32,7 @@ class TestHTTPCookieStorage: XCTestCase {
             ("test_cookiesForURL", test_cookiesForURL),
             ("test_cookiesForURLWithMainDocumentURL", test_cookiesForURLWithMainDocumentURL),
             ("test_cookieInXDGSpecPath", test_cookieInXDGSpecPath),
+            ("test_descriptionCookie", test_descriptionCookie)
```
alblue: This needs a comma at the end, so it doesn't cause unnecessary noise later

ianpartridge:
Please remove the `XCTAssertEqual(storage.description, ...)` lines from other tests now that we are testing the description in its own test.

TestFoundation/TestHTTPCookieStorage.swift
```diff
@@ -246,6 +252,23 @@ class TestHTTPCookieStorage: XCTestCase {
         XCTAssertEqual(storage.cookies(for: url1!)!.count, 0)
     }
 
+    func descriptionCookie(with storageType: _StorageType) {
```
ianpartridge: As this function is only used inside `test_descriptionCookie()` why not make it a nested function?

TestFoundation/TestHTTPCookieStorage.swift
```diff
@@ -246,6 +252,23 @@ class TestHTTPCookieStorage: XCTestCase {
...
+        let storage = getStorage(for: storageType)
+        XCTAssertEqual(storage.description, "<NSHTTPCookieStorage cookies count:\(storage.cookies!.count)>")
```
ianpartridge: I know this is done elsewhere in the tests but I don't like force unwrapping. If it fails, the whole testsuite blows up.  It's better to do:

```swift
guard let cookies = storage.cookies else {
    XCTFail("No cookies")
    return
}
etc...
```

TestFoundation/TestHTTPCookieStorage.swift
vipinmenon: @ianpartridge I see that this pattern (calling a function XXX two times from test_XXX) is used in all the tests in TestHTTPCookieStorage. Do you think doing this as a separate PR will make more sense?

vipinmenon:
@ianpartridge @alblue Thanks for the review. I have responded to the comments. Kindly approve.

pushkarnk:
@swift-ci please test

pushkarnk:
I see that all of the comments have been addressed. 

PR #1362, opened 2017-12-07
Restore .concurrent for block-based enumerateObjects(…) methods in NSArray

millenomi:
- These methods are implemented as (relevant NSIndexSet).enumerate(options: options…), so the fix goes in NSIndexSet.

 - Includes NSArray tests.


millenomi:
@swift-ci please test.

PR #1365, opened 2017-12-08
Add initial implementation of URLSession.shared

pushkarnk:
Basic users of URLSession may want to use the shared URLSession instance`URLSession.shared`. 

The shared URLSession instance uses shared instances of HTTPCookieStorage, URLCache and URLCredentialStorage, the last two still remain unimplemented. It is likely that some of the users that want to use `URLSession.shared` aren't interested in using the shared storages and cache! So, we need not wait for URLCache and URLCredentialStorage to be fully available here (we can plug them in later, see the TODOS). In that way, we are offering `URLSession.shared` to at least those users!

### DIFF
--- a/Foundation/URLSession/URLSession.swift
+++ b/Foundation/URLSession/URLSession.swift
@@ -201,8 +201,19 @@ open class URLSession : NSObject {
      * The shared session uses the currently set global URLCache,
      * HTTPCookieStorage and URLCredential.Storage objects.
      */
-    open class var shared: URLSession { NSUnimplemented() }
-    
+    open class var shared: URLSession {
+        return _shared
+    }
+
+    fileprivate static let _shared: URLSession = {
+        var configuration = URLSessionConfiguration.default
+        configuration.httpCookieStorage = HTTPCookieStorage.shared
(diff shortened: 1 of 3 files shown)


Foundation/URLSession/URLSession.swift
```diff
@@ -201,8 +201,19 @@ open class URLSession : NSObject {
      * The shared session uses the currently set global URLCache,
      * HTTPCookieStorage and URLCredential.Storage objects.
      */
-    open class var shared: URLSession { NSUnimplemented() }
-    
+    open class var shared: URLSession {
+        return _shared
+    }
+
+    fileprivate static let _shared: URLSession = {
```
ianpartridge: Lazy?

millenomi: IIRC, all global stored variables (including all static stored variables) that are initialized with a block call are initialized lazily. `lazy` cannot be applied to them.

pushkarnk: Yeah, static properties are implicitly lazy.

pushkarnk:
@swift-ci please test

ianpartridge:

PR #1366, opened 2017-12-08
Restore the .concurrent option for NSArray's enumerateObjects(…)

millenomi:
(Let's try this again. The previous attempt had an issue in the test for enumerateObjects(at:options:using:) where it could've set the stop flag to true before all indexes corresponding to our expectation had been set to true, which could lead it to inconsistently failing.)

NSArray implements these methods as: `someIndexSet.enumerate(options: options) { block(self[$0]…) }`, so the fix goes in NSIndexSet.

Includes NSArray tests.


millenomi:
@swift-ci please test.

PR #1367, opened 2017-12-09
Implement copying, indexing and enumeration for NSOrderedSet using NSMutableArray/NSMutableSet

amorde:
This is a follow up to #385

Were this to be merged, the only things left unimplemented would be `description(withLocale:)` and `description(withLocale:indent:)`, excluding the sort options that are not yet implemented by `NSArray`

### DIFF
--- a/Foundation/NSOrderedSet.swift
+++ b/Foundation/NSOrderedSet.swift
@@ -9,15 +9,19 @@
 
 /****************       Immutable Ordered Set   ****************/
 open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, ExpressibleByArrayLiteral {
-    internal var _storage: Set<NSObject>
-    internal var _orderedStorage: [NSObject]
+    internal var _storage: NSMutableSet
+    internal var _orderedStorage: NSMutableArray
     
     open override func copy() -> Any {
         return copy(with: nil)
     }
     
     open func copy(with zone: NSZone? = nil) -> Any {
-        NSUnimplemented()
+        if type(of: self) == NSOrderedSet.self || type(of: self) == NSMutableOrderedSet.self {
+            // Return self for immutable type
+            return self
+        }
+        return NSOrderedSet(array: allObjects)
     }
 
     open override func mutableCopy() -> Any {
@@ -25,7 +29,17 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
     }
 
     open func mutableCopy(with zone: NSZone? = nil) -> Any {
-        NSUnimplemented()
+        let setType = type(of: self)
+        if setType == NSOrderedSet.self || setType == NSMutableOrderedSet.self {
+            let mutableSelf = NSMutableOrderedSet()
+            mutableSelf._storage = _storage.mutableCopy() as! NSMutableSet
+            mutableSelf._orderedStorage = _orderedStorage.mutableCopy() as! NSMutableArray
+            return mutableSelf
+        } else {
+            let mutableSelf = NSMutableOrderedSet(capacity: count)
+            mutableSelf.addObjects(from: self.allObjects)
+            return mutableSelf
+        }
     }
     
     public static var supportsSecureCoding: Bool {
@@ -46,10 +60,11 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
         }
     }
     
-    public required convenience init?(coder aDecoder: NSCoder) {
+    public required init?(coder aDecoder: NSCoder) {
         guard aDecoder.allowsKeyedCoding else {
             preconditionFailure("Unkeyed coding is unsupported.")
         }
+        
         var idx = 0
         var objects : [AnyObject] = []
         while aDecoder.containsValue(forKey: ("NS.object.\(idx)")) {
@@ -59,7 +74,13 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
             objects.append(object as! NSObject)
             idx += 1
         }
-        self.init(array: objects)
+        
+        _orderedStorage = NSMutableArray(array: objects)
+        _storage = NSMutableSet(array: objects)
+        
+        guard _orderedStorage.count == _storage.count else {
+            return nil
+        }
     }
     
     open var count: Int {
@@ -67,20 +88,26 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
     }
 
     open func object(at idx: Int) -> Any {
-        return _SwiftValue.fetch(nonOptional: _orderedStorage[idx])
+        return _orderedStorage.object(at: idx)
     }
 
     open func index(of object: Any) -> Int {
-        return _orderedStorage.index(of: _SwiftValue.store(object)) ?? NSNotFound
+        return _orderedStorage.index(of: object)
     }
 
     public convenience override init() {
-        self.init(objects: [], count: 0)
+        self.init(objects: nil, count: 0)
     }
 
-    public init(objects: UnsafePointer<AnyObject>!, count cnt: Int) {
-        _storage = Set<NSObject>()
-        _orderedStorage = [NSObject]()
+    public convenience init(objects: UnsafePointer<AnyObject>, count cnt: Int) {
+        self.init()
+
+        _insertObjects(objects, count: cnt)
+    }
+
+    public init(objects: UnsafePointer<AnyObject>?, count cnt: Int) {
+        _storage = NSMutableSet()
+        _orderedStorage = NSMutableArray()
 
         super.init()
 
@@ -88,11 +115,11 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
     }
     
     required public convenience init(arrayLiteral elements: Any...) {
-      self.init(array: elements)
+        self.init(array: elements)
     }
 
     public convenience init(objects elements: Any...) {
-      self.init(array: elements)
+        self.init(array: elements)
     }
     
     open subscript (idx: Int) -> Any {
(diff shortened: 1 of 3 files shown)


Foundation/NSOrderedSet.swift
```diff
@@ -46,67 +49,67 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
...
-        return _orderedStorage.index(of: _SwiftValue.store(object)) ?? NSNotFound
+        return _orderedStorage.index(of: object)
     }
 
     public convenience override init() {
         self.init(objects: [], count: 0)
     }
 
     public init(objects: UnsafePointer<AnyObject>!, count cnt: Int) {
-        _storage = Set<NSObject>()
```
spevans: According to https://developer.apple.com/documentation/foundation/nsorderedset the two init functions that take an `UnsafePointer` are:
`init(objects: UnsafePointer<AnyObject>?, count: Int)` and
`init(objects: UnsafePointer<AnyObject>, count: Int)`. This one needs to be fixed to match one of those, and also could you possibly add the other one?

spevans:
@swift-ci please test

Foundation/NSOrderedSet.swift
amorde: Absolutely - good catch

Foundation/NSOrderedSet.swift
```diff
@@ -46,85 +49,79 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
...
     }
 
     public convenience override init() {
         self.init(objects: [], count: 0)
     }
 
-    public init(objects: UnsafePointer<AnyObject>!, count cnt: Int) {
-        _storage = Set<NSObject>()
-        _orderedStorage = [NSObject]()
+    public convenience init(objects: UnsafePointer<AnyObject>, count cnt: Int) {
```
amorde: Inspecting the init in Xcode showed this specific variant as a convenience init while the one below was not, so I matched the behavior here as well

spevans: If you follow the link for any of the methods in https://developer.apple.com/documentation/foundation/nsorderedset you will see the full function signature. It would be a good  idea to do a quick check of all of the methods and bring them into line with the Foundation documentation.

spevans:
From https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/486/console it looks like there was a test failure:

`TestFoundation/TestNSOrderedSet.swift:177: error: TestNSOrderedSet.test_RemoveObject : XCTAssertEqual failed: ("3") is not equal to ("2") - `

Foundation/NSOrderedSet.swift
amorde: Yep, I was checking this as I implemented them but missed this one. I will check again

amorde: Found a couple more that need to be optional instead of implicitly unwrapped

spevans:
@swift-ci please test

parkera:
Do we need any additional tests for this functionality?

Foundation/NSOrderedSet.swift
```diff
@@ -9,23 +9,26 @@
...
 
     open override func mutableCopy() -> Any {
         return mutableCopy(with: nil)
     }
 
     open func mutableCopy(with zone: NSZone? = nil) -> Any {
-        NSUnimplemented()
+        let mutableSelf = NSMutableOrderedSet(capacity: count)
+        mutableSelf.addObjects(from: self.array)
+        return mutableSelf
```
michael-lehew: this is sufficient, but doesn't take advantage of the fact that we already know the elements are a set, so everything re-hashes, etc.

Foundation/NSOrderedSet.swift
```diff
@@ -46,85 +49,81 @@ open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
...
         while aDecoder.containsValue(forKey: ("NS.object.\(idx)")) {
             guard let object = aDecoder.decodeObject(forKey: "NS.object.\(idx)") else {
                 return nil
             }
-            objects.append(object as! NSObject)
+            _orderedStorage.add(object)
+            _storage.add(object)
             idx += 1
         }
-        self.init(array: objects)
```
michael-lehew: the previous implementation is a little bit more correct I think.  You don't want to have multiple initializers for coding. Coding should use the regular initializer as much as possible -- philosophically:  in doing so its easier to maintain the run time invariants of your object, since they're automatically enforced. 

Also from a correctness perspective -- its possible to construct a payload that will not actually come from a valid runtime ordered set (e.g. contain duplicates) and this does not detect such a malformed archive.  Best approach there would be just create the ordered set and ask if its count == count of the array, returning nil if not equal.

Foundation/NSOrderedSet.swift
```diff
@@ -9,23 +9,26 @@
 
 /****************       Immutable Ordered Set   ****************/
 open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, ExpressibleByArrayLiteral {
-    internal var _storage: Set<NSObject>
-    internal var _orderedStorage: [NSObject]
+    internal var _storage: NSMutableSet
+    internal var _orderedStorage: NSMutableArray
```
michael-lehew: this is essentially how the ObjC-based Darwin implementations are handled, so thats "A Nice Thing™"

Foundation/NSOrderedSet.swift
```diff
@@ -492,39 +493,33 @@ extension NSMutableOrderedSet {
...
+        minusSet(other.set)
     }
 
     open func union(_ other: NSOrderedSet) {
         other.forEach(add)
     }
     
     open func intersectSet(_ other: Set<AnyHashable>) {
-        for case let item as AnyHashable in self where !other.contains(item) {
+        for case let item as AnyHashable in self.reversed() where !other.contains(item) {
```
michael-lehew: i assume you're doing this for performance reasons?    there are still O(k) 'shifts'  of the array to fill the holes from the removals.  if you're going for clarity i'd remove the .reversed()  if you're going for clarity combine the shift with a single O(n) pass.   Something like:

```
// pseudo code
i = 0
while i < count{
    nextIndex = i + 1
   if !other.contains(_ordereStorage[i]) {
    if (i+1 < count) {
    _orderedStorage[i] = _orderedStorage[i + 1]
   } else {
      _orderedStorage.removeLastObject()
   }
   // remove from the set too
  } else {
    i = i + 1 // move onto the next index
  }
}

```

Foundation/NSPredicate.swift
```diff
@@ -196,9 +196,9 @@ extension NSMutableSet {
 
 extension NSOrderedSet {
     open func filtered(using predicate: NSPredicate) -> NSOrderedSet {
-        return NSOrderedSet(array: self.allObjects.filter({ object in
+        return NSOrderedSet(array: self.array.filter { object in
```
michael-lehew: why go through the snapshot (and thus bridged) array here?

Foundation/NSOrderedSet.swift
michael-lehew: Actually looking more closely -- I realize this is the NSOrderedSet impl -- why are these mutable?

Foundation/NSOrderedSet.swift
amorde: I looked at how `NSArray` is implemented and it copies over the `_storage` to the next object - should I use a similar mechanism here?

Foundation/NSOrderedSet.swift
amorde: Doing it without `reversed()` resulted in an index out of bounds error, my guess was because the enumerator was not keeping track of mutations to the array. I could definitely take some time to rewrite this to improve performance

Foundation/NSPredicate.swift
amorde: I took this portion from the other PR - I'll take another look for better options

Foundation/NSOrderedSet.swift
amorde: Thank you for the insight, I'll take another look and see if I can get it closer to was it was before

Foundation/NSOrderedSet.swift
amorde: Good catch - these no longer need to be vars

amorde: Although it does allow internal implementation to copy the `_storage` from one set to another, like when implementing `copy`

amorde: Sorry, I see I misunderstood your question. I could override these in `NSMutableOrderedSet` to be mutable and keep these ones as their non-mutable counter parts, but keeping these as mutable makes it easier to build the set incrementally in initializers

Foundation/NSPredicate.swift
amorde: I can't see a way around group through bridging, since the initializers to NSOrderedSet accept `[Any]`

I may be misunderstanding this though. Do you have any suggestions on a better way to do this?

amorde:
I will clean up the commits a bit once the implementation is finalized

Foundation/NSOrderedSet.swift
amorde: Here's a similar method in `NSMutableArray` for reference:

```swift
open func removeObjects(in otherArray: [Any]) {
    let set = Set(otherArray.map { $0 as! AnyHashable } )
    for idx in (0..<count).reversed() {
        if let value = object(at: idx) as? AnyHashable {
            if set.contains(value) {
                removeObject(at: idx)
            }
        }
    }
}
```

Maybe that could be improved a bit as well

spevans:
@swift-ci please test

Foundation/NSOrderedSet.swift
michael-lehew: Building incrementally isn't the best for performance (sans capacity hints that are obeyed) - since you'll hit the internal vector growth storage of array/set many times.

Foundation/NSOrderedSet.swift
michael-lehew: I think I agree that array could be improved as well.  :) something for the future

Foundation/NSOrderedSet.swift
michael-lehew: arrays don't get to make extra assumptions about their contents -- my point here was that the its possible to avoid the set-membership check (the set will still have to hash etc to maintain correct layout) but since we're already a runtime set, we already have satisfied uniqueness.  its probably a small win -- but for common currency types, small wins matter a bit more.

Foundation/NSPredicate.swift
michael-lehew: I was thinking just using the ivar -- but I see that this in an extension and perhaps doesn't have access.   The self.array will force a bridge from the NS type to the swift type, and so there's an extra semantic copy here (its perhaps not as bad as on Darwin platforms where we're crossing the objc->swift divide).

Foundation/NSOrderedSet.swift
amorde: I have updated this to use an implementation that is based on your pseudocode - I think it could definitely be adapted to NSArray too!

Foundation/NSPredicate.swift
amorde: I added back the `allObjects` accessor, so this change is no longer present

millenomi:
@swift-ci please test.

amorde:
@michael-lehew Do you have any other comments or suggestions for this? Since the discussion is growing quite large I am considering starting from scratch with a new PR, unless you feel this isn't the right direction to go at all

alblue:
@swift-ci please test

PR #1370, opened 2017-12-14
Carving out common message parsing code out of HTTPMessage

saiHemak:
Moving out the common code from HTTPMessage  to Message.swift as it can be used with multiple protocols.

Eg: HTTP headers or FTP headers basically terminates with CRLF. However the format of headers in HTTP is different from HTTP ,the basic parsing remains same in both cases.
Hence it makes sense to carve out the common implementation which does the basic parsing to a common place.
Also, rename HTTPCharacters to Delimiters as CRLF characters are not specific to HTTP protocol.

@alblue  , I raised this ahead of TransferState.swift refactoring as TransferState.swift  depends on this PR.


pushkarnk:
@swift-ci please test

ianpartridge:
I'm sorry but again this is both moving and changing code in one commit. Please can we move the files in one commit, and make the changes in another.

saiHemak:
@ianpartridge , I am not moving files here ,Its part of code from HTTPMessage has been carved out into a  new file.

ianpartridge:
Take the example of `struct _HTTPCharacters`. This has both moved from `URLSession/http/HTTPMessage.swift` to `URLSession/Message.swift` and renamed to `struct _Delimiters`. That means git can't generate a meaningful diff.

PR #1373, opened 2017-12-18
[SR-6581] Mark FileManager.attributesOfFileSystem unavailable on Android

ruipfcosta:
Fixes [SR-6581](https://bugs.swift.org/browse/SR-6581).

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -340,10 +340,13 @@ open class FileManager : NSObject {
      
         This method replaces fileSystemAttributesAtPath:.
      */
-    open func attributesOfFileSystem(forPath path: String) throws -> [FileAttributeKey : Any] {
 #if os(Android)
+    @available(*, unavailable, message: "Not available on Android")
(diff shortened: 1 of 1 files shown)


millenomi:
@ruipfcosta Can you note in the comment that this fixes SR-6581 at a minimum?

@swift-ci Please test.

parkera:
What was the problem with Android here, specifically? It has no way to retrieve this info?

millenomi:
Android was #if'ing just this method to crash — I'm not sure what to do here, but the correct outcome would be that someone implement the Android version of this method.

In the meantime, I wanted to make this apparent to clients rather than not

millenomi:
(It would be real weird if Android couldn't do the equivalent of stat(), though, honestly.)

alblue:
Someone who is working on the Android version should probably opine on this; it's a no-op on other platforms. The question is, are there any places in the Android build that refer to this? Has it been built correctly on the Android platform with this change?

millenomi:
Looks like a similar change already went in.

PR #1378, opened 2017-12-23
 Finishing up NS[Mutable]AttributedString

eyeplum:
This pull request finishes up unimplemented methods in `NS[Mutable]AttributedString`.

### DIFF
--- a/TestFoundation/TestNSAttributedString.swift
+++ b/TestFoundation/TestNSAttributedString.swift
@@ -23,6 +23,7 @@ class TestNSAttributedString : XCTestCase {
     static var allTests: [(String, (TestNSAttributedString) -> () throws -> Void)] {
         return [
             ("test_initWithString", test_initWithString),
+            ("test_attributedSubstring", test_attributedSubstring),
             ("test_initWithStringAndAttributes", test_initWithStringAndAttributes),
             ("test_longestEffectiveRange", test_longestEffectiveRange),
             ("test_enumerateAttributeWithName", test_enumerateAttributeWithName),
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

eyeplum:
Hi @spevans, thanks for triggering the test!
I have implemented a few more methods, can you please trigger it again?
Also, Merry Christmas!! 🎄

spevans:
@swift-ci please test

millenomi:
@swift-ci Please test.

eyeplum:
Implemented more methods (since I've found out that `isEqual(to:)` `init(attributedString:)` `setAttributes` are kind of related, and can take advantage of some already implemented methods).

Now only `NSCoding` remains unimplemented, other methods should be good :)

I will investigate how to start implementing `NSCoding` for attributed string and will create a new pull request for that.

spevans:
@swift-ci please test

spevans:
@eyeplum FYI that test finished successfully https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/526/

I don't think you need to worry about supporting `NSCoding` as it isn't supported anywhere else in swift-corelibs-foundation - Im not even sure it actually can be outside of Darwin.

I had a quick look though some of the PR, could you possibly add tests for `NSMutableAttributedString(string:attributes:)` and `NSMutableAttributedString(attributedString:)`.
Its a good idea to explicitly check all of the initialisers work.

eyeplum:
@spevans Looks like the behaviour of those initializers are a bit different in mutable attributed string, since it needs the string to be mutable.
I have added the overrides and some tests.

spevans:
@swift-ci please test

TestFoundation/TestNSAttributedString.swift
```diff
@@ -251,12 +355,303 @@ class TestNSMutableAttributedString : XCTestCase {
...
+            return
+        }
+        XCTAssertEqual(validAttribute, "attribute.placeholder.value")
+    }
+    
+    func test_initWithAttributedString() {
+        let string1 = "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
+        let attrKey1 = NSAttributedStringKey("attribute.placeholder.key1")
+        let attrValue1 = "attribute.placeholder.value1"
+        let attrRange1 = NSRange(location: 0, length: string1.utf8.count)
```
parkera: You could also use the new `NSRange` and `Range` conversions, e.g. https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/NSRange.swift#L180 and https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/NSRange.swift#L148

eyeplum: Thanks for the review.

These `string.utf8.count` were mistakes, I meant `utf16` to get the `length` as `NSString` does, sorry for the confusion.

I did some cleanups:
- Use `NSMutableAttributedString.init(string:attributes:)` when possible, this reduces some uses of the string ranges
- Convert `NSRange` from `Range` when trying to get the full range of the string

spevans:
@swift-ci please test

eyeplum:
Looks like the last test run finished successfully: https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/541/

millenomi:

PR #1380, opened 2017-12-24
Misc warnings fixes

spevans:
- Fixes for the following warnings:

  'BidirectionalSlice' is deprecated: renamed to 'Slice'
  'DefaultRandomAccessIndices' is deprecated: renamed to 'DefaultIndices'
  'index(_:offsetBy:)' is deprecated: all index distances are now of type Int
  initializer for struct 'CFRange' must use "self.init(...)" or "self = ..." because the struct was imported from C
  deinitialize()' is deprecated: the default argument to deinitialize(count:) has been removed, please specify the count explicitly
  'deallocate(capacity:)' is deprecated: Swift currently only supports freeing entire heap blocks, use deallocate() instead


spevans:
@swift-ci please test

pushkarnk:

PR #1381, opened 2017-12-24
Added guard around #include <xlocale.h> for Linux

tachoknight:
`xlocale.h` was removed in glibc 2.26 so any version of Linux with that or later versions will not be able to compile successfully. This change adds a guard to check that if we're *not* compiling on Linux, to go ahead and include it.


millenomi:
@swift-ci please test.

pushkarnk:
@tachoknight Thanks for this pull request! 

This means CF never used anything from `xlocale.h` on Linux. Though it makes sense to exclude it on Linux, another option could have been including `locale.h` which [I learn](https://sourceware.org/glibc/wiki/Release/2.26#Removal_of_.27xlocale.h.27) is a superset of `locale.h`. Just mentioning this for future references.

pushkarnk:

parkera:
I would rather we have used the DEPLOYMENT_TARGET macros here.

PR #1382, opened 2017-12-26
Initial implementation of URLProtectionSpace

pushkarnk:
This is one of the pre-reqs for an upcoming implementation of basic auth with URLSession. URLProtectionSpace is a simple class with a handful of read-only properties which are initialised using one of the two initialisers - I am not sure if there's value in adding tests here (however, I am open to suggestions!)


pushkarnk:
@swift-ci please test

ianpartridge:

PR #1287, opened 2017-10-28
NSData: Align methods with Darwin

spevans:
- Remove 'convenience' from:
  init()
  init(bytes:length:)
  init(bytesNoCopy:length:)
  init(bytesNoCopy:length:freeWhenDone:)
  init(bytesNoCopy:length:deallocator:)
  init(contentsOfFile:options:)
  init?(contentsOfFile:)
  init(data:)
  init(contentsOf:,options:)
  init?(base64Encoded base64String:options:)
  init?(base64Encoded base64Data:options:)

- Add:
  init?(contentsOf:)
  (deprecated) base64Encoding()

- Add (unimplemented):
  (deprecated) init?(contentsOfMappedFile:)
  (deprecated) class func dataWithContentsOfMappedFile()
  contains()
  contains(where:)

NSMutableData:

- Remove 'convenience' from:
   init?(capacity:)
   init?(length:)

- Add:
  init(bytes:length:)

- Add description documentation to public methods.

- Update tests.

PR #1385, opened 2017-12-27
NSMutableData: Restore missing initialisers

spevans:
- When aligning the init() methods with Darwin in PR #1287,
  converting the NSData `convenience` methods to designated
  initialisers stopped them from being inherited by NSMutableData
  as it declared its own designed initialisers.

- Add in all of the missing initialisers and add tests to ensure
  their presence is checked at compile time.

- Fix init(data: Data) to use the correct length of the data
  argument when copying the bytes.

### DIFF
--- a/TestFoundation/TestNSData.swift
+++ b/TestFoundation/TestNSData.swift
@@ -206,7 +206,6 @@ class TestNSData: XCTestCase {
             ("test_dataHash", test_dataHash),
             ("test_genericBuffers", test_genericBuffers),
             ("test_writeFailure", test_writeFailure),
-            ("testBasicConstruction", testBasicConstruction),
             ("testBridgingDefault", testBridgingDefault),
             ("testBridgingMutable", testBridgingMutable),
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

TestFoundation/TestNSData.swift
```diff
@@ -206,7 +206,6 @@ class TestNSData: XCTestCase {
             ("test_dataHash", test_dataHash),
             ("test_genericBuffers", test_genericBuffers),
             ("test_writeFailure", test_writeFailure),
-            ("testBasicConstruction", testBasicConstruction),
```
phausler: why was this test removed?

spevans: It was a duplicate line, that test is called on line 198

phausler: ah ok, I see it past the code fold

spevans:

PR #1386, opened 2017-12-28
[SR-6671] Decimal.leastNonzeroMagnitude, leastNormalMagnitude should have -128 exponent

takasek:
Fixes https://bugs.swift.org/browse/SR-6671 .


takasek:
@swift-ci please test

parkera:
This one is shared with the overlay so we should make coordinated changes there if we go forward with it.

phausler:
Can we add some tests for this to verify the expected behavior?

takasek:
> we should make coordinated changes there

Do you mean that these codes should be fixed at the same time?

https://github.com/apple/swift/blob/49a8cf050304e3503960f0f1815afafe71dba28a/test/stdlib/TestDecimal.swift#L130-L133
https://github.com/apple/swift/blob/8f8f00012a391b7c254e2056ef382c768ca7671a/stdlib/public/SDK/Foundation/Decimal.swift#L22-L23

> Can we add some tests for this to verify the expected behavior?

Sure, please add them.

parkera:
@takasek I think @phausler was asking for you to write some additional tests and include them in this PR.

takasek:
OK, I'll add some tests.

TestFoundation/TestJSONSerialization.swift
```diff
@@ -1257,8 +1257,8 @@ extension TestJSONSerialization {
...
         XCTAssertEqual(try trySerialize([Decimal.leastFiniteMagnitude]), "[-3402823669209384634633746074317682114550000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000]")
-        XCTAssertEqual(try trySerialize([-Decimal.leastNonzeroMagnitude]), "[-0.0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001]")
-        XCTAssertEqual(try trySerialize([Decimal.leastNonzeroMagnitude]), "[0.0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001]")
+        XCTAssertEqual(try trySerialize([-Decimal.leastNonzeroMagnitude]), "[-0.00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001]")
+        XCTAssertEqual(try trySerialize([Decimal.leastNonzeroMagnitude]), "[0.00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001]")
```
takasek: These tests failed because of changes in this PR.

Foundation/Decimal.swift
```diff
@@ -101,31 +101,31 @@ public struct Decimal {
 
 extension Decimal {
     public static let leastFiniteMagnitude = Decimal(
-        _exponent: 127,
+        _exponent: Int32(Int8.max),
```
takasek: What do you think of such expression change?

millenomi:
@swift-ci Please test.

alblue:
@phausler @parkera can you check this over - are the added tests sufficient?

Foundation/Decimal.swift
stephentyrone: What's going on here, and where did this property come from? "Magnitudes" are always non-negative, but this is a huge negative number. I don't know why this property even exists, but the name and documentation are *extremely* misleading.

(@takasek This is tangential to your change, but this is the first I've seen this).

millenomi:
Are the objections above understood? Do they still need correction?

alblue:
This seems to have gone stale. Is it still needed, or can it be closed?

spevans:
@swift-ci test

takasek:
Any problem still on this issue?

alblue:
@swift-ci test

takasek:
This issue is yet stale.
`NSDecimalNumber` defines its exponent's range is down to -128.
https://developer.apple.com/documentation/foundation/nsdecimalnumber
Thus `Foundation.Decimal` should follow it.

shahmishal:

PR #1392, opened 2018-01-12
[FHS bundles] Proposal

millenomi:
Please leave comments here or at the thread I'm about to start on the swift-corelibs-dev list.

### DIFF
--- a/Docs/FHS Bundles.md
+++ b/Docs/FHS Bundles.md
@@ -23,7 +23,7 @@ The base name of this bundle is the name of this directory, removing the extensi
 
 ### Bundle Paths
 
-The bundle's `.resources` directory contains resources much like an iOS-style flat bundle. Unlike on iOS, however, executables are not contained directly inside this directory. Instead, given that the bundle exists in `share/MyFramework.resources`, then an installed bundle will source from the following additional paths within the prefix the bundle is installed in (the parent directory of the `share` directory):
+The bundle's `.resources` directory contains resources much like an [iOS-style flat bundle](https://developer.apple.com/library/content/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW1). Unlike on iOS, however, executables are not contained directly inside this directory. Instead, given that the bundle exists in `share/MyFramework.resources`, then an installed bundle will source from the following additional paths within the prefix the bundle is installed in (the parent directory of the `share` directory):
 
  - The main executable will be searched in the `bin`, `sbin`, or `lib…` directories.
 	 - Executable bundles will search `bin`, then `sbin`, for an executable with the same base name as the resources folder, plus any platform prefix or suffix for executables. e.g.: `bin/MyApp`.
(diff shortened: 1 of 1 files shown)


millenomi:
As a formality:

@swift-ci Please test.

Docs/FHS Bundles.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Bundle Paths
+
+The bundle path for an installed bundle is the same as its resources path. That's the path that needs to be passed into `CFBundleCreate()` and equivalents for the bundle to be successfully created; passing associated paths, including the path to the executable, will return `NULL`. [⊗](#nullForInnerPaths)
+
+The `.resources` directory functions exactly like an iOS bundle, returning the same paths. By way of example, for a freestanding bundle created with `CFBundleCreate(NULL, …(…"./MyFramework.resources"))`:
+
+Function | Path returned
+---|---
+`CFBundleCopyBundleURL` | `…/MyFramework.resources`
```
e78l: Just to be sure: does `...` mean the parent directory?

millenomi: It meant 'the same directory as … in the other entries', but I see how it can be confusing. I will clarify this.

Docs/FHS Bundles.md
```diff
@@ -0,0 +1,123 @@
...
+
+ - The main executable will be searched in the `bin`, `sbin`, or `lib…` directories.
+	 - Executable bundles will search `bin`, then `sbin`, for an executable with the same base name as the resources folder, plus any platform prefix or suffix for executables. e.g.: `bin/MyApp`.
+	 - Framework bundles will search the appropriate `lib…` directories for a library of the same name, adding appropriate platform prefixes or suffixes for shared library file (falling back to just `lib` otherwise). e.g.: `lib/libMyFramework.so`.
+ - Auxiliary executables and libraries will be in, or in subdirectories of, `libexec/MyFramework.executables`;
+ - For framework bundles, `include/MyFramework` will contain headers, though this directory does affect the return values of the runtime API.
+
+Both paths are optional. If they don't exist, the API will behave much like on Darwin platforms.
```
aciidb0mb3r: Can you expand a bit here? Which API? How does it behave on Darwin?

Docs/FHS Bundles.md
```diff
@@ -0,0 +1,123 @@
...
+An installed bundle exists if there is:
+
+ - A directory with the `.resources` extension;
+ - contained in a directory named `share`.
+
+The base name of this bundle is the name of this directory, removing the extension. (E.g., for `/usr/share/MyFramework.resources`, the base name is `MyFramework`.)
+
+### Bundle Paths
+
+The bundle's `.resources` directory contains resources much like an iOS-style flat bundle. Unlike on iOS, however, executables are not contained directly inside this directory. Instead, given that the bundle exists in `share/MyFramework.resources`, then an installed bundle will source from the following additional paths within the prefix the bundle is installed in (the parent directory of the `share` directory):
```
aciidb0mb3r: Can you include reference to some document which explains iOS-style flat bundle? I see many references to the iOS bundle in this proposal.

ianpartridge:
Hi @millenomi - from my reading of FHS `/usr/local/share` is intended for architecture-independent data.  Do you intend for installed bundles to only contain architecture-independent data?

As Foundation runs on platforms including Linux on z Systems (LinuxONE) and Linux on POWER it's important that we have a cross-architecture approach.  Just checking that this has been considered in your thinking.  Perhaps you could elaborate how this would work.  Thanks!

millenomi:
The idea is that bundle resources, which are for all bundles we intend to ship this way architecture-independent, will live under `share`; architecture-dependent parts will live in `bin`, `sbin`, `lib<suffix>`, `lib` (for the main executable) and `libexec` (for auxiliary executables and embedded frameworks.) So, no, not all bundle content will live in `share`.

Docs/FHS Bundles.md
millenomi: I will revise with a reference to https://developer.apple.com/library/content/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW1.

ianpartridge:
Thanks @millenomi that sounds good.

One other thing. I imagine that in future we intend to ship bundles containing Foundation resources as part of the Linux toolchains from swift.org (for things like localization content), so we'd have a `/usr/local/share/Foundation.resources` directory?

How would this work when a user wants to have multiple versions of Swift installed on their Linux machine, which would ship with different bundle content?

Docs/FHS Bundles.md
aciidb0mb3r: Perfect, thanks!

millenomi:
I'm going to circle back to you about that shortly, but the preliminary thought is that right now no version of Swift can coinhabit the same `--prefix`. However, we _do_ want to have a proposal for when that can occur, but we may postpone the details if it turns out we can do it in a purely additive way.

millenomi:
Perfunctorily,

@swift-ci Please test.

Docs/FHS Bundles.md
```diff
@@ -0,0 +1,123 @@
...
+
+ - The main executable will be searched in the `bin`, `sbin`, or `lib…` directories.
+	 - Executable bundles will search `bin`, then `sbin`, for an executable with the same base name as the resources folder, plus any platform prefix or suffix for executables. e.g.: `bin/MyApp`.
+	 - Framework bundles will search the appropriate `lib…` directories for a library of the same name, adding appropriate platform prefixes or suffixes for shared library file (falling back to just `lib` otherwise). e.g.: `lib/libMyFramework.so`.
+ - Auxiliary executables and libraries will be in, or in subdirectories of, `libexec/MyFramework.executables`;
+ - For framework bundles, `include/MyFramework` will contain headers, though this directory does affect the return values of the runtime API.
```
ddunbar: Will we also support a Modules directory like we do for frameworks? One reason this distinction is important is that a framework module map is written differently (`framework module`) than one for flat headers, and there may be something we want to do to reconcile this (presumably Clang itself isn't going to start understanding this convention).

Docs/FHS Bundles.md
```diff
@@ -0,0 +1,123 @@
...
+This setup complies with the [Filesystem Hierarchy Standard (FHS) 3.0](https://refspecs.linuxfoundation.org/fhs.shtml), used by most Linux distributions. It also fits with the intent of the `/usr` and `/usr/local` directories of many BSD systems (e.g. from FreeBSD's [hier(7)][].)
+
+[hier(7)]: https://www.freebsd.org/cgi/man.cgi?hier(7)
+
+### Definition
+
+An installed bundle exists if there is:
+
+ - A directory with the `.resources` extension;
+ - contained in a directory named `share`.
```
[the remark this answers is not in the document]
millenomi: This is complete. As long as the CFBundle code finds that its client is trying to create a bundle at `$PREFIX/share/$NAME.resources`, we treat it as a FHS installed bundle. This is to support installing bundles in any `--prefix=$PREFIX`.

Docs/FHS Bundles.md
millenomi: Good question. I'll circle back to you on that.

Docs/FHS Bundles.md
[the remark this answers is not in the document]
ddunbar: That makes sense, thanks!

Docs/FHS Bundles.md
millenomi: So.

This document _does_ affect the discussion you mention, but right now we're a little more interested in the runtime aspect of this. The compile-time aspect is additive, and we can set that up with the clang team as a follow-up from implementing this — there's no runtime API that is affected by the presence (or absence) of the headers or of the module file, as far as I can tell.

ddunbar: That makes sense. I don't see any issues with tackling that later.

millenomi:
We seem to have reached some consensus. I will write SRs for the remaining tasks, and we can continue discussion on the forums if needed once they're up.

millenomi:
Ahem.

millenomi:
Looks like there was a transient issue.

I'll just force-merge this, since it has no code changes.

PR #1394, opened 2018-01-15
Pass -sil-merge-partial-modules flag

slavapestov:
The build scripts for foundation bypass the Swift driver and use
the -frontend mode directly. The command line interface for
-frontend is not officially supported and is subject to change.

A recent change was that when merging partial modules to form a
final module file, the -sil-merge-partial-modules flag must be
passed in so that serialized SIL can be preserved in the final
module file.

Preserving serialized SIL is now a requirement, since soon
default argument generators and stored property initializers
will be emitted with non-public linkage, requiring clients to
deserialize SIL when calling these entry points.

### DIFF
--- a/lib/script.py
+++ b/lib/script.py
@@ -189,7 +189,7 @@ rule CompileSwift
     depfile = $out.d
 
 rule MergeSwiftModule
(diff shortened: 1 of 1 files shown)


slavapestov:
@swift-ci Please test Linux

millenomi:

PR #1397, opened 2018-01-18
SR-6449: FileManager.removeItem(atPath) leaks

spevans:
- FileManager.fileSystemRepresentation returns a heap allocated
  pointer that needs to be freed by the caller.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -461,6 +461,7 @@ open class FileManager : NSObject {
             let stream = fts_open(ps, FTS_PHYSICAL | FTS_XDEV | FTS_NOCHDIR, nil)
             ps.deinitialize(count: 2)
             ps.deallocate()
+            fsRep.deallocate()
 
             if stream != nil {
                 defer {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

parkera:
I think we may need to instead introduce a new API here that gives you the representation in a closure, somewhat like withUnsafeBytes. Plus mark this one as unavailable on Linux.

These APIs that rely on an autorelease pool are problematic for us.

spevans:
@parkera I agree that replacing it with a better API is preferable (as we should with  all of the methods that return an autoreleased result). Can we not just return a `String` and try and avoid any mention of `Unsafe`?

spevans:
@parkera Is it worth just fixing this in the 4.1 branch to remove the leak but then coming up with something better for 5 onwards?

parkera:
I think we're stuck with this until Swift 5, since we shouldn't introduce a source breaking change in a dot release. The Swift 5 solution probably has to be a replacement method.

parkera:
FileManager could perhaps switch to some other mechanism of getting the C string.

spevans:
@parkera Sorry, when I said fixing it in 4.1 I meant applying this PR to 4.1-branch to just fix the leak, not change the API.

parkera:
I thought about this some and I guess this makes sense to do in the short term. Eventually we'll have to just deprecate this API (or remove it, maybe).

spevans:

spevans:

spevans:

PR #1399, opened 2018-01-22
Use FHS/XDG paths for UserDefaults on Linux & BSD-y OSes

millenomi:
On FHS/XDG systems, the paths for UserDefaults are now:

 - `/usr/local/etc` for AnyUser preferences (replaces `/Library/Preferences`)
 - `$XDG_CONFIG_DIR` for CurrentUser preferences (replaces `$HOME/Library/Preferences`)

(`$XDG_CONFIG_DIR` is typically, and defaults to, `$HOME/.config/`.)

### DIFF
--- a/CoreFoundation/Base.subproj/CFFileUtilities.c
+++ b/CoreFoundation/Base.subproj/CFFileUtilities.c
@@ -1179,7 +1179,7 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
 }
 
 
-#if DEPLOYMENT_RUNTIME_SWIFT
+#if !DEPLOYMENT_RUNTIME_OBJC
 
 // https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
 // Version 0.8
@@ -1356,5 +1356,5 @@ CF_PRIVATE CFArrayRef _CFCreateCFArrayByTokenizingString(const char *values, cha
     return CFArrayCreate(kCFAllocatorSystemDefault, NULL, 0, &kCFTypeArrayCallBacks);
 }
 
-#endif
+#endif // !DEPLOYMENT_RUNTIME_OBJC
 
--- a/CoreFoundation/Base.subproj/CFKnownLocations.c
+++ b/CoreFoundation/Base.subproj/CFKnownLocations.c
@@ -59,7 +59,9 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
         assert(user == _kCFKnownLocationUserCurrent);
         
         if (user == _kCFKnownLocationUserCurrent) {
-            location = _CFXDGCreateConfigHomePath();
+            CFStringRef path = _CFXDGCreateConfigHomePath();
+            location = CFURLCreateWithFileSystemPath(kCFAllocatorSystemDefault, path, kCFURLPOSIXPathStyle, true);
+            CFRelease(path);
         }
     }
     
--- a/CoreFoundation/Base.subproj/CFKnownLocations.c
+++ b/CoreFoundation/Base.subproj/CFKnownLocations.c
@@ -15,10 +15,7 @@
 
 #include <assert.h>
 
-CONST_STRING_DECL(_kCFKnownLocationUserAny, " == _kCFKnownLocationUserAny");
-CONST_STRING_DECL(_kCFKnownLocationUserCurrent, " == _kCFKnownLocationUserCurrent");
-
-CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUser user) {
+CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUser user, CFStringRef _Nullable username) {
     CFURLRef location = NULL;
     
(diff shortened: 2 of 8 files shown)


millenomi:
@swift-ci Please test.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1198,7 +1198,7 @@ static CFStringRef _CFXDGCreateHome(void) {
 }
 
 /// a single base directory relative to which user-specific data files should be written. This directory is defined by the environment variable $XDG_DATA_HOME.
-CF_SWIFT_EXPORT
```
parkera: This should still only be exported on non-Darwin platforms. We may need another version of this `CF_EXPORT` macro.

CoreFoundation/Base.subproj/CFKnownLocations.c
```diff
@@ -0,0 +1,75 @@
...
+
+#include "CFKnownLocations.h"
+
+#include <CoreFoundation/CFString.h>
+#include "CFPriv.h"
+#include "CFInternal.h"
+
+#include <assert.h>
+
+CONST_STRING_DECL(_kCFKnownLocationUserAny, " == _kCFKnownLocationUserAny");
```
parkera: Hm, the ` == ` at the start is something unusual. In most cases the string is equal to the key name.

CoreFoundation/Base.subproj/CFKnownLocations.c
```diff
@@ -0,0 +1,75 @@
...
+    
+/*
+ Building for a Darwin OS. (We use these paths on Swift builds as well, so that we can interoperate a little with Darwin's defaults(1) command and the other system facilities; but you want to use the system version of CF if possible on those platforms, which will talk to cfprefsd(8) and has stronger interprocess consistency guarantees.)
+ 
+ User:
+ - Any: /Library/Preferences
+ - Current: $HOME/Library/Preferences
+ */
+    
+    if (user == _kCFKnownLocationUserAny) {
```
parkera: If we're just going to do an `==` check, perhaps we should just declare an enum?

CoreFoundation/Base.subproj/CFKnownLocations.h
```diff
@@ -0,0 +1,41 @@
...
+ - For platforms that use the XDG spec to identify a configuration path in a user's home, we cannot determine that path for any user other than the one we're currently running as.
+ 
+ So:
+  - We're keeping that behavior when building Core Foundation for Darwin/ObjC for compatibility, hence the _EXTENSIBLE above; but
+  - For Swift and any new platform, we are enforcing the documented constraint. Using a user value other than …Any or …Current above will assert (or return NULL if asserts are off).
+ 
+ See CFKnownLocations.c for a summary of what paths are returned.
+ */
+
+extern CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUser user);
```
parkera: We'll need to make sure this is only exported on non-Darwin platforms.

CoreFoundation/Base.subproj/CFKnownLocations.c
millenomi: Yeah, that makes sense. It'll be the enum plus a NULLable username string argument.

CoreFoundation/Base.subproj/CFKnownLocations.h
millenomi: Ok!

CoreFoundation/Base.subproj/CFKnownLocations.c
millenomi: Yeah — the idea here was to make sure it couldn't be a valid username. Since we're switching to an enum, this problem goes away.

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -327,8 +327,6 @@ CF_EXPORT _Nullable CFErrorRef CFReadStreamCopyError(CFReadStreamRef stream);
 
 CF_EXPORT _Nullable CFErrorRef CFWriteStreamCopyError(CFWriteStreamRef stream);
 
-CF_SWIFT_EXPORT Boolean _CFBundleSupportsFHSBundles(void);
```
millenomi: This removal is a merge issue and incorrect.

millenomi:
@swift-ci Please test.

millenomi:
I'm going to have a proposal for CF_SWIFT_EXPORT as a separate patch, and then rebase this one.

achivetta:
Out of curiosity, why `/usr/local/etc` rather than `/etc`?

millenomi:
`/etc` is for system-level configuration (think `/System/Library/` on macOS). There is some confusion between the two, and [the FHS actually allows](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.pdf) linking `/usr/local/etc` to `/etc/local` so that all configuration files are in `/etc`, but the intent is that this directory be e.g. never replaced by a system update if it's modified (think `/Library` on macOS.)

millenomi:
@parkera This has been rebased on master and has fixes for your review comments.

@swift-ci Please test.

millenomi:
@swift-ci Please test.

millenomi:

PR #1400, opened 2018-01-22
Implement forced keys in UserDefaults.

millenomi:
If we’re running this version of Foundation, there is no MCX nor configuration profiles that can force a defaults key. Therefore, by definition, unless in the future we want to introduce such a facility in non-Apple OSes, this is always false.


millenomi:

millenomi:
A lower-layer issue.

PR #1405, opened 2018-01-24
Fix crashes when creating directories at single component relative paths

mattrajca:
Bail early if after calling `deletingLastPathComponent` the path is empty. There is
no point in doing a traversal and trying to create an empty directory at the parent
path. In addition, this ends up tripping a pre-condition when we try finding the next
parent path component again.

This addresses SR-4570.


millenomi:
@swift-ci Please test.

millenomi:

PR #1406, opened 2018-01-24
ExtraStringAPI: Sync with the overlay version

spevans:
- The three functions are obsoleted in Swift 4.0. The bodies from
  the overlay wont compile due to use of internal symbols so replace
  with a fatalError() call which isnt used anyway.

- Update callers of advanced(by:) to use index(offsetBy:).

### DIFF
--- a/Foundation/ExtraStringAPIs.swift
+++ b/Foundation/ExtraStringAPIs.swift
@@ -2,21 +2,28 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
-// See http://swift.org/LICENSE.txt for license information
-// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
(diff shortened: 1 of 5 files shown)


millenomi:
@swift-ci Please test.

Foundation/ExtraStringAPIs.swift
```diff
@@ -2,21 +2,28 @@
...
 
-// Random access for String.UTF16View, only when Foundation is
-// imported.  Making this API dependent on Foundation decouples the
-// Swift core from a UTF16 representation.
 extension String.UTF16View.Index {
-    public func advanced(by n: Int) -> String.UTF16View.Index {
-        return String.UTF16View.Index(encodedOffset: encodedOffset.advanced(by: n))
-    }
-}
+  /// Construct from an integer offset.
```
alblue: Is there a way we can change this from an obsoleted to a `renamed` with instead, so that callers have the opportunity of a fixit to change? Alternatively, instead of just using "Unavailable" can we have a more descriptive message as to what they need to replace this with instead?

spevans: For `advanced(by:)` would this suffice (I don't think a `rename` can be used in its place. ) ?
`@available(swift, obsoleted: 4.0, message: "Use index(_,offsetBy:) instead.")`

Im not sure what the alternatives are for the other 2 although I could remove them entirely since adding obsoleted functions does not really make sense anyway.

parkera: We seem to be moving towards keeping them in for a release or two because then the compile error is clearer than 'can't find it'.

millenomi:

PR #1403, opened 2018-01-23
UserDefaults: Implement volatile domains, argument domain

millenomi:
This implements:
 - The `.volatileDomain(forName:)` API and related calls.
 - The volatile argument domain, `UserDefaults.argumentDomain`.

A note: since there is no language-supported bridging in Swift Foundation, we opt to accept and return value types (Array<_>, Dictionary<_,_>, Date, URL…) rather than reference types (NSArray, NSDictionary, NSDate, NSURL…). We internally convert to reference types them to a. check that you are only saving plist types and b. to support existing persistent domain API (since these objects go to CF and need to be reference types).

PR #1408, opened 2018-01-25
Implement missing parts of UserDefaults and fix consistency issues

millenomi:
This supersedes the patch at https://github.com/apple/swift-corelibs-foundation/pull/1403.

While implementing persistent domains as a follow-up to volatile domains (and the arguments domain), I've found several issues in UserDefaults that required a whole bit of rethinking of the way this class works. It has been substantially changed.

This patch:
 - Implements volatile domains
 - Implements the arguments domain as a volatile domain that is searched at the appropriate time
 - Implements persistent domains
 - Changes UserDefaults so that all methods that return Any values return value types instead of reference types. On Linux, where we don't (yet?) have Foundation reference <-> value type bridging, we have to pick one of the two, and I feel it is more likely and desirable for an end user to want the value type rather than the reference type.
 - Changes the internals of the class so that any reference types are converted to value types ASAP where possible, and adopt `_SwiftValue` instead of trying to do ad-hoc conversions.


millenomi:
@swift-ci Please test.

Foundation/UserDefaults.swift
```diff
@@ -469,30 +349,44 @@ open class UserDefaults: NSObject {
...
+            var removedAny = false
             for key in defaults._dictionaryRepresentation(searchingOutsideOfSuite: false).keys {
                 defaults.removeObject(forKey: key)
+                removedAny = true
             }
             
-            NotificationCenter.default.post(name: UserDefaults.didChangeNotification, object: self)
+            _ = defaults.synchronize()
+            
+            if removedAny {
```
Catfish-Man: This notification is posted unconditionally in the ObjC implementation

Foundation/UserDefaults.swift
```diff
@@ -469,30 +349,44 @@ open class UserDefaults: NSObject {
...
+                NotificationCenter.default.post(name: UserDefaults.didChangeNotification, object: self)
+            }
         }
     }
     
     open func synchronize() -> Bool {
-        return CFPreferencesAppSynchronize(kCFPreferencesCurrentApplication)
+        return CFPreferencesAppSynchronize(suite?._cfObject ?? kCFPreferencesCurrentApplication)
```
Catfish-Man: Something I did in the ObjC implementation is add CF SPI for “synchronize everything” and call it here, since a ubiquitous pattern of misuse is synchronizing standard user defaults when you set something else.

Foundation/UserDefaults_Arguments.swift
```diff
@@ -0,0 +1,57 @@
...
+    static func _parseArguments(_ arguments: [String]) -> [String: Any] {
+        var result: [String: Any] = [:]
+        
+        let count = arguments.count
+        
+        var index = 0
+        while index < count - 1 { // We're looking for pairs, so stop at the second-to-last argument.
+            let current = arguments[index]
+            let next = arguments[index + 1]
+            if current.hasPrefix("-") && !next.hasPrefix("-") {
```
Catfish-Man: I bet we have subtle unicode behavior handling differences due to Swift string vs NSString, but I don’t think that’s a problem we need to solve here, at least not right now

millenomi:
@swift-ci Please test.

Foundation/UserDefaults_Arguments.swift
millenomi: As far as I can see here, this allows a subset of all defaults we allow on Darwin — namely, we skip all defaults that start with a '-' code point but where that code point forms a grapheme, e.g. '-⃣'. It still allows basically all defaults keys of actual interest, though. But yep.

millenomi: … all defaults _arguments_.

millenomi:
@swift-ci Please test.

millenomi:
@swift-ci please test

Foundation/UserDefaults.swift
millenomi: Will change.

Foundation/UserDefaults.swift
millenomi: … ah. hm. I'm inclined to have a follow-up on this one.

millenomi:
OK, this is ready for review and test (once CI comes back up.) @Catfish-Man @parkera

millenomi:
@swift-ci Please test

millenomi:
@swift-ci please test

alblue:
@swift-ci please test

Foundation/UserDefaults.swift
```diff
@@ -73,147 +131,103 @@ open class UserDefaults: NSObject {
...
         }
-        if let bVal = aVal as? NSString {
-            return bVal.doubleValue
+        if let bVal = aVal as? String {
+            return NSString(string: bVal).doubleValue
         }
         return 0
     }
+    
     open func bool(forKey defaultName: String) -> Bool {
```
Catfish-Man: One minor incompatibility between this and what darwin does, which we probably don’t want/need to fix: NSUserDefaults historically coerced values to int/bool/float using different logic from CF. I changed it so that subclasses maintain the old behavior (for compatibility reasons), but un-subclassed NSUserDefaults always uses the CF logic.

I haven’t checked whether this implements the NSUserDefaults version or the CFPrefs version, but I think we should use the CFPrefs behavior.

Foundation/UserDefaults.swift
```diff
@@ -12,17 +12,85 @@ import CoreFoundation
...
+        return value
+    }
+}
+
 open class UserDefaults: NSObject {
+    static private func _isValueAllowed(_ nonbridgedValue: Any) -> Bool {
+        let value = bridgeFromNSCFTypeIfNeeded(nonbridgedValue)
+        
+        if let value = value as? [Any] {
+            for innerValue in value {
```
Catfish-Man: The swift tinkerer in me wants to say .lazy.filter { _isValueAllowed($0) }.count > 0, but tbh I’m not sure it’s worth it :)

Foundation/UserDefaults.swift
```diff
@@ -299,31 +288,105 @@ open class UserDefaults: NSObject {
...
     
-    open func dictionaryRepresentation() -> [String : Any] {
-        guard let aPref = CFPreferencesCopyMultiple(nil, kCFPreferencesCurrentApplication, kCFPreferencesCurrentUser, kCFPreferencesCurrentHost),
-            let bPref = (aPref._swiftObject) as? [NSString: Any] else {
-                return registeredDefaults
+    open func dictionaryRepresentation() -> [String: Any] {
+        return _dictionaryRepresentation(searchingOutsideOfSuite: true)
+    }
+    
+    private func _dictionaryRepresentation(searchingOutsideOfSuite: Bool) -> [String: Any] {
```
Catfish-Man: searchingOutsideOfSuite seems like a slightly odd name for this. “Suite” is a very fuzzy term in NSUD, but to the extent that it means anything I tend to think of it as referring to a search list, which would generally include registered prefs, etc...

I’m guessing the sense you mean here is “outside of the set of domains matching this identifier”, which is a subset of the search list for that identifier?

Foundation/UserDefaults.swift
millenomi: I will reply with my standard reply, which is 'if it turns out to be a hot spot'.

Foundation/UserDefaults.swift
millenomi: Yup. What about `_dictionaryRepresentation(includingOnlySuiteNameDomains:)`?

millenomi: We chatted privately and agreed on `_dictionaryRepresentation(includingVolatileDomains:)`.

millenomi:

millenomi:

millenomi:
@swift-ci please test

PR #1409, opened 2018-01-25
Avoid crashing when XPath queries return no results.

mattrajca:
Check before dereferencing `nodesetval` because it may be `NULL` if an XPath query
returns no results (but is a valid query).

Fixes SR-4628.


alblue:

PR #1413, opened 2018-01-28
[URLRequest] Copy body

dtorres:
This is the behaviour in Darwin Foundation, so.. ```¯\_(ツ)_/¯```


spevans:
@swift-ci please test

parkera:
Let's add a unit test to verify the behavior.

natecook1000:
It looks like there are still a few instance variables missing from this method, which will result in unexpected behavior just like the reported issue.

alblue:
This PR has gone stale. Please rebase and resubmit.

PR #1416, opened 2018-01-30
Add test for SR-3941.

mattrajca:
<SR-3941: AffineTransform.rotate(byRadians:) can be wrong> looks fixed already.
Add a test for it.


millenomi:
@swift-ci Please test.

PR #1419, opened 2018-02-03
Update Status to reflect HTTPCookieStorage implementation status

AndrewSB:
I'm not sure why these weren't updated in previous commits. Maybe I've misunderstood something. There appear to be a substantial number of tests, and a pretty complete looking implementation.


pushkarnk:
Hey @AndrewSB Can you please resolve the conflicts so that we can merge this one? Thanks!

pushkarnk:
There's one method that's still `NSUnimplemented()`. I think we should keep the status as `Mostly Implemented` as it is today (after a rebase with the latest master).

AndrewSB:
@pushkarnk rebased 👍 but it looks like #1422 has implemented the right thing in the interim.

I'm sorry about not doing this sooner, I was having issues with my GitHub notifications, and it looked like I missed this one.

PR #1423, opened 2018-02-06
Restore Data.Indices to be a CountableRange

airspeedswift:
Needed to keep master compiling against the Swift 5.0 branch. To be reverted once the Swift repo rebranches 5.0.


airspeedswift:
@swift-ci please test

parkera:
Do we need a 5.0 branch for SCL-foundation to match everything else?

airspeedswift:
That would help avoid this problem going forward. I guess the main question is, how much maintenance burden would that introduce.

PR #1424, opened 2018-02-07
Restore two more instances of CountableRange

airspeedswift:
Similar to #1423 to keep compatibility with both the 5.0 branch and master. To be reverted once swift rebranches 5.0.


airspeedswift:
@swift-ci test

PR #1425, opened 2018-02-08
Implement NSMeasurement

Dante-Broggi:
This is implementing the `NSUnimplemented` methods.

### DIFF
--- a/Foundation/NSMeasurement.swift
+++ b/Foundation/NSMeasurement.swift
@@ -64,7 +64,7 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
         } else {
             guard let dimension = unit as? Dimension,
                     let otherDimension = rhs.unit as? Dimension else {
-                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: dimension)) rhs: \(type(of: otherDimension))")
+                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: unit)) rhs: \(type(of: rhs.unit))")
                 }
             let selfValueInTermsOfBase = dimension.converter.baseUnitValue(fromValue: self.doubleValue)
             let rhsValueInTermsOfBase = otherDimension.converter.baseUnitValue(fromValue: rhs.value)
@@ -78,7 +78,7 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
         } else {
             guard let dimension = unit as? Dimension,
                     let otherDimension = rhs.unit as? Dimension else {
-                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: dimension)) rhs: \(type(of: otherDimension))")
+                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: unit)) rhs: \(type(of: rhs.unit))")
                 }
             let selfValueInTermsOfBase = dimension.converter.baseUnitValue(fromValue: self.doubleValue)
             let rhsValueInTermsOfBase = otherDimension.converter.baseUnitValue(fromValue: rhs.value)
--- a/Foundation/NSMeasurement.swift
+++ b/Foundation/NSMeasurement.swift
@@ -69,7 +69,7 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
         }
     }
     
-    open func copy(with zone: NSZone? = nil) -> Any { NSUnimplemented() }
+    open func copy(with zone: NSZone? = nil) -> Any { return self }
     
     open class var supportsSecureCoding: Bool { return true }
     
--- a/Foundation/NSMeasurement.swift
+++ b/Foundation/NSMeasurement.swift
@@ -24,7 +24,15 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
     }
     
     open func canBeConverted(to otherUnit: Unit) -> Bool {
-        return otherUnit.isKind(of: type(of: unit))
+        // this really should be something like 
+        // `(otherUnit is type(of: unit))` but 
+        // that requires `is` to take a 
+        // dynamic metatype argument on the rhs.
+        #if DEPLOYMENT_RUNTIME_OBJC //|| os(Linux)
+            return otherUnit.isKind(of: type(of: unit))
+        #else
+            return true // see above
+        #endif
     }
     
     open func converting(to otherUnit: Unit) -> Measurement<Unit> {
@@ -53,8 +61,12 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
         if self.unit.isEqual(rhs.unit) {
             return Measurement(value: self.doubleValue + rhs.value, unit: self.unit)
         } else {
-            let selfValueInTermsOfBase = self.unit.converter.baseUnitValue(fromValue: self.doubleValue)
-            let rhsValueInTermsOfBase = rhs.unit.converter.baseUnitValue(fromValue: rhs.value)
+            guard let dimension = unit as? Dimension,
+                    let otherDimension = otherUnit as? Dimension else {
+                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: unit)) rhs: \(type(of: otherUnit))")
+                }
+            let selfValueInTermsOfBase = dimension.converter.baseUnitValue(fromValue: self.doubleValue)
+            let rhsValueInTermsOfBase = otherDimension.converter.baseUnitValue(fromValue: rhs.value)
             return Measurement(value: selfValueInTermsOfBase + rhsValueInTermsOfBase, unit: type(of: self.unit).baseUnit())
         }
     }
@@ -63,8 +75,12 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
         if self.unit.isEqual(rhs.unit) {
             return Measurement(value: self.doubleValue - rhs.value, unit: self.unit)
         } else {
-            let selfValueInTermsOfBase = self.unit.converter.baseUnitValue(fromValue: self.doubleValue)
-            let rhsValueInTermsOfBase = rhs.unit.converter.baseUnitValue(fromValue: rhs.value)
+            guard let dimension = unit as? Dimension,
+                    let otherDimension = otherUnit as? Dimension else {
+                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: unit)) rhs: \(type(of: otherUnit))")
+                }
+            let selfValueInTermsOfBase = dimension.converter.baseUnitValue(fromValue: self.doubleValue)
+            let rhsValueInTermsOfBase = otherDimension.converter.baseUnitValue(fromValue: rhs.value)
             return Measurement(value: selfValueInTermsOfBase - rhsValueInTermsOfBase, unit: type(of: self.unit).baseUnit())
         }
     }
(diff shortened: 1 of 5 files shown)


ianpartridge:
Thanks for your contribution! Can you add some tests too?

P.S. We use 4 spaces for indentation here.

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,72 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
+    open func copy(with zone: NSZone? = nil) -> Any { return self }
     
     open class var supportsSecureCoding: Bool { return true }
     
-    open func encode(with aCoder: NSCoder) { NSUnimplemented() }
+    open func encode(with aCoder: NSCoder) { 
+      guard aCoder.allowsKeyedCoding else {
+            preconditionFailure("Unkeyed coding is unsupported.")
+        }
+        aCoder.encode(self.doubleValue, forKey:"NS.dblval")
```
parkera: `NS.value`

Dante-Broggi:
For tests, as there isn't a `TestNSMeasurement.swift` file, do I have to do anything special when adding it?

millenomi:
Please add it to the Xcode project and the list of test source files in the build.py file when you do.

Dante-Broggi:
The `build.py` includes `+ glob.glob('./TestFoundation/Test*.swift')) # all TestSomething.swift are considered sources to the test project in the TestFoundation directory` so, actually I don't need to add `TestNSMeasurement.swift`, but I did add it to the Xcode project.

spevans:
@swift-ci please test

spevans:
Looks like there were a couple of compilation errors, see https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/635/consoleFull#18229161593122a513-f36a-4c87-8ed7-cbc36a1ec144

spevans:
@swift-ci please test

spevans:
@Dante-Broggi there are still a couple of compile errors in CI: https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/636/console

Dante-Broggi:
So, the one thing I couldn't fix is `canBeConverted(to:)`, the original implementation of which used objc-only `isKind(of:)` and I cannot think of how to replace that.
The nearest I can think of is `(otherUnit is type(of: unit))`, except `is` would require a dynamic metatype argument on the rhs, and it currently only accepts static metatypes.

Dante-Broggi:
I thought of something that doesn't use `.isKind(of:)` for `canBeConverted(to:)`
*just attempt the conversion*, and return false instead of `fatalError()`ing.

spevans:
@swift-ci please test

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,95 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
+        }
+    }
     
-    open func adding(_ measurement: Measurement<Unit>) -> Measurement<Unit> { NSUnimplemented() }
+    open func adding(_ rhs: Measurement<Unit>) -> Measurement<Unit> {
+        if self.unit.isEqual(rhs.unit) {
+            return Measurement(value: self.doubleValue + rhs.value, unit: self.unit)
+        } else {
+            guard let dimension = unit as? Dimension,
+                    let otherDimension = otherUnit as? Dimension else {
```
spevans: Do you mean `rhs` instead of `otherUnit` ? (same on line 82)

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,95 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
+            return otherUnit.isKind(of: type(of: unit))
+        #else
+            // just check conversion
+            if let dimension = unit as? Dimension,
+               let otherDimension = otherUnit as? Dimension {
+                return true
+            } else {
+                return unit.isEqual(otherUnit)
+            }
+            re
```
spevans: Extra typo here

Foundation/NSMeasurement.swift
Dante-Broggi: Yes, indeed.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,94 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
     
-    open func adding(_ measurement: Measurement<Unit>) -> Measurement<Unit> { NSUnimplemented() }
+    open func adding(_ rhs: Measurement<Unit>) -> Measurement<Unit> {
+        if self.unit.isEqual(rhs.unit) {
+            return Measurement(value: self.doubleValue + rhs.value, unit: self.unit)
+        } else {
+            guard let dimension = unit as? Dimension,
+                    let otherDimension = rhs.unit as? Dimension else {
+                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: dimension)) rhs: \(type(of: otherDimension))")
+                }
```
spevans: This needs to be `(type(of: unit))` and `(type(of: rhs.unit))` as `dimension` and `otherDimension` don't exist in th `guard` block if the `as?` fails.

btw, are you testing on macOS or Linux?

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,94 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
     
-    open func canBeConverted(to unit: Unit) -> Bool { NSUnimplemented() }
+    open func canBeConverted(to otherUnit: Unit) -> Bool {
+        #if DEPLOYMENT_RUNTIME_OBJC //|| os(Linux)
+            return otherUnit.isKind(of: type(of: unit))
+        #else
+            // just check conversion
+            if let dimension = unit as? Dimension,
+               let otherDimension = otherUnit as? Dimension {
+                return true
```
spevans: `dimension` and `otherDimension` are not used so could be replaced with `_` or perhaps rewrite as `unit is Dimension && otherUnit is Dimension`.

Dante-Broggi: I think I thought of that, then forgot to do it.

Foundation/NSMeasurement.swift
Dante-Broggi: I am on a mac, but I really have no idea how to actually even build any of the swift projects.
So, realistically, nowhere. I am sorry, but I am just an enthusiast with no idea how to use the infrastructure. 🙁

spevans:
@Dante-Broggi Ok here's what to do assuming you are on either Sierra (10.12) or High Sierra (10.13) and have Xcode already installed:

Get the latest snapshot from https://swift.org/download/#snapshots Trunk development (master) 'Xcode' . This is a package you install on your mac.

Next, in Xcode, goto Xcode > Toolchains and you should see the snapshot you just downloaded in the menu, select it as your default tool chain. It will be called something like 'Swift Development Snapshot 2018-02-08 (a)' or something like that.

Now in the same directory that you checked out the `swift-corelibs-foundation` repo you also need to checkout the `swift-corelibs-xctest` repo.

Now in the `swift-corelibs-foundation` directory, open the `Foundation.xcworkspace` workspace (NOT the xcodeproj). The workspace includes the Foundation Xcode project and the XCTest Xcode project.

In the menu at the top the Xcode window, select 'TestFoundation > My Mac' then do 'Product > Run' and that will compile up CodeFoundation, Swift Foundation, XCtest and TestFoundation and then it will try and run the tests. Any compile errors should show up here. Don't worry about all of the warnings about deprecation in code that you haven't changed, there is usually because of changes in the main swift core in the snapshot you downloaded.

That should be enough to get you started, let me know if you have any questions about the above.

spevans:
@swift-ci please test

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,93 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
         self.unit = unit
     }
     
-    open func canBeConverted(to unit: Unit) -> Bool { NSUnimplemented() }
+    open func canBeConverted(to otherUnit: Unit) -> Bool {
+        #if DEPLOYMENT_RUNTIME_OBJC //|| os(Linux)
```
xwu: Nit: the stuff inside a compilation conditional should be indented as though there weren't a conditional surrounding it; in most files, the conditional statement itself should be outdented to the left margin, but the specific style can vary depending on the file.

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,93 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
+            return otherUnit.isKind(of: type(of: unit))
+        #else
+            // just check conversion
+            if unit is Dimension && otherUnit is Dimension {
+                return true
+            } else {
+                return unit.isEqual(otherUnit)
+            }
+        #endif
```
xwu: Does this implementation match Darwin Foundation? It seems to be very limited.

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,93 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
     
-    open func converting(to unit: Unit) -> Measurement<Unit> { NSUnimplemented() }
+    open func converting(to otherUnit: Unit) -> Measurement<Unit> {
+        if canBeConverted(to: otherUnit) {
+            if unit.isEqual(otherUnit) {
+                return Measurement(value: doubleValue, unit: otherUnit)
+            } else {
+                guard let dimension = unit as? Dimension,
+                    let otherDimension = otherUnit as? Dimension else {
+                        fatalError("Cannot convert differing units that are non-dimensional! lhs: \(type(of: unit)) rhs: \(type(of: otherUnit))")
```
xwu: Try to align these error messages as closely as possible with Darwin Foundation.

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,93 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
+    }
     
-    open func converting(to unit: Unit) -> Measurement<Unit> { NSUnimplemented() }
+    open func converting(to otherUnit: Unit) -> Measurement<Unit> {
+        if canBeConverted(to: otherUnit) {
+            if unit.isEqual(otherUnit) {
+                return Measurement(value: doubleValue, unit: otherUnit)
+            } else {
+                guard let dimension = unit as? Dimension,
+                    let otherDimension = otherUnit as? Dimension else {
```
xwu: Nit: the stuff inside the braces should be indented only four spaces as compared to `guard`; it's fine to indent the second line of the condition but it doesn't affect how the inside stuff is indented (the same applies elsewhere below).

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,93 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
+    open func copy(with zone: NSZone? = nil) -> Any { return self }
     
     open class var supportsSecureCoding: Bool { return true }
     
-    open func encode(with aCoder: NSCoder) { NSUnimplemented() }
+    open func encode(with aCoder: NSCoder) {
+        guard aCoder.allowsKeyedCoding else {
+            preconditionFailure("Unkeyed coding is unsupported.")
+        }
+        aCoder.encode(self.doubleValue, forKey:"NS.value")
```
xwu: Nit: space after `forKey:`; likewise one line below.

Foundation/NSMeasurement.swift
```diff
@@ -23,21 +23,93 @@ open class NSMeasurement : NSObject, NSCopying, NSSecureCoding {
...
     
-    public required init?(coder aDecoder: NSCoder) { NSUnimplemented() }
+    public required init?(coder aDecoder: NSCoder) {
+        guard aDecoder.allowsKeyedCoding else {
+            preconditionFailure("Unkeyed coding is unsupported.")
+        }
+        let doubleValue = aDecoder.decodeDouble(forKey: "NS.value")
+        let possibleUnit = aDecoder.decodeObject(forKey: "NS.unit")
+        guard let unit = possibleUnit as? Unit else {
+            return nil // or should we `fatalError()`?
```
xwu: What does Darwin Foundation do?

malhal: One thing it does is validation, e.g. @"Cannot add measurements of differing unit types!"

alblue:
@swift-ci please test

millenomi:
@Dante-Broggi are you still interested in chaperoning this patch?

ikesyo:
Looks like this is superseded by #2095.

PR #1427, opened 2018-02-08
Implement FileHandle#waitForDataInBackgroundAndNotify

RLovelett:
I proposed working on this method implementation on the [Swift forums](https://forums.swift.org/t/is-anyone-working-on-filehandles-wait-for-data-in-background-and-notify/9570?u=rlovelett).

This change is a work in progress. It needs a few details to be addressed. I am soliciting feedback.

1. [From the docs](https://developer.apple.com/documentation/foundation/filehandle/1409270-waitfordatainbackgroundandnotify) it says the function must be called from a thread that has an active run loop. How can you check if a thread has an active run-loop? Would that be an `assert` or `precondition` or something else entirely?

2. [From the docs](https://developer.apple.com/documentation/foundation/filehandle/1409270-waitfordatainbackgroundandnotify) it says the function will post the `NSFileHandleDataAvailable` notification on the current thread. Presumably that means the thread that was used to call `waitForDataInBackgroundAndNotify` in the first place. How do you specify the thread when posting a notification?

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -8,6 +8,7 @@
 //
 
 import CoreFoundation
+import Dispatch
 
 #if os(OSX) || os(iOS)
 import Darwin
@@ -345,7 +346,19 @@ extension FileHandle {
     }
 
     open func waitForDataInBackgroundAndNotify() {
-        NSUnimplemented()
+        let global = DispatchQueue.global(qos: .background)
+        let channel = DispatchIO(type: .stream, fileDescriptor: fileDescriptor, queue: global) { _ in }
+
+        // The desire is to trigger the handler blocka as soon as there is any data in the channel.
+        channel.setLimit(lowWater: 1)
+
+        channel.read(offset: 0, length: Int.max, queue: global) { (_, _, error) in
+            var notification = Notification(name: .NSFileHandleDataAvailable, object: nil, userInfo: nil)
+            if error != 0 {
+                notification.userInfo = ["NSFileHandleError": NSNumber(value: error)]
+            }
+            NotificationCenter.default.post(notification)
+        }
     }
     
     open var readabilityHandler: ((FileHandle) -> Void)? {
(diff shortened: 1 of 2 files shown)


Foundation/FileHandle.swift
```diff
@@ -345,7 +346,19 @@ extension FileHandle {
...
 
     open func waitForDataInBackgroundAndNotify() {
-        NSUnimplemented()
+        let global = DispatchQueue.global(qos: .background)
+        let channel = DispatchIO(type: .stream, fileDescriptor: fileDescriptor, queue: global) { _ in }
+
+        // The desire is to trigger the handler blocka as soon as there is any data in the channel.
+        channel.setLimit(lowWater: 1)
+
+        channel.read(offset: 0, length: Int.max, queue: global) { (_, _, error) in
```
weissi: - this reads all the data for no reason
- also if this is a socket/pipe this is a permanent leak of a file descriptor
 - you could just use a `DispatchSource.makeReadSource` and get the notifications but not the reading. As soon as you get the first notification you then want to `.cancel` the `DispatchSource` as you only want to get notified once.

Foundation/FileHandle.swift
```diff
@@ -345,7 +346,19 @@ extension FileHandle {
...
+            var notification = Notification(name: .NSFileHandleDataAvailable, object: nil, userInfo: nil)
+            if error != 0 {
+                notification.userInfo = ["NSFileHandleError": NSNumber(value: error)]
+            }
+            NotificationCenter.default.post(notification)
```
weissi: doesn't that need to happen on the main thread? CC @parkera

millenomi:
@RLovelett are you still interested in chaperoning this patch?

RLovelett:
@millenomi I am interested in finishing this. I might need some pointers on how to get continue to move forward. But I am interested it getting this implemented (that way I can stop maintaining my own private fork with this patch).

shahmishal:

PR #1428, opened 2018-02-09
Sync with the Swift repository on the _silgen_name thing

jrose-apple:
That's apple/swift#14484. This only affects the Apple-platform implementation anyway.


jrose-apple:
@swift-ci Please test

PR #1429, opened 2018-02-10
Fixed `addObserver`'s obsolescence

felix91gr:
As per [this post](https://forums.swift.org/t/found-a-bug-in-foundations-obsolescence-mechanism-in-linux/7730) in the forums.

Fixed with Xiaodi's suggestion.
It passed `ninja test` in my computer and it fixed the problem I described after I rebuilt my binaries with this patch.

I think we need to run a `ninja test` in CI, just to be sure. After that, it should be ready for merge.

### DIFF
--- a/Foundation/NSNotification.swift
+++ b/Foundation/NSNotification.swift
@@ -188,8 +188,8 @@ open class NotificationCenter: NSObject {
         })
     }
 
-    @available(*, unavailable, renamed: "addObserver(forName:object:queue:using:)")
-    open func addObserver(forName name: NSNotification.Name?, object obj: Any?, queue: OperationQueue?, usingBlock block: @escaping (Notification) -> Void) -> NSObjectProtocol {
+    @available(*,obsoleted:4.0,renamed:"addObserver(forName:object:queue:using:)")
+    open func addObserver(forName name: NSNotification.Name?, object obj: Any?, queue: OperationQueue?, usingBlock block: @escaping (Notification) -> Void, _: Void = ()) -> NSObjectProtocol {
         return addObserver(forName: name, object: obj, queue: queue, using: block)
     }
 


spevans:
@swift-ci please test

ikesyo:
This would revert the changes in https://github.com/apple/swift-corelibs-foundation/pull/1323. Please reconsider this. I think `@available(swift,` should be used here instead of `@available(*,`.

ikesyo:
See also https://bugs.swift.org/browse/SR-6419

cc @parkera @alblue @ianpartridge

ikesyo:
I added a reply on that thread: https://forums.swift.org/t/found-a-bug-in-foundations-obsolescence-mechanism-in-linux/7730/14?u=ikesyo

felix91gr:
Oh. There was already context for this! Nice.

Let’s see... hmm. Did #1323 solve SR-6419?

The problem this tries to fix is that the function is still ambiguous, at least in Linux (which is my setup).

Quoting Xiaodi,
> An immediate source-compatible fix would be to disable trailing closure syntax selectively for the obsolete function by tacking on a defaulted Void parameter. That way, any explicit use of the obsolete function still works and prompts the warning, and any use of trailing closure syntax becomes unambiguous.

If I understood them correctly, it should suffice to keep #1323’s obsolescence but change the function’s signature. Is that correct?

ikesyo:
> The problem this tries to fix is that the function is still ambiguous, at least in Linux (which is my setup).

The issue is in Swift 4.0.2 and 4.0.3 on Linux (and not in 4.0). #1323 is included in recent 4.1-DEVELOPMENT-SNAPSHOT builds. Did you try building your code with the snapshots?

felix91gr:
> Did you try building your code with the snapshots?

I tried, but I did it wrong. I can try it again, now that I’ve remembered how to change my toolchains.

Gimme a day. 

(I’m going to be without internet starting from tomorrow, so if I don’t come back soon-ish, you can expect me to appear again around the 17th)

felix91gr:
You were right all along. I tested the 4.1 snapshots and the bug was long gone! Thank you :)

PR #1431, opened 2018-02-11
Added userName and fullUserName to ProcessInfo

pvieito:
- Added ProcessInfo.userName and ProcessInfo.fullUserName as in Darwin Foundation
- Added NSFullUserName() as in Darwin Foundation
- Added underlying implementation in Core Foundation CFCopyFullUserName() for POSIX OS.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -284,8 +284,10 @@ CF_EXPORT CFStringRef CFCopyUserName(void) {
 #else
 #error Dont know how to compute user name on this platform
 #endif
-    if (!result)
+    if (!result) {
         result = (CFStringRef)CFRetain(CFSTR(""));
+    }
+    
     return result;
 }
 
@@ -301,8 +303,9 @@ CF_EXPORT CFStringRef CFCopyFullUserName(void) {
 #else
 #error Don't know how to compute full user name on this platform
 #endif
-    if (!result)
+    if (!result) {
         result = (CFStringRef)CFRetain(CFSTR(""));
+    }
     
     return result;
 }
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -289,6 +289,24 @@ CF_EXPORT CFStringRef CFCopyUserName(void) {
     return result;
 }
 
+CF_EXPORT CFStringRef CFCopyFullUserName(void) {
+    CFStringRef result = NULL;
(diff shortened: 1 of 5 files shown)


Foundation/NSPathUtilities.swift
```diff
@@ -588,6 +588,10 @@ public func NSUserName() -> String {
     return userName._swiftObject
 }
 
+public func NSFullUserName() -> String {
+    NSUnimplemented()
+}
```
spevans: Could you use something like the following to extract the gecos info?

```
public func NSFullUserName() -> String {
    let euid = _CFGetEUID()
    let uid = euid != 0 ? euid : getuid()
    var pwd = passwd()
    let maxSize = sysconf(Int32(_SC_GETPW_R_SIZE_MAX))
    let buf = UnsafeMutablePointer<Int8>.allocate(capacity: maxSize)
    defer { buf.deallocate() }
    var ptr: UnsafeMutablePointer<passwd>? = nil
    if getpwuid_r(uid, &pwd, buf, maxSize, &ptr) == 0 {
        if let gecos = pwd.pw_gecos {
            return String(cString: gecos)
        }
    }

    return ""
}
```

pvieito: I would prefer to leave the NSFullUserName implementation for other Pull Request/Contributor.

pvieito: Thinking about this, shouldn't this code be implemented in Core Foundation in a function similar to CFCopyUserName and made portable between multiple platforms?

spevans: I would probably keep it in Swift and only add it into CF if the multi-platform support became too messy and a lot easier to write using the C preprocessor.

pvieito:
Added NSFullUserName() and underlying implementation in CFCopyFullUserName() for POSIX OS.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -289,6 +289,24 @@ CF_EXPORT CFStringRef CFCopyUserName(void) {
...
+    __CFGetUGIDs(&euid, NULL);
+    struct passwd *upwd = getpwuid(euid ? euid : getuid());
+    if (upwd && upwd->pw_gecos) {
+        result = CFStringCreateWithCString(kCFAllocatorSystemDefault, upwd->pw_gecos, kCFPlatformInterfaceStringEncoding);
+    }
+#else
+#error Don't know how to compute full user name on this platform
+#endif
+    if (!result)
+        result = (CFStringRef)CFRetain(CFSTR(""));
```
parkera: Style nit here: please always use `{ }` or make one line for `if` statements.

parkera:
Other than minor style point, looks good to me.

pvieito:
Updated style with braces on if statements.

parkera:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #1432, opened 2018-02-14
CountableRange is deprecated, renamed to Range

spevans:
- Replace NSRange.toCountableRange() with Range.init?(_:)

- Update Test cases.

### DIFF
--- a/Foundation/NSArray.swift
+++ b/Foundation/NSArray.swift
@@ -264,7 +264,7 @@ open class NSArray : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCo
             return
         }
         
-        objects += range.toCountableRange()!.map { self[$0] }
+        objects += Range(range)!.map { self[$0] }
     }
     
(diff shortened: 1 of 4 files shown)


spevans:
@swift-ci please test

parkera:
@airspeedswift does this line up with the changes you've been working on here too?

airspeedswift:
Please don't make this change. I already made it originally and had to revert it, because we don't have a 5.0 branch for corelibs-foundation, so Linux testing of the 5.0 swift branch started failing.

That said, it is a good point that the `toCountableRange` method should be replaced by just a `Range.init`. But that'll break stuff too so please hold off for now. 5.0 branch will rebranch soon, then we can do it.

spevans:
@swift-ci please test

airspeedswift:
Now that the conditionally-conforming `Range` is available on 4.2, I think this is gtm.

spevans:

PR #1436, opened 2018-02-16
 Fixed prefix path for FHS Installed Bundles

pvieito:
The initial implementation was using the `share` directory as the
prefix path.

### DIFF
--- a/TestFoundation/TestBundle.swift
+++ b/TestFoundation/TestBundle.swift
@@ -220,7 +220,7 @@ class BundlePlayground {
         self.playgroundPath = nil
         
         do {
-            //try FileManager.default.removeItem(atPath: path)
+            try FileManager.default.removeItem(atPath: path)
         } catch _ {
             // ¯\_(ツ)_/¯ We did what we could.
         }
--- a/TestFoundation/TestBundle.swift
+++ b/TestFoundation/TestBundle.swift
@@ -413,6 +413,8 @@ class TestBundle : XCTestCase {
         
         try! _withEachPlaygroundLayout { (playground) in
             let bundle = Bundle(path: playground.bundlePath)!
+            
+            // Must throw as the main executable is a dummy empty file.
             XCTAssertThrowsError(try bundle.preflight())
         }
     }
--- a/TestFoundation/TestBundle.swift
+++ b/TestFoundation/TestBundle.swift
@@ -243,7 +243,8 @@ class TestBundle : XCTestCase {
             ("test_bundleLoad", test_bundleLoad),
             ("test_bundleLoadWithError", test_bundleLoadWithError),
             ("test_bundleWithInvalidPath", test_bundleWithInvalidPath),
-            ("test_bundleFindExecutables", test_bundleFindExecutables),
+            ("test_bundlePreflight", test_bundlePreflight),
+            ("test_bundleFindExecutable", test_bundleFindExecutable),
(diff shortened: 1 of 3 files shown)


pvieito:
Executable search order for bundle at “Bundles/Darwin/InstalledBundles/share/Bundle.resources”:

Darwin/InstalledBundles/bin/Bundle
Darwin/InstalledBundles/bin/libBundle.dylib
Darwin/InstalledBundles/sbin/Bundle
Darwin/InstalledBundles/sbin/libBundle.dylib
Darwin/InstalledBundles/lib/Bundle
Darwin/InstalledBundles/lib/libBundle.dylib
Darwin/InstalledBundles/libexec/Bundle.executables/MacOS/Bundle
Darwin/InstalledBundles/libexec/Bundle.executables/MacOS/libBundle.dylib
Darwin/InstalledBundles/libexec/Bundle.executables/Bundle
Darwin/InstalledBundles/libexec/Bundle.executables/libBundle.dylib
Darwin/InstalledBundles/share/Bundle.resources/Bundle
Darwin/InstalledBundles/share/Bundle.resources/libBundle.dylib

millenomi:
Can you check that the bundle tests are correct also?

pvieito:
@millenomi Yes, they seem correct.

millenomi:
I'm a little confused — why is it that they're passing if our search path was incorrect? Can we shore them up as part of this?

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

pvieito:
@millenomi I found an error in the tests that was allowing the bug in the initial implementation. In [TestBundle.swift:406](https://github.com/apple/swift-corelibs-foundation/blob/458977b5bc8911832e3bdacb0af9c2b3bc676e1f/TestFoundation/TestBundle.swift#L406) it is using `XCTAssertThrowsError` instead of `XCTAssertNoThrow`. I will add a new commit to the Pull Request with an updated test.

TestFoundation/TestBundle.swift
```diff
@@ -234,41 +243,41 @@ class TestBundle : XCTestCase {
             ("test_bundleLoad", test_bundleLoad),
             ("test_bundleLoadWithError", test_bundleLoadWithError),
             ("test_bundleWithInvalidPath", test_bundleWithInvalidPath),
-            ("test_bundlePreflight", test_bundlePreflight),
```
millenomi: Why was this test removed?

pvieito: I changed the name of the test function to `test_bundleFindExecutables`. For the `testBundle()` the test will try a `Bundle.preflight()` which must success and for the `withEachPlaygroundLayout` bundles the test will try to locate the executable (trying to `preflight()` them will always fail because the executable are empty file).

alblue: The `test_bundleFindExecutables` does not appear to be added here, though.

pvieito: Yes, it does.

millenomi: That makes more sense, but I'm not quite sure why we're not testing that invalid executables can't be preflighted anymore.

pvieito:
@millenomi Ok, I have implemented the `test_bundleFindExecutable` as a separated test function and re-added the original `test_bundlePreflight`.

alblue:
Can we squash all of these commits into one and rebase onto current master? There are changes in 3 files but spread over 8 commits, which is going to cause issues if trying to git bisect issues later on.

PR #1440, opened 2018-02-17
closure: make it build on Windows x86_64

compnerd:
Windows x86_64 is a LLP64 environment.  `unsigned long` is not `size_t`.
Adjust the parameter to be the correct size for the allocator.  This is
compatible with the previous implementation on LP64 environments.

Prefer to use the `__sync_bool_compare_and_swap` over the MSVC
intrinsic which is overly strong.  While in the area, remove an unused
compare and swap method.

### DIFF
--- a/closure/runtime.c
+++ b/closure/runtime.c
@@ -25,25 +25,22 @@
 #define os_assert(_x) assert(_x)
 #endif
 
-#if TARGET_OS_WIN32
-#define _CRT_SECURE_NO_WARNINGS 1
-#include <windows.h>
-static __inline bool OSAtomicCompareAndSwapLong(long oldl, long newl, long volatile *dst) 
-{ 
-    // fixme barrier is overkill -- see objc-os.h
-    long original = InterlockedCompareExchange(dst, newl, oldl);
-    return (original == oldl);
-}
+#if !defined(__has_builtin)
+#define __has_builtin(builtin) 0
+#endif
 
-static __inline bool OSAtomicCompareAndSwapInt(int oldi, int newi, int volatile *dst) 
-{ 
-    // fixme barrier is overkill -- see objc-os.h
-    int original = InterlockedCompareExchange(dst, newi, oldi);
-    return (original == oldi);
+#if __has_builtin(__sync_bool_compare_and_swap)
+#define OSAtomicCompareAndSwapInt(Old, New, Ptr)                               \
+  __sync_bool_compare_and_swap(Ptr, Old, New)
+#elif TARGET_OS_WIN32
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @parkera

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

The test failure seemed unrelated

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

parkera:
Seems reasonable. I thought we were actually depending on the swift-corelibs-libdispatch version of this file now, though.

closure/runtime.c
```diff
@@ -142,7 +139,7 @@ GC support stub routines
 
 
 
-static void *_Block_alloc_default(const unsigned long size, const bool initialCountIsOne, const bool isObject) {
+static void *_Block_alloc_default(uintptr_t size, const bool initialCountIsOne, const bool isObject) {
```
parkera: Shouldn't this still be size_t though? I can see using the ptr types for the CAS.

compnerd:
Hmm, actually, yeah, the `size_t` probably is better.  Hmm, using the version that is in libdispatch is fine, but, we should then have a single copy of the sources.  Would it be okay to remove these then?

compnerd:
Switched to `size_t`.  I took a look at the `build.py`.  Seems that if `LIBDISPATCH_SOURCE_DIR` is not specified, then it will fall back to this source.  Is it really possible to build CoreFoundation/Foundation without libdispatch at this point (i.e. is that just dead code)?

compnerd:
@swift-ci please test

parkera:
@compnerd I have been wondering that myself recently as well. We did need this during bringup (dispatch was not yet ready), and I wonder if it is still useful and still working for other platforms that are in that stage.

parkera:
Like Windows 64 bit, apparently. :)

compnerd:
@swift-ci please test

alblue:
@swift-ci please test

compnerd:
That failure is really weird, its not at all related to this change, and the crasher there doesn't even reference Foundation.

shahmishal:
@swift-ci please test

spevans:
@swift-ci please test

compnerd:
@spevans guess it likes you better than the rest of us :-p.  @parkera good to merge?  I have the mirror'ed commit in libdispatch up as well so that we don't drift apart until we can unify the codebase.

parkera:
Yup, ok.

PR #1451, opened 2018-02-21
[4.1] SR-7017: JSONDecoder will decode booleans as numbers.

spevans:
- If an NSNumber was initialised from a boolean then dont allow it
  to be converted to a numeric value.

- Use same boolean check as Foundation overlay method.


spevans:
@swift-ci please test

spevans:

PR #1452, opened 2018-02-22
Add support for custom JSON values

neon12345:
A small change to show how custom values in JSON data could be supported by the JSON decoder.

For example one could use

struct MyData : Codable {
      let test : String
}

with

{"test":<"my""strange""data">}

and easily get 

"\"my\"\"strange\"\"data\"\"

into test. 

This can be used to bypass JSON validation and decode for some data e.g. to have JSON as String inside of JSON without double quotes escape madness.

### DIFF
--- a/Foundation/JSONEncoder.swift
+++ b/Foundation/JSONEncoder.swift
@@ -900,6 +900,12 @@ open class JSONDecoder {
         let decoder = _JSONDecoder(referencing: topLevel, options: self.options)
         return try T(from: decoder)
     }
+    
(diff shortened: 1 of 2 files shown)


ianpartridge:
As this changes public API, we would need matching changes in the Darwin implementation of Foundation, which is not part of this project. @parkera can comment on whether that's likely 🙂

parkera:
@neon12345 thanks for the idea. Let's discuss it on the [Swift forums](https://forums.swift.org) before proposing a source change here.

xwu:
@neon12345 Discussion on custom encoders and decoders is occurring in several threads on the Swift forums, including:

https://forums.swift.org/t/pitch-export-jsonencoder-jsondecoder/6945
https://forums.swift.org/t/writing-new-encoders-and-decoders/10207

PR #1453, opened 2018-02-22
Add icu-config invocation to linux LDFLAGS to link the correct lib

lanza:
This change simply adds an invocation of `icu-config` to include the proper `searchdir` for icu.


lanza:
@compnerd

build.py
```diff
@@ -16,7 +16,7 @@
 swift_cflags = ['-DDEPLOYMENT_RUNTIME_SWIFT']
 if Configuration.current.target.sdk == OSType.Linux:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE -DCF_CHARACTERSET_DATA_DIR="CoreFoundation/CharacterSets"'
-	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc -Wl,-Bsymbolic '
+	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc -Wl,-Bsymbolic `icu-config --ldflags`'
```
compnerd: Where was ICU being linked previously?  I think that we should ensure that that area is updated.

PR #1454, opened 2018-02-23
Add icu-config invocation to linux LDFLAGS to link the correct lib

lanza:
This change simply adds an invocation of icu-config to include the proper searchdir for icu.

### DIFF
--- a/build.py
+++ b/build.py
@@ -16,7 +16,7 @@ foundation.GCC_PREFIX_HEADER = 'CoreFoundation/Base.subproj/CoreFoundation_Prefi
 swift_cflags = ['-DDEPLOYMENT_RUNTIME_SWIFT']
 if Configuration.current.target.sdk == OSType.Linux:
 	foundation.CFLAGS = '-DDEPLOYMENT_TARGET_LINUX -D_GNU_SOURCE -DCF_CHARACTERSET_DATA_DIR="CoreFoundation/CharacterSets"'
-	foundation.LDFLAGS = '${SWIFT_USE_LINKER} -Wl,@./CoreFoundation/linux.ld -lswiftGlibc -Wl,-Bsymbolic '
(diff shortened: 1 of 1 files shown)


lanza:
@compnerd

spevans:
Whats the advantage of using `icu-config` over `pkg-config` ? The latter allowed ICU to be compiled as part of the swift build and the build path passed through to the foundation build.

parkera:
I'm curious if this builds, but I have the same question as @spevans

parkera:
@swift-ci please test

lanza:
Ah, I was not aware you could build ICU as part of Swift. I'll look into that.

The advantage was in working with `/usr/local/lib` installs of icu4c (in particular - working with CentOS). The system package manager installs version 50 (which is outdated) to `/usr/lib` and invoking `pkg-config` simply provides `-licuXYZ` on both Ubuntu and CentOS. Since `icu-devtools` and `libicu-dev` are both prerequisites for building on Ubuntu `icu-config` generates the same `-l` flags as the `pig-config` version as well as the proper `-L` flag for both systems (not accounting for the ability to build icu as part of Swift).

spevans:
@swift-ci please test

spevans:
@lanza I think you would also need to update the main swift build as well since that also links against ICU. It could be compiling the runtime and stdlib against the `/usr/lib` version whilst compiling Foundation against `/usr/local/lib`. This can happen because ICU can be built to use versioned symbols eg `ucol_55` v `ucol_57` so you may actually be linking different parts of swift against different libraries.

However on a related note, I have had success with building swift on Linux against Apple's fork of ICU so it may well be the case that in the future Swift is simply built against this version for all platforms which fixes a variety of bugs and keeps every platform on the same version and the `pkg-config` override is removed from the build completely.

lanza:
@spevans Yes, sorry this is not

PR #1459, opened 2018-03-02
Remove hasPrefix/Suffix as they are now on String for Linux 

airspeedswift:
Needs coordination with apple/swift#14390, which introduces the `String` versions.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -1479,36 +1479,6 @@ extension String : _NSBridgeable, _CFBridgeable {
     internal var _cfObject: CFType { return _nsObject._cfObject }
 }
 
-#if !(os(OSX) || os(iOS))
-extension String {
-    public func hasPrefix(_ prefix: String) -> Bool {
-        if prefix.isEmpty {
-            return true
-        }
-
-        let cfstring = self._cfObject
-        let range = CFRangeMake(0, CFStringGetLength(cfstring))
-        let opts = CFStringCompareFlags(
-            kCFCompareAnchored | kCFCompareNonliteral)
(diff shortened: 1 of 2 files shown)


airspeedswift:
Test with https://github.com/apple/swift/pull/14390

@swift-ci please test

spevans:
@airspeedswift `Foundation/NSString.swift` has `hasPrefix/hasSuffix` extensions for both `NSString` and `String`. I think the wrong pair got removed here.

airspeedswift:
@spevans yup/oops

airspeedswift:
Test with apple/swift#14390

@swift-ci please test

airspeedswift:
Test with apple/swift#14390

@swift-ci please test

airspeedswift:
Test with apple/swift#14390

@swift-ci please test

airspeedswift:
@parkera OK with you to merge this? I'll merge the stdlib change onto the 5.0 branch too to keep the 5.0 bots working.

parkera:
Does this really have the same behavior? The previous code used `kCFCompareNonliteral` ("If specified, loose equivalence is performed (o-umlaut == o, umlaut)").

parkera:
If so perhaps we should encapsulate that behavior in a unit test just so we have it documented as behaving as expected.

airspeedswift:
@parkera Yes, this will consider the two compositions of `ö` the same, since it uses `Character.==` which has that behavior. The tests on the stdlib version are pretty extensive already for this and cover lots of different compositions.

airspeedswift:
(and the Linux version will now be covered by those tests too)

parkera:
Ok

PR #1460, opened 2018-03-02
Changes done to align NSURL.description with Darwin code

saiHemak:
On Linux , the URL.description seems to be incomplete with respect to Darwin, when the URL has a component string associated with it .


alblue:
@swift-ci please test

PR #1461, opened 2018-03-04
Added modern URL API to Foundation.Process

pvieito:
- Deprecated old path API as Darwin Foundation

### DIFF
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -164,12 +164,39 @@ open class Process: NSObject {
     
     }
     
-    // these methods can only be set before a launch
-    open var launchPath: String?
-    open var arguments: [String]?
-    open var environment: [String : String]? // if not set, use current
+    @available(*, deprecated)
+    open var launchPath: String? {
+        set {
+            guard let newValue = newValue else {
+                self.executableURL = nil
+                return
+            }
+            
+            self.executableURL = URL(fileURLWithPath: newValue)
+        }
+        
+        get {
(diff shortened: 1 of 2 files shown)


Foundation/Process.swift
```diff
@@ -164,13 +164,40 @@ open class Process: NSObject {
     
     }
     
+    @available(*, deprecated)
```
parkera: These are actually not marked as deprecated yet in the SDK.

parkera: The new implementation looks reasonable but we can leave the deprecation annotation off for now.

pvieito: It is marked as deprecated in the documentation: https://developer.apple.com/documentation/foundation/process/1413110-currentdirectorypath

Maybe it is just an error.

pvieito: If the path API is not deprecated I would prefer to do the inverted implementation (just add the new URL API as an overlay of the path counterparts).

parkera: The documentation actually got ahead of us here. It is "soft deprecated", indicating that will eventually be deprecated.

It may seem like splitting hairs but I think we should remain consistent with the annotations in the SDK. Once the SDK marks it deprecated (with a `API_DEPRECATED` macro, e.g.) then we can add the attribute here.

If you want to invert the implementation, that's fine with me. That's actually how we do it in the ObjC Foundation too.

pvieito:
Final request: #1462

PR #1462, opened 2018-03-06
Added URL API to Foundation.Process

pvieito:
Refs #1461

### DIFF
--- a/Foundation/NSLocale.swift
+++ b/Foundation/NSLocale.swift
@@ -84,11 +84,11 @@ open class NSLocale: NSObject, NSCopying, NSSecureCoding {
 }
 
 extension NSLocale {
-    open class var current: Locale {
+    open static var current: Locale {
         return CFLocaleCopyCurrent()._swiftObject
     }
     
-    open class var system: Locale {
+    open static var system: Locale {
         return CFLocaleGetSystem()._swiftObject
     }
 }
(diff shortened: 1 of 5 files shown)


pvieito:
I've also added a small fix to Locale static properties that produces crashes on Linux by bridging String to NSString.

pushkarnk:
@swift-ci please test

alblue:
@swift-ci please test

pvieito:
@swift-ci please test

Foundation/NSLocale.swift
```diff
@@ -84,11 +84,11 @@ open class NSLocale: NSObject, NSCopying, NSSecureCoding {
 }
 
 extension NSLocale {
-    open class var current: Locale {
+    open static var current: Locale {
```
millenomi: Why `static`? This doesn't match the Darwin signature:

```swift
class var current: Locale { get }
```

Foundation/NSLocale.swift
```diff
@@ -84,11 +84,11 @@ open class NSLocale: NSObject, NSCopying, NSSecureCoding {
...
         return CFLocaleCopyCurrent()._swiftObject
     }
     
-    open class var system: Locale {
+    open static var system: Locale {
```
millenomi: As above for all `static` members. Please revert?

pvieito:
Ups, sorry, didn't think about that. I will resend two separated pull requests.

PR #1464, opened 2018-03-09
Fix for NSKeyedArchiver.archiveRootObject api failure on Linux

mamabusi:
Test case:
```
        let filePath = NSTemporaryDirectory() + "testdir\(NSUUID().uuidString)"
        let result = NSKeyedArchiver.archiveRootObject("Hello", toFile: filePath)
        print(result)
```

Expected Result:  `true`
Actual result on Linux: `false`
Actual result on Darwin: `true`

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -590,7 +590,7 @@ public func NSFullUserName() -> String {
 }
 
 internal func _NSCreateTemporaryFile(_ filePath: String) throws -> (Int32, String) {
-    let template = "." + filePath + ".tmp.XXXXXX"
+    let template = filePath + ".tmp.XXXXXX"
     let maxLength = Int(PATH_MAX) + 1
     var buf = [Int8](repeating: 0, count: maxLength)
     let _ = template._nsObject.getFileSystemRepresentation(&buf, maxLength: maxLength)
--- a/TestFoundation/TestNSKeyedArchiver.swift
+++ b/TestFoundation/TestNSKeyedArchiver.swift
@@ -107,6 +107,8 @@ class TestNSKeyedArchiver : XCTestCase {
             ("test_archive_uuid_bvref", test_archive_uuid_byref),
             ("test_archive_uuid_byvalue", test_archive_uuid_byvalue),
             ("test_archive_unhashable", test_archive_unhashable),
+            ("test_archiveRootObject_String", test_archiveRootObject_String),
+            ("test_archiveRootObject_URLRequest()", test_archiveRootObject_URLRequest),
         ]
     }
 
@@ -345,4 +347,29 @@ class TestNSKeyedArchiver : XCTestCase {
             XCTFail("test_archive_unhashable, de-serialization error \(error)")
         }
     }
+
+    func test_archiveRootObject_String() {
+        let filePath = NSTemporaryDirectory() + "testdir\(NSUUID().uuidString)"
+        let result = NSKeyedArchiver.archiveRootObject("Hello", toFile: filePath)
+        XCTAssertTrue(result)
+        do {
+            try FileManager.default.removeItem(atPath: filePath)
+        } catch {
+            XCTFail("Failed to clean up file")
+        }
+    }
+
+    func test_archiveRootObject_URLRequest() {
+        let filePath = NSTemporaryDirectory() + "testdir\(NSUUID().uuidString)"
+        let url = URL(string: "http://swift.org")!
+        let request = URLRequest(url: url)._bridgeToObjectiveC()
+        let result = NSKeyedArchiver.archiveRootObject(request, toFile: filePath)
+        XCTAssertTrue(result)
+        do {
+            try FileManager.default.removeItem(atPath: filePath)
+        } catch {
+            XCTFail("Failed to clean up file")
+        }
+    }
+
 }


Foundation/NSKeyedArchiver.swift
```diff
@@ -132,20 +132,21 @@ open class NSKeyedArchiver : NSCoder {
     ///   - path:       The path of the file in which to write the archive.
     /// - Returns:      `true` if the operation was successful, otherwise `false`.
     open class func archiveRootObject(_ rootObject: Any, toFile path: String) -> Bool {
+        var filePath = "/" + path
```
mamabusi: Path input like "Users/sdsfd/test.txt" is accepted on Darwin and a "/" is prepended to the path. Hence, prepending the path since a double "/" will not impact.

Foundation/NSPathUtilities.swift
```diff
@@ -590,7 +590,7 @@ public func NSFullUserName() -> String {
 }
 
 internal func _NSCreateTemporaryFile(_ filePath: String) throws -> (Int32, String) {
-    let template = "." + filePath + ".tmp.XXXXXX"
```
mamabusi: With the "." in place, the `fd` returned by system library call mkstemp() is always `-1` and the temporary file creation always fails. Also, since the file path sent to  _NSCreateTemporaryFile() is absolute, making it relative here does not seems right.

ianpartridge:
@swift-ci please test

spevans:
@swift-ci  please test

pushkarnk:
The failure doesn't appear to be related to the change. 
`07:26:54 /home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux@2/Ninja-ReleaseAssert/swiftpm-linux-x86_64/x86_64-unknown-linux/release/swift-test: error while loading shared libraries: libswiftSwiftOnoneSupport.so: cannot open shared object file: No such file or directory`

I'm firing up another test, just in case we have some intermittent issue at hand.

@swift-ci please test

pushkarnk:
@swift-ci please test

Foundation/NSKeyedArchiver.swift
ianpartridge: Is this the only normalisation that Darwin does? It seems unlikely to me... I would guess that there is more to it than this. Also, why not only prepend if required - allowing a double // seems a bit hacky.

parkera:
cc @itaiferber

mamabusi:
@itaiferber Could you please review this PR.

itaiferber:
@mamabusi I must've missed this; looking now.

Foundation/NSKeyedArchiver.swift
itaiferber: I don't think prepending with `/` is valid here. If I try to write to a relative path (say, `"./my-archive"`), we'd be making it absolute (`"/./my-archive"`), which is not correct.

Why is the current behavior a failure on Linux? What paths are failing? Note that in Darwin Foundation, we don't modify this path at all; the underlying writing code validates the path and adds a `/` if necessary.

Foundation/NSPathUtilities.swift
itaiferber: The file path to `_NSCreateTemporaryFile` should not need to be absolute; we're indeed using it as a template to create a new file, so we should be able to make whatever modifications necessary internally to make it work. What path are you seeing failure for? And what's the error? We should fix the underlying problem, not the call sites which use this.

mamabusi:
@itaiferber Thanks for taking time and looking into this PR.  
Below are more details in response to the review:

`internal func _NSCreateTemporaryFile(_ filePath: String) throws -> (Int32, String)`

The above internal function is called only from the function `NSKeyedArchiver.archiveRootObject` in the whole of SCLF. There are no other existing call sites to `_NSCreateTemporaryFile`

Here’s  the result of `NSTemporaryDirectory() + "testdir\(NSUUID().uuidString)"` on Linux box:

`/tmp/testdir60CABFCC-3F02-4EC7-90C0-42F4E1C8F55F`

The above result looks fine and there are no issues there. 

When I pass the resulting filePath `/tmp/testdir60CABFCC-3F02-4EC7-90C0-42F4E1C8F55F` to 
`NSKeyedArchiver.archiveRootObject("Hello", toFile: filePath)` , it fails to write to the file. 

What is happening internally is: 
When  `_NSCreateTemporaryFile(“/tmp/testdir60CABFCC-3F02-4EC7-90C0-42F4E1C8F55F”) ` is called, it prepends `"."` to the path.

Now the path looks like this:
`./tmp/testdir60CABFCC-3F02-4EC7-90C0-42F4E1C8F55F`

 and then when the system call “mkstemp(&buf)” is made, the fd returned is always “-1”  causing a failure in temporary file creation. 

Is there a reason why the function ` _NSCreateTemporaryFile()`  has to convert the given path to a relative path from the current directory?

itaiferber:
@mamabusi I agree in that I don't think that it should make the path relative; it doesn't do that on the Darwin side AFAICT. I think it's worth trying to figure out from the git history why (if for any reason) it was there in the first place, especially if `NSKeyedArchiver` is currently the only remaining place using it. It seems that the solution here could be as simple as just removing the leading `"." +` from `_NSCreateTemporaryFile`, without modifying `NSKeyedArchiver` itself.

mamabusi:
@itaiferber 
Here’s what I got from the history: 
The internal function `_NSCreateTemporaryFile()` was introduced as part of the commit https://github.com/apple/swift-corelibs-foundation/commit/b4d207eaf0783411b1ed1484d4a82e8aebe11d1c#diff-c4535e3015bf3d52374e84bcaf592d95 that maps to the SR: https://bugs.swift.org/browse/SR-432 which tells us that this internal api was created to help the implementation of the `NSKeyedArchiver` apis. The implementation has not changed since then (not sure about the reason for "." prefixed to the `filepath`). Also, there is no test-case in the Foundation test-suite that covers `NSKeyedArchiver.archiveRootObject()` api  which uses the internal api `_NSCreateTemporaryFile()` and hence this bug is not caught. 

On the other aspect, the reason I decided to prefix "/" to the `filePath` in `NSKeyedArchiver` was because the below test-cases pass on Darwin platform:

```
let result = NSKeyedArchiver.archiveRootObject("Hello", toFile: "tmp/test.txt")
let result = NSKeyedArchiver.archiveRootObject("Hello", toFile: "Users/mamatha/Desktop/../Documents/test.txt")

```
Without adding the prefix of "/" , the above test-cases fail on Linux. 

This test-case: 
`let result = NSKeyedArchiver.archiveRootObject("Hello", toFile: "./../Documents/dd")`

fails on both Darwin and Linux. The behaviour is the same even if Linux prefixes "/", i.e. we are not deviating from any of the behaviour of Darwin for all the above test-cases. 

If there is something more I need to consider, please let me know.

itaiferber:
@mamabusi Thanks for taking the time to uncover the history of this! It sounds like removing the preceding `"." +` from `_NSCreateTemporaryFile` makes sense, so I'm definitely on board with that.

Regarding prepending `/`: I don't know how those tests pass on Darwin. When you create a path that is not rooted in `/` to begin with, the path you have is relative. Archiving to that path, then, depends on the current working directory. Consider the following:

```swift
import Foundation

let path = "Users/itai/Desktop/../Documents/test.txt"
let url = URL(fileURLWithPath: path)
print(url.path)

if NSKeyedArchiver.archiveRootObject("Hello, world!", toFile: path) {
    print("Write succeeded")
} else {
    print("Write failed")
}
```

When I run the above from my home directory (`cd ~; swift Test.swift`), the output is

```
/Users/itai/Users/itai/Documents/test.txt
Write failed
```

When I run it from `/` (`cd /; swift ~/Test.swift`), the output is

```
/Users/itai/Documents/test.txt
Write succeeded
```

In fact, that can only succeed when run when the current working directory is `/`; the path given to `NSKeyedArchiver` is turned into a `URL` for writing exactly as above (`NSURL *fileURL = [NSURL fileURLWithPath:tempFilePath isDirectory:NO];`). That path is then used to create a stream in the same way, and the stream resolves the `URL` as all relative `URL`s are resolved. Those tests should pass only if the current working directory is `/`.

Given that, I _don't_ think that making the path absolute by forcibly rooting it against `/` is correct; I think the current behavior is correct. If the write fails, the path in the test should probably be updated.

mamabusi:
@itaiferber  Thank you! I understand and agree with you on "not making the path absolute". Have made the necessary changes on the PR.

alblue:
@swift-ci please test

itaiferber:
@mamabusi Thanks for taking the time to fix this up! LGTM

pushkarnk:

pushkarnk:

PR #1468, opened 2018-03-11
Fix and reenable tests that depend on specific hash values

lorentey:
Hash values are now randomly seeded, and the hash algorithm may change between any two releases of the standard library — code should not depend on specific hash values or Set/Dictionary orderings.


lorentey:
@swift-ci test

PR #1476, opened 2018-03-14
[string] Wean off of non-String-API.

milseman:
String._fromCodeUnits is about to disappear; use already existing
String.init calls instead.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -273,7 +273,7 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
     }
     
     public init(characters: UnsafePointer<unichar>, length: Int) {
-        _storage = String._fromWellFormedCodeUnitSequence(UTF16.self, input: UnsafeBufferPointer(start: characters, count: length))
+        _storage = String.init(decoding: UnsafeBufferPointer(start: characters, count: length), as: UTF16.self)
     }
     
     public required convenience init(unicodeScalarLiteral value: StaticString) {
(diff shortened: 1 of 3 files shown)


milseman:

milseman:
@phausler who is the right person to review this change?

milseman:
@itaiferber can you review?

Foundation/NSString.swift
```diff
@@ -1357,23 +1349,19 @@ open class NSMutableString : NSString {
     }
     
     public required init(stringLiteral value: StaticString) {
-        if value.hasPointerRepresentation {
-            super.init(String._fromWellFormedCodeUnitSequence(UTF8.self, input: UnsafeBufferPointer(start: value.utf8Start, count: Int(value.utf8CodeUnitCount))))
-        } else {
-            var uintValue = value.unicodeScalar.value
-            super.init(String._fromWellFormedCodeUnitSequence(UTF32.self, input: UnsafeBufferPointer(start: &uintValue, count: 1)))
-        }
+        super.init(value.description)
```
itaiferber: Does `value.description` just return a `String` representing `StaticString`? Is there a performance hit to not initializing directly from the contents?

Foundation/NSString.swift
```diff
@@ -273,7 +273,7 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
     }
     
     public init(characters: UnsafePointer<unichar>, length: Int) {
-        _storage = String._fromWellFormedCodeUnitSequence(UTF16.self, input: UnsafeBufferPointer(start: characters, count: length))
+        _storage = String.init(decoding: UnsafeBufferPointer(start: characters, count: length), as: UTF16.self)
```
itaiferber: We can drop `.init` for brevity, no?

milseman: Yup

Foundation/NSString.swift
milseman: `value.description` is direct initialization of a String with the same contents. It will be faster in the future than trying to do it yourself, as StaticString can poke at String's innards.

Currently, it will encode a single scalar as UTF-8 prior to initialization, and then allocate storage on the heap to store it. We could try to skip that transcoding temporarily, however, I'm about to land small strings and plan on supporting UTF-8 in them. In this case, we still want to encode as UTF-8  there, so we'd do a encoding anyways and skip the heap allocation. This will be better than what this code was doing before: before, even a single scalar gets a heap allocation.

It is equivalent for non-scalar StaticString.

itaiferber: Okay, sounds reasonable, then!

milseman:
@swift-ci please test

milseman:
@swift-ci please test

PR #1477, opened 2018-03-15
SR-2100 - Add support for multi-lib systems

capent:
[SR-2100 - Add support for multi-lib systems](https://bugs.swift.org/browse/SR-2100)

Create the --libdir build option on config.py that set library dir name, default to "lib",
so platforms that uses "lib64" can set this as the libdir name.

This works by setting a new LIBDIR variable, used on file copies by the installer.


TestFoundation/TestURLSession.swift
```diff
@@ -30,7 +30,7 @@ class TestURLSession : LoopbackServerTest {
             ("test_finishTaskAndInvalidate", test_finishTasksAndInvalidate),
             ("test_taskError", test_taskError),
             ("test_taskCopy", test_taskCopy),
-            ("test_cancelTask", test_cancelTask),
+ //           ("test_cancelTask", test_cancelTask),
```
spevans: Why has this test been disabled?

spevans:
@swift-ci please test

TestFoundation/TestURLSession.swift
capent: This test resulted in a odd intermittent failure, but i never bothered to track why since i put all my effort on SR-2100 and it was clearly unrelated to my modifications on the build script, maybe this test failure was specific of my build setup, likely due my firewall.

I forgot to uncomment before doing the push request since i commented that long ago on my multilib branch, I will re-enable. Please, ask the swift-ci bot to retest it just to be sure.

spevans:
@swift-ci please test

parkera:
Seems reasonable to me.

alblue:
@swift-ci please test

configure
```diff
@@ -136,6 +136,7 @@ def main():
     config.script_path            = config.source_root.path_by_appending("build.py")
     config.build_script_path      = config.source_root.path_by_appending("build.ninja")
     config.config_path            = config.source_root.path_by_appending(".configuration")
+    config.libdir                 = os.getenv("LIBDIR", "lib")
```
millenomi: I'm always a little scared about repeatable builds when the config script picks up from the build-host environment. Is there a way to do this via the config setup that we have instead?

capent: Sure, i can add it as a configure command line option, creating a "--libdir" switch using the parser with a default "lib" value. It will imply in a minor modification on build-script-impl at swift repository, but there is no problem since nobody is reviewing that pull request.

millenomi: Yeah, I'd prefer this approach. Thanks!

millenomi:
@capent What's the corresponding build-script pull request?

capent:
@millenomi It's [here](https://github.com/apple/swift/pull/15260).

millenomi:
Please test with following pull request:
apple/swift#15260

@swift-ci please test

capent:
@millenomi This swift-ci test will not work, because the pull request apple/swift#15260 requires the following pull requests to work: 
[apple/swift-corelibs-libdispatch#348](https://github.com/apple/swift-corelibs-libdispatch/pull/348)
[apple/swift-corelibs-xctest#213](https://github.com/apple/swift-corelibs-xctest/pull/213)
[apple/swift-package-manager#1538](https://github.com/apple/swift-package-manager/pull/1538)
[apple/swift-lldb#397](https://github.com/apple/swift-lldb/pull/397)
[apple/swift-llbuild#245](https://github.com/apple/swift-llbuild/pull/245)

I should have explained myself better in my last answer, sorry!

The issue with the apple/swift#15260 is that it integrates also the changes needed to build all 'SR-2100' pull requests that i did these repositories above, since it incorporates the "master" build script, and the build script expects the new command line arguments from the other repositories as well.

As it is, this pull request here is designed to work both with current build script from swift, and the one from the pull request, since it define default values and you can omit the suffix from the build command (the --libdir-suffix is optional).

millenomi:
Please let me know if assistance is required in merging. LGTM.

configure
```diff
@@ -144,10 +144,13 @@ def main():
     parser.add_argument('--linker', dest='linker', type=str, default=None, help="which linker program to use")
     parser.add_argument('--pkg-config', dest='pkg_config', type=str, default="pkg-config")
     parser.add_argument('--bootstrap', dest='bootstrap', type=str, default=os.path.join(os.path.dirname(os.path.abspath(__file__)), "bootstrap"), help="a directory for bootstrapping commonly used headers and libraries not associated directly with the target operating system")
+    parser.add_argument('--libdir-suffix', dest='libdir_suffix', type=str, default=None, help="library folder suffix")
     parser.add_argument('--reconfigure', dest='reconfigure', action="store_true", help="re-generate the build script given the previous configuration")
     parser.add_argument('-v', '--verbose', dest='verbose', action="store_true")
     args, extras = parser.parse_known_args()
     
+    config.libdir = "lib" + args.libdir_suffix
+
```
millenomi: If the default is None, concatenation without that flag passed will throw an exception. Perhaps check?

millenomi:
Actually, one problem.

configure
capent: True, it will throw an exception. Changed for a explicit default before the parse, and then check before concatenating the string.

capent:
I believe that this change will do it.

millenomi:
@swift-ci please test

alblue:
Can we rebase these changes onto master, instead of merging upstream tracking branches in? It will be cleaner in the history when it's finally merged if we do this.

capent:
@alblue Ok, i will try, but it may make a mess out of this chat history.

millenomi:
@swift-ci please test

millenomi:
This PR no longer has an author. I'm closing it for now, but please comment here if you want to take ownership of this.

PR #1480, opened 2018-03-16
[Calendar] `var _autoupdating` could be `private`

ikesyo:


ikesyo:
@swift-ci please test

PR #1481, opened 2018-03-16
FileManager fixes

spevans:
- Use a common subfunction to read directory contents in
  contentsOfDirectory(atPath:) and subpathsOfDirectory(atPath:).

- Use readdir_r() instead of readdir().

- Call fileSystemRepresentation(withPath:) on the requested directory
  to match behaviour on Darwin.

- Replace the test case of a bad directory using "" with "/..." which
  now fails the precondition (to match Darwin).

### DIFF
--- a/TestFoundation/TestFileManager.swift
+++ b/TestFoundation/TestFileManager.swift
@@ -329,7 +329,7 @@ class TestFileManager : XCTestCase {
         } catch _ {
             XCTFail()
         }
-        
+
         if let e = FileManager.default.enumerator(at: URL(fileURLWithPath: path), includingPropertiesForKeys: nil, options: [], errorHandler: nil) {
             var foundItems = [String:Int]()
             while let item = e.nextObject() as? URL {
@@ -442,8 +442,8 @@ class TestFileManager : XCTestCase {
         }
         
         do {
-            let _ = try fm.contentsOfDirectory(atPath: "")
-            
+            // Check a bad path fails
+            let _ = try fm.contentsOfDirectory(atPath: "/...")
             XCTFail()
         }
         catch _ {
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci please test

TestFoundation/TestFileManager.swift
```diff
@@ -442,8 +442,8 @@ class TestFileManager : XCTestCase {
         }
         
         do {
-            let _ = try fm.contentsOfDirectory(atPath: "")
-            
+            // Check a bad path fails
```
millenomi: Why the change here, for the record?

TestFoundation/TestFileManager.swift
```diff
@@ -492,8 +492,8 @@ class TestFileManager : XCTestCase {
         }
         
         do {
-            let _ = try fm.subpathsOfDirectory(atPath: "")
-            
+            // Check a bad path fails
```
millenomi: As above.

TestFoundation/TestFileManager.swift
spevans: Its for compatibility with Darwin's native Foundation. This call will throw an exception under native Foundation as the filename is passed through `fileSystemRepresentation(withPath:)` which treats `nil` and `""` as invalid paths. This PR modifies `contentsOfDirectory(atPath:)` and `subpathsOfDirectory(atPath:)` to now also pass the path through this function so `""` is now going to fail the `precondition(path != "", "Empty path argument")` check.

FYI, if you open the Xcode project `DarwinCompatibilityTests` in the `swift-corelibs-foundation` directory you can run all of the tests as unit tests against native Foundation to see what differences currently exist.

spevans:

PR #1483, opened 2018-03-16
[RFC] Fix parsing long numbers from json

zayass:
Allow to parse long long and unsigned long long even on 32bit platforms
Fixes overflow crash when casting big double to Int

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -812,6 +812,35 @@ private struct JSONReader {
         0x2E, 0x2D, 0x2B, 0x45, 0x65, // . - + E e
     ]
     func parseNumber(_ input: Index, options opt: JSONSerialization.ReadingOptions) throws -> (Any, Index)? {
+       func castInt64(_ value: Int64) -> Any {
+            if Int.bitWidth == 64 {
+                return Int(value)
+            } else {
+                if let result = Int(exactly: value) {
+                    return result
+                }
+
+                return value
+            }
+        }
+
+        func castUInt64(_ value: UInt64) -> Any {
+            if let result = Int(exactly: value) {
+                return result
+            }
+
+            if Int.bitWidth == 64 {
+                return UInt(value)
+            } else {
+                if let result = Int64(exactly: value) {
+                    return result
+                }
+
+                return value
+            }
+        }
+
+
         func parseTypedNumber(_ address: UnsafePointer<UInt8>, count: Int) -> (Any, IndexDistance)? {
             let temp_buffer_size = 64
             var temp_buffer = [Int8](repeating: 0, count: temp_buffer_size)
@@ -823,8 +852,21 @@ private struct JSONReader {
                 defer { intEndPointer.deallocate() }
                 let doubleEndPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
                 defer { doubleEndPointer.deallocate() }
-                let intResult = strtol(startPointer, intEndPointer, 10)
+
+                let isNegative = startPointer.pointee == UInt8(ascii: "-")
+                let intResult: Int64
+                let uintResult: UInt64
+
+                if isNegative {
+                    intResult = strtoll(startPointer, intEndPointer, 10)
+                    uintResult = 0
+                } else {
+                    intResult = 0
+                    uintResult = strtoull(startPointer, intEndPointer, 10)
+                }
+
                 let intDistance = startPointer.distance(to: intEndPointer[0]!)
+
                 let doubleResult = strtod(startPointer, doubleEndPointer)
                 let doubleDistance = startPointer.distance(to: doubleEndPointer[0]!)
 
@@ -835,15 +877,23 @@ private struct JSONReader {
                 let shouldUseReferenceType = opt.contains(.useReferenceNumericTypes)
 
                 if intDistance == doubleDistance {
-                    return (shouldUseReferenceType ? NSNumber(value: intResult) : intResult,
-                            intDistance)
+                    if shouldUseReferenceType {
+                        return isNegative
+                            ? (NSNumber(value: intResult) as Any, intDistance)
+                            : (NSNumber(value: uintResult) as Any, intDistance)
+                    }
+
+                    return isNegative
+                        ? (castInt64(intResult), intDistance)
+                        : (castUInt64(uintResult), intDistance)
                 }
+
                 guard doubleDistance > 0 else {
                     return nil
                 }
 
-                if doubleResult == doubleResult.rounded() {
-                    return (shouldUseReferenceType ? NSNumber(value: Int(doubleResult)) : Int(doubleResult),
+                if doubleResult == doubleResult.rounded(), let result = Int(exactly: doubleResult)  {
+                    return (shouldUseReferenceType ? NSNumber(value: result) : result,
                             doubleDistance)
                 }
 


zayass:
It little bit change behavior and not contain tests
If this behavior is ok I will add tests

Foundation/JSONSerialization.swift
```diff
@@ -823,8 +858,21 @@ private struct JSONReader {
                 defer { intEndPointer.deallocate() }
                 let doubleEndPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
                 defer { doubleEndPointer.deallocate() }
-                let intResult = strtol(startPointer, intEndPointer, 10)
+
+                let isNegative = startPointer.pointee == 0x2D
+                let intResult: Int64
```
spevans: 0x2D is better written as UInt8(ascii: "-") so the value is more obvious

Foundation/JSONSerialization.swift
```diff
@@ -812,6 +812,41 @@ private struct JSONReader {
         0x2E, 0x2D, 0x2B, 0x45, 0x65, // . - + E e
     ]
     func parseNumber(_ input: Index, options opt: JSONSerialization.ReadingOptions) throws -> (Any, Index)? {
+        #if !arch(i386) && !arch(arm) // x64
+            func castInt64(_ value: Int64) -> Any {
```
spevans: Its better to use `Int.bitWidth == 64` rather than adding a list of architectures which would need updating

Foundation/JSONSerialization.swift
```diff
@@ -823,8 +858,21 @@ private struct JSONReader {
...
+                let uintResult: UInt64
+
+                if isNegative {
+                    intResult = strtoll(startPointer, intEndPointer, 10)
+                    uintResult = 0
+                } else {
+                    intResult = 0
+                    uintResult = strtoull(startPointer, intEndPointer, 10)
+                }
+
```
spevans: The `sttroll` and `strtoull` functions can both return 0 on overflow and underflow so this needs to be detected and possibly `nil` returned if the `strtod` also can't parse the number

Foundation/JSONSerialization.swift
zayass: I agree this looks ugly, but I want to move this to compile time. Do you know some better way to make a check like this without extra runtime cost?

Foundation/JSONSerialization.swift
zayass: `strtol`, `sttroll`, `strtoull` returns max allowed value on overflow and min on underflow.
It is main problem with original code. It returns Int32.max(on android which is 32bit currently) even if number is bigger without any failing.

As I understand original code case you are described handled below
```
guard intDistance > 0 || doubleDistance > 0 else {
    return nil
}
```

Foundation/JSONSerialization.swift
zayass: Do you expect more 32bit arch in future? I think all new archs would be 64bit and this if will be work correctly.

zayass:
The main difference in this branch compared to master is using `strtoll` or `strtoull` depends on sign instead of `strtol`.

On 64bit platforms this allow to parse UInt64 which lead to overflow early.
On 32bit platforms even more cases lead to overflow - UInt32, Int64, UInt64.

But this introduce kind of API changes. Early every json integer would be an Int.
After this change it can be Int or UInt on 64bit platform depends on input.
On 32bit it can be Int, Int64, UInt64.

Foundation/JSONSerialization.swift
xwu: The compiler knows how to optimize `Int.bitWidth == 64` to a compile-time constant.

Foundation/JSONSerialization.swift
```diff
@@ -812,6 +812,35 @@ private struct JSONReader {
...
+                return value
+            }
+        }
+
+        func castUInt64(_ value: UInt64) -> Any {
+            if let result = Int(exactly: value) {
+                return result
+            }
+
+            if Int.bitWidth == 64 {
```
itaiferber: Should this be `UInt.bitWidth`?

Foundation/JSONSerialization.swift
itaiferber: These checks aren't necessarily correct, by the way; usually the right way to validate the result of `stro*` is to check the value of where the end pointer is pointing. For instance, `strtoull("123f", intEndPointer, 10)` will return `123` and have `intEndPointer` pointing to `'f'`. Since we have a `count` and are copying into a null-terminated buffer, we should be checking the result of `intEndPointer.pointee == 0` to validate whether we parsed a full number or not.

Foundation/JSONSerialization.swift
```diff
@@ -823,8 +852,21 @@ private struct JSONReader {
                 defer { intEndPointer.deallocate() }
                 let doubleEndPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
                 defer { doubleEndPointer.deallocate() }
-                let intResult = strtol(startPointer, intEndPointer, 10)
+
+                let isNegative = startPointer.pointee == UInt8(ascii: "-")
+                let intResult: Int64
+                let uintResult: UInt64
+
+                if isNegative {
```
itaiferber: Rather than repeating the result here, can we please instead advance past a `-` if present and always parse the remaining digits as `UInt64`? It will simplify the casting above and below to always work in terms of one type (`UInt64`) until you need to down cast it and check the range (esp. if we do have a `-`).

Foundation/JSONSerialization.swift
zayass: Maybe. But I think it is impossible to Int.bitWidth != UInt.bitWidth

Foundation/JSONSerialization.swift
zayass: Good idea. Will do

zayass:
Hi @itaiferber thanks for comments

What do you think about API? `Int`/`Int64`/`UInt64` on 32bit and `Int`/`UInt` on 64bit is ok?
About test, of course, it should be fully covered but I want to check if everybody ok with API firstly.

And one more about tests, do you know if any 32bit test env presented on CI? Or any possibility to emulate this in tests?

Foundation/JSONSerialization.swift
```diff
@@ -812,6 +812,35 @@ private struct JSONReader {
...
+                    return result
+                }
+
+                return value
+            }
+        }
+
+
         func parseTypedNumber(_ address: UnsafePointer<UInt8>, count: Int) -> (Any, IndexDistance)? {
             let temp_buffer_size = 64
```
zayass: Anybody knows why 64 char length hardcoded here?

itaiferber: I'm guessing the idea here is that we won't be able to represent any number >64 digits long faithfully; this isn't necessarily a correct way to do this, though.

zayass: I am trying to parse `String(repeating: "9", count: 165)` on Darwin and it parsed as 1.0e165 which is much better than 1.0e64
But 166 failed with an exception, which works fine in js in chrome.
```
Domain=NSCocoaErrorDomain Code=3840 "Number wound up as NaN around character 1."
```

itaiferber: What OS version are you running?

zayass: macOS 10.12.6 Xcode 9.2

itaiferber: So yes, on macOS Sierra (10.12) and previous, you'll see the older behavior of a reduced range of values which parse as a `double` and more numbers which fall into the case parsing an `NSDecimalNumber` (which has a more limited range).

zayass: Ok long floating point numbers works even worse then I think. Looks like another big issue for another PR

This code fails on swift 4.0.2 on Ubuntu 16.04
```
import Foundation

let json = "[\(Int.max - 100).5]".data(using: .utf8)!
let data = try! JSONSerialization.jsonObject(with: json)

print(data)
```

millenomi:
@zayass Are you still interested in chaperoning this patch?

shahmishal:

PR #1489, opened 2018-03-25
Display absolute path for Foundation lib dir during manual testing.

spevans:
- This path is shown when 'ninja test' is run.

- The LD_LIBRARY_PATH passed to TestFoundation via manual testing is
  inherited by xdgTestHelper. If the path is relative and the
  current directory has been changed then libraries will not be
  found. This is the case if compiled with debug and libFoundation
  is linked to libswiftSwiftOnone.

### DIFF
--- a/build.py
+++ b/build.py
@@ -539,7 +539,7 @@ foundation.add_phase(plutil)
 
 script.add_product(foundation)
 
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

PR #1496, opened 2018-03-31
FileManager: Implement mountedVolumeURLs(includingResourceValuesForKeys:options:)

spevans:
- propertyKeys and options arent fully implemented.

- Only works on Linux and macOS.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -32,10 +32,64 @@ open class FileManager : NSObject {
         }
     }
     
-    /* Returns an NSArray of NSURLs locating the mounted volumes available on the computer. The property keys that can be requested are available in NSURL.
-     */
+    /// Returns an array of URLs that identify the mounted volumes available on the device.
     open func mountedVolumeURLs(includingResourceValuesForKeys propertyKeys: [URLResourceKey]?, options: VolumeEnumerationOptions = []) -> [URL]? {
-        NSUnimplemented()
+        var urls: [URL]
+
+#if os(Linux)
+        guard let procMounts = try? String(contentsOfFile: "/proc/mounts", encoding: .utf8) else {
+            return nil
+        }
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

parkera:
Neat

parkera:
cc @kperryua

Foundation/FileManager.swift
```diff
@@ -32,10 +32,64 @@ open class FileManager : NSObject {
...
+            return nil
+        }
+        urls = []
+        for line in procMounts.components(separatedBy: "\n") {
+            let mountPoint = line.components(separatedBy: " ")
+            if mountPoint.count > 2 {
+                urls.append(URL(fileURLWithPath: mountPoint[1], isDirectory: true))
+            }
+        }
+#elseif os(macOS)
```
millenomi: Can you use canImport(Darwin) here instead? I'm ok with os(Linux) for the above because /proc/mounts is explicitly a Linux feature, but I'd rather not tie one Darwin API to a specific OS.

millenomi: I'm also half-thinking whether we also need to start using:

```swift
#else
#error("Requires a platform-specific implementation")
#endif
```

now that we _do_ have `#error`.

spevans:
@swift-ci  please test

spevans:
@millenomi I added both of those changes although the `canImport(Darwin)` does feel a bit hacky.

spevans:
@swift-ci please test

millenomi:
@spevans I think I want to push standardizing on canImport(Darwin) in the rest of corelibs-foundation, too. (i.e.: if it's a Darwin-based OS, we should ask if Darwin API are available, not which particular flavor of it it is.)

millenomi:

spevans:

PR #1499, opened 2018-04-01
Fix nsdata fails to access contents of url

tid-kijyun:
`NSData (contentsOf :)` was always initialized with length 0.
Also, Testing `NSData (contentsOf :)` on a test HTTP server(`_HTTPServer`) will cause deadlock, so I changed it to work on sub thread.

### DIFF
--- a/TestFoundation/HTTPServer.swift
+++ b/TestFoundation/HTTPServer.swift
@@ -207,7 +207,7 @@ class _HTTPServer {
             deadlineTime = .now()
         }
 
-        DispatchQueue.main.asyncAfter(deadline: deadlineTime) {
+        DispatchQueue.global().asyncAfter(deadline: deadlineTime) {
             do {
                 try self.socket.writeData(header: response.header, body: response.body, sendDelay: sendDelay, bodyChunks: bodyChunks)
                 semaphore.signal()
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -218,7 +218,7 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
             guard let data = resData else {
                 throw resError!
             }
-            _init(bytes: UnsafeMutableRawPointer(mutating: data._nsObject.bytes), length: length, copy: true)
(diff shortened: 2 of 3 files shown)


spevans:
@swift-ci please test

TestFoundation/HTTPServer.swift
```diff
@@ -207,7 +207,7 @@ class _HTTPServer {
             deadlineTime = .now()
         }
 
-        DispatchQueue.main.asyncAfter(deadline: deadlineTime) {
+        DispatchQueue.global().asyncAfter(deadline: deadlineTime) {
```
parkera: Can we split up the data fix itself from the change to the URLSession test code?

tid-kijyun: No, we can't. Without this change, deadlock will occur in testing this PR. However If you can merge split PRs in order, I should be able to do that.
Would you like to split this?

spevans: I think it should be merged as one PR as the `HTTPServer.swift` change is required for the test case to work.

parkera: I don't think there is anything specific to `HTTPServer` about the change to `NSData`. How is it using `NSData` that triggers this bug?

spevans: If the test case that was added (`test_contentsOfURL`) is calling `NSData(contentsOf: url)` and that is running on the main thread then won't the HTTP server have to run on another thread to service the request, or am I mis-understanding how `DispatchQueue` works?

tid-kijyun: This is a code that reproduces this bug.

```swift
import Foundation

if let data = NSData(contentsOf: URL(string: "https://swift.org")!) {
    print(data.length) // 0
}
```

millenomi:
It makes sense to me that in the case the same thread is responding to the HTTP request initiated by NSData as the one that is blocked executing NSData(contentsOf: URL), then the test code should move its serving to a different thread.

TestFoundation/HTTPServer.swift
parkera: Ok, thanks for the additional info.

millenomi:

PR #1497, opened 2018-04-01
[SE-0193] Rename the attributes to their final names

ikesyo:
Ref: https://github.com/apple/swift/pull/15594

PR #1502, opened 2018-04-03
Revert "[SE-0193] Rename the attributes to their final names"

moiseev:
Reverts apple/swift-corelibs-foundation#1497 as it [caused problems](https://ci.swift.org/job/oss-swift-4.2-incremental-RA-linux-ubuntu-16_04/516/) for Linux builds with swift-4.2-branch, where new attributes are not yet available.

corelibs-foundation code should be compatible with both swift-4.2-branch and master branches of the main Swift repo.

### DIFF
--- a/Foundation/Boxing.swift
+++ b/Foundation/Boxing.swift
@@ -14,7 +14,7 @@
 ///
 /// Note: This assumes that the result of calling copy() is mutable. The documentation says that classes which do not have a mutable/immutable distinction should just adopt NSCopying instead of NSMutableCopying.
 internal final class _MutableHandle<MutableType : NSObject> where MutableType : NSCopying {
-    @usableFromInline internal var _pointer : MutableType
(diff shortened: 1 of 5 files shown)


moiseev:

millenomi:
Noted.

ikesyo:
So what is the purpose of the swift -4.2-branch of this repo? I thought the issue was addressed with https://github.com/apple/swift/pull/15698 and https://github.com/apple/swift/pull/15699.

moiseev:
@ikesyo apparently (and I did not know this when I reverted this PR) there is indeed a separate 4.2-branch for corelibs-foundation, but! there is a catch: there is an automerge set-up that unconditionally applies all the changes from master to the swift-4.2-branch. It should have been enough to simple remove this change in the 4.2 branch alone and not in master. And yes, you're right, once the rebranch happens, we can revert the revert.

PR #1504, opened 2018-04-03
Marked UserDefaults.synchronize() as @discardableResult

pvieito:
- Same as in Darwin

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Here is a simple `main.swift` file which uses Foundation. This guide assumes you
 import Foundation
 
 // Make an URLComponents instance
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -33,7 +33,7 @@ Here is a simple `main.swift` file which uses Foundation. This guide assumes you
 import Foundation
 
 // Make an URLComponents instance
-let swifty = NSURLComponents(string: "https://swift.org")!
+let swifty = URLComponents(string: "https://swift.org")!
```
millenomi: Let's not mix different things in the same patch? PRs are cheap :)

pvieito: Seems reasonable :)

PR #1507, opened 2018-04-04
TestXMLDocument: Un-XFAIL DTD tests.

spevans:
- Serve the DTD from the built in HTTPServer so the tests
  will work without internet access.

- Resolves rdar://31567922.


spevans:
@swift-ci please test

alblue:
This LGTM. What do you think, @pushkarnk ?

millenomi:
More tests are nice, and this was open long enough that we can open a separate PR for feedback if needed.

millenomi:

PR #1510, opened 2018-04-08
FileManager: Implement contentsEqual(atPath:andPath:)

spevans:
- Requires some more tests, especially for comparing directories.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -637,11 +637,161 @@ open class FileManager : NSObject {
     open func isDeletableFile(atPath path: String) -> Bool {
         NSUnimplemented()
     }
-    
+
+    private func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64) -> Bool {
+        let bufSize = min(size, 1024 * 1024)
+        let fd1 = open(file1Rep, O_RDONLY)
+        guard fd1 >= 0 else {
+            return false
+        }
+        defer { close(fd1) }
+
+        let fd2 = open(file2Rep, O_RDONLY)
+        guard fd2 >= 0 else {
+            return false
+        }
+        defer { close(fd2) }
+
+        let buffer1 = UnsafeMutablePointer<UInt8>.allocate(capacity: Int(bufSize))
+        let buffer2 = UnsafeMutablePointer<UInt8>.allocate(capacity: Int(bufSize))
+        defer {
+            buffer1.deallocate()
+            buffer2.deallocate()
+        }
+
+        var bytesLeft = size
+        while bytesLeft > 0 {
+            let bytesToRead = Int(min(bufSize, bytesLeft))
+            guard read(fd1, buffer1, bytesToRead) == bytesToRead else {
+                return false
+            }
+            guard read(fd2, buffer2, bytesToRead) == bytesToRead else {
+                return false
+            }
+            guard memcmp(buffer1, buffer2, bytesToRead) == 0 else {
+                return false
+            }
+            bytesLeft -= Int64(bytesToRead)
+        }
+        return true
+    }
+
+    private func _compareSymlinks(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64) -> Bool {
+        let bufSize = Int(size) + 1
+        let buffer1 = UnsafeMutablePointer<CChar>.allocate(capacity: Int(bufSize))
+        let buffer2 = UnsafeMutablePointer<CChar>.allocate(capacity: Int(bufSize))
+
+        let size1 = readlink(file1Rep, buffer1, bufSize)
+        let size2 = readlink(file2Rep, buffer2, bufSize)
+
+        let compare: Bool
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci please test

Foundation/FileManager.swift
```diff
@@ -637,11 +637,161 @@ open class FileManager : NSObject {
...
+        while let item = enumerator1.nextObject() as? String {
+            path1entries.insert(item)
+        }
+
+        while let item = enumerator2.nextObject() as? String {
+            if path1entries.remove(item) == nil {
+                return false
+            }
+        }
+        return path1entries.isEmpty
```
kperryua: I'd expect this to recurse for each item in the two directories. If the contents of 'A/C' doesn't match 'B/C', then this is supposed to fail.

Foundation/FileManager.swift
```diff
@@ -637,11 +637,161 @@ open class FileManager : NSObject {
...
+            guard memcmp(buffer1, buffer2, bytesToRead) == 0 else {
+                return false
+            }
+            bytesLeft -= Int64(bytesToRead)
+        }
+        return true
+    }
+
+    private func _compareSymlinks(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64) -> Bool {
+        let bufSize = Int(size) + 1
```
kperryua: On Darwin at least, `readlink` does not append a NUL byte, so the +1 isn't strictly necessary.

Foundation/FileManager.swift
```diff
@@ -637,11 +637,161 @@ open class FileManager : NSObject {
...
+            return false
+        }
+        let file2Type = file2.st_mode & S_IFMT
+
+        // Are paths the same type: file, directory, symbolic link etc.
+        guard file1Type == file2Type else {
+            return false
+        }
+
+        if file1Type == S_IFCHR {
```
kperryua: The same could be done here for `S_IFBLK`, no?

TestFoundation/TestFileManager.swift
```diff
@@ -620,4 +621,83 @@ class TestFileManager : XCTestCase {
...
+        XCTAssertFalse(fm.contentsEqual(atPath: emptyFile.path, andPath: testDir1.appendingPathComponent("null1").path))
+        XCTAssertFalse(fm.contentsEqual(atPath: emptyFile.path, andPath: testDir1.appendingPathComponent("unreadable_file").path))
+
+        XCTAssertTrue(fm.contentsEqual(atPath: testFile1URL.path, andPath: testFile1URL.path))
+        XCTAssertFalse(fm.contentsEqual(atPath: testFile1URL.path, andPath: testFile2URL.path))
+        XCTAssertFalse(fm.contentsEqual(atPath: testFile3URL.path, andPath: testFile4URL.path))
+        XCTAssertFalse(fm.contentsEqual(atPath: symlink, andPath: testFile1URL.path))
+        XCTAssertTrue(fm.contentsEqual(atPath: testDir1.path, andPath: testDir1.path))
+        XCTAssertTrue(fm.contentsEqual(atPath: testDir2.path, andPath: testDir3.path))
+        XCTAssertFalse(fm.contentsEqual(atPath: testDir1.path, andPath: testDir2.path))
```
kperryua: A test case that verifies recursive (in)equality is needed.

Foundation/FileManager.swift
```diff
@@ -637,11 +637,161 @@ open class FileManager : NSObject {
     open func isDeletableFile(atPath path: String) -> Bool {
         NSUnimplemented()
     }
-    
+
+    private func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64) -> Bool {
+        let bufSize = min(size, 1024 * 1024)
```
kperryua: It might be worth using `st_blksize` to pick an optimal block size. 1 MB (*2) seems like a rather large allocation.

spevans: Im not sure how to relate it to block size, given there are only a few fixed block sizes. How does Darwin's Foundation decide on the buffer size to use?

spevans:
@swift-ci please test

Foundation/FileManager.swift
kperryua: `st_blksize` is documented as "The optimal I/O block size for the file". I've taken that to mean that, especially if we're not reading the entire contents of the file into a single buffer (a la NSData), that this value is the system-declared preferred buffer size for I/O operations. I won't link directly to any here, but various implementations of 'cat' agree (including Darwin's).

Darwin's implementation of this method is fairly old and happens not to use this value—opting for a constant 8K size instead—though it should probably should.

spevans: Ah right, I mis-read this as using some multiple of `st_blksize` rather than the value directly.

spevans:
@swift-ci please test

alblue:
@swift-ci please test

spevans:
I added some extra tests for deep subdirectories to check the recursion. It exposed a bug in `copyItem(atPath:toPath:)` (#1515) which will need to be merged before this and also I came across a bug in `fileExists(atPath:isDirectory)` (#1514)  which should also go in .

parkera:
Great, thanks @spevans!

millenomi:
@swift-ci please test

PR #1514, opened 2018-04-14
FileManager: Fix fileExists(atPath:isDirectory:) with dangling symlink

spevans:
- When the path item is a symlink and the target of the symlink does
  not exist, return false instead of true.

- Combine the calls to stat() when the file is a symlink.

Issue found whilst testing https://github.com/apple/swift-corelibs-foundation/pull/1510


spevans:
@swift-ci please test

PR #1516, opened 2018-04-15
Remove XCTAssertNotNil calls where parameter is a non-optional type

otaviolima:


parkera:
@swift-ci please test

parkera:
What happened here in the first place? Did we change the results from nullable to non-null?

otaviolima:
>What happened here in the first place? Did we change the results from nullable to non-null?

I did not check the code changes over time, but I have a feeling that is related to the annotations on the Darwin counter-part being updated which then make those `init` swing from `nullable` to `non-nullable` annotation. But that is just an assumption with no fact check what so ever. :)

ianpartridge:

PR #1518, opened 2018-04-16
Changes required for swift-4.1-branch port to Android

johnno1962:
Hi Apple,

Could you please consider merging this PR derived from the changes required to get the swift-4.1-branch running on Android. The bulk of the PR has to adjust the memory model of CF instances that use NSbridged classes for storage. I found that the sizeof the CF structure (which is more of a “stride”) was greater than the instance size Swift allocated resulting in the initialisation of the CF instance writing off the end of allocated Swift instance memory. This is catered for by the _CFPad struct and probably applies to all 32 bit ports of Swift. The remainder of changes are various tweaks to get tests running and to accommodate the environment inside an Android application for things such as NSUserDefaults.

Test results are:
```
Test Suite 'All tests' failed at 2018-04-16 16:16:34.606
	 Executed 1239 tests, with 23 failures (0 unexpected) in 17.16 (17.16) seconds
```

### DIFF
--- a/CoreFoundation/PlugIn.subproj/CFBundle_Main.c
+++ b/CoreFoundation/PlugIn.subproj/CFBundle_Main.c
@@ -79,20 +79,12 @@ static CFBundleRef _CFBundleGetMainBundleAlreadyLocked(void) {
         CFStringRef str = NULL;
         CFURLRef executableURL = NULL, bundleURL = NULL;
         _initedMainBundle = true;
-#if DEPLOYMENT_TARGET_ANDROID
-        const char *bundlePath = getenv("CFFIXED_USER_HOME") ?: getenv("TMPDIR") ?: "/data/local/tmp";
-        CFStringRef bundleRef = CFStringCreateWithFileSystemRepresentation(kCFAllocatorNull, bundlePath);
-        bundleURL = CFRetain(CFURLCreateWithFileSystemPath(kCFAllocatorSystemDefault, bundleRef,
-                                                           PLATFORM_PATH_STYLE, false));
-        CFRelease(bundleRef);
-#else
         processPath = _CFProcessPath();
         if (processPath) {
             str = CFStringCreateWithFileSystemRepresentation(kCFAllocatorSystemDefault, processPath);
             if (!executableURL) executableURL = CFURLCreateWithFileSystemPath(kCFAllocatorSystemDefault, str, PLATFORM_PATH_STYLE, false);
         }
         if (executableURL) bundleURL = _CFBundleCopyBundleURLForExecutableURL(executableURL);
-#endif
         if (bundleURL) {
             // make sure that main bundle has executable path
             //??? what if we are not the main executable in the bundle?
--- a/CoreFoundation/Preferences.subproj/CFPreferences.c
+++ b/CoreFoundation/Preferences.subproj/CFPreferences.c
@@ -440,11 +440,7 @@ static CFURLRef _CFPreferencesURLForStandardDomainWithSafetyLevel(CFStringRef do
     CFURLRef theURL = NULL;
     CFAllocatorRef prefAlloc = __CFPreferencesAllocator();
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_LINUX || DEPLOYMENT_TARGET_WINDOWS
-#if DEPLOYMENT_TARGET_ANDROID
-    CFURLRef prefDir = CFBundleCopyBundleURL(CFBundleGetMainBundle());
-#else
     CFURLRef prefDir = _preferencesDirectoryForUserHostSafetyLevel(userName, hostName, safeLevel);
-#endif
     CFStringRef  appName;
     CFStringRef  fileName;
     Boolean mustFreeAppName = false;
--- a/Foundation/Bundle.swift
+++ b/Foundation/Bundle.swift
@@ -53,15 +53,9 @@ open class Bundle: NSObject {
         self.init(path: url.path)
     }
     
-#if os(Android)
-    public convenience init(for aClass: AnyClass) {
-        self.init(path: Bundle.main.bundlePath)!
-    }
-#else
     public init(for aClass: AnyClass) {
         NSUnimplemented()
     }
-#endif
 
     public init?(identifier: String) {
         super.init()


CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -448,6 +448,12 @@ CF_EXPORT double kCFCoreFoundationVersionNumber;
 #define kCFCoreFoundationVersionNumber_iOS_9_x_Max 1299
 #endif
 
+#ifdef __ANDROID__
+typedef uint16_t _swift_rc_type;
```
millenomi: Types that are private to CF should be in camel-case style with the _CF prefix. `_CFSwiftRCCount`?

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -216,7 +216,7 @@ extern void __CFGenericValidateType_(CFTypeRef cf, CFTypeID type, const char *fu
 #define __CFBitfield64GetValue(V, N1, N2)	(((V) & __CFBitfield64Mask(N1, N2)) >> (N2))
 #define __CFBitfield64SetValue(V, N1, N2, X)	((V) = ((V) & ~__CFBitfield64Mask(N1, N2)) | ((((uint64_t)X) << (N2)) & __CFBitfield64Mask(N1, N2)))
 
-#if __LP64__
+#if __LP64__ || defined(__ANDROID__)
```
millenomi: What is the need here? Can we document why we need to special-case Android, or can we use a compiler capability flag if that's not the case?

CoreFoundation/Base.subproj/CFRuntime.h
```diff
@@ -193,8 +193,8 @@ CF_EXPORT void _CFRuntimeUnregisterClassWithTypeID(CFTypeID typeID);
 typedef struct __CFRuntimeBase {
     // This matches the isa and retain count storage in Swift
     uintptr_t _cfisa;
-    uint32_t _swift_strong_rc;
-    uint32_t _swift_weak_rc;
+    _swift_rc_type _swift_strong_rc;
```
millenomi: (See above.)

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -1000,6 +1000,7 @@ void CFLog1(CFLogLevel lev, CFStringRef message) {
         CFStringGetCString(message, buffer, maxLength, encoding);
     
     __android_log_print(priority, tag, "%s", buffer);
+    fprintf(stderr, "%s\n", buffer);
```
millenomi: What is this for? Isn't `__android_log_print` enough?

CoreFoundation/PlugIn.subproj/CFBundle.c
```diff
@@ -714,7 +714,7 @@ static CFBundleRef _CFBundleCreate(CFAllocatorRef allocator, CFURLRef bundleURL,
 
     bundle->_url = newURL;
     
-#if !DEPLOYMENT_RUNTIME_OBJC && !DEPLOYMENT_TARGET_WINDOWS
+#if !DEPLOYMENT_RUNTIME_OBJC && !DEPLOYMENT_TARGET_WINDOWS && !defined(__ANDROID__)
```
millenomi: I'm thinking if, for consistency, there shouldn't be a `DEPLOYMENT_TARGET_ANDROID` that we could use. I'm also pretty sure there are other FHS bits — it should be entirely enabled or entirely disabled.

CoreFoundation/PlugIn.subproj/CFBundle_Main.c
```diff
@@ -79,12 +79,20 @@ static CFBundleRef _CFBundleGetMainBundleAlreadyLocked(void) {
         CFStringRef str = NULL;
         CFURLRef executableURL = NULL, bundleURL = NULL;
         _initedMainBundle = true;
+#ifdef __ANDROID__
```
millenomi: This seems to be setting the bundle URL to `$CFFIXED_USER_HOME` or (a) `$TMPDIR`, which is highly surprising for downstream users — this is supposed to be a directory path where the application (as 'application' is defined by the platform) is storing read-only resource files.

Some quick research suggests this would be covered by the `AAssetManager` API on Android; can we go the distance for that? Why isn't the default (creating a bundle out of the path the executable is in) correct on this platform?

CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -440,7 +440,11 @@ static CFURLRef _CFPreferencesURLForStandardDomainWithSafetyLevel(CFStringRef do
     CFURLRef theURL = NULL;
     CFAllocatorRef prefAlloc = __CFPreferencesAllocator();
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_LINUX || DEPLOYMENT_TARGET_WINDOWS
+#ifdef __ANDROID__
+    CFURLRef prefDir = CFBundleCopyBundleURL(CFBundleGetMainBundle());
```
millenomi: It is not correct to write from or to the main bundle. The system must function if the application is running with its main bundle path read only.

This should be a location where application-specific user information can be read or written to. A quick search shows the closest would be the `internalDataPath` vended through a `ANativeActivity`, but that may not be the layering you want. Either way, this is incorrect in the face of the semantics of this API.

CoreFoundation/String.subproj/CFString.h
```diff
@@ -155,8 +155,8 @@ since it is the default choice with Mac OS X developer tools.
 struct __CFConstStr {
     struct {
         uintptr_t _cfisa;
-        uint32_t _swift_strong_rc;
-        uint32_t _swift_weak_rc;
+        _swift_rc_type _swift_strong_rc;
```
millenomi: (See above.)

Foundation/Bundle.swift
```diff
@@ -53,8 +53,16 @@ open class Bundle: NSObject {
         self.init(path: url.path)
     }
     
-    public init(for aClass: AnyClass) { NSUnimplemented() }
-    
+#if os(Android)
```
millenomi: Can you explain your intent here?

Foundation/NSCFString.swift
```diff
@@ -61,8 +61,8 @@ internal final class _NSCFConstantString : _NSCFString {
     internal var _ptr : UnsafePointer<UInt8> {
         // FIXME: Split expression as a work-around for slow type
         //        checking (tracked by SR-5322).
-        let offTemp1 = MemoryLayout<OpaquePointer>.size + MemoryLayout<Int32>.size
-        let offTemp2 = MemoryLayout<Int32>.size + MemoryLayout<_CFInfo>.size
+        let offTemp1 = MemoryLayout<OpaquePointer>.size + MemoryLayout<_swift_rc_type>.size
```
millenomi: (See above.)

Foundation/NSCFArray.swift
```diff
@@ -12,7 +12,9 @@ import CoreFoundation
 
 internal final class _NSCFArray : NSMutableArray {
     deinit {
+#if !os(Android)
```
millenomi: I'm pretty sure this is semantically incorrect. Can you explain why not this?

Foundation/NSObjCRuntime.swift
```diff
@@ -228,6 +228,14 @@ internal struct _CFInfo {
     }
 }
 
+internal struct _CFPad {
+#if os(Android)
```
millenomi: I'd want compiler people to chime in on this to understand if the guarantee this code relies on re: sizing will hold. Who could we cc here?

Foundation/NSString.swift
```diff
@@ -630,6 +630,10 @@ extension NSString {
     }
 
     public var boolValue: Bool {
+#if os(Android)
```
millenomi: This makes it inconsistent from other platform only on Android. What's the reasoning here?

TestFoundation/TestFileManager.swift
```diff
@@ -228,6 +228,9 @@ class TestFileManager : XCTestCase {
     }
     
     func test_fileSystemAttributes() {
+#if os(Android)
```
millenomi: Patches with tests that fail can't be merged. Would you like to instead figure out a way to mark the attributes API as not available on Android?

TestFoundation/TestNSNumber.swift
```diff
@@ -1087,33 +1087,33 @@ class TestNSNumber : XCTestCase {
             XCTAssertEqual(NSNumber(value: UInt.min).stringValue, "0")
             XCTAssertEqual(NSNumber(value: UInt.min + 1).stringValue, "1")
             XCTAssertEqual(NSNumber(value: UInt.max).stringValue, "4294967295")
-            XCTAssertEqual(NSNumber(value: UInt.max - 1).stringValue, "4294967294")
+            XCTAssertEqual(NSNumber(value: UInt.max - 1 as UInt).stringValue, "4294967294")
```
millenomi: Is this and similar a 32-bit guarantee violation? Why doesn't it work on Android as written? If it is incorrect, it should be diagnosed as incorrect on all platforms to avoid forking the language.

CoreFoundation/Base.subproj/CFBase.h
parkera: This should not be in CFBase.h at all. This is a public header and this is totally internal to CF.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -297,6 +297,9 @@ CF_EXPORT CFStringRef CFCopyFullUserName(void) {
     uid_t euid;
     __CFGetUGIDs(&euid, NULL);
     struct passwd *upwd = getpwuid(euid ? euid : getuid());
+#ifdef __ANDROID__
```
parkera: I would like to use a `TARGET_OS_ANDROID` value instead, if possible, to match the rest of our macros.

CoreFoundation/PlugIn.subproj/CFBundle_Main.c
parkera: I would stick with the CF behavior of using the current path in this case.

CoreFoundation/Base.subproj/CFBase.h
johnno1962: Thanks for the comments. I can change this type to _CFSwiftRCCount and #ifdef __ANDROID__’s in this PR to DEPLOYMENT_TARGET_ANDROID no problem. CFBase.h seemed to be the only place I could define this type which is included by both [CFString.h](https://github.com/SwiftJava/swift-corelibs-foundation/blob/master/CoreFoundation/String.subproj/CFString.h#L158) and [CFRuntime.h](https://github.com/SwiftJava/swift-corelibs-foundation/blob/master/CoreFoundation/Base.subproj/CFRuntime.h#L196). It is also important it is available to Swift in [NSCFString.swift](https://github.com/SwiftJava/swift-corelibs-foundation/blob/master/Foundation/NSCFString.swift#L64) for the code to hold together so if an accommodation could be made here that would make the most sense to me.

CoreFoundation/Base.subproj/CFInternal.h
johnno1962: This #if __LP64__ seems to be a leftover as __CFInfoType is always uint64_t in the actual definition in [CFRuntime.h](https://github.com/SwiftJava/swift-corelibs-foundation/blob/master/CoreFoundation/Base.subproj/CFRuntime.h#L199). I added this defined(__ANDROID__) to silence the many warnings telling me something wasn’t quite right. This #if should probably be removed altogether.

CoreFoundation/Base.subproj/CFPlatform.c
johnno1962: Sure.

CoreFoundation/Base.subproj/CFUtilities.c
johnno1962: This is helpful when you are running an Android binary (such as the tests) from the command line using an “adb shell". Otherwise NSLogs are only visible using an “adb logcat” by which time the messages have lost their context. This doesn’t affect normal running of an actual application and the logging to stderr goes into the void.

CoreFoundation/PlugIn.subproj/CFBundle.c
johnno1962: What exactly is FHS?? There may be other areas I should be #ifdeffing out but this is the only one that was causing a crash.

Foundation/Bundle.swift
johnno1962: This is for compatibility for Open Source code notably Alamofire to run (almost) without modification. It uses the main bundle to try to read the Info.plist to get the app version. Non of this works but with this change at least it doesn’t crash.

Foundation/NSCFArray.swift
johnno1962: Ahhh, yes… this does rather stand out as being an area that should be investigated. Basically if this call is made the app will crash further down the line due presumably to freeing invalid block or freeing them when they shouldn’t be. If made a few passes at trying to get to the bottom of this but had no luck. Avoiding making this call solves the problem. I’m not even sure self is a CFArray instance at this point as even just asking for the count will crash the app and have no idea where the pointer is coming from.

Foundation/NSString.swift
johnno1962: For whatever reason the NSString.boolValue implementation crashes and it exercises quite a bit of code including charsets which is where the problem most likely lies. I put this in as a stop-gap to keep moving. If you ask me this is what the implementation should be anyway. The Scanner/Charset version seems to make a bit of a meal of it.

TestFoundation/TestFileManager.swift
johnno1962: [Filemanager.swift](https://github.com/SwiftJava/swift-corelibs-foundation/blob/master/Foundation/FileManager.swift#L380) marks this as NSUnimplmented so I didn’t have much option if the tests were to run. Perhaps it should be made unavailable if you can suggest the syntax for me but then what should the test do?

TestFoundation/TestNSNumber.swift
johnno1962: This is a strange problem indeed which I don’t know the solution for other than the change you see. It doesn’t seem to be 32 bit specific but something not quite right about the compiler. I’m seeing warnings I don’t expect and the errors I had to fix with this change.

Sources/TestNSNumber.swift:1148:50: warning: '-' is deprecated: Please use explicit type conversions or Strideable methods for mixed-type arithmetics.
        XCTAssertEqual(NSNumber(value: Int64.max - 1).stringValue, "9223372036854775806")
                                                 ^
Sources/TestNSNumber.swift:1090:53: error: arithmetic operation '4294967295 + 4294967295' (on unsigned 32-bit integer type) results in an overflow
            XCTAssertEqual(NSNumber(value: UInt.max - 1).stringValue, "4294967294")
                                           ~~~~~~~~ ^ ~

The compiler is basically built right out of the swift-4.1-branch with only three minor unrelated patches.

CoreFoundation/PlugIn.subproj/CFBundle_Main.c
johnno1962: There is no useful equivalent to a “bundle directory” on Android as you would think of it on iOS. App resources are held compressed and the only means of access to them is through a stream which isn’t useful for example in specifying a “cacerts.pem” file for lib curl to use for accessing https: sites. The nearest equivalent is the “cacheDir” or “internalDataPath” as it is made available in C which you could consider the "Documents" and “tmp” dir from iOS all rolled into one. This is the only directory an application knows the absolute path for and has write access to as far as I know.

For bundle resources all you can do is copy your resource out of it’s stream into this directory on the Java side of things, as in, for example, [this code](https://github.com/SwiftJava/swift-android-kotlin/blob/master/app/src/main/java/com/example/user/myapplication/MainActivity.kt#L48) and press this directory into service as the bundle directory so the following code works in Swift:

```Swift
    func setCacheDir( cacheDir: String? ) {
        setenv( "TMPDIR", cacheDir!, 1 )

        // Required for SSL to work
        let pemFile = Bundle.main.path(forResource:"cacert", ofType:"pem")
        setenv( "URLSessionCertificateAuthorityInfoFile", pemFile, 1 )
```
Using the AAssetManager api would increase the surface area of foundation exposed to Android unnecessarily and would require support from the Java end of things to get hold of the activity instance to get hold of the AssetManager etc which isn’t straightforward. Far easier to just pass the path of the cacheDir to the Swift code as that is all it’s interested in and keep things simple. The only dependency foundation has at the moment is the __android_log_print function.

CoreFoundation/Preferences.subproj/CFPreferences.c
johnno1962: If you can accept using the cacheDirectory as the bundle directory then this follows rather than replicate the dependency on various environment variables.

CoreFoundation/Base.subproj/CFBase.h
johnno1962: I think it’s unlikely it’s actually 16 bits on Android, more likely there is no _swift_weak_rc at all but using 16 bits is a way to confine the changes required for Android. What I do know is that Swift ivars start 8 bytes into the instance so this is one way to model this.

CoreFoundation/PlugIn.subproj/CFBundle_Main.c
ephemer: `AAssetManager` is also only available from a prohibitively high API version compared to what we'd like to support. I couldn't find the exact version in a quick google search but the AAssetManager was only added to the NDK in 2016 from what I could see.

It's a valid point though, what happens if you just enter no path whatsoever? Logically this should default to a path relative to the currently running app's... bundle directory... but I have a vague suspicion that Android doesn't do that and instead just throws an exception / assertion failure of sorts. Have you tested this recently @johnno1962 ?

johnno1962: The default "bundle directory" in this PR is /data/local/tmp. How do you mean “specify no path"? I agree we really don’t want to go down the C route. The AAssetManager api is just a limited version of what is available from Java.

CoreFoundation/PlugIn.subproj/CFBundle.c
millenomi: FHS is the Filesystem Hierarchy Spec, but in this context it refers to bundles that store contents pursuant that spec: https://forums.swift.org/t/proposal-foundation-allowing-for-system-bundles-on-linux/7417/6

Android is not a FHS system and should not have those on.

CoreFoundation/PlugIn.subproj/CFBundle_Main.c
millenomi: I appreciate the discussion. My confusion is: how does the code under it, which defines the bundle path to be essentially dirname(executable), not work for the Android case?

CoreFoundation/Preferences.subproj/CFPreferences.c
millenomi: The semantics of a bundle directory is that it, loosely, is where the executable is. I assume that is not writable on Android, and the API contract for CFBundle is that bundles may be read-only. No, this doesn't work.

What is a correct path for storing _user_-specific data on Android? Note that on FHS systems we are falling back to `$HOME/.local` (per [XDG](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html)); is that good on Android as well, or is there a better Android equivalent?

Foundation/Bundle.swift
millenomi: The main bundle is fetched via `Bundle.main`, though. How is this path exercised by that?

TestFoundation/TestFileManager.swift
millenomi: If this only fails on Android, sure — it will be more noise if anyone sets up CI for this, though. In general, on platforms we CI, we'd just `#if` the test content out with an explanatory comment.

CoreFoundation/Base.subproj/CFBase.h
millenomi: If the layout doesn't include a weak retain count, something is deeply, deeply weird in compiler land.

CoreFoundation/PlugIn.subproj/CFBundle_Main.c
johnno1962: It comes up with /system/bin which is no use in practical terms.

CoreFoundation/Preferences.subproj/CFPreferences.c
johnno1962: The only writable directory known to an android app is  the “cacheDir”.

Foundation/NSString.swift
johnno1962: The semantics do not diverge. The Android code is exactly equivalent to the the existing implementation.

Foundation/Bundle.swift
johnno1962: Alamofire calls Bundle(for: aClass)

TestFoundation/TestFileManager.swift
johnno1962: ok, I’ll remove the XCFail as there is nothing that can be done about this.

CoreFoundation/Base.subproj/CFBase.h
johnno1962: ¯\_(ツ)_/¯

Foundation/NSCFArray.swift
johnno1962: This is resolved - see below.

TestFoundation/TestNSNumber.swift
spevans: You may want to add a comment for this line (or jira if there is one) as otherwise it may get reverted by someone 'cleaning up' the tests in the future.

johnno1962: https://bugs.swift.org/browse/SR-7469

johnno1962:
Dunno. Why did Hillary climb Everest? Why did the chicken cross the road? Why do people learn Haskel? I think the main idea is to open the possibility of code sharing across the two platforms.

spevans:
@johnno1962 Regarding `test_fileSystemAttributes()`, it might be best to mark the original function unavailable as you suggest above. Its a bit ugly but I think you can use the following:

```
#if os(Android)
@available(*, unavailable, message: "Unsuppported on this platform")
   NSUnsupported()
}
#else
  .
  .
  .
}
#endif
```

johnno1962:
Thanks @spevans this works I’ll add it to the next commit.

johnno1962:
Just an observation here.. I’m sure you’re wondering @parkera why all this padding (8 bytes extra to instances ), 4 bytes extra to CF structs is necessary when it wasn’t for last year’s ~4.0 release or for Linux. I’ve done a bit of digging and the odd thing is, while Swift ivars start at offset 8 from self in binaries built with the new swift-4.1-branch compiler they were 12 bytes from self using the last release which is more what you would expect. This, along with the extra 4 bytes in CF structs to align the now int64_t _cfinfoa (a.k.a CFInfo) which used to be char[8] result in the instances being smaller than the CFstructs which use them. Is there a friendly local compiler bod we can consult at this stage whether this offset change is explicable?

parkera:
If the swift runtime changed the layout of that type, then we need to change the layout of our matching struct to match on all platforms.

johnno1962:
This is what changed: https://github.com/SwiftJava/swift/commit/ae1c9849200b7c72d08eb7f329109f73d97b80af, mystery solved. strong count is 22 bits, weak is 7 bits

parkera:
Can we start a separate PR just to fix up the size of the ref count for CFBase then?

johnno1962:
Now the memory model is tidied up in #1525 I’ve been able to tidy up this PR by removing the messy padding of allocations and the tests run through fine now on Android 5.1.1 though 7.0. Ready to merge.

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -403,6 +403,8 @@ static inline _Bool _withStackOrHeapBuffer(size_t amount, void (__attribute__((n
 static inline int _direntNameLength(struct dirent *entry) {
 #ifdef _D_EXACT_NAMLEN  // defined on Linux
     return _D_EXACT_NAMLEN(entry);
+#elseif DEPLOYMENT_TARGET_ANDROID
```
spevans: @johnno1962 Build is failing, this needs to be `#elif`

TestFoundation/TestNSNumber.swift
parkera: I think we still need @spevans' suggested comment here

parkera:
Looking pretty good, just a few minor things and we can get this merged.

CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -440,7 +440,11 @@ static CFURLRef _CFPreferencesURLForStandardDomainWithSafetyLevel(CFStringRef do
     CFURLRef theURL = NULL;
     CFAllocatorRef prefAlloc = __CFPreferencesAllocator();
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_LINUX || DEPLOYMENT_TARGET_WINDOWS
+#if DEPLOYMENT_TARGET_ANDROID
```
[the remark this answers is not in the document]
johnno1962: In the current build script both are set. For android the existing conditionals for Linux apply.

CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -25,10 +25,10 @@
 #include <unistd.h>
 #include <sys/stat.h>
 #include <sys/types.h>
-#if __has_include(<sys/syslog.h>)
-#include <sys/syslog.h>
-#elif __has_include(<syslog.h>)
+#if __has_include(<syslog.h>)
 #include <syslog.h>
+#else
+#include <sys/syslog.h>
```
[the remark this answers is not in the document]
johnno1962: This is just a minor portability fix. We need a syslog.h but probably not both and if you include sys/syslog.h on android which has syslog.h you can end up including the host os one.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -813,6 +813,14 @@ static CFTimeZoneRef __CFTimeZoneCreateSystem(void) {
         CFRelease(name);
         if (result) return result;
     }
+#if DEPLOYMENT_TARGET_ANDROID
+    // Timezone database by name not available on Android.
+    // Approximate with gmtoff - could be general default.
+    struct tm info;
+    time_t now = time(NULL);
+    if (NULL != localtime_r(&now, &info))
```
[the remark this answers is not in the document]
johnno1962: Will do in the next —amend if we can reach agreement on what to do about bundles.

TestFoundation/TestNSNumber.swift
johnno1962: Will include in next push

CoreFoundation/PlugIn.subproj/CFBundle.c
```diff
@@ -714,7 +714,7 @@ static CFBundleRef _CFBundleCreate(CFAllocatorRef allocator, CFURLRef bundleURL,
 
     bundle->_url = newURL;
     
-#if !DEPLOYMENT_RUNTIME_OBJC && !DEPLOYMENT_TARGET_WINDOWS
+#if !DEPLOYMENT_RUNTIME_OBJC && !DEPLOYMENT_TARGET_WINDOWS && !DEPLOYMENT_TARGET_ANDROID
```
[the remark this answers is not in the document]
johnno1962: I’ve gone through the code again and gated anything I could find about FHS out for Android.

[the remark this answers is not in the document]
johnno1962: There seems to be a bug in github where I can’t reply directly to your next comment about path(toResource:ofType:) so I’ll add a note here. One of the goals of the Android port is that code for iOS should run on Android and it’s likely that code will use the path(toResource:ofType:) pattern somewhere. On Android It’s up to the Java side of the app to copy resources required into the cache/tmp directory then the Swift/Foundation side can find these resources inside this directory as if it was a bundle so the code runs unmodified.

Foundation/Bundle.swift
```diff
@@ -53,8 +53,16 @@ open class Bundle: NSObject {
...
+    public convenience init(for aClass: AnyClass) {
+        self.init(path: Bundle.main.bundlePath)!
```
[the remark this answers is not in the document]
johnno1962: The goal is compatibility. If at all possible OSS code should run unmodified.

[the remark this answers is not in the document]
johnno1962: I think we need to get a bit of perspective here. The code currently executes NSUnimplemented() isn’t much use. This is a pragmatic Android-specific step up in functionality that returns “a bundle” if this method happens to be called rather than crashing the app.

CoreFoundation/PlugIn.subproj/CFBundle.c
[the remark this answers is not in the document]
millenomi: I appreciate the use case, but again, it breaks our contract that Bundle.main be where the resources for a specific bundle _actually are_ (not _potentially will be if the author is careful_), and that this be a directory that's either specific to the application or at worst contains existing application executable and resource data for that one application.

I outlined a workaround above — since apps need to write custom code to copy resources to some filesystem directory, it is no stretch, if they want `Bundle` API to fetch their own resources, to ask that they construct a `Bundle` instance to the copied data.

I will also reiterate that any library code using `Bundle(for:)` that builds with the Swift Package Manager and then asserts the returned bundle is exclusive to that library is essentially in undefined behavior territory, and should be fixed upstream.

Foundation/Bundle.swift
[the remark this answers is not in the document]
millenomi: Alamofire is relying on an assumption that isn't true on all platforms, and its OSS nature makes it trivial to propose a patch upstream. That's up to Alamofire to fix, not to Foundation to work around, and I think it's safe to say the core team is in broad agreement here.

CoreFoundation/PlugIn.subproj/CFBundle.c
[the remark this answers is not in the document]
millenomi: (Also — that breaks if multiple components on Android are using Swift, have equally-named resources or resource directories {which isn't that unlikely: I assume careless copying code could easily clobber an en.lproj folder}, and are running at the same time.)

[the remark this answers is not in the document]
johnno1962: Perhaps there just is a misunderstanding here. This is not a system shared directory. It is the app's home, combination Documents and tmp directory on iOS with a path something like: /data/data/com.example.user.myapplication/cache and is the only directory an app has read write access to I am aware of. There's no candidate for a bundle directory other than the approach I suggest of copying resources on Android as resources are not unpacked from their .apk (zip file) on demand. This is what the zip file for an Android app looks like:

```
Users-iMac:swift-corelibs-foundation user$ unzip -l ~/swift-android-kotlin/app/build/outputs/apk/app-debug.apk
Archive:  /Users/user/swift-android-kotlin/app/build/outputs/apk/app-debug.apk
  Length      Date    Time    Name
---------  ---------- -----   ----
     2788  00-00-1980 00:00   AndroidManifest.xml
     1211  00-00-1980 00:00   META-INF/CERT.RSA
    31429  00-00-1980 00:00   META-INF/CERT.SF
    31367  00-00-1980 00:00   META-INF/MANIFEST.MF
      246  01-01-1970 01:00   META-INF/kotlin-runtime.kotlin_module
       43  01-01-1970 01:00   META-INF/kotlin-stdlib-jre7.kotlin_module
     2090  01-01-1970 01:00   META-INF/kotlin-stdlib.kotlin_module
    46044  00-00-1980 00:00   classes.dex
      752  00-00-1980 00:00   classes2.dex
      926  01-01-1970 01:00   kotlin/annotation/annotation.kotlin_builtins
     3689  01-01-1970 01:00   kotlin/collections/collections.kotlin_builtins
      726  01-01-1970 01:00   kotlin/internal/internal.kotlin_builtins
    14202  01-01-1970 01:00   kotlin/kotlin.kotlin_builtins
     2296  01-01-1970 01:00   kotlin/ranges/ranges.kotlin_builtins
     4866  01-01-1970 01:00   kotlin/reflect/reflect.kotlin_builtins
 10933048  00-00-1980 00:00   lib/armeabi-v7a/libFoundation.so
   359812  00-00-1980 00:00   lib/armeabi-v7a/libXCTest.so
   661056  00-00-1980 00:00   lib/armeabi-v7a/libc++_shared.so
  2009280  00-00-1980 00:00   lib/armeabi-v7a/libcurl.so
   668356  00-00-1980 00:00   lib/armeabi-v7a/libdispatch.so
 25911876  00-00-1980 00:00   lib/armeabi-v7a/libscudata.so
  2045828  00-00-1980 00:00   lib/armeabi-v7a/libscui18n.so
  1418788  00-00-1980 00:00   lib/armeabi-v7a/libscuuc.so
  1079184  00-00-1980 00:00   lib/armeabi-v7a/libsqlite3.so
  5259060  00-00-1980 00:00   lib/armeabi-v7a/libswiftCore.so
    31020  00-00-1980 00:00   lib/armeabi-v7a/libswiftGlibc.so
   506816  00-00-1980 00:00   lib/armeabi-v7a/libswiftRemoteMirror.so
   383140  00-00-1980 00:00   lib/armeabi-v7a/libswiftSwiftOnoneSupport.so
  2810068  00-00-1980 00:00   lib/armeabi-v7a/libswifthello.so
  1783196  00-00-1980 00:00   lib/armeabi-v7a/libxml2.so
      620  00-00-1980 00:00   res/anim-v21/design_bottom_sheet_slide_in.xml
      620  00-00-1980 00:00   res/anim-v21/design_bottom_sheet_slide_out.xml
...
```
What would you have me do? Mark the whole Bundle api unavailable? There is also no prospect of a OSS fix upstream to Alamofire as the code I’m using is a fork of a fork. If it were me I would make the fix I suggest apply to all Linux systems in the interests of compatibility. If I were to fix Alamofire there just be some other library which falls at this hurdle.

But we’ve discussed these issues a few times and we’re going around in circles. What is the minimum changes I have proposed you would have me remove to get this PR merged? I’ll have to maintain a fork (undesirable as it may be) to get what I need to get done done.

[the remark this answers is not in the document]
millenomi: The following changes are, I believe, at issue:

 - `Bundle(for:)` should not return an arbitrary bundle. It should stay callable. (My personal intent is to implement it on Linux as soon as resources allow.)

 - `Bundle.main` should not pick an arbitrary path whose forming substantially differs from other platforms. It should go on the existing fallback path unless that breaks existing _well-formed_ code.

 - `CFPreferences…` should not assume that any `CFBundle` path is writable or a correct location for user data. All paths returned by `CFBundle` must be treated as read-only.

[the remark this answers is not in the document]
millenomi: Also, someone should send Alamofire a patch, I believe.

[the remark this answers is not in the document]
johnno1962: OK.. how does it look now?

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

johnno1962:
Hi @millenomi, is were anything more you need me to do to get this merged? It would be good to get this tidied away.

johnno1962:
Thank @millenomi, Happy hunting with bridging.

PR #1519, opened 2018-04-16
FileManager: Reduce memory usage of copyItem(atPath:toPath:)

spevans:
- Enumerate a directory directly instead of using
  subpathsOfDirectory(atPath:) and creating an array.

- When copying a regular file, use a buffer sized to st_blksize instead
  of doing a whole file read/write.


spevans:
@swift-ci please test

spevans:

PR #1520, opened 2018-04-17
Unwrap potentially boxed attributes returned by NSAttributedString.attribute

ephemer:
When adding an attribute to a NSMutableAttributedString with `.addAttribute`, the Swift value provided is potentially boxed in a `_SwiftValue` via `_SwiftValue.store(value)`, as seen here: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSAttributedString.swift#L345

The boxed value is never unboxed again though on retrieval. And given that `_SwiftValue` is a Foundation-internal type, it's impossible for an end-user to ever access the real raw value again (`.attribute(at: index)` always just returns the boxed version).

The current change aims to fix this (see Changed Files in this PR for details) by defensively unboxing any individual attributes returned. `_SwiftValue.fetch(value)` checks whether the value needs unboxing and is thus safe to use on any `AnyObject(?)`.

The only question is whether we should extend this change to multiple attributes (here, probably: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSAttributedString.swift#L254), or whether there's a more foolproof way about this.

### DIFF
--- a/Foundation/NSAttributedString.swift
+++ b/Foundation/NSAttributedString.swift
@@ -251,7 +251,9 @@ private extension NSAttributedString {
                 guard let stringKey = (key as? NSString)?._swiftObject else {
                     continue
                 }
-                results[NSAttributedStringKey(stringKey)] = _SwiftValue.fetch(value)
+
+                let unwrappedValue = _SwiftValue.fetch(value as? AnyObject) ?? value
+                results[NSAttributedStringKey(stringKey)] = unwrappedValue
             }
             
             // Update effective range and return the results
--- a/Foundation/NSAttributedString.swift
+++ b/Foundation/NSAttributedString.swift
@@ -263,7 +263,7 @@ private extension NSAttributedString {
     
     func _attribute(_ attrName: NSAttributedStringKey, atIndex location: Int, rangeInfo: RangeInfo) -> Any? {
         var cfRange = CFRange()
-        return withUnsafeMutablePointer(to: &cfRange) { (cfRangePointer: UnsafeMutablePointer<CFRange>) -> AnyObject? in
+        let attribute = withUnsafeMutablePointer(to: &cfRange) { (cfRangePointer: UnsafeMutablePointer<CFRange>) -> AnyObject? in
             // Get attribute value using CoreFoundation function
             let attribute: AnyObject?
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci please test

parkera:
You're right that we should not be exposing `_SwiftValue` to callers. We must have this box/unbox pattern elsewhere as well.

Adding your test case to our XCTests seems reasonable.

ephemer:
@parkera thank you for your speedy review. Can you run the tests again? The test that failed last run made little sense to me as written: at the top of the very same test that failed we were trying to cast the attribute's value to `String`, and below we were trying to cast it to `NSString` (which now failed).

As far as I'm aware we're not able to cast String to NSString on Linux as we are on Darwin. So it seems to me the way the test was written was bound to fail, or was relying on broken behaviour. Or have I misunderstood something here? For now I've changed the expectation to be internally consistent (it now always checks for a `String`).

spevans:
@swift-ci please test

ephemer:
@spevans sorry, I got too hasty with that one. I have pushed an update but am not able to test the build locally until tomorrow. It should work so you could try again if you're feeling lucky, otherwise I'll write again tomorrow.

spevans:
@swift-ci please test

alblue:
@swift-ci please test

ephemer:
@spevans I can't build the current master on my machine with Xcode 9.3 - I have a bunch of errors like `Data does not conform to protocol 'Collection'`.

There was a problem with the build yesterday because of a typo. It should work now but I unfortunately don't have time to fix the Foundation build errors to get it working here. Can you try again?

spevans:
If you are building with Xcode, may sure to download the latest nightly snapshot of the toolchain, available at https://swift.org/download/#snapshots.

Also, remember to checkout swift-corelibs-xctest and open `Foundation.xcworkspace` so that it also builds XCTest at the same time.

millenomi:
@swift-ci please test

millenomi:
The fix is minimal, so I commandeered this.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
I've slept on this; since `NSString(…) as? String` now works on Linux, I'd rather have returning values be a constant time operation and the downstream casts perform the bridging like on Darwin. I'm closing it, but if this is in error, please ping me.

PR #1522, opened 2018-04-18
Implement NSTextCheckingResult.range(withName:)

nevil:
- Added the missing range(withName:) and corresponding CF function.
The implementation relies on uregex_groupNumberFromName which is
available as a draft API from ICU 55.

- Added tests related to named capture groups.

### DIFF
--- a/CoreFoundation/String.subproj/CFRegularExpression.h
+++ b/CoreFoundation/String.subproj/CFRegularExpression.h
@@ -57,6 +57,7 @@ _CFRegularExpressionRef _Nullable _CFRegularExpressionCreate(CFAllocatorRef allo
 void _CFRegularExpressionDestroy(_CFRegularExpressionRef regex);
 
 CFIndex _CFRegularExpressionGetNumberOfCaptureGroups(_CFRegularExpressionRef regex);
+CFIndex _CFRegularExpressionGetCaptureGroupNumberWithName(_CFRegularExpressionRef regex, CFStringRef groupName);
 void _CFRegularExpressionEnumerateMatchesInString(_CFRegularExpressionRef regexObj, CFStringRef string, _CFRegularExpressionMatchingOptions options, CFRange range, void *_Nullable context, _CFRegularExpressionMatch match);
 
 CFStringRef _CFRegularExpressionGetPattern(_CFRegularExpressionRef regex);
--- a/CoreFoundation/String.subproj/CFRegularExpression.c
+++ b/CoreFoundation/String.subproj/CFRegularExpression.c
@@ -171,6 +171,41 @@ CFIndex _CFRegularExpressionGetNumberOfCaptureGroups(_CFRegularExpressionRef reg
     return (CFIndex)uregex_groupCount(regex->regex, &errorCode);
 }
 
+CFIndex _CFRegularExpressionGetCaptureGroupNumberWithName(_CFRegularExpressionRef regex, CFStringRef groupName) {
+    UniChar stackBuffer[STACK_BUFFER_SIZE], *nameBuffer = NULL;
+    Boolean freeNameBuffer = false;
+
+    CFIndex nameLength = CFStringGetLength(groupName);
+    UErrorCode errorCode = U_ZERO_ERROR;
+
+    nameBuffer = (UniChar *)CFStringGetCharactersPtr(groupName);
+    if (!nameBuffer) {
+        if (nameLength <= STACK_BUFFER_SIZE) {
+            nameBuffer = stackBuffer;
+            CFStringGetCharacters(groupName, CFRangeMake(0, nameLength), nameBuffer);
+        } else {
+            nameBuffer = (UniChar *)malloc(sizeof(UniChar) * nameLength);
+            if (nameBuffer) {
+                CFStringGetCharacters(groupName, CFRangeMake(0, nameLength), nameBuffer);
+                freeNameBuffer = true;
+            } else {
+                HALT;
+            }
+        }
+    }
+
+    CFIndex idx = uregex_groupNumberFromName(regex->regex, nameBuffer, nameLength, &errorCode);
+    if (U_FAILURE(errorCode) || idx < 0) {
+        idx = kCFNotFound;
+    }
+
+    if (freeNameBuffer) {
+        free(nameBuffer);
+    }
+
+    return idx;
+}
+
 struct regexCallBackContext {
     void *context;
     void (*match)(void *context, CFRange *ranges, CFIndex count, _CFRegularExpressionMatchingFlags flags, Boolean *stop);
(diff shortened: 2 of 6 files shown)


spevans:
@swift-ci please test

nevil:
I realized that I'm not sure what to do with the availability story.
Should I do the following:
1) add ```API_AVAILABLE(macosx(10.13), ios(11.0), watchos(4.0), tvos(11.0))``` to the CF API
2) add ```@available(macOS 10.13, iOS 11.0, watchOS 4.0, tvOS 11.0, *)``` to the NSTextCheckingResult API
3) Condition the unit tests based on availability?

parkera:
@swift-ci please test

nevil:
@parkera Seems like the test wasn't run or the result not reported back?

spevans:
@swift-ci please test

nevil:
The build passed but the status seems to be reported for an old commit that no longer exist (0aedad0d4c122b2a59598bd0b6d8c1c14ff7c9cb) ... 
https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/918/

nevil:
Would a kind soul please rerun the test? :-)

spevans:
@swift-ci please test

nevil:
@parkera 
Is there anything more I should do for this PR?

Not sure what the normal review time frames are...

CoreFoundation/String.subproj/CFRegularExpression.h
```diff
@@ -57,6 +57,7 @@ _CFRegularExpressionRef _Nullable _CFRegularExpressionCreate(CFAllocatorRef allo
 void _CFRegularExpressionDestroy(_CFRegularExpressionRef regex);
 
 CFIndex _CFRegularExpressionGetNumberOfCaptureGroups(_CFRegularExpressionRef regex);
+CFIndex _CFRegularExpressionGetCaptureGroupNumberWithName(_CFRegularExpressionRef regex, CFStringRef groupName) API_AVAILABLE(macosx(10.13), ios(11.0), watchos(4.0), tvos(11.0));
```
millenomi: Why availability annotations? Internal calls do not need them, and in this case they would also be inaccurate (this call isn’t on any of the mentioned OSes.)

Just remove them, here and in the Swift code?

spevans:
@swift-ci please test

nevil:
@spevans Thanks for starting the build.
Unfortunately it really looks like the swift CI infrastructure do not like git rebases...
The current commit is 4b22b29 but the build is running on e167c459c191e6, which was the commit before the rebase...

https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/961/console

spevans:
There is just one issue I see with this change and its that Ubuntu 14.04 uses ICU 52 and so doesn't have the `uregex_groupNumberFromName` function from what I can see (checking the libs on my 14.04 install).

Is it possible to use the `U_ICU_VERSION_MAJOR_NUM ` (defined as 52 on 14.04 and 55 on 16.04) to mark this function as unavailable on the older version (or maybe there is an alternative solution).

I think this will cause problems if merged as it will then break the 14.04 package build (although it is current broken due to a different 14.04/ ICU issue anyway, see https://github.com/apple/swift-corelibs-foundation/pull/1537)

spevans:
@nevil swift-ci looks fine, its just misleading with the first comment. 
There is a log entry that says 
```
Commit message: "Merge 4b22b2929e253bf50564498cf4e208443e8559d0 into 55bb988867eb5c8e174d67a0f91a15d9efdd072a"
```
 which means its merging in the correct version.
Rebases and force pushes work fine, I use them all the time without issue.

nevil:
@spevans You're correct. Looks like the commit hash mismatch is just for the GitHub build status reporting...

I could of course use U_ICU_VERSION_MAJOR_NUM (or some other flag) to conditionally enable the _CFRegularExpressionGetCaptureGroupNumberWithName, but the question is if this is an acceptable workaround until all supported Linux versions use the same ICU library?

It would introduce a version difference that can not be checked through the use of availability annotation checks...

spevans:
You could do something like the following:

Define `HAS_ICU_GROUP_NUMBER_FROM_NAME` based on the ICU version and conditionally compile the support functions, then add:

```
#if !HAS_ICU_GROUP_NUMBER_FROM_NAME
@available(*, unavailable, message: "Unuspported with this version of ICU")
override func range(withName name: String) -> NSRange {
  NSUnsupported()
}
#else
override func range(withName name: String) -> NSRange {
.
.
.
}
#endif
```

millenomi:
@nevil What do you think about @spevans's suggestion?

nevil:
@millenomi 
I will try to implement something like what he suggested.
Unfortunately I haven't had time to study the build system until know.

nevil:
@millenomi 
My intention was to complete the task ASAP, but I had the problem that my local machine doesn't have enough RAM to do a complete linux build in a VM. I thought I would have access to a better machine soon, but didn't and then time passed quickly...

Sorry for that!
I still have an interest and would like to complete the PR.

Is it best to try to get this in with build time version detection of ICU (as suggested here 
https://github.com/apple/swift-corelibs-foundation/pull/1522#issuecomment-387017925) or wait for 

https://forums.swift.org/t/building-libicu-as-part-of-the-linux-swift-build/16526/8 to settle?

spevans:
@nevil The PR to always build the latest version of ICU for linux should hopefully be ready in the next day or so - I just have to fix one LLDB test - so you can probably leave the version detection out now. In fact it will be good to since it should catch builds using the old version since the symbols won't be defined.

nevil:
@spevans Great. Thank you very much!

I will rebase this PR on the latest master branch in preparation for your PR to be merged.

nevil:
Nothing had really changed in the touched files so the rebase was straight forward.

nevil:
Waiting for https://github.com/apple/swift/pull/19860

nevil:
@spevans 
Now that https://github.com/apple/swift/pull/19860 is merged, my understanding is that we will always use a compatible ICU version.
Is there anything build related I should consider in this PR?

spevans:
@swift-ci test

nevil:
@millenomi
I think the build passed.
Please review again when you have time.

Thanks!

nevil:
@millenomi 
Hi, I understand that you are busy with Swift 5.
Just read the latest https://nshipster.com/swift-regular-expressions/ and noticed that named capture groups are mentioned...

Any chance of reviewing again?

spevans:
@swift-ci please test

parkera:
This looks good to me.

parkera:
Sorry this took so long and thank you for the contribution!

nevil:
@parkera 
We all have lots of things to do :-)
Thank you for merging!

PR #1523, opened 2018-04-18
Replace XCTAssertNotNil with enhanced tests on TestDate and TestTimer

otaviolima:
Similar to #1516 but in cases where simply removing the `XCTAssertNotNil` would not bring any benefit or even harm the test suite.

### DIFF
--- a/TestFoundation/TestDate.swift
+++ b/TestFoundation/TestDate.swift
@@ -42,26 +42,27 @@ class TestDate : XCTestCase {
     
     func test_BasicConstruction() {
         let d = Date()
-        XCTAssertNotNil(d)
+        XCTAssert(d.timeIntervalSince1970 != 0)
+        XCTAssert(d.timeIntervalSinceReferenceDate != 0)
     }
 
     func test_descriptionWithLocale() {
         let d = NSDate(timeIntervalSince1970: 0)
         XCTAssertEqual(d.description(with: nil), "1970-01-01 00:00:00 +0000")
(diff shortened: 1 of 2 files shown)


TestFoundation/TestDate.swift
```diff
@@ -42,26 +42,27 @@ class TestDate : XCTestCase {
...
-        XCTAssertNotNil(d)
+        XCTAssert(d.timeIntervalSince1970 != 0)
+        XCTAssert(d.timeIntervalSinceReferenceDate != 0)
     }
 
     func test_descriptionWithLocale() {
         let d = NSDate(timeIntervalSince1970: 0)
         XCTAssertEqual(d.description(with: nil), "1970-01-01 00:00:00 +0000")
-        XCTAssertNotNil(d.description(with: Locale(identifier: "ja_JP")))
+        XCTAssertEqual(d.description(with: Locale(identifier: "en_US")), "Thursday, January 1, 1970 at 12:00:00 AM Greenwich Mean Time")
```
parkera: We explicitly did not want to test the exact string output here because it can change from time to time due to formatting differences.

otaviolima: Or perhaps we could use `en_US_POSIX` locale as it should not change it's formatting.

parkera: I'm fine with checking that `isEmpty` is false

otaviolima: Done

parkera:
@swift-ci test and merge

PR #1525, opened 2018-04-18
Cater for new Swift reference count representation

johnno1962:
The representation of reference counts for 32 bit systems has changed https://github.com/apple/swift/commit/ae1c9849200b7c72d08eb7f329109f73d97b80af and needs to be adjusted in the matching structs in CoreFoundation (__CFRuntimeBase & __CFConstStr). Swift ivars now start at offset 8 bytes from self on 32 bit systems. Part of https://github.com/apple/swift-corelibs-foundation/pull/1518


CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -449,11 +449,13 @@ CF_EXPORT double kCFCoreFoundationVersionNumber;
 #endif
 
 #if __LLP64__
+typedef uint32_t _CFSwiftRCCount;
```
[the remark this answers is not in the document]
parkera: The reason is that we're adding a new API and name here, and for all platforms - even those that don't use the Swift RC pattern.

gparker42:
A better placeholder for Swift's data would be a single uintptr_t on all architectures. The refcount storage is certainly not divided into two equal-size fields as described here; better to not tease the possibility of reading useful information from them.

CoreFoundation/Base.subproj/CFRuntime.h
```diff
@@ -193,8 +193,8 @@ CF_EXPORT void _CFRuntimeUnregisterClassWithTypeID(CFTypeID typeID);
 typedef struct __CFRuntimeBase {
     // This matches the isa and retain count storage in Swift
     uintptr_t _cfisa;
-    uint32_t _swift_strong_rc;
-    uint32_t _swift_weak_rc;
+    _CFSwiftRCCount _swift_strong_rc;
+    _CFSwiftRCCount _swift_weak_rc;
```
parkera: So here I would write:

```
     uintptr_t _cfisa;
#if __LLP64__
    uint32_t _swift_strong_rc;
    uint32_t _swift_weak_rc;
#else
    int16_t _swift_strong_rc;
    int16_t _swift_weak_rc;
#endif
    _Atomic(uint64_t) _cfinfoa;
```

and so forth, for the others

parkera:
@gparker42 thanks, even better

CoreFoundation/Base.subproj/CFBase.h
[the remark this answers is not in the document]
johnno1962: Ammended.

johnno1962:
How does it look now?

millenomi:
LGTM. @parkera?

Foundation/NSCFString.swift
```diff
@@ -61,18 +61,17 @@ internal final class _NSCFConstantString : _NSCFString {
...
     }
 
     private var _lenOffset : Int {
         // FIXME: Split expression as a work-around for slow type
         //        checking (tracked by SR-5322).
-        let offTemp1 = MemoryLayout<OpaquePointer>.size + MemoryLayout<Int32>.size
-        let offTemp2 = MemoryLayout<Int32>.size + MemoryLayout<_CFInfo>.size
+        let offTemp1 = MemoryLayout<OpaquePointer>.size + MemoryLayout<uintptr_t>.size
+        let offset = offTemp1 + MemoryLayout<_CFInfo>.size
         return offTemp1 + offTemp2 + MemoryLayout<UnsafePointer<UInt8>>.size
```
gparker42: This doesn't compile, does it?

johnno1962: yeah, noticed that.. should be ok now

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -412,11 +412,11 @@ CF_EXPORT void *__CFConstantStringClassReference[];
 #endif
 
 #define CONST_STRING_DECL(S, V) \
-const struct __CFConstStr __##S CONST_STRING_SECTION = {{(uintptr_t)&__CFConstantStringClassReference, _CF_CONSTANT_OBJECT_STRONG_RC, 0, 0x000007c8U}, (uint8_t *)(V), sizeof(V) - 1}; \
+const struct __CFConstStr __##S CONST_STRING_SECTION = {{(uintptr_t)&__CFConstantStringClassReference, _CF_CONSTANT_OBJECT_STRONG_RC, 0x000007c8U}, (uint8_t *)(V), sizeof(V) - 1}; \
```
gparker42: Do you need to change the definition of `_CF_CONSTANT_OBJECT_STRONG_RC`? My guess is that the bits generated here will be wrong if you do not.

johnno1962: This will work on little-endian only.. are there are any big-endian Swift ports?

parkera: I think the point of splitting it into strong/weak here was so we could set that pinned flag. There may now be a better way.

johnno1962: The Android tests run through ok. Are you saying there is still something to change?

gparker42: I don't think you were setting the pinned flag correctly before. The pinned flag is the LSB of that `uintptr_t` on all platforms.

I don't think you are setting the refcount correctly now. The strong refcount is stored differently on 32-bit versus 64-bit, but on all platforms it is biased: a strong refcount of 1 is represented as bit value 0.

johnno1962: #define _CF_CONSTANT_OBJECT_STRONG_RC (_CF_SWIFT_RC_PINNED_FLAG) then?

spevans:
@swift-ci please test

alblue:
@swift-ci please test

parkera:
once @gparker42 approves (since it's his struct we are messing with) we are good to merge

millenomi:
Actually bypassing swift-ci since it's not starting.

johnno1962:
Thanks @millenomi et all.

PR #1526, opened 2018-04-19
Adopt ‘as’ bridging on Linux.

millenomi:
This patch adopts 'as' bridging on Linux. It requires https://github.com/apple/swift/pull/16022 — see that PR for more details.

### DIFF
--- a/Foundation/AffineTransform.swift
+++ b/Foundation/AffineTransform.swift
@@ -476,7 +476,7 @@ open class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
     }
 }
 
-extension AffineTransform : _ObjectTypeBridgeable {
+extension AffineTransform : _ObjectiveCBridgeable {
     public static func _isBridgedToObjectiveC() -> Bool {
         return true
     }
--- a/Foundation/Bridging.swift
+++ b/Foundation/Bridging.swift
@@ -12,27 +12,8 @@
 
 import CoreFoundation
 
-// Support protocols for casting
-public protocol _ObjectBridgeable {
-    func _bridgeToAnyObject() -> AnyObject
-}
-
 public protocol _StructBridgeable {
     func _bridgeToAny() -> Any
-}
-
-/// - Note: This is a similar interface to the _ObjectiveCBridgeable protocol
-public protocol _ObjectTypeBridgeable : _ObjectBridgeable {
-    associatedtype _ObjectType : AnyObject
-    
-    func _bridgeToObjectiveC() -> _ObjectType
-    
-    static func _forceBridgeFromObjectiveC(_ source: _ObjectType, result: inout Self?)
-    
-    @discardableResult
-    static func _conditionallyBridgeFromObjectiveC(_ source: _ObjectType, result: inout Self?) -> Bool
-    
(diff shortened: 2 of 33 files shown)


Foundation/AffineTransform.swift
```diff
@@ -476,7 +476,7 @@ open class NSAffineTransform : NSObject, NSCopying, NSSecureCoding {
     }
 }
 
-extension AffineTransform : _ObjectTypeBridgeable {
+extension AffineTransform : _ObjectiveCBridgeable {
```
millenomi: The main change is removing the parallel type `_ObjectTypeBridgeable` and replacing it with the newly-exposed-on-Linux `_ObjectiveCBridgeable`.

Foundation/NSError.swift
```diff
@@ -1378,3 +1378,14 @@ extension POSIXError {
     /// Interface output queue is full.
     public static var EQFULL: POSIXError.Code { return .EQFULL }
 }
+
+public // COMPILER_INTRINSIC
+func _convertNSErrorToError(_ error: NSError?) -> Error {
```
millenomi: Converting `Error` to `NSError` and back is in scope, but support is WIP.

parkera:
Ultimately we should also change `NSJSONSerialization` to create the `NS` types to finally fix the `NSNumber` casting problem described in the forum post. I think it can be a follow-up to this though.

`NSPropertyListSerialization` can probably lose its eager conversion as well.

parkera:
Looks good!

millenomi:

benlangmuir:
The corresponding compiler change in had to be reverted.

benlangmuir:
I canceled the "and merge" build in CI so that this wouldn't be automatically merged.  I think the underlying linux test will continue though.

millenomi:
On hold as we figure out the 14.04 issues on the compiler patch.

millenomi:
Please test with the following:
https://github.com/apple/swift/pull/16736

@swift-ci please test

millenomi:
Please test with the following:
https://github.com/apple/swift/pull/16736

@swift-ci please test

millenomi:
Please test with the following:
https://github.com/apple/swift/pull/16736

@swift-ci please test

millenomi:
Simultaneous merging with https://github.com/apple/swift/pull/16736 imminent…

Foundation/Bridging.swift
```diff
@@ -108,10 +89,8 @@ internal final class _SwiftValue : NSObject, NSCopying {
     static func store(_ value: Any) -> NSObject {
         if let val = value as? NSObject {
             return val
-        } else if let val = value as? _ObjectBridgeable {
-            return val._bridgeToAnyObject() as! NSObject
         } else {
-            return _SwiftValue(value)
+            return (value as AnyObject) as! NSObject
```
zayass: Looks like this change leads to crash in cases like this:
```
class A {}
let array: NSArray = [A()]
```

zayass: @millenomi Can you look into this https://github.com/apple/swift-corelibs-foundation/pull/2500/files ?

PR #1529, opened 2018-04-20
Require gold on linux and freebsd by default.

gottesmm:
This commit just changes the default to require gold. The user can still override from the command line if they choose to.

The reason why I am making this change is to match Swift's behavior here. There
really isn't any good reason to have separate linkers compiling separate parts
of the same build.

rdar://39456714

-----

In terms of review, my main interest is in am I missing anything in terms of integrating this fix into your build system.


gottesmm:
@swift-ci test

millenomi:

PR #1531, opened 2018-04-22
Package.swift for testing foundation

johnno1962:
Hi Apple,

As a step toward ci, this PR adds a Package.swift that can be used to build “TestFoundation”. This needs to relax the access control on the “internal” _ProgressFraction class to “public" for tests to compile as, as far as I understand @testable is not available on Linux.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,19 @@
+// swift-tools-version:4.0
+import PackageDescription
+
+let package = Package(
+    name: "TestFoundation",
+    products: [
+        .executable(name: "TestFoundation", targets: ["TestFoundation"])
+    ],
+    dependencies: [
+    ],
+    targets: [
+        .target(
+            name: "TestFoundation",
+            dependencies: [],
+            path: "TestFoundation",
+            exclude: ["XDGTestHelper.swift", "xdgTestHelper"]
+        ),
+    ]
+)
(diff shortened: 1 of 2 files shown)


Foundation/ProgressFraction.swift
```diff
@@ -9,18 +9,18 @@
 
 // Implementation note: This file is included in both the framework and the test bundle, in order for us to be able to test it directly. Once @testable support works for Linux we may be able to use it from the framework instead.
 
-internal struct _ProgressFraction : Equatable, CustomDebugStringConvertible {
-    var completed : Int64
-    var total : Int64
-    private(set) var overflowed : Bool
+public struct _ProgressFraction : Equatable, CustomDebugStringConvertible {
```
parkera: I know why you want to do this, but we can't make this public. What other options do we have?

Package.swift
```diff
@@ -0,0 +1,19 @@
+// swift-tools-version:4.0
```
parkera: So what does this get us? A package just for testing? What's the end goal here?

johnno1962: I’ve amended the PR to not expose _ProgressFraction but still build for Android.

millenomi:
Can you articulate what this buys us over `ninja test`?

millenomi:
Closing per request.

PR #1532, opened 2018-04-24
Truncate fix

nevil:
Originally reported by @DagAgren here https://github.com/apple/swift-corelibs-foundation/pull/1297

- Accepting truncating to any position, not just 0.
- Reposition to the truncate offset.
- Darwin throws exceptions on error so we use fatalError() instead of silently failing.


spevans:
@swift-ci please test

millenomi:

PR #1537, opened 2018-04-25
NumberFormatter: Disable test for currencyPlural that breaks on Ubuntu 14.04

spevans:
- Cause could be ICU issues.

This looks to be causing the build to fail here https://ci.swift.org/job/oss-swift-package-linux-ubuntu-14_04/1594/console

### DIFF
--- a/TestFoundation/TestNumberFormatter.swift
+++ b/TestFoundation/TestNumberFormatter.swift
@@ -38,7 +38,7 @@ class TestNumberFormatter: XCTestCase {
             ("test_scientificMinimumIntegerDigits", test_scientificMinimumIntegerDigits),
             ("test_spellOutMinimumIntegerDigits", test_spellOutMinimumIntegerDigits),
             ("test_ordinalMinimumIntegerDigits", test_ordinalMinimumIntegerDigits),
-            ("test_currencyPluralMinimumIntegerDigits", test_currencyPluralMinimumIntegerDigits),
+            // XFAIL: breaks on Ubuntu 14.04 probably ICU - ("test_currencyPluralMinimumIntegerDigits", test_currencyPluralMinimumIntegerDigits),
             ("test_currencyISOCodeMinimumIntegerDigits", test_currencyISOCodeMinimumIntegerDigits),
             ("test_currencyAccountingMinimumIntegerDigits", test_currencyAccountingMinimumIntegerDigits),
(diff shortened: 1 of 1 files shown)


spevans:
cc @shahmishal @itaiferber

itaiferber:
@spevans Does the whole test need to be commented out? This sounds bizarre to me but I do see other `.currency` and `.currencyPlural` tests whose contents are commented out referencing [SR-250](https://bugs.swift.org/browse/SR-250). (I guess the tests are so old the value was called `.CurrencyPluralStyle`.)

Even then, though — why would this be causing `27993 Illegal instruction     (core dumped) "$@"`?

spevans:
TBH its hard to tell but I can't really think of another way of testing it unless we could get ubuntu-14.04 CI added to the normal `swift-corelibs-foundation` `swift-ci test` on jenkins. (I don't have access to 14.04 locally)

spevans:
I was just guessing at ICU but I believe the version shipped with 14.04 is quite old and has had issues in the past as you have noted above.

parkera:
Do we have a way to continue to run this on newer ICU versions?

spevans:
Swift on Linux can be built whilst compiling a newer version of ICU using the `--libicu` options to `swift-build-impl`. I have also had success building swift against Apple's version of ICU (available at https://opensource.apple.com/tarballs/ICU/ICU-59152.0.1.tar.gz) which also fixed the issues in https://bugs.swift.org/browse/SR-250. One of these is probably the only solution to carry on supporting Ubuntu 14.04. Personally I would aim to use Apple's version and always build it as part of the swift build since it would mean the same version is used on all platforms.

alblue:
You could modify the implementation of the test function to look for a line `DISTRIB_CODENAME=trusty` in `/etc/lsb-release` and then just skip the contents of the test. That way it would continue to run for other platforms and other versions of Ubuntu.

parkera:
@spevans A long term goal I have is to get that Apple ICU imported into the swift tree, so that on Linux we can all use the same one regardless of which platform we're on. The stdlib could also depend on it.

shahmishal:
@parkera @itaiferber Can we merge this PR? We have not been able to release new toolchain since April 25th.

itaiferber:
@shahmishal We should merge this once CI passes

PR #1538, opened 2018-04-26
[RFC] Fix canceling and dependency resolution

zayass:
We found two major issues with OperationQueue which not present in Darwin implementation.
Tests for this to cases presented in this PR.

First
If you cancel some operation other operation dependent on it will be never executed.

Second
If dependency add before the dependent both and queue is serial it causes deadlock https://bugs.swift.org/browse/SR-5361

This PR contains fixes for both cases.
But the main concern is `_checkReadyness` function. It uses waiting with zero timeout to check is _depGroup locked.

Could anyone validate/comment this solution?

### DIFF
--- a/Foundation/Operation.swift
+++ b/Foundation/Operation.swift
@@ -285,7 +285,7 @@ internal struct _OperationList {
             }
         }
     }
-    
+
     mutating func dequeue() -> Operation? {
         if !veryHigh.isEmpty {
             return veryHigh.remove(at: 0)
@@ -327,7 +327,7 @@ open class OperationQueue: NSObject {
     }
     let queueGroup = DispatchGroup()
 #endif
-    
+
     var _operations = _OperationList()
 #if DEPLOYMENT_ENABLE_LIBDISPATCH
     internal var _concurrencyGate: DispatchSemaphore? {
@@ -353,17 +353,10 @@ open class OperationQueue: NSObject {
             } else {
                 effectiveName = "NSOperationQueue::\(Unmanaged.passUnretained(self).toOpaque())"
             }
-            let attr: DispatchQueue.Attributes
-            if maxConcurrentOperationCount == 1 {
-                attr = []
-                __concurrencyGate = DispatchSemaphore(value: 1)
-            } else {
-                attr = .concurrent
-                if maxConcurrentOperationCount != OperationQueue.defaultMaxConcurrentOperationCount {
-                    __concurrencyGate = DispatchSemaphore(value:maxConcurrentOperationCount)
-                }
+            if maxConcurrentOperationCount != OperationQueue.defaultMaxConcurrentOperationCount {
+                __concurrencyGate = DispatchSemaphore(value:maxConcurrentOperationCount)
             }
-            let queue = DispatchQueue(label: effectiveName, attributes: attr)
+            let queue = DispatchQueue(label: effectiveName, attributes: .concurrent)
             if _suspended {
                 queue.suspend()
             }
@@ -398,15 +391,16 @@ open class OperationQueue: NSObject {
         addOperations([op], waitUntilFinished: false)
     }
     
-    internal func _runOperation() {
-        if let op = _dequeueOperation() {
-            if !op.isCancelled {
-                op._waitUntilReady()
-                if !op.isCancelled {
-                    op.start()
-                }
-            }
+    internal func _dequeueReadyOperation() -> Operation? {
+        guard let op = _dequeueOperation() else {
+            return nil
         }
+
+        if !op.isCancelled {
+            op._waitUntilReady()
+        }
+
+        return op
     }
     
     open func addOperations(_ ops: [Operation], waitUntilFinished wait: Bool) {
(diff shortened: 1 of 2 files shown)


millenomi:
@swift-ci please test

Foundation/Operation.swift
```diff
@@ -168,6 +168,21 @@ open class Operation : NSObject {
 #endif
         _ready = true
     }
+
+    internal func _checkReadyness() -> Bool {
+        if _ready {
+            return true
+        }
+
+        #if DEPLOYMENT_ENABLE_LIBDISPATCH
```
millenomi: I don't think at this point that we have a viable setup wherein we ship Foundation without libdispatch, but this is for another patch to clean up, I believe.

Foundation/Operation.swift
```diff
@@ -168,6 +168,21 @@ open class Operation : NSObject {
 #endif
         _ready = true
     }
+
+    internal func _checkReadyness() -> Bool {
```
millenomi: typo: `_checkReadiness`.

Foundation/Operation.swift
```diff
@@ -168,6 +168,21 @@ open class Operation : NSObject {
...
+        if _ready {
```
millenomi: If a dependency is added to this operation, shouldn't it reset its readiness?

Foundation/Operation.swift
zayass: oh thanks

Foundation/Operation.swift
zayass: Good question
But now if `_ready` is set then operation will be immediately executed. So this if have no many sense.
Maybe this `if` should be removed to not confuse. And proper isReady mangement should be in different PR.

Foundation/Operation.swift
zayass: Yes, this is strange but I trying to save style :)

alblue:
@swift-ci please test

zayass:
Strange. Why test not even started 7 hours after request?

spevans:
Looks like there was a failure in `test_deadlock`, see https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/938/consoleFull#console-section-14

FYI, sometimes when ci.swift.org starts a test it doesn't reliably update GitHub that the test as started/finished

zayass:
Ok. I understand an issue.
And looks like found even better solution without `.wait(timeout: .now())`
I just move `_waitUnitilReady` away of `__concurrencyGate` and now it counting only actual execution, not waiting for readiness.

Just one controversial thing. Queue should be always concurrent

zayass:
And one more bad thing. Each `addOperation` call occupy thread till operation finishes. But this is unchanged behavior

spevans:
@swift-ci please test

millenomi:
@phausler Can you give this a look?

Foundation/Operation.swift
```diff
@@ -398,15 +391,16 @@ open class OperationQueue: NSObject {
...
-                    op.start()
-                }
-            }
+    internal func _dequeueReadyOperation() -> Operation? {
+        guard let op = _dequeueOperation() else {
+            return nil
+        }
+
+        if !op.isCancelled {
+            op._waitUntilReady()
```
phausler: this isnt exactly how it works, the dequeue should be re-triggered upon finished states. The current implementation is a bit simplified since we dont have KVO.

TestFoundation/TestOperationQueue.swift
```diff
@@ -178,6 +180,51 @@ class TestOperationQueue : XCTestCase {
...
+            expectation2.fulfill()
+        }
+
+        op1.addDependency(op2)
+
+        queue.addOperation(op1)
+        sleep(1)
+        queue.addOperation(op2)
+
+        waitForExpectations(timeout: 1)
```
phausler: the sleep 1 might play havoc with the timeout here btw ( would suggest reducing the sleep to something smaller and that way the test is quicker too...)

Foundation/Operation.swift
```diff
@@ -439,13 +433,16 @@ open class OperationQueue: NSObject {
             }
 
             let block = DispatchWorkItem(flags: .enforceQoS) { () -> Void in
-                if let sema = self._concurrencyGate {
-                    sema.wait()
-                    self._runOperation()
-                    sema.signal()
+                let op = self._dequeueReadyOperation()
```
phausler: this really isnt the way it works internally to NSOQ; In my spare time I should probably re-draft up a non-KVO variation again that avoids the semaphore. Since the Objective-C implementation does not use a semaphore and if there are no external observers it does not use KVO (it is lazy in that manner). And I think that might be transliteratable to Swift.

millenomi:
@zayass I know it's been a long while; are you still interested in chaperoning this patch?

alblue:
This seems to have gone stale, and has conflicts. I suggest we close it, and then later re-open if necessary.

zayass:
Replaced with more complete implementation https://github.com/apple/swift-corelibs-foundation/pull/1858

PR #1539, opened 2018-04-26
RecoverableError's resultHandler needs @escaping

itaiferber:
Resolves [SR-4405](https://bugs.swift.org/browse/SR-4405)/rdar://problem/36911792 by adding `@escaping` to `resultHandler`.

[This mirrors apple/swift#16182]

### DIFF
--- a/Foundation/NSError.swift
+++ b/Foundation/NSError.swift
@@ -239,7 +239,7 @@ public protocol RecoverableError : Error {
     /// This entry point is used for recovery of errors handled at a
     /// "document" granularity, that do not affect the entire
     /// application.
-    func attemptRecovery(optionIndex recoveryOptionIndex: Int, resultHandler handler: (_ recovered: Bool) -> Void)
+    func attemptRecovery(optionIndex recoveryOptionIndex: Int, resultHandler handler: @escaping (_ recovered: Bool) -> Void)
 
     /// Attempt to recover from this error when the user selected the
     /// option at the given index. Returns true to indicate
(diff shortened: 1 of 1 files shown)


itaiferber:
@swift-ci Please test

millenomi:
@itaiferber: `@escaping` is implied for optional closures, since they need to be stored in the `Optional<_>`.

Foundation/NSError.swift
```diff
@@ -147,7 +147,7 @@ open class NSError : NSObject, NSCopying, NSSecureCoding, NSCoding {
     internal typealias UserInfoProvider = (_ error: Error, _ key: String) -> Any?
     internal static var userInfoProviders = [String: UserInfoProvider]()
     
-    open class func setUserInfoValueProvider(forDomain errorDomain: String, provider: (/* @escaping */ (Error, String) -> Any?)?) {
+    open class func setUserInfoValueProvider(forDomain errorDomain: String, provider: (@escaping (Error, String) -> Any?)?) {
```
millenomi: This must drop the `@escaping` before it can build. Optional<X> where X is a closure type is always implicitly escaping.

itaiferber:
@millenomi Thanks — hadn't had the time to fix this up. I dropped that commit, though I wonder if changing `/* @escaping */` to `/* implicitly @escaping */` would make it clearer that this isn't something that we couldn't add before (which we've had in the past), but rather, this is something that happens to be true

What do you think?

millenomi:
We have that pattern nowhere, and I'm not sure it adds anything.

itaiferber:
@swift-ci Please test

ikesyo:
@swift-ci Please test

itaiferber:
@ikesyo Thanks! Completely forgot about this.

PR #1543, opened 2018-05-01
CoreFoundation: simplify Linux libdispatch handling

compnerd:
Rather than special case the Linux handling, use the more prevalent
codepath to enable libdispatch support.  NFCI


compnerd:
CC: @phausler

compnerd:
@swift-ci please test

parkera:
Seems fine

PR #1544, opened 2018-05-01
build: add CMake build system

compnerd:
Add a CMake based build system for CoreFoundation.  This helps moves to
a single unified build for swift and the corelibs.  It also makes it
easier to cross-compile.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,424 @@
+
+cmake_minimum_required(VERSION 3.4.3)
+
+list(APPEND CMAKE_MODULE_PATH
+     "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+
+project(Foundation
+        LANGUAGES
+          C)
+enable_testing()
+
+option(FOUNDATION_ENABLE_LIBDISPATCH "Enable GCD Support" YES)
+option(FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE "Path to libdispatch source" "")
+option(FOUNDATION_PATH_TO_LIBDISPATCH_BUILD "Path to libdispatch build" "")
+option(FOUNDATION_PATH_TO_XCTEST_BUILD "Path to XCTest build" "")
+
+find_package(CURL REQUIRED)
+find_package(ICU COMPONENTS uc i18n REQUIRED)
+find_package(LibXml2 REQUIRED)
+find_package(UUID REQUIRED)
+
+include(SwiftSupport)
+include(GNUInstallDirs)
+include(ExternalProject)
+
+ExternalProject_Add(CoreFoundation
+                    SOURCE_DIR
+                      ${CMAKE_CURRENT_SOURCE_DIR}/CoreFoundation
+                    CMAKE_COMMAND
+                      ${CMAKE_COMMAND}
+                    CMAKE_ARGS
+                      -DBUILD_SHARED_LIBS=NO
+                      -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+                      -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+                      -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+                      -DCMAKE_INSTALL_LIBDIR=usr/lib
+                      -DCF_DEPLOYMENT_SWIFT=YES
+                      -DCF_ENABLE_LIBDISPATCH=${FOUNDATION_ENABLE_LIBDISPATCH}
+                      -DCF_PATH_TO_LIBDISPATCH_SOURCE=${FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE}
+                      -DCF_PATH_TO_LIBDISPATCH_BUILD=${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD})
+ExternalProject_Get_Property(CoreFoundation install_dir)
+
+set(swift_optimization_flags)
+if(CMAKE_BUILD_TYPE MATCHES Release)
+  set(swift_optimization_flags -O)
+endif()
+
+set(deployment_enable_libdispatch)
+set(libdispatch_cflags)
+set(libdispatch_ldflags)
+if(FOUNDATION_ENABLE_LIBDISPATCH)
+  set(deployment_enable_libdispatch -DDEPLOYMENT_ENABLE_LIBDISPATCH)
+  set(libdispatch_cflags -I;${FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE};-I;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src/swift;-Xcc;-fblocks)
+  set(libdispatch_ldflags -L;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src;-ldispatch)
+endif()
+
+add_swift_library(Foundation
+                  MODULE_NAME
+                    Foundation
+                  MODULE_LINK_NAME
+                    Foundation
+                  MODULE_PATH
+                    ${CMAKE_CURRENT_BINARY_DIR}/swift/Foundation.swiftmodule
+                  OUTPUT
+                    ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}Foundation${CMAKE_SHARED_LIBRARY_SUFFIX}
+                  SOURCES
+                    Foundation/AffineTransform.swift
+                    Foundation/Array.swift
+                    Foundation/Boxing.swift
+                    Foundation/Bridging.swift
+                    Foundation/Bundle.swift
+                    Foundation/ByteCountFormatter.swift
+                    Foundation/Calendar.swift
+                    Foundation/CGFloat.swift
+                    Foundation/CharacterSet.swift
+                    Foundation/Codable.swift
+                    Foundation/Data.swift
+                    Foundation/DateComponentsFormatter.swift
+                    Foundation/DateComponents.swift
+                    Foundation/DateFormatter.swift
+                    Foundation/DateIntervalFormatter.swift
+                    Foundation/DateInterval.swift
+                    Foundation/Date.swift
+                    Foundation/Decimal.swift
+                    Foundation/Dictionary.swift
+                    Foundation/EnergyFormatter.swift
+                    Foundation/ExtraStringAPIs.swift
+                    Foundation/FileHandle.swift
+                    Foundation/FileManager.swift
+                    Foundation/Formatter.swift
+                    Foundation/FoundationErrors.swift
+                    Foundation/Host.swift
+                    Foundation/HTTPCookieStorage.swift
+                    Foundation/HTTPCookie.swift
+                    Foundation/IndexPath.swift
+                    Foundation/IndexSet.swift
+                    Foundation/ISO8601DateFormatter.swift
+                    Foundation/JSONEncoder.swift
+                    Foundation/JSONSerialization.swift
+                    Foundation/LengthFormatter.swift
+                    Foundation/Locale.swift
+                    Foundation/MassFormatter.swift
+                    Foundation/MeasurementFormatter.swift
+                    Foundation/Measurement.swift
+                    Foundation/NotificationQueue.swift
+                    Foundation/Notification.swift
+                    Foundation/NSArray.swift
+                    Foundation/NSAttributedString.swift
+                    Foundation/NSCache.swift
+                    Foundation/NSCalendar.swift
+                    Foundation/NSCFArray.swift
+                    Foundation/NSCFBoolean.swift
+                    Foundation/NSCFCharacterSet.swift
+                    Foundation/NSCFDictionary.swift
+                    Foundation/NSCFSet.swift
+                    Foundation/NSCFString.swift
+                    Foundation/NSCharacterSet.swift
+                    Foundation/NSCoder.swift
+                    Foundation/NSComparisonPredicate.swift
+                    Foundation/NSCompoundPredicate.swift
+                    Foundation/NSConcreteValue.swift
+                    Foundation/NSData.swift
+                    Foundation/NSDate.swift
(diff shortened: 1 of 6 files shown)


compnerd:
CC: @gottesmm

CoreFoundation/CMakeLists.txt
```diff
@@ -0,0 +1,348 @@
...
+elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
+  target_compile_definitions(CoreFoundation
+                             PRIVATE
+                               -DDEPLOYMENT_TARGET_MACOSX)
+endif()
+target_compile_definitions(CoreFoundation
+                           PRIVATE
+                             -DU_SHOW_DRAFT_API
+                             -DCF_BUILDING_CF
+                             -DDEPLOYMENT_RUNTIME_C
```
phausler: only one deployment runtime should be specified; if you are building for swift you need the RUNTIME_SWIFT version and if you are building for stand-alone you need the RUNTIME_C version.

CoreFoundation/cmake/modules/CoreFoundationAddFramework.cmake
```diff
@@ -0,0 +1,63 @@
...
+
+  if(AF_MODULE_MAP)
+    file(COPY
+           ${AF_MODULE_MAP}
+         DESTINATION
+           ${CMAKE_BINARY_DIR}/${NAME}.framework/Headers/${NAME}
+         NO_SOURCE_PERMISSIONS)
+  endif()
+  if(AF_PUBLIC_HEADERS)
+    file(COPY
```
phausler: do the header copy phases get run before the compile sources phase? that is a strict requirement for clean builds and non darwin builds

phausler: cc @parkera do you have any thoughts on this Tony?

phausler: also cc @millenomi what are your thoughts Lily? do you think this is more maintainable?

compnerd: Yeah, the headers get copied before the build into a framework layout of:

```
    CoreFoundation.framework/Headers/...
    CoreFoundation.framework/PrivateHeaders/...
```

This can be changed if needed.

CoreFoundation/CMakeLists.txt
phausler: it could also be a bug btw ;)

CoreFoundation/cmake/modules/CoreFoundationAddFramework.cmake
phausler: the important point here is that before any C file is compiled it needs to copy the headers into place so they can be properly included. and that copy phase could be just a temporary copy phase into a placeholder directory if needed. e.g. `$build_dir/usr/include/CoreFoundation` for the public headers and `$build_dir/usr/local/include/CoreFoundation` for private headers to fake out the `#include <CoreFoundation/CoreFoundation.h>` style includes.

compnerd: Oh, interesting, you would prefer the traditional Unix style layout?  Currently, I actually create a Framework style layout to permit the `<CoreFoundation/CoreFoundation.h>` style inclusion.

phausler: -F is a Darwin only thing iirc

compnerd: Sure, but, that should be easy enough to extend to other targets.  Its just a header search order thing.  It would mean that this can be built similarly on all the targets.  I can probably do that change in clang pretty easily.

phausler: It could be but that might be a tall order to get accepted I would guess. It also impacts linking too so you would need to fix gold or switch over to lld. In my opinion the path of least resistance is to just make a library and not a framework.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,75 @@
+
+include(CMakeParseArguments)
+
+function(add_swift_library library)
```
phausler: It would be super keen if this was maintained by the swift team and/or a standard cmake thing

CMakeLists.txt
```diff
@@ -0,0 +1,214 @@
...
+                      -DCF_DEPLOYMENT_SWIFT=YES
+                      -DCF_PATH_TO_LIBDISPATCH_SOURCE=${FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE}
+                      -DCF_PATH_TO_LIBDISPATCH_BUILD=${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD})
+ExternalProject_Get_Property(CoreFoundation install_dir)
+
+set(swift_optimization_flags)
+if(CMAKE_BUILD_TYPE MATCHES Release)
+  set(swift_optimization_flags -O)
+endif()
+add_swift_library(Foundation
```
phausler: Does this build incrementally or as a giant glob?

CMakeLists.txt
```diff
@@ -0,0 +1,214 @@
...
+          C)
+
+option(FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE "Path to libdispatch source" "")
+option(FOUNDATION_PATH_TO_LIBDISPATCH_BUILD "Path to libdispatch build" "")
+
+include(SwiftSupport)
+include(GNUInstallDirs)
+include(ExternalProject)
+
+ExternalProject_Add(CoreFoundation
```
phausler: The libCoreFoundation.a needs to be a whole archive linkage iirc

phausler:
It would be good to also update the readme to include instructions on how to build this via cmake

CMakeLists.txt
compnerd: That is one of the areas of improvement :-).  It is a giant glob, but it should be possible to do it incrementally (I need to figure out how to do that).

CMakeLists.txt
compnerd: Interesting, I didnt see that flag used.  But, that should be possible to do.  We explicitly build it static here to allow us to incorporate it completely.

cmake/modules/SwiftSupport.cmake
compnerd: I would absolutely love that too.  Newer versions of CMake have introduced some support for swift, and I could see this going away in the future.

CoreFoundation/CMakeLists.txt
```diff
@@ -0,0 +1,400 @@
...
+                StringEncodings.subproj/CFUniChar.h
+                StringEncodings.subproj/CFUniCharPriv.h
+                StringEncodings.subproj/CFUnicodeDecomposition.h
+                StringEncodings.subproj/CFUnicodePrecomposition.h
+                # URL
+                URL.subproj/CFURL.inc.h
+                URL.subproj/CFURLPriv.h
+                URL.subproj/CFURLSessionInterface.h
+              PUBLIC_HEADERS
+                # FIXME: PrivateHeaders referenced by public headers
```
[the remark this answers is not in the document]
phausler: For each of these can you give me the places they are referenced? We may just have a header that should be private that isn’t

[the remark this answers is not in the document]
compnerd:
```
CoreFoundation.h
  +- Base.subproj/ForSwiftFoundationOnly.h
      +- CoreFoundation/ForFoundationOnly.h
          +- Plugin.subproj/CFBundlePriv.h
          `- Stream.subproj/CFStreamPriv.h
              +- CoreFoundation/CFRuntime.h
      +- Base.subproj/CFLogUtilities.h
      +- Base.subproj/CFPriv.h
          `- Base.subproj/CFKnownLocations.h
      +- Locale.subproj/CFLocaleInternal.h
      +- Parsing.subproj/CFXMLInternal.h
      +- String.subproj/CFRegularExpression.h
      `- URL.subproj/CFURLSessionInternal.h
  +- String.subproj/CFRunArray.h
  `- URL.subproj/CFURLPriv.h
```

Perhaps `ForSwiftFoundation.h` should be brought into the module some other way?

compnerd:
Okay, docs are updated, and we now do incremental builds instead of a single glob.

CoreFoundation/CMakeLists.txt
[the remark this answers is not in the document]
compnerd: Hmm, actually, it just occurred to me, we should be able to create a private modulemap for the framework and split out the `ForSwiftFoundationOnly.h` from `CoreFoundation.h`.

compnerd:
Okay it seems to be pretty far along now, with XCTest being required for the testsuite

compnerd:
@parkera @millenomi - how do you break the dependency cycle between XCTest and Foundation?

millenomi:
@compnerd We usually don't. `build-script` builds in this order:

 - Foundation
 - XCTest
 - Foundation tests

and I think that dependency is handled in `build-script-impl`, not here.

compnerd:
Okay, this is now tested against XCTest in the sense that XCTest builds and runs against this (and passes tests).  I think that this is in a good state now.  The remaining pieces is the test suite/infrastructure.

CMakeLists.txt
```diff
@@ -0,0 +1,340 @@
...
+                    SOURCE_DIR
+                      ${CMAKE_CURRENT_SOURCE_DIR}/CoreFoundation
+                    CMAKE_COMMAND
+                      ${CMAKE_COMMAND}
+                    CMAKE_ARGS
+                      -DBUILD_SHARED_LIBS=NO
+                      -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                      -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+                      -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+                      -DCMAKE_INSTALL_LIBDIR=usr/lib
```
capent: The "lib" part really should be a variable, like "SWIFT_LIBDIR", or something like that, with a default "lib" value. That way, Linux package maintainers could define it in build time, and it will handle the whole lib|lib64 debacle themselves.

Something like that before the options on this file:
set(SWIFT_LIBDIR "lib" CACHE STRING "Library folder name, defined by swift main buildscript")

CMake [do have a way to handle this by itself](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), with the GNUInstallDirs module, and that maybe a ever more elegant solution where you can use the standard CMAKE_INSTALL_LIBDIR generated by it and remove this manual definition from you file, but i don't know how it behaves on OS X. In theory it should revert to plain "lib".

compnerd: @capent - the install is already using GNUInstallDirs.  This is for the internal build of CoreFoundation for Foundation.  This allows us to simplify some of the logic in the build by assuming it is always lib.

millenomi:
We already have some lib/lib64 infrastructure in the main build scripts; can we follow the example set there?

compnerd:
@millenomi - there shouldn't be anything special needed for the lib vs lib64 here, its using the GNUInstallDirs, so the standard cmake handling should be sufficient.  Is there something specific that you believe is unsupported with the current implementation?  This is just missing the xctest option to enable the tests AFAIK.

compnerd:
Okay, at this point, its pretty close.  The tests require copying the resource bundles and two tests are currently failing (NSLocale and UserDefaults).  This does require the annotations to be synchronized as well.

compnerd:
Okay - getting closer; need to figure out the right location for xdgTestHelper, but, seems that the rest of the test suite is now passing (the callbacks in CoreFoundation weren't being setup properly).  I think that once #1612 is merged, this is going to be really close on its heals.

spevans:
@compnerd the `xdgTestHelper` binary needs to be in the same directory as `TestFoundation` so that is can be found using `testBundle().url(forAuxiliaryExecutable: "xdgTestHelper")`

compnerd:
Okay, so, this is now complete AFAICT.  I am able to build and run tests and have the full test suite pass with the CMake based build system :-)

alblue:
@swift-ci please test

compnerd:
@phausler - what do you think is the best approach here?  We could merge this as it will have no impact, but it is not usable without the CF changes that are still pending (#1612).

compnerd:
@swift-ci please test

alblue:
@swift-ci please test

alblue:
The test seems to have failed with something unrelated to this change. Maybe it's worth trying to rebase this on top of master and we can try again?

> orelibs-foundation/../Ninja-ReleaseAssert/foundation-linux-x86_64/Foundation/usr/lib/swift/CoreFoundation/module.modulemap:2:21: error: umbrella header 'CoreFoundation.h' not found
> 13:12:12     umbrella header "CoreFoundation.h"
> 13:12:12                     ^
> 13:12:12 Foundation/NSObject.swift:15:8: error: could not build C module 'CoreFoundation'
> 13:12:12 import CoreFoundation
> 13:12:12        ^
>

compnerd:
@alblue, sure I can rebase it, but that won't go away :-(.  The thing is that the module map is altered to use a Framework layout on all targets, which is perfectly fine, except for the fact that the old build doesn't generate a framework layout ... *sigh*  I guess we could duplicate the module map temporarily.  Thoughts on that?  It seems sad to have to maintain two versions of the module map.

compnerd:
Okay, rebased with a small tweak to disable the hidden visibility of CoreFoundation.  This is less than ideal, but perhaps we can get the CMake build enabled while we fix the symbol visibility.

alblue:
@compnerd was about to merge this, then I noticed that the .modulemap for Foundation has been changed. Can you remove that file from this patch, so that we just have the CMake files? Apologies for the delay in this.

compnerd:
@alblue sure, I can do that, however, that won't really work per se ... the CMake build is configured to use the framework style build for CoreFoundation.  I can duplicate the module map and that should be enough to use the CMake build.  Would that be acceptable?

alblue:
For now, let's get the cmake scripts in without adding or changing any modulemaps. We can then tackle the modulemap changes with visibility changes in e.g. #1612 at a later stage. This way we can start iterating on the cmake scripts instead of keeping this open for a long time.

compnerd:

PR #1545, opened 2018-05-02
CoreFoundation: reference CFNumber by the canonical name

compnerd:
CFNumber is a public header in the CoreFoundation framework.  Reference
it as `<CoreFoundation/CFNumber.h>` rather than `"CFNumber.h"`.


compnerd:
CC: @parkera

PR #1548, opened 2018-05-07
FileManager: Implement missing Path and URL Enumerator functionality.

spevans:
- NSURLDirectoryEnumerator: Implement .skipsPackageDescendants and
  .skipsHiddenFiles options when recursing through sub-directories.

- NSPathDirectoryEnumerator: Implement directoryAttributes,
  fileAttributes, level and skipDescendants.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -943,9 +943,6 @@ open class FileManager : NSObject {
      */
     // Note: Because the error handler is an optional block, the compiler treats it as @escaping by default. If that behavior changes, the @escaping will need to be added back.
     open func enumerator(at url: URL, includingPropertiesForKeys keys: [URLResourceKey]?, options mask: DirectoryEnumerationOptions = [], errorHandler handler: (/* @escaping */ (URL, Error) -> Bool)? = nil) -> DirectoryEnumerator? {
-        if mask.contains(.skipsPackageDescendants) || mask.contains(.skipsHiddenFiles) {
-            NSUnimplemented("Enumeration options not yet implemented")
-        }
         return NSURLDirectoryEnumerator(url: url, options: mask, errorHandler: handler)
     }
     
@@ -1307,19 +1304,25 @@ extension FileManager {
     internal class NSPathDirectoryEnumerator: DirectoryEnumerator {
         let baseURL: URL
         let innerEnumerator : DirectoryEnumerator
+        private var _currentItemPath: String?
+
         override var fileAttributes: [FileAttributeKey : Any]? {
-            NSUnimplemented()
+            guard let currentItemPath = _currentItemPath else {
+                return nil
+            }
+            return try? FileManager.default.attributesOfItem(atPath: baseURL.appendingPathComponent(currentItemPath).path)
         }
+
         override var directoryAttributes: [FileAttributeKey : Any]? {
-            NSUnimplemented()
+            return try? FileManager.default.attributesOfItem(atPath: baseURL.path)
         }
         
         override var level: Int {
-            NSUnimplemented()
+            return innerEnumerator.level
         }
         
         override func skipDescendants() {
-            NSUnimplemented()
+            innerEnumerator.skipDescendants()
         }
         
         init?(path: String) {
@@ -1337,9 +1340,9 @@ extension FileManager {
                 return nil
             }
             let path = url.path.replacingOccurrences(of: baseURL.path+"/", with: "")
+            _currentItemPath = path
             return path
         }
-
     }
 
     internal class NSURLDirectoryEnumerator : DirectoryEnumerator {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

Foundation/FileManager.swift
```diff
@@ -1385,25 +1388,27 @@ extension FileManager {
...
 
                 _current = fts_read(stream)
                 while let current = _current {
+                    let filename = NSString(bytes: current.pointee.fts_path, length: Int(strlen(current.pointee.fts_path)), encoding: String.Encoding.utf8.rawValue)!._swiftObject
+                    let skipHidden = (_options.contains(.skipsHiddenFiles) && filename[filename._startOfLastPathComponent] == ".")
+
                     switch Int32(current.pointee.fts_info) {
                         case FTS_D:
-                            if _options.contains(.skipsSubdirectoryDescendants) {
+                            if skipHidden || _options.contains(.skipsSubdirectoryDescendants) || (_options.contains(.skipsPackageDescendants) && filename.hasSuffix(".app")) {
```
kperryua: The definitions of "isHidden" and "isPackage" should be abstracted away into some private functions or something. Eventually this abstraction should be URLResourceValues-based, but internal functions should suffice for now.

Perhaps even better, you could have a single function taking the path or filename and the enumeration options that decides whether the file should be hidden. This avoids the variable naming nit above.


That all said, since packages aren't an actual concept on Linux, so I feel that the option should be a no-op.

The . prefix definition of "hidden" seems fine, though incomplete on Darwin.

Foundation/FileManager.swift
```diff
@@ -1385,25 +1388,27 @@ extension FileManager {
                     
                     // Skip the root.
                     _current = fts_read(stream)
-                    
                 }
 
                 _current = fts_read(stream)
                 while let current = _current {
+                    let filename = NSString(bytes: current.pointee.fts_path, length: Int(strlen(current.pointee.fts_path)), encoding: String.Encoding.utf8.rawValue)!._swiftObject
+                    let skipHidden = (_options.contains(.skipsHiddenFiles) && filename[filename._startOfLastPathComponent] == ".")
```
kperryua: Nit on the variable name: I expected it to just have the value of `_options.contains(.skipsHiddenFiles)`. I'd suggest `skipBecauseHidden` or something like that. But see suggested refactoring down below.

Foundation/FileManager.swift
```diff
@@ -1381,6 +1381,30 @@ extension FileManager {
...
+            func match(filename: String, to options: DirectoryEnumerationOptions, isDir: Bool) -> (Bool, Bool) {
+                var showFile = true
+                var skipDescendants = false
+
+                if isDir {
+                    if options.contains(.skipsSubdirectoryDescendants) {
+                        skipDescendants = true
+                    }
+#if canImport(Darwin)
+                    if options.contains(.skipsPackageDescendants) && filename.hasSuffix(".app") {
```
kperryua: Let's not do this .app stuff at all, even on Darwin. There are a plethora of extensions that can represent a package, and even more non-filename-based conditions. In this instance, I'd say it's better to have consistent behavior between platforms than have an inadequate Darwin behavior.

I know this means that the API's implementation does nothing, but I don't think there's a better alternative at the moment.

Foundation/FileManager.swift
```diff
@@ -1381,6 +1381,30 @@ extension FileManager {
...
+                        skipDescendants = true
+                    }
+#endif
+                }
+                if options.contains(.skipsHiddenFiles) && (filename[filename._startOfLastPathComponent] == ".") {
```
kperryua: Seems like this could be checked first, letting us skip the `isDir` block if true, since it overrides the previous settings. But not a big deal, since the `skipDescendants` check is not expensive.

Foundation/FileManager.swift
```diff
@@ -1393,16 +1417,20 @@ extension FileManager {
                 _current = fts_read(stream)
                 while let current = _current {
                     let filename = NSString(bytes: current.pointee.fts_path, length: Int(strlen(current.pointee.fts_path)), encoding: String.Encoding.utf8.rawValue)!._swiftObject
-                    let skipHidden = (_options.contains(.skipsHiddenFiles) && filename[filename._startOfLastPathComponent] == ".")
 
                     switch Int32(current.pointee.fts_info) {
                         case FTS_D:
-                            if skipHidden || _options.contains(.skipsSubdirectoryDescendants) || (_options.contains(.skipsPackageDescendants) && filename.hasSuffix(".app")) {
+                            let (showFile, skipDescendants) = match(filename: filename, to: _options, isDir: true)
```
kperryua: The tuple return is interesting… I'll allow it. :-)

The main concern is that if the `skipDescendants` check ended up touching the file system or doing some other expensive work, it would be wasted work if you just assign it to `_`.

spevans:

spevans:

PR #1551, opened 2018-05-11
SR-7570: Initializing XMLDocument crashes on Linux with nodePreserveAll

spevans:
- Converting the options to an Int32 caused an overflow, so make
  _CFXMLDocPtrFromDataWithOptions() take the options as an
  unsigned int instead.


spevans:

PR #1552, opened 2018-05-16
Use withUnsafeMutableBufferPointer in NSData test.

rudkx:
This ensures that:
- We produce a well scoped stable pointer for the duration of the use.

- We aren't attempting to do an implicit conversion from an array to
  pointer in an argument position which is an autoclosure result,
  which is now disallowed in
  https://github.com/apple/swift/pull/16623.


rudkx:
@swift-ci Please test

rudkx:
@swift-ci Please test

rudkx:
@swift-ci Please test

PR #1554, opened 2018-05-17
Fix URLRequest copying

zayass:
A lot of fields not copied
It works very bad when `Content-Length` set but `httpBody` loosed after COW


spevans:
@swift-ci please test

zayass:
@parkera @ianpartridge @pushkarnk
Is it ok to merge?

millenomi:
I'm hoping @pushkarnk can take a look here; I'll merge unless they have an objection.

PR #1557, opened 2018-05-19
https://bugs.swift.org/browse/SR-6812

chnmrc:
Resolve a bug that affects ARMv7 system
https://bugs.swift.org/browse/SR-6812

### DIFF
--- a/CoreFoundation/Base.subproj/CFRuntime.h
+++ b/CoreFoundation/Base.subproj/CFRuntime.h
@@ -195,10 +195,18 @@ typedef struct __CFRuntimeBase {
     uintptr_t _cfisa;
     uintptr_t _swift_rc;
     // This is for CF's use, and must match _NSCFType layout
+#if __LP64__
     _Atomic(uint64_t) _cfinfoa;
+#else
+    _Atomic(uint32_t) _cfinfoa;
+#endif
 } CFRuntimeBase;
 
+#if __LP64__
 #define INIT_CFRUNTIME_BASE(...) {0, _CF_CONSTANT_OBJECT_STRONG_RC, 0x0000000000000080ULL}
+#else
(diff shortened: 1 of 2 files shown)


hpux735:
Thanks @chnmrc !  This seems pretty straight-forward to me.

millenomi:
I assume this just moves the CF types to the memory layout Swift objects have on 32 bit platforms? Is it the same on other 32 bit platforms or does it need a specific armv7 guard?

hpux735:
That's a good question.  We don't have testing on other non-supported 32-bit platforms.  As there aren't any other 32-bit platforms in community CI, I can't be sure who represents those platforms, if they're being used at all.

CoreFoundation/Base.subproj/CFRuntime.h
```diff
@@ -195,10 +195,18 @@ typedef struct __CFRuntimeBase {
     uintptr_t _cfisa;
     uintptr_t _swift_rc;
     // This is for CF's use, and must match _NSCFType layout
+#if __LP64__
```
parkera: Presumably `LLP64`, too.

chnmrc: Yes, for Windows also, maybe we must add an || condition.

chnmrc: I agree

parkera: Looks like the .c file was updated but not the .h file?

chnmrc: Is that a question to me?

parkera:
@swift-ci please test

millenomi:

spevans:

PR #1558, opened 2018-05-20
Process: Deprecate old methods and properties, implement run()

spevans:
- Mark launchPath and currentDirectoryPath as deprecated,
  replace with executableURL and currentDirectoryURL.

- Implement run() and Process.run(_:arguments:terminationHandler:)
  and mark launch() and Process.launchedProcess(path:arguments:) as
  deprecated.

- Update DarwinCompatibility Tests to 10.13 to use run().


spevans:
@swift-ci please test

spevans:
@swift-ci please test

Foundation/Process.swift
```diff
@@ -162,48 +162,36 @@ open class Process: NSObject {
...
-                return nil
-            }
-            
-            return URL(fileURLWithPath: launchPath)
-        }
-        set {
-            self.launchPath = newValue?.path
-        }
+
+    @available(*, deprecated: 4, renamed: "executableURL")
```
millenomi: While this is allowed, I don't think the semantics for this are well-defined. (The '4' in the `deprecated:` portion is meant to be the version of the platform specified just after `@available(`, but in this case there's none.)

The semantically appropriate thing to do here, I think, is just to mark it:

```swift
@available(*, deprecated, renamed:…)
```

millenomi: Unless it is undocumented that the Swift version is used even if the tag is not `@available(swift, …)`

spevans: I wasn't sure if a version number was required, its fixed now.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:

PR #1559, opened 2018-05-22
Build Foundation with DEPLOYMENT_RUNTIME_SWIFT disabled

mario-vera:
Expand Linux platforms supported under DEPLOYMENT_RUNTIME_SWIFT=0 umbrella.

### DIFF
--- a/CoreFoundation/Parsing.subproj/CFXMLInterface.c
+++ b/CoreFoundation/Parsing.subproj/CFXMLInterface.c
@@ -176,8 +176,6 @@ _CFXMLInterfaceSAXHandler _CFXMLInterfaceCreateSAXHandler() {
     saxHandler->externalSubset = (externalSubsetSAXFunc)__CFSwiftBridge.NSXMLParser.externalSubset;
     
     saxHandler->initialized = XML_SAX2_MAGIC; // make sure start/endElementNS are used
-#else
-    memset(&saxHandler, 0, sizeof(saxHandler));
 #endif //if DEPLOYMENT_RUNTIME_SWIFT
     return saxHandler;
 }
--- a/CoreFoundation/Parsing.subproj/CFXMLInterface.c
+++ b/CoreFoundation/Parsing.subproj/CFXMLInterface.c
@@ -12,6 +12,7 @@
  */
 
 #include <CoreFoundation/CFRuntime.h>
+#include <CoreFoundation/CFInternal.h>
 #include <libxml/globals.h>
 #include <libxml/xmlerror.h>
 #include <libxml/parser.h>
@@ -22,7 +23,7 @@
 #include <libxml/xpath.h>
 #include <libxml/xpathInternals.h>
 #include <libxml/dict.h>
-#include "CFInternal.h"
+#include "CFXMLInterface.h"
 
 /*
  libxml2 does not have nullability annotations and does not import well into swift when given potentially differing versions of the library that might be installed on the host operating system. This is a simple C wrapper to simplify some of that interface layer to libxml2.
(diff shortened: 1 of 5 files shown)


CoreFoundation/Parsing.subproj/CFXMLInterface.c
```diff
@@ -22,7 +23,7 @@
 #include <libxml/xpath.h>
 #include <libxml/xpathInternals.h>
 #include <libxml/dict.h>
-#include "CFInternal.h"
```
parkera: So we don't use CFInternal.h here at all?

CoreFoundation/Parsing.subproj/CFXMLInterface.c
```diff
@@ -168,6 +176,9 @@ _CFXMLInterfaceSAXHandler _CFXMLInterfaceCreateSAXHandler() {
     saxHandler->externalSubset = (externalSubsetSAXFunc)__CFSwiftBridge.NSXMLParser.externalSubset;
     
     saxHandler->initialized = XML_SAX2_MAGIC; // make sure start/endElementNS are used
+#else
+    memset(&saxHandler, 0, sizeof(saxHandler));
```
parkera: Doesn't the calloc above already promise us zeroed memory?

lib/script.py
```diff
@@ -226,7 +226,15 @@ def generate_products(self):
...
-            script += product.generate()
+            items = product.generate()
+            for item in items:
+                if isinstance(item, list):
+                    for subitem in item:
+                        #TODO: What to do with this elements?
+                        #script += subitem
+                        continue
+                else:
+                    script += item
```
parkera: Can you explain a bit more about what this is doing?

CoreFoundation/Parsing.subproj/CFXMLInterface.c
mario-vera: It wasn't removed, it moved up & uses qualified path. Check line 15: <CoreFoundation/CFInternal.h>

CoreFoundation/Parsing.subproj/CFXMLInterface.c
mario-vera: Agreed, this line is not needed, I'll remove it.

lib/script.py
mario-vera: Please add Lily to review this one. script is a string. items is a list containing strings and lists. On python 2.7 we can not append a list into a string, so we need to walk it, appending the strings on the array. 
The comment is to highlight the fact that I found no usage for the  nested lists... Not sure what the intended usage was.

CoreFoundation/Parsing.subproj/CFXMLInterface.c
parkera: Got it, thanks. Misread the diff.

lib/script.py
parkera: cc @millenomi

millenomi: IIRC: This path is triggered by the `+= [ … ]` command that adds the block runtime, which is invoked if libdispatch is not available. Can you just build libdispatch as part of your CF build instead?

millenomi: You can also fix that command to add single items instead of a list.

millenomi: Sorry, for context: this is the portion of build.py that is triggering that behavior:

```python
# This code is already in libdispatch so is only needed if libdispatch is
# NOT being used
if "LIBDISPATCH_SOURCE_DIR" not in Configuration.current.variables:
    sources += (['closure/data.c', 'closure/runtime.c'])
```

It'd be nice to fix this bit, but in general you want to set your `LIBDISPATCH_SOURCE_DIR` environment variable correctly.

mario-vera: Applying LIBDISPATCH_SOURCE_DIR did not work for me. I just auggested a more pythonic way of applying the product commands... hopefully that is better.

mario-vera: BTW, I'm compiling from within CoreFoundation, so the build.py I'm using is in CoreFoundation/build.py, that scripts do not have any references to LIBDISPATCH_SOURCE_DIR.

millenomi:
@swift-ci please test

millenomi:

PR #1560, opened 2018-05-23
Avoid runtime crashes by removing undefined behavior.

mario-vera:
From pthread docs:
"The effect of calling pthread_getspecific() or pthread_setspecific() with a key value not obtained from pthread_key_create() or after key has been deleted with pthread_key_delete() is undefined."

On some Linux flavors this results in getspecific returning uninitialized pointer with results in runtime crashes after dereferencing invalid memory.

Also NULL format for fprintf is invalid and results on crashes when calling CFShow.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -628,6 +628,7 @@ static void *__CFTSDGetSpecific() {
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_EMBEDDED_MINI
     return pthread_getspecific(__CFTSDIndexKey);
 #elif DEPLOYMENT_TARGET_LINUX
+    __CFTSDInitialize();
     return pthread_getspecific(__CFTSDIndexKey);
 #elif DEPLOYMENT_TARGET_WINDOWS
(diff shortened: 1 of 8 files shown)


CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -628,6 +628,7 @@ static void *__CFTSDGetSpecific() {
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_EMBEDDED_MINI
     return pthread_getspecific(__CFTSDIndexKey);
 #elif DEPLOYMENT_TARGET_LINUX
+    __CFTSDInitialize();
```
parkera: I think this actually belongs here:

https://github.com/apple/swift-corelibs-foundation/blob/ac7f8030ad381f0f3d97359f574a6c97a6cf66e1/CoreFoundation/Base.subproj/CFRuntime.c#L1080

where we can probably just remove the requirement for `DEPLOYMENT_RUNTIME_SWIFT` and make it `DEPLOYMENT_TARGET_LINUX`.

millenomi:
I think you have the commits for your other PR merged into this one. Can you separate them?

millenomi:
@swift-ci please test

millenomi:

PR #1561, opened 2018-05-23
Port JSONEncoder overlay changes to swift-corelibs-foundation

saiHemak:
This PR is based on PR #1347  by @parkera .

@parkera  I have applied some bridging fixes on top of your PR  to make the tests pass.


saiHemak:
@parkera  All the JSONEncoder tests are passing successfully.However, two tests (``` TestCodable.test_URL_JSON``` and ```TestCodable.test_Decimal_JSON```). I have commented out these two tests as of now. 
Please review

Foundation/JSONEncoder.swift
```diff
@@ -34,7 +34,7 @@ open class JSONEncoder {
         public static let prettyPrinted = OutputFormatting(rawValue: 1 << 0)
 
         /// Produce JSON with dictionary keys sorted in lexicographic order.
-        @available(macOS 10.13, iOS 11.0, watchOS 4.0, tvOS 11.0, *)
+        @available(OSX 10.13, iOS 11.0, watchOS 4.0, tvOS 11.0, *)
```
parkera: I think this reverts some earlier changes made to normalize the availability macros.

TestFoundation/TestCodable.swift
```diff
@@ -568,13 +568,13 @@ extension TestCodable {
         return [
             ("test_PersonNameComponents_JSON", test_PersonNameComponents_JSON),
             ("test_UUID_JSON", test_UUID_JSON),
-            ("test_URL_JSON", test_URL_JSON),
+           // ("test_URL_JSON", test_URL_JSON),
```
parkera: I think we'll need to figure out the root cause of these before we can do a final merge. What were the failures you were seeing? Any clue as to what's causing them to fail?

parkera:
@millenomi is working on bringing bridging to SCL-foundation in parallel, so we may be able to simplify this patch if we can get that to land first.

TestFoundation/TestCodable.swift
spevans: FYI, the two tests that now fail here are the only two tests in `TestCodable` that fail when run against Darwin's native foundation, so the tests could be wrong (although they don't look to be).

saiHemak:
@parkera Sure ..Thanks ..

ianpartridge:
Now that #1526 is merged we should rework this - hopefully it's easier now!

saiHemak:
Thanks @millenomi for the work on bridging. @ianpartridge I am working on this PR with bridging changes in place

millenomi:
@saiHemak A thing that may not be obvious and I need to document is that you still need to use _SwiftValue.store()/.fetch() in the bridging code for this class to avoid bringing in non-Swift Foundation values when this code runs on Darwin.

ianpartridge:
@swift-ci please test

Foundation/JSONEncoder.swift
ianpartridge: Yes @saiHemak this needs to match the version in the overlay: https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/JSONEncoder.swift#L35

Foundation/JSONEncoder.swift
```diff
@@ -270,7 +272,7 @@ fileprivate struct _JSONEncodingStorage {
     }
 
     fileprivate mutating func popContainer() -> NSObject {
-        precondition(!self.containers.isEmpty, "Empty container stack.")
+        precondition(self.containers.count > 0, "Empty container stack.")
```
ianpartridge: @saiHemak this is also wrong: https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/JSONEncoder.swift#L359

Foundation/JSONEncoder.swift
```diff
@@ -861,7 +904,7 @@ open class JSONDecoder {
     open func decode<T : Decodable>(_ type: T.Type, from data: Data) throws -> T {
```
ianpartridge: This function looks very different to the overlay - https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/JSONEncoder.swift#L1102 ??

saiHemak: I have applied overlay changes on top of #1347. However still this implementation depends on ``` bridgeToObjectiveC``` instead of ```ConevertedKey``` method.Hence only the changes pertaining to the ```ConvertedKey``` have been ignored.

saiHemak:
@ianpartridge  please review

ianpartridge:
Hi @itaiferber this is our latest effort to update `JSONEncoder` in accordance to the current version in the overlay. Could you review please?

millenomi:
Unfortunately we need _slightly_ different code because the bridging syntax on Darwin can produce Darwin Foundation objects — cfr. 
https://github.com/apple/swift-corelibs-foundation/pull/1579#issuecomment-395238181. But there's no reason we can't have helpers for this on both the overlay and Swift Foundation sides to help until we figure out how to switch those classes out at runtime.

millenomi:
@swift-ci please test

itaiferber:
@millenomi Got it. How much do we support running s-cl-f on Darwin? (Makes sense for testing but also feels a bit unfortunate that it prevents us from unifying these sorts of things right now.)

millenomi:
For the record, the result of that discussion so far is:

- for now, we don't want people using SCF on Darwin for general purpose applications; _but_
- not letting SCF run on Darwin is a significant barrier of entry for community members who want to contribute 

thus it's not acceptable to break it on Darwin.

The longer-term solution is to move the compile-time conditionals for Linux bridging to be runtime checks, so that we can e.g. have a `-bridging-framework SwiftFoundation` flag or similar somewhere to change which classes are used for `.none` and boxing. (Or runtime hooks that SCF can call into.) But we can create e.g. `box()`, `unbox()` helpers in the overlay that use the bridging syntax, and in SCF to delegate to `_SwiftValue.store()/.fetch()`, and have the same code in both places that way.

itaiferber:
@millenomi Fair enough, and sounds good! This is one such place that suffers in our code due to the bridging split, so if we can get to source stability internally it'll make things _significantly_ easier to apply from the overlay, and vice versa.

spevans:
@millenomi  I agree that SCF on Darwin only has value for developing SCF itself (inside Xcode) which is easier than on Linux. Maybe an alternative solution is to build an Xcode toolchain with `SWIFT_OBJC_INTEROP=0` to make it the same as Swift running on any other platform.

This should mean that any bridging code that has to work with the ObjectiveC runtime can be removed and reduces the maintenance burden for SCF on Darwin.

A secondary advantage may be that anyone targeting Swift on the (Linux) Server can use it to develop in Xcode using SCF so there are no surprises when they move to their code to Linux as the `Foundation` functionality will be the same (hopefully).

Im not sure if a toolchain can be built with `SWIFT_OBJC_INTEROP=0` but it might be worth examining. Even now the SCF tests don't actually build/run 100% on Darwin due to bridging issues.

millenomi:
Building a toolchain with `-DSWIFT_OBJC_INTEROP=0` on Darwin requires a tiny bit of nonpublic information, which makes it a problem on a number of levels. (It can be done, just not easily outside Apple.)

There is little value in bifurcating the toolchain builds, though. It is infinitely more valuable to have a single toolchain that can be set up to understand either mode (so, turning the `#if SWIFT_OBJC_INTEROP` changes I wrote into runtime checks) than there is in shipping two different binaries for such a narrow scope.

parkera:
To summarize the thread: we can pursue the objc interop discussion separately because it can't be resolved in this particular PR.

PR #1562, opened 2018-05-24
Test fixes for RedHat

spevans:
- Process: Make xdgTestHelper into a self-contained file
    
- Build as part of DarwinCompatibilityTests.
    
- Add TestBundle.xdgTestHelperURL() to return the URL of the
  executable.

- Process: When setting the current directory for a sub-process,
  set the PWD environment variable to this directory if the
  default environment is used. This matches Darwin's behaviour.
    
- Use xdgTestHelper as an external test program for TestProcess,
  taking options for getcwd() and getenv("PWD").
    
- SR-7748: Use /tmp as a test directory and run it through realpath()
  to test against the result of getcwd(). On macOS, /tmp is a symlink
  to private/tmp.
    
- SR-7747: FileManager.attributesOfFileSystem(forPath:) may return
  .systemNumber as 0 on some systems (eg tmpfs on Redhat). Remove
  this invalid test.


spevans:
@swift-ci please test

davezarzycki:
1) I think `test_current_working_directory()` is trying to verify that the caller's current working directory doesn't change when running a task in a *different* working directory. Therefore, changing the task's working directory to the root directory is at greater risk of being a "no op" test. I'd just drop the "pwd" assert and add a comment to the test about what the *goal* of the test is.
2) I'd also add a comment to the code that initializes `systemNumber` to note that tmpfs on Linux can return zero.

davezarzycki:
Seems reasonable

spevans:
@swift-ci please test

TestFoundation/TestBundle.swift
```diff
@@ -27,6 +27,21 @@ internal func testBundleName() -> String {
...
 }
 
+internal func xdgTestHelperURL() -> URL {
+#if DARWIN_COMPATIBILITY_TESTS
+    let exeName = "/xdgTestHelper"
+#elseif os(macOS)
+    let exeName = "/xdgTestHelper.app/Contents/MacOS/xdgTestHelper"
+#else
+    let exeName = "/xdgTestHelper/xdgTestHelper"
+#endif
```
parkera: This seems ok, but it's somewhat ironic that bundle can't find this helper itself. That's what it's supposed to be there fore in the first place.

spevans:
@davezarzycki Are you able to give this a test on redhat?

millenomi:
@swift-ci please test

TestFoundation/TestBundle.swift
millenomi: The issue here is that we are not building xdgTestHelper in a bundle directory structure. It may be worth it to figure that out.

spevans: I'll aim to do that in a follow-up PR as it is complicated by the fact that there are 3 separate targets being built (Linux, macOS and the Darwin Compatibility testing). However it should act as a good test of the `Bundle` class so it is definitely worth doing.

davezarzycki:
I haven't tested them, but they look fine.

spevans:

spevans:

PR #1563, opened 2018-05-25
NSDecimalNumber: Add missing init(value:) initialisers.

spevans:
- Fix init(mantissa:exponent:isNegative:) to set the mantissa
  correctly.


spevans:

PR #1565, opened 2018-05-27
SR-7620: Use charset encoding if available for NSString(contentsOf:usedEncoding:)

spevans:
NSString(contentsOf:usedEncoding:) would only test for UTF16 and UTF32 before falling back to UTF8. Using the char-set from the `Content-Type` header if available allows the specified encoding to be tried.

This matches the behaviour on Darwin for non-file URLs.


spevans:
@swift-ci please test

spevans:

parkera:
This looks lined up for Swift 5.0. We'll need to cherry-pick to the swift-4.2-branch to make sure it lands there.

PR #1566, opened 2018-05-27
Added support for recursively resolving symlinks

pvieito:
Used by URL.resolvingSymlinksInPath().

The current Swift Foundation implementation only tries to solve the first symlink while Darwin Foundation tries to solve recursively all the symlinks.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -1045,7 +1045,7 @@ open class FileManager : NSObject {
         let initialPath = initialPath ?? path
         let resolvedPath = _appendSymlinkDestination(destination, toPath: path)
         
-        // As in Darwin Foundation, after a recursion limit we return the intial path without resolution.
+        // As in Darwin Foundation, after a recursion limit we return the initial path without resolution.
         guard recursionLevel <= 100 else {
             return initialPath
         }
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -1046,7 +1046,13 @@ open class FileManager : NSObject {
         let resolvedPath = _appendSymlinkDestination(destination, toPath: path)
         
         // As in Darwin Foundation, after a recursion limit we return the initial path without resolution.
-        guard recursionLevel <= 100 else {
+        var maximum_recursion = sysconf(Int32(_SC_SYMLOOP_MAX))
+        
+        if maximum_recursion <= 0 {
+            maximum_recursion = Int(_POSIX_SYMLOOP_MAX)
+        }
+        
+        guard recursionLevel <= maximum_recursion else {
             return initialPath
         }
         
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -609,7 +609,7 @@ open class FileManager : NSObject {
             return
         } else if errno == ENOTEMPTY {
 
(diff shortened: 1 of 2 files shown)


pushkarnk:
@swift-ci please test

pvieito:
I hace set a recursion limit of 50, but I would like to know what Darwin Foundation does to maintain both implementations aligned.

millenomi:
This feels like it would need a test.

pvieito:
Added tests

pvieito:
@swift-ci please test

pvieito:
@swift-ci please test

parkera:
@swift-ci please test

TestFoundation/TestFileManager.swift
```diff
@@ -770,6 +771,80 @@ class TestFileManager : XCTestCase {
         }
         XCTAssertNil(try? fm.linkItem(atPath: srcLink, toPath: destLink), "Creating link where one already exists")
     }
+    
+    func test_resolvingSymlinksInPath() {
+        do {
+
+            // Initialization
+            var baseURL = FileManager.default.temporaryDirectory.appendingPathComponent("test_resolvingSymlinksInPath")
```
parkera: I think we have some of our other tests here stick a UUID in the path, so that if something gets stuck around from a previous run it doesn't hose the next test.

pvieito: Yes, good idea. Added!

Foundation/FileManager.swift
```diff
@@ -1033,16 +1033,39 @@ open class FileManager : NSObject {
         NSUnimplemented()
     }
     
-    internal func _tryToResolveTrailingSymlinkInPath(_ path: String) -> String? {
+    internal func _tryToResolveTrailingSymlinkInPath(_ path: String, recursionLevel: Int = 0, initialPath: String? = nil) -> String? {
         guard _pathIsSymbolicLink(path) else {
             return nil
         }
```
spevans: The call to `_pathIsSymbolicLink()` can be removed as `destinationOfSymbolicLink()` should fail anyway if the path is not a symbolic link (according to `readlink(2)`).

pvieito: The function `_pathIsSymbolicLink()` was already there so I decided not to touch it. It was introduced as is with the [initial `URL.resolvingSymlinksInPath()` implementation](https://github.com/apple/swift-corelibs-foundation/commit/df1b258).

pvieito:
@parkera / @millenomi Why [`NSURL.resolvingSymlinksInPath`](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURL.swift#L824) and [`NSString.resolvingSymlinksInPath`](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURL.swift#L824) have independent implementations?

millenomi:
That seems like an oversight; we should have only the one.

Generally, we want to keep these implementations closer to URL and related types.

spevans:
Is this functionality the same as provided by `realpath(3)`? If so, it might be simpler to just use this function.

pvieito:
Opened Swift Bug Report: [SR-7857](https://bugs.swift.org/browse/SR-7857)

PR #1567, opened 2018-05-29
SR-3126: Decimal(Double.infinity) hangs forever

spevans:
- Decimal doesnt fully support floating point so trap when initialised
  with Double.infinity.

- Also mark Decimal properties .infinity and .signalingNaN and
 formTruncatingRemainder(dividingBy:) as unavailable.

`Decimal.swift` requires a big sync-up with `swift/stdlib/public/SDK/Foundation/Decimal.swift` at some point, but this prevents an infinite-loop and provides a useful error message.

### DIFF
--- a/Foundation/Decimal.swift
+++ b/Foundation/Decimal.swift
@@ -326,6 +326,9 @@ extension Decimal : SignedNumeric {
         return answer;
     }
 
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

PR #1568, opened 2018-05-29
[SR-3077] Bundle.main.executableURL returns incorrect values on Linux

millenomi:
Fixes https://bugs.swift.org/browse/SR-3077

We were doing nothing with the executable path here unless the platform was Darwin or Windows. Use the PLATFORM_PATH_STYLE macro here instead to ensure all platforms execute this.


millenomi:
@swift-ci please test

PR #1571, opened 2018-05-31
Fix NSString.boolValue not matching Darwin Foundation

otaviolima:
There were some cases which were being evaluated to `true` but on Darwin/Objective-C they evaluate to `false`.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -633,13 +633,19 @@ extension NSString {
         let scanner = Scanner(string: _swiftObject)
         // skip initial whitespace if present
         let _ = scanner.scanCharactersFromSet(.whitespaces)
+
+        if scanner.scanCharactersFromSet(CharacterSet(charactersIn: "tTyY")) != nil {
+            return true
+        }
+
         // scan a single optional '+' or '-' character, followed by zeroes
         if scanner.scanString("+") == nil {
             let _ = scanner.scanString("-")
         }
+        
         // scan any following zeroes
         let _ = scanner.scanCharactersFromSet(CharacterSet(charactersIn: "0"))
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

otaviolima:
I pushed again because previous implementation changed the tests in a wrong way. 

Instead of `"-00000, +t"` it should be two different items in the array, like this -> `"-00000", "+t",`

Foundation/NSString.swift
```diff
@@ -633,12 +633,16 @@ extension NSString {
         let scanner = Scanner(string: _swiftObject)
         // skip initial whitespace if present
         let _ = scanner.scanCharactersFromSet(.whitespaces)
-        // scan a single optional '+' or '-' character, followed by zeroes
-        if scanner.scanString("+") == nil {
-            let _ = scanner.scanString("-")
-        }
+        // scan a single optional '+' or '-' character
+        let hasSignal = scanner.scanString("+") != nil || scanner.scanString("-") != nil
         // scan any following zeroes
```
parkera: Looking at the Darwin implementation, it seems we check only the first character for `t T y Y` and return `true`. After that is when we do the scan for `+`/`-` and the digits.

millenomi: @otaviolima Can you switch this with the check in the return statement below? (i.e., check that first, then this.)

otaviolima: You mean check in the same way @parkera explained it is in Darwin implementation?

millenomi: Yeah.

otaviolima: Yup, doing that right now. ^-^

otaviolima: Done.

millenomi:

millenomi:

PR #1576, opened 2018-06-03
Make Range-from-NSRange generic over StringProtocol

karwa:
This means you can take an NSRange and resolve it relative to a Substring.

This can come up sometimes when working with Substrings on the Swift level and working with Objective-C APIs which return NSRange (e.g. NSRegularExpression). You need to copy to a String when performing matching, but the matched ranges can be resolved relative to the original String.

### DIFF
--- a/Foundation/NSRange.swift
+++ b/Foundation/NSRange.swift
@@ -290,17 +290,14 @@ extension Range where Bound == Int {
 }
 
 extension Range where Bound == String.Index {
-    public init?(_ range: NSRange, in string: String) {
-        let u = string.utf16
-        guard range.location != NSNotFound,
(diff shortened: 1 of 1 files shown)


parkera:
I think this one counts as an API change, so we'll have to run it through review.

Can all callers of the current version use the new generic version? It's not immediately clear to me if it covers all possible cases, since this new one has the generic constraint (although hopefully that one is true for the concrete String type).

karwa:
Yes, it should be a source-compatible change.

The generic constraint is true for both `String` and `Substring`. Alternatively, we could write this in a non-generic way by adding an overload for the concrete `Substring` type. The documentation for `StringProtocol` says that custom types shouldn't conform to it anyway.

shahmishal:

PR #1569, opened 2018-05-30
HTTP Basic Auth implementation

saiHemak:
This PR is based on  #1407 . I have picked up the implementation from #1407  PR and added HTTPServer changes which are required for testing this functionality.

@pushkarnk please review

PR #1577, opened 2018-06-04
[SR-7861] Fix for HTTP redirect failure on Linux

nethraravindran:
Codepaths in `HTTPURLProtocol` missed to assign the port for the new redirect request which leads to the failure in redirection.


ianpartridge:
@swift-ci please test

ianpartridge:
I've merged this without a test because we need this fix in a toolchain asap. We do have a test for it, and it passes, but it's written on top of #1569 which isn't ready yet. Once #1569 is merged we will add the test for this issue too.

PR #1579, opened 2018-06-04
[SR-7865] swift-corelibs-foundation does not compile on macOS after 'as' bridging was merged

millenomi:
The new primitives are only available if Objective-C isn't. Guard them with `#canImport(ObjectiveC)`.

(Additionally, see below for discussion.)

Resolves [SR-7865](https://bugs.swift.org/browse/SR-7865).

### DIFF
--- a/Foundation/Bridging.swift
+++ b/Foundation/Bridging.swift
@@ -54,8 +54,13 @@ internal protocol _NSBridgeable {
 }
 
 
+#if !canImport(ObjectiveC)
+// The _NSSwiftValue protocol is in the stdlib, and only available on platforms without ObjC.
+extension _SwiftValue: _NSSwiftValue {}
+#endif
+
 /// - Note: This is an internal boxing value for containing abstract structures
-internal final class _SwiftValue : NSObject, NSCopying, _NSSwiftValue {
+internal final class _SwiftValue : NSObject, NSCopying {
     public private(set) var value: Any
     
     static func fetch(_ object: AnyObject?) -> Any? {
--- a/Foundation/Bridging.swift
+++ b/Foundation/Bridging.swift
@@ -150,7 +150,7 @@ internal final class _SwiftValue : NSObject, NSCopying {
             // as if we returned the unboxed value directly.
         }
         
-        // On Linux, case 2 is handled below, and case 3 can't happen —
+        // On Linux, case 2 is handled by the stdlib bridging machinery, and case 3 can't happen —
         // the compiler will produce SwiftFoundation._SwiftValue boxes rather than ObjC ones.
         #endif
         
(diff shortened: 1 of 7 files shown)


millenomi:
@swift-ci please test mac

millenomi:
@swift-ci please test

millenomi:
cc @spevans

spevans:
Ignore that last message, I see now that your change is different to what I tried.

pushkarnk:
@millenomi @spevans Can we have this merged please?

millenomi:
I forgot to run tests in Xcode; once I do today.

millenomi:
@swift-ci please test

millenomi:
Here's what I found:

 - On Darwin, `x as AnyObject` can return the _native_ `_SwiftValue` class, or `Foundation.NSNull()` — **where Foundation is the native framework from /System/Frameworks, not SwiftFoundation** for nil values.

The solution is to prevent usage of `as AnyObject` in the SCF codebase — it is not safe to assume this expression returns a `SwiftFoundation.NSObject` all the time (as we instead guarantee on Linux). SCF code must use `_SwiftFoundation.store()` instead to produce a `SwiftFoundation.NSObject`; this patch uses `SwiftFoundation.NSNull()` and intercepts and substitutes native `_SwiftValue` boxes with `SwiftFoundation._SwiftValue` boxes.

This patch isn't done; we need to make sure that objects passed in by an author work, and to this extent I need to make `SwiftFoundation._SwiftValue.fetch()` able to unbox native `_SwiftFoundation` boxes.

millenomi:
After some experimentation: we can just return the native box from `.fetch()`, since it will be casted immediately afterward to the desired type.

millenomi:
@swift-ci please test

millenomi:
This addresses the latter point.

millenomi:
cc @spevans

spevans:
Confirmed working on 2018-06-04 toolchain (macOS 10.13.5, Xcode 9.4) and Ubuntu 16.04.

millenomi:
@swift-ci please test

millenomi:
It’s breaking the SwiftPM build.

millenomi:
I cannot reproduce locally. Triggering a new build here in case it was transient.

millenomi:
@swift-ci please test

millenomi:
Found the issue:

We are implementing NSNull roundtrip in `_SwiftValue.fetch()`, but it is spurious. Darwin does not unpack `NSNull`s unless they are explicitly casted to an optional type:

```swift
let x: [Int?] = [nil]
print((x as [AnyObject]) as [Any]) // -> [<null>]
print(((x as [AnyObject]) as [Any]).map { type(of: $0) }) // -> [NSNull]
```

Since bridging already works on both platforms to unpack it in that case, we may want to stop explicitly unpacking it in our case. However, we need to figure out how to treat `SwiftFoundation.NSNull` in that case, since Darwin bridging will not unpack it. (A solution could be to make it `_ObjectiveCBridgeable` on Darwin only.)

millenomi:
I'm going to check if us not unpacking is enough to let things be unblocked while we work on a longer term solution (which hinges on whether we desire SCF to be used on Darwin as a general purpose library or not).

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

PR #1580, opened 2018-06-07
[gardening] Remove nil coalescing operator where right hand operand is literal nil

otaviolima:


spevans:
@swift-ci please test

PR #1585, opened 2018-06-08
make tests on darwin run the same way as linux

barbosa:
There were 16 failing tests on Darwin (more specifically on Decodable).
This PR removes the #if statements that were checking for linux only.


spevans:
@swift-ci please test

PR #1592, opened 2018-06-09
Add URLSessionDownloadTask.cancel(byProducingResumeData:)

joshavant:
This is an attempt at resolving the implementation for `URLSessionDownloadTask.cancel(byProducingResumeData:)`.

This function seems to represent an edge case in portability. After communicating with @phausler, it appears the Objective-C implementation of this relies on an Apple-maintained XPC process to manage the bookmarking of partially downloaded data. Of course, this behavior is not portable to the framework project.

One approach that was considered was serializing the `URLSessionDownloadTask` instance and returning _that_ to the caller as the opaque blob given to the completion handler. However, the Objective-C implementation of `URLSessionTask` does not implement any `NSCoding` flavor publicly, so conformance would need to be implemented in this framework with `internal` or greater access control. But, unfortunately, that functionality is not currently supported in Swift.

The approach suggested by @phausler was simply to... hardcode calling the completion block. This is what was done in this PR.

The only thing worth highlighting about the approach is that a token `Data` instance is passed to the completion block to prevent unexpected failures that may be caused from passing `nil` instead.

## Open Issues:
- [ ] `downloadTask(withResumeData:)` (and related) would remain broken by this approach. What should be done with these functions?
- [ ] Unit test(s)

### DIFF
--- a/Foundation/URLSession/URLSessionTask.swift
+++ b/Foundation/URLSession/URLSessionTask.swift
@@ -437,11 +437,8 @@ open class URLSessionDownloadTask : URLSessionTask {
          * original behavior cannot be directly ported, here.
          *
          * Instead, we just call the completionHandler directly.
-         *
-         * A token Data value is passed to the client to prevent unexpected failures
-         * that may be caused from passing nil instead.
          */
-        completionHandler(Data())
+        completionHandler(nil)
     }
 }
 
--- a/Foundation/URLSession/URLSessionTask.swift
+++ b/Foundation/URLSession/URLSessionTask.swift
@@ -428,7 +428,13 @@ open class URLSessionDownloadTask : URLSessionTask {
      * If resume data cannot be created, the completion handler will be
      * called with nil resumeData.
      */
-    open func cancel(byProducingResumeData completionHandler: @escaping (Data?) -> Void) { NSUnimplemented() }
+    open func cancel(byProducingResumeData completionHandler: @escaping (Data?) -> Void) {
+        super.cancel()
+        
+        // A token Data value is passed to the client to prevent unexpected
+        // failures that may be caused by passing nil instead
+        completionHandler(Data())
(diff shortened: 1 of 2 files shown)


pushkarnk:
@swift-ci please test

Foundation/URLSession/URLSessionTask.swift
```diff
@@ -428,7 +428,21 @@ open class URLSessionDownloadTask : URLSessionTask {
...
+        super.cancel()
+        
+        /*
+         * In Objective-C, this method relies on an Apple-maintained XPC process
+         * to manage the bookmarking of partially downloaded data. Therefore, the
+         * original behavior cannot be directly ported, here.
+         *
+         * Instead, we just call the completionHandler directly.
+         *
+         * A token Data value is passed to the client to prevent unexpected failures
```
millenomi: This would be prudent if data was _usually_ expected, but [the contract of this method goes](https://developer.apple.com/documentation/foundation/urlsessiondownloadtask/1411634-cancel):

> If the download is resumable, the completion handler is provided with a resumeData object. Your app can later pass this object to a session’s downloadTask(withResumeData:) or downloadTask(withResumeData:completionHandler:) method to create a new task that resumes the download where it left off.

Since we do not yet support resuming on Linux, no downloads are resumable, and so we can safely pass `nil` here. Let's not confuse downstream clients into thinking the download is resumable when we can't resume it.

joshavant: Great point. PR updated, accordingly.

pushkarnk:
@swift-ci please test

millenomi:
@swift-ci please test

joshavant:
🎉 

What's the process to merge this in, now that everything is passing?

millenomi:
Me hitting a button.

PR #1593, opened 2018-06-11
Mark assembly files as SafeSEH compatible

smeenai:
These assembly files contain no exception handlers, so we can just emit
the special @feat.00 symbol to indicate that they're trivially SafeSEH
compatible. See [1] for details.

[1] https://msdn.microsoft.com/en-us/library/windows/desktop/ms680547(v=vs.85).aspx#the_.sxdata_section_


compnerd:
These are data files, so, yes, this makes sense

compnerd:
@swift-ci please test

compnerd:
CC: @phausler

PR #1456, opened 2018-02-23
Implement FileManager.url{s,}(for:…) and NSSearchPathForDirectoriesInDomains.

millenomi:
Implement urls(for:…) as the primitive directory method:
 - In Darwin, match paths with the Objective-C implementation wherever possible.
 - On platforms that use FHS/XDG, implement reading user-dirs.* files to determine XDG directories;

This patch also enables the `@testable import` syntax in TestFoundation in the Debug configuration.

PR #1596, opened 2018-06-12
Revert "Implement FileManager.url{s,}(for:…) and NSSearchPathForDirectoriesInDomains."

rudkx:
Reverts apple/swift-corelibs-foundation#1456


rudkx:
@swift-ci Please test

millenomi:
Let's revert for now. I guess we need to conditionalize the use of `@testable import` to debug configurations only.

millenomi:
My aim is, whenever a revert is generated, to write a postmortem to show what occurred.

A while ago in the discussion for #1456 we asked @shahmishal to change our build process on CI to build Foundation in the debug configuration (`build-script --debug-founfortdation`), and we made conditional to that configuration the use of `-enable-testing` during the Foundation build. This enables the use of `@testable import`.

Unfortunately, this change is set up only on the Foundation PR CI testing and not on other CI setups. This made it fail when running tests as part of other CI configurations.

I think the long-term solution here is to have a `--build-foundation-testable-import` flag in `build-script`, and that all CI configurations run with that flag set to true, whether Foundation is built debug or release. That is:

| Command line invocation | Optimization settings | `@testable import` available |
|--|--|--|
| `--foundation` | Release | No |
| `--debug-foundation` | Debug | Yes |
| `--foundation --build-foundation-testable-import true` | Release | *Yes* |
| `--debug-foundation --build-foundation-testable-import false` | Debug | *No* |

For now, I'm going to pipe down to the Swift build whether `@testable import` is available or not, and keep its enabledness as it is (on for Debug, off for Release), and we can tackle adding the flag and changing CI later in the game.

rudkx:
Thanks so much for the write-up. I think that's an excellent process to follow and appreciate the explanation.

PR #1598, opened 2018-06-12
Revert "[Attempt 2] Implement FileManager.url{s,}(for:…) and NSSearchPathForDirectoriesInDomains."

rudkx:
Reverts apple/swift-corelibs-foundation#1597

The CI builders are broken and this blocks PR testing.

PR #1597, opened 2018-06-12
[Attempt 2] Implement FileManager.url{s,}(for:…) and NSSearchPathForDirectoriesInDomains.

millenomi:
This re-proposes the patch at #1456 — see there for discussion.

Unlike that patch, this one changes the test setup and build so that `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT` is defined iif the build product can be used in a `@testable import` expression, and conditionalizes white-box tests to the presence of that flag. That flag is never available if `DEPLOYMENT_RUNTIME_OBJC`, since macOS's Foundation will never be usable with `@testable import`.

PR #1600, opened 2018-06-14
[Attempt 3] Implement FileManager.url{s,}(for:…) and NSSearchPathForDirectoriesInDomains.

millenomi:
This re-reproposes the patch at #1597 that was reverted by #1598. It should fix the issues from the previous two attempts. (Whoops.)

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -128,21 +128,38 @@ open class FileManager : NSObject {
         case network
         case user
         
+        static let correspondingValues: [UInt: _SearchPathDomain] = [
+            SearchPathDomainMask.systemDomainMask.rawValue: .system,
+            SearchPathDomainMask.localDomainMask.rawValue: .local,
+            SearchPathDomainMask.networkDomainMask.rawValue: .network,
+            SearchPathDomainMask.userDomainMask.rawValue: .user,
+        ]
+        
+        static let searchOrder: [SearchPathDomainMask] = [
+            .systemDomainMask,
+            .localDomainMask,
+            .networkDomainMask,
+            .userDomainMask,
+        ]
+        
         init?(_ domainMask: SearchPathDomainMask) {
-            if domainMask == .systemDomainMask {
-                self = .system; return
+            if let value = _SearchPathDomain.correspondingValues[domainMask.rawValue] {
+                self = value
+            } else {
+                return nil
             }
-            if domainMask == .localDomainMask {
-                self = .local; return
-            }
-            if domainMask == .networkDomainMask {
-                self = .network; return
-            }
-            if domainMask == .userDomainMask {
-                self = .user; return
+        }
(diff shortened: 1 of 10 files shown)


millenomi:
@swift-ci please test

millenomi:
cc @rudkx @shahmishal I'm attempting again. I'm trying to figure out why the buildbot found that issue but both local Ubuntu 16.10 and Mac builds didn't find this.

felix91gr:
> Ubuntu 16.10

Did you try it with 16.04? Since it's LTS, it's usually more representative of how Ubuntu behaves :)

rudkx:
We can certainly try merging again and see what the effects are. As @felix91gr says, the same release might provide more confidence this will work, but I don't know any of the details of how the releases are configured (or any customizations in our CI), so I don't know how much of a concern that might be.

@millenomi If you do merge, can you keep an eye on the bots and revert ASAP if there are issues?

millenomi:
We can retry but I have been unavailable to check on the patch's progress and CI for a little bit as other work priorities have weighted.

Foundation/FileManager.swift
```diff
@@ -120,18 +122,333 @@ open class FileManager : NSObject {
...
     }
     
+    private enum _SearchPathDomain {
+        case system
+        case local
+        case network
+        case user
+        
+        init?(_ domainMask: SearchPathDomainMask) {
+            if domainMask == .systemDomainMask {
```
amosavian: 1. domainMask is an OptionSet afaik, and urls function will return an array with multiple items if multiple masks are passes I think. I’m not sure however.

Foundation/FileManager.swift
```diff
@@ -120,18 +122,333 @@ open class FileManager : NSObject {
...
+                self = .system; return
+            }
+            if domainMask == .localDomainMask {
```
amosavian: Why not use switch case instead?

Foundation/FileManager.swift
millenomi: No, they're documented to accept a single mask at a time. Weirdness of the Darwin API, but yeah.

Foundation/FileManager.swift
millenomi: IIRC I tried, Swift complained, and I switched over without investigating.

Foundation/FileManager.swift
millenomi: My bad — you're correct, and this patch is wrong. (_url…_ accepts a single mask, but _urls…_ does not.)

millenomi:
@swift-ci please test

millenomi:
Turns out it is indeed helpful if I remember to ask the bot for testing.

millenomi:
This has gone on long enough. I'm going to hit that merge button, and we'll deal with bugs in future follow-ups.

PR #1601, opened 2018-06-14
closure: port use of dlsym to Windows

compnerd:
This adjusts the initialization path to ensure that the lookup for
`_Block_destructInstance` is portable to Windows.


compnerd:
CC: @parkera

compnerd:
@swift-ci please test

millenomi:
It would be amazing at this point if we could have a CI node hooked up to ci-external for Windows.

PR #1603, opened 2018-06-15
Improve the performance of IndexSet.union

RobertPieta:
Removed double loop insert, leveraged swift struct copy. Should achieve performance improvement.

### DIFF
--- a/Foundation/IndexSet.swift
+++ b/Foundation/IndexSet.swift
@@ -476,16 +476,23 @@ public struct IndexSet : ReferenceConvertible, Equatable, BidirectionalCollectio
     
     /// Union the `IndexSet` with `other`.
     public func union(_ other: IndexSet) -> IndexSet {
-        // This algorithm is naïve but it works. We could avoid calling insert in some cases.
-        
-        var result = IndexSet()
-        for r in self.rangeView {
-            result.insert(integersIn: r)
+        var result: IndexSet
+        var dense: IndexSet
+
+        // Prepare to make a copy of the more sparse IndexSet to prefer copy over repeated inserts
+        if self.rangeView.count > other.rangeView.count {
+            result = self
+            dense = other
+        } else {
+            result = other
+            dense = self
         }
-        
-        for r in other.rangeView {
-            result.insert(integersIn: r)
+
+        // Insert each range from the less sparse IndexSet
+        dense.rangeView.forEach {
+            result.insert(integersIn: $0)
         }
+
         return result
     }
     


spevans:
@swift-ci please test

parkera:
This code is probably shared with the overlay stuff in the Swift project too.

rauhul:
@parkera whats the "overlay stuff"?

parkera:
https://github.com/apple/swift/blob/master/stdlib/public/SDK/Foundation/IndexSet.swift

RobertPieta:
@parkera if a PR is merged into this project and is shared with the Swift project, should I submit a PR to the Swift project with the same change?

parkera:
We need two PRs -- unfortunately they are basically copied source code in two git repositories at this time.

RobertPieta:
Not a problem. I'll submit a PR for this in the Swift project after setting up the Swift repo on my machine and ensuring the tests pass. Thank you!

rauhul:
@parkera is there a reason why this repo is not linked as a submodule in the Swift project?

parkera:
The sources are not exactly the same, and the build process is different for the two projects as well. Plus, I think we would need to have a larger discussion about introducing submodules into the main Swift repo, since that complicates the development workflow for everyone.

RobertPieta:
Profiling this change produces interesting behavior between this Core Libs project, the main Swift repo, and the existing Foundation on my machine. If I am interpreting Instruments correctly, SwiftFoundation's mutable copy of the IndexSet (backed by an internal NSIndexSet, implemented in Swift) is slower than Foundation's mutable copy (backed by an internal NSIndexSet, implemented in ObjC).

While this change achieves a performance improvement when linked to freshly compiled Swift or the existing Foundation on my machine, the change is actually slower in the Core Libs project when SwiftFoundation is compiled with -O and -whole-module-optimization.

I was able to reproduce this with a short benchmark, running the following 1,000,000 times and comparing between Foundation and SwiftFoundation:

> let a = IndexSet()
> var b = a
> b.insert(1)

On my machine SwiftFoundation was much slower than Foundation for this benchmark. Instruments points to _swift_release and swift_slowAlloc as main contributors to the time taken in SwiftFoundation.

Is there a way to compile SwiftFoundation to match the performance of Foundation? Or if not, would there by any particular places to investigate inside Core Libs that may be the source of additional allocs/deallocs during a mutable copy that could be optimized?

spevans:
Compiling SwiftFoundation in Release mode should use `-O`, I don't think there is another optimisation flag that can be applied. Have you tried benchmarking the underlying `NSIndexSet` on SwfitFoundation v ObjC Foundation as that should remove a layer of code.

parkera:
I wonder if the compiler has enough info to optimize away the retain/release of the underlying reference type. There was a change in calling convention this year (detailed in the What's new in Swift talk at WWWDC); that may have an impact as well.

RobertPieta:
If we benchmarked the following for 1,000,000 iterations:

> let a = NSMutableIndexSet()
> a.add(1)

The result is the same, Foundation is faster than SwiftFoundation. Inspecting the two in Instruments shows _swift_allocObject_ performs better (40ms) than NSAllocateObject (52ms). The largest differences occur between:

- _swift_release_dealloc (171ms) and NSIndexSet.dealloc + objc_release (45ms combined) 
- NSMutableIndexSet.add (196ms) and NSMutableIndexSet.addIndexesInRange (20ms)

Diving deeper into the .add difference, seems like the _swift_* calls related to memory management that take more time than the ObjC variants. Not sure yet if the SwiftFoundation add has a deeper call tree, or if there are more types/classes/variables to manage along the way.

Also possible the root of the above is at the Swift compiler level, outside of the scope of at least this repo.

I'll watch the talk next. Happy to read/watch any other docs/videos that might be relevant.

spevans:
Do you have any counts of the number of `_swift_allocObject/_swift_release_dealloc` and `NSAllocateObject/NSIndexSet.dealloc` ?

RobertPieta:
The Allocations profiler I think provides that insight. Foundation calls `NSAllocateObject` (eventually calls `malloc_zone_calloc`) and then `NSIndexSet.dealloc` (eventually calling `free`) for each iteration. A million of each in total. 

SwiftFoundation makes two memory allocations for each iteration using `swift_slowAlloc` (eventually calling `malloc_zone_malloc`) and then two deallocations using `_swift_release_dealloc` (eventually calling `free`):
- `NSMutableIndexSet`
- `_ContiguousArrayStorage<_NSRange>` from `_ArrayBufferProtocol._forceCreateUniqueMutableBufferImpl`

This would explain why SwiftFoundation takes longer, since it makes double the memory calls. Additionally, the `_ContiguousArrayStorage` allocation is part of the `NSMutableIndexSet.add` subtree explaining what the add difference is large.

Specifically:
1. `NSMutableIndexSet.add(_:)`
2. `NSMutableIndexSet.add(in:)` 
3. specialized `NSMutableIndexSet.add(in:)`
4. specialized `Array.replaceSubrange<A>(_:with:)`
5. specialized `_ArrayBufferProtocol._arrayOutOfPlaceReplace<A>(_:with:count:)`
6. specialized `_ArrayBufferProtocol._forceCreateUniqueMutableBufferImpl`
7. `_swift_allocObject_` leading to `malloc_zone_malloc`

spevans:
Foundation might be doing an optimisation for a single element or small number of elements. What happens if you add say 32elements to maybe try and force it to allocate the storage array?

RobertPieta:
Analyzing the following for 10,000 runs:

```
let a = NSMutableIndexSet()
for i in 0..<1000 {
    a.add(i)
}
```

Foundation makes 10,000 calls to malloc, SwiftFoundation makes 20,000 (2 for each run). Does `_ContiguousArrayStorage<_NSRange>` map to `internal var _ranges = [NSRange]()` in `NSIndexSet`? And if so, why is the  `_ContiguousArrayStorage<_NSRange>` `malloc` call separate from the `NSIndexSet` `malloc`?

Runtime wise, `_ranges[index] = range` in [_replaceRangeAtIndex](https://github.com/apple/swift-corelibs-foundation/blob/95905cf523c1a96262b3530b9662fce690bcf56c/Foundation/NSIndexSet.swift#L568) is called most here. Conceptually, the length and location of the NSRange just need to be updated.

Would you know the differences of replacing a struct in place within an Array in Swift vs ObjC? Or possibly when/why `_swift_retain_(swift::HeapObject*)` and `_swift_release_(swift::HeapObject*)` are called?

spevans:
I think you are triggering an optimisation whereby the Set can be held as a range since indexes in the loop can simple be expressed as 0-999.

Try a loop such as the following which should break any such simplification:

```
    let a = NSMutableIndexSet()
    var num = 0
    for _ in 0...1000 {
        num += Int.random(in: 1..<10)
        a.add(num)
    }

```

RobertPieta:
Thank you so much for all the prompts. With all the snippets behaving unexpectedly I started diff-ing my the code and finally the .pbxcodeproj with master. I was finally able to reproduce the slowdown with a single toggle.

The issue was "Exclusive Access To Memory" was set to `Full Enforcement (Run-time Checks in All Builds` vs `Full Enforcement (Run-time Checks in Debug Builds Only)`. This setting made a lot of sense, since Instruments was pointing to memory behavior as well.

Within my Xcode window (at the size I was using), both settings truncate to display `Full Enforcement (Run-time Check...`, which is why I didn't notice the difference without using diff.

This brings a followup question: in the Release configuration when `Exclusive Access To Memory` is set to `Full Enforcement (Run-time Checks in Debug Builds Only)`, what does it do?

RobertPieta:
With the configuration fixed I was able to identify why the union implementation had a different performance impact in the Swift repo versus this one. `NSIndexSet.mutableCopy` in this repo is currently implemented as:

```
let set = NSMutableIndexSet()
enumerateRanges(options: []) { (range, _) in
    set.add(in: range)
}
return set
```

The `_enumerateWithOptions` resulting from `enumerateRanges` and repeated `add` create additional work. Instead, `NSIndexSet.mutableCopy` could be implemented as:

```
let set = NSMutableIndexSet()
set._ranges = _ranges
set._count = _count
return set
```

This implementation should achieve a good performance improvement. If this looks reasonable I can make a PR.

spevans:
Thats seems reasonable, you may even be able to eliminate the `_count` since I think the `_ranges.count` is O(1) and `_count` is just mirroring this value.

I think a PR is a good idea, can you include some benchmarks comparing to both the current code and native Foundation? Remember that array copy is probably copy-on-write so you may want to do some updates on the new set to force a memory allocation in both cases.

parkera:
Great work @RobertPieta!

spevans:
@RobertPieta  Ive just reread `NSIndexSet` and now understand that its holding the set of indexes as an array of ranges for compactness so `_count` is the sum of the length of these ranges, so its cant be implemented using just the `.count` of the underlying array.

RobertPieta:
@spevans ok! Makes sense to leave as is then. Thank you!

PR #1605, opened 2018-06-15
NSIndexset isEqual performance enhancement

RobertPieta:
Remove expensive map and exit earlier. Should achieve performance improvement, significantly in cases where the indexSet has many intervals.


spevans:
@swift-ci please test

alblue:
A further optimisation would be to store the `indexSet.rangeView` into a constant at the start, and then use that in both places it is used in the function.

millenomi:

PR #1608, opened 2018-06-16
Use a counter instead of array in lengthOfInt

otaviolima:


Foundation/ByteCountFormatter.swift
```diff
@@ -354,14 +354,13 @@ open class ByteCountFormatter : Formatter {
     // A helper method to return the length of an int
     private func lengthOfInt(number: Int) -> Int {
         var num = abs(number)
-        var length: [Int] = []
+        var length = 0
```
spevans: I see this code is only used to test for length of == 3 or > 3 but maybe length should be at least 1 for a value of 0?

otaviolima: That is true, should I fix it in same PR or open a new one?

parkera: I think this one is small enough that we can put that fix here as well.

otaviolima: Done.

parkera:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #1609, opened 2018-06-16
[gardening] Remove explicit use of .some & .none when dealing with optionals

otaviolima:
The codebase _seems_ to not make much use of `.some`/`.none` explicit syntax.

So the main idea of this PR is just to make it more consistent.


otaviolima:
@swift-ci please test

spevans:
@swift-ci please test

PR #1610, opened 2018-06-17
Fix more tests to run correctly on Darwin's native Foundation

spevans:
HTTPCookie fixes for version 0 cookies

- For expiry date, prefer maximum-age over expires-date but only
  use maximum-age for version 1 cookies.

- For version 0 cookies only return the first port number,
  if available.

NSURLRequest: Capitalise header names
    
- Correctly override httpAdditionalHeaders headers with
  allHTTPHeaderFields.

NSURL fix checkResourceIsReachable()
    
- If URL is a file, throw error with code .fileReadUnsupportedScheme
  instead of .fileNoSuchFile to match Darwin.

Disable tests that wont work on native Foundation

### DIFF
--- a/Foundation/HTTPCookie.swift
+++ b/Foundation/HTTPCookie.swift
@@ -266,54 +266,55 @@ open class HTTPCookie : NSObject {
         }
         _version = version
 
-        if let portString = properties[.port] as? String, _version == 1 {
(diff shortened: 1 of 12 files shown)


spevans:
@swift-ci please test

PR #1612, opened 2018-06-19
CoreFoundation: synchronize CF_EXPORT and CF_PRIVATE decorations

compnerd:
This is needed to ensure that CoreFoundation is built with the correct
visibility to enable re-export of symbols from Foundation.  The
adjustment to the sources are needed to synchronize the headers with the
sources.

The changes to the locale constants are needed as variable declarations
cannot be aliases according to clang.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -27,6 +27,9 @@
 #include <CoreFoundation/CFDateIntervalFormatter.h>
 #include <CoreFoundation/ForFoundationOnly.h>
 #include <CoreFoundation/CFCharacterSetPriv.h>
+#include <CoreFoundation/CFURLPriv.h>
+#include <CoreFoundation/CFURLComponents.h>
+#include <CoreFoundation/CFRunArray.h>
 
 #if TARGET_OS_WIN32
 #define NOMINMAX
--- a/CoreFoundation/Base.subproj/SwiftRuntime/CoreFoundation.h
+++ b/CoreFoundation/Base.subproj/SwiftRuntime/CoreFoundation.h
@@ -85,18 +85,16 @@
 #include <CoreFoundation/CFUUID.h>
 #include <CoreFoundation/CFUtilities.h>
 #include <CoreFoundation/CFBundle.h>
+
 #include <CoreFoundation/CFMessagePort.h>
 #include <CoreFoundation/CFPlugIn.h>
 #include <CoreFoundation/CFRunLoop.h>
 #include <CoreFoundation/CFStream.h>
 #include <CoreFoundation/CFSocket.h>
 #include <CoreFoundation/CFMachPort.h>
+
 #include <CoreFoundation/CFAttributedString.h>
 #include <CoreFoundation/CFNotificationCenter.h>
-
-#include <CoreFoundation/CFURLPriv.h>
-#include <CoreFoundation/CFURLComponents.h>
-#include <CoreFoundation/CFRunArray.h>
 
 #include <CoreFoundation/ForSwiftFoundationOnly.h>
 


compnerd:
CC: @parkera @phausler @millenomi

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

phausler:
This has some severe consequences for synchronizing this back to macOS/iOS. I don’t want to merge this until we can reconsile those issues.

phausler:
Well first off you are exposing symbols that we are not willing to make API; which is a no-go for merging back unless we are willing to devote support for it. We don’t have these symbols as public export on Darwin, so what is the issue here? Perhaps a deeper explanation of the problem at hand would be good. Because it works today on Linux and Darwin and this brings in a huge level of risk/need for support.

compnerd:
The problem is that the headers and the source files are in conflict.  Its unclear which API and SPI need to be exposed to support Foundation.  A concrete example is `_GetThreadName`.  That is an SPI which is used, and needs to be exported so that it can be used.  However, I'm happy to go through and change the functions which are now exported as being internal only (so that the header alters the decoration).  Would you mind going through and pointing out which are SPIs and which are not?

phausler:
The best way to do this is to rely on the symbolic exports detailed by nm (using the system CF as reference)

compnerd:
@phausler - okay, reconciled it with that list.  There are a couple that need to be adjusted to support the tests/SPIs.  But, the vast majority are now marked as internal.

phausler:
This looks more in the right direction. I will need to take some time with this to ensure it won’t break the Darwin side.

compnerd:
@phausler - sure.  I would appreciate it if you could take a look at this sooner though as it as we are *really* close to having the CMake based build working now :-).

alblue:
@swift-ci please test

compnerd:
@swift-ci please test

parkera:
I'm not sure about the effects of this one. Need some more time to review.

CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -138,6 +138,10 @@
             #define CF_EXPORT extern __declspec(dllimport)
         #endif
     #endif
+#elif TARGET_OS_LINUX || TARGET_OS_BSD || TARGET_OS_ANDROID
+
+#define CF_EXPORT __attribute__((__visibility__("default")))
```
parkera: Why don't we just define `CF_EXPORT` to include `extern`?

CoreFoundation/Preferences.subproj/CFPreferences.c
```diff
@@ -51,7 +51,7 @@ CONST_STRING_DECL(kCFPreferencesCurrentUser, "kCFPreferencesCurrentUser")
 
 
 static CFAllocatorRef _preferencesAllocator = NULL;
-CF_PRIVATE CFAllocatorRef __CFPreferencesAllocator(void) {
+CF_EXPORT CFAllocatorRef __CFPreferencesAllocator(void) {
```
parkera: Do we use these from the Foundation binary? Otherwise they seem like they should be internal.

CoreFoundation/Preferences.subproj/CFXMLPreferencesDomain.c
```diff
@@ -46,7 +46,7 @@ static void getXMLKeysAndValues(CFAllocatorRef alloc, CFTypeRef context, void *x
 static CFDictionaryRef copyXMLDomainDictionary(CFTypeRef context, void *domain);
 static void setXMLDomainIsWorldReadable(CFTypeRef context, void *domain, Boolean isWorldReadable);
 
-CF_PRIVATE const _CFPreferencesDomainCallBacks __kCFXMLPropertyListDomainCallBacks = {createXMLDomain, freeXMLDomain, fetchXMLValue, writeXMLValue, synchronizeXMLDomain, getXMLKeysAndValues, copyXMLDomainDictionary, setXMLDomainIsWorldReadable};
+CF_EXPORT const _CFPreferencesDomainCallBacks __kCFXMLPropertyListDomainCallBacks = {createXMLDomain, freeXMLDomain, fetchXMLValue, writeXMLValue, synchronizeXMLDomain, getXMLKeysAndValues, copyXMLDomainDictionary, setXMLDomainIsWorldReadable};
```
parkera: This also seems like it should be internal.

CoreFoundation/StringEncodings.subproj/CFStringEncodingConverter.h
```diff
@@ -72,16 +72,16 @@ CF_EXPORT uint32_t CFStringEncodingBytesToUnicode(uint32_t encoding, uint32_t fl
 typedef CFIndex (*CFStringEncodingToBytesFallbackProc)(const UniChar *characters, CFIndex numChars, uint8_t *bytes, CFIndex maxByteLen, CFIndex *usedByteLen);
 typedef CFIndex (*CFStringEncodingToUnicodeFallbackProc)(const uint8_t *bytes, CFIndex numBytes, UniChar *characters, CFIndex maxCharLen, CFIndex *usedCharLen);
 
-CF_EXPORT bool CFStringEncodingIsValidEncoding(uint32_t encoding);
+CF_PRIVATE bool CFStringEncodingIsValidEncoding(uint32_t encoding);
```
parkera: `CF_PRIVATE` is not defined for people importing CF's headers, so if something is appearing in a public or private header that is not actually exported, we need to just delete it.

CoreFoundation/Base.subproj/CFBase.h
compnerd: Because the definition should not contain extern.  `CF_EXPORT` can be used on both the declaration and the definition.

CoreFoundation/Preferences.subproj/CFPreferences.c
compnerd: Precisely :-)  They are SPI for Foundation.

CoreFoundation/StringEncodings.subproj/CFStringEncodingConverter.h
compnerd: Ah, cool, will remove.  I was just worried about doing that since it was previously in the header.  I'll remove it.

CoreFoundation/Preferences.subproj/CFXMLPreferencesDomain.c
compnerd: Hmm, it seemed that I couldn't do that due to the usage of it as SPI.

CoreFoundation/Base.subproj/CFBase.h
parkera: We define it to `extern` for our other platforms.

compnerd: Hmm, actually, now that I think about it, we should be able to use the visibility attribute on Darwin too and unify the codepaths.

CoreFoundation/StringEncodings.subproj/CFStringEncodingConverter.h
compnerd: Hmm, I actually took another look at this.  This header is private and will not be exposed to the users, so I think that we can leave this as is.  Do you still want to move it to another header?  The implementation and the use are across two files, so we need it to be in some header.

parkera: On Darwin (where this source code is shared), this header is used by other projects outside CF, so if we change this to CF_PRIVATE then we will break them. We call that "private" vs an "internal" header. Unfortunately there is not really a way for you to see that from this project.

parkera: It doesn't help that `CF_PRIVATE` is a misnomer; it really should be `CF_INTERNAL`, but anyway that's something we shouldn't try to change here.

compnerd: Okay, I can move the declaration, but do need a new home for it.

parkera: `CFStringEncodingConverterPriv.h`, despite the name looking like it's private, is actually internal. That is a good place.

compnerd:
Okay, moved the declarations and switched Darwin to also use the visibility to permit building it with hidden visibility by default there as well.

compnerd:
@parkera ping

compnerd:
@parkera - hey, so, this is going to block a bunch of other work too :-( ... can you please take a look at this soon?

alblue:
@swift-ci please test

parkera:
Sorry for the delay on this. I'll take a look.

parkera:
I'm still stuck on changing the definition of `CF_EXPORT` here. It's used all over the rest of the projects in the OS itself, and anyone who uses the SDK can also use this macro. If we remove `extern` from it then we will change its meaning, which is clear for CF itself from the diff required here to add it back to various constants.

parkera:
I think we could change the definition to:

```
#define CF_EXPORT extern __attribute__((__visibility__("default")))
```

and fix up the mismatches between `CF_EXPORT` and `CF_PRIVATE`. The symbols that are used by the Swift Foundation code can be marked `CF_CROSS_PLATFORM_EXPORT` instead of `CF_EXPORT`.

compnerd:
@parkera - just to make sure that we are both looking at the same bits, the second commit (`extern`-ing all the constants) should be okay right?  I'll adjust the top commit and see if I can do what you are suggesting

parkera:
I tried a simple C++ library in VS2017 and it seems to accept `extern __declspec(dllexport)` on the declaration and definition of a function and constant value. [This documentation](https://msdn.microsoft.com/en-us/library/y4h7bcy6.aspx) also seems to suggest that `extern __declspec(dllexport)` is ok:

```
extern __declspec( dllexport ) int k;  // Okay; extern implies declaration. 
```

What am I missing? Do we have enough ready that I could try to build this for Windows and see the errors? I want to get this figured out so we can unblock you.

compnerd:
@parkera, yeah, I think I ran into issues with clang building with that.  That was the original motivation for splitting out the `extern` from that.

parkera:
I see; so is the problem with clang's interpretation of Microsoft's rules about the dllexport decoration?

millenomi:
@compnerd (is there a clang output that we can look at for the issue?)

compnerd:
@parkera, well, its just more strict about it.  @millenomi, sorry, I don't have it on hand.  I can see if I can get my laptop dusted off to get another build though.

millenomi:
I want to discuss this further.

alblue:
This has a bunch of conflicting files now; @compnerd do you want to rebase or close?

compnerd:
@alblue - I think that I should try to rebase this.  This would be a nice thing to get right.

PR #1613, opened 2018-06-20
NSCondition fixes: Ensure correct use of lock()/action/unlock() sequence

spevans:


spevans:
@swift-ci please test

PR #1614, opened 2018-06-22
closure: make `Block_descriptor_1` conform to the ABI

compnerd:
The Blocks ABI v1 used `unsigned long int` for the `size` and `reserved`
fields.  On LLP64 targets (e.g. Windows x86_64), `uintptr_t` is larger
than `unsigned long int`.  Adjust the types to make the definition
conform to the ABI specification.

### DIFF
--- a/closure/Block_private.h
+++ b/closure/Block_private.h
@@ -38,8 +38,8 @@ enum {
 
 #define BLOCK_DESCRIPTOR_1 1
 struct Block_descriptor_1 {
-    uintptr_t reserved;
-    uintptr_t size;
+    unsigned long int reserved;
+    unsigned long int size;
 };
 
 #define BLOCK_DESCRIPTOR_2 1


compnerd:
CC: @parkera @phausler

compnerd:
@swift-ci please test

compnerd:
@parkera: correct.  It uses `unsigned long int` (which is going to be 32-bits on LLP64).

compnerd:
@parkera seems like a pretty simple case?

parkera:
The root of the question is me wondering if the first two values should be 64 bits on LLP64, like they would be on LP64. I suppose that would require an update to clang when targeting Windows. However, when looking at the rest of the source here it doesn't appear to be actually used by the runtime so I guess this is ok.

compnerd:
Sure, we could look at altering the definition in LLP64 since we don't expect there to be many users right now.  I think that is beyond the scope of this change and we should discuss that with @rjmccall first.

rjmccall:
That's an interesting question.  I think the intent of the specification is probably to always get a pointer-sized integer, yes; that is, the `reserved` field should allow it to be used for a pointer, and the `size` field should not impose artificial restrictions on the size (even though a > 4GB block literal is pretty clearly impossible).

I think that is a change we can reasonably make, but yeah, it's really up to the libclosure mantainers (CC @gparker42), as well as making changes to clang and (I think) gcc.

gparker42:
Yes, the intent is `uint64_t`. The original author of this code was a classic C89 / LP64 / Unix guy so `unsigned long int` is how he would have spelled that intent.

PR #1616, opened 2018-06-23
SR-8078: Some time zone identifiers are not supported by TimeZone on Linux

spevans:
- TimeZone.knownTimeZoneIdentifiers was returning time zone names
  that could not be instantiated by TimeZone(identifier:) due to
  CFTimeZone.c:__nameStringOK() filtering out some names.

- Make CFTimeZoneCopyKnownNames(), used by .knownTimeZoneIdentifiers,
  also filter these names using __nameStringOK().

- __nameStringOK calls ICU function ucal_getCanonicalTimeZoneID
  which ultimately decides if the time zone is valid.

https://bugs.swift.org/browse/SR-8078


spevans:
@swift-ci please test

alblue:
LGTM

PR #1618, opened 2018-06-27
build: link against shared BlocksRuntime

compnerd:
This allows libdispatch to build BlocksRuntime shared.

### DIFF
--- a/build.py
+++ b/build.py
@@ -125,6 +125,7 @@ if "LIBDISPATCH_SOURCE_DIR" in Configuration.current.variables:
 		'-Xcc -fblocks'
 	])
 	foundation.LDFLAGS += '-ldispatch -L'+Configuration.current.variables["LIBDISPATCH_BUILD_DIR"]+'/src/.libs -rpath \$$ORIGIN '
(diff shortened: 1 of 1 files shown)


parkera:
Are we going to ditch the implementation in Foundation then?

alblue:
@swift-ci please test

compnerd:
@parkera - that would totally be an option - we could end up with a single implementation, which would be an improvement.

parkera:
I always thought that the Swift runtime should include the C blocks library on Linux.

compnerd:
@parkera - yeah, that makes sense to me, especially since blocks are woven into the language itself.  Unfortunately, I don't know how to best accomplish that move.  I think that once we can have a shared BlocksRuntime, it should be possible to actually move the source into the swift repository and have it be provided by the swift runtime.

parkera:
Ok

compnerd:
Lets get this bit merged.  Since it doesn't break the build even without the library being converted, it reduces the number of outstanding changes and simplifies the effort to switch to it.

PR #1620, opened 2018-06-29
O(n) replacement in NSString.replaceOccurrances

gmilos:
Bug discussed here:
https://bugs.swift.org/browse/SR-7749

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -1420,18 +1420,35 @@ extension NSMutableString {
             return _replaceOccurrencesOfRegularExpressionPattern(target, withTemplate:replacement, options:options, range: searchRange)
         }
         
+        guard let findResults = CFStringCreateArrayWithFindResults(kCFAllocatorSystemDefault, _cfObject, target._cfObject, CFRange(searchRange), options._cfValue(true)) else {
+            return 0
+        }
+        let numOccurrences = CFArrayGetCount(findResults)
 
-        if let findResults = CFStringCreateArrayWithFindResults(kCFAllocatorSystemDefault, _cfObject, target._cfObject, CFRange(searchRange), options._cfValue(true)) {
-            let numOccurrences = CFArrayGetCount(findResults)
+        guard type(of: self) == NSMutableString.self else {
+            // If we're dealing with non NSMutableString, mutations must go through `replaceCharacters` (documented behavior)
             for cnt in 0..<numOccurrences {
                 let rangePtr = CFArrayGetValueAtIndex(findResults, backwards ? cnt : numOccurrences - cnt - 1)
                 replaceCharacters(in: NSRange(rangePtr!.load(as: CFRange.self)), with: replacement)
             }
             return numOccurrences
-        } else {
-            return 0
         }
 
+        let start = _storage.utf16.startIndex
+        var newStorage = Substring()
+        var sourceStringCurrentIndex = _storage.startIndex
+        for cnt in 0..<numOccurrences {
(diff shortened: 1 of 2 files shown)


spevans:
I don't see any existing tests for this method, could you add some test cases?

Foundation/NSString.swift
```diff
@@ -1423,10 +1423,20 @@ extension NSMutableString {
 
         if let findResults = CFStringCreateArrayWithFindResults(kCFAllocatorSystemDefault, _cfObject, target._cfObject, CFRange(searchRange), options._cfValue(true)) {
             let numOccurrences = CFArrayGetCount(findResults)
+            let start = _storage.utf16.startIndex
+            var newStorage = Substring()
+            var sourceStringCurrentIndex = _storage.startIndex
             for cnt in 0..<numOccurrences {
-                let rangePtr = CFArrayGetValueAtIndex(findResults, backwards ? cnt : numOccurrences - cnt - 1)
-                replaceCharacters(in: NSRange(rangePtr!.load(as: CFRange.self)), with: replacement)
```
phausler: removing this changes the behavior for subclassing

Foundation/NSString.swift
```diff
@@ -1423,10 +1423,20 @@ extension NSMutableString {
 
         if let findResults = CFStringCreateArrayWithFindResults(kCFAllocatorSystemDefault, _cfObject, target._cfObject, CFRange(searchRange), options._cfValue(true)) {
             let numOccurrences = CFArrayGetCount(findResults)
+            let start = _storage.utf16.startIndex
```
phausler: _storage should only be accessed in concrete methods, it may not always be there

parkera: This might be tricky, without factory patterns which allow us to return a special NSMutableString subclass.

Foundation/NSString.swift
phausler: Currently the concrete class is NsMutableString itself so ‘type(of: self) === NSMutableString.self’ would be true for the concrete case but not the subclass case

parkera: Checking the class seems like a reasonable workaround to me.

spevans: Can you use something like the following but specialise the `substring(with:)` for `NSMutableString` to avoid the `String()` copy it uses:

```
var newStorage = ""
var currentIndex = 0
for cnt in 0..<numOccurrences {
    let rangePtr = CFArrayGetValueAtIndex(findResults, backwards ? numOccurrences - cnt - 1 : cnt)
    let range = NSRange(rangePtr!.load(as: CFRange.self))
    let copyRange = NSRange(location: currentIndex, length: range.location)
    newStorage += substring(with: copyRange)
    newStorage += replacement
    currentIndex = range.location + range.length
}
let copyRange = NSRange(location: currentIndex, length: length - currentIndex)
newStorage += substring(with: copyRange)
setString(newStorage)
```

gmilos:
@phausler @parkera I updated the patch (finally), testing for `NSMutableString` type, as per your suggestion.

parkera:

alblue:

kevints:
@gmilos @alblue Looks like this was never merged?

millenomi:
Thanks for the ping. I’ll put it back on its merry way when CI unblocks.

millenomi:

kevints:
@gmilos looks like this change needs to be rebased

alblue:
@gmilos can you rebase your change so that we can retest?

weissi:
ping @gmilos

weissi:
ping @gmilos

kevints:
Thanks for the patch @gmilos, I've rebased it in #2534.

PR #1621, opened 2018-07-02
Performance improvement for NSIndexSet.mutableCopy

RobertPieta:
Replaces repeated `add` in `mutableCopy` with faster assignment. Implementation now matches [`NSMutableIndexSet.copy` implementation](https://github.com/apple/swift-corelibs-foundation/blob/95905cf523c1a96262b3530b9662fce690bcf56c/Foundation/NSIndexSet.swift#L536).

Benchmarked using [`swift-profiling`](https://github.com/RobertPieta/swift-profiling), specifically [`test_indexset_mutable_copy.swift`](https://github.com/RobertPieta/swift-profiling/blob/master/tests/test_indexset_mutable_copy.swift).

Previous implementation: 9.95s
New implementation: 0.08s

Comparison to native Foundation:
Native Foundation: 0.06s

Since #1603 relies on `NSIndexSet.mutableCopy`, a similar speedup is observed for `IndexSet.union`.

### DIFF
--- a/Foundation/NSIndexSet.swift
+++ b/Foundation/NSIndexSet.swift
@@ -93,8 +93,19 @@ open class NSIndexSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
     
     open func mutableCopy(with zone: NSZone? = nil) -> Any {
         let set = NSMutableIndexSet()
-        set._ranges = _ranges
-        set._count = _count
+
+        if type(of: self as Any) == NSMutableIndexSet.self {
+            set._ranges = _ranges
+            set._count = _count
+        }
+        // On Darwin, NSMutableSet is abstract, and does not have ivars.
+        // Therefore assignment to _ranges and _count would not be correct.
+        else {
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

parkera:
I think this has the same problem as we were discussion in another PR (which I'm having trouble locating) about subclassing correctness. On Darwin, `NSMutableSet` is abstract, so it doesn't even have the ivars in the first place.

That said, we don't have a way yet in Swift to implement our factory method pattern, so this is probably a reasonable way to preserve the performance we need.

millenomi:
@RobertPieta random thought: can you check if the object is of the parent class type (`type(of: self as Any) == NSMutableIndexSet.self`) and only enumerate for subclasses? Should give us the win in the common case while preserving correctness in the copy case.

spevans:
@swift-ci test

RobertPieta:
@millenomi done! Let me know if the added branch is correct and what you were looking for.

shahmishal:

PR #1623, opened 2018-07-02
Improved performance of isSubset for NSOrderedSet and NSSet

RobertPieta:
Improve performance by exiting early if self is larger than the potential containing set, since a larger set cannot be contained by a smaller set.


spevans:
Looks good but I don't see any pre-existing tests for `NSSet.isSubset` only `NSOrderedSet.isSubset`. Could you add some for NSSet to ensure this new code path is covered?

RobertPieta:
@spevans absolutely!

spevans:
@swift-ci please test

PR #1625, opened 2018-07-11
make sure writes to URLSession.taskRegistry happen on work queue

ahti:
Addresses [SR-8017](https://bugs.swift.org/browse/SR-8017).

The docs for taskRegistry specify any use should occur on the sessions workQueue, which wasn't respected with some of the removals.

I'm not sure all callers of `URLSession.behaviour(for:)` are run on the work queue, but the changes in this PR resolved the crashes I was seeing, so I didn't touch `URLSession.behaviour(for:)`.

A test for this would be nice, but I couldn't really think of a way to test this, as the crashes occur randomly and require maybe a few thousand requests to semi-reliably reproduce.

### DIFF
--- a/Foundation/URLSession/URLSessionTask.swift
+++ b/Foundation/URLSession/URLSessionTask.swift
@@ -552,24 +552,30 @@ extension _ProtocolClient : URLProtocolClient {
             session.delegateQueue.addOperation {
                 delegate.urlSession(session, task: task, didCompleteWithError: nil)
                 task.state = .completed
-                task.workQueue.async {
+                session.workQueue.async {
                     session.taskRegistry.remove(task)
                 }
             }
         case .noDelegate:
             task.state = .completed
-            session.taskRegistry.remove(task)
+            session.workQueue.async {
+                session.taskRegistry.remove(task)
+            }
         case .dataCompletionHandler(let completion):
             session.delegateQueue.addOperation {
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

pushkarnk:

pushkarnk:
Thanks @ahti !

ahti:
I would say yes, but I'm not yet familiar enough with the codebase to say so with much authority ^_^

ianpartridge:
Please put up a PR and we'll get it merged. Otherwise, the fix won't be released until Swift 5 next year.

Foundation/URLSession/URLSessionTask.swift
```diff
@@ -552,24 +552,30 @@ extension _ProtocolClient : URLProtocolClient {
...
                 task.state = .completed
-                task.workQueue.async {
+                session.workQueue.async {
                     session.taskRegistry.remove(task)
                 }
             }
         case .noDelegate:
             task.state = .completed
-            session.taskRegistry.remove(task)
+            session.workQueue.async {
```
zayass: Sorry for necroposting but this and this https://github.com/apple/swift-corelibs-foundation/pull/1625/files#diff-14c6cb724eead661596e4468ae5ca638R625 looks like overkill
There is no thread switching with `session.delegateQueue.addOperation`
And if this method itself not synchronized race will be earlier at line 545

PR #1632, opened 2018-07-13
Remove unavailable overloads of ++/-- for CGFloat.

rudkx:
I've opened https://github.com/apple/swift/pull/17918 for the
compiler, which removes the operator declarations as well as the
similar unavailable overloads from the stdlib.

The compiler change includes a diagnostic suggesting the use of
'+= 1', just like the message that is used in these removed
overloads (and the ones in the stdlib).


rudkx:
@swift-ci Please test

PR #1633, opened 2018-07-13
Use reference type NSNumber for JSONSerialization

johnno1962:
As discussed in PR https://github.com/apple/swift-corelibs-foundation/pull/1443, now that NSNumber can be cast to numeric types outside Darwin, switch JSONSerialization code to use it. Also removes some ill-advised conversion to Int of double values that happen to round which will fail for double values > Int.max


johnno1962:
Closed in favour of https://github.com/apple/swift-corelibs-foundation/pull/1634

PR #1634, opened 2018-07-13
[JSONSerialization] Always use reference numeric types to match Darwin behavior

xwu:
This PR removes the undocumented internal serialization option `useReferenceNumericTypes` and aligns behavior with Darwin Foundation. (Note: there remains a divergence between platforms when it comes to 128-bit integers.)

The parsing logic is simplified and tests are expanded to ensure that `"[10]"` is deserialized to a value that can be cast to `Int` or `Double`.

### DIFF
--- a/Foundation/JSONEncoder.swift
+++ b/Foundation/JSONEncoder.swift
@@ -859,7 +859,7 @@ open class JSONDecoder {
     /// - throws: `DecodingError.dataCorrupted` if values requested from the payload are corrupted, or if the given data is not valid JSON.
     /// - throws: An error if any value throws an error during decoding.
     open func decode<T : Decodable>(_ type: T.Type, from data: Data) throws -> T {
-        let topLevel = try JSONSerialization.jsonObject(with: data, options: [.useReferenceNumericTypes])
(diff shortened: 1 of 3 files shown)


xwu:
@swift-ci please test

xwu:
@johnno1962 Whoops, looks like we did the same thing. Sorry!

johnno1962:
Pinch my PR will you 😀!. You’re is better, run with it. Don’t forget to cherry pick into 4.2

parkera:
Looks good to me. I'm assuming we can do this now because of the bridging improvements?

johnno1962:
Yes, this is following through on https://github.com/apple/swift-corelibs-foundation/pull/1443

xwu:
@itaiferber Certainly a good idea; I'll keep this PR limited to the changes here, and that can be a follow-on improvement. As I mention, we still have some discrepancies in that department with respect to Darwin support for 128-bit integers.

itaiferber:

itaiferber:
@xwu Now that this is merged, please also cherry pick into swift-4.2-branch

PR #1635, opened 2018-07-13
Add @discardableResult to NSArray.write* methods

zayass:
Make result discardable as it is on Darwin


millenomi:

PR #1638, opened 2018-07-20
build: use the block headers from internal copy

compnerd:
Rather than relying on the system having a copy of the blocks runtime headers,
use the headers from our local copy.


parkera:
I suppose dispatch is always present, so we can rely on this?

compnerd:
@parkera - yeah, it should always be present

compnerd:
@parkera - okay to merge then?

parkera:
Yup, thanks!

PR #1639, opened 2018-07-22
Fix configure script Python 3 compatibility

ahti:
This makes Foundation build on platforms where the `python` executable is Python 3.


parkera:
I'm not a Python expert, but this will also continue to work for Python 2.x, right?

parkera:
@swift-ci please test

ahti:
Yes. The only change in behavior could occur if the command output was not valid UTF-8, which should not happen with pkg-config output I think.

parkera:
Sorry about that. Looks good to me.

PR #1642, opened 2018-07-31
SR-8407: NSNumber as? Bool should only work for 0 or 1

spevans:
- NSNumber as? Bool should only work for NSNumber values 0 and 1 to
  match Darwin.

### DIFF
--- a/TestFoundation/TestNSNumber.swift
+++ b/TestFoundation/TestNSNumber.swift
@@ -35,6 +35,7 @@ class TestNSNumber : XCTestCase {
             ("test_objCType", test_objCType ),
             ("test_stringValue", test_stringValue),
             ("test_Equals", test_Equals),
+            ("test_boolValue", test_boolValue),
         ]
     }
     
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci please test

parkera:
I recall @phausler doing a lot of work in this area last year as far as bridging goes; do we need to update it elsewhere here now that we have more consistent bridging with linux?

TestFoundation/TestNSNumber.swift
```diff
@@ -1214,4 +1215,51 @@ class TestNSNumber : XCTestCase {
...
+
+        XCTAssertEqual(NSNumber(value: Int32.max).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int32.max - 1).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int32.min).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int32.min + 1).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int32(-1)).boolValue, true)
+
+        XCTAssertEqual(NSNumber(value: Int64.max).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int64.max - 1).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int64.min).boolValue, false) // Darwin compatibility
```
phausler: this is no longer the case, we finally fixed that in High Sierra.

TestFoundation/TestNSNumber.swift
```diff
@@ -1214,4 +1215,51 @@ class TestNSNumber : XCTestCase {
...
+        XCTAssertEqual(NSNumber(value: Int64.min + 1).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int64(-1)).boolValue, true)
+
+        XCTAssertEqual(NSNumber(value: Int.max).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int.max - 1).boolValue, true)
+        XCTAssertEqual(NSNumber(value: Int.min).boolValue, false)   // Darwin compatibility
```
phausler: ditto here

TestFoundation/TestNSNumberBridging.swift
```diff
@@ -623,6 +624,37 @@ class TestNSNumberBridging : XCTestCase {
             XCTAssertEqual(value, ns_value)
         }
     }
+
+    func testNSNumberToBool() {
```
phausler: this is correct behavior, thanks for the update!

TestFoundation/TestNSNumber.swift
spevans: Are you sure? I tested on HS 10.13.6:

```
print(NSNumber(value: Int64.min).boolValue)
false
```

spevans:
@parkera The behaviour for `as? Bool` now matches Darwin, although I think when the bridging was merged to `master` and `swift-4.2-branch` it actually regressed the previous behaviour which as also correct, so this probably warrants a cherry-pick to 4.2 as well.

PR #1646, opened 2018-08-02
NSURLRequest.hash: Implement

lorentey:
`NSURLRequest` inherits `hash` from `NSObject`, but overrides `isEqual(_:)` to compare contents. The default hash uses object identity, so two `NSURLRequest` instances that compare equal under `isEqual(_:)` can produce non-equal hash values. This also breaks hashing for `URLRequest`, which calls into `NSURLRequest.hash`.

Add an implementation for `NSURLRequest.hash` that matches `isEqual(_:)`.

Resolves https://bugs.swift.org/browse/SR-8449

### DIFF
--- a/TestFoundation/TestURLRequest.swift
+++ b/TestFoundation/TestURLRequest.swift
@@ -18,6 +18,7 @@ class TestURLRequest : XCTestCase {
             ("test_mutableCopy_1", test_mutableCopy_1),
             ("test_mutableCopy_2", test_mutableCopy_2),
             ("test_mutableCopy_3", test_mutableCopy_3),
+            ("test_hash", test_hash),
             ("test_methodNormalization", test_methodNormalization),
             ("test_description", test_description),
         ]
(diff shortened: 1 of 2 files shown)


lorentey:
@swift-ci please test

TestFoundation/TestURLRequest.swift
```diff
@@ -194,6 +195,15 @@ class TestURLRequest : XCTestCase {
         XCTAssertNil(originalRequest.allHTTPHeaderFields)
     }
 
+    func test_hash() {
```
millenomi: Can we test also altering all the entries that contribute to the hash?

lorentey: `NSObject.hash` doesn't allow for custom seeding, so the possibility of collisions makes this unfeasible for properties that have less than a handful possible values -- but we can check most of them!

I think I'll consolidate these hashing PRs to allow for the introduction of some helper functions.

lorentey:
Closing; please see followup PR at https://github.com/apple/swift-corelibs-foundation/pull/1656

PR #1647, opened 2018-08-02
NSLocale.hash: Implement

lorentey:
`NSLocale` inherits hash from `NSObject`, but overrides `isEqual(_:)` to compare contents. The default hash uses object identity, so two `NSLocale` instances that compare equal under `isEqual(_:)` can produce non-equal hash values. This also breaks hashing for `Locale`, which calls into `NSLocale.hash`.

Add an implementation for `NSLocale.hash` that matches `isEqual(_:)`.

Resolves https://bugs.swift.org/browse/SR-8448


lorentey:
@swift-ci test

millenomi:

PR #1648, opened 2018-08-02
NSURLComponents.hash: Implement

lorentey:
`NSURLComponents` inherits `hash` from `NSObject`, but overrides `isEqual(_:)` to compare contents. The default hash uses object identity, so two `NSURLComponent` instances that compare equal under `isEqual(_:)` can produce non-equal hash values. This also breaks hashing for `URLComponents`, which calls into `NSURLComponents.hash`.

Add an implementation for `NSURLComponents.hash` that matches `isEqual(_:)`.

Resolves https://bugs.swift.org/browse/SR-8450

### DIFF
--- a/TestFoundation/TestURL.swift
+++ b/TestFoundation/TestURL.swift
@@ -521,6 +521,7 @@ class TestURLComponents : XCTestCase {
             ("test_port", test_portSetter),
             ("test_url", test_url),
             ("test_copy", test_copy),
+            ("test_hash", test_hash),
             ("test_createURLWithComponents", test_createURLWithComponents),
             ("test_path", test_path),
             ("test_percentEncodedPath", test_percentEncodedPath),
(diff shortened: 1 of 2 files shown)


lorentey:
@swift-ci please test

TestFoundation/TestURL.swift
```diff
@@ -615,7 +616,15 @@ class TestURLComponents : XCTestCase {
         /* Assert that NSURLComponents.copy is actually a copy of NSURLComponents */ 
         XCTAssertTrue(copy.isEqual(urlComponent))
     }
-    
+
+    func test_hash() {
+        let c1 = URLComponents(string: "https://www.swift.org/path/to/file.html?id=name")!
```
millenomi: Same as in the other review, can we test changing all parts that contribute to the hash?

lorentey:
Closing; please see followup PR at https://github.com/apple/swift-corelibs-foundation/pull/1656

PR #1615, opened 2018-06-22
[4.2] Ubuntu18.04 fixes for TestURLSession

spevans:
This is a back port of various fixes to allow TestURLSession to work on Ubuntu18.04 without intermittent errors. It includes the following (some required because of what later fixes are based on)

https://github.com/apple/swift-corelibs-foundation/pull/1602
https://github.com/apple/swift-corelibs-foundation/pull/1613
https://github.com/apple/swift-corelibs-foundation/pull/1565
https://github.com/apple/swift-corelibs-foundation/pull/1498
https://github.com/apple/swift-corelibs-foundation/pull/1650

Tested on macOS, Ubuntu 16.04 and Ubuntu 18.04

PR #1650, opened 2018-08-02
When redirecting with a relative Location, add the port number to the new url

spevans:
- If the new location includes a host but no port number, then DONT add
  the port number.


spevans:
@swift-ci please test

spevans:
Not setting the port correctly along with an occasional apache server running on the CI hosts was causing #1615 to not pass the tests. The fix however should apply to `master` as well

spevans:

PR #1652, opened 2018-08-03
Fix Foundation build error with --install-prefix

loganchien:
This commit fixes a header-not-found build error when Foundation is
built with --install-prefix other than "/usr".


alblue:
@swift-ci please test

alblue:

PR #1653, opened 2018-08-05
Decimal and NSDecimalNumber fixes.

spevans:
- NSDecimalNumber int/uint getters: Dont convert the Decimal value
  using the doubleValue as this can lose the lower digits for large
  integers which cant be accuately represented as a Double.

- NSDecimalNumber.isEqual: Dont use as? to cast every value
  to an NSDecimalNumber but test individual types.

- NSNumber.compare: If the other value is a NSDecimalNumber, upgrade
  self to an NSDecimalNumber to avoid loss of precison in the other
  value.


spevans:
@swift-ci please test

spevans:
@swift-ci test

spevans:
@swift-ci test

PR #1655, opened 2018-08-06
JSONSerialization: Use NSDecimalNumber for parsing json numbers.

spevans:
- Replace strol()/strtod() parsing with Scanner.scanDecimal().
  This allows parsing of Integers > Int64.max (upto UInt64.max)
  without losing information due to the fallback parsing provided
  by parsing as a Double.

- Also allows parsing larger integers into a Decimal.

Requires PR #1653 for some `NSDecimalNumber` and `NSNumber` fixes

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -822,45 +822,61 @@ private struct JSONReader {
     ]
 
     func parseNumber(_ input: Index, options opt: JSONSerialization.ReadingOptions) throws -> (Any, Index)? {
-        func parseTypedNumber(_ address: UnsafePointer<UInt8>, count: Int) -> (Any, IndexDistance)? {
-            let temp_buffer_size = 64
-            var temp_buffer = [Int8](repeating: 0, count: temp_buffer_size)
-            return temp_buffer.withUnsafeMutableBufferPointer { (buffer: inout UnsafeMutableBufferPointer<Int8>) -> (Any, IndexDistance)? in
-                memcpy(buffer.baseAddress!, address, min(count, temp_buffer_size - 1)) // ensure null termination
-                
-                let startPointer = buffer.baseAddress!
-                let intEndPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
-                defer { intEndPointer.deallocate() }
-                let doubleEndPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
-                defer { doubleEndPointer.deallocate() }
-                let intResult = strtol(startPointer, intEndPointer, 10)
-                let intDistance = startPointer.distance(to: intEndPointer[0]!)
-                let doubleResult = strtod(startPointer, doubleEndPointer)
-                let doubleDistance = startPointer.distance(to: doubleEndPointer[0]!)
+        let ZERO = UInt8(ascii: "0")
+        let ONE = UInt8(ascii: "1")
+        let NINE = UInt8(ascii: "9")
+        let MINUS = UInt8(ascii: "-")
 
-                guard doubleDistance > 0 else { return nil }
-                if intDistance == doubleDistance {
-                    return (NSNumber(value: intResult), intDistance)
-                }
-                return (NSNumber(value: doubleResult), doubleDistance)
+        func parseTypedNumber(_ string: String) throws -> (Any, IndexDistance)? {
+            let scan = Scanner(string: string)
+            var decimal = Decimal()
+            guard scan.scanDecimal(&decimal) else {
+                return nil
             }
+            return (NSDecimalNumber(decimal: decimal), scan.scanLocation)
         }
-        
+
+        // Validate the first few characters look like a JSON encoded number:
+        // Optional '-' sign at start only 1 leading zero if followed by a decimal point.
+        var index = input
+        func nextDigit() -> UInt8? {
+            guard let (ascii, nextIndex) = source.takeASCII(index) else { return nil }
+            index = nextIndex
+            return ascii
+        }
+
+        guard var digit = nextDigit() else { return nil }
+        guard digit == MINUS || (digit >= ZERO && digit <= NINE) else { return nil }
+        if digit == MINUS {
+            guard let d = nextDigit() else { return nil }
+            digit = d
+        }
+
+        if digit == ZERO {
+            if let digit2 = nextDigit(), digit2 >= ZERO && digit2 <= NINE {
+                throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue,
+                              userInfo: ["NSDebugDescription" : "Leading zeros not allowed character \(input)." ])
+            }
+        } else if digit < ONE || digit > NINE {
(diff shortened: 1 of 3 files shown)


alblue:
@swift-ci please test

spevans:
@alblue This will currently fail tests until #1653 is merged

Foundation/JSONSerialization.swift
```diff
@@ -822,45 +822,61 @@ private struct JSONReader {
...
+                return nil
             }
+            return (NSDecimalNumber(decimal: decimal), scan.scanLocation)
         }
-        
+
+        // Validate the first few characters look like a JSON encoded number:
+        // Optional '-' sign at start only 1 leading zero if followed by a decimal point.
+        var index = input
+        func nextDigit() -> UInt8? {
```
itaiferber: Should this more clearly be called `nextASCII()`/`nextCharacter()`/similar?

Foundation/JSONSerialization.swift
```diff
@@ -822,45 +822,61 @@ private struct JSONReader {
...
-                    return (NSNumber(value: intResult), intDistance)
-                }
-                return (NSNumber(value: doubleResult), doubleDistance)
+        let ZERO = UInt8(ascii: "0")
+        let ONE = UInt8(ascii: "1")
+        let NINE = UInt8(ascii: "9")
+        let MINUS = UInt8(ascii: "-")
+
+        func parseTypedNumber(_ string: String) throws -> (Any, IndexDistance)? {
+            let scan = Scanner(string: string)
```
itaiferber: It's worth testing the performance of creating a new `Scanner` for every single number we try to parse — I suspect that this is a potentially prohibitively expensive operation.

itaiferber:
Note also that this conflicts with the work in #1654 so we might need to coordinate something here.

itaiferber:
@spevans I was quickly putting together an approach similar to #1654 — I agree that it can involve up to 3 parses but they should be relatively short and quick as the input is limited in length. Otherwise, though: how do we know how we want to use a parsed number? Aren't these values just returned to the developer directly? Or are you proposing a _further_ deferral that will only parse the value once it's read and cast out (e.g. a number that totally defers evaluation until you try to access it)?

(If so, it doesn't necessarily improve things as it's totally valid to write `if let num as? Int /* fail */ { ... } else if let num as? UInt64 /* fail */ { ... } else if let num as? Double /* success */ { ... }` which doesn't decrease the number of attempted parses.)

spevans:
I was thinking out not parsing until its accessed and with `Codable` doesn't it know the type it wants? With the #1654 solution every `Double` is parsed 3 times, which only matches the worst case you listed above.

Also, if it is a subclass of `NSNumber` the other properties are still available (eg `intValue`) which is currently lossy anyway for out of range numbers

spevans:
FYI, here is the version I had using the `strto*` functions:
```
func parseNumber(_ input: Index, options opt: JSONSerialization.ReadingOptions) throws -> (Any, Index)? {
        func parseTypedNumber(_ address: UnsafePointer<UInt8>, count: Int) throws -> (Any, IndexDistance)? {
            let temp_buffer_size = 64
            var temp_buffer = [CChar](repeating: 0, count: temp_buffer_size)

            return try temp_buffer.withUnsafeMutableBufferPointer { (buffer: inout UnsafeMutableBufferPointer<CChar>) throws -> (Any, IndexDistance)? in
                memcpy(buffer.baseAddress!, address, min(count, temp_buffer_size - 1)) // ensure null termination

                var chIndex = 0
                var ch = buffer[chIndex]
                let isNegative: Bool
                if ch == CChar(UInt8(ascii: "-")) {
                    isNegative = true
                    chIndex = chIndex + 1
                    ch = buffer[chIndex]
                } else {
                    isNegative = false
                }

                guard ch >= CChar(UInt8(ascii: "0")) && ch <= CChar(UInt8(ascii: "9")) else {
                    return nil
                }

                // Only allow 1 leading zero and it must be followed by a decimal point
                if ch == CChar(UInt8(ascii: "0")) && buffer[chIndex + 1] == CChar(UInt8(ascii: "0")) {
                    throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue,
                                  userInfo: ["NSDebugDescription" : "Number with leading zero around character 16." ])
                }

                let startPointer = buffer.baseAddress!
                let endPointer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 1)
                defer { endPointer.deallocate() }

                func parseAsDouble() -> (Double, IndexDistance) {
                    let doubleResult = strtod(startPointer, endPointer)
                    let distance = startPointer.distance(to: endPointer[0]!)
                    return (doubleResult, distance)
                }

                func parseAsInt64() -> (Int64?, IndexDistance) {
                    errno = 0
                    let int64Result = strtoll(startPointer, endPointer, 10)
                    let outOfRange = ((int64Result == Int64.max || int64Result == Int64.min) && errno == ERANGE)
                    let invalid = (int64Result == 0 && errno == EINVAL)
                    let distance = startPointer.distance(to: endPointer[0]!)

                    if distance == 0 || outOfRange || invalid {
                        return (nil, distance)
                    } else {
                        return (int64Result, distance)
                    }
                }

                func parseAsUInt64() -> (UInt64?, IndexDistance) {
                    // strtoull ignores a leading '-' and always returns a positive value
                    // so check against the flag set earlier
                    guard !isNegative else { return (nil, 0) }
                    errno = 0
                    let uint64Result = strtoull(startPointer, endPointer, 10)
                    let outOfRange = (uint64Result == UInt64.max && errno == ERANGE)
                    let invalid = (uint64Result == 0 && errno == EINVAL)
                    let distance = startPointer.distance(to: endPointer[0]!)

                    if distance == 0 || outOfRange || invalid {
                        return (nil, distance)
                    } else {
                        return (uint64Result, distance)
                    }
                }

                // Try parsing as a Double, checks that it is some sort of number
                let (doubleResult, doubleDistance) = parseAsDouble()
                guard doubleDistance > 0 else { return nil }

                let (int64Result, int64distance) = parseAsInt64()
                if let result = int64Result {
                    if int64distance == doubleDistance {
                        // The parsed number can be represented as an Int64
                        return (NSNumber(value: result), int64distance)
                    } else {
                        return (NSNumber(value: doubleResult), doubleDistance)
                    }
                }

                // Int64 wasnt sufficient, try parsing as a UInt64
                let (uint64Result, uint64distance) = parseAsUInt64()
                if let result = uint64Result {
                    return (NSNumber(value: result), uint64distance)
                }

                // This double result will be 0 or infinity for out of range values. How should they be
                // treated?
                return (NSNumber(value: doubleResult), doubleDistance)
            }
        }

        if source.encoding == .utf8 {
            return try parseTypedNumber(source.buffer.baseAddress!.advanced(by: input), count: source.buffer.count - input).map { return ($0.0, input + $0.1) }
        }
        else {
            var numberCharacters = [UInt8]()
            var index = input
            while let (ascii, nextIndex) = source.takeASCII(index), JSONReader.numberCodePoints.contains(ascii) {
                numberCharacters.append(ascii)
                index = nextIndex
            }
            numberCharacters.append(0)

            return try numberCharacters.withUnsafeBufferPointer {
                try parseTypedNumber($0.baseAddress!, count: $0.count)
                }.map { return ($0.0, index) }
        }
    }
```

spevans:
Closed in favour of #1657

PR #1649, opened 2018-08-02
Add custom AnyHashable representations to Foundation types with bridging

lorentey:
When converted to `AnyHashable`, both forms of bridged types should hash the same way and compare as identical. To implement this, we need to add `_HasCustomAnyHashableRepresentation` conformances to all bridgable `NSObject` subclasses.

This PR brings the set of types with custom `AnyHashable` representations in swift-corelibs-foundation in sync with the Foundation overlay in apple/swift.

Resolves https://bugs.swift.org/browse/SR-7436

Note that the new tests require the hashing fixes in #1645, #1646, #1647 and #1648.

PR #1645, opened 2018-08-02
NSDateComponents.hash: Fix arithmetic overflows

lorentey:
`NSDateComponents.hash` currently hashes 13 components, while `isEqual(_:)` compares 17 components. The expression that produces the hash value often traps because of arithmetic overflows. 

Replace the `hash` implementation with one that uses `Hasher`, feeding it with the exact same components that are compared in equality checks.

Resolves https://bugs.swift.org/browse/SR-8441

PR #1656, opened 2018-08-06
DateComponents, URLRequest, URLComponents: Fix hashing

lorentey:
This PR collects fixes previously submitted separately as #1645, #1646 and #1648, fixing the implementation of hashing for the following Foundation types:

- `DateComponents`
- `NSDateComponents`
- `URLRequest`
- `NSURLRequest`
- `URLComponents`
- `NSURLComponents`

Note that this isn't a full audit of Foundation hashing -- this was the bare minimum necessary for #1649. I suspect there may be more types with issues similar to these.

Resolves:
- https://bugs.swift.org/browse/SR-8441
- https://bugs.swift.org/browse/SR-8449
- https://bugs.swift.org/browse/SR-8450

### DIFF
--- a/TestFoundation/Utilities.swift
+++ b/TestFoundation/Utilities.swift
@@ -1,0 +1,105 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+
+func checkHashableMutations_ValueType<Item: Hashable, S: Sequence>(
+    _ item: Item,
+    _ keyPath: WritableKeyPath<Item, S.Element>,
+    _ values: S,
+    file: StaticString = #file,
+    line: UInt = #line
+) {
+    _checkHashableMutations(
+        Item.self, Item.self,
+        item,
+        { $0 },
+        keyPath,
+        values)
+}
+
+func checkHashableMutations_NSCopying<Item: NSObject & NSCopying, S: Sequence>(
+    _ item: Item,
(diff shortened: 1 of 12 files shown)


lorentey:
@swift-ci please test

TestFoundation/Utilities.swift
```diff
@@ -0,0 +1,105 @@
...
+        Source.self, Target.self,
+        item,
+        { $0.mutableCopy() as! Target },
+        keyPath,
+        values)
+}
+
+// Check that mutating `object` via the specified key path affects its
+// hash value.
+func _checkHashableMutations<Source: Hashable, Target: Hashable, S: Sequence>(
```
millenomi: I know these are a ton of arguments and it's essentially fileprivate, but honestly I'd have a more descriptive set of labels because it is _really_ hard to parse the invocations above at a glance.

e.g.: `_checkHashing(of type, matches type, initialValue:, copyHandler:, byMutating keyPaths, throughRange:)`

lorentey:

lorentey:
Oh, swift-ci can’t merge here!

@swift-ci test

spevans:
@lorentey There is an Xcode project, `DarwinCompatibilityTests` in the toplevel that can be used to run the `TestFoundation` tests against Darwin's native foundation. The tests should show up in the unit test pane, although not all of them currently pass.

lorentey:
I managed to run these tests on Darwin's Foundation, and filed rdar://problem/43422017 with my findings. While Darwin Foundation (erroneously) neglects to mix some of the properties that affect equality into the hash value, I believe there is no need for swift-corelibs-foundation to emulate this behavior. (Emulating exact hash values would not be a worthy exercise anyway, and given the lack of random seeding, it may even be actively harmful to do so in server contexts.) So I think the definitions can stay as defined here.

lorentey:
Now if only I could get PR testing to update GitHub status... The tests seem to finish successfully, but the PR stays yellow.

@swift-ci test

PR #1657, opened 2018-08-08
JSONSerialization: Improve parsing of numbers

spevans:
- Check the number looks like a JSON number and exit early if not.

- Use the native Int64(), UInt64(), Double() parsers to avoid creating
  a C string and passing to strtol()/strtod(). This also eliminates a
  memcpy() and removes the 63 digit restriction which would fail to
  parse numbers expressible by Double's full exponent.

- For numbers with a leading '-' sign, parse using Int64() falling
  back to Double(), otherwise parse using UInt64() failling back to
  Double().

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -816,66 +816,145 @@ private struct JSONReader {
     }
     
     //MARK: - Number parsing
-    static let numberCodePoints: [UInt8] = [
-        0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, // 0...9
-        0x2E, 0x2D, 0x2B, 0x45, 0x65, // . - + E e
-    ]
+    private static let ZERO = UInt8(ascii: "0")
+    private static let ONE = UInt8(ascii: "1")
+    private static let NINE = UInt8(ascii: "9")
+    private static let MINUS = UInt8(ascii: "-")
+    private static let PLUS = UInt8(ascii: "+")
+    private static let LOWER_EXPONENT = UInt8(ascii: "e")
+    private static let UPPER_EXPONENT = UInt8(ascii: "E")
+    private static let DECIMAL_SEPARATOR = UInt8(ascii: ".")
+    private static let allDigits = (ZERO...NINE)
+    private static let oneToNine = (ONE...NINE)
+
+    private static let numberCodePoints: [UInt8] = {
+        var numberCodePoints = Array(ZERO...NINE)
+        numberCodePoints.append(contentsOf: [DECIMAL_SEPARATOR, MINUS, PLUS, LOWER_EXPONENT, UPPER_EXPONENT])
+        return numberCodePoints
+    }()
+
 
     func parseNumber(_ input: Index, options opt: JSONSerialization.ReadingOptions) throws -> (Any, Index)? {
-        let ZERO = UInt8(ascii: "0")
-        let ONE = UInt8(ascii: "1")
-        let NINE = UInt8(ascii: "9")
-        let MINUS = UInt8(ascii: "-")
 
         var isNegative = false
         var string = ""
-
-        // Validate the first few characters look like a JSON encoded number:
-        // Optional '-' sign at start only 1 leading zero if followed by a decimal point.
+        var isInteger = true
+        var exponent = 0
+        var positiveExponent = true
         var index = input
-        func nextASCII() -> UInt8? {
-            guard let (ascii, nextIndex) = source.takeASCII(index),
-                JSONReader.numberCodePoints.contains(ascii) else { return nil }
-            index = nextIndex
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci please test

Foundation/JSONSerialization.swift
```diff
@@ -822,46 +822,64 @@ private struct JSONReader {
...
-                guard doubleDistance > 0 else { return nil }
-                if intDistance == doubleDistance {
-                    return (NSNumber(value: intResult), intDistance)
+        let ZERO = UInt8(ascii: "0")
+        let ONE = UInt8(ascii: "1")
+        let NINE = UInt8(ascii: "9")
+        let MINUS = UInt8(ascii: "-")
+
+        var isNegative = false
+        var string = ""
```
itaiferber: Should we try to preserve the UTF-8 behavior here of walking a pointer along, rather than appending a single character at a time here? (Given that the vast majority of JSON provided is given to us in UTF-8, it'd be nice to maintain the performance there.)

spevans: I did think about this but strings on x86_64/ARM64 upto 15 ASCII characters should actually fit into a SSO so avoid a memory allocation which probably covers most numbers to be parsed.

The other issue is that strings passed to `Int64()`, `UInt64()` and `Double()` cant have any invalid trailing characters so this avoids creating a string of all the available characters using `String(bytesNoCopy:)` which I believe still  get validated according to the encoding which could end up reading through the whole of the rest of the JSON document and then scanning through it to determine the new shorter count.

As an performance enhancement, when validating the characters its possible to count the number of integers and look for `.eE` and directly jump to parsing as a `Double()`

spevans:
For `Decimal` do you mean with a change like:
```
                 return (NSNumber(value: uintValue), index)
             }
         }
+        let decimalNumber = NSDecimalNumber(string: string)
+        if decimalNumber.isFinite {
+            return (decimalNumber, index)
+        }
         if let doubleValue = Double(string) {
             return (NSNumber(value: doubleValue), index)
         }
```
?

I think #1653 is needed before `Decimal` works properly but I can always do it as a follow-up

Foundation/JSONSerialization.swift
itaiferber: I was thinking more along the lines of [`String.init(bytesNoCopy:length:encoding:freeWhenDone:)`](https://developer.apple.com/documentation/swift/string/1408702-init) which would allow us to avoid reading to the end of the document and wouldn't necessitate doing any further validation. Might be worth doing some small perf tests, just to see. (Or is this not available in s-cl-f?)

As for looking for `[.eE]` — we do just this on Darwin: as soon as we encounter one of those characters we avoid parsing as an integer unnecessarily, which does save some time in common situations.

itaiferber:
As for `Decimal`, I was thinking about the heuristic I mentioned briefly in #1655. The string representation of a `Double` can have at most `DBL_DECIMAL_DIG` digits of precision (17 for an IEEE `Double`) before you start to lose precision. This means that if a string is longer than:

* 1 (if it has a sign) +
* 1 (if it has a decimal point and starts with a leading `0` like `0.xxxxxxxxxx`) +
* 1 (if it has a decimal point) +
* `DBL_DECIMAL_DIG` +
* `E` (if it has an exponent of length `E`, max 5 digits for `e±308`)

then you will lose precision on the parse. For instance,

* Strings in the form `"xxxxxxxxxx"`: max length 17
* Strings in the form `"±xxxxxxxxxx"`: max length 18
* Strings in the form `"xxxxx.yyyyy"`: max length 18
* Strings in the form `"xxxxx.yyyyye±zzz"`: max length 23
* etc.

If the string is longer than this (we're losing precision) but parsing succeeds (it's a valid `Double`) and the magnitude would fit in a `Decimal`, then it might be worth parsing as a `Decimal` to avoid losing that precision. (This is what we do on Darwin, at least.)

Foundation/JSONSerialization.swift
spevans: Unfortunately from https://github.com/apple/swift-corelibs-foundation/blob/a2b40951e8365da696d5105fd57a19c1f1c220ef/Foundation/NSString.swift#L1237:

```
public convenience init?(bytesNoCopy bytes: UnsafeMutableRawPointer, length len: Int, encoding: UInt, freeWhenDone freeBuffer: Bool) /* "NoCopy" is a hint */ {
        if freeBuffer { // dont take the hint
            free(bytes)
        }
    }
```

So I don't think its that useful at the moment. I will look into bypassing the integer parsing where possible

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

millenomi:

PR #1658, opened 2018-08-10
SR-7011: Fatal Error when Using Calendar.current.dateComponents(_:from:to:)

spevans:
- _CFCalendarGetComponentDifferenceV() doesn't support a component
  character of 'l' (leap year) as it make no sense for the difference
  between two dates.

- Dont add 'l' to the component string or parse the result into the
  .isLeapMonth property when called from
  NSCalendar.components(_:from:to:options)

### DIFF
--- a/CoreFoundation/Locale.subproj/CFCalendar.c
+++ b/CoreFoundation/Locale.subproj/CFCalendar.c
@@ -953,7 +953,12 @@ Boolean _CFCalendarGetComponentDifferenceV(CFCalendarRef calendar, CFAbsoluteTim
     while (ch) {
         UCalendarDateFields field = __CFCalendarGetICUFieldCodeFromChar(ch);
         const int multiple_table[] = {0, 0, 16, 19, 24, 26, 24, 28, 14, 14, 14};
-        int multiple = direction * (1 << multiple_table[flsl(__CFCalendarGetCalendarUnitFromChar(ch)) - 1]);
+        int unit = __CFCalendarGetCalendarUnitFromChar(ch);
+        if (unit < 0) {
+            // Not a valid calendar unit character
+            return false;
+        }
+        int multiple = direction * (1 << multiple_table[flsl(unit) - 1]);
         Boolean divide = false, alwaysDivide = false;
         int result = 0;
         while ((direction > 0 && curr < goal) || (direction < 0 && goal < curr)) {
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci please test

TestFoundation/TestCalendar.swift
```diff
@@ -223,4 +224,16 @@ class TestNSDateComponents: XCTestCase {
         XCTAssertEqual(components.hour, 12)
         XCTAssertEqual(copy.hour, 14)
     }
+
+    func test_dateDifferenceComponents() {
+        let date1 = Date(timeIntervalSince1970: 0)
+
+        // 1971-06-21
+        let date2 = Date(timeIntervalSince1970:  46310400)
+        let difference = Calendar.current.dateComponents([.month, .year, .day], from: date1, to: date2)
```
phausler: for leap months you probably need the Chinese calendar iirc.

itaiferber: On Darwin we actually almost always add "is leap month" to `DateComponents` in output, if that's what you're referring to. (Leap months are also applicable in the Hebrew calendar, and otherwise.)

itaiferber: (But yes, in this case, on Darwin we don't add "is leap month" here.)

CoreFoundation/Locale.subproj/CFCalendar.c
```diff
@@ -980,8 +985,13 @@ Boolean _CFCalendarGetComponentDifferenceV(CFCalendarRef calendar, CFAbsoluteTim
             divide = alwaysDivide;
         }
         }
+        if (count < 1) {
```
itaiferber: Should this be moved further down (right after the `--`) so we don't go through the loop an additional time and do work just to discover we can't get the difference?

itaiferber:
I know the unit tests were non-existent before, but it'd be nice to expand them further if we can before we merge. A variety of dates (future and past relative to one another) and a variety of components would be great!

CoreFoundation/Locale.subproj/CFCalendar.c
spevans: It should probably go at the top of the while loop in case `count` starts at `0`. I will fix that

itaiferber: Ah, yes — good catch

spevans:
@swift-ci please test

spevans:
@itaiferber Ive added some extra tests and validated them against Darwin. There are probably some more cases to catch but this should cover a reasonable amount and does add some extra validation into the CF function.

spevans:

millenomi:

PR #1660, opened 2018-08-14
Allow header files to be parsed by C++ parsers

alblue:
By replacing `namspace` with `name_space` and changing the `restrict` into `__restrict__` in C++ mode, the headers can now be parsed by clang++ as well as clang.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -82,6 +82,17 @@
 #define _CF_RELEASES_ARGUMENT_OBJ
 #endif
 
+#if DEPLOYMENT_TARGET_WINDOWS
+  // No C99 support
+  #define _CF_RESTRICT
+#else
+  #if defined(__cplusplus)
+    #define _CF_RESTRICT __restrict__
+  #else
+    #define _CF_RESTRICT restrict
+  #endif
+#endif
+
 CF_EXTERN_C_BEGIN
 
 #include <CoreFoundation/CFBase.h>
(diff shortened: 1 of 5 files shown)


alblue:
@swift-ci please test

CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -144,6 +144,17 @@
...
+  // No C99 support
+  #define CF_RESTRICT
+#else
+  #if defined(__cplusplus)
+    #define CF_RESTRICT __restrict__
+  #else
+    #define CF_RESTRICT restrict
+  #endif
+#endif
+
```
parkera: This is a public header so we can't just add the macro here. Perhaps `ForSwiftFoundationOnly.h`?

alblue: The `restrict` is used in CFPlatform.c and CFStorage.c, so could move it into CFInternal.h instead?

parkera: I was thinking CFInternal as well, but then it's used in ForSwiftFoundationOnly.h. I'd prefer the internal header if possible.

alblue: So it looks like if we put it in `CFInternal.h` then it can't be used for the ForSwiftFoundationOnly.h. We could have the macro in CFInternal for ease of maintenance and then have a special-case for the ForSwiftFoundationOnly that duplicates it, but it's not elegant.

alblue:
@swift-ci please test

CoreFoundation/Base.subproj/CFBase.h
alblue: So @parkera I have moved the definitions to be internal only, but can't use it in the ForSwiftFoundation file. I'm not sure which approach you'd like me to take here in terms of conditionally defining the value of `restrict` -- I've duplicated the line here, but it's introducing duplicate code which I don't like. Any other preferences/suggestions?

alblue:
@swift-ci please test

alblue:
@parkera any thoughts as to the new approach suggested here?

alblue:

PR #1662, opened 2018-08-15
SR-8519: Linux standardizingPath syscall performance

spevans:
- Filemanager should not use attributesOfItem(atPath:) internally
  when only the file type of the path is required, to avoid
  unnecessary getpwuid() and getgrgid() calls.


spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:

PR #1664, opened 2018-08-20
SR-8560: XMLDocument on Linux crashes with a Segmentation Fault

spevans:
- When returning an XMLNode, take a reference to the parent
  XMLDocument to prevent the underlying libxml xml document
  from being freed.

- This prevents the document being freed if a reference is still
  held to a XMLNode but not the XMLDocument.

### DIFF
--- a/Foundation/XMLNode.swift
+++ b/Foundation/XMLNode.swift
@@ -91,6 +91,7 @@ open class XMLNode: NSObject, NSCopying {
     }
 
     internal let _xmlNode: _CFXMLNodePtr
+    internal var _xmlDocument: XMLDocument?
 
     open func copy(with zone: NSZone? = nil) -> Any {
         let newNode = _CFXMLCopyNode(_xmlNode, true)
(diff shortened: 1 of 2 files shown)


parkera:
Looks good to me.

spevans:

Foundation/XMLNode.swift
```diff
@@ -790,6 +793,11 @@ open class XMLNode: NSObject, NSCopying {
         withUnretainedReference {
             _CFXMLNodeSetPrivateData(_xmlNode, $0)
         }
+        if let documentPtr = _CFXMLNodeGetDocument(_xmlNode) {
+            if documentPtr != ptr {
+                _xmlDocument = XMLDocument._objectNodeForNode(documentPtr)
```
michelf: Won't this create a cyclic strong reference? The document has a reference to its children, and children have a reference to the document. The objects will never be deallocated unless you detach all the children to break the cycle.

PR #1668, opened 2018-08-26
SR-5829: Fix NSString.commonPrefix(with:options:)

spevans:
- Also implement .caseInsensitive option.

Current implementation was buggy and would crash for some strings.
Tests now match Darwin

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -434,49 +434,31 @@ extension NSString {
     }
     
     public func commonPrefix(with str: String, options mask: CompareOptions = []) -> String {
-        var currentSubstring: CFMutableString?
-        let isLiteral = mask.contains(.literal)
-        var lastMatch = NSRange()
-        let selfLen = length
-        let otherLen = str.length
-        var low = 0
-        var high = selfLen
-        var probe = (low + high) / 2
-        if (probe > otherLen) {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

spevans:
@itaiferber Could you have a look at this string one as well please :)

spevans:
According to https://developer.apple.com/documentation/foundation/nsstring/1408169-commonprefix only `NSCaseInsensitiveSearch` and `NSLiteralSearch` are supported for `commonPrefix(with:options:)` unless Im missing something?

I haven't done any benchmarking because the binary search method didn't actually work and I didn't fully understand how to fix it. I decided to work directly treating the string as a `Character` array and that does not have O(1) performance for random access so I used iterators instead.

itaiferber:
@spevans Hmm, indeed — it looks like we don't support things like `.diacriticInsensitive` or `.widthInsensitive` there on Darwin. Scratch that, then. :)

itaiferber:
Re: perf testing, understandable. The original code does look very much like the actual implementation in `NSString` on Darwin, so I wonder what the differential was there. I'll try to take a look when I have time to investigate.

spevans:
@swift-ci test

itaiferber:

PR #1669, opened 2018-08-26
SR-936: NSMutableData.hash() causes a segmentation fault on Linux

spevans:
- CFHash() doesnt see objects of type NSMutableData as CF objects
  so calls back to NSObject.hash which causes infinite recursion.

- Directly call _CFNonObjCHash() to bypass this check.

### DIFF
--- a/Foundation/NSData.swift
+++ b/Foundation/NSData.swift
@@ -288,7 +288,7 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
 
     // MARK: - NSObject methods
     open override var hash: Int {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci please test

parkera:
Yikes. This would be good for the 4.2 branch.

spevans:
@parkera Would this affect other classes? Should `_CFNonObjCHash` be used instead of `CFHash` for all implementations of `hash()` in the `NS*` classes?

parkera:
That is reasonable. It looks like most of our Darwin ObjC implementations simply implement their own hash instead of using the CFHash function in the first place.

PR #1670, opened 2018-08-27
isDeletableFile Implementation + Tests

rauhul:
Tried to best match the suggested implementation from the [Swift Forum](https://forums.swift.org/t/swiftfoundation-filemanager-isdeletablefile-implementation/13465/7).

I wasn't exactly sure how to write tests for these methods. My tests attempt to check that the result of `isReadableFile`, `isWritableFile`, `isExecutableFile` match the value one would receive by directly checking the file's permission bits.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -744,9 +744,41 @@ open class FileManager : NSObject {
             access($0, X_OK) == 0
         })
     }
-    
+
+    /**
+     - parameters:
+        - path: The path to the file we are trying to determine is deletable.
+
+      - returns: `true` if the file is deletable, `false` otherwise.
+     */
     open func isDeletableFile(atPath path: String) -> Bool {
-        NSUnimplemented()
+        // Get the parent directory of supplied path
+        let parent = path._nsObject.deletingLastPathComponent
+
+        return _fileSystemRepresentation(withPath: parent, andPath: path, { parentFsRep, fsRep  in
+            // Check the parent directory is writeable, else return false.
+            guard access(parentFsRep, W_OK) == 0 else {
+                return false
+            }
+
+            // Stat the parent directory, if that fails, return false.
+            guard let parentS = try? _lstatFile(atPath: path, withFileSystemRepresentation: parentFsRep) else {
+                return false
+            }
+
+            // Check if the parent is 'sticky' if it exists.
+            if (parentS.st_mode & S_ISVTX) == S_ISVTX {
+                guard let s = try? _lstatFile(atPath: path, withFileSystemRepresentation: fsRep) else {
+                    return false
+                }
+
+                // If the current user owns the file, return true.
+                return s.st_uid == getuid()
+            }
+
+            // Return true as the best guess.
+            return true
+        })
     }
 
     private func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64, bufSize: Int) -> Bool {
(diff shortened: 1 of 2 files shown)


TestFoundation/TestFileManager.swift
```diff
@@ -202,6 +206,62 @@ class TestFileManager : XCTestCase {
...
+            XCTAssertTrue(fileIsReadableFile == fmIsReadableFile)
+        } catch let e {
+            XCTFail("\(e)")
+        }
+    }
+
+    func test_isDeletableFile() {
+        // TODO: Implement test
+        // how to test?
+    }
```
spevans: You could create a file and check its is deletable. Also check a file that should fail eg `/dev/null` and also check `/` to validate the parent checking code

rauhul: Just to confirm, we would expect the created file to be deletable, correct?

spevans: Yes, thats correct. You should be able to create a file that fails by putting it into a subdirectory with no permissions:

eg 
```
mkdir dir1
touch dir1/file1
chmod 000 dir1

isDeletableFile("dir1/file1") -> false

chmod 755 dir1

isDeletableFile("dir1/file1") -> true
```

rauhul:
@spevans does the `test_isDeletableFile` implementation look good?

spevans:
@swift-ci test

TestFoundation/TestFileManager.swift
```diff
@@ -225,41 +225,61 @@ class TestFileManager : XCTestCase {
...
         do {
             let attrs = try fm.attributesOfItem(atPath: path)
             let permissions = attrs[FileAttributeKey.posixPermissions] as! UInt16
-            let fileIsReadableFile = (permissions & S_IXUSR == S_IXUSR)
-            let fmIsReadableFile = fm.isReadableFile(atPath: path)
-            XCTAssertTrue(fileIsReadableFile == fmIsReadableFile)
+            let fileIsExecutableFile = (permissions & S_IXUSR == S_IXUSR)
+            let fmIsExecutableFile = fm.isExecutableFile(atPath: path)
+            XCTAssertTrue(fileIsExecutableFile == fmIsExecutableFile)
         } catch let e {
```
spevans: I would modify this test to set the executable bit on the file and then simply test it with:
`XCTAssertTrue(fm.isExecutableFile(atPath: path))` (both before and after setting the bit)

Testing equality this way could mask an issue if the `let fileIsExecutableFile = (permissions & S_IXUSR == S_IXUSR)` was wrong.

rauhul: Makes sense, will do

spevans:
@swift-ci test

rauhul:
@parkera @spevans can this be merged?

alblue:
Seems reasonable to me. WDYT @millenomi @phausler ?

parkera:
@millenomi you want to do the final approval here?

Foundation/FileManager.swift
```diff
@@ -744,9 +744,41 @@ open class FileManager : NSObject {
...
+    /**
+     - parameters:
+        - path: The path to the file we are trying to determine is deletable.
+
+      - returns: `true` if the file is deletable, `false` otherwise.
+     */
     open func isDeletableFile(atPath path: String) -> Bool {
-        NSUnimplemented()
+        // Get the parent directory of supplied path
+        let parent = path._nsObject.deletingLastPathComponent
```
millenomi: This is pretty close to the Darwin implementation! That implementation also special-cases the empty string to mean the current directory, but it's not behavior that's documented, so I don't think we need to port it.

millenomi:

PR #1672, opened 2018-08-27
SR-8572: Rework OperationQueue.isSuspended to stop suspending underly…

maksimorlovich:
…ing DispatchQueue

- Suspending the underlying DispatchQueue causes issues when an app needs to
  use the same DispatchQueue for scheduling work from multiple OperationQueue's
  and control which OperationQueue's are suspended or not


maksimorlovich:
This needs to be tested against master please. @parkera @spevans Thanks!

parkera:
@swift-ci test and merge

PR #1677, opened 2018-08-29
SR-7455: Allow NUL in Strings to match Darwin

spevans:
- String._conditionallyBridgeFromObjectiveC:
  Use String(decoding:as: UTF8.self) instead of String(cString:) to
  allow NULs in the decoded string.

- Update tests which use buffers with trailing zeros to create strings,
  making the tests work correctly against Darwin's native Foundation.


spevans:
@swift-ci test

Foundation/String.swift
```diff
@@ -29,10 +29,13 @@ extension String : _ObjectiveCBridgeable {
             result = source._storage
         } else if type(of: source) == _NSCFString.self {
             let cf = unsafeBitCast(source, to: CFString.self)
+            let length = CFStringGetLength(cf)
             if let str = CFStringGetCStringPtr(cf, CFStringEncoding(kCFStringEncodingUTF8)) {
-                result = String(cString: str)
+                result = str.withMemoryRebound(to: UInt8.self, capacity: length) { ptr in
+                    let buffer = UnsafeBufferPointer(start: ptr, count: length)
+                    return String(decoding: buffer, as: UTF8.self)
+                }
```
parkera: This looks good. If you're open to a small addition to your fix, we may want to also consider trying to call `CFStringGetCharactersPtr` if `CFStringGetCStringPtr` fails. That one may return a pointer to a UTF16 encoded buffer (or NULL), which we can also use without re-allocating in the slow path below.

parkera:
```
/* These functions attempt to return in O(1) time the desired format for the string.
Note that although this means a pointer to the internal structure is being returned,
this can't always be counted on. Please see note at the top of the file for more
details.
*/

CF_EXPORT
const UniChar *CFStringGetCharactersPtr(CFStringRef theString);  /* May return NULL at any time; be prepared for NULL, if not now, in some other time or place. See discussion at top of this file. */
```

spevans: @pakera I will look into making that change. BTW, is it correct to use `CFStringGetLength` to find the number of bytes in the UTF8 string or is there something better to use?

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@parkera does the 2nd commit look ok?

millenomi:
(Where ‘characters’ means ‘UTF-16 code points’.)

spevans:
@millenomi Are `CFString`s always UTF-16? Is there a way to get the string length in bytes?

xwu:
I've made four attempts at this and no longer have the time to do so, so I'm glad you're taking a look. Theres are the four attempts, which hopefully will give some insight:

#496  
#549  
#604  
#1253

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@millenomi I think I have addressed the UTF-8 multibyte issue when using `CFStringGetCStringPtr()` by requesting  `kCFStringEncodingASCII` instead of `kCFStringEncodingUTF8` so the code point count from `CFStringGetLength()` should be the number of ASCII bytes.

Some testing actually revealed that constructing a multibyte UTF-8 `CFString` made `CFStringGetCStringPtr(cfstr, kCFStringEncodingUTF8)` return `NULL` anyway.

spevans:
@millenomi Can you check if my last comment about only requesting an ASCII string when using ` CFStringGetCStringPtr()` and `CFStringGetLength()` makes sense?

millenomi:
Yeah, in that encoding byte length is guaranteed to be equal to character length.

spevans:
@swift-ci test and merge

spevans:
@swift-ci test and merge

millenomi:

spevans:

millenomi:
@spevans I missed the fact this has conflicts; even with them, tho, the build failure is worrying. Let's try again tomorrow.

spevans:
@swift-ci test

spevans:
The Mojave merge had some whitespace that caused a conflict, Ive just rebased. 

But yes, that build error is odd and not one Ive seen before.

spevans:

spevans:

PR #1678, opened 2018-08-29
Use FixedWidthInteger rather than BinaryInteger in NSError interoperability

DougGregor:


DougGregor:
@swift-ci please test

PR #1679, opened 2018-08-31
Fix argument order for CFArraySetValueAtIndex.

drodriguez:
The Swift dispatching of CFArraySetValueAtIndex had the arguments for
value and index switched around. If someone creates a Swift NSArray
which is eventually bridged to CFArray, and some other piece calls
CFArraySetValueAtIndex, CF_SWIFT_FUNCDISPATCHV will delegate to the
Swift method, but will send a pointer where CFIndex is expected, and a
CFIndex where a pointer is expected, probably causing a segmentation
fault.

The test tries to simplify the scenario (not easy to create the right
conditions) and uses unsafeBitCast (which is the same that happens in
the internal _cfObject). The test crashes before the patch is applied.

### DIFF
--- a/CoreFoundation/Collections.subproj/CFArray.c
+++ b/CoreFoundation/Collections.subproj/CFArray.c
@@ -628,7 +628,7 @@ void CFArrayAppendValue(CFMutableArrayRef array, const void *value) {
 }
 
 void CFArraySetValueAtIndex(CFMutableArrayRef array, CFIndex idx, const void *value) {
-    CF_SWIFT_FUNCDISPATCHV(CFArrayGetTypeID(), void, (CFSwiftRef)array, NSMutableArray.setObject, idx, value);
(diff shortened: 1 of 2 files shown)


CoreFoundation/Collections.subproj/CFArray.c
```diff
@@ -628,7 +628,7 @@ void CFArrayAppendValue(CFMutableArrayRef array, const void *value) {
 }
 
 void CFArraySetValueAtIndex(CFMutableArrayRef array, CFIndex idx, const void *value) {
-    CF_SWIFT_FUNCDISPATCHV(CFArrayGetTypeID(), void, (CFSwiftRef)array, NSMutableArray.setObject, idx, value);
+    CF_SWIFT_FUNCDISPATCHV(CFArrayGetTypeID(), void, (CFSwiftRef)array, NSMutableArray.setObject, value, idx);
```
millenomi: We should really name all the argument in order in the bridge struct field to prevent this from occurring again. Is there any way I can convince you to add an edit so this reads NSMutableArray.setObjectAtIndex?

✅ whether this is done or not.

PR #1681, opened 2018-08-31
Fix DateFormatter behaviour for nil locale and timeZone.

drodriguez:
The locale and timeZone properties of DateFormatter seems to be "null
resetteable" in macOS, but the Swift Foundation behaviour differs.

Before this patch, setting locale to nil would segfault the program the
next time the formatter is used, while timeZone will incorrectly return
nil. After this patch a nil locale will simply reset to the current
locale and a nil timeZone will reset to the system one.

The tests check that the initial defaults are also correctly set.


millenomi:
@swift-ci please test

PR #1682, opened 2018-09-01
Fix Dimension.encode(with:) to use same key as init(coder:).

drodriguez:
The encoding method and the decoding method used slightly different
keys. The incorrect key made Dimension (and the subtypes) impossible to
decode (and with Swift Foundation a crash since decoding cannot fail).

Includes a test to check for correctly encoding/decoding cycle.


spevans:
@swift-ci test

maksimorlovich:
LGTM!

parkera:
Thanks!

PR #1686, opened 2018-09-07
Fix unnecessary (and broken) preprocessor exclusion

copumpkin:
I'm not 100% sure this is the correct fix, but the `__nameStringOK` static function gets used even if `DEPLOYMENT_RUNTIME_SWIFT` is unset, so this just breaks the build (linkage) if this block of code gets excluded. It also just doesn't seem very Swift-specific, to my uneducated eye.

cc @DougGregor or anyone willing to suggest a better fix


parkera:
LGTM

parkera:
@swift-ci test and merge

copumpkin:
Thanks @parkera!

PR #1687, opened 2018-09-08
SR-8689: Fix crash with NSMutableString.replaceOccurrences(of:with:options:)

spevans:
- "\r\n" is treated as a single character, but so are "\r" and "\n"
  so use rangeOfComposedCharacterSequence(at:) to find the range of the
  decomposed character.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -1331,12 +1331,14 @@ open class NSMutableString : NSString {
         }
 
         let start = _storage.utf16.startIndex
-        let min = _storage.utf16.index(start, offsetBy: range.location).samePosition(in: _storage) ??
-            _storage.utf16.index(start, offsetBy: _storage.rangeOfComposedCharacterSequence(at: range.location).location)
+        let min = _storage.utf16.index(start, offsetBy: range.location).samePosition(in: _storage) ?? {
+            let characterRange = _storage.rangeOfComposedCharacterSequence(at: range.location)
+            return _storage.utf16.index(start, offsetBy: characterRange.location)
+        }()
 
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

Foundation/NSString.swift
```diff
@@ -1331,8 +1331,13 @@ open class NSMutableString : NSString {
         }
 
         let start = _storage.utf16.startIndex
-        let min = _storage.utf16.index(start, offsetBy: range.location).samePosition(in: _storage)!
-        let max = _storage.utf16.index(start, offsetBy: range.location + range.length).samePosition(in: _storage)!
+        let min = _storage.utf16.index(start, offsetBy: range.location).samePosition(in: _storage) ??
+            _storage.utf16.index(start, offsetBy: _storage.rangeOfComposedCharacterSequence(at: range.location).location)
+
+        let max = _storage.utf16.index(start, offsetBy: range.location + range.length).samePosition(in: _storage) ?? {
```
itaiferber: Looks reasonable, though from a style perspective it might be nice to make these consistent (either both as a closure or not — I prefer the closure).

In the cases where we might expect to fall into the `rangeOfComposedCharacterSequence(at:)` case for both `min` and `max`, it's likely worth factoring out that computation, I think — there are cases where it's not so cheap.

spevans: Am I correct that the computation for `rangeOfComposedCharacterSequence(at:)` can only be reused if `range.length == 0`. Is that a valid input?

itaiferber: 🤦‍♂️
You're totally right, of course — read too hastily.

spevans:
@swift-ci test

itaiferber:
Looks good!

spevans:
@swift-ci test and merge

PR #1688, opened 2018-09-10
<rdar://problem/43884430> Allow C code to link NSCFConstantString

alblue:
When C code links against Foundation, it can't link to the NSCFConstantString because the symbol isn't exposed by the library. Using this attribute makes it visible, but not usable from Swift.


alblue:
@swift-ci please test

PR #1689, opened 2018-09-12
(Arm32) Fix 32-bit Variable Offset Issues with CoreFoundation

Kaiede:
_CFInfo Assumes a 64-bit _cfinfoa type, when it is really 32-bit on 32-bit systems.
_NSCFType assumes a 32-bit _cfinfoa too, making things even more complicated.

This impacts Swift on arm32v7 platforms like Raspberry Pi 3, and has been causing crashes when using Foundation types that bridge between C and Swift, because shared instance variables are offset by 4 bytes (the extra padding). String being a common example. 

There's a couple approaches to addressing this:
1) Back _CFInfo with a UInt that, on paper, matches size with __CFInfoType
2) Add padding so we can guarantee that _CFInfo can always be 64-bit.

Which one you pick depends on if this is guaranteed to be true: Swift's UInt == CoreFoundation's __CFInfoType

Because Android has some interesting macros in the code that prevented me from answering this question with high confidence, this change picks option 2 which is less risky, even though it does mean an extra 4 bytes for each bridgable object. Option 1 would use less memory on 32-bit systems. 

In either case, it makes sense that __NSCFType should be backed by a _CFInfo instead
of a raw int type so that things remain consistent. Share types when possible so we have consistent, understandable behavior. Question is, does __NSCFType even get used anymore? Is it meant to back something else that always uses a 32-bit _cfinfo? 

In terms of validation, this was tested on Raspbian Stretch, using armhf architecture (armv7 specifically), using an existing project (http://github.com/Kaiede/RPiLight) that was able to produce the original bug against the swift-4.1.2-RELEASE tag doing a 'swift test' on the package. To validate against swift-4.1.2-RELEASE, I also needed to cherry-pick commit 1086b3bb8b859dae10a2a6d6fda4dbbcac2698ae which fixes another 32-bit issue. The swift RC field is 32-bit on arm32, not 64, so that commit is required for 32-bit to work as well.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -227,7 +227,7 @@ extern void __CFGenericValidateType_(CFTypeRef cf, CFTypeID type, const char *fu
 #define __CFBitfield64GetValue(V, N1, N2)	(((V) & __CFBitfield64Mask(N1, N2)) >> (N2))
 #define __CFBitfield64SetValue(V, N1, N2, X)	((V) = ((V) & ~__CFBitfield64Mask(N1, N2)) | ((((uint64_t)X) << (N2)) & __CFBitfield64Mask(N1, N2)))
 
-#if __LP64__ || DEPLOYMENT_TARGET_ANDROID
+#if defined(__LP64__) || defined(__LLP64__)
 typedef uint64_t __CFInfoType;
 #define __CFInfoMask(N1, N2) __CFBitfield64Mask(N1, N2)
 #else
--- a/CoreFoundation/Base.subproj/CFRuntime.h
+++ b/CoreFoundation/Base.subproj/CFRuntime.h
@@ -194,19 +194,18 @@ typedef struct __CFRuntimeBase {
     // This matches the isa and retain count storage in Swift
     uintptr_t _cfisa;
     uintptr_t _swift_rc;
-    // This is for CF's use, and must match _CFInfo layout
+    // This is for CF's use, and must match _CFInfo / _NSCFType layout
 #if defined(__LP64__) || defined(__LLP64__)
     _Atomic(uint64_t) _cfinfoa;
 #else
     _Atomic(uint32_t) _cfinfoa;
-    uint32_t          _padding;
 #endif
 } CFRuntimeBase;
 
 #if defined(__LP64__) || defined(__LLP64__)
 #define INIT_CFRUNTIME_BASE(...) {0, _CF_CONSTANT_OBJECT_STRONG_RC, 0x0000000000000080ULL}
 #else
-#define INIT_CFRUNTIME_BASE(...) {0, _CF_CONSTANT_OBJECT_STRONG_RC, 0x00000080UL, 0}
+#define INIT_CFRUNTIME_BASE(...) {0, _CF_CONSTANT_OBJECT_STRONG_RC, 0x00000080UL}
 #endif
 
 #else
--- a/CoreFoundation/Base.subproj/CFRuntime.h
+++ b/CoreFoundation/Base.subproj/CFRuntime.h
@@ -194,18 +194,19 @@ typedef struct __CFRuntimeBase {
     // This matches the isa and retain count storage in Swift
     uintptr_t _cfisa;
     uintptr_t _swift_rc;
-    // This is for CF's use, and must match _NSCFType layout
+    // This is for CF's use, and must match _CFInfo layout
 #if defined(__LP64__) || defined(__LLP64__)
     _Atomic(uint64_t) _cfinfoa;
 #else
     _Atomic(uint32_t) _cfinfoa;
+    uint32_t          _padding;
 #endif
 } CFRuntimeBase;
 
 #if defined(__LP64__) || defined(__LLP64__)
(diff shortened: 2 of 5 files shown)


Kaiede:
Seems like I should try to page @johnno1962 on the issue with __CFInfoType, as they checked in the change initially that added `DEPLOYMENT_TARGET_ANDROID`.

Kaiede:
Here's Option 1. May need more changes, I haven't validated it yet. Waiting for my local Arm32 swift build to finish so I can test it against a couple real projects I have affected by this bug.

millenomi:
In the meantime, let’s see if it regresses anything on our side:

millenomi:
@swift-ci please test

Kaiede:
So, option 1 doesn't actually seem to work, although it's not clear exactly what piece of code is being naughty. In simple cases, alignment/stride of member variables works and matches between C and Swift, and so bridging is okay. But there are other cases where they don't match which I have yet to find which causes heap corruption. 

I guess my question is, after so long using 8 bytes (4 byte cfinfo + 4 byte pad, and later an 8 byte cfinfo) which kept alignment clean on 8-byte boundaries (16 bytes on 32-bit, 24 bytes on 32-bit), why is it anathema to effectively revert the change that dropped the padding? The history behind the change to use uint32_t is messy:

- @johnno1962 Submits 1086b3b which fixes the Swift refcount block on Apr 18 to be a uintptr_t instead of two uint32_ts. That's this year. This doesn't change the size of the struct for 64 bit, but does for 32-bit. Shrinking it from 24 bytes (20 + 4 padding) to 16 bytes. This is because _cfinfoa was 8-byte aligned, which kept the CFRuntimeBase struct 8-byte aligned as a side-effect. 
- @chnmrc Runs across the bug in 4.1.2-RELEASE (https://bugs.swift.org/browse/SR-6812) in early May. Doesn't see the other fix that's in the master branch, and does his own fix that changes _cfinfoa to a uint32_t, but because he's writing it against 4.1.2-RELEASE, it interacts poorly with @johnno1962's change. Now CFRuntimeBase is 12 bytes, and no longer 8-byte aligned. It also leaves behind other bugs when applied to 4.1.2, because of a mismatch of offset for _cfinfoa between Swift (which has it at offset 8 on 32-bit) and CFRuntimeBase (which has it at offset 12 on 32-bit). Crashes still ensue in some cases. 

Here's really what I want to do: Foundation has effectively had an 8-byte aligned CFRuntimeBase since it's origin in this repository. Because of the interaction of these two things, it is no longer 8-byte aligned, and there's problems because of it. Those problems are difficult to track down. 

While my first commit isn't ideal, it returns Foundation to the status quo that has been working, and has allowed Swift 3 to run so well on these devices. The change to uint32_t wasn't to save memory, and without resources to track down the side-effects of that change, it seems safer to simply revert it and keep CFRuntimeBase 8-byte aligned.

johnno1962:
There is a bit of history behind this problem. In Swift 3, the preamble of a Swift Object was 12 bytes on a 32 bit system (isa + two 32 bit refcounts) and the type of _cfinfoa was char[8]. in Swift4 after the foundation code was synced _cfinfoa became uint64_t which with 8 byte memory alignment added a new 4 bytes padding to the __CFRuntimeBase structure after the ref counts while in Swift4 preamble became 8 bytes (isa plus 32 but unified refcount) just before 4.0 was released. This created a problem because while the __CFRuntimeBase structure increased in size the memory allocated for swift instances decreased so the core foundation code was writing off the end of Swift objects when they were bridged and anarchy ensuing. This was fixed by https://github.com/apple/swift-corelibs-foundation/pull/1525. 

This has interacted badly with another subsequent fix (for the same problem?) https://github.com/apple/swift-corelibs-foundation/commit/a3e72a90c008d64d320dbcb9a58b12e1fb017aff and should be reverted if it is still in master as it results in an ungainly size for __CFRuntimeBase of 12 bytes which may end up padded and didn’t update _CFInfo on the Swift side. I recommend _cfinfoa always be 64 bits.

If you’re looking for a reference point of something that has run the foundation tests and works on 32 bits use [This branch](https://github.com/SwiftJava/swift-corelibs-foundation/commits/android-toolchain-4.1) which contains a single commit relative to the original 4.1 to get things working.

The DEPLOYMENT_TARGET_ANDROID conditional on the type for __CFInfoType is a red herring that is not visible to Swift, was only added to prevent warnings during the C compile and should only affect an Android build.

Kaiede:
Yeah, a revert approach would be cleaner. I’ll just do local testing to be sure. :)

johnno1962:
Sounds good. I’ve just built the 4.2 branch and apart from minor changes the tests are in good shape.

Kaiede:
There's some concern that we need a change to fix a more specific issue being caused by __NSCFType using Int32 instead _CFInfo, a couple of us on swift-arm are still looking at it. We may still need parts of my PR here that align __NSCFType if that's true.

johnno1962:
__NSCFType should probably have a _CFInfo ivar from the look of things.

Kaiede:
> __NSCFType should probably have a _CFInfo ivar from the look of things.

Yeah, that's one of the things this PR does, so I think we should preserve that change with a separate PR.

And boo, I had a branch just ready to pull when you beat me to it. :(

Kaiede:
Closing this one, because of #1693. A new PR will be opened with the __NSCFType part of the change.

PR #1690, opened 2018-09-12
Fix CoreFoundation.h header in standalone CF build

copumpkin:
The SwiftRuntime header has a big warning in it saying not to use it unless you're building with the swift runtime, which this build.py doesn't do.

cc @millenomi 

Also, as a related aside: does this CoreFoundation/build.py get exercised by @swift-ci? I couldn't find any information on how to affect its behavior but I'd love to add a skeleton build to avoid it getting broken in future

### DIFF
--- a/CoreFoundation/build.py
+++ b/CoreFoundation/build.py
@@ -110,7 +110,7 @@ module = 'Base.subproj/module.modulemap',
 public = [
 	'Stream.subproj/CFStream.h',
 	'String.subproj/CFStringEncodingExt.h',
-	'Base.subproj/SwiftRuntime/CoreFoundation.h',
+	'Base.subproj/CoreFoundation.h',
 	'Base.subproj/SwiftRuntime/TargetConditionals.h',
 	'RunLoop.subproj/CFMessagePort.h',
 	'Collections.subproj/CFBinaryHeap.h',


parkera:
I don't think our CI has any way to trigger the standalone build. I'd love if it did, along with building for macOS, actually.

parkera:
@swift-ci test and merge

copumpkin:
@parkera thanks! Is that something I could realistically do? I tried to figure out what it was doing but it looks like a jenkins config that is set up server-side and thus isn't something I could meaningfully affect from within the repo (like in a PR).

shahmishal:
@parkera Did you want to update the current Linux PR job, or add new macOS job?

parkera:
There are a few things. The first is to add a build of the Darwin build of swift-corelibs-foundation to our CI, so we know that it builds correctly when accepting PRs.

copumpkin:
@shahmishal that looks awesome! I haven't quite figured out how those jenkins builds work yet: is that going to test the `CoreFoundation/build.py` or the root `build.py`? Or both? 😁

shahmishal:
it's using build-script from apple/swift repository. 

`swift/utils/build-script -R -T --llbuild --swiftpm --xctest --debug-foundation -- --reconfigure --lit-args=-v` 

Looks like it might be using xcodebuild -- https://github.com/apple/swift/blob/master/utils/build-script-impl#L2515

millenomi:
From my work on merging, it looks like you might want to expect some breakage. I’ll be on standby if that happens.

copumpkin:
@millenomi sorry, was that to me, or to Mishal about the CI stuff? If to me I actually pin to known-good revisions so I'll keep an eye out for it but it shouldn't be anything too pressing 😄

millenomi:
Yeah, sorry, the above was to @shahmishal.

PR #1691, opened 2018-09-13
Add back unnecessarily redacted headers that now exist

copumpkin:
I think this might have been an artifact of the old CFLite redactions? Either way, these headers are now available in this repository. The real CoreFoundation guards these includes with `#ifndef CF_OPEN_SOURCE` but that doesn't seem relevant here. There are also other includes guarded
by that which are not (yet? 😆) available in this repository, so I did not include them here.

cc @millenomi @parkera it's hard for me as an outsider to know the "right" thing to do around `CF_OPEN_SOURCE` avoidance. I see a little bit of it still in this repo, but it's hard for me to know if that's appropriate here (in order to be more similar to the official CF) or if it's just going to be unconditionally available from now on. Any advice if I encounter more stuff like this?

Edit: I'm also unclear on how I should be keeping this CoreFoundaiton.h in sync with the one in SwiftRuntime. Should I make the same change there?

### DIFF
--- a/CoreFoundation/Base.subproj/CoreFoundation.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation.h
@@ -77,6 +77,8 @@
 #include <CoreFoundation/CFSocket.h>
 #include <CoreFoundation/CFMachPort.h>
 
+#include <CoreFoundation/CFAttributedString.h>
+#include <CoreFoundation/CFNotificationCenter.h>
 
 #endif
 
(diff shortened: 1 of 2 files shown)


copumpkin:
I'm honestly not 100% sure this is necessary, because other headers (like CFTimeZone.h for example) do `#include` these headers indirectly, but it seemed good to not rely on the internals of those and to match the official framework more closely wherever possible.

parkera:
I think we need to make the same change in both CF.h headers.

I agree with you that the exclusion is an artifact of before we had these files available in the open source CF.

copumpkin:
Cool, I'll update the other header, thanks!

copumpkin:
@parkera turns out the other header already had `CFAttributedString.h` so I just added the notification center

parkera:
@swift-ci test and merge

copumpkin:
Hmm, I think @swift-ci might be stuck

PR #1703, opened 2018-09-23
Cleanup Conditional Bridge Behavior of NSNumber to Bool

Kaiede:
-- Main Change:
Bring over the implementation of the conditional bridge from the Swift stdlib for NSNumber to Bool, which matches with the initializer of NSNumber to Bool. The reason for this is

It also adds a couple checks to the Foundation Tests to avoid regressing Double/Float to Bool behavior. However, because of SR-8119 (see below), these tests will continue to produce failures. 

-- Side Changes:
I discovered a couple side issues which was causing problems on Darwin while trying to run the tests so I could investigate that have been present for a while in the code. I've included these fixes as well as they are useful. 

There is one remaining test build break on Darwin that I don't have a fix for, and I hacked around it.

-- Further Issues:
Investigating this uncovered another underlying issue where casting NSNumber to Bool conditionally doesn't work on Linux's Foundation, but does on Darwin. I've opened https://bugs.swift.org/browse/SR-8819 to track that particular issue since it can be worked around, and probably needs more thinking to solve correctly than I have time to give. 

This bug also prevents the Foundation tests from passing completely. The XCTAssertNil pieces of testNSNumberToBool will continue to fail until SR-8819 is fixed.

This bug is also responsible for a regression in http://github.com/IBM-Swift/SwiftyJSON/ which was depending a bit on 'as? Bool' returning nil which was true in 4.1.3.


Foundation/NSNumber.swift
```diff
@@ -537,17 +537,20 @@ extension Bool : _ObjectiveCBridgeable {
     }
     
     public static func _forceBridgeFromObjectiveC(_ x: NSNumber, result: inout Bool?) {
-        result = _unconditionallyBridgeFromObjectiveC(x)
+        let _ = _conditionallyBridgeFromObjectiveC(x, result: &result)
```
Kaiede: This isn't needed and wasn't intended to be part of the change, but the swift stdlib does something similar here, so I was testing this. I'll undo this change.

xwu:
Good catch. LGTM except the change that's not needed, which you've already noted.

Foundation/NSNumber.swift
Kaiede: Removed

Kaiede:
I've now added the commit to undo the unnecessary change. The Foundation tests look good on Darwin (except SR-8819), and my own unit tests that discovered this bug are passing again, so I don't have anything else I can think of to test on this.

parkera:
@swift-ci test

Kaiede:
@itaiferber One thing to point out, testNSNumberToBool() should already be failing right now. How is it not?

itaiferber:
@Kaiede I'm not sure, to be honest — it might be related to architecture/OS differences between the version of Swift you tested on and what we've got in CI. It's worth seeing if this test run passes or not; if it does then we have some signal that SR-8819 affects your current config, but not what we have in CI (for better or worse).

Kaiede:
@itaiferber I guess I answered my own question, the Linux CI is passing testNSNumberToBool() properly, even with my changes. So I guess SR-8819 does only affect building Foundation on Darwin. I'll go ahead and add that information to the bug then, since the limited scope is probably good to know about (and reduces any priority on fixing it, I bet).

itaiferber:
@Kaiede Ah, I had misread your initial report — I mistakenly assumed that this was happening _on_ Linux (as I had read `Linux 4.2 (Originally Discovered on aarch64)` and missed the explicit callout to Darwin); indeed, this is less surprising, then.

Kaiede:
I'll go ahead and cherry-pick the Darwin build fixes into their own PR. They were mostly required so that I could at least use the unit tests to debug/validate the change on Darwin. I haven't setup an x86_64 Linux VM yet (my focus has been on armv7 and aarch64). Probably should.

Kaiede:
I've pulled them out, squashed the partial revert, and updated the primary comment to reflect the true state of the fix.

itaiferber:
@swift-ci Please test

millenomi:

millenomi:

PR #1704, opened 2018-09-24
Fix DateFormatter TimeZone Setter

Kaiede:
Because of the use of 'timeZone' instead of 'newValue' in the setter, setting the timeZone of a DateFormatter doesn't actually work. Instead, it winds up re-setting the old value, which happens to be whatever was originally set, usually what we get from copying the default down in CoreFoundation.

A good fix here is to avoid any ambiguity with the property name in the setter and getter.

Also adds a unit test to help prevent breaks like this again.

### DIFF
--- a/TestFoundation/TestDateFormatter.swift
+++ b/TestFoundation/TestDateFormatter.swift
@@ -24,6 +24,7 @@ class TestDateFormatter: XCTestCase {
             ("test_dateFormatString", test_dateFormatString),
             ("test_setLocaleToNil", test_setLocaleToNil),
             ("test_setTimeZoneToNil", test_setTimeZoneToNil),
+            ("test_setTimeZone", test_setTimeZone),
         ]
     }
     
@@ -355,5 +356,23 @@ class TestDateFormatter: XCTestCase {
         f.timeZone = nil
         // Time zone should go back to the system one.
         XCTAssertEqual(f.timeZone, NSTimeZone.system)
+    }
+
+    func test_setTimeZone() {
+        // Test two different time zones. Should ensure that if one
+        // happens to be TimeZone.current, we still get a valid test.
+        let newYork = TimeZone(identifier: "America/New_York")!
+        let losAngeles = TimeZone(identifier: "America/Los_Angeles")!
+
+        XCTAssertNotEqual(newYork, losAngeles)
+
+        // Case 1: New York
+        let f = DateFormatter()
+        f.timeZone = newYork
+        XCTAssertEqual(f.timeZone, newYork)
+
+        // Case 2: Los Angeles
+        f.timeZone = losAngeles
+        XCTAssertEqual(f.timeZone, losAngeles)
     }
 }
--- a/TestFoundation/TestDateFormatter.swift
+++ b/TestFoundation/TestDateFormatter.swift
@@ -25,6 +25,7 @@ class TestDateFormatter: XCTestCase {
             ("test_setLocaleToNil", test_setLocaleToNil),
             ("test_setTimeZoneToNil", test_setTimeZoneToNil),
             ("test_setTimeZone", test_setTimeZone),
+            ("test_ExpectedTimeZone", test_ExpectedTimeZone),
         ]
     }
     
@@ -374,5 +375,37 @@ class TestDateFormatter: XCTestCase {
         // Case 2: Los Angeles
         f.timeZone = losAngeles
         XCTAssertEqual(f.timeZone, losAngeles)
+    }
+
+    func test_ExpectedTimeZone() {
+        let gmt = TimeZone(abbreviation: DEFAULT_TIMEZONE)
+        let newYork = TimeZone(identifier: "America/New_York")!
+        let losAngeles = TimeZone(identifier: "America/Los_Angeles")!
+
+        XCTAssertNotEqual(newYork, losAngeles)
+
+        let now = Date()
+
+        let f = DateFormatter()
+        f.dateFormat = "z"
+
+        // Case 1: TimeZone.current
+        f.timeZone = TimeZone.current
+        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+
+        // Case 2: New York
+        f.timeZone = newYork
+        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+
+        // Case 3: Los Angeles
+        f.timeZone = losAngeles
+        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+
+        guard gmt != TimeZone.current else {
+            print("Inconclusive: This test checks to see if the formatter produces the same TZ as TimeZone.current")
+            print("When it fails, TimeZone.current formats as GMT instead of normal.")
+            print("Unfortunately, we can't use GMT as TimeZone.current for this test to be conclusive.")
+            return
+        }
     }
 }
--- a/TestFoundation/TestDateFormatter.swift
+++ b/TestFoundation/TestDateFormatter.swift
@@ -25,7 +25,7 @@ class TestDateFormatter: XCTestCase {
             ("test_setLocaleToNil", test_setLocaleToNil),
             ("test_setTimeZoneToNil", test_setTimeZoneToNil),
             ("test_setTimeZone", test_setTimeZone),
-            ("test_ExpectedTimeZone", test_ExpectedTimeZone),
+            ("test_expectedTimeZone", test_expectedTimeZone),
         ]
     }
     
@@ -377,7 +377,7 @@ class TestDateFormatter: XCTestCase {
         XCTAssertEqual(f.timeZone, losAngeles)
     }
 
-    func test_ExpectedTimeZone() {
+    func test_expectedTimeZone() {
         let gmt = TimeZone(abbreviation: DEFAULT_TIMEZONE)
         let newYork = TimeZone(identifier: "America/New_York")!
         let losAngeles = TimeZone(identifier: "America/Los_Angeles")!
@@ -388,24 +388,24 @@ class TestDateFormatter: XCTestCase {
 
         let f = DateFormatter()
         f.dateFormat = "z"
+        f.locale = Locale(identifier: "en_US_POSIX")
 
         // Case 1: TimeZone.current
+        // This case can catch some issues that cause TimeZone.current to be
+        // treated like GMT, but it doesn't work if TimeZone.current is GMT.
+        // If you do find an issue like this caused by this first case,
+        // it would benefit from a more specific test that fails when
+        // TimeZone.current is GMT as well.
+        // (ex. TestTimeZone.test_systemTimeZoneName)
         f.timeZone = TimeZone.current
-        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+        XCTAssertEqual(f.string(from: now), TimeZone.current.abbreviation())
 
         // Case 2: New York
         f.timeZone = newYork
-        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+        XCTAssertEqual(f.string(from: now), newYork.abbreviation())
 
         // Case 3: Los Angeles
         f.timeZone = losAngeles
-        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
-
-        guard gmt != TimeZone.current else {
-            print("Inconclusive: This test checks to see if the formatter produces the same TZ as TimeZone.current")
-            print("When it fails, TimeZone.current formats as GMT instead of normal.")
-            print("Unfortunately, we can't use GMT as TimeZone.current for this test to be conclusive.")
-            return
-        }
+        XCTAssertEqual(f.string(from: now), losAngeles.abbreviation())
     }
 }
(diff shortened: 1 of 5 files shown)


Kaiede:
I found this by accident trying to debug/find a similar break in 4.1 (which is where we are stuck on arm32 at the moment due to 32-bit alignment problems in the compiler with 4.2). Still trying to find that one, sadly.

spevans:
@swift-ci test

Kaiede:
What's the general rules for bundling fixes? I found the bug in 4.1/4.2 I was looking for. It's related enough I could just add the commit to this PR, but it's also somewhat unrelated (TimeZone.current has effectively been "GMT" for at least a year, no matter what it actually reports).

Kaiede:
I've added caa8d0a to the PR, since that's also required to unblock my scenario of using the system timezone to parse/format date strings (think logging, and interpreting time-of-day schedules without requiring a TZ in the string).

TestFoundation/TestDateFormatter.swift
```diff
@@ -375,4 +376,36 @@ class TestDateFormatter: XCTestCase {
...
+
+        // Case 2: New York
+        f.timeZone = newYork
+        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+
+        // Case 3: Los Angeles
+        f.timeZone = losAngeles
+        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+
+        guard gmt != TimeZone.current else {
```
Kaiede: I'm not 100% happy with the way this test works, which pretty much is trying to make sure that we can use "TimeZone.current" for zones other than GMT and get what we expect when we then run it through DateFormatter.

If this is too much, I can try to write a more targeted test, although I'm not entirely sure that would be any less fragile.

TestFoundation/TestDateFormatter.swift
```diff
@@ -356,4 +358,54 @@ class TestDateFormatter: XCTestCase {
...
+        let f = DateFormatter()
+        f.dateFormat = "z"
+
+        // Case 1: TimeZone.current
+        f.timeZone = TimeZone.current
+        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
+
+        // Case 2: New York
+        f.timeZone = newYork
+        XCTAssertEqual(f.string(from: now), f.timeZone.abbreviation())
```
maksimorlovich: Would it make sense to us f.string(from: now), newYork.abbreviation())? Ie, if f.timeZone does not actually change the timezone, this check will succeed (with off-by-1 fix).

I understand that's not the purpose of this test case, but it is to catch that off-by-one error in the name, but I don't see a disadvantage in using the actual timezone variable instead of f.timeZone. Thoughts?

Kaiede: That's fair. I generally agree that tests should avoid behavior like this. This code was mostly easier to copy/paste 3 times without errors as written, to be honest, and I forgot to clean it up after confirming the test failed without the fix and passed with it.

maksimorlovich:
LGTM! Thanks @Kaiede!

parkera:
Once this is in master, a PR with a cherry-pick to 4.2 will get it into the next version of that. Generally I approve those if they are manageable bug fixes that unblock clients or address serious issues. Other stuff can target the next major release (5.0).

TestFoundation/TestDateFormatter.swift
```diff
@@ -356,4 +358,54 @@ class TestDateFormatter: XCTestCase {
...
+        let f = DateFormatter()
+        f.timeZone = newYork
+        XCTAssertEqual(f.timeZone, newYork)
+
+        // Case 2: Los Angeles
+        f.timeZone = losAngeles
+        XCTAssertEqual(f.timeZone, losAngeles)
+    }
+
+    func test_ExpectedTimeZone() {
```
itaiferber: Tiny non-blocking nit: `s/Expected/expected/`

TestFoundation/TestDateFormatter.swift
```diff
@@ -356,4 +358,54 @@ class TestDateFormatter: XCTestCase {
...
+        let gmt = TimeZone(abbreviation: DEFAULT_TIMEZONE)
+        let newYork = TimeZone(identifier: "America/New_York")!
+        let losAngeles = TimeZone(identifier: "America/Los_Angeles")!
+
+        XCTAssertNotEqual(newYork, losAngeles)
+
+        let now = Date()
+
+        let f = DateFormatter()
```
itaiferber: `DateFormatter` defaults to `Locale.current` as its default locale. This is unlikely to affect CI too too much, but is a potential source of confusing test failures when running tests on a machine whose locale is not `en_US` or similar.

Explicitly setting `f.locale = Locale(identifier: "en_US_POSIX")` is a good way to make sure this test is 100% deterministic.

TestFoundation/TestDateFormatter.swift
```diff
@@ -356,4 +358,54 @@ class TestDateFormatter: XCTestCase {
...
+
+        // Case 2: New York
+        f.timeZone = newYork
+        XCTAssertEqual(f.string(from: now), newYork.abbreviation())
+
+        // Case 3: Los Angeles
+        f.timeZone = losAngeles
+        XCTAssertEqual(f.string(from: now), losAngeles.abbreviation())
+
+        guard gmt != TimeZone.current else {
```
itaiferber: Should there be another test case after this guard for GMT? (If not, I don't see much of a reason to keep the print statements around.)

Kaiede: I really want an XCTInconclusive(). The issue is that because TimeZone.current gets treated as GMT when it isn’t, if you happen to be using GMT, this test can’t tell you conclusively that the bug doesn’t exist.

Kaiede: So I guess my question is, should this test just pass in this case without the “warnings”? Should a more specific test be found checking TimeZone.current’s identifier string so this test doesn’t need to test for the fallout of that case?

It’s kinda why I’m not a huge fan of how I wrote this, and was hoping for pointers on what would be more acceptable.

Kaiede: Assume TimeZone.current is "America/Los_Angeles"

You can ask it's identifier, and it's abbreviation and it will give you what you expect. However, because of this:

```
CFStringCreateWithBytes(kCFAllocatorSystemDefault, (uint8_t *)linkbuf + zoneInfoDirLen,
                                            strlen(linkbuf) - zoneInfoDirLen + 1, kCFStringEncodingUTF8, false);
```

Which allocates one byte too many, and includes the null terminator in the CFString, if I then give DateFormatter this malformed TimeZone, I get GMT, despite it being PDT/PST. It also affects the intepretation of strings in both directions. Instead of getting say, "Midnight, PST", I get "7-8am, GMT". 

i.e. If you have a malformed TimeZone, DateFormatter assumes that it is GMT, not your actual time zone. So if you are trying to format strings that implicitly assume local time, "08:00" for example, then you get bunk results from DateFormatter that throw your time off by your GMT offset.

Kaiede: Either way, I'm going to try to write a more focused test that checks for the bad length, which means the warning can become a comment on case 1 as information for folks down the road wondering why non-GMT time zones are suddenly being treated as GMT by DateFormatter for no apparent reason.

Kaiede: There's one more subtle wrinkle here:

`TimeZone(identifier: "America/Los_Angeles")`: Works
`TimeZone(identifier: TimeZone.current.identifier)`: Works
`TimeZone.current`: Broken

The extra character was only being appended in CFCreateSystemTimeZone(), and so it doesn't affect you if you can declare the time zone directly.

Does ICU take in the string length? That would explain it if it relies on the buffer length and not a terminator like CFString.

itaiferber: ICU _can_ search for a NUL-terminating character if you tell it to, but `CFStringRef` does not require NUL termination; we pass the contents of the `CFStringRef` along with its length directly to ICU, so if the `CFStringRef` contains a NUL character, ICU will get that. It appears that ICU reads this NUL character in a way that makes `"America/Los_Angeles\0" != "America/Los_Angeles"`.

Reproduces easily with

```objc
#import <Foundation/Foundation.h>
#include <unicode/udat.h>

int main()
{
    @autoreleasepool {
#define U_CHECK(status, message) if (U_FAILURE(status)) { NSLog(message @": %s", u_errorName(status)); exit(EXIT_FAILURE); }
        UErrorCode status = U_ZERO_ERROR;
        UDateFormat *formatter = udat_open(UDAT_PATTERN, UDAT_PATTERN, "en_US_POSIX", u"America/Los_Angeles\0", 20, u"z", -1, &status);
        U_CHECK(status, @"Unable to create date formatter");
        
        UChar result[BUFSIZ] = {0};
        int32_t length = udat_format(formatter, ucal_getNow(), result, BUFSIZ, NULL, &status);
        U_CHECK(status, @"Unable to format");
        
        NSLog(@"%@", [[NSString alloc] initWithCharactersNoCopy:result length:length freeWhenDone:NO]);
    }
}
```

producing `GMT`. Changing the length from `20` to `19` produces `PDT` as expected.

That being said: are you saying that `TimeZone.current` is _still_ broken after this change, or was broken only before the change? I'm trying to figure out what the state of things are with this fix.

Kaiede: With this PR, it is no longer broken. 

The point of this test was to try to prevent a similar regression that breaks DateFormatter in a similar way.

itaiferber: This seems reasonable for now — I think in the future we can just add a test to see if `TimeZone.current.identifier != "GMT` but `formatter.string(from: date) == "GMT"` to catch this too, but we shouldn't need to hold up this PR any further over that.

itaiferber:
@swift-ci Please test

millenomi:

Kaiede:
@millenomi My bet is on newly uncovered. This was not something I was seeing fail with my changes locally, *but*, my sync point is now 16 days old, and the last PR push was 8 days old. A lot of stuff recently got merged here, and it's possible that for some reason I just wasn't paying close enough attention.

In general, the calendar test failing is interesting because it is relying on the current time zone to function properly, in addition to the current Calendar. The bug that I'm fixing is that TimeZone.current is always acting like GMT when using ICU/DateFormatter. So it kinda makes sense that my "fix" could make this test appear to fail, especially depending on the time zone the machine running the tests happens to be using. 

Instead, the failing test should be using Calendar current with a fixed UTC timezone to avoid this sort of problem. It's testing too many variables at once in practice.

Kaiede:
Yeah, my suspicions were right. If the test runs in GMT, it passes, if it runs in PDT, it fails with the exact failures the CI also hit.

I also realize I didn't see it before because there's another bug when running these tests on macOS using the xcworkspace where CoreFoundation can't find the current time zone and also defaults to GMT. I had hacked around it to run my own tests, but I didn't run the whole suite that way when I discovered that problem. 

Is the xcworkspace not really used anymore? It seems to be getting progressively broken. I think I ran into a few new breaks attempting to pull the latest set of changes from master, and couldn't even get SwiftFoundation to link on Mojave anymore.

millenomi:
I ran tests with the Mojave merge on Mac, so I'm surprised. @shahmishal was looking into possibly adding CI to run those tests.

Kaiede:
@millenomi I just open the xcworkspace from the Terminal and attempt to run/build from there. So I'm wondering what I'm doing differently. Either way, the fix for the test has been added to the PR.

millenomi:
@Kaiede Swift Foundation is revlocked to the Swift toolchain it ships with — so you _must_ either build your own via `$SOURCES/swift/utils/build-toolchain`, or, much better, download the latest one from https://swift.org/download/ for your branch (in this case, the one under 'Trunk Development (master)').

I have just built and ran tests for #1600 successfully, though I am seeing the test failures you describe with the off-by-oneness. But I can build and run correctly.

millenomi:
(Switch toolchains from Xcode > Preferences > Components > Toolchains.)

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

Kaiede:
> @Kaiede Swift Foundation is revlocked to the Swift toolchain it ships with — so you _must_ either build your own via `$SOURCES/swift/utils/build-toolchain`, or, much better, download the latest one from https://swift.org/download/ for your branch (in this case, the one under 'Trunk Development (master)').

Okay, that's what I'm doing wrong. Thanks.

millenomi:
We really ought to have that written down somewhere easy to find.

gparker42:
This broke CI. TestDateFormatter.swift's test_expectedTimeZone is failing. Reverting in #1721.

```
TestFoundation/TestDateFormatter.swift:401: error: TestDateFormatter.test_expectedTimeZone : XCTAssertEqual failed: ("Optional("GMT")") is not equal to ("Optional("UTC")")
```

There is a suspicious build warning in that test:
```
TestFoundation/TestDateFormatter.swift:381:13: warning: initialization of immutable value 'gmt' was never used; consider replacing with assignment to '_' or removing it
        let gmt = TimeZone(abbreviation: DEFAULT_TIMEZONE)
```

gparker42:
Looks like the build czar's force-merge privileges don't apply to swift-corelibs-foundation, so you have a bit of time to fix it before then.

millenomi:
Per side discussion: let's revert, then we will put the PR back up again.

Kaiede:
> This broke CI. TestDateFormatter.swift's test_expectedTimeZone is failing. Reverting in #1721.

Sorry about that, looks like the pre-checkin CI just happened to be on a PT time zone, which doesn't expose this issue with the test. Same with my local test environment.

> There is a suspicious build warning in that test:
> 
> ```
> TestFoundation/TestDateFormatter.swift:381:13: warning: initialization of immutable value 'gmt' was never used; consider replacing with assignment to '_' or removing it
>         let gmt = TimeZone(abbreviation: DEFAULT_TIMEZONE)
> ```

Ironically, it looks suspicious, but is entirely unrelated code that was part of an "inconclusive test" message that was removed based on feedback, but not the let statement that helped feed the guard. 

The issue with the break is that if TimeZone.current.abbreviation gives you "GMT", but then ask a DateFormatter using that TimeZone what the abbreviated zone is, it gives you the string "UTC". Neat. :(

PR #1705, opened 2018-09-24
Fix a Couple Darwin Build Breaks on Xcode 10

Kaiede:
A couple build break fixes for 10.14 + Xcode 10. Doesn't get Darwin completely building at master, but really helps clear out the noise. 

- NSString(_:) is internal, use NSString(string:) instead. Probably a better test as well.
- Clean up the cast for NSError.


spevans:
@swift-ci test

millenomi:

PR #1707, opened 2018-09-26
Remove two-digit year cookie format support & unit test (fails on Ubu…

maksimorlovich:
…ntu 14.04)

### DIFF
--- a/Foundation/HTTPCookie.swift
+++ b/Foundation/HTTPCookie.swift
@@ -104,17 +104,8 @@ open class HTTPCookie : NSObject {
         return formatter
     }()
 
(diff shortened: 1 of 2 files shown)


jrose-apple:
@swift-ci Please test

jrose-apple:
Unfortunately I don't think we have any way to test 14.04 from CI right now. @shahmishal?

shahmishal:
Correct, we don't currently support PR testing on 14.04.

jrose-apple:
I'm going to merge this ahead of Foundation reviewers to get the bots back, but there's probably a way to write this test so that it can dynamically check for a new enough ICU. That's outside my area, though, so for now let's take this.

PR #1396, opened 2018-01-17
Added missing __checkint macros for 32bit systems

uraimo:
These two defines, similar to the int64 variants already present, are required on 32bit systems (e.g. linux-arm).

PR #1708, opened 2018-09-26
The Mojave Core Foundation merge.

millenomi:
Welcome to macOS Mojave and iOS 12. This patch merges improvements from the Core Foundation version found in these operating systems into Swift Foundation.

For more information on changes to clients of Core Foundation on these operating systems, see [the Foundation release notes](https://developer.apple.com/documentation/ios_release_notes/ios_12_release_notes/foundation_release_notes). In particular, macros for closed enumerations and CFMessagePort improvements apply to this codebase as well.

The changes in this merge mostly revolve around performance improvements. For example, type IDs are now constants and we can lay out constant objects' memory directly into CF instead of requiring `dispatch_once`s and allocations in `__CFInitialize`. We also audited which symbols are exported, picked more specific kinds of locks to improve performance in some situations, and improved the correctness of some code (e.g. `CFURLComponents` parsing; ensuring subclasses of `NSData` that override `copy` interact correctly with `CFDataCreateCopy`). Finally, we bring headers to closer parity with the macOS and iOS ones.


millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
If you're working on a port of Swift, or on ABI stability, I'd highly recommend you read the patch. In particular, @compnerd, @johnno1962, @kaiede, @tinysun212 (for the Cygwin parts), @mikeash, would you like to take a look at this?

CoreFoundation/Locale.subproj/CFLocale.c
```diff
@@ -1429,20 +1478,16 @@ static bool __CFLocaleCopyCollatorID(CFLocaleRef locale, bool user, CFTypeRef *c
...
+STATIC_CONST_STRING_DECL(_metricUnitsKey, "AppleMetricUnits");
+STATIC_CONST_STRING_DECL(_measurementUnitsKey, "AppleMeasurementUnits");
+STATIC_CONST_STRING_DECL(_measurementUnitsCentimeters, "Centimeters");
+STATIC_CONST_STRING_DECL(_measurementUnitsInches, "Inches");
+STATIC_CONST_STRING_DECL(_temperatureUnitKey, "AppleTemperatureUnit");
 
 static bool __CFLocaleGetMeasurementSystemForPreferences(CFTypeRef metricPref, CFTypeRef measurementPref, UMeasurementSystem *outMeasurementSystem) {
     if (metricPref || measurementPref) {
         if (metricPref == kCFBooleanTrue && measurementPref && CFEqual(measurementPref, _measurementUnitsInches)) {
-#if U_ICU_VERSION_MAJOR_NUM >= 55
```
spevans: Will removing this check break Ubuntu14 builds which I think only has version 52?

millenomi:
@swift-ci please test

CoreFoundation/Locale.subproj/CFLocale.c
millenomi: Good catch.

millenomi:
@swift-ci please test

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -564,13 +572,13 @@ CF_EXPORT void _CFGetUGIDs(uid_t *euid, gid_t *egid) {
     __CFGetUGIDs(euid, egid);
 }
     
-CF_EXPORT uid_t _CFGetEUID(void) {
+CF_EXPORT uid_t _CFGetEUID() {
```
parkera: Should we have lost the void here?

CoreFoundation/Base.subproj/DarwinSymbolAliases
```diff
@@ -1,7 +1,7 @@
 
 
 # This is specific to Swift open source; Foundation's NSCFConstantString is used as our constant string class reference
-__TMC15SwiftFoundation19_NSCFConstantString ___CFConstantStringClassReference
+_$s15SwiftFoundation19_NSCFConstantStringCN ___CFConstantStringClassReference
```
parkera: Just a note here to double check this vs recent discussions about the mangling changing.

CoreFoundation/Collections.subproj/CFData.c
```diff
@@ -322,19 +324,16 @@ static const CFRuntimeClass __CFDataClass = {
...
 
 void _CFDataInit(CFMutableDataRef memory, CFOptionFlags variety, CFIndex capacity, const uint8_t *bytes, CFIndex length, Boolean noCopy) {
     Boolean isMutable = ((variety & __kCFMutableMask) != 0);
     Boolean isGrowable = ((variety & __kCFGrowableMask) != 0);
-    Boolean isDontDeallocate = ((variety & __kCFDontDeallocate) != 0);
     
     __CFDataSetNumBytesUsed(memory, 0);
     __CFDataSetLength(memory, 0);
-    __CFDataSetDontDeallocate(memory, isDontDeallocate);
+    __CFDataSetDontDeallocate(memory, true);
```
parkera: I think this is reverting an earlier change

CoreFoundation/Collections.subproj/CFData.c
```diff
@@ -460,7 +460,7 @@ CFDataRef CFDataCreateCopy(CFAllocatorRef allocator, CFDataRef data) {
     Boolean allowRetain = true;
     if (allowRetain) {
         CF_OBJC_FUNCDISPATCHV(CFDataGetTypeID(), CFDataRef, (NSData *)data, copy);
-        CF_SWIFT_FUNCDISPATCHV(CFDataGetTypeID(), CFDataRef, (CFSwiftRef)data, NSObject.copyWithZone, NULL);
+        CF_SWIFT_FUNCDISPATCHV(CFDataGetTypeID(), CFDataRef, (CFSwiftRef)data, NSData.copy);
```
parkera: I think this may also be reverting an earlier change. Not sure.

CoreFoundation/Collections.subproj/CFStorage.c
```diff
@@ -242,7 +245,7 @@ static inline void __CFStorageSetChild(CFStorageNode *parentNode, CFIndex childI
     *((void **)&parentNode->info.notLeaf.child[childIndex]) = newChild;
 }
 
-static inline void __CFStorageGetChildren(const CFStorageNode *parent, CFStorageNode ** _CF_RESTRICT resultArray, bool shouldRetain, bool shouldFreeze) {
+static inline void __CFStorageGetChildren(const CFStorageNode *parent, CFStorageNode ** restrict resultArray, bool shouldRetain, bool shouldFreeze) {
```
parkera: This was recently added for C++ support so we need to use the macro.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -696,50 +701,6 @@ static void __InitTZStrings(void) {
     }
     __CFUnlock(&__CFTZDirLock);
 }
-
-#elif DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED
```
parkera: We sure it's correct to remove this?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -813,15 +769,6 @@ static CFTimeZoneRef __CFTimeZoneCreateSystem(void) {
         CFRelease(name);
         if (result) return result;
     }
-#if DEPLOYMENT_TARGET_ANDROID
```
parkera: I suspect most of the changes in `CFTimeZone.c` should probably prefer the SCL-F version.

CoreFoundation/Base.subproj/CFUtilities.c
millenomi: Nope. To fix.

CoreFoundation/Base.subproj/DarwinSymbolAliases
millenomi: Yup; it's using `$s`, which is the new prefix. (The other class names in the source have been updated accordingly as well.)

CoreFoundation/Collections.subproj/CFData.c
millenomi: Yup. To fix.

CoreFoundation/Collections.subproj/CFData.c
millenomi: There was at least one test failing because a `NSData` subclass overrode `copy()` instead of `copy(with:)`.

CoreFoundation/Collections.subproj/CFStorage.c
millenomi: I will make sure all mentions remain in the codebase.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
millenomi: Tests pass, and it matches the original.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
millenomi: Yeah. This bit absolutely needs to be preserved.

CoreFoundation/Base.subproj/CFSortFunctions.c
```diff
@@ -52,8 +52,6 @@ enum {
 
 #define __checkint_int64_mul(x, y, err)         (x * y)
 #define __checkint_uint64_add(x, y, err)        (x + y)
-#define __checkint_int32_mul(x,y,err)           (x * y)
```
uraimo: Hi, while this is not required on 64bits platforms it should still be on arm32 and x86 (committed by me with #1396).

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -279,6 +277,24 @@ static inline void __CFRuntimeSetFlag(CFTypeRef cf, uint8_t n, Boolean flag) {
...
+#define _CFRUNTIME_BASE_INIT_SWIFT_RETAIN_COUNT ._swift_rc = _CF_CONSTANT_OBJECT_STRONG_RC
+#else
+#define _CFRUNTIME_BASE_INIT_SWIFT_RETAIN_COUNT
+#endif
+
+// A note on these macros.
+// On systems where we have ObjC support (DEPLOYMENT_RUNTIME_OBJC), STATIC_CLASS_REF(…) can statically produce a reference to the ObjC class symbol that ties into this particular type.
+// When compiling for Swift Foundation, STATIC_CLASS_REF returns a Swift class. There's a mapping of ObjC name classes to Swift symbols in the header that defines it that should be kept up to date if more constant objects are defined.
+// On all other platforms, it returns NULL, which is okay; we only need the type ID if CF is to be used by itself.
+#if __LP64__
```
Kaiede: This set of macros is a little weird for a couple reasons:

It is using a ULL in the 32-bit portion.

It assumes that Darwin and non-Darwin both use effectively “unsigned” as the cfinfoa type. Probably not fatal, but confusing to read.

Also curious why this was added here rather than where the other initialization macros are, and why do we need another set?

Out of curiosity, does anyone know if _CFInfo is only used on Linux (non-ObjC) targets? Seems like it is?

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -190,11 +188,11 @@ CF_PRIVATE CFIndex __CFActiveProcessorCount(void);
...
-    #define CFAssert1(cond, prio, desc, a1) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1); HALT; } } while (0)
-    #define CFAssert2(cond, prio, desc, a1, a2) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1, a2); HALT; } } while (0)
-    #define CFAssert3(cond, prio, desc, a1, a2, a3) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1, a2, a3); HALT; } } while (0)
-    #define CFAssert4(cond, prio, desc, a1, a2, a3, a4) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1, a2, a3, a4); HALT; } } while (0)
+    #define CFAssert(cond, prio, desc) do { if (!(cond)) { CFLog(prio, CFSTR(desc)); /* HALT; */ } } while (0)
+    #define CFAssert1(cond, prio, desc, a1) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1); /* HALT; */ } } while (0)
+    #define CFAssert2(cond, prio, desc, a1, a2) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1, a2); /* HALT; */ } } while (0)
+    #define CFAssert3(cond, prio, desc, a1, a2, a3) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1, a2, a3); /* HALT; */ } } while (0)
+    #define CFAssert4(cond, prio, desc, a1, a2, a3, a4) do { if (!(cond)) { CFLog(prio, CFSTR(desc), a1, a2, a3, a4); /* HALT; */ } } while (0)
```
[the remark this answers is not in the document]
millenomi: I'll research, and I'll uncomment if the reason doesn't apply to SCF.

CoreFoundation/Base.subproj/CFInternal.h
millenomi: `_cfinfoa` is used on all targets. Its bits correspond to the ObjC object flag word semantics (please, runtime people, correct me if I'm wrong?), and it is used to store the refcount in ObjC and standalone (non-Swift) builds.

CoreFoundation/Base.subproj/CFSortFunctions.c
millenomi: I'll fix it!

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1374,15 +1375,15 @@ CF_EXPORT char **_CFEnviron(void) {
...
     return open(path, opts, mode);
 }
 int _CFOpenFile(const char *path, int opts) {
     return open(path, opts);
 }
 
-void *_CFReallocf(void *ptr, size_t size) {
-#if DEPLOYMENT_TARGET_WINDOWS | DEPLOYMENT_TARGET_LINUX
+CF_CROSS_PLATFORM_EXPORT void *_CFReallocf(void *ptr, size_t size) {
+#if TARGET_OS_WIN32 | DEPLOYMENT_TARGET_LINUX
```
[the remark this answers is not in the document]
millenomi: Good catch.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1354,10 +1356,9 @@ CF_CROSS_PLATFORM_EXPORT int _CFThreadSetName(pthread_t thread, const char *_Non
 #endif
 }
 
-CF_CROSS_PLATFORM_EXPORT int _CFThreadGetName(char *_Nonnull buf, int length) {
+CF_CROSS_PLATFORM_EXPORT int _CFThreadGetName(char *buf, int length) {
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_EMBEDDED_MINI
     return pthread_getname_np(pthread_self(), buf, length);
-#elif DEPLOYMENT_TARGET_ANDROID
```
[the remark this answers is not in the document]
millenomi: Not really. It would've just dropped into the `return -1;` path, and the code is equivalent.

millenomi:
@swift-ci please test.

millenomi:
@swift-ci please test macOS

millenomi:
@shahmishal (is swift-ci working? it has responded only once to commands on this patch).

shahmishal:
@swift-ci please test

CoreFoundation/Base.subproj/CFInternal.h
Kaiede: Okay, so these macros add more parameters that `INIT_CFRUNTIME_BASE` in `CFRuntime.h` doesn't. That's fine. It's just a little hard to compare the two sets of macros that do incredibly similar things because they are in different files. But since this is for internal, I guess that's the point.

My comment about _cfinfoa is more that on 32-bit, you've got a `ULL` (64-bit) instead of a `UL` (32-bit) const. 

As for _CFInfo, I bring it up because it's padded to 64-bits, at least on Linux, which makes macros like this... interesting.

millenomi:
This will be merged Monday morning, if nothing else pops up in the meanwhile.

millenomi:
I'm going to ask swift-ci to test and merge around 10.15 PST today.

millenomi:

millenomi:

millenomi:
Seems like an upstream issue.

millenomi:

millenomi:
@swift-ci test and merge

millenomi:
https://bugs.swift.org/browse/SR-8899 has been filed to track the blocker.

millenomi:
- Uses `bool` instead of `_Bool` to match the API surface in Core Foundation

- You may notice we are moving some `DEPLOYMENT_TARGET_*` flags to `TARGET_OS_*`. In general, we will strongly prefer `TARGET_OS_*` going forward wherever possible. (`DEPLOYMENT_RUNTIME_SWIFT` and `DEPLOYMENT_RUNTIME_OBJC` remain.)

- Changing some `extern function(void);` was premature for source compatibility reasons.

millenomi:
@swift-ci please test

millenomi:
(on the off chance these changes actually happen to not trigger the blocker anymore.)

millenomi:
@swift-ci please test

millenomi:
Turns out: no.

millenomi:

Foundation/NSData.swift
```diff
@@ -1138,3 +1138,7 @@ extension NSData : _StructTypeBridgeable {
         return Data._unconditionallyBridgeFromObjectiveC(self)
     }
 }
+
+internal func _CFSwiftDataCreateCopy(_ data: AnyObject) -> Unmanaged<AnyObject> {
+    return Unmanaged<AnyObject>.passRetained((data as! NSDictionary).copy() as! NSObject)
```
drodriguez: I don’t think `as! NSDictionary` is correct. Shouldn’t it be `as! NSData`?

millenomi: A copy-and-paste error. I postponed adding tests due to the size, but this may merit one. Thanks!

CoreFoundation/Locale.subproj/CFDateFormatter.c
```diff
@@ -1517,7 +1524,7 @@ static UDate __CFDateFormatterCorrectTimeToARangeAroundCurrentDate(UCalendar *ca
         }
     } else {
         if (period < INT_MAX && futureMax > period) {
-            futureRange.location = 1;
+            futureRange.location = 1,
```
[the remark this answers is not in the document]
millenomi: Oh. This must've reverted a fix inadvertently.

millenomi:
@swift-ci please test

CoreFoundation/Locale.subproj/CFDateFormatter.c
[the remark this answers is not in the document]
mikeash: Just for the record, in case it's somehow relevant in the future, the comma operator has a lower precedence than the assignment operator so this code does work as expected with the comma, although it's definitely weird. It's equivalent to `(futureRange.location = 1), (futureRange.length = futureMax - period);`. This very occasionally has a practical use in `for` loops, where you can use it to do multiple assignments in the increment section of the `for()`, but mostly this tidbit is only useful for making strange code.

CoreFoundation/Base.subproj/CFPriv.h
```diff
@@ -26,14 +26,21 @@
...
-    #if !DEPLOYMENT_RUNTIME_OBJC
-        #define CF_CROSS_PLATFORM_EXPORT extern
-    #else
-        #define CF_CROSS_PLATFORM_EXPORT static __attribute__((used))
-    #endif
+#define CF_CROSS_PLATFORM_EXPORT extern
+
+#if TARGET_OS_WIN32
+  // No C99 support
+  #define _CF_RESTRICT
```
compnerd: We should conditional this over `#if defined(_MSC_VER) && !defined(__clang__)`

millenomi: I’ll write a follow-up bug to do it tomorrow.

millenomi: I’ll write a follow-up bug to do it tomorrow.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:

millenomi:

PR #1694, opened 2018-09-13
Align __NSCFType with CFRuntimeBase / _CFInfo

Kaiede:
__NSCFType has been using Int32 for _cfinfo. More recently in Swift 4, it’s been noticed that this can cause problems on 32-bit ARM and should be aligned with _CFInfo instead.

Currently, there are no known issues on 32-bit ARM because of this. However, it'd be nice to fix this before some future change breaks because _cfinfo is an Int32.

PR #1709, opened 2018-09-27
Fix CFInfo on 64-bit big endian platforms.

mundaym:
Prior to this change 64-bit platforms treated _cfinfoa as both a
64-bit and 32-bit variable. When accessing _cfinfoa as a 32-bit
variable the code makes the assumption that it is accessing the
lowest (or rightmost) bits of _cfinfoa. Unfortunately on big-endian
platforms it is actually accessing the high (or leftmost) bits.

Fix this issue by unconditionally treating _cfinfoa as a 64-bit
value when targeting a 64-bit platform.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -434,11 +434,7 @@ CF_EXPORT void * __CFConstantStringClassReferencePtr;
 #define CONST_STRING_SECTION
 #endif
 
-#if __BIG_ENDIAN__
-#define _CF_CONST_STR_CFINFOA 0x00000000C8070000
-#else // Little endian:
 #define _CF_CONST_STR_CFINFOA 0x07C8
-#endif // __BIG_ENDIAN__
 
 #define _CF_CONST_STR_CONTENTS(cStr) {{(uintptr_t)&_CF_CONSTANT_STRING_SWIFT_CLASS, _CF_CONSTANT_OBJECT_STRONG_RC, _CF_CONST_STR_CFINFOA}, (uint8_t *)(cStr), sizeof(cStr) - 1}
 
--- a/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/CoreFoundation/Base.subproj/CFRuntime.c
@@ -362,17 +362,10 @@ void _CFEnableZombies(void) {
  On 64 bit, also includes 32 bits more of retain count from 32-63
  */
 
-#if __CF_BIG_ENDIAN__
-#define RC_INCREMENT		(1ULL)
-#define RC_CUSTOM_RC_BIT	(0x800000ULL << 32)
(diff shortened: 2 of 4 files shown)


Foundation/NSSwiftRuntime.swift
```diff
@@ -87,7 +87,13 @@ extension ObjCBool : CustomStringConvertible {
 #endif
 
 internal class __NSCFType : NSObject {
+#if arch(x86_64) || arch(arm64) || arch(s390x) || arch(powerpc64) || arch(powerpc64le)
```
parkera: Do we not have a word-size macro like C does here?

parkera:
Seems reasonable. Want to make sure this lines up with the work @millenomi is doing to merge right now.

Foundation/NSSwiftRuntime.swift
Kaiede: This change also conflicts with a backout we've been trying to do for 32-bit that keeps getting shuffled, reopened, and closed, and expected to be re-opened again. And really, this particular use should be a _CFInfo, not a raw Int.

Kaiede: Because this _cfinfo should match the size of _CFInfo

Kaiede: See https://github.com/apple/swift-corelibs-foundation/pull/1693 (which is a followup of https://github.com/apple/swift-corelibs-foundation/pull/1689 which attempted a similar forward fix as you are applying to 32-bit with this change).

Kaiede:
In addition to the specific comment, I'll point out that #1689 already fixes _CFInfo in a cleaner way.

millenomi:
As soon as I'm finished with the merge, I want to sit down and figure out what is the best solution here. #1694 we want for sure, and let's chat about the definition in CFRuntimeBase.

millenomi:
Yeah, sorry; it's just been a bit of bad timing at the moment because of the behind-the-scenes work for the merge.

You can contact me via private message on the [Swift forums](http://forums.swift.org) if necessary!

millenomi:
And I can also see to set up something for multiple people if needed.

millenomi:
Sounds good.

I have had trouble setting up appropriate Slack stuff, but if you’re all comfortable with iMessage, please send me your contact (mention your GitHub username) via iMessage at millenomi@me.com and I’ll figure out the rest. (If that’s not okay either, I’ll figure out another way.)

Kaiede:
Worked for me.

mundaym:
@swift-ci Please test.

mundaym:
Hi, I've brought this PR up to date with the changes upstream. Can you take another look? Thanks!

mundaym:
Friendly ping.

levivic:
friendly ping

johnno1962:
ping

shahmishal:

PR #1710, opened 2018-09-28
Non-Darwin SDKs product results in per-arch directories.

drodriguez:
Depends on apple/swift#19432

Unix (other than Darwin) and Windows do not support fat binaries.
However, the build system was placing the build results in the
platform/OS folder, instead of the per-architecture folder. Having two
architectures side by side was impossible.

After applying the above patch in the Swift compiler, non-Darwin
platforms will look into the per-architecture folders, so the sibling
projects need to leave the products in the right place.

The change is performed in the CMakeList.txt file (which as far as I
understand is not used to build Foundation at the moment), and the
Python script.

### DIFF
--- a/build.py
+++ b/build.py
@@ -551,12 +551,18 @@ if "LIBDISPATCH_BUILD_DIR" in Configuration.current.variables:
 
 Configuration.current.variables["LIBS_DIRS"] = LIBS_DIRS
 
-extra_script = """
+if "darwin" in triple.lower():
+    extra_script = "INSTALL_TARGET_DIR=${OS}"
+else:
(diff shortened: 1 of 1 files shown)


millenomi:
I will trigger testing on this as needed when the main patch has been vetted.

compnerd:
I think that we should split this up into two changes.  The CMake changes can go in first since they aren't in use just yet.

drodriguez:
I left in this PR the changes in the script. I will create another PR with the CMake changes. We know (because Jenkins passed in the dependent PR) that this changes work and should be enough for most people.

alblue:
@swift-ci please test linux

PR #1711, opened 2018-10-01
Implement part of the value(forKey) method

ddunn2:
In [NSDictionary](https://developer.apple.com/documentation/foundation/nsdictionary) the `value(forKey)` method is currently unsupported. 

This PR aims to implement part of this method based on the following [documentation](https://developer.apple.com/documentation/foundation/nsdictionary/1410210-value). That is the non-KVC path which calls out to the already implemented `object(forKey)` method. 

I've added a unit test for the new method which tests that `value(forKey)` does return the value we expect. I've also added a test for nested dictionaries. 

The KVC path, where the key has a `@` prefix, has been left unsupported.

### DIFF
--- a/Foundation/NSDictionary.swift
+++ b/Foundation/NSDictionary.swift
@@ -33,7 +33,12 @@ open class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
     }
     
     open func value(forKey key: String) -> Any? {
-        NSUnsupported()
+        if key.hasPrefix("@") {
+            NSUnsupported()
+        } else {
+            return object(forKey: key)
+        }
+        
     }
     
     open func keyEnumerator() -> NSEnumerator {
--- a/TestFoundation/TestNSDictionary.swift
+++ b/TestFoundation/TestNSDictionary.swift
@@ -21,6 +21,8 @@ class TestNSDictionary : XCTestCase {
             ("test_writeToFile", test_writeToFile),
             ("test_initWithContentsOfFile", test_initWithContentsOfFile),
             ("test_settingWithStringKey", test_settingWithStringKey),
+            ("test_valueForKey", test_valueForKey),
+            ("test_valueForKeyWithNestedDict", test_valueForKeyWithNestedDict),
         ]
     }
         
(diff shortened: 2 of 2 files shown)


ianpartridge:
@swift-ci please test

TestFoundation/TestNSDictionary.swift
```diff
@@ -217,6 +219,19 @@ class TestNSDictionary : XCTestCase {
         // has crashed in the past
         dict["stringKey"] = "value"
     }
+    
+    func test_valueForKey() {
+        let dict: NSDictionary = ["foo": "bar"]
+        let result = dict.value(forKey: "foo")
+        XCTAssert(result as? String == "bar")
+    }
```
spevans: Not a bit issue buts its generally better to use `XCTAssertEqual()` as it shows the incorrect value if the test fails.

spevans:
@swift-ci please test

ianpartridge:

Foundation/NSDictionary.swift
```diff
@@ -33,7 +33,12 @@ open class NSDictionary : NSObject, NSCopying, NSMutableCopying, NSSecureCoding,
     }
     
     open func value(forKey key: String) -> Any? {
-        NSUnsupported()
+        if key.hasPrefix("@") {
```
millenomi: It is part of our API, but I'm kinda conflicted bringing this in; as it is, we do not have the ability to implement the full contract of KVC in Swift, and I don't know if the presence of a method is dangerous enough.

I would probably at the very least have some sort of diagnostic (via deprecation maybe?) on usage of this method. But that's for another patch.

millenomi:

ianpartridge:
Yeah we found an SDK that wouldn't run on Linux because it was using this method in the non-KVC path. So I figured it was worth fixing that as it's so easy to do.

millenomi:
It is, but we need to be mindful of the consequences. I’m trying to figure out if there’s a way I can mark the superclass’s method as unavailable while keeping this one as available.

If the SDK is open source or the provider is amenable, it may also be useful to fix the issue upstream.

millenomi:
@douggregor — this is a much smaller patch that still triggers the blocking issue, if working with mine is problematic.

ianpartridge:
Yes we're working with upstream to move them to `object(forKey:)` (which they probably meant to use anyway, understandably since dictionaries have "key"s and "value"s).

millenomi:

PR #1713, opened 2018-10-07
CFString: correct section for CFString section

compnerd:
Re-synchronize the section that the CFString structure is emitted into
to match the behaviour of clang.  This is actually not determined by the
target OS but rather the target file format.  This makes the behaviour
match that of `__builtin___CFStringMakeConstantString`.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -416,10 +416,12 @@ CF_EXPORT void * __CFConstantStringClassReferencePtr;
 
 CF_EXPORT void *__CFConstantStringClassReference[];
 
-#if DEPLOYMENT_TARGET_LINUX
-#define CONST_STRING_SECTION __attribute__((section(".cfstr.data")))
+#if defined(__ELF__)
+#define CONST_STRING_SECTION __attribute__((__section__("cfstring")))
+#elif defined(__MACH__)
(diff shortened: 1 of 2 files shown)


millenomi:
I’m not inclined to take this unless there’s a compelling need; the plan is to delete this implementation and figure out if we can use `__builtin__CFStringMake…` in a Swift context instead.

compnerd:
@millenomi I would love to do that instead as well.  I think that getting the sections correct is helpful in at least comparing the results (so we can start looking at the before and after).  That said, that doesn't necessarily require it to get merged :-).  LMK if I can help with the cleanup to switch over to the builtin BTW.

gottesmm:
@millenomi @compnerd I was thinking about this last night and really if you think about it this is a larger problem. I wonder if we can create a general builtin for this pattern. You would construct a literal array element and as a separate argument give the section (it would be optional). Then if we want to do something similar to this, the builtin would just work. That seems like it would be really useful for low level programmers to have given the constraints of a pure c environment. Basically it would allow one to have a constant array that is in a specific section without specifying a storage location (which I think would be required for the attribute).

millenomi:
@gottesmm Can you capture this in a bug so we don’t forget? It sounds like it could be helpful in general.

compnerd:
@millenomi, @gottesmm - any reason to keep this around now that we should be able to just go with `__builtin___CFStringMakeConstantString`?

gottesmm:
If we can use the builtin, I don't see any reason to have this. But I defer to @millenomi

PR #1716, opened 2018-10-08
Fix _CFIterateDirectory behavior on some old file systems

jtxapl:
On some old systems readdir does not fill correct filetypes. Tasks like loading CFBundle will fail on these systems since _CFIterateDirectory relies on readdir. Adding a second check with stat when readdir returns unknown filetype.

### DIFF
--- a/CoreFoundation/Base.subproj/CFFileUtilities.c
+++ b/CoreFoundation/Base.subproj/CFFileUtilities.c
@@ -1075,6 +1075,23 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
         while ((dent = readdir(dirp))) {
 #if DEPLOYMENT_TARGET_LINUX
             CFIndex nameLen = strlen(dent->d_name);
+            if (dent->d_type == DT_UNKNOWN) {
+                // on some old file systems readdir may always fill d_type as DT_UNKNOWN (0), double check with stat
+                struct stat statBuf;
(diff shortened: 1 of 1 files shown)


millenomi:
Which file systems are affected?

Is there some way to test this?

millenomi:
@swift-ci please test

jtxapl:
> Which file systems are affected?

Earlier versions of CentOS / RHEL / Oracle Linux 7 format XFS partitions with ftype=0 (disabling d_type) unless specially configured.

Oracle Linux 7 [Update 4 release notes](https://docs.oracle.com/cd/E52668_01/E88149/html/ol7-features-xfschange.html) mentioned:

> "For systems installed with the Oracle Linux 7 Update 4 installer, when formatting a device using XFS, d_type support is enabled automatically, which means all XFS-formatted partitions are created using the ftype=1 parameter as the default. Whereas, in previous Oracle Linux 7 updates, ftype=0 was the default parameter, meaning d_type was disabled or off and XFS-formatted partitions were created using ftype=0 as the default."

More details can be found [here](https://linuxer.pro/2017/03/what-is-d_type-and-why-docker-overlayfs-need-it/)

> Is there some way to test this?

One can check if filesystem supports d_type via xfs_info. If it shows ftype = 0, can verify readdir always return d_type = 0 and subsequently CFBundle load will always fail without this patch.

parkera:
The diff seems reasonable to me.

PR #1717, opened 2018-10-10
Non-Darwin SDKs product results in per-arch directories (CMake)

drodriguez:
Unix (other than Darwin) and Windows do not support fat binaries.
However, the build system was placing the build results in the
platform/OS folder, instead of the per-architecture folder. Having two
architectures side by side was impossible.

After applying the above patch in the Swift compiler, non-Darwin
platforms will look into the per-architecture folders, so the sibling
projects need to leave the products in the right place.

/cc @compnerd


compnerd:
The CMake changes seem reasonable to me.

compnerd:
@swift-ci please test

drodriguez:
Seems like the same failure as apple/swift#19823. Probably unrelated to this changes, but I will check tomorrow.

compnerd:

compnerd:
Going to merge this since the CMake changes are valid and will not break existing runs.  The work to switch over to CMake for the builds themselves is on a different PR and this shouldn't impact it (maybe will require a rebase).

PR #1719, opened 2018-10-11
CoreFoundation build clean ups

compnerd:
I noticed these things when trying to build CoreFoundation with the CMake based system.

### DIFF
--- a/CoreFoundation/RunLoop.subproj/CFRunLoop.c
+++ b/CoreFoundation/RunLoop.subproj/CFRunLoop.c
@@ -50,7 +50,9 @@ CF_INLINE uint64_t check_uint64_add(uint64_t x, uint64_t y, int32_t* err) {
 #include <dispatch/private.h>
 #elif __HAS_DISPATCH__
 extern dispatch_queue_t _dispatch_runloop_root_queue_create_4CF(const char *_Nullable label, unsigned long flags);
+#if USE_DISPATCH_SOURCE_FOR_TIMERS
 extern mach_port_t _dispatch_runloop_root_queue_get_port_4CF(dispatch_queue_t queue);
+#endif
 extern void _dispatch_source_set_runloop_timer_4CF(dispatch_source_t source, dispatch_time_t start, uint64_t interval, uint64_t leeway);
 extern bool _dispatch_runloop_root_queue_perform_4CF(dispatch_queue_t queue);
 #endif
--- a/CoreFoundation/RunLoop.subproj/CFRunLoop.c
+++ b/CoreFoundation/RunLoop.subproj/CFRunLoop.c
@@ -46,15 +46,24 @@ CF_INLINE uint64_t check_uint64_add(uint64_t x, uint64_t y, int32_t* err) {
 };
 #endif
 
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @millenomi

millenomi:
@swift-ci please test

CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -95,7 +97,7 @@ DISPATCH_EXPORT void _dispatch_main_queue_callback_4CF(void);
 #include <sys/eventfd.h>
 #include <sys/timerfd.h>
 
-dispatch_runloop_handle_t _dispatch_get_main_queue_handle_4CF(void);
```
millenomi: What is this necessary for? Why did we compile earlier without and can't now?

compnerd: I believe that it may be an artifact of my configuration, but, include ordering is also worrisome.  This requires a private header from libdispatch (`private/private.h`) which is attempted to be included below.  I can try to dig through that a bit more, but I thought that desugaring was better than trying to get through to the private headers (or is there a nice approach that apple has for getting to private headers for SPI?)

parkera: Can we move the include up?

compnerd: @parkera, sure, I don't think that should result in anything breaking.  It does require a bit more work to move around as it was embedded in a weird way (or maybe it was the time of day).

compnerd: @parkera, @millenomi - so, I looked into this when I wasn't half asleep.  So, the problem is that the header that it is using (`dispatch/private.h`) is a private header (I know, shocking given the name of the file).  This is possible to install via the `INSTALL_PRIVATE_HEADERS` option.  However, it seems bad to have to install the private headers if you are sharing the libdispatch build (it will get installed into the toolchain image).  We *could* add the libdispatch source path to the search order as a workaround if you don't want to desugar the type.

parkera: Hm, this is the only thing from dispatch's private headers that we use in the whole project?

compnerd: Yeah, it was the only thing that I hit that needed the private header explicitly.  The other things are declared and don't have any sugar that is not replicated here.

compnerd: @parkera, @millenomi - so what is your opinion on this?  The three options that I see (in reverse order of [my] preference):
1) require private headers to always be installed for swift
2) require path to libdispatch sources to get the private header
3) desugar the type

parkera: I could imagine a `has_include` style check here, perhaps redefining the type if the private header is not available?

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1722, opened 2018-10-12
SR-7195: JSONSerialization encodes Double differently on Darwin and Linux

spevans:
- Use .description property of numbers to encode as String
  instead of using CFNumberFormatter for Float/Double,
  and serialising integers explictly.


spevans:
@swift-ci please test

itaiferber:
Looks reasonable — can we add a unit test to cover the specific values mentioned in SR-7195 so we can ensure that this at least helps with the round-tripping?

spevans:
@swift-ci test

spevans:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #1723, opened 2018-10-14
build: strip DESTDIR when performing subinstall

compnerd:
When setting up CoreFoundation for the Foundation build, we need to
strip out any preset DESTDIR to ensure that the installation is not
redirected to the install location for the Foundation build.


compnerd:
@swift-ci please test

compnerd:
CC: @parkera @millenomi

PR #1721, opened 2018-10-12
Revert "Fix DateFormatter TimeZone Setter"

gparker42:
Reverts apple/swift-corelibs-foundation#1704

Test test_expectedTimeZone is failing in CI.

PR #1725, opened 2018-10-14
Reapply "Fix DateFormatter TimeZone Setter"

Kaiede:
This is a fix for PR #1704, which was reverted by #1721 due to a break in the test I added in the original PR. 

This includes rebased versions of the commit history of #1704, plus a new commit that disables Case 1 of the test which was failing in the case of GMT, because DateFormatter was spitting out UTC as the abbreviation. I've opened https://bugs.swift.org/browse/SR-8994 to track the issue of TimeZone's abbreviation and DateFormatter not agreeing on GMT vs UTC.


spevans:
@swift-ci test

PR #1732, opened 2018-10-20
SR-8999: NSData contentsOf expects st_size to be the size of the file.

spevans:
- Modify FileHandle._readDataOfLength(untilEOF:options:) to ignore
  stat.st_size and just use the result of read() to determine whether
  to keep reading.

- Also convert syscall error handling to throw instead of calling
  fatalError().

- NSData.readBytesFromFileWithExtendedAttributes(options:) now uses
  FileHandle._readDataOfLength(untilEOF:options:) to read from a file
  rather than duplicating the file reading logic.

- (Linux) Add a test to read a file in /sys which reports
  stat.st_size > size of the file's contents.

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -31,7 +31,12 @@ open class FileHandle : NSObject, NSSecureCoding {
     }
 
     open var availableData: Data {
-        return _readDataOfLength(Int.max, untilEOF: false)
+        do {
+            let readResult = try _readDataOfLength(Int.max, untilEOF: false)
+            return readResult.toData()
+        } catch {
+            return Data()
+        }
     }
     
     open func readDataToEndOfFile() -> Data {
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test

spevans:
@swift-ci test

Foundation/FileHandle.swift
```diff
@@ -31,102 +31,95 @@ open class FileHandle : NSObject, NSSecureCoding {
...
-                // Make sure there is always at least amountToRead bytes available in the buffer.
-                if (currentAllocationSize - total) < amountToRead {
-                    currentAllocationSize *= 2
-                    dynamicBuffer = _CFReallocf(dynamicBuffer!, currentAllocationSize)
-                    if dynamicBuffer == nil {
-                        fatalError("unable to allocate backing buffer")
+
+        let readBlockSize: Int
+        if statbuf.st_mode & S_IFMT == S_IFREG {
+            // TODO: Should files over a certain size always be mmap()'d?
```
millenomi: Immediate answer: no, unless if options contains `.alwaysMapped` or `.mappedIfSafe`.

(I don't see `.mappedIfSafe` handled here at all? Are you just assuming mapping is always unsafe? This is allowed by the contract but we should have a follow-up.)

millenomi:
@itaiferber @phausler If you want to chip in today, or I'll merge.

Foundation/FileHandle.swift
spevans: `.mappedIfSafe` wasn't handled in the previous code so I haven't added it but  I can do a separate PR for it.

phausler:
Seems reasonable to me

PR #1733, opened 2018-10-23
build: correctly handle dependencies

compnerd:
Ensure that the dependencies are wired up to the build commands.  We
would previously not track them correctly.


compnerd:

PR #1734, opened 2018-10-23
build: adjust for split libdispatch SDK overlay

compnerd:
This adjust the build to explicitly link against swiftDispatch to enable
the split SDK overlay on non-Darwin targets.  This would be fixed when
we switch over to the CMake based build which uses the swift driver to
link which will use the auto-link-extract functionality to make the ELF
targets build.


compnerd:
https://github.com/apple/swift-corelibs-libdispatch/pull/401

@swift-ci please test

compnerd:
apple/swift-corelibs-libdispatch#401
apple/swift-corelibs-xctest#227

@swift-ci please test

compnerd:

millenomi:

PR #1735, opened 2018-10-26
Add support for compiler-managed CFSTR() on non-Darwin systems

millenomi:
This gets the swift-corelibs-foundation implementation closer to the Darwin implementation by making the compiler lay out constant CFStrings produced by CFSTR(). It requires a [corresponding clang patch](https://github.com/apple/swift-clang/pull/213) to work.

This is only enabled for not-`TARGET_OS_MAC` for now, since the clang used to build Swift here is not the same as the one used by Xcode when using a Swift toolchain. When Xcode gets the new clang, we can enable this for `TARGET_OS_MAC` as well.

rdar://42898576

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -421,7 +421,7 @@ CF_PRIVATE Boolean __CFProcessIsRestricted(void);
 
 CF_EXPORT void * __CFConstantStringClassReferencePtr;
 
-#if DEPLOYMENT_RUNTIME_SWIFT
+#if DEPLOYMENT_RUNTIME_SWIFT && TARGET_OS_MAC
 
 #if DEPLOYMENT_TARGET_LINUX
 #define CONST_STRING_SECTION __attribute__((section(".cfstr.data")))
--- a/CoreFoundation/String.subproj/CFString.h
+++ b/CoreFoundation/String.subproj/CFString.h
@@ -152,14 +152,16 @@ since it is the default choice with Mac OS X developer tools.
 #endif
 
 #if DEPLOYMENT_RUNTIME_SWIFT
-
-#if TARGET_OS_MAC
-#define _CF_CONSTANT_STRING_SWIFT_CLASS $s15SwiftFoundation19_NSCFConstantStringCN
-#else
-#define _CF_CONSTANT_STRING_SWIFT_CLASS $s10Foundation19_NSCFConstantStringCN
-#endif
+    #if TARGET_OS_MAC
+        #define _CF_CONSTANT_STRING_SWIFT_CLASS $s15SwiftFoundation19_NSCFConstantStringCN
+    #else
+        #define _CF_CONSTANT_STRING_SWIFT_CLASS $s10Foundation19_NSCFConstantStringCN
+    #endif
 
 CF_EXPORT void *_CF_CONSTANT_STRING_SWIFT_CLASS[];
+#endif
+
+#if DEPLOYMENT_RUNTIME_SWIFT && TARGET_OS_MAC
 
 struct __CFConstStr {
     struct {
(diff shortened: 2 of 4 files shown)


millenomi:
Please test with this pull request:
https://github.com/apple/swift-clang/pull/213

@swift-ci please test

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -421,7 +421,7 @@ CF_PRIVATE Boolean __CFProcessIsRestricted(void);
 
 CF_EXPORT void * __CFConstantStringClassReferencePtr;
 
-#if DEPLOYMENT_RUNTIME_SWIFT
+#if DEPLOYMENT_RUNTIME_SWIFT && TARGET_OS_MAC
 
 #if DEPLOYMENT_TARGET_LINUX
 #define CONST_STRING_SECTION __attribute__((section(".cfstr.data")))
```
compnerd: Is this still used?  The section itself is not compatible with what clang has done for a very long time.

CoreFoundation/String.subproj/CFString.h
```diff
@@ -152,14 +152,16 @@ since it is the default choice with Mac OS X developer tools.
 #endif
 
 #if DEPLOYMENT_RUNTIME_SWIFT
+    #if TARGET_OS_MAC
+        #define _CF_CONSTANT_STRING_SWIFT_CLASS $s15SwiftFoundation19_NSCFConstantStringCN
+    #else
+        #define _CF_CONSTANT_STRING_SWIFT_CLASS $s10Foundation19_NSCFConstantStringCN
+    #endif
```
compnerd: Do we care about the different versions of the ABI?

CoreFoundation/String.subproj/CFString.h
```diff
@@ -175,20 +177,14 @@ struct __CFConstStr {
 #endif // defined(__LP64__) || defined(__LLP64__)
 };
```
compnerd: Should the structure definition be updated to match the reality of the compiler?

build.py
```diff
@@ -63,6 +63,7 @@
 	'-Wno-int-conversion',
 	'-Wno-unused-function',
 	'-I./',
+	'-fcf-runtime-abi=swift',
```
compnerd: We should update the CMake based build system as well.

CoreFoundation/String.subproj/CFString.h
millenomi: This structure definition is used only when the compiler is not involved, and is now ifdef'd away. I checked that the actual struct used by CFStringRef (`__CFString`) does use intptr_t lengths as we discussed.

CoreFoundation/String.subproj/CFString.h
millenomi: No, because Foundation is in lockstep with Swift anyway.

CoreFoundation/Base.subproj/CFInternal.h
millenomi: Yeah; this is the path used when clang is _not_ involved in laying out the CFSTR(). We still need it for Mac as outlined in the PR description.

compnerd: Should the section be updated to be correct then?  The literal is placed into `.rodata.str1.1`, the structures into `cfstring`.

CoreFoundation/String.subproj/CFString.h
compnerd: SGTM

CoreFoundation/String.subproj/CFString.h
compnerd: Any reason to not make the two be identical?

CoreFoundation/Base.subproj/CFInternal.h
millenomi: I'll just remove the dead code.

millenomi: Per discussion: this seems to be diff'd against an older revision than master, and that code is no longer in apple/swift-corelibs-foundation.

millenomi:
Please test with this pull request:
apple/swift-clang#213

@swift-ci please test

millenomi:
Ready to merge whenever the clang change is.

PR #1737, opened 2018-10-28
build: pass `-target` when linking the swift content

compnerd:
This allows us to cross-link when building with CMake.


compnerd:

PR #1738, opened 2018-10-29
CoreFoundation: mark tentative definitions as declarations

compnerd:
This converts the tentative definitions in the CoreFoundation header set
to explicit declarations.  This should ensure that we get the correct
definition in the final image.  This was caught by the CMake based build
and swift-package-manager getting a tentative definition for the
CFString converter.


compnerd:
CC: @millenomi @phausler

compnerd:
@swift-ci please test

millenomi:
`../Ninja-ReleaseAssert/foundation-linux-x86_64/Foundation/CoreFoundation/Base.subproj/CFRuntime.c.o:CFRuntime.c:__CFRuntimeClassTable: error: undefined reference to '__CFMessagePortClass'`

CFMessagePort.c isn't being compiled.

CoreFoundation/AppServices.subproj/CFUserNotification.h
```diff
@@ -131,53 +131,53 @@ CF_INLINE CFOptionFlags CFUserNotificationPopUpSelection(CFIndex n) {return ((CF
 /* Keys */
 
 CF_EXPORT
-const CFStringRef kCFUserNotificationIconURLKey;
+extern const CFStringRef kCFUserNotificationIconURLKey;
```
millenomi: `CF_EXPORT` _must_ imply `extern`. If the current definition doesn't, we must change its `#define`.

millenomi: The current definition _does_ imply it:

```C
#if TARGET_OS_WIN32
    #if !defined(CF_EXPORT)
        #if defined(CF_BUILDING_CF) && defined(__cplusplus)
            #define CF_EXPORT extern "C" __declspec(dllexport)
        #elif defined(CF_BUILDING_CF) && !defined(__cplusplus)
            #define CF_EXPORT extern __declspec(dllexport)
        #elif defined(__cplusplus)
            #define CF_EXPORT extern "C" __declspec(dllimport)
        #else
            #define CF_EXPORT extern __declspec(dllimport)
        #endif
    #endif
#else
#define CF_EXPORT extern
#endif
```

compnerd: `CF_EXPORT` really shouldn't imply `extern`.  Consider the following bit in the implementation:

```
CF_EXPORT void CFFunction(void) {
}
```

The definition itself should not get marked as `extern`.  I think that this is why this was mixed into the other change.  We should split the two apart.  There is a fair amount of usage of this pattern through out the current source tree.

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -591,8 +591,8 @@ CF_PRIVATE void _CF_dispatch_once(dispatch_once_t *, void (^)(void));
 #endif
 
 #if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_EMBEDDED_MINI
-CF_PRIVATE _Atomic(uint8_t) __CF120293;
-CF_PRIVATE _Atomic(uint8_t) __CF120290;
+CF_PRIVATE exxtern _Atomic(uint8_t) __CF120293;
```
millenomi: This cannot be right. Likewise, `CF_PRIVATE` must imply `extern`.

CoreFoundation/AppServices.subproj/CFUserNotification.h
millenomi: The function is incorrect in this instance. We should split that into:

```
CF_EXTERN void CFFunction(void);
void CFFunction() { … }
```

millenomi: @compnerd is there any platform we want to build on that doesn't support extern on function declarations? I just tested with clang on Linux and it does support it.

millenomi: (by which I mean functions with bodies.)

compnerd: Windows :-)

CoreFoundation/Base.subproj/CFInternal.h
compnerd: Okay, so we should just change the definition of `CF_PRIVATE` then?

CoreFoundation/AppServices.subproj/CFUserNotification.h
millenomi: Even on clang?

compnerd: Yeah, `clang-cl` matches the behaviour

PR #1739, opened 2018-10-29
Implemented url resourceValues and updating file timestamps

amosavian:
I did my best to implement all possible keys, both on getting and setting.

Some keys were platform specific and has not any meaning on other platforms, thus marked as `NSUnsupported()`

keys related to drive capabilities were not implemented because there was no straight-forward way to obtain informations.

i marked some Windows specific ways as `NSUnimplmented()` as I did not know how to import Win32 functions.

### DIFF
--- a/CoreFoundation/URL.subproj/CFURL.c
+++ b/CoreFoundation/URL.subproj/CFURL.c
@@ -5215,59 +5215,78 @@ CFURLRef CFURLCreateFileReferenceURL(CFAllocatorRef alloc, CFURLRef url, CFError
 #include <linux/fs.h>
 #endif
 
-#if DEPLOYMENT_TARGET_WINDOWS
+#if __has_include(<sys/statvfs.h>)
+#include <sys/statvfs.h>
+#endif
+
+#if TARGET_OS_LINUX
+#include <mntent.h>
+#endif
+
+#if TARGET_OS_WINDOWS
 #include <windows.h>
 #endif
 
-Boolean __CFURLResourceValueHasFlag(CFURLRef url, CFURLResourcePropertyFlags key, errno_t *error) {
+CONST_STRING_DECL(kVolumeInfoMountFrom, "mountfrom")
+CONST_STRING_DECL(kVolumeInfoMountTo, "mountto")
+CONST_STRING_DECL(kVolumeInfoType, "typename")
+
+__nullable CFBooleanRef _CFURLResourceValueHasFlag(CFURLRef url, CFURLResourcePropertyFlags key, errno_t *error)
+{
     char pathBuf[CFMaxPathSize];
     CFStringRef path = CFURLCopyPath(url);
-    if (!CFStringGetFileSystemRepresentation(path, pathBuf, CFMaxPathSize)) return false;
+    if (!CFStringGetFileSystemRepresentation(path, pathBuf, CFMaxPathSize)) return NULL;
+    CFRelease(path);
     
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_FREEBSD
+#if TARGET_OS_DARWIN || TARGET_OS_BSD
     struct stat st;
 #endif
     switch (key) {
         case kCFURLResourceIsHidden:
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_FREEBSD
-            if (stat(pathBuf, &st)) {
+        {
+#if TARGET_OS_DARWIN || TARGET_OS_BSD
+            if (lstat(&pathBuf[0], &st)) {
                 error = errno;
-                return false;
+                return NULL;
             }
-            return st.st_flags & UF_HIDDEN;
+            return st.st_flags & UF_HIDDEN ? kCFBooleanTrue : kCFBooleanFalse;
 #endif
-#if DEPLOYMENT_TARGET_WINDOWS
+#if TARGET_OS_WINDOWS
             int attr = GetFileAttributes(path);
-            return (attr & FILE_ATTRIBUTE_HIDDEN) != 0;
+            return (attr & FILE_ATTRIBUTE_HIDDEN) != 0 ? kCFBooleanTrue : kCFBooleanFalse;
 #else
             // Unix-based systems
-            if (CFStringHasPrefix(CFURLCopyLastPathComponent(url), CFSTR("."))) {
-                return true;
-            }
+            CFStringRef lastPathComponent = CFURLCopyLastPathComponent(url);
+            Boolean isHidden = CFStringHasPrefix(lastPathComponent, CFSTR("."));
+            CFRelease(lastPathComponent);
+            return isHidden ? kCFBooleanTrue : kCFBooleanFalse;
 #endif
-            return false;
+            break;
+        }
             
         case kCFURLResourceIsUserImmutable:
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_FREEBSD
-            if (stat(pathBuf, &st)) {
+        {
+#if TARGET_OS_DARWIN || TARGET_OS_BSD
+            if (lstat(&pathBuf[0], &st)) {
                 error = errno;
                 return false;
             }
-            return st.st_flags & UF_IMMUTABLE;
-#else
-            return false;
+            return st.st_flags & UF_IMMUTABLE ? kCFBooleanTrue : kCFBooleanFalse;
 #endif
+            break;
+        }
             
         case kCFURLResourceIsSystemImmutable:
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_FREEBSD
-            if (stat(pathBuf, &st)) {
+        {
+#if TARGET_OS_DARWIN || TARGET_OS_BSD
+            if (lstat(&pathBuf[0], &st)) {
                 error = errno;
                 return false;
             }
-            return st.st_flags & SF_IMMUTABLE;
+            return st.st_flags & SF_IMMUTABLE ? kCFBooleanTrue : kCFBooleanFalse;
 #endif
 #if __has_include(<linux/fs.h>)
-            int fd = _CFOpenFile(pathBuf, O_RDONLY);
+            int fd = _CFOpenFile(&pathBuf[0], O_RDONLY);
             if (fd < 0) {
                 error = errno;
                 return false;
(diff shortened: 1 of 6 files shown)


Foundation/FileManager.swift
```diff
@@ -509,28 +509,172 @@ open class FileManager : NSObject {
...
-        for attribute in attributes.keys {
-            if attribute == .posixPermissions {
-                guard let number = attributes[attribute] as? NSNumber else {
-                    fatalError("Can't set file permissions to \(attributes[attribute] as Any?)")
+        for (key, value) in attributes {
+            switch key {
+            case .creationDate:
+                guard let date = value as? Date else {
+                    fatalError("Can't set file modification date to \(value as Any?)")
+                }
```
spevans: Since this function `throws` it is better to throw an error rather than call `fatalError()`

Foundation/FileManager.swift
```diff
@@ -509,28 +509,172 @@ open class FileManager : NSObject {
...
-            } else {
-                fatalError("Attribute type not implemented: \(attribute)")
+                
+            case .ownerAccountID:
+                guard let id = value as? NSNumber else {
+                    fatalError("Can't set file owner to \(value as Any?)")
+                }
+                _ = self._fileSystemRepresentation(withPath: path) {
+                    chown($0, uid_t(id.uint32Value), 0 &- 1)
+                }
```
spevans: All of the `chmod` / `chown` system calls should throw with `errno` if the call fails

Foundation/FileManager.swift
amosavian: I've duplicated Darwin's version behavior. It does crash with `NSException` if you pass another type, like string.

Foundation/FileManager.swift
amosavian: You are right. I shall fix

Foundation/NSURL.swift
```diff
@@ -638,13 +665,464 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
...
+    }
+    
+    open func resourceValues(forKeys keys: [URLResourceKey]) throws -> [URLResourceKey : Any] {
+        
+        func hasFlag(path: String, flag: Int32) throws -> Bool {
+            #if canImport(Darwin) || os(FreeBSD)
+            let stat = try FileManager.default._lstatFile(atPath: path)
+            return stat.st_flags & UInt32(flag) != 0
+            #elseif canImport(Glibc)
+            let fd = open(path, O_RDWR)
```
spevans: `path` needs to be converted using `_fileSystemRepresentation` but it might actually be better to replace the `open()` with something like:
```
let fh = try FileHandle(..)
var attr: Int32 = 0
ioctl(fh.fileDescriptor, FS_IOC_GETFLAGS, &attr)
return attrs & flag != 0
```

amosavian:
@spevans I've pushed changes. I would squash it after you've checked changes.

spevans:
@swift-ci test

Foundation/FileManager.swift
spevans: In that case it might be better to use `as!` instead of the `as?` / `fatalError()` as it will have the same outcome and should simplify things a bit.

Foundation/FileManager.swift
```diff
@@ -704,43 +856,41 @@ open class FileManager : NSObject {
      
         This method replaces fileSystemAttributesAtPath:.
      */
- #if os(Android)
-    @available(*, unavailable, message: "Unsuppported on this platform")
     open func attributesOfFileSystem(forPath path: String) throws -> [FileAttributeKey : Any] {
-        NSUnsupported()
```
spevans: Its better to leave in the `@available` for Android as it will catch usage at compile time  rather than the `NSUnsupported()` below,  at runtime.

Foundation/FileManager.swift
amosavian: These messages are exactly the NSException's messages which Darwin throws. That means we would have just a cryptic crash when final user uses compiled version of SF.

spevans: Ah, that makes sense. However in the `setResourceValues` method `fatalError()` is called without a message so those ones could be changed.

Foundation/NSURL.swift
```diff
@@ -638,13 +665,462 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
...
+            let stat = try FileManager.default._lstatFile(atPath: path)
+            return stat.st_flags & UInt32(flag) != 0
+            #elseif canImport(Glibc)
+            let fh = try FileHandle(forUpdating: self._swiftObject)
+            var attrs: Int32 = 0
+            if ioctl(fh.fileDescriptor, FS_IOC_GETFLAGS, &attrs) != 0 {
+                throw _NSErrorWithErrno(errno, reading: true, path: path)
+            }
+            return attrs & flag != 0
+            #elseif os(Windows)
```
spevans: For now it might be bets to change this whole section  to 
```
#else
#error("Requires a platform-specific implementation")
#endif
```

As it will notify the platform maintainers for Windows/Android etc to add the appropriate support.

Foundation/NSURL.swift
```diff
@@ -638,13 +665,462 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
...
+                newValues[.volumeTotalCapacityKey] = attributes?[.systemSize]
+                newValues[.volumeAvailableCapacityKey] = attributes?[.systemFreeSize]
+                newValues[.volumeAvailableCapacityForImportantUsageKey] = attributes?[.systemFreeSize]
+                newValues[.volumeAvailableCapacityForOpportunisticUsageKey] = attributes?[.systemFreeSize]
+                newValues[.volumeResourceCountKey] = attributes?[.systemNodes]
+                
+            case .volumeURLKey, .isVolumeKey:
+                let rootPath: String
+                #if canImport(Darwin) || os(FreeBSD)
+                var root = try _statfs(path: path).f_mntonname
```
spevans: It looks like `_statfs(path: path)` is called in multiple places so it might be best to move it outside of the loop and only call it once (and only if needed).

The same applies to `mountedVolumeURLS` as it could be called twice.

Foundation/FileManager.swift
amosavian: Ok done

Foundation/FileManager.swift
amosavian: You are right

Foundation/NSURL.swift
amosavian: There are only two places which I use `_statfs()` in this loop. Furthermore it is only available in Darwin and FreeBSD thus moving this would add far more complexity to code.

Foundation/NSURL.swift
```diff
@@ -638,13 +665,462 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
...
+    open func resourceValues(forKeys keys: [URLResourceKey]) throws -> [URLResourceKey : Any] {
+        
+        func hasFlag(path: String, flag: Int32) throws -> Bool {
+            #if canImport(Darwin) || os(FreeBSD)
+            let stat = try FileManager.default._lstatFile(atPath: path)
+            return stat.st_flags & UInt32(flag) != 0
+            #elseif canImport(Glibc)
+            let fh = try FileHandle(forUpdating: self._swiftObject)
+            var attrs: Int32 = 0
+            if ioctl(fh.fileDescriptor, FS_IOC_GETFLAGS, &attrs) != 0 {
```
amosavian: @spevans In test I see linux compile error which does not recongize `FS_IOC_GETFLAGS` and other `ioctl iflags`. I don't know how to import related `<linux/fs.h>` header.

Foundation/FileManager.swift
amosavian: I still wonder why it is unsupported on Android!

Foundation/NSURL.swift
amosavian: @spevans I removed this code from Swift entirely and moved it CoreFoundation

parkera:
This is great work, thanks @amosavian.

CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -5208,3 +5208,152 @@ CFURLRef CFURLCreateFilePathURL(CFAllocatorRef alloc, CFURLRef url, CFErrorRef *
...
+#if DEPLOYMENT_TARGET_WINDOWS
+#include <windows.h>
+#endif
+
+Boolean __CFURLResourceValueHasFlag(CFURLRef url, CFURLResourcePropertyFlags key, errno_t *error) {
+    char pathBuf[CFMaxPathSize];
+    CFStringRef path = CFURLCopyPath(url);
+    if (!CFStringGetFileSystemRepresentation(path, pathBuf, CFMaxPathSize)) return false;
+    
+#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_FREEBSD
```
parkera:
```suggestion
#if TARGET_OS_MAC || TARGET_OS_FREEBSD
```

Do we have a `TARGET_OS_FREEBSD`? Worth checking.

CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -5208,3 +5208,152 @@ CFURLRef CFURLCreateFilePathURL(CFAllocatorRef alloc, CFURLRef url, CFErrorRef *
...
+    switch (key) {
+        case kCFURLResourceIsHidden:
+#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_FREEBSD
+            if (stat(pathBuf, &st)) {
+                error = errno;
+                return false;
+            }
+            return st.st_flags & UF_HIDDEN;
+#endif
+#if DEPLOYMENT_TARGET_WINDOWS
```
parkera:
```suggestion
#if TARGET_OS_WINDOWS
```

(I think you get the idea on these)

CoreFoundation/URL.subproj/CFURLPriv.h
```diff
@@ -432,6 +432,12 @@ typedef unsigned long long CFURLResourcePropertyFlags;
 CF_EXPORT
 Boolean _CFURLGetResourcePropertyFlags(CFURLRef url, CFURLResourcePropertyFlags mask, CFURLResourcePropertyFlags *flags, CFErrorRef *error) API_AVAILABLE(macos(10.6), ios(4.0), watchos(2.0), tvos(9.0));
 
+CF_EXPORT
+Boolean __CFURLResourceValueHasFlag(CFURLRef url, CFURLResourcePropertyFlags key, errno_t *error);
+CF_EXPORT
+Boolean _CFURLResourceValueSetFlag(CFURLRef url, CFURLResourcePropertyFlags key, Boolean set, errno_t *error);
+
```
parkera: I think these should be in `ForSwiftFoundationOnly` and decorated with `CF_SWIFT_EXPORT`, correct?

Foundation/FileManager.swift
parkera: Agreed here that passing the wrong type in is a developer error worthy of trapping. I think the message could be clearer about the problem -- it's not important to me to maintain exactly the same string for a fatal error as an `NSException` on Darwin.

CoreFoundation/URL.subproj/CFURLPriv.h
amosavian: @parkera I applied your suggestions and have done some refactoring and enhancements.

Foundation/FileManager.swift
```diff
@@ -509,28 +478,168 @@ open class FileManager : NSObject {
...
+                            throw _NSErrorWithErrno(errno, reading: false, path: path)
+                        }
+                    }
+                }
+                
+            case .immutable:
+                try NSURL(fileURLWithPath: path).setResourceValue(value, forKey: .isUserImmutableKey)
+                
+            case .busy, .extensionHidden:
+                NSUnimplemented()
```
amosavian: @parkera These keys are only meaningful on Darwin, should we just ignore these instead of crash.

Foundation/NSURL.swift
```diff
@@ -638,13 +654,464 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
...
+                fallthrough
+                
+            // These keys are Darwin specific. Some can be implemented using xattrs, but still no
+            // linux disto would honor them!
+            case .hasHiddenExtensionKey, .isPackageKey , .isApplicationKey, .typeIdentifierKey,
+                 .localizedTypeDescriptionKey, .labelNumberKey, .labelColorKey, .localizedLabelKey,
+                 .effectiveIconKey, .customIconKey, .fileSecurityKey, .isExcludedFromBackupKey,
+                 .tagNamesKey, .isMountTriggerKey, .fileResourceIdentifierKey, .quarantinePropertiesKey,
+                 .thumbnailDictionaryKey, .thumbnailKey, .isAliasFileKey:
+                NSUnsupported()
```
amosavian: @parkera same here, we have no equal functionality on other OSes.

Foundation/FileManager.swift
amosavian: @parkera I'm not a native speaker. I would appreciate if you have alternative sentences to replace.

Foundation/FileManager.swift
parkera: Hm, I think we should probably ignore them.

amosavian: Thanks. I will change these.

shahmishal:

PR #1740, opened 2018-10-30
Fix linkage for Linux and BSD symbol declarations that use `CF_PRIVATE`

millenomi:
Symbols using `CF_PRIVATE` are meant to be available outside the compilation unit where they're declared, but not outside Core Foundation. Headers that use `CF_PRIVATE` thus mean to declare, not define, these symbols. Match Darwin by adding `extern` to the `CF_PRIVATE` definition, and then add `-fno-common` to also match how Darwin compiles and detect double symbol definitions.

This fixes Linux crashes in CFString encoding introduced by the Mojave merge.


millenomi:
@compnerd ping.

CMake flag addition incoming.

millenomi:
@swift-ci please test

compnerd:
LGTM, I do believe that this should fix the issue as the types were marked with `CF_PRIVATE`.

parkera:
Thank you @compnerd for your work on this!

PR #1741, opened 2018-10-30
SR-8970 Fix for handling of post with empty post body

mkistler:
This PR makes a small fix to the URLSession code for handling POST requests that avoids adding a `content-type` header when the POST body is empty.


spevans:
@swift-ci test

PR #1742, opened 2018-10-30
Mark FileManager.createFile as @discardableResult

keith:
On iOS / macOS this API comes from Objective-C which is implicitly
@discardableResult, this means it doesn't warn on those platforms, but
previously did on Linux


dduan:
@swift-ci test

millenomi:
I'm really not into adding this piecemeal. It's fine, but we should do it to all the equivalent API, or none.

PR #1744, opened 2018-11-02
Update CMake build to pass -I ICU_INCLUDE_DIR

spevans:
- This allows building with a non system libicu.

- Required to build https://github.com/apple/swift/pull/19860

### DIFF
--- a/CoreFoundation/CMakeLists.txt
+++ b/CoreFoundation/CMakeLists.txt
@@ -352,15 +352,16 @@ if(CMAKE_SYSTEM_NAME STREQUAL Linux OR CMAKE_SYSTEM_NAME STREQUAL Android)
   target_include_directories(CoreFoundation
                              PRIVATE
                                ${CURL_INCLUDE_DIRS})
+  find_package(ICU COMPONENTS uc i18n REQUIRED)
+  target_include_directories(CoreFoundation
+                             PRIVATE
+                               ${ICU_INCLUDE_DIR})
 else()
   target_include_directories(CoreFoundation
                              PRIVATE
                                "/usr/include"
                                "/usr/include/libxml2")
 endif()
-target_include_directories(CoreFoundation
-                           PRIVATE
-                             ${ICU_INCLUDE_DIR})
 
 target_compile_options(CoreFoundation
                        PRIVATE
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test

CoreFoundation/CMakeLists.txt
```diff
@@ -358,6 +358,10 @@ else()
                                "/usr/include"
                                "/usr/include/libxml2")
 endif()
+target_include_directories(CoreFoundation
+                           PRIVATE
+                             ${ICU_INCLUDE_DIR})
+
```
compnerd: There is no reason to assume that `ICU_INCLUDE_DIR` is defined here.  This is built as an external project (since you can build CF outside of Foundation).  You need to replicate `FindICU.cmake` and add the appropriate `find_package`.

spevans:
@swift-ci test

CoreFoundation/CMakeLists.txt
spevans: I moved it into the Linux section and used `FindICU`. Although I think even if building CF outside of Foundation, ICU is still required as its CF that actually makes the ICU calls.

compnerd: Yeah, it is required.  I was just saying that you need the `find_package` if you are going to do this :-).

CoreFoundation/cmake/modules/FindICU.cmake
```diff
@@ -0,0 +1 @@
+../../../cmake/modules/FindICU.cmake
```
compnerd: Ugh, the symlink is painful since it will break checkouts on Windows.  Can you please copy?

tkremenek:
https://github.com/apple/swift/pull/19860 has been merged

tkremenek:
@swift-ci clean test

CoreFoundation/cmake/modules/FindICU.cmake
spevans: Fixed

spevans:
@swift-ci test

spevans:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #1746, opened 2018-11-02
[cmake] Find ICU in CoreFoundation.

drodriguez:
ICU is also included in CoreFoundation, but the include paths were
not used, which meant that the headers were probably the system
headers, if they were there. This might have work in Ubuntu, but
might not work in other systems with older ICUs.

The patch passes the ICU_ROOT from the top level CMakeList.txt into
the CoreFoundation one and performs a find_package to find the right
include path.

/cc @compnerd


spevans:
Should probably keep the addition of `"${CMAKE_CURRENT_SOURCE_DIR}/../cmake/modules"` to `CMAKE_MODULE_PATH` as well and get rid of the duplicate of `FindICU.cmake` that I made

drodriguez:
I hope this was what you both were intending: I removed the duplicated script and I passed `ICU_ROOT` instead of `ICU_INCLUDE_DIR`.

compnerd:
@swift-ci please test

compnerd:

millenomi:

PR #1747, opened 2018-11-02
build: repair the build on 14.04

compnerd:
CMake 3.4.3 does not support the <files>... <directory> pattern for
copy_if_different.  Adjust for that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,6 +229,7 @@ add_swift_library(Foundation
                     ${CMAKE_C_COMPILER_TARGET}
                   CFLAGS
                     ${deployment_enable_libdispatch}
+                    -F${install_dir}/System/Library/Frameworks
                   LINK_FLAGS
                     -L${install_dir}/usr/lib
                     -lCoreFoundation
@@ -240,7 +241,6 @@ add_swift_library(Foundation
                   SWIFT_FLAGS
                     -DDEPLOYMENT_RUNTIME_SWIFT
                     ${deployment_enable_libdispatch}
-                    -Fsystem;${install_dir}/System/Library/Frameworks
                     -I;${ICU_INCLUDE_DIR}
                     ${libdispatch_cflags}
                     ${swift_enable_testing}
@@ -252,6 +252,7 @@ add_swift_executable(plutil
                        Tools/plutil/main.swift
                      CFLAGS
                        ${deployment_enable_libdispatch}
+                       -F${install_dir}/System/Library/Frameworks
                      LINK_FLAGS
                        -L${CMAKE_CURRENT_BINARY_DIR}
                        ${libdispatch_ldflags}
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

gottesmm:
@swift-ci test linux platform

spevans:
@swift-ci test

compnerd:
@swift-ci please test

spevans:
Looking at the Ubuntu14.04 build host logs they show the cmake version as 3.4.3 but https://github.com/apple/swift/blob/master/docs/Ubuntu14.md says that 3.5.2 should be installed.

Maybe we should just get the 14.04/16.04 hosts updated to 3.10 and then set a minimum version check in Swift's `CMakeLists.txt`. Now that we are using CMake everywhere having to use such a low version may cause problems in the future.

gottesmm:
@swift-ci please test

gottesmm:
@spevans A few things:

1. We can not just update the minimum accepted cmake version across all projects. This is because changing the minimum required cmake for LLVM/clang is a non-trivial thing that takes community input. IIRC from talking with @llvm-beanz there isn't much of a push in LLVM to upgrade to a newer cmake yet. But I might be wrong. @llvm-beanz has this changed?

2. If we can't change the minimum version accepted by LLVM/clang, I would be hesitant to upgrade the minimum version for swift and all downstream dependencies since we use cmake chaining to pass information in between different cmake invocations. I don't think it is likely, but you could have a situation where the export format is different in between cmakes that are compatible with 3.4.3 and those that have later compatibility.

spevans:
I didn't realise there would be those issues. But could we at least get the Ubuntu14 CI hosts installed with 3.5.2 as listed in the docs? - I have it on my Ubuntu14 test host and the full toolchain builds and tests ok without the need for the fix in this PR.

gottesmm:
@spevans two things:

1. I do think it **could** be reasonable to upgrade swift's as long as we put something in build-script to check this. Otherwise if we upgraded swift's you wouldn't get a failure until after llvm/clang had been built (which would be annoying).

2. If we do not do such a thing, I would be worried about upgrading ubuntu14 CI unless we had some other bot to catch for incompatibilities (i.e. one bot running with a 3.4.3 cmake).

Doing 1 would require a post to swift-dev and potential discussion I think.

gottesmm:
Btw, it looks like the macOS changes are related to some clang changes: https://ci.swift.org/job/oss-swift-incremental-RA-osx/5513/ and not to this PR. I am going to merge this to unbreak the 14.04 build.

spevans:
@gottesmm This PR https://github.com/apple/swift-corelibs-foundation/pull/1750 fixes the macOS build (and also building sclf when built in Xcode) but breaks linux - although I don't understand why tbh.

PR #1750, opened 2018-11-04
cmake: Revert adding 'framework' into module.modulemap

spevans:


spevans:
@swift-ci test

spevans:
@compnerd This seems to fix the macOS build (and also building corelibs-foundation in Xcode) but breaks the Linux build. Do you know the correct fix?

gottesmm:
@spevans I think that the macOS build is already broken for other reasons. Notice how the incremental bot started to fail here:

https://ci.swift.org/job/oss-swift-incremental-RA-osx/5513/

All clang commits and the thing you are hitting is the same.

PR #1751, opened 2018-11-04
build: compatibility for CMake 3.4.3

compnerd:
CMake 3.4.3 does not support the copy files... <directory> pattern.  Add
compatibility for it.


compnerd:
@swift-ci please test

gottesmm:
NOTE: The linux test here will be on 16.04 machine, so it will have a newer cmake. I am testing this locally to be sure this works on 14.04 with cmake-3.4.3.

gottesmm:
Got through to that weird XCTest that fails sometime. But we got through all of the foundation tests where we were failing.

gottesmm:
Reran skipping the earlier tests and XCTest passed. I think we need to increase the timeout on that test.

gottesmm:
Merging to unbreak the build.

PR #1754, opened 2018-11-06
CoreFoundation: use #include instead of #import

compnerd:
`#import` when building for Windows will be treated as the import of a
type library rather than the semantics of Objective-C style `#import`
which does an include once type of operation.  Convert to the `#include`
syntax instead to allow this to be built on Windows as well (e.g. via
clang-cl).

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -110,7 +110,7 @@ CF_EXTERN_C_BEGIN
 #include <pthread.h>
 
 #if !DEPLOYMENT_RUNTIME_SWIFT && __has_include(<os/log.h>)
-#import <os/log.h>
+#include <os/log.h>
 #else
 typedef struct os_log_s *os_log_t;
 #define os_log(...) do { } while (0)
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -1311,7 +1311,7 @@ CF_PRIVATE int asprintf(char **ret, const char *format, ...) {
 #endif
 
 #if DEPLOYMENT_RUNTIME_SWIFT
-#import <fcntl.h>
+#include <fcntl.h>
 
 extern void swift_retain(void *);
 extern void swift_release(void *);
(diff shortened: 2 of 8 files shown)


compnerd:
CC: @millenomi @phausler

compnerd:
@swift-ci please test

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -110,7 +110,7 @@ CF_EXTERN_C_BEGIN
 #include <pthread.h>
 
 #if !DEPLOYMENT_RUNTIME_SWIFT && __has_include(<os/log.h>)
-#import <os/log.h>
+#include <os/log.h>
```
phausler: Does windows have <os/log.h>?

CoreFoundation/String.subproj/CFAttributedString.c
```diff
@@ -17,7 +17,7 @@
 #include "CFRuntime_Internal.h"
 
 #if (TARGET_OS_MAC || TARGET_OS_WIN32) && DEPLOYMENT_RUNTIME_OBJC
-#import <Foundation/NSAttributedString.h>
+#include <Foundation/NSAttributedString.h>
```
phausler: This should be import

CoreFoundation/Base.subproj/CFInternal.h
compnerd: No, it doesn't, just going through an getting them all in a single go.

CoreFoundation/String.subproj/CFAttributedString.c
compnerd: The `TARGET_OS_WIN32` means that we will come down this path, which does mean that even clang will see that and object.

phausler: DEPLOYMENT_RUNTIME_OBJC means I wont for swift-corelibs-foundation.

compnerd: Ah, fair enough.  I'll revert this bit to an import then.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1745, opened 2018-11-02
[cmake] Fix capitalization in find_path argument.

drodriguez:
/cc @compnerd

PR #1756, opened 2018-11-08
Revert "[cmake] Find ICU in CoreFoundation."

milseman:
Reverts apple/swift-corelibs-foundation#1746

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ ExternalProject_Add(CoreFoundation
                       -DCF_ENABLE_LIBDISPATCH=${FOUNDATION_ENABLE_LIBDISPATCH}
                       -DCF_PATH_TO_LIBDISPATCH_SOURCE=${FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE}
                       -DCF_PATH_TO_LIBDISPATCH_BUILD=${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}
-                      -DICU_ROOT=${ICU_ROOT}
+                      -DICU_INCLUDE_DIR=${ICU_INCLUDE_DIR}
                     INSTALL_COMMAND
                       ${CMAKE_COMMAND} -E env --unset=DESTDIR ${CMAKE_COMMAND} --build . --target install)
 ExternalProject_Get_Property(CoreFoundation install_dir)
--- a/CoreFoundation/CMakeLists.txt
+++ b/CoreFoundation/CMakeLists.txt
@@ -1,8 +1,7 @@
 
 cmake_minimum_required(VERSION 3.4.3)
 list(APPEND CMAKE_MODULE_PATH
-     "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules"
-     "${CMAKE_CURRENT_SOURCE_DIR}/../cmake/modules")
+     "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
 project(CoreFoundation
         VERSION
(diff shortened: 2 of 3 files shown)


milseman:
Reverting because this seems to be the most likely culprit for https://ci.swift.org//job/oss-swift-incremental-RA-linux-ubuntu-14_04/6459/consoleFull#-3356114363122a513-f36a-4c87-8ed7-cbc36a1ec144

drodriguez:
I’m OK with reverting this. I asked Michael to do it. It will give me some time to setup a VM to check that I can reproduce the problem.

spevans:
@drodriguez The problem is that the `find_package(ICU..)` in CoreFoundation never finds the correct location and defaults to the system one with the header file in `/usr/include`.
 It works in the parent Foundation directory because `build-script-impl` passes overriding `-DICU_..` flags to `cmake` for `swift-corelibs-foundation`. I would remove the `find_package(ICU)` in CoreFoundation altogether and just let the parent pass the location of the include directory as that is all that needs to be correct anyway.

milseman:
@drodriguez @spevans I'm now seeing this on the package bot:

10:57:10 env: '/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-16_04/icu/icu4c/source/runConfigureICU': No such file or directory

Any ideas what could be going on? Is this related to the revert or something completely different?

https://ci.swift.org/job/oss-swift-package-linux-ubuntu-16_04//2577/console

drodriguez:
@milseman: for some reason, the script has checked out the first commit of the repository:

```
10:45:26  > git show-ref --tags -d # timeout=10
10:45:26 JENKINS-19022: warning: possible memory leak due to Git plugin usage; see: https://wiki.jenkins-ci.org/display/JENKINS/Remove+Git+Plugin+BuildsByBranch+BuildData
10:45:26 Multiple candidate revisions
10:45:26 Scheduling another build to catch up with OSS - Swift Package - Ubuntu 16.04 (master)
10:45:26 Checking out Revision 1568e2d239301084ab451c597c6c7c5c2effab3e (refs/tags/icu4j-release-3-0-d02)
10:45:26  > git config core.sparsecheckout # timeout=10
10:45:26  > git checkout -f 1568e2d239301084ab451c597c6c7c5c2effab3e
10:45:27 Commit message: "This commit was manufactured by cvs2svn to create tag 'release-3-0-d02'."
```

I don’t know if the Jenkins message is the one causing the problem. I can look at the checkout script later if you want, but probably someone else can find the problem faster than me.

spevans:
@milseman I don't think the ICU branch has been added to the swift-5 section of `utils/update_checkout/update-checkout-config.json`. Could that be causing issues?

`` "swift-5.0-branch" : {
            "aliases": ["swift-5.0-branch"],
            "repos": {
                "llvm": "swift-5.0-branch",
                "clang": "swift-5.0-branch",
                "compiler-rt": "swift-5.0-branch",
                "swift": "swift-5.0-branch",
                "lldb": "swift-5.0-branch",
                "cmark": "master",
                "llbuild": "master",
                "swiftpm": "master",
                "swift-syntax": "master",
                "swift-corelibs-xctest": "master",
                "swift-corelibs-foundation": "master",
                "swift-corelibs-libdispatch": "master",
                "swift-integration-tests": "master",
                "swift-xcode-playground-support": "master",
                "ninja": "release"
            }
        }

```

drodriguez:
@spevans: I think I found the problem. The patch that this reverted is OK. It forwards `ICU_ROOT`, which is one of the variables passed from the build script, and it should be the only variable that  `find_package` needs (but we can forward all of them if you prefer).

The problem, for me at least, is the typo in #1745. When I fix that, the script works only with `ICU_ROOT`. Probably the top level will simply need `ICU_ROOT` if we fix that typo, but works because all the variables are specified anyway.

I will try to figure out what’s going on with #1745. Then I will try to redo the patch that was reverted with this lessons applied.

Sorry for all the problems.

PR #1758, opened 2018-11-09
Return Self instead of UnitLength in UnitLength override method

tkremenek:
Fixes “cannot override a Self return type with a non-Self return type” error in Swift 5 compiler.


tkremenek:
@swift-ci test

slavapestov:
@tkremenek This changed because the default is now -swift-version 5 right? Should we just be passing -swift-version 4 instead?

millenomi:
We want to be on the latest version of the language as much as possible.

millenomi:
@swift-ci please test

tkremenek:
I'm going to try and address this problem when moving to Swift 5 with a different approach.

PR #1759, opened 2018-11-09
Decimal: Fix .doubleValue to use _length to only index valid mantissa elements.

spevans:
Fix for https://forums.swift.org/t/severe-decimal-error-in-foundation-on-linux-needs-workaround/14359/6


spevans:
@swift-ci test linux

PR #1760, opened 2018-11-12
docs: update for the CMake migration

compnerd:
Some of the commands for building s-c-f and running tests on Linux have
changed.  Update the documentation to reflect this.

### DIFF
--- a/Docs/GettingStarted.md
+++ b/Docs/GettingStarted.md
@@ -49,14 +49,13 @@ This will build and run the Foundation tests, in the Debug configuration.
 After the complete Swift build has finished, you can iterate quickly on Foundation itself by simply invoking `ninja` in the Foundation directory.
 
 ```
-cd swift-corelibs-foundation
-ninja
+ninja -C build/Ninja-DebugAsserts/swift-corelibs-foundation Foundation
 ```
 
 This will build Foundation. To build and run the tests, use the `test` target:
 
 ```
-ninja test
+ninja -C build/Ninja-DebugAsserts/swift-corelibs-foundation test
 ```
 
 The ninja build script will print the correct command-line invocation for both running the tests and debugging the tests. The exact library path to use will depend on how Foundation itself was configured by the earlier `build-script`. For example:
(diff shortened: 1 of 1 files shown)


Docs/GettingStarted.md
```diff
@@ -49,14 +49,13 @@ This will build and run the Foundation tests, in the Debug configuration.
...
-cd swift-corelibs-foundation
-ninja
+ninja -C build/Ninja-DebugAsserts/swift-corelibs-foundation Foundation
 ```
 
 This will build Foundation. To build and run the tests, use the `test` target:
 
 ```
-ninja test
+ninja -C build/Ninja-DebugAsserts/swift-corelibs-foundation test
```
spevans: That doesn't seem to work:

```$ ninja -C build/Ninja-ReleaseAssert/foundation-linux-x86_64  test
ninja: Entering directory `build/Ninja-ReleaseAssert/foundation-linux-x86_64'
[0/1] Running tests...
Test project /home/spse/swift-source/build/Ninja-ReleaseAssert/foundation-linux-x86_64
No tests were found!!!
```

Docs/GettingStarted.md
```diff
@@ -78,7 +77,7 @@ Just copy & paste the correct line.
 When new source files or flags are added to the `build.py` script, the project will need to be reconfigured in order for the build system to pick them up. The top-level `swift/utils/build-script` can be used, but for quicker iteration you can use the following command to limit the reconfiguration to just the Foundation project:
```
millenomi: We should remove references to build.py.

Docs/GettingStarted.md
compnerd: @spevans - did you enable the tests?  `grep ENABLE_TESTING build/Ninja-ReleaseAssert/foundation-linux-x86_64/CMakeCache.txt`

millenomi: This _DEFINITELY_ must make into the documentation if it isn't there. It's not obvious.

spevans:
```
$ grep ENABLE_TESTING build/Ninja-ReleaseAssert/foundation-linux-x86_64/CMakeCache.txt 
ENABLE_TESTING:BOOL=NO
```

How are the tests enabled? Is it only if `--test-foundation` is passed to `build-script-impl`?
I don't really know much about CMake but could the tests always be enabled by `build-script-impl` even it `--test-xxx` flags aren't passed? I take it actually running the tests is still a separate step in the build script?

ddunbar: I'm confused, why are tests disabled by default?

millenomi: because testing enables support for `@testable import`, which we don't want to have on in the release build.

compnerd:
I believe that @spevans already updated the docs.

PR #1761, opened 2018-11-13
Mark subclasses of Dimension final.

tkremenek:
Possible resolution of blocking adoption of `-swift-version 5`:

https://github.com/apple/swift/pull/20374

### DIFF
--- a/Foundation/Unit.swift
+++ b/Foundation/Unit.swift
@@ -248,7 +248,7 @@ open class Dimension : Unit {
     }
 }
 
-open class UnitAcceleration : Dimension {
+final class UnitAcceleration : Dimension {
     
     /*
      Base unit - metersPerSecondSquared
@@ -297,7 +297,7 @@ open class UnitAcceleration : Dimension {
     }
 }
 
-open class UnitAngle : Dimension {
+final class UnitAngle : Dimension {
     
     /*
      Base unit - degrees
@@ -378,7 +378,7 @@ open class UnitAngle : Dimension {
     }
 }
 
-open class UnitArea : Dimension {
+final class UnitArea : Dimension {
     
     /*
      Base unit - squareMeters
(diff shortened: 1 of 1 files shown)


tkremenek:
@swift-ci test

Foundation/Unit.swift
```diff
@@ -248,7 +248,7 @@ open class Dimension : Unit {
     }
 }
 
-open class UnitAcceleration : Dimension {
+final class UnitAcceleration : Dimension {
```
millenomi: I don't think we can take this patch? These subclasses are open in ObjC, and we want as-close-as-possible source parity.

millenomi: (I assume we're trying to work around the fact that Swift 5 prevents us from having a pure Swift class that has the equivalent of `+ (instancetype)baseUnit`?)

tkremenek: At this point I'm testing the option for consideration.  If we took it there are other things to consider, such as parity with Objective-C and expectations for users of the APIs.

millenomi: IIRC, subclassing is part of the contract of this API. (I will check tomorrow.)

I have a patch I can PR that moves .baseUnit() to be of Dimension type everywhere, erasing the type but allowing the build to work. I will discuss it with @parkera tomorrow morning.

tkremenek:
@swift-ci test

Foundation/Unit.swift
millenomi: Yeah, and Swift is just highlighting it. :(

I'll circle back before EOD.

millenomi: (It feels like Swift could allow `Self` class methods if they were similar to `required` initializers, wherein it is not valid to express a class that doesn't override the method.)

tkremenek: Yes, that sounds like potentially a reasonable language extension that does not exist today.

parkera: For now, to unblock Swift 5 adoption, we will take this patch.

millenomi:
We had a side discussion between me, @tkremenek and @parkera and decided that the right way forward is to require subclasses to be made `final`.

millenomi:

PR #1763, opened 2018-11-13
Remove deprecated (now an error) conversion to Range.

tkremenek:
Needed to build Foundation with Swift 5.


tkremenek:
@swift-ci test

PR #1765, opened 2018-11-15
Mark open properties in final class public.

tkremenek:
Swift 5 mode is more strict; treating this as an error.


tkremenek:
@swift-ci test

PR #1769, opened 2018-11-15
SR-9251: Add $ORIGIN into libFoundation.so RUNPATH

spevans:
`libFoundation.so` does not have `$ORIGIN` in its `RUNPATH` so it cant find the ICU libraries installed in `usr/lib/swift/linux`

With this PR the path now looks like : 
```
$ readelf -aW ~/swift-install/usr/lib/swift/linux/libFoundation.so |grep RUNPATH
0x000000000000001d (RUNPATH)            Library runpath: [/home/spse/swift-source/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/spse/swift-source/build/buildbot_linux/libdispatch-linux-x86_64/src:$ORIGIN]
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -241,6 +241,7 @@ add_swift_library(Foundation
                     ${LIBXML2_LIBRARIES}
                     ${libdispatch_ldflags}
                     ${uuid_LIBRARIES}
+		    -Xlinker;-rpath;-Xlinker;"\\\$\$ORIGIN"
                   SWIFT_FLAGS
                     -DDEPLOYMENT_RUNTIME_SWIFT
                     ${deployment_enable_libdispatch}


spevans:
@swift-ci test

benlangmuir:
@spevans what do you think of adding a canary to https://github.com/apple/swift-integration-tests for running a simple file that imports Foundation?

eeckstein:
Can we merge this?

spevans:
@benlangmuir I had a look and there is actually a test for Foundation already in `swift-integration-tests` however it passes because the build directory is currently in the RUNPATH. I will do a 2nd PR to remove this but it will require some CMake changes to setup `LD_LIBRARY_PATH` for the `TestFoundation` tests to run so will be a while.

spevans:
cc @millenomi

eeckstein:
ok, thanks. I'll merge it now to unblock ci

PR #1770, opened 2018-11-15
Use forced ‘cast’ in NSNumber convenience initalizer.

tkremenek:
Needed to build in Swift 5.  This enforces an invariant using a
runtime check.


tkremenek:
@swift-ci test linux

PR #1774, opened 2018-11-16
Build Foundation tests in Swift 4 mode to work around @autoclosure regression

tkremenek:
There is a regression in `-swift-version 5` mode when forwarding `@autoclosure` closures from
one function to another.  Temporarily build `TestFoundation` in `-swift-version 4` mode to unblock
bumping the compiler's default to `-swift-version 5`.  This should be removed later.


tkremenek:
Thanks to @slavapestov and @rjmccall for pointing out this is not a regression.  Backing this PR out and creating a new patch.

PR #1776, opened 2018-11-16
Remove extra installed copy of libFoundation.so

spevans:
- Currently libFoundation.so exists in 2 places in the distribution
  tarfile:

  ./usr/lib/swift/linux/x86_64/libFoundation.so
  ./usr/lib/swift/linux/libFoundation.so

- ld.so will load the first one located in x86_64 subdirectory which
  will then set $ORIGIN to this directory. Consequently, the libICU
  shared libraries required by libFoundation.so will not be found.

- Update CMakeLists.txt to not install libFoundation.so in the x86_64
  directory.

- Also remove an unneeded RUNPATH pointing to the build directory for
  libdispatch. This is already reached using LD_LIBRARY_PATH during
  testing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,9 +59,6 @@ if(FOUNDATION_ENABLE_LIBDISPATCH)
   set(deployment_enable_libdispatch -DDEPLOYMENT_ENABLE_LIBDISPATCH)
   set(libdispatch_cflags -I;${FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE};-I;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src/swift;-Xcc;-fblocks)
   set(libdispatch_ldflags -L;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD};-L;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src;-ldispatch)
-  if(CMAKE_SYSTEM_NAME STREQUAL Linux OR CMAKE_SYSTEM_NAME STREQUAL Android OR CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
-    list(APPEND libdispatch_ldflags -Xlinker;-rpath;-Xlinker;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src)
-  endif()
 endif()
 
 add_swift_library(Foundation
@@ -465,12 +462,6 @@ install(FILES
           CoreFoundation/Base.subproj/module.map
         DESTINATION
           lib/swift/CoreFoundation)
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
-  install(FILES
-            ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}Foundation${CMAKE_SHARED_LIBRARY_SUFFIX}
-          DESTINATION
-            lib/swift/${swift_os}/${swift_arch})
-endif()
 install(FILES
           ${CMAKE_CURRENT_BINARY_DIR}/plutil
         DESTINATION


spevans:
@swift-ci test

spevans:
```
$ find swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/ -name libFoundation.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/x86_64/libFoundation.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so
$ ./hello
./hello: error while loading shared libraries: libicuucswift.so.61: cannot open shared object file: No such file or directory
$ rm swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/x86_64/libFoundation.so
$ ./hello
Hello
```

Note that even though the `RUNPATH` for `libswiftcore.so` specifies `$ORIGIN` which points to `swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux`, `ld.so` will try several subdirectories automatically, including `tls/x86_64` and `x86_64`. For full details try running:
`LD_DEBUG=libs ./hello 2>&1|less`

alblue:
Wasn't the plan on this to put all the libraries under the x86_64 directory instead? Or to put it another way, why isn't the correct fix to stop putting the libFoundation at the top level, and move the ICU libraries to under the x86_64 directory?

spevans:
Looking at the current layout:
```
$ find swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/ -name 'lib*.so'
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/x86_64/libFoundation.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libicudataswift.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libicui18nswift.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libswiftSwiftOnoneSupport.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libswiftCore.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libBlocksRuntime.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libXCTest.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libdispatch.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libicuucswift.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libswiftDispatch.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libswiftRemoteMirror.so
swift-DEVELOPMENT-SNAPSHOT-2018-11-15-a-ubuntu18.04/usr/lib/swift/linux/libswiftGlibc.so
```

I think the only thing that is important is that all of the libraries are in either the `x86_64` subdirectory or not. Its only a problem if it appears in both.

alblue:
I concur – having them in only one place is the right solution. But I thought the re-org had happened; maybe it was rolled back in the Swift repo but not here? ISTR @compnerd was doing some refactoring in this area, so wanted to get visibility on what the right fix was.

spevans:
I think the only issue will be that some changes will be required in the main swift repository to move the `libicu` libraries into the per arch subdirectory along with `libswiftCore` etc and only critical PRs are being taken for `swift`'s `swift-5.0-branch` so I think we need to merge this PR as is (since it will get auto merged to `swift-corelibs-foundation`'s `swift-5.0-branch`) and then move everything into the subdirectory as a post 5.0 change. Does that make sense?

spevans:
@compnerd I tried a test removing the version in the os directory but it failed because the linker commands used by `swiftc` when building binaries assume all of the libraries are in  the os directory not the arch subdirectory. So moving everything into the arch directory will have to be a post 5.0 change, this change is just to get the current swift-5.0-branch and master working.

```
$ ~/swift-test/usr/bin/swiftc -v  ~/src/hello.swift 
Swift version 5.0-dev (LLVM dcb9eb74a7, Clang 95cdf7c9af, Swift 89f9ce839d)
Target: x86_64-unknown-linux-gnu
/home/spse/swift-test/usr/bin/swift -frontend -c -primary-file /home/spse/src/hello.swift -target x86_64-unknown-linux-gnu -disable-objc-interop -color-diagnostics -module-name hello -o /tmp/hello-751723.o
/home/spse/swift-test/usr/bin/swift-autolink-extract /tmp/hello-751723.o -o /tmp/hello-d88c13.autolink
/usr/bin/clang++ -fuse-ld=gold -pie -target x86_64-unknown-linux-gnu -Xlinker -rpath -Xlinker /home/spse/swift-test/usr/lib/swift/linux /home/spse/swift-test/usr/lib/swift/linux/x86_64/swiftrt.o /tmp/hello-751723.o @/tmp/hello-d88c13.autolink -L /home/spse/swift-test/usr/lib/swift/linux -lswiftCore --target=x86_64-unknown-linux-gnu -v -o hello
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/7
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/8
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/7
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/7.3.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/8
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0
Candidate multilib: .;@m64
Selected multilib: .;@m64
 "/usr/bin/ld.gold" -pie -z relro --hash-style=gnu --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o hello /usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/../../../x86_64-linux-gnu/Scrt1.o /usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/../../../x86_64-linux-gnu/crti.o /usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/crtbeginS.o -L/home/spse/swift-test/usr/lib/swift/linux -L/usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0 -L/usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/../../../x86_64-linux-gnu -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/../../.. -L/usr/lib/llvm-6.0/bin/../lib -L/lib -L/usr/lib -rpath /home/spse/swift-test/usr/lib/swift/linux /home/spse/swift-test/usr/lib/swift/linux/x86_64/swiftrt.o /tmp/hello-751723.o -lFoundation -lswiftCore -lswiftGlibc -lpthread -lutil -ldl -lm -lswiftDispatch -ldispatch -lBlocksRuntime -lswiftSwiftOnoneSupport -lswiftCore -lstdc++ -lm -lgcc_s -lgcc -lc -lgcc_s -lgcc /usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/crtendS.o /usr/bin/../lib/gcc/x86_64-linux-gnu/7.3.0/../../../x86_64-linux-gnu/crtn.o
```
Note the  `-L/home/spse/swift-test/usr/lib/swift/linux` and `-rpath /home/spse/swift-test/usr/lib/swift/linux` arguments to `ld.gold`

compnerd:
Hmm, in that case, why not commit directly into just the swift-5.0-branch rather than in master, and go ahead and start making the other changes for master?  @drodriguez was working on this which is why I haven't been pushing on it ... there were a couple of patches that still need to be merged and need to be coordinated across repositories.

compnerd:
Oh, but, the change to the `DT_RUNPATH` ... ❤️. The fewer the entries the happier I am.  Let's get that merged on both.

compnerd:
I really would like to try to see if we can get the subdirectory move in the 5.0 timeframe.  Because we are installing the runtime libraries into the toolchain (why are we doing that btw?), the RPATHs will have to change.  It seems that it would be nicer to have the paths corrected for the 5.0 release.  I believe that all the patches for this are already up, its a matter of coordinating and merging them.  The problem is that getting the build changes through tends to be very difficult (I've had build changes get reverted multiple times for various reasons, including Apple internal merge conflicts).  I don't see a good way to ensure that we can get such changes in quickly because of the wonkiness in the build system that has spread out through all the repositories (primarily due to build-script and swift's resistance to using CMake's support for cross-compiling and installation).

spevans:
I don't think you will get anymore changes into the main swift repo for swift-5 as I believe only essential patches are now being taken for the main repo and I believe at least 2 changes are required for that:
1. Moving the libs into the subdirectory
2. Modifying `getRuntimeLibraryPath` https://github.com/apple/swift/blob/a820992cb4741b333a3e9c98746d2a4f3e8c4d0b/lib/Driver/ToolChains.cpp#L1051 to point to the new subdirectory.

So I think it will end up being a Swift6 change unfortunately.

compnerd:
We really do need to figure out how to get more of these changes which potentially may break Apple's builds in more easily.  In this particular case, the tree structure is not suitable for non-Darwin targets :-(.

spevans:
Isn't it only an issue when we want to ship multi architecture toolchains? How is it not suitable for non -Darwin targets at the moment ?

spevans:
@swift-ci test and merge

drodriguez:
It is unfortunate that this change was merged, as it has undone one of the preparation steps to merge the changes in the compiler. The full process, was adding this duplication workaround, merge the compiler changes and remove the workarounds. Sadly the compiler change is talking a long time to be reviewed and merged, so this was catch in the middle.

As @compnerd explains, it is impossible to target with the same toolchain two architectures for the same os, except you use the special support for fat libraries from Mach-O. That means that it is not possible to have a toolchain building for two Android architectures, two Linux architectures, etc. The installation process will copy both architectures into one destination with the same name, which fails. The changes in the compiler will allow to look into the right architecture subdirectory instead, not needing to do any copy (except that Darwin keeps working as always). It is also very strange that libraries that are not exactly related to the compiler are installed into the compiler files

I’m really sorry I didn’t see this PR before, and I also sorry I didn’t realize that ICU was being installed in the OS directory instead of the architecture. I would have provided the same workaround until the compiler changes were merged. We will try to figure out if there’s a way of doing this in a simple way without too many inter-repo dependencies, but it would be awesome if this is not limited to “after 5.0”, since this only affects platforms that are not Darwin (we were very careful about it) and it makes distributing toolchains outside of Darwin a really complicated process.

PR #1779, opened 2018-11-18
Windows Port

compnerd:
This set of changes allows building for windows.  There are some implementations and some things which can be cleaned up, but, this at least identifies the areas that we need to fix up.


compnerd:
Superseded by #1783.  The patches have been formalized and extended.  They can still deal with a bit of clean up, but I think that there is no longer any value in this reference work.

PR #1783, opened 2018-11-23
CoreFoundation Windows Port

compnerd:


compnerd:
CC: @phausler @parkera 

Note that this is not fully complete, I'm still working through the patches.  But, I needed a way to track the patch set.  Most of the changes are pretty small and very targeted at individual classes of issues, so this is going to look much worse than it really is.  It is one of those death by a thousand paper cuts scenario.

compnerd:
CC: @millenomi

CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -390,6 +393,14 @@ CF_INLINE void __CFPortSetFree(__CFPortSet portSet) {
 typedef HANDLE __CFPort;
 #define CFPORT_NULL NULL
 
+#ifndef __unused
```
phausler: this should probably be at a header like CFInternal.h

CoreFoundation/AppServices.subproj/CFUserNotification.c
```diff
@@ -17,13 +17,18 @@
...
 #include <CoreFoundation/CFMachPort.h>
 #include <stdlib.h>
-#include <unistd.h>
 #include <stdio.h>
 #include <mach/mach.h>
 #include <mach/error.h>
 #include <limits.h>
 #include <errno.h>
+
+#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
```
millenomi: We use `__has_include` elsewhere for this.

CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -446,17 +446,10 @@ CF_EXPORT double kCFCoreFoundationVersionNumber;
...
-typedef unsigned long long CFHashCode;
-typedef signed long long CFIndex;
-#else
-typedef unsigned long CFTypeID;
-typedef unsigned long CFOptionFlags;
-typedef unsigned long CFHashCode;
-typedef signed long CFIndex;
-#endif
+typedef uintptr_t CFTypeID;
+typedef uintptr_t CFOptionFlags;
```
millenomi: This affects Darwin.

@parkera @phausler Do you see any scenario in which this wouldn’t be strictly equivalent?

CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -411,7 +422,7 @@ static __CFPortSet __CFPortSetAllocate(void) {
     result->used = 0;
     result->size = 4;
     result->handles = (HANDLE *)CFAllocatorAllocate(kCFAllocatorSystemDefault, result->size * sizeof(HANDLE), 0);
-    CF_SPINLOCK_INIT_FOR_STRUCTS(result->lock);
+    CF_LOCK_INIT_FOR_STRUCTS(result->lock);
```
phausler: I am not sure we can change this lock type here btw.

CoreFoundation/Base.subproj/CFFileUtilities.c
```diff
@@ -1023,7 +1024,7 @@ CF_PRIVATE void _CFIterateDirectory(CFStringRef directoryPath, Boolean appendSla
     if (!CFStringGetFileSystemRepresentation(directoryPath, directoryPathBuf, CFMaxPathSize)) return;
     
 #if DEPLOYMENT_TARGET_WINDOWS
-#error this path does not support calculateFullResultPath but it must do so someday
+#warning this path does not support calculateFullResultPath but it must do so someday
```
millenomi: We need a SR for this.

CoreFoundation/AppServices.subproj/CFUserNotification.c
millenomi: Other instances below as well.

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -517,53 +519,46 @@ typedef pthread_mutex_t CFLock_t;
...
 
-CF_INLINE Boolean __CFLockTry(volatile CFLock_t *lock) {
-    return (InterlockedCompareExchange((LONG volatile *)lock, ~0, 0) == 0);
-}
+typedef int32_t OSSpinLock;
+#define OS_SPINLOCK_INIT 0
+#define OSSpinLockLock(lock) ({ while (__sync_val_compare_and_swap(lock, 0, ~0) != 0) sleep(0); })
+#define OSSpinLockUnlock(lock) ({ __sync_synchronize(); *lock = 0; })
 
-#elif DEPLOYMENT_TARGET_LINUX || DEPLOYMENT_TARGET_FREEBSD
```
millenomi: Why remove this support?

CoreFoundation/String.subproj/CFString.h
```diff
@@ -147,9 +147,6 @@ equivalents in the encoding the compiler will use to interpret them (for instanc
 O-umlaut is \303\226 in UTF-8). UTF-8 is the recommended encoding here, 
 since it is the default choice with Mac OS X developer tools.
 */
-#if TARGET_OS_WIN32
-#undef __CONSTANT_CFSTRINGS__
```
phausler: im sure this might require compiler changes

CoreFoundation/Base.subproj/CFKnownLocations.c
```diff
@@ -72,6 +72,25 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
         }
     }
     
+#elif DEPLOYMENT_TARGET_WINDOWS
+
+    switch (user) {
+        case _kCFKnownLocationUserAny:
+            location = CFURLCreateWithFileSystemPath(kCFAllocatorSystemDefault, CFSTR("\\Users\\All Users\\AppData\\Local"), kCFURLWindowsPathStyle, true);
```
millenomi: Isn’t there API to get this path, and the user path below?

CoreFoundation/Base.subproj/CFBase.h
phausler: are we in any danger with C++? or swift imports here?

This is a public API change so lets be super careful here.

CoreFoundation/Base.subproj/CFPriv.h
```diff
@@ -90,7 +90,7 @@ CF_EXPORT void CFPreferencesFlushCaches(void);
 CF_EXPORT Boolean _CFURLGetWideFileSystemRepresentation(CFURLRef url, Boolean resolveAgainstBase, wchar_t *buffer, CFIndex bufferLength);
 #endif
 
-#if !__LP64__
+#if !__LP64__ && !__LLP64__
```
millenomi: Given the prevalence o this, I’m beginning to suspect a _CF_64_BIT macro may not be a bad idea.

CoreFoundation/Locale.subproj/CFCalendar_Internal.h
```diff
@@ -93,10 +93,10 @@ CF_PRIVATE void __CFCalendarZapCal(CFCalendarRef calendar);
 
 CF_PRIVATE CFCalendarRef _CFCalendarCreateCopy(CFAllocatorRef allocator, CFCalendarRef calendar);
 
-CF_PRIVATE Boolean _CFCalendarComposeAbsoluteTimeV(CFCalendarRef calendar, /* out */ CFAbsoluteTime *atp, const char *componentDesc, int32_t *vector, int32_t count);
-CF_PRIVATE Boolean _CFCalendarDecomposeAbsoluteTimeV(CFCalendarRef calendar, CFAbsoluteTime at, const char *componentDesc, int32_t *_Nonnull * _Nonnull vector, int32_t count);
-CF_PRIVATE Boolean _CFCalendarAddComponentsV(CFCalendarRef calendar, /* inout */ CFAbsoluteTime *atp, CFOptionFlags options, const char *componentDesc, int32_t *vector, int32_t count);
-CF_PRIVATE Boolean _CFCalendarGetComponentDifferenceV(CFCalendarRef calendar, CFAbsoluteTime startingAT, CFAbsoluteTime resultAT, CFOptionFlags options, const char *componentDesc, int32_t *_Nonnull * _Nonnull vector, int32_t count);
+CF_EXPORT Boolean _CFCalendarComposeAbsoluteTimeV(CFCalendarRef calendar, /* out */ CFAbsoluteTime *atp, const char *componentDesc, int32_t *vector, int32_t count);
```
millenomi: Why this?

I can’t export these symbols on Darwin. If it is incorrect that thy be private, they should be `CF_CROSS_PLATFORM_EXPORT`.

millenomi: (Is there any way to achieve parity such that s-c-f can use these symbols on Windows as it can on Linux?)

CoreFoundation/NumberDate.subproj/CFDate.c
```diff
@@ -104,6 +116,9 @@ CF_EXPORT CFTimeInterval CFGetSystemUptime(void) {
         HALT;
     }
     return (double)res.tv_sec + ((double)res.tv_nsec)/1.0E9;
+#elif DEPLOYMENT_TARGET_WINDOWS
+    ULONGLONG ullTickCount = GetTickCount64();
+    return ullTickCount / 1000;
```
millenomi: Should this be floating-point division?

CoreFoundation/PlugIn.subproj/CFBundle_Executable.c
```diff
@@ -119,7 +119,7 @@ static CFURLRef _CFBundleCopyExecutableURLRaw(CFURLRef urlPath, CFStringRef exeN
 #elif DEPLOYMENT_TARGET_WINDOWS
     if (!executableURL) {
         executableURL = CFURLCreateWithFileSystemPathRelativeToBase(kCFAllocatorSystemDefault, exeName, kCFURLWindowsPathStyle, false, urlPath);
-        if (executableURL && !_urlExists(executableURL)) {
+        if (executableURL && !_binaryLoadable(executableURL)) {
```
millenomi: I need to double-check that this matches Darwin behavior. Wednesday.

millenomi:
@swift-ci please test

CoreFoundation/Base.subproj/CFBase.h
compnerd: I suppose that we could just map these specifically in the clang-importer.  I think that we should talk to @jrose-apple about that.

I suppose, that this *could* cause issue where there are template specializations using `CFTypeID`, `CFOptionsFlags`.

CoreFoundation/Base.subproj/CFFileUtilities.c
compnerd: I would love to do that.  I am afraid that I do not fully understand what the warning is about.  I would have just implemented it otherwise.

CoreFoundation/AppServices.subproj/CFUserNotification.c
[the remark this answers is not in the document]
compnerd: Yeah, I can switch to `__has_include`.  I really wish that Apple would define `__unix__` as well.  It really is a unix header.

CoreFoundation/Base.subproj/CFAsmMacros.h
```diff
@@ -16,7 +16,7 @@
 
 /* Use .note.GNU-stack to explicitly indicate a non-exec stack, b/c of bad   */
 /* default behaviour when translating handwritten assembly files (needed on  */
-/* GNU/* platforms, Android and FreeBSD, as tests have shown).               */
+/* GNU/.* platforms, Android and FreeBSD, as tests have shown).               */
```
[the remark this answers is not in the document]
compnerd: SGTM

CoreFoundation/String.subproj/CFString.h
compnerd: I took care of those changes a while ago.  constant CFStrings are generated properly on Windows by clang.  Unfortunately, gcc and cl are not viable options due to the ABI level changes for supporting the CFString to swift.String bridging.

CoreFoundation/Base.subproj/CFKnownLocations.c
compnerd: No, the path is just "\Users\All Users\AppData\Local" ... the last component is a decision (Local vs LocalLow vs Roaming).  The user name substitution is usually a `_snprintf` type of thing

CoreFoundation/Base.subproj/CFPriv.h
compnerd: Hehe, yes, I really did want that previously too :-).

CoreFoundation/Locale.subproj/CFCalendar_Internal.h
compnerd: These are mismatches between the declarations in the different headers.  Yes, the attempt to get the parity was not well received (the hidden visibility build would do that).  I think that we can do that change separately.

CoreFoundation/NumberDate.subproj/CFDate.c
compnerd: No, its integral, the tick count is an integral value.

CoreFoundation/Base.subproj/CFPriv.h
jrose-apple: Maybe switch to `__POINTER_WIDTH__` like the Swift runtime?

CoreFoundation/Base.subproj/CFInternal.h
compnerd: Its not really removing support, just merging codepaths to just use the posix threads like on Darwin.  However, Philippe points out that it may be preferable to make them futexes on Linux.

CoreFoundation/Base.subproj/CFPriv.h
parkera: I think that would work. There are probably still places where `LP64` means a few different things (e.g., historically some have used it to indicate presence of modern runtime, incorrectly, on Darwin), but I've tried to fix those up as I find them.

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -307,29 +304,21 @@ CF_EXPORT int _NS_access(const char *name, int amode);
 #define WIN32_LEAN_AND_MEAN
 
 #ifndef WINVER
-#define WINVER  0x0501
+#define WINVER  0x0601
```
parkera: This is basically defining the min deployment target for Windows, correct?

CoreFoundation/Base.subproj/CFPriv.h
parkera: Or, I guess, `TARGET_RT_64_BIT` from `TargetConditionals.h` below, which @compnerd fixed up as well.

CoreFoundation/Locale.subproj/CFCalendar_Internal.h
parkera: I think we may have needed these in the ObjC `NSCalendar.m`, so the cross-platform export would make sense here.

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
jakepetroules: Not really, this is more like the SDK version, which automatically implies the "deployment target" is not higher than that.

You could set WINVER to 0x0603 (Windows 8.1) for example, and then use LoadLibrary/GetProcAddress (dlopen/dlsym), to conditionally use APIs introduced in Windows 8.1, so that your app still works when running on Windows 7 (0x0601). If you have WINVER at 0x0601, then the definitions of Windows 8.1 APIs would not be visible to you in the SDK headers.

0x0601 seems like a good value here (means: "allows statically binding to APIs introduced no later than Windows 7"), I can't imagine anyone is interested in deploying Swift code to anything less than Windows 7 at this point in time.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -117,7 +122,7 @@ static CFMutableArrayRef __CFCopyWindowsTimeZoneList() {
     TCHAR lpName[MAX_PATH+1];
     DWORD dwIndex, retCode;
 
-    if (RegOpenKey(HKEY_LOCAL_MACHINE,_T(TZZONEINFO),&hkResult) !=
+    if (RegOpenKey(HKEY_LOCAL_MACHINE, _T("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Time Zones"), &hkResult) !=
```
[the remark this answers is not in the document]
compnerd: I couldn't really get it to work with a macro.  It's really no worse than with the macro for `TZZONEINFO`.

CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -689,13 +688,21 @@ struct __CFRunLoopMode {
 };
 
 CF_INLINE void __CFRunLoopModeLock(CFRunLoopModeRef rlm) {
-    pthread_mutex_lock(&(rlm->_lock));
+#if DEPLOYMENT_TARGET_WINDOWS
+    EnterCriticalSection(&(rlm->_lock));
+#else
+    __CFLock(&(rlm->_lock));
```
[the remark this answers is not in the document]
compnerd: The `CFLock_t` is always a `pthread_t` on Darwin.  That's what I had changed Linux to, but seems that it was meant to be a lighter weight primitive like futex on Linux.

compnerd:
I realize that this work is mostly useful only when available in aggregate, however, I think that it would help reduce some of the cognitive overhead and management burden to pick out the patches that are ready to go and merge those in.  Is that okay with you @parkera, @millenomi?

parkera:
Yes I think so.

CoreFoundation/Locale.subproj/CFCalendar_Internal.h
compnerd: Splintered into #1800 with the change to `CF_CROSS_PLATFORM_EXPORT`.

CoreFoundation/Base.subproj/CFFileUtilities.c
compnerd: @millenomi - could you please explain what `calculateFullResultPath` actually means?

compnerd:
@phausler, @parkera - so, for the locking changes, what do you guys think of introducing a new type alias that we can use to differentiate between what needs to be a fully mutex (`pthread_mutex_t`) and what can be an unfair lock (`CFLock_t`)?  Would `CFRecursiveMutex` and `CFMutex` be reasonable to have?  That would allow most of this work to get merged.  The patches that remain are mostly because I am afraid I don't know how to best address to them.  Those are explicitly called out with a `HACK/` in the commit message.

parkera:
As far as I know we never use an actually recursive lock - just the `pthread_lock_t` vs a spin lock. Is `CFSPINLOCK` actually a spin lock anymore (on Darwin)? We've gone back and forth recently on this, and you're probably right that the macro name is now a misnomer.

phausler:
Really the two lock types are fair and unfair. pthread_mutex is fair, others are not. I am not convinced we really ever need fair locks even in Darwin but that is something we need to confirm to change. The unfair case for Darwin donates QoS via os_unfair_lock and Linux i don’t think really has QoS anyhow so a lightweight lock is fine (most Linuxes pthread_mutex is lighter than Darwin anyhow). So in the end the two concepts are heavy but fair lock and lightweight lock.

compnerd:
@parkera, @phausler - there are definitely cases of recursive mutexes in the code base.  Thats why I ended up using `CRITICAL_SECTION`s.  They are used in the bundle loader.  I don't see any use of `OSSpinLock` anywhere, so I do suspect that there is no use of spin locks on Darwin.

The problem is that the locking mechanism on Windows doesn't have posix locking, so we need a different set of locks.  For lightweight mutexes that do not need to recursive, the Slim Reader/Writer locks work well.  For recursive locks, we need to resort to `CRITICAL_SECTION`.  The type aliasing should enable us to do that.  Unifying and sharing the codepaths makes it easier to restructure the code and

CoreFoundation/PlugIn.subproj/CFBundle_Executable.c
compnerd: @millenomi - ping on this item!  If this matches the Darwin behavior, I'd like to get this patch merged as well.

CoreFoundation/Base.subproj/CFFileUtilities.c
compnerd: Spoke offline, this is vestigial, and can be safely removed.

CoreFoundation/PlugIn.subproj/CFBundle_Executable.c
compnerd: Spoke offline, this is in fact correct.

CoreFoundation/Base.subproj/CFPriv.h
compnerd: I think I ended up favoring @parkera's suggestion of `TARGET_RT_64_BIT` as it fits more inline with the rest of the project.  Split into #1808

compnerd:
Once #1808 is merged, this change becomes primarily locking with a small sprinkling of TLS.  We can then focus on getting that right and finishing up the Windows port for CF.

CoreFoundation/RunLoop.subproj/CFRunLoop.c
compnerd: L415 shows that this is of type `CFLock_t`.

compnerd:
Alright, its that time ... good night PR.  I've split this up into #1810 and #1811 .  #1810 should be pretty much uncontroversial.  The threading issues being discussed here are all handled in #1811 via a new `_CFMutex` and `_CFRecursiveMutex` type for the RunLoop handling.

PR #1790, opened 2018-11-29
Only call pthread_key_create once in __CFTSDInitialize

benlangmuir:
We were calling pthread_key_create more than PTHREAD_KEYS_MAX times
(1024 on the version of Linux I tested with). It should only need to be
called once.

Fixes https://bugs.swift.org/browse/SR-9241


benlangmuir:
I didn't see an obvious way to test this change, but let me know if you have suggestions!

benlangmuir:
@swift-ci test

millenomi:
LGTM after looking at the fact __CFTSDGetTable invokes this at every invocation.

PR #1791, opened 2018-11-29
Change `rangeOfString(other)` to `range(of: other)` in the docs

lucatorella:
In the `NSStringAPI.swift` there was still a reference to the old `rangeOfString()` method. This method was renamed to `range(to:)` so I fixed the documentation otherwise it can create some confusion 🙂

I did the same fix here https://github.com/apple/swift/pull/20860


ianpartridge:
@swift-ci test and merge

PR #1792, opened 2018-12-03
UUID.uuidString should return an upper String like on Darwin

pvieito:
Report for NSUUID: https://bugs.swift.org/browse/SR-865

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -410,6 +410,7 @@ if(ENABLE_TESTING)
                          TestFoundation/TestURL.swift
                          TestFoundation/TestUserDefaults.swift
                          TestFoundation/TestUtils.swift
+                         TestFoundation/TestUUID.swift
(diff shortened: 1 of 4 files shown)


xwu:
Could you please write a more descriptive comment above as to what's being fixed? The link is to a bug that's already been fixed for several years.

This PR also needs a test that fails without the patch and passes with the patch. Also, the same change should be committed to the Darwin Foundation [overlay](https://github.com/apple/swift/blob/master/stdlib/public/Darwin/Foundation/UUID.swift).

compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

pvieito:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1793, opened 2018-12-03
Fix mixup with _CFRegularExpressionMatchingOptions and Flags.

drodriguez:
The Swift function
`NSRegularExpression.enumerateMatches(in:options:range:using:)` receives
a `NSRegularExpression.MatchingOptions`, but the `using:` block receives
a `NSRegularExpression.MatchingFlags`. At one point in the code of
CoreFoundation, the results of a call the obtained `Flags` were shoved
into an `Options` variable, instead of a `Flags` one. This error was
perpetuated in the function points signature in the header, which Swift
imported, forcing Swift to do a conversion between
`_CFRegularExpressionMatchingOptions` into
`NSRegularExpression.MatchingFlags` using the raw value.


compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

PR #1796, opened 2018-12-07
Simple portions of the Windows Port

compnerd:
This is set of changes that are simple and non-controversial for the Windows port.  This allows us to work on a smaller set of changes that have more interesting consequences (like threading and locking) without having to worry about the additional context of these patches.  It also makes it easier to work on the port as it reduces the likelihood of conflicts.


compnerd:
@swift-ci please test

compnerd:
CC: @parkera @millenomi 
As discussed on #1783, this splits off the "simple" parts of the change so that we can focus on the possibly worrisome patches.

compnerd:
hahaha, `__unused` in the header is causing the issue with glibc!

compnerd:
@swift-ci please test

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -307,29 +304,21 @@ CF_EXPORT int _NS_access(const char *name, int amode);
 #define WIN32_LEAN_AND_MEAN
 
 #ifndef WINVER
-#define WINVER  0x0501
+#define WINVER  0x0601
```
jakepetroules: 👍

CoreFoundation/String.subproj/CFBurstTrie.c
```diff
@@ -351,7 +351,7 @@ CFBurstTrieRef CFBurstTrieCreateFromFile(CFStringRef path) {
     if (stat(filename, &sb) != 0) return NULL;
 
     /* Check if file can be opened */
-    if ((fd=open(filename, CF_OPENFLGS|O_RDONLY)) < 0) return NULL;
+    if ((fd = open(filename, CF_OPENFLGS|O_RDONLY, 0)) < 0) return NULL;
```
jakepetroules: This needs to use `_wsopen_s` on Windows, not `open`.

Unfortunately, the story with C APIs is the same as with the *A vs *W variants of Win32 functions -- you *must* use the wide-character variants for Unicode to work at all.

Please go through and audit any C API calls through SCF and make sure that the _w variants are used on Windows. It's probably a good idea to separate all C APIs which take strings into an abstraction layer somewhere.

I'd do something like this. Have an abstraction layer which provides both UTF-8 and UTF-16 variants of each C API that we use:

    int scf_open(const char *);
    int scf_wopen(const uint16_t *, size_t len); // do NOT use wchar_t, it's different size on different platforms
    int scf_cfopen(CFStringRef);

In the implementations (pseudocode):

    int scf_open(const char *s) {
        if (windows) {
            // We have a UTF-8 string and are on Windows, convert to UTF-16
            auto x = convert_to_utf16(s);
            return scf_wopen(x.ptr, x.length);
        } else {
            // We have a UTF-8 string and are on UNIX, no conversion needed
            return ::open(s);
        }
    }

    int scf_wopen(const uint16_t *s, size_t len) {
        if (windows) {
            // We have a UTF-16 string and are on Windows, no conversion needed
            return ::_wsopen_s(s);
        } else {
            // We have a UTF-16 string and are on UNIX, convert to UTF-8
            return scf_open(convert_to_utf8(s));
        }
    }

    int scf_cfopen(CFStringRef s) {
        if (windows) {
            return scf_wopen(CFStringGetCharactersPtr(s), CFStringGetLength(s));
        } else {
            char filename[PATH_MAX];
            CFStringGetCString(s, filename, PATH_MAX, kCFStringEncodingUTF8);
            return scf_open(filename);
        }
    }

That way you can call either the UTF-8 or UTF-16 variant depending on what kind of string you have coming from the input, and it'll do the right thing on each platform in order to minimize conversions.

For example, when you have a CFString and need to pass it to a native Win32 API, you can actually avoid the UTF-8 conversion you're doing here (which again is currently broken due to UTF-8 not being supported).

compnerd:

CoreFoundation/String.subproj/CFBurstTrie.c
compnerd: `open` is not `open`, it is a macro, and replaces it with `_NS_open`, and it does go through `_wopen_s`.  I believe that all of the ones have already been replaced with `_NS_*` equivalents.

compnerd: Yeah, I agree, that would be a great follow up!

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1325,21 +1336,35 @@ static void _CFThreadSpecificDestructor(void *ctx) {
...
+  return (CFTypeRef)FlsGetValue(key);
+#else
     return (CFTypeRef)pthread_getspecific(key);
+#endif
 }
 
 void _CFThreadSpecificSet(_CFThreadSpecificKey key, CFTypeRef _Nullable value) {
+#if DEPLOYMENT_TARGET_WINDOWS
+    if (value != NULL)
+        swift_retain((void *)value);
```
parkera: Nit pick here - we always use curly braces or single-line style for if statements.

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -759,6 +761,15 @@ static bool also_do_stderr(const _cf_logging_style style) {
     return result;
 }
 
+#if DEPLOYMENT_TARGET_WINDOWS
+static struct tm *localtime_r(time_t *tv, struct tm *result) {
+  struct tm *tm = localtime(tv);
+  if (tm)
```
parkera:
```suggestion
  if (tm) {
```

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -759,6 +761,15 @@ static bool also_do_stderr(const _cf_logging_style style) {
...
+    *result = *tm;
```
parkera:
```suggestion
    *result = *tm;
}
```

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1325,21 +1336,35 @@ static void _CFThreadSpecificDestructor(void *ctx) {
...
+#if DEPLOYMENT_TARGET_WINDOWS
+  return (CFTypeRef)FlsGetValue(key);
+#else
     return (CFTypeRef)pthread_getspecific(key);
+#endif
 }
 
 void _CFThreadSpecificSet(_CFThreadSpecificKey key, CFTypeRef _Nullable value) {
+#if DEPLOYMENT_TARGET_WINDOWS
+    if (value != NULL)
```
parkera:
```suggestion
    if (value != NULL) {
```

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1325,21 +1336,35 @@ static void _CFThreadSpecificDestructor(void *ctx) {
...
+        swift_retain((void *)value);
+    FlsSetValue(key, value);
```
parkera:
```suggestion
    FlsSetValue(key, value);
}
```

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
I think that I can improve some of the Windows implementations, but, lets get the basic bits in and do that in a follow up.

PR #1798, opened 2018-12-10
Use strongSelf instead of `self`

lukabratos:
Replace using:
```swift
`self`
```
which is relying on a [compiler bug](https://github.com/apple/swift-evolution/blob/master/proposals/0079-upgrade-self-from-weak-to-strong.md#relying-on-a-compiler-bug) with `strongSelf`.

Unwrapping `self` by using `strongSelf` is already used in [other places](https://github.com/search?q=org%3Aapple+strongSelf&type=Code) so this will make the code consistent across the projects.

### DIFF
--- a/Foundation/URLSession/URLSession.swift
+++ b/Foundation/URLSession/URLSession.swift
@@ -295,10 +295,10 @@ open class URLSession : NSObject {
 
            let invalidateSessionCallback = { [weak self] in
                //invoke the delegate method and break the delegate link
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci test and merge

linqingmo:
```swift
guard let self = self else { return }
```
would be better. [SE-0079](https://github.com/apple/swift-evolution/blob/master/proposals/0079-upgrade-self-from-weak-to-strong.md)

PR #1801, opened 2018-12-12
Fix NSString.substring returning an address.

drodriguez:
The second branch of NSString.substring(with:) creates a buffer where
the contents of the substring are copied. However, instead of returning
the contents of the buffer, the result was the pointer value of the
UnsafeMutablePointer. This commit changes to use
String(utf16CodeUnits:count:) to return the contents.


parkera:
Seems reasonable. Can we find out why it's not hit in your tests, so that we know that we are getting coverage on the change?

spevans:
@swift-ci test

compnerd:
@swift-ci please test

compnerd:
@parkera - I believe that this addresses your concerns now?

PR #1805, opened 2018-12-12
Windows Port Continuation

compnerd:
This replaces some of the threading types to allow us to use the Windows threading model rather than POSIX threads.  This is still incomplete, but these should also be pretty safe without changing the semantics for the other platforms.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -558,15 +558,14 @@ CF_EXPORT void _NS_pthread_setname_np(const char *name) {
     }
 }
 
-static _CFThreadRef __initialPthread = { NULL, 0 };
+static _CFThreadRef __initialPthread = INVALID_HANDLE_VALUE;
 
 CF_EXPORT int _NS_pthread_main_np() {
-    _CFThreadRef me = pthread_self();
-    if (NULL == __initialPthread.p) {
-        __initialPthread.p = me.p;
-        __initialPthread.x = me.x;
-    }
-    return (pthread_equal(__initialPthread, me));
+    if (__initialPthread == INVALID_HANDLE_VALUE)
+      DuplicateHandle(GetCurrentProcess(), GetCurrentThread(),
+                      GetCurrentProcess(), &__initialPthread, 0, FALSE,
+                      DUPLICATE_SAME_ACCESS);
+    return CompareObjectHandles(__initialPthread, GetCurrentThread());
 }
 
 #endif
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -558,10 +558,10 @@ CF_EXPORT void _NS_pthread_setname_np(const char *name) {
     }
 }
 
-static pthread_t __initialPthread = { NULL, 0 };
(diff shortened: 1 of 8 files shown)


compnerd:
CC: @millenomi @parkera @phausler

compnerd:
@swift-ci please test

CoreFoundation/RunLoop.subproj/CFSocket.c
```diff
@@ -2625,7 +2625,7 @@ static CFSocketRef _CFSocketCreateWithNative(CFAllocatorRef allocator, CFSocketN
 #endif
         pthread_create(&tid, &attr, __CFSocketManager, 0);
         pthread_attr_destroy(&attr);
-//warning CF: we dont actually know that a pthread_t is the same size as void *
+//warning CF: we dont actually know that a _CFThreadRef is the same size as void *
```
jakepetroules: Could add (C11 has this now):

    _Static_assert(sizeof(void* ) == sizeof(CFThreadRef), "sizeof(void* ) != sizeof(CFThreadRef)");

compnerd: Is CF built assuming C11 or does it attempt to compatible with C99?  If so, I'm happy to add the assertion.

compnerd:
@swift-ci please test

CoreFoundation/RunLoop.subproj/CFSocket.c
jakepetroules: Seems unlikely that anyone would attempt to compile this without a C11-compatible compiler at this point, but you could guard it with an ifdef, or check with @parkera, who should know for sure.

compnerd:
@swift-ci please test

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -558,15 +559,14 @@ CF_EXPORT void _NS_pthread_setname_np(const char *name) {
...
 
 CF_EXPORT int _NS_pthread_main_np() {
-    pthread_t me = pthread_self();
-    if (NULL == __initialPthread.p) {
-        __initialPthread.p = me.p;
-        __initialPthread.x = me.x;
-    }
-    return (pthread_equal(__initialPthread, me));
+    if (__initialPthread == INVALID_HANDLE_VALUE)
+      DuplicateHandle(GetCurrentProcess(), GetCurrentThread(),
```
millenomi: What are the threading characteristics of this call? The review is conditional on this being thread-safe, or at least harmless if done more than once.

compnerd: The handle is process-wide, so, it is safe.  Calling `GetCurrentProcess` and `GetCurrentThread` is perfectly fine, the former returns the process handle, the latter the current thread's handle.  There is a note above the function that it must be called on the main thread only before any threads are created.

compnerd: Spoke with @millenomi offline, the concern was over this being called from multiple threads, which is not permitted.  This is invoked from `__CFInitialize` before any threads are created, and thus guaranteed to be executed on the main thread.

parkera:
I'm late on this one, but re: the C11 thing - we use it for Darwin, and I'm ok with it here, assuming that all our target platforms support it. I didn't think VS had C11 support but happy to hear I'm wrong on that, if so.

PR #1806, opened 2018-12-13
[CoreFoundation] Fix incorrect cast in _CFRuntimeCreateInstance.

drodriguez:
_cfinfoa field of CFRuntimeBase is an _Atomic(uint64_t), but for some
reason it was being accessed from a uint32_t pointer (which I think
violates strict aliasing rules). This was working for little endian
architectures, but would have failed for big endian architectures when
writing the value back into the MSB instead of the LSB.

The fix uses the right size (64 bits) and removes the castings where
possible. However, to avoid the _Atomic to be honored, the field is
still casted to an uint64_t, which might hide future errors (like the
situation this is trying to fix).

### DIFF
--- a/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/CoreFoundation/Base.subproj/CFRuntime.c
@@ -443,10 +443,10 @@ CFTypeRef _CFRuntimeCreateInstance(CFAllocatorRef allocator, CFTypeID typeID, CF
     memset(&memory->_cfinfoa, 0, size - (sizeof(memory->_cfisa) + sizeof(memory->_swift_rc)));
 
     // Set up the cfinfo struct
-    uint32_t *cfinfop = (uint32_t *)&(memory->_cfinfoa);
+    uint64_t *cfinfop = (uint64_t *)&(memory->_cfinfoa);
     // The 0x80 means we use the default allocator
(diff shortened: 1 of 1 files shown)


CoreFoundation/Base.subproj/CFRuntime.c
```diff
@@ -443,10 +443,10 @@ CFTypeRef _CFRuntimeCreateInstance(CFAllocatorRef allocator, CFTypeID typeID, CF
     memset(&memory->_cfinfoa, 0, size - (sizeof(memory->_cfisa) + sizeof(memory->_swift_rc)));
 
     // Set up the cfinfo struct
-    uint32_t *cfinfop = (uint32_t *)&(memory->_cfinfoa);
+    uint64_t *cfinfop = (uint64_t *)&(memory->_cfinfoa);
```
compnerd: The type of `_cfinfoa` is different on different targets.  IIRC, its if its android or 64-bit, its 64-bit, otherwise 32-bits.

drodriguez: It is different for non-Swift runtimes, but it always 64 for the Swift runtime (and this code only applies to the Swift runtime). See https://github.com/apple/swift-corelibs-foundation/blob/master/CoreFoundation/Base.subproj/CFRuntime.h#L198

compnerd:
@swift-ci please test

PR #1807, opened 2018-12-14
[Foundation] Mark Decimal as @_fixed_layout.

dcci:


dcci:
@swift-ci please test

itaiferber:
Is there a measured performance need for this? What's spurring this change?

phausler:
@itaiferber NSDecimal is defined in C and therefore is naturally fixed layout iiuc.

itaiferber:
@phausler Sure, but here in s-cl-f are we still bridging to `NSDecimal` from C? If so, then understandable

stephentyrone:
Swift doesn't normally guarantee the binary layout of fields within a struct. Do you need / want the binary layout of this type to match NSDecimal?

phausler:
I dont think we make a promise that layouts are cross platform compatible (even binary forms between Mac and iOS differ with other structures). It so happens that there are a few that do, and there are some via spec (ala Double). Is there some sort of archival form that splats NSDecimals down in binary form that we need to expose for reasons like GDPR etc?

itaiferber:
@dcci @stephentyrone I... don't know if we want to lock down the `Decimal` ABI for the purposes of writing a summary provider. We should discuss this internally to see what we think about this sort of stuff going forward.

swift-corelibs-foundation does not guarantee a stable ABI to begin with. @millenomi Do you have thoughts/directions otherwise?

millenomi:
Per offline conversation: we are not going to lock down our ABI on Linux just to make this more debuggable. If that has to take the form of `po …`, then may it; not all types need to 100% have a debug representation in LLDB, and certainly we shouldn't make ABI guarantees just because of that.

dcci:
Withdrawn this.

PR #1808, opened 2018-12-17
CoreFoundation: handle LLP64 as LP64

compnerd:
Adjust the various sites that consider LP64 to also consider LLP64
environments.  This is needed to support Windows x86_64.  Normalize the checks
across the code base to use `TARGET_RT_64_BIT` which is defined in
TargetConditionals.h.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -229,7 +229,7 @@ extern void __CFGenericValidateType_(CFTypeRef cf, CFTypeID type, const char *fu
 #define __CFBitfield64GetValue(V, N1, N2)	(((V) & __CFBitfield64Mask(N1, N2)) >> (N2))
 #define __CFBitfield64SetValue(V, N1, N2, X)	((V) = ((V) & ~__CFBitfield64Mask(N1, N2)) | ((((uint64_t)X) << (N2)) & __CFBitfield64Mask(N1, N2)))
 
-#if __LP64__ || DEPLOYMENT_TARGET_ANDROID
+#if TARGET_RT_64_BIT || DEPLOYMENT_TARGET_ANDROID
 typedef uint64_t __CFInfoType;
 #define __CFInfoMask(N1, N2) __CFBitfield64Mask(N1, N2)
 #else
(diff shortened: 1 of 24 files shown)


compnerd:
CC: @parkera @millenomi @phausler

compnerd:
@swift-ci please test

CoreFoundation/NumberDate.subproj/CFBigNumber.c
```diff
@@ -9,6 +9,7 @@
 	Original author: Zhi Feng Huang
 */
 
+#include <CoreFoundation/TargetConditionals.h>
```
phausler: We cant do this since it only exists on non-darwin platforms. The Darwin variant is just <TargetConditionals.h> additionally the proper TargetConditionals header should be included by CFBase.h so any additions should not be needed.

This applies to all other instances of the addition of this header.

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -229,7 +229,7 @@ extern void __CFGenericValidateType_(CFTypeRef cf, CFTypeID type, const char *fu
 #define __CFBitfield64GetValue(V, N1, N2)	(((V) & __CFBitfield64Mask(N1, N2)) >> (N2))
 #define __CFBitfield64SetValue(V, N1, N2, X)	((V) = ((V) & ~__CFBitfield64Mask(N1, N2)) | ((((uint64_t)X) << (N2)) & __CFBitfield64Mask(N1, N2)))
 
-#if __LP64__ || DEPLOYMENT_TARGET_ANDROID
+#if TARGET_RT_64_BIT || DEPLOYMENT_TARGET_ANDROID
```
phausler: TARGET_ are usually only from TargetConditionals.h which CoreFoundation does not control on Darwin. So we probably need to first define something along the lines of CFRUNTIME_64BIT. Additionally this should be defined if it does not get defined in a public header; CFBase.h is a good spot to do this.

```c
#ifndef CFRUNTIME_64BIT
#if __LLP64__ || __LP64__
#define CFRUNTIME_64BIT 1
#else
#define CFRUNTIME_64BIT 0
#endif
#endif
```

phausler: hmm I take that back: TARGET_RT_64_BIT is defined by TargetConditionals on Darwin.

Do we have that defined in <CoreFoundation/TargetConditionals.h>?

CoreFoundation/NumberDate.subproj/CFBigNumber.c
phausler: even though TARGET_RT_64_BIT is ok, the CoreFoundation/TargetConditionals thing still holds. These header includes would be a breaking change for cycling back to the Darwin implementation.

compnerd: Okay, I'll adjust the sites to use `CFBase.h` instead.  The locations that I included this header were not already including `CFBase.h`.

CoreFoundation/Base.subproj/CFInternal.h
compnerd: Oh, so we shouldn't be using `TARGET_RT_64_BIT` and define and use `CF_RUNTIME_64_BIT` through out?

compnerd: Yes, that is defined in the `<CoreFoundation/TargetConditionals.h>` header.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@phausler - does this look good now?  or are there other changes to be made?

compnerd:

compnerd:

PR #1809, opened 2018-12-17
[SR-9503] Match NSNull's description to Darwin's

ogres:
Fixes SR-9503.

NSNull has `<null>` as description on Darwin.
It had `<NSNull: 0x00000000075ce1c0>` elsewhere.

Overridden description method to match Darwin's behaviour.

I've also noticed that `testCase(TestNSNull.allTests)` was listed twice in `TestFoundation/main` and removed extra entry.


millenomi:

PR #1814, opened 2018-12-21
build: pass along more variables to the subonfigure

compnerd:
Explicitly call out the CPU/OS that are building for.  Pass along the
ICU, CURL, and libxml2 library/headers to the subconfgiure.  This is
particularly helpful for cross-compilation where the configuration is
being pass through via the commandline or through CMake caches.


compnerd:

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1815, opened 2018-12-22
removed dead code

mxswd:
Neither of these functions use the bootstrap mach port.


parkera:
There is some darwin-specific code that is `#ifdef`ed out during export from the closed source CF that uses this. It would be better to take care of it on our side. I'll file a radar for it and we should be able to sync it up again next time we have a source drop.  rdar://problem/47012113

parkera:
Thanks for noticing this though!

PR #1816, opened 2019-01-06
Use NSArray/NSMutableArray and NSSet/NSMutableSet in NSOrderedSet

amorde:
Taking another stab at #1367 

This cleans up a lot of the implementation since we can defer work to `NS{Mutable}Array` and `NS{Mutable}Set`


Foundation/NSOrderedSet.swift
```diff
@@ -252,22 +228,16 @@ extension NSOrderedSet {
     }
     
     public func objectEnumerator() -> NSEnumerator {
-        guard type(of: self) === NSOrderedSet.self || type(of: self) === NSMutableOrderedSet.self else {
-            NSRequiresConcreteImplementation()
-        }
-        return NSGeneratorEnumerator(_orderedStorage.map { __SwiftValue.fetch(nonOptional: $0) }.makeIterator())
+        return _orderedStorage.objectEnumerator()
```
amorde: This has the interesting side effect of making these enumerators behave similar to Darwin - they will now break if mutation occurs during iteration. The previous implementation made a copy of the all the objects using `map` and therefore did not have this issue.

TestFoundation/TestNSOrderedSet.swift
```diff
@@ -62,6 +61,126 @@ class TestNSOrderedSet : XCTestCase {
             index += 1
         }
     }
+    
+    func test_enumerationUsingBlock() {
```
amorde: This test was copied from `NSArray` tests for coverage on concurrent enumeration

Foundation/NSOrderedSet.swift
```diff
@@ -544,7 +581,8 @@ extension NSMutableOrderedSet {
     }
     
     open func intersectSet(_ other: Set<AnyHashable>) {
-        for case let item as AnyHashable in self where !other.contains(item) {
+        let objects = Array(self)
+        for case let item as AnyHashable in objects where !other.contains(item) {
```
amorde: since the enumerator now breaks if mutation occurs, we have to make a copy here

millenomi:
@swift-ci please test

millenomi:

PR #1818, opened 2019-01-12
Fix Xcode build

otaviolima:
Using latest master (60651e26396c968a3a70159dc2bf912826da265f) and same toolchain that this commits points to (swift-DEVELOPMENT-SNAPSHOT-2019-01-10-a) I'm unable to build using Xcode unless I make the following changes.

Not sure if those fixes goes through PR's too.

### DIFF
--- a/TestFoundation/TestURL.swift
+++ b/TestFoundation/TestURL.swift
@@ -35,7 +35,7 @@ private func getTestData() -> [Any]? {
         XCTFail("Unable to deserialize property list data")
         return nil
     }
-    guard let parsingTests = testRoot[kURLTestParsingTestsKey] as? [Any] else {
+    guard let parsingTests = testRoot?[kURLTestParsingTestsKey] as? [Any] else {
         XCTFail("Unable to create the parsingTests dictionary")
         return nil
     }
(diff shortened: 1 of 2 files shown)


ikesyo:
@swift-ci Please test

TestFoundation/TestURL.swift
```diff
@@ -35,7 +35,7 @@ private func getTestData() -> [Any]? {
         XCTFail("Unable to deserialize property list data")
         return nil
     }
-    guard let parsingTests = testRoot[kURLTestParsingTestsKey] as? [Any] else {
+    guard let parsingTests = testRoot?[kURLTestParsingTestsKey] as? [Any] else {
```
ikesyo: This is related to [SE-0230](https://github.com/apple/swift-evolution/blob/master/proposals/0230-flatten-optional-try.md).

```swift
guard let testRoot = try? PropertyListSerialization.propertyList(from: data, options: [], format: nil) as? [String : Any] else {
```

This would break builds which uses Swift 5 mode.

spevans:
I think `TestUUID.swift` also needs to be added into the project / `TestFoundation` target

PR #1819, opened 2019-01-12
Use bundled UUID

compnerd:
Setup the build to build the bundled UUID and port to windows in preparation for supporting Windows as a build target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,6 @@ option(FOUNDATION_PATH_TO_XCTEST_BUILD "Path to XCTest build" "")
 find_package(CURL REQUIRED)
 find_package(ICU COMPONENTS uc i18n REQUIRED)
 find_package(LibXml2 REQUIRED)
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
CC: @millenomi @parkera

jakepetroules:
Why not just use [CoCreateGuid](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-cocreateguid) instead of adding the libuuid dependency?

compnerd:
@jakepetroules - because there are auxiliary functions here that need to be re-implemented then.  The conversion to from string can be done using Win32 APIs, but, it seems that if we do this, we can avoid the dependency on e2fsprogs on Linux as well.  Its a pretty tiny code base, already here, and means that all the ports have the same code base.

PR #1821, opened 2019-01-15
NSDecimalValue: Dont use .doubleValue for integer properties.

spevans:
- Decimal: Add an internal .int64Value and .uint64Value.

- NSDecimalValue: For the .int*Value and .uint*Value properties
  use a truncated value from .int64Value and .uint64Value.

- Using .doubleValue lost information when returning large numbers
  > 2e53 due to Double to Int conversion.

- The returned values mostly match Darwin, moreso than just returning 0
  when an exact conversion cannot be made.


spevans:
@swift-ci test

PR #1822, opened 2019-01-15
SR-9322: DateFormatter difference between macOS and Linux

spevans:
- When parsing a date, return nil if the date to be parsed was not
  fully consumed by the parse format.


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

PR #1827, opened 2019-01-17
build: explicitly link against swiftDispatch

compnerd:
swift-corelibs-foundation is an implementation of the Foundation APIs in
swift.  It has a dependency on libdispatch and uses some of the swift
interfaces.  Ensure that we enumerate the dependency for linking
purposes.  This was noticed as missing when building for Windows.


compnerd:
@swift-ci please test

PR #1828, opened 2019-01-17
CoreFoundation cleanup for Windows

compnerd:
This has most of the cleanup needed for using CoreFoundation for Windows.  There are a couple of minor macro definitions that are needed, but that can be done as a follow up.


compnerd:
@swift-ci please test

CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -127,16 +127,20 @@
...
-            #define CF_EXPORT extern __declspec(dllexport)
-        #elif defined(__cplusplus)
-            #define CF_EXPORT extern "C" __declspec(dllimport)
+    #if defined(__cplusplus)
+        #define _CF_EXTERN extern "C"
+    #else
+        #define _CF_EXTERN extern
+    #endif
+
+    #if defined(_USRDLL)
```
parkera: This is defined by the Windows SDK or compiler when building shared libraries?

CoreFoundation/Base.subproj/CFPriv.h
```diff
@@ -57,10 +57,12 @@ CF_EXPORT void _CFRuntimeSetCFMPresent(void *a);
 CF_EXPORT const char *_CFProcessPath(void);
 CF_EXPORT const char **_CFGetProcessPath(void);
 CF_EXPORT const char **_CFGetProgname(void);
+
+#if !TARGET_OS_WIN32
```
parkera: We like to keep these in the positive sense if possible

CoreFoundation/Base.subproj/CFPriv.h
```diff
@@ -220,8 +222,10 @@ typedef CF_OPTIONS(CFOptionFlags, CFSearchPathDomainMask) {
     kCFAllDomainsMask = 0x0ffff	/* all domains: all of the above and more, future items */
 };
 
+#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED
```
parkera: Since this is a private header, the macro must be the `TARGET_OS` style, as those are defined in a public header but `DEPLOYMENT_TARGET` are internal-only.

We're going to eventually remove all of these in favor of `TARGET_OS` everywhere but that will come later.

CoreFoundation/Base.subproj/CFRuntime.c
```diff
@@ -1193,11 +1193,6 @@ void __CFInitialize(void) {
         CFNumberGetTypeID();		// NB: This does other work
 
         __CFCharacterSetInitialize();
-        
-#if DEPLOYMENT_TARGET_WINDOWS
-        __CFWindowsNamedPipeInitialize();
-#endif
-
```
parkera: Did we remove this completely?

CoreFoundation/Error.subproj/CFError.c
```diff
@@ -495,7 +495,7 @@ static CFTypeRef _CFErrorPOSIXCallBack(CFErrorRef err, CFStringRef key) CF_RETUR
     if (!errStr) return NULL;
     if (CFEqual(key, kCFErrorDescriptionKey)) return errStr;	// If all we wanted was the non-localized description, we're done
     
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_WINDOWS
```
parkera: Hm, why is this concept not portable to Windows?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -679,14 +679,10 @@ CFTimeZoneRef CFTimeZoneCreateWithWindowsName(CFAllocatorRef allocator, CFString
...
 
-extern CFStringRef _CFGetWindowsAppleSystemLibraryDirectory(void);
 static void __InitTZStrings(void) {
     static CFLock_t __CFTZDirLock = CFLockInit;
     __CFLock(&__CFTZDirLock);
-    if (!__tzZoneInfo) {
-        CFStringRef winDir = _CFGetWindowsAppleSystemLibraryDirectory();
-        __tzZoneInfo = CFStringCreateWithFormat(NULL, NULL, CFSTR("%@\\etc\\zoneinfo"), winDir);
-    }   
+    // TODO(compnerd) figure out how to initialize __tzZoneInfo
```
parkera: I think we'll need to make sure this works before merging right?

CoreFoundation/Base.subproj/CFBase.h
compnerd: Its a slight abuse of the normal behaviour for Windows.  `_USRDLL` is pretty commonly defined (by Visual Studio's default build rules), but it technically is meant for ATL/MFC applications building a COM server for a DLL (InProc server).

CoreFoundation/Base.subproj/CFPriv.h
compnerd: Hmm, would you prefer:

```
#if TARGET_OS_WIN32
#else
...
#endif
```

Trying to exhaustively enumerate the other targets seems like a failing exercise since most OSes are Unixy :-(.  But, if that really is preferable, I can attempt to do that.

CoreFoundation/Base.subproj/CFPriv.h
compnerd: Nice!  Oooh, thank you so much for that explanation!  I never understood the difference between the `DEPLOYMENT_*` and `TARGET_OS_*` macros.  I'll change this.

CoreFoundation/Base.subproj/CFRuntime.c
compnerd: Nearly completely ... had two straggling references remaining which I just cleaned up :-).

CoreFoundation/Error.subproj/CFError.c
compnerd: Windows doesn't really have POSIX-y errors.  We can translate errors using `FormatMessage` which will take care of the localization.  However, Rather than trying to implement something for that, it seemed that `_CFErrorPOSIXCallBack` was invoked from one site, `_CFErrorInitializeCallBackTable`, which is used by `CFError.*CallBackBlockForDomain` which is currently not in use in Foundation ... so I thought I would take the lazy way out :-p.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: So, the interesting thing is that there is `__CFCopyWindowsTimeZoneList` which will enumerate the data from the registry.  We reference that in `CFTimeZoneCopyKnownNames`.  I don't have the context on why CoreFoundation attempted to provide its own copy of the timezone data rather than using the system one on Windows (as it already uses the system provided one on Darwin, BSD, and Linux).

parkera: I think our Windows version of CF used to bring its own, but using the system TZ data is almost certainly better, since it can change underneath us.

CoreFoundation/Base.subproj/CFPriv.h
parkera: Hm, maybe it's ok. I'm not sure if we are super consistent about it anyway.

compnerd:
@parkera please LMK if I've missed anything else that needs to be addressed.

PR #1831, opened 2019-01-18
Improve flags handling for Windows build

compnerd:
Update the flags handling to enable building for the Windows target which does not support `-rpath` and requires an indication that the shared C library is being used.


compnerd:
@swift-ci please test

compnerd:

PR #1833, opened 2019-01-18
Foundation: use `numericCast` in places for LLP64

compnerd:
LLP64 causes the `Int` type to be different on LLP64 hosts and LP64 as
Swift does not have `intptr_t`.  This prepares for some of those changes
by using `numericCast` to hide the type checks.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -775,7 +775,7 @@ open class FileManager : NSObject {
         var bytesRead = 0
 
         repeat {
-            bytesRead = read(fd, buffer, bytesToRead)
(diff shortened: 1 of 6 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @millenomi @parkera @phausler

millenomi:
So, we talked at length about this. There are assumptions in the compiler that make Int 32-bit on Windows to facilitate C interop, and they're harder to reverse than just touching the few spots where C interop is using different types than on LP32/LP64. The Foundation interface will remain unaffected.

If Windows port maintenance is able to catch any leaking of Int that needs cast and propose patches (e.g. via CI), I'm okay pulling those in as things happen.

PR #1838, opened 2019-01-23
Foundation: DCE `sliceRange` (NFC)

compnerd:
Remove unnecessary construction of `NSRange` and assignment to a
variable where the value was unused.


compnerd:
@swift-ci please test

compnerd:
CC: @phausler @millenomi @parkera

phausler:
This will be subsumed by the import of Data I have from the overlay

compnerd:
Seems like @phausler is already going to take care of this, I'm going to close it so I don't accidentally submit it.

PR #1839, opened 2019-01-23
Foundation: remove duplicated conformance (NFC)

compnerd:
Remove the duplicated conformance specification which causes a warning
to be emitted.


compnerd:
@swift-ci please test

compnerd:
CC: @phausler @millenomi @parkera

PR #1841, opened 2019-01-23
SR-9454: StringProtocol.lineRange(for:) returns invalid range on Linux

norio-nomura:
Fixes [SR-9454](https://bugs.swift.org/browse/SR-9454)

### DIFF
--- a/Foundation/NSStringAPI.swift
+++ b/Foundation/NSStringAPI.swift
@@ -443,7 +443,7 @@ extension StringProtocol where Index == String.Index {
   /// Return an `Index` corresponding to the given offset in our UTF-16
   /// representation.
   func _index(_ utf16Index: Int) -> Index {
-    return Index(encodedOffset: utf16Index + _substringOffset)
+    return self.utf16.index(self.utf16.startIndex, offsetBy: utf16Index)
   }
 
   @inlinable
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

norio-nomura:
We may also need to examine other places that use `String.Index.init(encodedOffset:)`. 🤔

norio-nomura:
> We may also need to examine other places that use `String.Index.init(encodedOffset:)`. 🤔

I filed a separate issue about `encodedOffset`: https://bugs.swift.org/browse/SR-9749

milseman:
@spevans this change looks good to me and is an important fix for Swift 5. But, I'm not familiar with corelibs-foundation, what do you think?

spevans:
I think this needs an extra test on a `SubString` to check that the `_substringOffset` can be removed as the comment for it is:
```// self can be a Substring so we need to subtract/add this offset when
  // passing _ns to the Foundation APIs. Will be 0 if self is String.
```

milseman:
That PR gives new API that you can use instead of having to advance the index yourself. Basically, it will let you simplify your one line of code, but it could be held up for a while awaiting core-team feedback and maybe even some SE work. This change is good and important no matter what happens with that PR, are you ok to merge it @spevans?

spevans:
Yup, thanks for the explanation, lets get this fix in.

spevans:
@swift-ci test and merge

norio-nomura:
Thanks!

PR #1842, opened 2019-01-24
Foundation: homogenise importing C, add MSVCRT

compnerd:
This is preliminary setup for enabling the Windows port.  Ensure that we
import MSVCRT in all the locations where we are importing the C library.
Take the opportunity to homogenise the import for linux/android/cygwin
as well.

### DIFF
--- a/Foundation/AffineTransform.swift
+++ b/Foundation/AffineTransform.swift
@@ -7,12 +7,6 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-#if os(macOS) || os(iOS)
-    import Darwin
-#elseif os(Linux) || CYGWIN
-    import Glibc
-#endif
-
 private let ε: CGFloat = CGFloat(2.22045e-16)
 
 
--- a/Foundation/Data.swift
+++ b/Foundation/Data.swift
@@ -12,16 +12,11 @@
 
 #if DEPLOYMENT_RUNTIME_SWIFT
 
-#if os(macOS) || os(iOS)
-import Darwin
-#elseif os(Linux)
-import Glibc
-
+#if os(Linux)
 @inlinable // This is @inlinable as trivially computable.
 internal func malloc_good_size(_ size: Int) -> Int {
     return size
 }
-
 #endif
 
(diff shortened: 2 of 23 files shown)


compnerd:
@swift-ci please test

Foundation/AffineTransform.swift
```diff
@@ -9,8 +9,10 @@
 
 #if os(macOS) || os(iOS)
```
jakepetroules: Why are tvOS and watchOS missing here...?

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
Well, that's part of the problem.  You can compile glibc for a number of platforms, and just because you can import the module does not mean it is correct.  As an example: I may be building for Linux, but importing glibc might be incorrect because I could have glibc and musl available and wish to build against musl.

jakepetroules:
OK, but `#if os(Linux)` doesn't solve that problem either, since it still hardcodes Glibc. I still think you should generally use `canImport` and later if we want to enable alternates we can do so via additional compiler flags.

compnerd:
There was a single instance of `canImport`, and everywhere else was just spelt out, so I think that switching to `canImport` can be a separate change/conversation.

compnerd:
@swift-ci please test

compnerd:
CC: @phausler @karwa @spevans

compnerd:
@swift-ci please test

spevans:
Looks good on macOS in Xcode

compnerd:
Well, the thing is that nothing really prevents me from building msvcrt on Darwin and then having the msvcrt module available on Darwin.  Why should Darwin be given precedence then?  Why not try to load msvcrt (even on macOS) and use that?  The module can be imported!

spevans:
Having the libc module with different names on different platforms does seems a potential source of problems. It a shame that an alias cant be created and shipped with the toolchain so that a simple `import Libc` could be used everywhere.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1843, opened 2019-01-24
port NSLock to Windows

compnerd:
This is the NSLock specific bits of the Windows port.  It is dependent on a previous change (#1842), but, this at least splits up the change into a digestible change.

### DIFF
--- a/Foundation/NSLock.swift
+++ b/Foundation/NSLock.swift
@@ -8,49 +8,78 @@
 //
 
 import CoreFoundation
+#if os(Windows)
+import WinSDK
+#endif
 
 public protocol NSLocking {
     func lock()
     func unlock()
 }
 
-#if CYGWIN
+#if os(Windows)
+private typealias _MutexPointer = UnsafeMutablePointer<SRWLOCK>
+private typealias _RecursiveMutexPointer = UnsafeMutablePointer<CRITICAL_SECTION>
+private typealias _ConditionVariablePointer = UnsafeMutablePointer<CONDITION_VARIABLE>
+#elseif CYGWIN
 private typealias _MutexPointer = UnsafeMutablePointer<pthread_mutex_t?>
+private typealias _RecursiveMutexPointer = UnsafeMutablePointer<pthread_mutex_t?>
 private typealias _ConditionVariablePointer = UnsafeMutablePointer<pthread_cond_t?>
 #else
 private typealias _MutexPointer = UnsafeMutablePointer<pthread_mutex_t>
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
CC: @millenomi @parkera @phausler @jakepetroules

Foundation/AffineTransform.swift
```diff
@@ -9,8 +9,10 @@
 
 #if os(macOS) || os(iOS)
     import Darwin
-#elseif os(Linux) || CYGWIN
+#elseif os(Linux) || os(Android) || CYGWIN
```
karwa: It would probably be cleaner to use `canImport` for these. [Some parts of the project](https://github.com/apple/swift-corelibs-foundation/blob/95906f238046e4f2aa7662d7ce442a175b00dd13/Foundation/FileManager.swift#L10) already do.

Foundation/AffineTransform.swift
```diff
@@ -9,8 +9,10 @@
...
     import Glibc
+#elseif os(Windows)
+    import MSVCRT
```
phausler: this seems to be growing yet again, perhaps we should find a way to do this in one spot via `@_exported import` for all of these in just one file.

phausler: just tested it out; it seems like we can delete all of them and just have them all in NSSwiftRuntime.swift which seems much better to have one spot of failure and one spot for changes in the future (also worth commentary there to that extent informing future contributors this is the spot to import platform libc type things)

spevans: Interestingly it seems that the `import Darwin` can be removed from `NSSwiftRuntime.swift` as well. I think the Xcode project might be pulling in a header file that ends up including all of the relevant C header files. However on Linux it still requires the `import Glibc` in `NSSwiftRuntime.swift`. Still a good simplification though.

compnerd: Updated #1842 for this.  I'm trying to split up the changes into something more digestible so that we can look at the specifics of the local changes (e.g. just NSLock) rather than everything at once.  Unfortunately, I have a bunch of changes that are foundational (ha ha) to the rest of the changes.

Foundation/AffineTransform.swift
compnerd: Yeah, #1842 is the better place for that.  Just updated that as per @phausler's comment and removed all the non-NSSwiftRuntime instances.

compnerd:

PR #1848, opened 2019-01-24
Foundation: port FileHandle to Windows

compnerd:
This ports the FileHandle interfaces to Windows.

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -10,11 +10,21 @@
 import CoreFoundation
 
 open class FileHandle : NSObject, NSSecureCoding {
+#if os(Windows)
+    private var _handle: HANDLE
+#else
     private var _fd: Int32
+#endif
     private var _closeOnDealloc: Bool
 
+    @available(windows, unavailable,
+               message: "cannot perform non-owning handle to fd conversion")
     open var fileDescriptor: Int32 {
+#if os(Windows)
+        return -1
+#else
         return _fd
+#endif
     }
 
     open var readabilityHandler: ((FileHandle) -> Void)? = {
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @phausler @parkera @millenomi @jakepetroules

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Foundation/FileHandle.swift
```diff
@@ -8,13 +8,26 @@
 //
 
 import CoreFoundation
+#if os(Windows)
+import WinSDK
+#endif
```
spevans: Should this be added to `Foundation/NSSwiftRuntime.swift` instead to keep all of the imports in one place?

Foundation/FileHandle.swift
```diff
@@ -47,6 +60,79 @@ open class FileHandle : NSObject, NSSecureCoding {
     }
 
     internal func _readDataOfLength(_ length: Int, untilEOF: Bool, options: NSData.ReadingOptions = []) throws -> NSData.NSDataReadResult {
+#if os(Windows)
+        if _handle == INVALID_HANDLE_VALUE {
+          return NSData.NSDataReadResult(bytes: nil, length: 0, deallocator: nil)
+        }
```
spevans: Should this not be `precondition(_handle != INVALID_HANDLE_VALUE, "Bad file handle")` to match the 
`precondition(_fd >= 0, "Bad file descriptor")` on the unix side?

Foundation/FileHandle.swift
```diff
@@ -47,6 +60,79 @@ open class FileHandle : NSObject, NSSecureCoding {
...
+          throw NSError(domain: NSPOSIXErrorDomain, code: Int(GetLastError()),
+                        userInfo: nil)
+        }
+
+        if fiFileInfo.dwFileAttributes & DWORD(FILE_ATTRIBUTE_NORMAL) == FILE_ATTRIBUTE_NORMAL {
+          if options.contains(.alwaysMapped) {
+            let hMapping: HANDLE =
+                CreateFileMappingA(_handle, nil, DWORD(PAGE_READONLY), 0, 0, nil)
+            if hMapping == HANDLE(bitPattern: 0) {
+              fatalError("CrateFileMappingA failed")
```
spevans: typo: `CreateFileMappingA failed`

Foundation/FileHandle.swift
```diff
@@ -47,6 +60,79 @@ open class FileHandle : NSObject, NSSecureCoding {
...
+            }
+          }
+        }
+
+        let blockSize: Int = 8 * 1024
+        var allocated: Int = blockSize
+        var buffer: UnsafeMutableRawPointer = malloc(allocated)!
+        var total: Int = 0
+
+        defer { free(buffer) }
```
spevans: The buffer passedd to the `NSData.NSDataReadResult()` call at the bottom isnt copied so this buffer shouldnt be freed at function exit.

Foundation/FileHandle.swift
```diff
@@ -47,6 +60,79 @@ open class FileHandle : NSObject, NSSecureCoding {
...
+          let BytesToRead: DWORD = DWORD(min(blockSize, remaining))
+
+          if (allocated - total) < BytesToRead {
+            allocated *= 2
+            buffer = _CFReallocf(buffer, allocated)
+          }
+
+          var BytesRead: DWORD = 0
+          if ReadFile(_handle, buffer.advanced(by: total), BytesToRead, &BytesRead, nil) == FALSE {
+            throw NSError(domain: NSPOSIXErrorDomain, code: Int(GetLastError()), userInfo: nil)
```
spevans: Add a `free(buffer)` here.

Foundation/FileHandle.swift
```diff
@@ -47,6 +60,79 @@ open class FileHandle : NSObject, NSSecureCoding {
...
+          }
+          total += Int(BytesRead)
+          if BytesRead == 0 || !untilEOF {
+            break
+          }
+        }
+
+        if total == 0 {
+          return NSData.NSDataReadResult(bytes: nil, length: 0, deallocator: nil)
```
spevans: Add `free(buffer)` here.

Foundation/FileHandle.swift
```diff
@@ -47,6 +60,79 @@ open class FileHandle : NSObject, NSSecureCoding {
...
+        }
+
+        buffer = _CFReallocf(buffer, total)
+        let data = buffer.bindMemory(to: UInt8.self, capacity: total)
+        return NSData.NSDataReadResult(bytes: data, length: total)
```
spevans: This needs to pass a deallocator, so should be:

```
return NSData.NSDataReadResult(bytes: data, length: total) { buffer, length in
    free(buffer)
}
```

Foundation/FileHandle.swift
compnerd: Sure, that is fine by me.  I just don't want this exported is all (I don't think that Foundation should pull in the entire Windows SDK).  Alas, the private imports will have to wait :-(.

Foundation/FileHandle.swift
compnerd: Huh?  I must've been asleep at the keyboard.

compnerd:
@swift-ci please test

Foundation/FileHandle.swift
spevans: Ah yes, just noticed it was a different module to the `MSVCRT` one!

Foundation/FileHandle.swift
```diff
@@ -125,44 +222,136 @@ open class FileHandle : NSObject, NSSecureCoding {
...
+#else
         precondition(_fd >= 0, "Bad file descriptor")
         return UInt64(lseek(_fd, 0, SEEK_END))
+#endif
     }
-    
+
     open func seek(toFileOffset offset: UInt64) {
+#if os(Windows)
+        if _handle == INVALID_HANDLE_VALUE { return }
```
spevans: `precondition(_handle != INVALID_HANDLE_VALUE, "Bad file handle")`

Foundation/FileHandle.swift
```diff
@@ -114,9 +198,22 @@ open class FileHandle : NSObject, NSSecureCoding {
         return NSData.NSDataReadResult(bytes: bytePtr, length: total) { buffer, length in
             free(buffer)
         }
+#endif
     }
-    
+
     open func write(_ data: Data) {
+#if os(Windows)
+      guard _handle != INVALID_HANDLE_VALUE else { return }
```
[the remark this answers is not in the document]
compnerd: The other side actually does the `guard _fd >= 0 else { return }` below?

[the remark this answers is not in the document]
spevans: I think the unix side is actually wrong and there should be a precondition there as well, according to https://developer.apple.com/documentation/foundation/filehandle/1410936-write "This method raises an exception if the file descriptor is closed or is not valid "

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Foundation/NSSwiftRuntime.swift
```diff
@@ -22,6 +22,10 @@ import CoreFoundation
 
 @_exported import Dispatch
 
+#if os(Windows)
+import WinSDK
```
parkera: Where did we end up on this? Will this cause people importing Foundation to also get `WinSDK`?

compnerd: Yeah, I think that currently it does currently result in WinSDK getting pulled in, since we do not yet have the private import (I think that @jrose-apple was working on the that?).  If there is a way to avoid that, I'm happy to try that out.  I don't think that going with Shims for this is very good though because as we start using more interfaces across Windows, we will effectively end up re-creating the SDK.

parkera: It's probably reasonable in the end; on Darwin it's not like you can't help but import standard Darwin headers if you import Foundation.

parkera:
Looks good to me - thanks!

PR #1851, opened 2019-01-25
Update NSKeyedArchiver.swift

SafelySwift:
Why can't we use an if let?


ikesyo:
@swift-ci Please test

PR #1852, opened 2019-01-25
Make NSError string constants more "Swifty"

SafelySwift:
Why don't we put the constants in a `enum`, so we don't have verbose names? `NSError.Domain.cocoa` looks much better than `NSCocoaErrorDomain`. 

The old constants are still there for backwards compatibility.

Not sure if this is something that can be done.

### DIFF
--- a/Foundation/NSError.swift
+++ b/Foundation/NSError.swift
@@ -13,33 +13,53 @@
 
 import CoreFoundation
 
-public typealias NSErrorDomain = NSString
+public let NSCocoaErrorDomain: String = NSError.Domain.cocoa.rawValue as String
+public let NSPOSIXErrorDomain: String = NSError.Domain.posix.rawValue as String
+public let NSOSStatusErrorDomain: String = NSError.Domain.osStatus.rawValue as String
+public let NSMachErrorDomain: String = NSError.Domain.mach.rawValue as String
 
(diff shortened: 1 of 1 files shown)


Foundation/NSError.swift
```diff
@@ -13,33 +13,53 @@
...
-// Other predefined domains; value of "code" will correspond to preexisting values in these domains.
-public let NSPOSIXErrorDomain: String = "NSPOSIXErrorDomain"
-public let NSOSStatusErrorDomain: String = "NSOSStatusErrorDomain"
-public let NSMachErrorDomain: String = "NSMachErrorDomain"
+public let NSCocoaErrorDomain: String = NSError.Domain.cocoa.rawValue as String
+public let NSPOSIXErrorDomain: String = NSError.Domain.posix.rawValue as String
+public let NSOSStatusErrorDomain: String = NSError.Domain.osStatus.rawValue as String
+public let NSMachErrorDomain: String = NSError.Domain.mach.rawValue as String
+
+public enum NSErrorUserInfoKey: NSString {
```
ianpartridge: I don't think we can add new public API can we?

millenomi:
We need to remain source compatible with Darwin, but thanks for the input.

shahmishal:

PR #1856, opened 2019-01-27
Foundation: port NSThread to Windows

compnerd:
Implement the Thread interfaces for Windows.  Extend the CoreFoundation
interface to support thread stack reservation sizing.

### DIFF
--- a/Foundation/Thread.swift
+++ b/Foundation/Thread.swift
@@ -38,7 +38,11 @@ private func NSThreadStart(_ context: UnsafeMutableRawPointer?) -> UnsafeMutable
     let thread: Thread = NSObject.unretainedReference(context!)
     Thread._currentThread.set(thread)
     if let name = thread.name {
+#if os(Windows)
+        _CFThreadSetName(GetCurrentThread(), name)
+#else
         _CFThreadSetName(pthread_self(), name)
+#endif
     }
     thread._status = .executing
     thread.main()
@@ -55,7 +59,11 @@ open class Thread : NSObject {
             if Thread.isMainThread {
                 return mainThread
             } else {
+#if os(Windows)
+                return Thread(thread: GetCurrentThread())
+#else
                 return Thread(thread: pthread_self())
+#endif
             }
         }
     }
@@ -89,6 +97,19 @@ open class Thread : NSObject {
     }
 
     open class func sleep(until date: Date) {
+#if os(Windows)
+        var hTimer: HANDLE = CreateWaitableTimerW(nil, TRUE, nil)
+        // FIXME(compnerd) how to check that hTimer is not NULL?
+        defer { CloseHandle(hTimer) }
+
+        // the timeout is in 100ns units
+        var liTimeout: LARGE_INTEGER =
+            LARGE_INTEGER(QuadPart: LONGLONG(date.timeIntervalSinceReferenceDate) * -10000000)
+        if SetWaitableTimer(hTimer, &liTimeout, 0, nil, nil, FALSE) == FALSE {
+          return
+        }
+        WaitForSingleObject(hTimer, WinSDK.INFINITE)
+#else
         let start_ut = CFGetSystemUptime()
         let start_at = CFAbsoluteTimeGetCurrent()
         let end_at = date.timeIntervalSinceReferenceDate
@@ -109,9 +130,23 @@ open class Thread : NSObject {
             }
             ti = end_ut - CFGetSystemUptime()
         }
+#endif
     }
 
     open class func sleep(forTimeInterval interval: TimeInterval) {
+#if os(Windows)
+        var hTimer: HANDLE = CreateWaitableTimerW(nil, TRUE, nil)
+        // FIXME(compnerd) how to check that hTimer is not NULL?
+        defer { CloseHandle(hTimer) }
+
+        // the timeout is in 100ns units
+        var liTimeout: LARGE_INTEGER =
+            LARGE_INTEGER(QuadPart: LONGLONG(interval) * -10000000)
+        if SetWaitableTimer(hTimer, &liTimeout, 0, nil, nil, FALSE) == FALSE {
+          return
+        }
+        WaitForSingleObject(hTimer, WinSDK.INFINITE)
+#else
         var ti = interval
         let start_ut = CFGetSystemUptime()
         let end_ut = start_ut + ti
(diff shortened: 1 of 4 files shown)


compnerd:
@swift-ci please test

Foundation/Thread.swift
```diff
@@ -183,6 +229,10 @@ open class Thread : NSObject {
         } else {
             _thread = nil
         }
+#elseif os(Windows)
+        _thread = self.withRetainedReference {
+            return _CFThreadCreate(_attr, NSThreadStart, $0)
+        }
```
spevans: Isnt this block redundant as it is a duplicate of the block below?

Foundation/Thread.swift
```diff
@@ -248,10 +312,13 @@ open class Thread : NSObject {
     private class func backtraceAddresses<T>(_ body: (UnsafeMutablePointer<UnsafeMutableRawPointer?>, Int) -> [T]) -> [T] {
         // Same as swift/stdlib/public/runtime/Errors.cpp backtrace
         let maxSupportedStackDepth = 128;
-        let addrs = UnsafeMutablePointer<UnsafeMutableRawPointer?>.allocate(capacity: maxSupportedStackDepth)
+        var addrs = UnsafeMutablePointer<UnsafeMutableRawPointer?>.allocate(capacity: maxSupportedStackDepth)
```
spevans: `addrs` doesnt seem to be mutated anywhere so can this stay as a `let`?

Foundation/Thread.swift
compnerd: Not when it was originally written :-).

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@jrose-apple - suggestions as to what may be the cause of the deserialization issue?

jrose-apple:
Does SwiftFoundation on Mac use the system CoreFoundation?

phausler:
It better not be using the system CF. The layout of the objects are different (and we have functions with differing signatures etc)

jrose-apple:
Hm, good point. We probably would have noticed that a long time ago.

Maybe the CoreFoundation overlay is causing problems? Do we need a SwiftCoreFoundation?

compnerd:
@phausler agreed ... that would be really bad, right, I didn't consider that (sorry, dealing with Linux, Android, and Windows a lot atm, so keep forgetting that Darwin has CF there)

jrose-apple:
Or is _CFThreadRef only defined when certain macros are passed? If that's the case, it can't be used in a stored property (yet).

phausler:
_CFThreadRef is only defined in the SCLF version of CF. The Darwin CF does not have such a notion.(also that typedef is private anyhow)

compnerd:
Woah, I think @phausler is on to something ... I think that it is using the **system** CF.  This is not in the Foundation build, but in XCTest.

compnerd:
Hmm, could it be a difference between the Xcode and the CMake build?  It seems from the jenkins build that this is being invoked from xcodebuild.

Foundation/Thread.swift
```diff
@@ -130,37 +165,48 @@ open class Thread : NSObject {
...
 #endif
     internal var _status = _NSThreadStatus.initialized
     internal var _cancelled = false
-    
+
     open private(set) var threadDictionary: NSMutableDictionary = NSMutableDictionary()
 
-    internal init(thread: pthread_t) {
-        // Note: even on Darwin this is a non-optional pthread_t; this is only used for valid threads, which are never null pointers.
+    internal init(thread: _CFThreadRef) {
```
phausler: this might be the cause of the failure: because it is marked as internal?

compnerd: Well, it was previously `internal` ... we should be allowed to have internal interfaces that are unexposed right?

phausler: it was previously the type of pthread_t which was from the Darwin module, where as _CFThreadRef pulls in the CoreFoundation module which has a naming conflict with the system CoreFoundation from my limited understanding.

phausler: specifically I think this might be associated with things around inlining visibility.

phausler:
> Hm, good point. We probably would have noticed that a long time ago.
> 
> Maybe the CoreFoundation overlay is causing problems? Do we need a SwiftCoreFoundation?

renaming it to SwiftCoreFoundation has other issues that crop up. We tried that early on in the initial bring-up of SCLF. It failed pretty horribly then, but perhaps with alot of work it could be done now that things are not so much in flux as they were back then?

Foundation/Thread.swift
compnerd: Hmm, I see how that got away in the past.  Suggestions on how to possibly address this then?

phausler: there are two potential solutions here:

1) #ifdef the windows version to use the underlying type that _CFThreadRef is typedef'd from and leave the darwin version to be pthread_t
2) make this method fileprivate if possible.

jrose-apple: Another hackaround: you could move the typedef into the Swift module, and then check that it's the same type as _CFThreadRef.

phausler: it seems that importing `typedef` is acting more like `newtype` than `typealias`.

jrose-apple: I *hope* it's not being treated as a CF type. That would be bad. But I *think* it's still a typealias; the compiler is just trying hard to preserve the sugar even though it doesn't need to.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Foundation/Thread.swift
compnerd: Okay, making it `fileprivate` is insufficient, its still failing on the import.  Does it make sense to actually `#if` check and desugar the conditional for Darwin and add a SR for ir?  Or just skip the SR entirely?

compnerd: @jrose-apple - can't just move the typedef as there are references to it in the C code.  I guess I could create a local typealias with a double underscore and see if that can get converted on the way down.

jrose-apple: You need to make something that *doesn't* reference the C one but is equivalent to it. Which stinks, but is probably the least bad answer.

compnerd: @jrose-apple - okay; think that this deserves a SR?

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Foundation/Thread.swift
compnerd: Hmm, this is something that is odd, so I've created SR-9811.

compnerd:

PR #1857, opened 2019-01-27
Update NSKeyedArchiver.swift

SafelySwift:
Similar to [this commit](https://github.com/apple/swift-corelibs-foundation/commit/f753b1c94c806109f01196d7060c0731cad9cd48) implementing a pull request I created, this PR changes occurrences of `if x != nil { /* use x! */ }` to `if let x = x { /* use x */ }`. It is arguably more "Swifty" in terms of syntax.

### DIFF
--- a/Foundation/NSKeyedArchiver.swift
+++ b/Foundation/NSKeyedArchiver.swift
@@ -487,7 +487,7 @@ open class NSKeyedArchiver : NSCoder {
         let className = NSStringFromClass(clsv)
         let mappedClassName = _classNameForClass(clsv)
         
-        if mappedClassName != nil && mappedClassName != className {
+        if let mappedClassName = mappedClassName && mappedClassName != className {
             // If we have a mapped class name, OS X only encodes the mapped name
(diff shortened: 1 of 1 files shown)


Foundation/NSKeyedArchiver.swift
```diff
@@ -487,7 +487,7 @@ open class NSKeyedArchiver : NSCoder {
         let className = NSStringFromClass(clsv)
         let mappedClassName = _classNameForClass(clsv)
         
-        if mappedClassName != nil && mappedClassName != className {
+        if let mappedClassName = mappedClassName && mappedClassName != className {
```
ikesyo: `&&` is not valid in this case; it should be `if let mappedClassName = mappedClassName, mappedClassName != className {`

millenomi:
@SafelySwift the change is needed to be syntactically correct. Are you still interested in shepherding this patch?

SafelySwift:
I mean, now that I think about it, it is not that bad of an issue. I'm sure there was a reason to not use unwrapping, as the core team is not stupid.

PR #1860, opened 2019-01-29
Foundation: port NSHost to Windows

compnerd:
This ports the NSHost functionality to Windows.

### DIFF
--- a/Foundation/Host.swift
+++ b/Foundation/Host.swift
@@ -33,6 +33,19 @@ open class Host: NSObject {
     }
     
     static internal func currentHostName() -> String {
+#if os(Windows)
+      var dwLength: DWORD = 0
+      GetComputerNameExA(ComputerNameDnsHostname, nil, &dwLength)
+      guard dwLength > 0 else { return "localhost" }
+
+      let hostname: UnsafeMutablePointer<Int8> =
+          UnsafeMutableBufferPointer<Int8>.allocate(capacity: Int(dwLength + 1)).baseAddress
+      defer { hostname.deallocate() }
+      guard GetComputerNameExA(ComputerNameDnsHostname, hostname, &dwLength) else {
+        return "localhost"
+      }
+      return String(cString: hostname)
+#else
         let hname = UnsafeMutablePointer<Int8>.allocate(capacity: Int(NI_MAXHOST))
         defer {
             hname.deallocate()
@@ -42,6 +55,7 @@ open class Host: NSObject {
             return "localhost"
         }
         return String(cString: hname)
+#endif
     }
     
     open class func current() -> Host {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @parkera @phausler @millenomi @jakepetroules

compnerd:
@swift-ci please test

Foundation/Host.swift
```diff
@@ -33,6 +33,18 @@ open class Host: NSObject {
...
     
     static internal func currentHostName() -> String {
+#if os(Windows)
+      var dwLength: DWORD = 0
+      GetComputerNameExA(ComputerNameDnsHostname, nil, &dwLength)
+
+      let hostname: UnsafeMutablePointer<Int8>? =
+          UnsafeMutableBufferPointer<Int8>
+              .allocate(capacity: Int(dwLength + 1))
+              .baseAddress
```
spevans: In scl-foundation source, long lines arent wrapped.

Foundation/Host.swift
```diff
@@ -33,6 +33,18 @@ open class Host: NSObject {
...
+      defer { hostname?.deallocate() }
+      GetComputerNameExA(ComputerNameDnsHostname, hostname, &dwLength)
+      return String(cString: hostname!)
```
spevans: Can a check be added to return `"localhost"` if `hostname` is empty?

Foundation/Host.swift
```diff
@@ -33,6 +33,18 @@ open class Host: NSObject {
     }
     
     static internal func currentHostName() -> String {
+#if os(Windows)
+      var dwLength: DWORD = 0
+      GetComputerNameExA(ComputerNameDnsHostname, nil, &dwLength)
+
+      let hostname: UnsafeMutablePointer<Int8>? =
```
spevans: Does `hostname` need to be an optional?

Foundation/Host.swift
compnerd: I really wish that `swift-format` was setup.  I'll join the lines if thats the style here.

Foundation/Host.swift
compnerd: I don't think that is possible (that is, `hostname` being empty).  We know the size of the host name due to the query above, so we should have plenty of space.  I suppose that it is possible to hit the race condition with the TOCTOU here ... but that does feel like an edge case.

Foundation/Host.swift
compnerd: hmm, I suppose not anymore, since it is being assigned a value immediately.

Foundation/Host.swift
spevans: I think its worth adding something like
```
guard dwLength > 0 && GetComputerNameExA(..) else {
  return "localhost"
}
```
just to cover the edge case for completeness, even if its unlikely to get used.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1867, opened 2019-01-31
SR-9758: `XMLParser.parse()` should return `false` if `abortParsing()` is called.

YOCKOW:
As [the Apple's document says](https://developer.apple.com/documentation/foundation/xmlparser/1411778-parse), `parse()` should return false in there is an error or if the parsing operation is aborted.

Resolves [SR-9758](https://bugs.swift.org/browse/SR-9758).

### DIFF
--- a/Foundation/XMLParser.swift
+++ b/Foundation/XMLParser.swift
@@ -465,16 +465,18 @@ open class XMLParser : NSObject {
     }
     
     internal func _handleParseResult(_ parseResult: Int32) -> Bool {
-        return true
-        /*
         var result = true
         if parseResult != 0 {
             if parseResult != -1 {
                 // TODO: determine if this result is a fatal error from libxml via the CF implementations
             }
+            
+            if _parserError == nil {
+                _parserError = NSError(domain: XMLParser.errorDomain, code: Int(parseResult))
+            }
+            result = false
         }
         return result
-        */
     }
 
     internal func parseData(_ data: Data) -> Bool {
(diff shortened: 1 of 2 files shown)


Foundation/XMLParser.swift
```diff
@@ -465,16 +465,18 @@ open class XMLParser : NSObject {
...
         var result = true
         if parseResult != 0 {
             if parseResult != -1 {
                 // TODO: determine if this result is a fatal error from libxml via the CF implementations
             }
+            
+            if _parserError == nil {
+                _parserError = NSError(domain: XMLParser.errorDomain, code: Int(parseResult))
+            }
+            result = false
```
YOCKOW: I know this is not the best way, but it would be better than just `return true`.

Foundation/XMLParser.swift
```diff
@@ -465,16 +465,18 @@ open class XMLParser : NSObject {
     }
     
     internal func _handleParseResult(_ parseResult: Int32) -> Bool {
-        return true
-        /*
         var result = true
         if parseResult != 0 {
             if parseResult != -1 {
```
spevans: It would be simpler to write this as \:
```
if parseResult == 0 {
  return true
} else {
    if _parserError == nil {
         _parserError = NSError(domain: XMLParser.errorDomain, code: Int(parseResult))
    }
} 
return false
```

If more interpretation of `parseResult` is added in the future, it will reduce the extra levels of indentation.

spevans:
@swift-ci test

Foundation/XMLParser.swift
YOCKOW: Thank you for the review. I will rewrite it.

spevans:
@swift-ci test

parkera:
Thank you!

spevans:
@swift-ci test

PR #1870, opened 2019-02-01
Switch to using correct form for unsafe type cast

airspeedswift:
A bug in the standard library allowed conversion between different pointees just using the `UnsafePointer.init` but this should not have been allowed. Updating to the correct form in order to permit apple/swift#22288 to land.

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -301,7 +301,7 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
         }
         if type(of: self) == NSString.self || type(of: self) == NSMutableString.self {
             if _storage._guts._isContiguousASCII {
-                return UnsafePointer<Int8>(_storage._guts.startASCII);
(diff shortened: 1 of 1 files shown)


airspeedswift:
@swift-ci please test

Foundation/NSString.swift
```diff
@@ -301,7 +301,7 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
         }
         if type(of: self) == NSString.self || type(of: self) == NSMutableString.self {
             if _storage._guts._isContiguousASCII {
-                return UnsafePointer<Int8>(_storage._guts.startASCII);
+                return UnsafeRawPointer(_storage._guts.startASCII).assumingMemoryBound(to: Int8.self)
```
itaiferber: The looks reasonable despite (I believe) this cast being unsafe to perform in the first place. But, we do it elsewhere too, so not worth addressing here.

PR #1871, opened 2019-02-01
Fix a bug in format localized number on s390x

samding01:
When building Foundation test on s390x, there is a warning:
    
       CoreFoundation/String.subproj/CFString.c:6030:36: warning: result of comparison of constant -1 with expression of type 'char' is always false [-Wtautological-constant-out-of-range-compare]

The debugging shows  `(disableLocalizedFormatting == -1)` is never `true` because `disableLocalizedFormatting` is defined as `char`, and  then the logic of format localized number is not correct on s390x.

Changing `disableLocalizedFormatting` to `short` to make the logic works on all platforms

### DIFF
--- a/CoreFoundation/String.subproj/CFString.c
+++ b/CoreFoundation/String.subproj/CFString.c
@@ -6028,7 +6028,7 @@ static Boolean __CFStringFormatLocalizedNumber(CFMutableStringRef output, CFLoca
     static SInt32 secondaryGroupingSize = 0;
     
     // !!! This code should be removed before shipping
-    static char disableLocalizedFormatting = -1;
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

compnerd:
@samding01 ... I assume that is due to the use of `-funsigned-char`.  I think that the better thing to do here might be to pass in `-fsigned-char` to the CoreFoundation build, since it does make that assumption in a bunch of places.  Alternatively,  use` _Bool` rather than changing this to a `short`.

samding01:
Thanks @compnerd  and @spevans  Updated to `int`

samding01:
Can you review and test?

compnerd:

PR #1872, opened 2019-02-01
NSData: fix an errno stomp

compnerd:
We would attempt to remove the temporary file.  However, that will
result in calls which can destroy the original errno.  Save the previous
errno before performing the cleanup.


compnerd:
CC: @spevans @phausler @parkera @millenomi

compnerd:
@swift-ci please test

compnerd:

PR #1873, opened 2019-02-01
[SR-7715] [Foundation] FileManager.DirectoryEnumeration converted to generic class to iterate Path(String) and URL types.

karthikkeyan:
Bug Link:
https://bugs.swift.org/browse/SR-7715

Discussion link: 
https://forums.swift.org/t/sr-7715-filemanager-directoryenumerator-from-enumerator-atoptions-enumerates-any-instead-of-url/20051

- `FileManager.DirectoryEnumeration` confirmed with `Sequence` protocol.
- `FileManager.DirectoryEnumeration` is no longer the subclass of `NSEnumerator`, which gives the class the flexibility to be a generic type class.
- `FileManager.DirectoryEnumeration` contains a `nextObject()` method with a generic return type, which should be overridden by all sub-classes.
- `NSPathDirectoryEnumerator` internal class is a subclass of `DirectoryEnumeration<String>` with generic type `String` as it returns a path
- `NSURLDirectoryEnumerator` internal class is a subclass of `DirectoryEnumeration<URL>` with generic type URL since it returns a file url.

url

path

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -1358,11 +1358,11 @@ open class FileManager : NSObject {
         }
 
         var path1entries = Set<String>()
-        while let item = enumerator1.nextObject() as? String {
+        while let item = enumerator1.nextObject() {
             path1entries.insert(item)
         }
 
-        while let item = enumerator2.nextObject() as? String {
+        while let item = enumerator2.nextObject() {
             if path1entries.remove(item) == nil {
                 return false
             }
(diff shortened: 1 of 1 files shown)


karthikkeyan:
@spevans @parkera   Please give me your feedback when you guys get a chance. 

I guess I tagged you guys correctly, as I have seen you two are actively involved in PR reviews. If not my apologies, hope I didn't spam you guys.

millenomi:
We can't change this. One of the primary goals of s-c-f is full source compatibility with Darwin, and Darwin's type is a `NSEnumerator`, which enumerates over `Any`.

The feedback is appreciated, though.

karthikkeyan:
@millenomi Thanks for looking in my PR. 

Sorry Im new to this process, Where do we go from here, do I have to close this PR? or wait till we reach the point to update `NSEnumerator`'s interface?

parkera:
We'll take this feedback into consideration for future Foundation API. I agree that the enumerator methods are not ideal in Swift vs their utility in ObjC.

parkera:
The challenge with taking this change as-is would be that our existing clients using this API would suddenly fail to compile, which is an experience we're trying really hard to avoid. That's frustrating to have happen to you.

parkera:
Thank you for the work here!

karthikkeyan:
@parkera  Thanks Tony, for looking into my PR.

PR #1878, opened 2019-02-02
CoreFoundation: fix CoreFoundation for Foundation static builds

compnerd:
CoreFoundation built statically for Foundation is used as an objects
library.  We want to re-export the symbols.  This requires that exports
are enabled for the target, but CMake defines _EXPORT symbols only for
shared libraries and `-rdynamic` executables.  Use the custom macro for
the static case for Foundation.


compnerd:

PR #1887, opened 2019-02-07
[4.2] SR-8560: XMLDocument on Linux crashes with a Segmentation Fault

spevans:
- When returning an XMLNode, take a reference to the parent
  XMLDocument to prevent the underlying libxml xml document
  from being freed.

- This prevents the document being freed if a reference is still
  held to a XMLNode but not the XMLDocument.

(cherry picked from commit aabed33c1695e1e607a603394f908e9b92554bca)

### DIFF
--- a/TestFoundation/TestXMLDocument.swift
+++ b/TestFoundation/TestXMLDocument.swift
@@ -34,6 +34,7 @@ class TestXMLDocument : LoopbackServerTest {
             ("test_addNamespace", test_addNamespace),
             ("test_removeNamespace", test_removeNamespace),
             ("test_optionPreserveAll", test_optionPreserveAll),
+            ("test_rootElementRetainsDocument", test_rootElementRetainsDocument),
         ]
     }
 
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test 4.2

spevans:
@kevints Another XML crash fix.

TestFoundation/TestXMLDocument.swift
```diff
@@ -527,6 +528,22 @@ class TestXMLDocument : LoopbackServerTest {
...
         XCTAssertEqual(expected, String(describing: document))
     }
+
+    func test_rootElementRetainsDocument() {
+        let str = """
+<?xml version="1.0" encoding="UTF-8"?>
+<plans></plans>
+"""
+
+        let data = str.data(using: .utf8)!
```
ianpartridge: It doesn't matter for this PR, but `let data = Data(str.utf8)` is nicer as there's no optional involved.

spevans: Yes, definitely need a cleanup of the tests in `master` at some point.

kevints:
Thanks @spevans!

@swift-ci test and merge

PR #1889, opened 2019-02-07
Supported FileHandle API for Swift

millenomi:
Adds API to FileHandle that allows clients to catch and handle errors, rather than panic with `fatalError` on I/O issues. This API will be supported going forward and is not experimental.

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -9,6 +9,18 @@
 
 import CoreFoundation
 
+// FileHandle has a .read(upToCount:) method. Just invoking read() will cause an ambiguity warning. Use _read instead.
+// Same with close()/.close().
+#if canImport(Darwin)
+import Darwin
+fileprivate let _read = Darwin.read(_:_:_:)
+fileprivate let _close = Darwin.close(_:)
+#elseif canImport(Glibc)
+import Glibc
+fileprivate let _read = Glibc.read(_:_:_:)
+fileprivate let _close = Glibc.close(_:)
+#endif
+
 open class FileHandle : NSObject, NSSecureCoding {
 #if os(Windows)
     private var _handle: HANDLE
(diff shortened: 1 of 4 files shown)


millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

Foundation/FileHandle.swift
```diff
@@ -405,6 +280,213 @@ open class FileHandle : NSObject, NSSecureCoding {
     public static var supportsSecureCoding: Bool {
         return true
     }
+    
+    private var _isPlatformHandleValid: Bool {
```
spevans: Might be a good idea to move this up into the Windows/Non-Windows  FD & handle blocks at the top to keep all the fd and fd checks in one place.

spevans:
I think the `_nulldeviceFileHandle` might need its own specialised versions of the new Swift 5 API methods and deprecations as it is using an invalid handle.

spevans:
Should also have at least one test for each of the new methods (one and the same for `nulDevice` as well ) just to check coverage.

millenomi:
Oh, right. We need to probably do a subclass check in the implementation, because the new entry points are (on purpose) `public` and not `open`.

pvieito:
Is this new API going to be added to the Foundation SDK overlay? Also, why some API (like `FileHandle.readToEnd()`) return `Data?` instead of `Data`, I think returning an empty `Data` value would be more correct.

Foundation/FileHandle.swift
```diff
@@ -405,6 +280,213 @@ open class FileHandle : NSObject, NSSecureCoding {
...
+        guard FlushFileBuffers(_handle) != FALSE else {
+            throw _NSErrorWithWindowsError(GetLastError(), reading: false)
+        }
+        #else
+        guard fsync(_fd) >= 0 else { throw _NSErrorWithErrno(errno, reading: false) }
+        #endif
+    }
+    
+    @available(swift 5.0)
+    public func close() throws {
```
compnerd: Might be nicer to write this as:

```
guard _isPlatformHandleValid else { return }
#if os(Windows)
guard CloseHandle(_handle) != FALSE else {
}
_handle = INVALID_HANDLE_VALUE
#else
gurad _close(_fd) >= 0 else {
  throw _NSErrorWithErrno(errno, reading: true)
}
_fd = -1
#endif
```

millenomi:
@pvieito I cannot comment on future Mac SDKs.

The API that return `nil` do so at, essentially, EOF.

TestFoundation/TestPipe.swift
```diff
@@ -35,20 +35,19 @@ class TestPipe: XCTestCase {
         pipes = []
     }
 
-    func test_Pipe() {
+    func test_Pipe() throws {
```
parkera: Doesn't this disqualify the test from XCTest (acknowledging that we do not automatically discover them at this time).

compnerd:
@millenomi - I don't think that @pvieito was asking about future macOS SDKs, I think he was asking about the Swift repository side.

TestFoundation/TestPipe.swift
millenomi: No; this is a pattern we use elsewhere that XCTest supports.

millenomi:
Tests will be forthcoming in a separate patch.

millenomi:
@swift-ci please test

spevans:
LGTM

millenomi:
@swift-ci please test

millenomi:
there it goes.

spevans:
@millenomi Should we be porting these changes to `swift-5.0-branch`?

parkera:
Our intent is to line this up with the next release.

PR #1893, opened 2019-02-11
[4.2] SR-7608: Pipe() crashes when the underlying system call fails.

spevans:
- Pipe does not have a failable or throwable initializer but currently
   calls fatalError() for all errors returned from pipe().
    
- If pipe() returns EMFILE or ENFILE then set the fileDescriptor
   to -1 for both .fileHandleForReading and .fileHandleForWriting.
   For all other errors call fatalError().
    
- TestFileHandle.swift: Remove duplicate test_pipe().
    
- TestPipe.swift: Add extra test to try and exhaust the process's pipes.
    
- Test the file descriptor is valid and open for certain methods
   and properties to match Darwin.
    
(cherry picked from commit 870d4259c9d6a7f33a0eadc0c0c214f53f9c550d)

NSPipe: make fileHandleFor{Reading,Writing} mutable
    
    In Objective-C, the fileHandleFor{Reading,Writing} property is mutable even on a
    const instance.  Change the property to read/write.
    
(cherry picked from commit 09cfc2105a752b215fc808e2005f080fd5a05758)

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -368,9 +368,9 @@ extension FileHandle {
 }
 
 open class Pipe: NSObject {
-    private let readHandle: FileHandle
-    private let writeHandle: FileHandle
-    
+    open let fileHandleForReading: FileHandle
+    open let fileHandleForWriting: FileHandle
+
     public override init() {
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test 4.2

PR #1895, opened 2019-02-11
 [4.2] SR-8999: NSData contentsOf expects st_size to be the size of the file.

spevans:
- Modify FileHandle._readDataOfLength(untilEOF:options:) to ignore
  stat.st_size and just use the result of read() to determine whether
  to keep reading.
    
- Also convert syscall error handling to throw instead of calling
  fatalError().
    
- NSData.readBytesFromFileWithExtendedAttributes(options:) now uses
  FileHandle._readDataOfLength(untilEOF:options:) to read from a file
  rather than duplicating the file reading logic.
    
- (Linux) Add a test to read a file in /sys which reports
  stat.st_size > size of the file's contents.
    
(cherry picked from commit 5733313eef587d400980ba1aeb40134dd8357b0a)

SR-8999: Ensure the FileHandle read methods call fatalError() on error
    
- readDataToEndOfFile(), readData(ofLength length: Int) and
  .availableData should call fataError() if there are IO errors,
  not return an empty Data().
   
(cherry picked from commit d58337726234323e6a96ca8c0e69056035793240)

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -24,7 +24,12 @@ open class FileHandle : NSObject, NSSecureCoding {
     }
 
     open var availableData: Data {
-        return _readDataOfLength(Int.max, untilEOF: false)
+        do {
+            let readResult = try _readDataOfLength(Int.max, untilEOF: false)
+            return readResult.toData()
+        } catch {
+            return Data()
+        }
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test 4.2

PR #1896, opened 2019-02-11
[4.2] NSDecimalNumber: Add missing init(value:) initialisers.

spevans:
- Fix init(mantissa:exponent:isNegative:) to set the mantissa
  correctly.

(cherry picked from commit 641e2cb52ee8fc9e75971f7a2cc030c4f1239eac)

### DIFF
--- a/Foundation/NSDecimalNumber.swift
+++ b/Foundation/NSDecimalNumber.swift
@@ -79,29 +79,21 @@ open class NSDecimalNumber : NSNumber {
 
     fileprivate let decimal: Decimal
     public convenience init(mantissa: UInt64, exponent: Int16, isNegative: Bool) {
-        var d = Decimal()
-        d._exponent = Int32(exponent)
+        var d = Decimal(mantissa)
+        d._exponent += Int32(exponent)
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test 4.2

spevans:
If I recall correctly I think these methods were missing originally but I can always remove them if necessary. The `.doubleValue` property part is a bugfix though.

weissi:
@spevans yes, I think we should try to not add API (in this case the inits) unless _strictly_ necessary. This might break code (because it might cause ambiguity).

spevans:
@swift-ci test 4.2

spevans:
@swift-ci test 4.2

compnerd:
@parkera - is there an ability to run a source compatibility test like on swift with Foundation?

spevans:
@weissi I removed the other `init`s

parkera:
@compnerd not sure about the source compatibility test; @shahmishal do you know if we can trigger that from here?

shahmishal:
Yes it’s supported

shahmishal:
@swift-ci test 4.2

PR #1898, opened 2019-02-12
[4.2] FileHandle.truncateFile should reposition on truncate

spevans:
- Accepting truncating to any position, not just 0.
- Reposition to the truncate offset.
- Darwin throws exceptions on error so we use fatalError() instead of silently failing.

(cherry picked from commit fb1e71bbe58329267b799695ff8ea1d307a752d2)


spevans:
@swift-ci test 4.2

PR #1899, opened 2019-02-12
[IMPROVEMENT] Double linked list used to improve performance of dequeue and remove operations in OperationQueue

karthikkeyan:
- `remove()` and `dequeue()` methods time complexity reduces to **O(1)** from **O(n)**.
- All arrays, veryHigh, high, normal, low, veryLow, all, converted to _DoubleLinkedList_.
- New internal class `_IndexedOperationLinkedList` introduced specifically to work with `Operation`.
- _HashMap_ used to index, node by it's operation, in the linked list, in order to avoid traversing the list to find a node, which in return reduces the time complexity to **O(1)**.
- Unit Test added to test cancel one operation and cancel operations of a specific queue priority

**Discussion Link**
https://forums.swift.org/t/operationqueue-dequeue-and-remove-operations-performance-improvement/20370/9

### DIFF
--- a/Foundation/Operation.swift
+++ b/Foundation/Operation.swift
@@ -244,79 +244,161 @@ extension OperationQueue {
     public static let defaultMaxConcurrentOperationCount: Int = Int.max
 }
 
+internal class _IndexedOperationLinkedList {
+    
+    internal class Node {
+        var operation: Operation
+        var next: _IndexedOperationLinkedList.Node? = nil
+        weak var previous: _IndexedOperationLinkedList.Node? = nil
+        
+        init(operation: Operation) {
+            self.operation = operation
+        }
+    }
+    
+    private(set) var root: _IndexedOperationLinkedList.Node? = nil
+    private(set) weak var tail: _IndexedOperationLinkedList.Node? = nil
+    private(set) var count: Int = 0
+    private var nodeForOperation: [Operation: _IndexedOperationLinkedList.Node] = [:]
+    
+    func insert(_ operation: Operation) {
+        let node = _IndexedOperationLinkedList.Node(operation: operation)
+        
+        if root == nil {
+            root = node
+            tail = node
+        } else {
+            node.previous = tail
+            tail?.next = node
+            tail = node
+        }
+        nodeForOperation[node.operation] = node
+        count += 1
+    }
+    
+    func remove(_ operation: Operation) {
+        guard let node = nodeForOperation.removeValue(forKey: operation) else {
+            return
+        }
+        
+        guard let unwrappedRoot = root, let unwrappedTail = tail else {
+            // There cannot be a case where `nodeForOperation` contains a node and either `root` or `tail` is nil
+            // When root and tail are `nil`, `nodeForOperation` dictionary must be empty.
+            fatalError()
+        }
+        
+        if node === unwrappedRoot {
+            let next = unwrappedTail.next
+            next?.previous = nil
+            root = next
+            count -= 1
+            return
+        }
+        
+        if node === unwrappedTail {
+            tail = node.previous
+            tail?.next = nil
+            count -= 1
+            return
+        }
+        
+        // Middle Node
+        let previous = node.previous
+        let next = node.next
+        
+        previous?.next = next
+        next?.previous = previous
+        
+        node.next = nil
+        node.previous = nil
+        count -= 1
+    }
+    
+    func removeFirst() -> Operation? {
+        guard let returnNode = root else {
+            return nil
+        }
+        
+        remove(returnNode.operation)
+        return returnNode.operation
+    }
+    
+    func map<T>(_ transform: (Operation) throws -> T) rethrows -> [T] {
+        var result: [T] = []
+        var current = root
+        while let node = current {
+            result.append(try transform(node.operation))
+            current = current?.next
+        }
+        return result
+    }
+    
+}
+
+
 internal struct _OperationList {
-    var veryLow = [Operation]()
-    var low = [Operation]()
-    var normal = [Operation]()
-    var high = [Operation]()
-    var veryHigh = [Operation]()
-    var all = [Operation]()
+    var veryLow = _IndexedOperationLinkedList()
+    var low = _IndexedOperationLinkedList()
+    var normal = _IndexedOperationLinkedList()
+    var high = _IndexedOperationLinkedList()
+    var veryHigh = _IndexedOperationLinkedList()
+    var all = _IndexedOperationLinkedList()
+    
+    var count: Int {
+        return all.count
+    }
     
     mutating func insert(_ operation: Operation) {
-        all.append(operation)
+        all.insert(operation)
+        
         switch operation.queuePriority {
         case .veryLow:
-            veryLow.append(operation)
+            veryLow.insert(operation)
         case .low:
-            low.append(operation)
+            low.insert(operation)
         case .normal:
-            normal.append(operation)
+            normal.insert(operation)
         case .high:
-            high.append(operation)
+            high.insert(operation)
         case .veryHigh:
-            veryHigh.append(operation)
+            veryHigh.insert(operation)
         }
     }
     
     mutating func remove(_ operation: Operation) {
-        if let idx = all.firstIndex(of: operation) {
-            all.remove(at: idx)
-        }
+        all.remove(operation)
+        
         switch operation.queuePriority {
         case .veryLow:
-            if let idx = veryLow.firstIndex(of: operation) {
-                veryLow.remove(at: idx)
-            }
+            veryLow.remove(operation)
         case .low:
-            if let idx = low.firstIndex(of: operation) {
-                low.remove(at: idx)
-            }
+            low.remove(operation)
         case .normal:
-            if let idx = normal.firstIndex(of: operation) {
-                normal.remove(at: idx)
-            }
+            normal.remove(operation)
         case .high:
-            if let idx = high.firstIndex(of: operation) {
-                high.remove(at: idx)
-            }
+            high.remove(operation)
         case .veryHigh:
-            if let idx = veryHigh.firstIndex(of: operation) {
-                veryHigh.remove(at: idx)
-            }
+            veryHigh.remove(operation)
         }
     }
     
     mutating func dequeue() -> Operation? {
-        if !veryHigh.isEmpty {
-            return veryHigh.remove(at: 0)
+        if let operation = veryHigh.removeFirst() {
+            return operation
         }
-        if !high.isEmpty {
-            return high.remove(at: 0)
+        if let operation = high.removeFirst() {
+            return operation
         }
-        if !normal.isEmpty {
-            return normal.remove(at: 0)
+        if let operation = normal.removeFirst() {
+            return operation
         }
-        if !low.isEmpty {
-            return low.remove(at: 0)
+        if let operation = low.removeFirst() {
+            return operation
         }
-        if !veryLow.isEmpty {
-            return veryLow.remove(at: 0)
+        if let operation = veryLow.removeFirst() {
+            return operation
         }
         return nil
-    }
-    
-    var count: Int {
-        return all.count
     }
     
     func map<T>(_ transform: (Operation) throws -> T) rethrows -> [T] {
--- a/Foundation/Operation.swift
+++ b/Foundation/Operation.swift
@@ -288,7 +288,7 @@ internal class _IndexedOperationLinkedList {
         }
         
         if node === unwrappedRoot {
-            let next = unwrappedTail.next
+            let next = node.next
             next?.previous = nil
             root = next
             count -= 1
--- a/Foundation/Operation.swift
+++ b/Foundation/Operation.swift
@@ -25,6 +25,12 @@ open class Operation : NSObject {
     internal var _depGroup = DispatchGroup()
     internal var _groups = [DispatchGroup]()
 #endif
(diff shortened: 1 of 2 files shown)


karthikkeyan:
@swift-ci Please test

spevans:
@swift-ci test

karthikkeyan:
@spevans Im debugging the CI failure. I will update the code soon.

karthikkeyan:
@spevans I have pushed another commit which will fix the CI failure.

spevans:
@swift-ci test

spevans:
@swift-ci test

Foundation/Operation.swift
```diff
@@ -234,81 +234,163 @@ extension OperationQueue {
...
+    var veryHigh = _IndexedOperationLinkedList()
+    var all = _IndexedOperationLinkedList()
+    
+    var count: Int {
+        return all.count
+    }
     
     mutating func insert(_ operation: Operation) {
-        all.append(operation)
+        all.insert(operation)
```
parkera: What are the required threading guarantees here?

Foundation/Operation.swift
```diff
@@ -234,81 +234,163 @@ extension OperationQueue {
     public static let defaultMaxConcurrentOperationCount: Int = Int.max
 }
 
+internal class _IndexedOperationLinkedList {
+    
+    internal class Node {
```
phausler: this is an extraneous allocation; the node contents can be referenced inside the operation themselves.

Foundation/Operation.swift
```diff
@@ -234,81 +234,163 @@ extension OperationQueue {
...
+        weak var previous: _IndexedOperationLinkedList.Node? = nil
+        
+        init(operation: Operation) {
+            self.operation = operation
+        }
+    }
+    
+    private(set) var root: _IndexedOperationLinkedList.Node? = nil
+    private(set) weak var tail: _IndexedOperationLinkedList.Node? = nil
+    private(set) var count: Int = 0
```
phausler: is this really needed? can't we just keep track of the count externally in the OperationQueue?

Foundation/Operation.swift
```diff
@@ -234,81 +234,163 @@ extension OperationQueue {
...
+    private var nodeForOperation: [Operation: _IndexedOperationLinkedList.Node] = [:]
```
phausler: if you flatten from the node into the operations themselves you wont need this and it would reduce execution complexity.

Foundation/Operation.swift
```diff
@@ -234,81 +234,163 @@ extension OperationQueue {
...
+    
+    func insert(_ operation: Operation) {
```
phausler: I would really call this append to avoid ambiguity

Foundation/Operation.swift
```diff
@@ -234,81 +234,163 @@ extension OperationQueue {
...
+    
+    func remove(_ operation: Operation) {
+        guard let node = nodeForOperation.removeValue(forKey: operation) else {
+            return
+        }
+        
+        guard let unwrappedRoot = root, let unwrappedTail = tail else {
+            // There cannot be a case where `nodeForOperation` contains a node and either `root` or `tail` is nil
+            // When root and tail are `nil`, `nodeForOperation` dictionary must be empty.
+            fatalError()
```
phausler: can we put a message here so that debugging this is super clear?

Foundation/Operation.swift
```diff
@@ -234,81 +234,163 @@ extension OperationQueue {
...
+        }
+        
+        if node === unwrappedRoot {
+            let next = node.next
+            next?.previous = nil
+            root = next
+            count -= 1
+            return
+        }
```
phausler: from a readability standpoint it might be nicer to have these read as:
```
if node === unwrappedRoot {
...
} else if node === unwrappedTail {
...
} else {
...
}
```

not absolutely needed but might be nice for making it super clear.

Foundation/Operation.swift
karthikkeyan: Im not sure what's the requirement around this logic. Looking at the implementation of OperationQueue, I see all the the usage of `_operations` is wrapped by `OperationQueue.lock`

karthikkeyan: So it is thread safe.

Foundation/Operation.swift
karthikkeyan: I think you are right. Since both classes, `_IndexedOperationLinkedList` and `_IndexedOperationLinkedList.Node`, are not used out side this file. I think `fileprivate` would be the correct access modifier.

Foundation/Operation.swift
karthikkeyan: I see your point here. I think I would be useful to have the count of operations for each priorities, may be in the future. 

If you think it don't haven any need now, as well as in the future, then it totally makes sense to remove this property from the linked list and move it over to `OperationQueue`.

Foundation/Operation.swift
karthikkeyan: Consider it done.

Foundation/Operation.swift
karthikkeyan: Of course

Foundation/Operation.swift
karthikkeyan: I think using of **if-elseif-else** improves the code readability. I will update the code.

Foundation/Operation.swift
karthikkeyan: Never mind my previous comment, I miss-read that. I read one of your comment below,

> if you flatten from the node into the operations themselves you wont need this and it would reduce execution complexity.

Now I understand. You want an implemented similar to objc as you mentioned in the forum. It would definitely reduce the space complexity if we use `Operation` itself as a node of the linked list. I will updated the implementation and let you know once it is ready for review.

karthikkeyan: @phausler  As I was implement the flatten-approach, I ran into an interesting problem. This property `OperationQueue.operations` is expected to return an array operations in the same order in which all the operations were added.

Given that we need two types of linked list,
**Type 1**: A list to maintain the operations based on the queue priority (`veryLow`, `low`, `normal`, `high` & `veryHigh`)
**Type 2**: A list to maintain the operations based on the order in which it is added (`all`)

So then, the properties `Operation.next` and `Operation.previous` cannot be reused by _Type 2_ as it being used by _Type 1_ linked list.

First I tried using Linked List for all Type 1, and Array for `all`(Type 2). This approach is definitely an improvement, but, this approach didn't provide the performance as I was expecting, as we still running in **O(n)** time complexity.

I liked your approach of using the `Operation` object as a node in linked list, which reduces the space complexity; I don't want to give up that too. Finally I found a solution that works for both, ordered-list and prioritized-list, by added additional two reference in the `Operation` class.

```swift
class Operation {
    /*
      prioritizedPrevious && prioritizedNext used to traverse in the order 
      in which the operations need to be dequeued
    */
    fileprivate weak var prioritizedPrevious: Operation?
    fileprivate var prioritizedNext: Operation?
    
    /*
      orderdedPrevious && orderdedNext used to traverse in the order 
      in which the operations added to the queue
    */
    fileprivate weak var orderdedPrevious: Operation?
    fileprivate var orderdedNext: Operation?
}
```

Let me know what you think.

Foundation/Operation.swift
karthikkeyan: I thought about this more, this property don't have any use inside `_IndexedOperationLinkedList`. I would make much more sense if the property `count` is moved inside `_OperationList` structure, as we just need to keep counting the operations just in `_OperationList.all`.

karthikkeyan:
@spevans I ran TestFoundation target in macOS using Swift 5.0 2019-02-06 (a) snapshot. I see 14 test cases failed `TestNSNumberBridging`, but those are not related to the changes introduced in this PR. I couldn't figure out the reasons for CI failure in Linux. Can you please point me to the right direction to debug this issue.

Foundation/Operation.swift
karthikkeyan: Heres the updated linked list's class looks like,

```swift
fileprivate class _IndexedOperationLinkedList {
    private(set) var root: Operation? = nil
    private(set) var tail: Operation? = nil
    
    func append(_ operation: Operation, inOrder: Bool = false) { }
    private func appendOperationInPriorityList(_ operation: Operation) { }
    private func appendOperationInOrderList(_ operation: Operation) { }
    
    func remove(_ operation: Operation, inOrder: Bool) { }
    private func removeOperationFromPriorityList(_ operation: Operation) { }
    private func removeOperationFromOrderList(_ operation: Operation) { }
    
    func removeFirstInPriority() -> Operation? { }
}

```

spevans:
The failure is https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/1922/consoleFull#1077775732ba62d58e-7248-467b-91e0-c7508d5cf947 but I think its actually an unrelated intermittent xctest error. I'll kick the tests off again and see if it makes a difference.

spevans:
@swift-ci test

Foundation/Operation.swift
phausler: The operations property is O(n) on Darwin and honestly that is fine since that property is honestly a race condition anyhow and developers should not really use it imho.

karthikkeyan: @phausler  Im couldn't understand your comment, Im kinda getting mixed signal. Do you think the approach I mentioned in the above comments will work or you suggesting some improvements?

phausler: I’m saying that you should not worry about the performance of the operations property and instead focus on the perf I’d add/remove of singular operations. What you have outlined is roughly the same as what the objc version has modulo weak (the objc side uses unsafe unretained pointers). As long as the operations are strongly referenced by the queue somehow the implementation will be on par.

millenomi:
cc @phausler for next steps?

Foundation/Operation.swift
karthikkeyan: @phausler. I think we are on the same page here. This PR is mainly focused on improving the performance of add/remove single operation. The public property `OperationQueue.operations` caught it the way. I will update the PR addressing the all comments and let you know once its ready for review.

karthikkeyan:
@phausler @spevans @parkera @millenomi This PR is ready to go for another round of review. Thanks for helping me out here guys, Im really exited!

Foundation/Operation.swift
```diff
@@ -25,6 +25,12 @@ open class Operation : NSObject {
     internal var _depGroup = DispatchGroup()
     internal var _groups = [DispatchGroup]()
 #endif
+    // prioritizedPrevious && prioritizedNext used to traverse in the order in which the operations need to be dequeued
+    fileprivate weak var prioritizedPrevious: Operation?
+    fileprivate var prioritizedNext: Operation?
+    // orderdedPrevious && orderdedNext used to traverse in the order in which the operations added to the queue
+    fileprivate weak var orderdedPrevious: Operation?
+    fileprivate var orderdedNext: Operation?
```
phausler: Out of these four variables only one needs to be strong (or none if you balanced via the finish which you are going to have to anyhow)

The others really should be Unmanaged. But I guess that can be for a future enhancement.

karthikkeyan: Thanks @phausler for the approval. 

Sure, I will think about more and find a way reduce the strong reference down to just one. I will send one more PR with that improvement, hope this implementation ok to merge.

Foundation/Operation.swift
karthikkeyan: Resolving this comment since it is addressed

Foundation/Operation.swift
karthikkeyan: Resolving this comment since it is addressed

Foundation/Operation.swift
karthikkeyan: Resolving this comment since it with be addressed in future PR

Foundation/Operation.swift
karthikkeyan: Resolving this comment since it is addressed

Foundation/Operation.swift
karthikkeyan: Resolving this comment since it is addressed

Foundation/Operation.swift
karthikkeyan: Resolving this comment since it is addressed

Foundation/Operation.swift
karthikkeyan: Resolving this comment since it is addressed

Foundation/Operation.swift
karthikkeyan: @parkera Please let me know if my comments address you concern.

karthikkeyan:
@spevans I think this PR is ready to merge.

millenomi:
@swift-ci please test

karthikkeyan:
@spevans @compnerd This PR looks ready for merge. Are there any concerns that need to be addressed?

compnerd:
None from me; if @phausler and @millenomi are happy with this, Im good too.

phausler:
Yep, this looks good to me

compnerd:
Ugh, @karthikkeyan - sorry, I didn't look carefully enough at the tests - this broke the Windows build

karthikkeyan:
@compnerd please let me know what I can do to fix this problem. We can even revert this comments until we fix the problem, if need to be.

compnerd:
@karthikkeyan - I just uploaded the test changes to deal with it.  This is more of a note for the future - please use `Thread.sleep(forTimeInterval:)` rather than `sleep`.

PR #1902, opened 2019-02-12
Foundation: remove future fixes

compnerd:
These changes will be needed in the future.  They accidentally snuck
into the previous changes.  They are caused by running a custom build of
clang with changes from the future in the past.


compnerd:
CC: @jakepetroules @spevans

compnerd:

PR #1904, opened 2019-02-13
Process: remove some duplication in RunLoop setup

compnerd:
This was being constructed twice as possibly an accident.  Preserve a
single instance of the constructor.


compnerd:
CC: @phausler @parkera @millenomi @spevans

compnerd:
@swift-ci please test

PR #1907, opened 2019-02-13
Process: port to Windows

compnerd:
This implements the `NSTask` API for Windows.

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -27,6 +27,10 @@ open class FileHandle : NSObject, NSSecureCoding {
 #if os(Windows)
     private var _handle: HANDLE
 
+    internal var handle: HANDLE {
+      return _handle
+    }
+
     @available(Windows, unavailable, message: "Cannot perform non-owning handle to fd conversion")
     open var fileDescriptor: Int32 {
         NSUnsupported()
@@ -49,7 +53,7 @@ open class FileHandle : NSObject, NSSecureCoding {
     private func _checkFileHandle() {
         precondition(_fd >= 0, "Bad file descriptor")
     }
-    
+
     private var _isPlatformHandleValid: Bool {
         return fileDescriptor >= 0
     }
@@ -248,7 +252,7 @@ open class FileHandle : NSObject, NSSecureCoding {
     }
 
 #if os(Windows)
-    public init(handle: HANDLE, closeOnDealloc closeopt: Bool) {
+    internal init(handle: HANDLE, closeOnDealloc closeopt: Bool) {
       _handle = handle
       _closeOnDealloc = closeopt
     }
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -232,8 +232,77 @@ open class Process: NSObject {
         }
     }
 
+    #if os(Windows)
+    private func _socketpair() -> (first: SOCKET, second: SOCKET) {
+      let listener: SOCKET = socket(AF_INET, SOCK_STREAM, 0)
+      if listener == INVALID_SOCKET {
+        return (first: INVALID_SOCKET, second: INVALID_SOCKET)
+      }
+      defer { closesocket(listener) }
+
+      var result: Int32 = SOCKET_ERROR
+
+      var address: sockaddr_in =
+          sockaddr_in(sin_family: ADDRESS_FAMILY(AF_INET), sin_port: USHORT(0),
+                      sin_addr: IN_ADDR(S_un: in_addr.__Unnamed_union_S_un(S_addr: ULONG("127.0.0.1")!)),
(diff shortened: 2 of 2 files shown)


compnerd:
CC: @parkera @phausler @millenomi @jakepetroules @spevans

compnerd:
@swift-ci please test

Foundation/Process.swift
```diff
@@ -232,23 +232,238 @@ open class Process: NSObject {
...
         Process.setup()
-        
+
         // Ensure that the launch path is set
         guard let launchPath = self.executableURL?.path else {
             throw NSError(domain: NSCocoaErrorDomain, code: NSFileNoSuchFileError)
         }
 
+#if os(Windows)
+        // TODO(compnerd) quote the commandline correctly
```
jakepetroules: apple/swift-llvm has an implementation of this, @jmittert should know where.

compnerd: @jakepetroules - apple/swift-llvm?  Why is that not upstream?  I know that llvm has code for unpacking the parameters, I didn't know that it did quoting too.

jmittert: No worries, it's upstream: https://github.com/llvm-mirror/llvm/blob/6b547686c5410b7528212e898fe30fc7ee7a70a3/lib/Support/Windows/Program.inc#L378

Foundation/FileHandle.swift
```diff
@@ -27,6 +27,10 @@ open class FileHandle : NSObject, NSSecureCoding {
 #if os(Windows)
     private var _handle: HANDLE
 
+    open var handle: HANDLE {
```
millenomi: This adds API surface :( I'm sympathetic to a Windows-only way to get a platform handle, but it needs to be 'not API' (either it's `internal`, or it's prefixed with `_` and marked experimental.)

Foundation/Process.swift
```diff
@@ -232,23 +232,238 @@ open class Process: NSObject {
         }
     }
 
+    #if os(Windows)
+    private func _socketpair() -> (first: SOCKET, second: SOCKET) {
```
millenomi: Does CFSocket not work on Windows as desired?

Foundation/FileHandle.swift
millenomi: (And since this is only used internally, perhaps `internal private(set) var _handle: HANDLE` is enough.)

compnerd: It would be nice to have a process handle access for use, but, I can make it `internal` for now.

Foundation/Process.swift
compnerd: There is no `socketpair` on Windows, this emulates the `socketpair` API.

Foundation/Process.swift
compnerd: @millenomi - does Foundation pull in LLVM to link against?

Foundation/FileHandle.swift
pvieito: [This one](
https://github.com/apple/swift-corelibs-foundation/blob/e154db158c38235a403606a2366b67a35b4dfd72/Foundation/FileHandle.swift#L255) should also be marked as `internal` or experimental, I think.

millenomi: I missed that one. Yes, it should be `internal`.

Foundation/Process.swift
millenomi: Ah!

millenomi: Good by me :)

Foundation/FileHandle.swift
compnerd: I got ya, don't worry.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Foundation/Process.swift
millenomi: No, we don’t link LLVM other than the language runtimes. You’ll have to port that code over to use it.

compnerd: I think its reasonable to do that as a follow up, what do you think?

millenomi: Please do so, but yup.

jckarter:
@compnerd It isn't obvious that this caused it, but the Ubuntu 18.04 bot started failing in a `Process` test after picking up this patch:

https://ci.swift.org/job/oss-swift-package-linux-ubuntu-18_04/1099/console

spevans:
I think that test has been intermittent before, maybe Jenkins is somehow setting some extra environment variables although it will be hard to track down as there are no Ubuntu18.04 CI test servers to debug against.

jckarter:
Can the test be disabled or changed to be more robust if it's intermittent?

spevans:
It has been disabled in 
https://github.com/apple/swift-corelibs-foundation/pull/1915

jckarter:
Thanks!

PR #1908, opened 2019-02-13
Create malloc_good_size for platforms other than Linux.

drodriguez:
Other platforms besides Linux do not define malloc_good_size (not even
BSD, it seems), so for everything but the Darwin platforms, create the
trivial malloc_good_size function.

/cc @phausler

### DIFF
--- a/Foundation/Data.swift
+++ b/Foundation/Data.swift
@@ -12,7 +12,7 @@
 
 #if DEPLOYMENT_RUNTIME_SWIFT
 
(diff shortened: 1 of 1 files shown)


spevans:
Looks like its covered by https://github.com/apple/swift-corelibs-foundation/pull/1903/files

drodriguez:
Looks like it. Closing this one. Saleem and me should talk more often.

PR #1909, opened 2019-02-13
[5.0] SR-9322: DateFormatter difference between macOS and Linux

spevans:
- When parsing a date, return nil if the date to be parsed was not
  fully consumed by the parse format.

(cherry picked from commit 8747de5134aba1281d5c641ea1bcb0d6b5f6e31d)


spevans:
@swift-ci test

PR #1910, opened 2019-02-13
Tests for the new FileHandle API

millenomi:
This includes a correctness fix where NSError.isEqual(_:) was not comparing errors the way ObjC NSError does.


millenomi:
@swift-ci please test

PR #1911, opened 2019-02-14
Windows: make _NSErrorWithWindowsError easier to use

compnerd:
`GetLastError` returns a `DWORD` rather than `Int32`.  Adjust the
signature to avoid the need for the explicit cast as it is an internal
helper.


compnerd:
@swift-ci please test

compnerd:
CC: @millenomi

PR #1914, opened 2019-02-14
Supported Scanner API for Swift

millenomi:
Add API on Scanner that's more idiomatic for use in Swift. This promotes some of the existing experimental API and deprecates the rest. The new API is intended to operate on graphemes, not single UTF-16 code units. This API is not experimental and will be supported going forward.

### DIFF
--- a/Foundation/Scanner.swift
+++ b/Foundation/Scanner.swift
@@ -7,8 +7,6 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-
-
 import CoreFoundation
 
 open class Scanner: NSObject, NSCopying {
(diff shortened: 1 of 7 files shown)


millenomi:
@swift-ci please test

millenomi:
@swift-ci Please test source compatibility

millenomi:
… worth a try.

millenomi:
@shahmishal is it possible to set up 'please test source compatibility' for s-c-f?

Foundation/ScannerAPI.swift
```diff
@@ -0,0 +1,198 @@
...
+                index = string._toUTF16Index(scanLocation + delta)
+            }
+            
+            return index
+        }
+        set { scanLocation = string._toUTF16Offset(newValue) }
+    }
+    
+    public func scanInt(representation: NumberRepresentation = .decimal) -> Int? {
+        #if arch(x86_64) || arch(arm64) || arch(s390x) || arch(powerpc64) || arch(powerpc64le)
```
spevans: Would it be better to write this as `if Int.bitWidth == Int64.bitWidth` to avoid having to maintain a list of archs?

millenomi: We use this in `Data` as well just to save the runtime check.

spevans: Wouldnt it be a constant expression and optimised away by the compiler?

millenomi: There were issues with that IIRC, and it may or may not be.

millenomi:
I bet the issue with the test is that one of the emoji isn't recognized as an emoji by ICU 61.

Foundation/ScannerAPI.swift
```diff
@@ -0,0 +1,198 @@
...
+        #error("This architecture isn't known. Add it to the 32-bit or 64-bit line; if the machine word isn't either of those, you need to implement appropriate scanning and handle the potential overflow here.")
+        #endif
+        return nil
+    }
+    
+    public func scanInt32(representation: NumberRepresentation = .decimal) -> Int32? {
+        var value = Int32.max
+        switch representation {
+        case .decimal: guard self.scanInt32(&value) else { return nil }
+        default:
```
spevans: should this be `case .hexadecimal` instead of `.default` ?

millenomi: Good catch, yep.

spevans:
Are `scanUInt(representation:)` and `scanUInt32(representation:)` deliberately  missing? Seems a bit lopsided not to have those.

spevans:
@millenomi because of https://bugs.swift.org/browse/SR-9699 the CI tests arent even building ICU so might be worth getting it fixed now ;)

TestFoundation/TestScanner.swift
```diff
@@ -1,58 +1,483 @@
...
+            expectEqual($0.scanInt64(), 0 as Int64, "Zero")
+            expectEqual($0.scanInt64(), -1 as Int64, "Minus one")
+            expectEqual($0.scanInt64(), -1 as Int64, "Minus one after whitespace")
+            expectEqual($0.scanInt64(), Int64.min, "Min")
+            expectEqual($0.scanInt64(), Int64.max, "Max again after min (no joining it with preceding min even with ignroed whitespace)")
+        }
+        
+        // Overflow:
+        withScanner(for: "\(Int64.max)0") {
+            expectEqual($0.scanInt64(), Int64.max, "Overflow")
```
spevans: So is there anyway to detect if a number being scanned is >`Int64.max` ? Or have a flag to say dont clip to `Int64.max` ?

millenomi: It can't not clip to a Int64 value, since it returns a Int64. This is consistent with the behavior of the older methods.

millenomi:
@spevans it is indeed https://bugs.swift.org/browse/SR-9699 — I pinged @shahmishal about it.

Foundation/Scanner.swift
```diff
@@ -72,6 +70,89 @@ open class Scanner: NSObject, NSCopying {
         _skipSet = Scanner.defaultSkipSet
         _scanLocation = 0
     }
+    
+    // On overflow, the below methods will return success and clamp
+    @discardableResult
+    open func scanInt32(_ result: UnsafeMutablePointer<Int32>) -> Bool {
```
spevans: Am I misreading the Scanner documentation or should all of these  `scanXXX(_ result: UnsafeMutablePointer<XXX>) -> Bool` methods actually take an optional pointer?
https://developer.apple.com/documentation/foundation/scanner/1410914-scanint32

Looks like they may have always been wrong?

millenomi: Yeah, good catch. I'll follow up to correct the signatures.

shahmishal:
> @shahmishal is it possible to set up 'please test source compatibility' for s-c-f?

Yes, can you file a Jira bug?

spevans:
@swift-ci test

PR #1916, opened 2019-02-14
Foundation: fix a couple of compile errors

compnerd:
FoundationErrors does not depend on NSSwiftRuntime and thus does not
transitively get the WinSDK import when not building in whole program
mode.  Add an import to get a definition of `DWORD`.

Remoe the `private(set)` from the computed `processIdentifier` property
which is already readonly due to the definition of the computed
property.


compnerd:
@swift-ci please test

compnerd:
CC: @parkera @phausler @millenomi

millenomi:

PR #1917, opened 2019-02-15
Add additional shims for Operation that mimic the KeyPath<> KVC methods.

millenomi:
Some implementations use these methods for implementing their `Operation`s. We probably still need generalized shims, but these are important to the functionality of the class.


millenomi:

PR #1922, opened 2019-02-16
 Foundation : partial port of NSFileManager for Windows

compnerd:
This leaves a few functions unimplemented and the DirectoryIterator unimplemented as they need to match the semantics of FTS which is not present on Windows.  The remainder of the API surface has been reimplemented in terms of the Windows API.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -28,7 +28,7 @@ open class FileManager : NSObject {
     
     /// Returns an array of URLs that identify the mounted volumes available on the device.
     open func mountedVolumeURLs(includingResourceValuesForKeys propertyKeys: [URLResourceKey]?, options: VolumeEnumerationOptions = []) -> [URL]? {
-        var urls: [URL]
+        var urls: [URL] = []
 
 #if os(Linux)
         guard let procMounts = try? String(contentsOfFile: "/proc/mounts", encoding: .utf8) else {
@@ -41,6 +41,36 @@ open class FileManager : NSObject {
                 urls.append(URL(fileURLWithPath: mountPoint[1], isDirectory: true))
             }
         }
+#elseif os(Windows)
+      var wszVolumeName: UnsafeMutableBufferPointer<WCHAR> = UnsafeMutableBufferPointer<WCHAR>.allocate(capacity: Int(MAX_PATH))
+      defer { wszVolumeName.deallocate() }
+
+      var hVolumes: HANDLE = FindFirstVolumeW(wszVolumeName.baseAddress, DWORD(wszVolumeName.count))
+      guard hVolumes != INVALID_HANDLE_VALUE else { return nil }
+      defer { FindVolumeClose(hVolumes) }
+
+      repeat {
+        var dwCChReturnLength: DWORD = 0
+        GetVolumePathNamesForVolumeNameW(wszVolumeName.baseAddress, nil, 0, &dwCChReturnLength)
+
+        var wszPathNames: UnsafeMutableBufferPointer<WCHAR> = UnsafeMutableBufferPointer<WCHAR>.allocate(capacity: Int(dwCChReturnLength + 1))
+        defer { wszPathNames.deallocate() }
+
+        if GetVolumePathNamesForVolumeNameW(wszVolumeName.baseAddress, wszPathNames.baseAddress, DWORD(wszPathNames.count), &dwCChReturnLength) == FALSE {
+          // TODO(compnerd) handle error
+          continue
+        }
+
+        var pPath: DWORD = 0
+        repeat {
+          let path: String = String(decodingCString: wszPathNames.baseAddress! + Int(pPath), as: UTF16.self)
+          if path.length == 0 {
+            break
+          }
+          urls.append(URL(fileURLWithPath: path, isDirectory: true))
+          pPath += DWORD(path.length + 1)
+        } while pPath < dwCChReturnLength
+      } while FindNextVolumeW(hVolumes, wszVolumeName.baseAddress, DWORD(wszVolumeName.count)) != FALSE
 #elseif canImport(Darwin)
 
         func mountPoints(_ statBufs: UnsafePointer<statfs>, _ fsCount: Int) -> [URL] {
@@ -184,21 +214,22 @@ open class FileManager : NSObject {
             return [ URL(fileURLWithPath: all, isDirectory: true, relativeTo: URL(fileURLWithPath: NSHomeDirectory(), isDirectory: true)) ]
         }
     }
-    
-    #if os(Windows) // Non-Apple OSes that do not implement FHS/XDG are not currently supported.
-    @available(*, unavailable, message: "Not implemented for this OS")
-    open func urls(for directory: SearchPathDirectory, in domainMask: SearchPathDomainMask) -> [URL] {
-        NSUnimplemented()
-    }
-    #else
+
     /* -URLsForDirectory:inDomains: is analogous to NSSearchPathForDirectoriesInDomains(), but returns an array of NSURL instances for use with URL-taking APIs. This API is suitable when you need to search for a file or files which may live in one of a variety of locations in the domains specified.
      */
     open func urls(for directory: SearchPathDirectory, in domainMask: SearchPathDomainMask) -> [URL] {
         let domains = _SearchPathDomain.allInSearchOrder(from: domainMask)
-        
+
+        var urls: [URL] = []
+
+#if os(Windows)
+        for domain in domains {
+          urls.append(contentsOf: windowsURLs(for: directory, in: domain))
+        }
+#else
         // We are going to return appropriate paths on Darwin, but [] on platforms that do not have comparable locations.
         // For example, on FHS/XDG systems, applications are not installed in a single path.
-        
+
         let useDarwinPaths: Bool
         if let envVar = ProcessInfo.processInfo.environment["_NSFileManagerUseXDGPathsForDirectoryDomains"] {
             useDarwinPaths = !NSString(string: envVar).boolValue
@@ -209,9 +240,7 @@ open class FileManager : NSObject {
                 useDarwinPaths = false
             #endif
         }
-        
-        var urls: [URL] = []
-        
+
         for domain in domains {
             if useDarwinPaths {
                 urls.append(contentsOf: darwinURLs(for: directory, in: domain))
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @millenomi @parkera @phausler @jakepetroules @spevans

Foundation/FileManager.swift
```diff
@@ -209,20 +240,93 @@ open class FileManager : NSObject {
...
     }
-    #endif
+
+#if os(Windows)
+    private class func getURL(for id: KNOWNFOLDERID) -> URL {
+      var pszPath: PWSTR?
+      _ = withUnsafePointer(to: id) { id in
+        SHGetKnownFolderPath(id, DWORD(KF_FLAG_DEFAULT.rawValue), nil, &pszPath)
+      }
+      let url: URL = URL(fileURLWithPath: String(decodingCString: pszPath!, as: UTF16.self), isDirectory: true)
```
millenomi: Does SHGetKnownFolderPath always return nonnil?

Foundation/FileManager.swift
```diff
@@ -209,20 +240,93 @@ open class FileManager : NSObject {
...
+        return []
+
+      case .desktopDirectory:
+        return [FileManager.getURL(for: FOLDERID_Desktop)]
+
+      case .documentDirectory:
+        return [FileManager.getURL(for: FOLDERID_Documents)]
+
+      case .cachesDirectory:
+        // FIXME(compnerd) where should this go?
```
millenomi: This should be a temporary directory that is regularly cleaned up by the OS.

Foundation/FileManager.swift
```diff
@@ -209,20 +240,93 @@ open class FileManager : NSObject {
...
+      _ = withUnsafePointer(to: id) { id in
+        SHGetKnownFolderPath(id, DWORD(KF_FLAG_DEFAULT.rawValue), nil, &pszPath)
+      }
+      let url: URL = URL(fileURLWithPath: String(decodingCString: pszPath!, as: UTF16.self), isDirectory: true)
+      CoTaskMemFree(pszPath)
+      return url
+    }
+
+    private func windowsURLs(for directory: SearchPathDirectory, in domain: _SearchPathDomain) -> [URL] {
+      guard domain == .user else { return [] }
```
millenomi: The Local domain can be \ProgramData, perhaps? As long as it’s a directory or set of directories all users can read from.

Foundation/FileManager.swift
```diff
@@ -209,20 +240,93 @@ open class FileManager : NSObject {
...
+
+      case .documentDirectory:
+        return [FileManager.getURL(for: FOLDERID_Documents)]
+
+      case .cachesDirectory:
+        // FIXME(compnerd) where should this go?
+        return []
+
+      case .applicationSupportDirectory:
+        // FIXME(compnerd) where should this go?
```
millenomi: This should be a directory where application can store auxiliary data that is not deleted by the OS. This is almost certainly somewhere inside %APPDATA%.

Foundation/FileManager.swift
```diff
@@ -646,22 +820,51 @@ open class FileManager : NSObject {
...
             }
+#endif
         })
         return contents
     }
-    
+
+
+#if os(Windows)
+    private func GetFileAttributes(atPath path: String) throws -> WIN32_FILE_ATTRIBUTE_DATA {
```
millenomi: windowsFileAttributes(atPath:)?

Foundation/FileManager.swift
```diff
@@ -209,20 +240,93 @@ open class FileManager : NSObject {
...
         }
-        
+#endif
+
         return urls
     }
-    #endif
+
+#if os(Windows)
+    private class func getURL(for id: KNOWNFOLDERID) -> URL {
```
millenomi: url(for:)?

Foundation/FileManager.swift
```diff
@@ -1080,7 +1417,10 @@ open class FileManager : NSObject {
         guard alreadyConfirmed || shouldRemoveItemAtPath(path, isURL: isURL) else {
             return
         }
-        
+
+#if os(Windows)
```
millenomi: Want to add an `@availability(Windows, deprecated, message: “Not yet implemented”)` to the method?

Foundation/FileManager.swift
```diff
@@ -1289,9 +1684,13 @@ open class FileManager : NSObject {
             // Return true as the best guess.
             return true
         })
+#endif
     }
 
     private func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64, bufSize: Int) -> Bool {
+#if os(Windows)
```
millenomi: Maybe this needs to be written in terms of FileHandle?

Foundation/FileManager.swift
```diff
@@ -1391,15 +1794,23 @@ open class FileManager : NSObject {
         }
         return statInfo
     }
+#endif
 
     internal func _permissionsOfItem(atPath path: String) throws -> Int {
+#if os(Windows)
```
millenomi: I’d rather have a `@availability(Windows, unavailable…)` here.

Foundation/FileManager.swift
```diff
@@ -1391,15 +1794,23 @@ open class FileManager : NSObject {
...
+#else
         let fileInfo = try _lstatFile(atPath: path)
         return Int(fileInfo.st_mode & 0o777)
+#endif
     }
 
     /* -contentsEqualAtPath:andPath: does not take into account data stored in the resource fork or filesystem extended attributes.
      */
     open func contentsEqual(atPath path1: String, andPath path2: String) -> Bool {
+#if os(Windows)
```
millenomi: Same. (Use `deprecated` for public methods.)

Foundation/FileManager.swift
```diff
@@ -1872,12 +2309,25 @@ extension FileManager {
             guard let url = o as? URL else {
                 return nil
             }
+
+#if os(Windows)
```
millenomi: Same.

Foundation/FileManager.swift
compnerd: It returns a HRESULT, we are only interested in the value of `pszPath` which is an out variable.

Foundation/FileManager.swift
compnerd: Hmm, so, this should just return `%TEMP%`?

Foundation/FileManager.swift
compnerd: `\ProgramData` is `%ALLUSERSPROFILE%`, its readable by all.

Foundation/FileManager.swift
compnerd: Sounds like `FOLDERID_LocalAppData` fits the bill.

Foundation/FileManager.swift
compnerd: Sure.

Foundation/FileManager.swift
compnerd: Sure

Foundation/FileManager.swift
compnerd: This is possible to port, but, I didn't write the mapping.

spevans: This is just a shortcut function I added to avoid calling `attributesOfItem(atPath:)` which can be quite heavy if the only thing needed is the file permissions and is really just a shortcut version of `attributesOfItem(atPath:)[.posixPermissions].intValue`

I notice that you dont support `.posixPermissions` at the moment, I guess you could always just return the permissions as somthing like:

```
var permissions: mode_t = isReadableFile(atPath: path) ? 0o400 : 0
permissions |= isWritableFile(atPath: path) ? 0o200 : 0
```

Foundation/FileManager.swift
compnerd: Oh, wait, I think I get what you mean.

Foundation/FileManager.swift
compnerd: Yeah, I think that would be a nice cleanup.

Foundation/FileManager.swift
millenomi: I'd rather not introduce any new potential crash paths as much as possible. If you don't support POSIX permissions yet, just don't return them.

Foundation/FileManager.swift
millenomi: Yeah, that serves pretty much the same function as `/Library`, or the Darwin Library in the local domain.

Now, this doesn't _have_ the same guarantees as /Library's structure so it's not appropriate to return as the path for `.libraryDirectory`, IMHO, but we can return a local domain app support path in there, for example.

Foundation/FileManager.swift
millenomi: If that's the best we can do, yeah.

Foundation/FileManager.swift
millenomi: I mean: is the out variable _always_ set? You're not checking for success or failure here.

compnerd: Right, the outer variable is the `HRESULT`.  True, I'm not checking for the error case, since the two known failure cases are: ENOMEM, in which case, death to all!  or the parameter validation failed.  I suppose I can check for the latter if it makes it seem better.

Foundation/FileManager.swift
compnerd: Okay, I'll sink the guard into the various cases and add the local domain for the app data

Foundation/FileManager.swift
compnerd: I think that crash paths are preferable to silent failures.  An explicit crash makes it easier to track down the issues once we get stuff running :-).

millenomi: _sigh_ okay, since all `NSUnimplemented`'s are `#if os(Windows)` and so _you_ have to deal with them. Just remember that any crash point you add is going to be a landmine in the codebase for a long while.

millenomi:
@swift-ci please test

Foundation/FileManager.swift
compnerd: Yeah, that's why I'm trying to keep that to a minimum and trying to implement as much as possible.  I've also annotated the functions as deprecated to try to catch them at build time if possible.

compnerd:
*sigh* time to fix this I guess

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

Foundation/FileManager.swift
compnerd: I'll add an assertion (`precondition`)

Foundation/FileManager.swift
compnerd: SR-9979

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1923, opened 2019-02-16
FileHandle: Implement .readabilityHandler and .writeabilityHandler.

spevans:


spevans:
@swift-ci test

spevans:
@millenomi @phausler Does this look like the right approach for the readability, writability handlers? If so I will add some more tests to this PR.

millenomi:
I have an in-flight patch that also does the notification methods, and it’s tricky to not do them all at once. Want to compare notes once I put it up on Monday?

PR #1928, opened 2019-02-19
Speedup for NSString -getCharacters:range:

Catfish-Man:


Catfish-Man:
@swift-ci please test

milseman:
Do you have a benchmark for this? UBP's subscript is probably too slow, I've found it significantly faster to get the underlying UP and dereference that.

Catfish-Man:
Heh whoops. A little TOO untested there. Will fix it up later.

@milseman it shows up on the regex test in the benchmark game

gottesmm:
@Catfish-Man Do we have a version of this benchmark? (We should probably add it).

gottesmm:
Maybe we even add it before this PR and then we can use the PR to validate the speed up! = )

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

Catfish-Man:
@swift-ci please test

Catfish-Man:
@swift-ci please test

Catfish-Man:
I whipped up a "simulated" benchmark for this since I don't have a Linux machine handy, and on my test at least the current patch is >10x faster than the old implementation.

Unfortunately we were also seeing crashes in NSKeyedArchiver tests with the previous version of it… we'll see how this one does.

millenomi:
We changed how testing occurs for SCF patches. Restarting test.

millenomi:
@swift-ci please test

Catfish-Man:
Hm. I wonder what "llvm-config: error: component libraries and shared library" means (from the linux build)

spevans:
@Catfish-Man When running your `unicopy` branch under Xcode 10.2b3 with `swift-DEVELOPMENT-snapshot-2019-02-19` I see the same crash in `TestNSKeyedArchiver.test_archive_locale` as in the Linux logs: 

EXC_BAD_ACCESS in NSString.swift line 349

Catfish-Man:
Oh excellent, thank you

Catfish-Man:
Ok, finally got myself set up properly to do corelibs debugging rather than stdlib, and I see what's wrong here. Fix shortly, thank you for your patience :)

Catfish-Man:
Should be good to go now. It wasn't handling _NSCFConstantString.

Catfish-Man:
@swift-ci please test

Foundation/NSString.swift
```diff
@@ -343,8 +343,18 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
 
 extension NSString {
     public func getCharacters(_ buffer: UnsafeMutablePointer<unichar>, range: NSRange) {
-        for idx in 0..<range.length {
-            buffer[idx] = character(at: idx + range.location)
+        if type(of: self) == NSString.self || type(of: self) == NSMutableString.self {
```
milseman: How much does this type check slow things down? I suppose anything is better than `character(at:)`...

Catfish-Man: There's a lot of things we could do to improve this further. My goal here is just to get it from n^2 to n so that it's not a huge problem. If we find important uses of it we can sink more time into it.

Catfish-Man:
@swift-ci please test

milseman:
You can also get up to 2x improvement on contiguous UTF-8 strings in general and several X when it's ASCII, depending on how many code paths you want to maintain here.

Pseudo-code (you'll have to ~~struggle with~~ handle optionality, because optional interfaces are ~~a bad idea~~ annoying) for a dedicated ASCII path .

```swift
// Fast-path Handle ASCII prefixes
_storage.utf8.withContiguousStorageIfAvailable {
  for byte in $0 {
    guard byte < 0x80 else { return }
    output[i] = UInt16(truncatingIfNeeded: byte)
  }
}

// Now we calculate the index corresponding to our (potentially further advanced) offsets...
```

PR #1866, opened 2019-01-30
Add ICU version check to ensure minimum of version 61 is used.

spevans:

PR #1929, opened 2019-02-19
[4.2]  SR-1464: NSNumber.description is not compatible between OS X and linux

spevans:
- Use Swift's stringification of numbers for .description and .stringValue
    
- Use String(format:locale:...) with the correct format for the
   NSNumber type if locale is not nil.
    
- Enable use of __CFStringFormatLocalizedNumber() on
   DEPLOYMENT_TARGET_LINUX.
    
(cherry picked from commit 6450221c7fea77b8c458ca7baac953cc939c41ae)

- Add NSDecimalNumber.description
    
(cherry picked from commit 7cb608f2d0d666223f5fd340eaa686c769b30bb4)

### DIFF
--- a/CoreFoundation/String.subproj/CFString.c
+++ b/CoreFoundation/String.subproj/CFString.c
@@ -5996,7 +5996,7 @@ enum {
     CFFormatDummyPointerType = 42	/* special case for %n */
 };
 
-#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_WINDOWS
+#if DEPLOYMENT_TARGET_MACOSX || DEPLOYMENT_TARGET_EMBEDDED || DEPLOYMENT_TARGET_WINDOWS || DEPLOYMENT_TARGET_LINUX
 /* Only come in here if spec->type is CFFormatLongType or CFFormatDoubleType. Pass in 0 for width or precision if not specified. Returns false if couldn't do the format (with the assumption the caller falls back to unlocalized).
 */
 static Boolean __CFStringFormatLocalizedNumber(CFMutableStringRef output, CFLocaleRef locale, const CFPrintValue *values, const CFFormatSpec *spec, SInt32 width, SInt32 precision, Boolean hasPrecision) {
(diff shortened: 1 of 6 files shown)


spevans:
@swift-ci test 4.2

weissi:
@swift-ci test 4.2

weissi:
@swift-ci test 4.2

futurejones:
As reported in https://bugs.swift.org/browse/SR-9985 since the addition of this PR
`TestNSNumber.test_descriptionWithLocale` and `TestNSString.test_initializeWithFormat3`
are now failing on Linux/Ubuntu/AArch64

weissi:
@futurejones thanks for reporting! So this PR passed all the tests here, are the ones you mention disabled? 

Could this (on swift5) maybe be an artefact of vendoring ICU on Swift 5?

We definitely need to sort this out before releasing. @spevans / @kevints you know more about the ICU interactions here. Any insight?

spevans:
Its odd that the swift-5.0-branch fails because none of the differences between 5.0 and master should matter. Also @futurejones tested 5.0 with #1866 which is a specific check for ICU >= 61, which is the only thing that could possibly make any difference so I cant see why that would fail either. Ubuntu18 comes with ICU60 so it must be getting built.

futurejones:
@weissi These 2 tests `TestNSNumber.test_descriptionWithLocale` and `TestNSString.test_initializeWithFormat3` were passing in 4.2.1 RELEASE and as far as I know they were passing in the 4.2.2 RELEASE. I am currently doing a 4.2.2 RELEASE build to verify this.
They are now failing in the current 4.2-branch. @spevans indicated that this PR may be the cause.
I have been testing on Ubuntu 16.04 only.

futurejones:
To clarify these tests were passing on a build done from the swift-4.2-branch on 2019-02-21.
```
Test Case 'TestNSNumber.test_descriptionWithLocale' started at 2019-02-25 10:19:02.135
Test Case 'TestNSNumber.test_descriptionWithLocale' passed (0.0 seconds)
```
```
Test Case 'TestNSString.test_initializeWithFormat2' passed (0.0 seconds)
Test Case 'TestNSString.test_initializeWithFormat3' started at 2019-02-25 10:19:06.142
```

spevans:
@weissi I think we should revert this one for now. I cant work out why it is breaking but it could just be an arm specific build issue. However the `swift-5.0-branch` breaking v `master` working is more worrying as I cant see any difference between the 2 branches wrt this change. But I think that would need to be looked at first.

weissi:
thanks for looking into this @spevans .

weissi:
@spevans I created https://github.com/apple/swift-corelibs-foundation/pull/1950 to revert this

spevans:
Further testing found that the fix for `signed char` in #1871 fixes this issue on ARM64 - the bug was that localised formatting was being disabled. So it should be ok to reintroduce after 4.2.3 has been released.

PR #1931, opened 2019-02-20
DateIntervalFormatter: Basic Implementation

spevans:
- Set the dateTemplate from the dateStyle and timeStyle.
    
- Cache the formatter unless one of the properties changes.

- Add wrapper functions to work around ICU function rename

- Update tests for differences with Apple's ICU

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -52,6 +52,8 @@
 #include <malloc/malloc.h>
 #endif
 
+#include <unicode/udateintervalformat.h>
+
 _CF_EXPORT_SCOPE_BEGIN
 
 CF_CROSS_PLATFORM_EXPORT Boolean _CFCalendarGetNextWeekend(CFCalendarRef calendar, _CFCalendarWeekendRange *range);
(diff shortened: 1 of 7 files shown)


spevans:
@swift-ci test

Foundation/DateIntervalFormatter.swift
```diff
@@ -57,7 +108,109 @@ open class DateIntervalFormatter : Formatter {
...
+
+        if _timeStyle == .none { return }
+
+        if _dateStyle == .short {
+            dateTemplate += ", "
+        } else if _dateStyle != .none {
+            dateTemplate += " 'at' "
+        }
+
+        switch _dateStyle {
```
e87l: _timeStyle?

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -484,6 +486,25 @@ static inline unsigned int _dev_minor(dev_t rdev) {
 }
 #endif
 
+// ICU function can be renamed on some platforms eg on Linux udtivfmt_open() -> udtitvfmt_open_61_swift() so wrap the functions
+// to provide a platform independant interface. These functions are used by DateIntervalFormatter.swift
```
e78l: 'independent' but also not sure what the comment about means... we can use version-specific ICU functions? But we are not?

Foundation/DateIntervalFormatter.swift
spevans: Good spot!

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
spevans: The ICU libraries  on Linux are built using the `--enable-renaming --with-library-suffix=swift` options which renames functions to include the library version and optional suffix to avoid symbol clash  allowing multiple versions of the library to be linked in at the same time

e78l: Ah okay thanks for the explanation
Didn't/not seeing the udtitvfmt_open_61_swift call but I suppose it's macro'd somewhere... not obvious online :-)

spevans: Yeah, there is a `<unicode/urename.h>` that  has 
```
#define udtitvfmt_close U_ICU_ENTRY_POINT_RENAME(udtitvfmt_close)
#define udtitvfmt_format U_ICU_ENTRY_POINT_RENAME(udtitvfmt_format)
#define udtitvfmt_open U_ICU_ENTRY_POINT_RENAME(udtitvfmt_open)
``` 

etc

spevans:
@swift-ci test

spevans:
Updated to use a `DateFormatter` to generate the template however this has now shown up some issues with `DateFormatter` where the `.dateFormat` doesnt update if the `.locale` is changed so that needs to be fixed first.

PR #1936, opened 2019-02-22
CoreFoundation: switch to `_WINDLL` from `_USRDLL`

compnerd:
`_USRDLL` is used to indicate that you are building a MFC DLL while
`_WINDLL` is specified for **ALL** DLL builds. Use the correct macro.


compnerd:

PR #1938, opened 2019-02-23
build: link in dependencies on Windows

compnerd:
This sets up the dependencies we need on Windows.  It helps reduce the
spew from the linker when building for windows.


compnerd:
CC: @millenomi @phausler @parkera @jakepetroules

compnerd:
@swift-ci please test

PR #1940, opened 2019-02-23
Foundation: adjust `NSTemporaryDirectory` for Windows

compnerd:
Windows provides `GetTempPath` for getting the temporary directory which
consults the environment variables and falls back through a cascading
set ultimately ending up at `%SystemRoot%\Temp` if nothing else is
found.  The last path is not exactly great, but, we do not expect to end
up there as modern Windows will use `%UserProfile%\Local Settings`
before that.

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -10,6 +10,14 @@
 import CoreFoundation
 
 public func NSTemporaryDirectory() -> String {
+#if os(Windows)
+    let cchLength: DWORD = GetTempPathW(0, nil)
+    var wszPath: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(cchLength + 1))
+    guard GetTempPathW(DWORD(wszPath.count), &wszPath) == cchLength else {
+      precondition(false, "GetTempPathW mutation race")
+    }
+    return String(decodingCString: wszPath, as: UTF16.self)
+#else
 #if canImport(Darwin)
     var length: Int = confstr(_CS_DARWIN_USER_TEMP_DIR, nil, 0)
     if length > 0 {
@@ -28,6 +36,7 @@ public func NSTemporaryDirectory() -> String {
         }
     }
     return "/tmp/"
+#endif
 }
 
 extension String {


compnerd:
CC: @millenomi @parkera @phausler @jakepetroules

compnerd:
@swift-ci please test

Foundation/NSPathUtilities.swift
```diff
@@ -10,13 +10,19 @@
 import CoreFoundation
 
 public func NSTemporaryDirectory() -> String {
-    #if os(macOS) || os(iOS)
+#if os(macOS) || os(iOS)
     var buf = [Int8](repeating: 0, count: 100)
```
gwynne: I really feel like this buffer should be dynamically allocated, as with the Windows version. `confstr()` does have the “returns the needed size” functionality.

compnerd: I think that's beyond the scope of this change, but I can do that in a different change.

gwynne:
The returned value should really be cached on all platforms, I think... I don’t know what the expected behavior of the function is if the env vars change between calls. @millenomi ?

compnerd:

compnerd:
Ugh, this is wrong.  This changed the semantics for the macOS target, as it would previously fall back to the TMP env var in the truncation case.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Foundation/NSPathUtilities.swift
```diff
@@ -10,13 +10,20 @@
...
 public func NSTemporaryDirectory() -> String {
-    #if os(macOS) || os(iOS)
+#if os(Windows)
+    var cchLength: DWORD = GetTempPathW(0, nil)
+    var wszPath: UnsafeMutableBufferPointer = UnsafeMutableBufferPointer<WCHAR>.allocate(capacity: Int(cchLength + 1))
+    defer { wszPath.deallocate() }
+    _ = GetTempPathW(cchLength + 1, wszPath.baseAddress)
+    return String(decodingCString: wszPath.baseAddress!, as: UTF16.self)
+#else
+#if os(macOS) || os(iOS)
```
millenomi: This is absolutely where `canImport(Darwin)` should go, since the point of this block is t guard from using API that exists in that module.

compnerd: I still hate `canImport(Darwin)`, but fine.  Although, I think that it should be done in #1941 rather than in this change.  I'll update that in a little while.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1941, opened 2019-02-23
Follow up for changes requested in #1940

compnerd:


compnerd:
@swift-ci please test

compnerd:
CC: @gwynne @millenomi

Foundation/NSPathUtilities.swift
```diff
@@ -10,13 +10,22 @@
...
-    #if os(macOS) || os(iOS)
-    var buf = [Int8](repeating: 0, count: 100)
-    let r = confstr(_CS_DARWIN_USER_TEMP_DIR, &buf, buf.count)
-    if r != 0 && r < buf.count {
-        return String(cString: buf, encoding: .utf8)!
-    }
-    #endif
+#if os(macOS) || os(iOS)
+    var length: Int = confstr(_CS_DARWIN_USER_TEMP_DIR, nil, 0)
+    guard length > 0 else { return "/tmp" }
```
spevans: Should this be `{ return "/tmp/" }` as it looks like trailing `/` are appended to the result.

gwynne: Yep, good catch.

compnerd:
@gwynne - I would like to get @millenomi to okay this as well since this has a subtle difference in behaviour (we will never go down the path of the env var check).

gwynne:
@compnerd 👍 Definitely want @millenomi 's sign-off on this one.

Foundation/NSPathUtilities.swift
```diff
@@ -10,13 +10,22 @@
...
 
 public func NSTemporaryDirectory() -> String {
-    #if os(macOS) || os(iOS)
-    var buf = [Int8](repeating: 0, count: 100)
-    let r = confstr(_CS_DARWIN_USER_TEMP_DIR, &buf, buf.count)
-    if r != 0 && r < buf.count {
-        return String(cString: buf, encoding: .utf8)!
-    }
-    #endif
+#if os(macOS) || os(iOS)
```
millenomi: I’m okay not fighting over this line since the old patch had it still, but it still is incorrect per the guidelines we’ve discussed.

Foundation/NSPathUtilities.swift
```diff
@@ -10,13 +10,22 @@
...
+    var buffer: [Int8] = Array<Int8>(repeating: 0, count: length)
+    length = confstr(_CS_DARWIN_USER_TEMP_DIR, &buffer, buffer.count)
+    precondition(length > 0 && length < buffer.count,
+                 "_CS_DARWIN_USER_TEMP_DIR mutation race")
+    return String(cString: buffer, encoding: .utf8)!
+#elseif os(Windows)
+    var cchLength: DWORD = GetTempPathW(0, nil)
+    var wszPath: UnsafeMutableBufferPointer = UnsafeMutableBufferPointer<WCHAR>.allocate(capacity: Int(cchLength + 1))
+    defer { wszPath.deallocate() }
+    _ = GetTempPathW(cchLength + 1, wszPath.baseAddress)
```
millenomi: The docs say this function can fail. You should check the return value and fall back to the env variable path.

millenomi: https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-gettemppathw

compnerd: We just invoked it earlier and checked the result (to determine the buffer size to ensure that we have the properly sized buffer.  The second invocation has already gone through the environment and failing that has returned `%SystemRoot%\Temp` (aka `C:\Windows\Temp`).

Foundation/NSPathUtilities.swift
compnerd: Well, I'll still express my disagreement and make the change (I just want to make that change here rather than in #1940 which is for Windows ... this PR has an extra patch that isn't mean to be there).

Foundation/NSPathUtilities.swift
gwynne: TOCTOU race - the environment may have been changed by another thread between the two calls.

Foundation/NSPathUtilities.swift
gwynne: FWIW, I'm not terribly delighted with either solution - but I do think `canImport(Darwin)` is slightly better than a parade of `os()` disjunctions for the time being.

Foundation/NSPathUtilities.swift
compnerd: Yeah, I have an assertion in the Darwin path for exactly that.  I suppose that I can add a second check here too.  But, that is going into #1940.

millenomi:

millenomi:

gwynne:
Holding off until I can push an additional correction to the Darwin error handling.

gwynne:
(Sorry for the spam about reviews, I'm still learning what some of these buttons do 😓)

PR #1942, opened 2019-02-23
Foundation: remove deprecated `bzero`

compnerd:
The `bzero` implementation is unnecessary as it is trivially
implementable by the "modern" `memset` API.  Additionally, the use of
the type `size_t` causes trouble for Windows which tries to interpret it
as `Foundation.size_t` rather than `ucrt.size_t` which results in a
horrific deserialisation fatalError.


compnerd:
CC: @gwynne @millenomi

compnerd:
@swift-ci please test

compnerd:

PR #1943, opened 2019-02-23
plutil: make `plutil` build on Windows

compnerd:
plutil uses the C runtime, make sure that we import `MSVCRT`.


compnerd:

PR #1944, opened 2019-02-23
Make PropertyListSerialization.MutabilityOptions constants public

keith:
These are meant to be public so that they can be passed when reading a
plist. Previously they were accessible on macOS, but not Linux.


dduan:
@swift-ci test

dduan:
lgtm

millenomi:
Whoops.

keith:
Thanks!

PR #1946, opened 2019-02-25
build: ensure that we build CoreFoundation for re-export

compnerd:
We wish to re-export CoreFoundation from Foundation.  In order to
accomplish this, the objects need to be built as if they are being built
for a DLL as they will be subsumed into the Foundation DLL.  We do not
necessarily wish to do this if we are building Foundation static
however.  Doing this ensures that we can re-export the symbols from
CoreFoundation on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,7 @@ ExternalProject_Add(CoreFoundation
                       -DBUILD_SHARED_LIBS=NO
                       -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
                       -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                      $<$<BOOL:BUILD_SHARED_LIBS>:-DCMAKE_C_FLAGS=-D_WINDLL>
                       -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
                       -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
                       -DCMAKE_INSTALL_LIBDIR=usr/lib


compnerd:
@swift-ci please test

compnerd:
The problem with that is that because it is an external project, we cant do a `TARGET_OBJECTS` to get the objects and statically link the library.  We want to have the ability to build the CoreFoundation project as shared or static, and in the case of Foundation, statically with shared semantics.  I guess I can add another flag.

parkera:
We haven't been sure we wanted to make CF "public" API on other platforms via Swift. It's not a great Swift API. I'd rather people use it as a C library.

compnerd:
@parkera - am I misunderstanding something, or are you recommending that we dynamically link to CoreFoundation then?  If so, that would solve the problem.

parkera:
I'm just curious what the desired effect is here. If, from Swift, you `import Foundation`, do you want API like `CFStringCreate` to be available?

compnerd:
@parkera - oh, the desired effect here is that users (i.e. XCTest) can use the CoreFoundation APIs just like they currently can on Darwin and Linux (as @millenomi has stated is the desired behaviour)

gwynne:
Another flag seems like the best option; overriding `CMAKE_C_FLAGS` will just leave us with the same Linux build failures as last time 😕

millenomi:
I talked with @parkera and @compnerd.

We want to limit how much people use CF; we want them to use the Foundation API instead, and we already prevent CF symbols from being imported by `import Foundation` on Linux.

I have proposed to @compnerd that, since Windows is a new platform here, if it requires going out of our way to achieve export, that we just don't on that platform, and we correct the XCTest issues by providing SPI for their use only. If we determine downstream support requires CF for interesting packages, we can always revisit. The idea here is to keep the status quo on Linux, so we don't break existing downstream code.

millenomi:
Addedum: @compnerd: can we limit XCTest SPI presence to Windows for now?

compnerd:
Closing this off, I'll add the SPI in a different PR; this was motivated for XCTest, so, no longer need this.

PR #1947, opened 2019-02-26
Revert "Add ICU version check to ensure minimum of version 61 is used."

dcci:
This reverts commit 021533bcc19e49cb7f373ed096caa1adfd03b862.
It broke lldb PR testing, as the version of `ubuntu` on the builders is 55.
It looks like I have an `icu` directory in my checkout, so I think the build might have moved to build its own version of `libicu` (although I'm not confident about this).

cc ing: @spevans

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,12 +20,6 @@ find_package(CURL REQUIRED)
 find_package(ICU COMPONENTS uc i18n REQUIRED)
 find_package(LibXml2 REQUIRED)
 
-set(ICU_MINIMUM_VERSION 61)
-if(ICU_VERSION VERSION_LESS ${ICU_MINIMUM_VERSION})
-  message(FATAL_ERROR "ICU version ${ICU_VERSION} is less than required version ${ICU_MINIMUM_VERSION}")
-endif()
-
-
 include(SwiftSupport)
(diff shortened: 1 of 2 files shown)


dcci:
@swift-ci test and merge

dcci:
@swift-ci test and merge

dcci:
@shahmishal can you please trigger a build? I'm not able to.

dcci:
@swift-ci please test

dcci:
@swift-ci please test

spevans:
@dcci So I had a look at the lldb PR testing on `ci.swift.org` and it looks to be using this command:

`swift/utils/build-script --foundation --libdispatch --test --release --lldb -- --skip-test-cmark --skip-test-swift --lldb-test-swift-only --skip-test-foundation=1`

For Linux we need to add the `--libicu` option so that the `icu` directory is built to bring it upto the same version as Foundation uses. This would make the full command

`swift/utils/build-script --libicu --foundation --libdispatch --test --release --lldb -- --skip-test-cmark --skip-test-swift --lldb-test-swift-only --skip-test-foundation=1`

This PR was designed to find these edge cases, its unfortunate that there are so many different build commands used by CI

(The alternative would be to remove the `--foundation` if thats possible)

PR #1949, opened 2019-02-26
Foundation: add SPI for XCTest

compnerd:
Foundation does not re-expose CoreFoundation interfaces on Windows.
This is desirable on Linux as well, however, in the interest of not
breaking existing applications, we have not yet done this.  For now
interfaces required are only available on Windows.

### DIFF
--- a/Foundation/NSDate.swift
+++ b/Foundation/NSDate.swift
@@ -165,6 +165,15 @@ open class NSDate : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
 }
 
+// SPI for XCTest
+#if os(Windows)
+extension NSDate {
+  public static func _currentAbsoluteTime() -> TimeInterval {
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

millenomi:

Foundation/NSDate.swift
```diff
@@ -165,6 +165,15 @@ open class NSDate : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
 }
 
+// SPI for XCTest
+#if os(Windows)
+extension NSDate {
+  public static func _currentAbsoluteTime() -> TimeInterval {
+    return CFAbsoluteTimeGetCurrent()
```
spevans: Isnt this just `Date.timeIntervalSinceReferenceDate` ? If so we could just fixup XCTest to use that directly

PR #1951, opened 2019-02-26
Better handle possible confstr() errors

gwynne:
In `NSTemporaryDirectory()`, `confstr()` is used to retrieve `_CS_DARWIN_USER_TEMP_DIR`. However, it does so in what is technically an incorrect fashion. This PR adds additional error checking to better diagnose various extremely rare and unlikely error conditions. If nothing else, it's a decent example of how finicky POSIX APIs can be.


millenomi:

PR #1958, opened 2019-03-01
[4.2] Fix PropertyListSerialization not doing a roundtrip of Data correctly.

millenomi:
In 4.x, bridging `Data` to an object produces a `_NSSwiftData`, a `NSData` subclass. `CFData…` functions weren’t set up for it, so now they must be. Unfortunately, due to the way the implementation of s-c-f differs from Darwin’s — since we have no class clusters in Swift — there’s some wrangling needed in NSMutableData to prevent infinite recursion when its base class’s methods are invoked. rdar://48420798

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -239,6 +239,13 @@ struct _NSNumberBridge {
     bool (*_Nonnull _getValue)(CFTypeRef number, void *value, CFNumberType type);
 };
 
+struct _NSDataBridge {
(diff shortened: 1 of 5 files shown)


millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
cc @shahmishal 4.2 macOS building seems broken in a way that's independent of this patch?

spevans:
@swift-ci test 4.2

shahmishal:
@millenomi 4.2 is only supported with 10.1 Xcode for this reason, I added a new PR testing config.  (test 4.2)

millenomi:
Whoops, thanks.

PR #1959, opened 2019-03-02
Use Response Files in CMake

jmittert:
Passing all of the swift source files to the swift compiler causes the
command line to go over Window's 32767 character limit command length.

This now writes the source file lists to response files to get around that.

I'm a CMake newb (it took me far too long to figure out how to join a list into a string), so it's fairly likely there's a better way of doing this.

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -1,6 +1,37 @@
 
 include(CMakeParseArguments)
 
+# Creates an output file map give a target and its list of sources at
+# output_path
+#
+# Usage:
+#   create_output_file_map(target sources output_path)
+function(create_output_file_map target sources output_path)
+  set(output_list)
+  set(output_file_map "{\n")
+  foreach(source ${sources})
+    get_filename_component(name ${source} NAME)
+
+    set(obj ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/${name}${CMAKE_C_OUTPUT_EXTENSION})
+    set(deps ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/${name}.d)
+    set(swiftdeps ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/${name}.swiftdeps)
+    set(dia ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/${name}.dia)
+    set(output_entry "\"${source}\": {\n\
+    \"object\": \"${obj}\",\n\
+    \"dependencies\": \"${deps}\",\n\
+    \"swift-dependencies\": \"${swiftdeps}\",\n\
+    \"diagnostics\": \"${dia}\"\n\
+},\n")
+    string(APPEND output_file_map ${output_entry})
+  endforeach()
+  set(master_deps ${CMAKE_CURRENT_BINARY_DIR}/${target}.swiftdeps)
+  string(APPEND output_file_map "\"\": {\n\
+  \"swift-dependencies\": \"${master_deps}\"\n\
+  }\n")
+  string(APPEND output_file_map "}\n")
+  file(WRITE ${output_path} ${output_file_map})
+endfunction()
+
 function(add_swift_target target)
   set(options LIBRARY;SHARED;STATIC)
   set(single_value_options MODULE_NAME;MODULE_LINK_NAME;MODULE_PATH;MODULE_CACHE_PATH;OUTPUT;TARGET)
@@ -11,6 +42,7 @@ function(add_swift_target target)
   set(compile_flags ${CMAKE_SWIFT_FLAGS})
   set(link_flags)
 
+  list(APPEND compile_flags -incremental)
   if(AST_TARGET)
     list(APPEND compile_flags -target;${AST_TARGET})
     list(APPEND link_flags -target;${AST_TARGET})
@@ -25,6 +57,11 @@ function(add_swift_target target)
   endif()
   if(AST_MODULE_CACHE_PATH)
     list(APPEND compile_flags -module-cache-path;${AST_MODULE_CACHE_PATH})
+  endif()
+  if(AST_MODULE_PATH)
+      get_filename_component(module_location ${AST_MODULE_PATH} PATH)
+      file(MAKE_DIRECTORY "${module_location}")
+      list(APPEND compile_flags "-emit-module-path;${AST_MODULE_PATH}")
   endif()
   if(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
     list(APPEND compile_flags -g)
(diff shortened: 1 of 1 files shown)


cmake/modules/SwiftSupport.cmake
```diff
@@ -97,6 +97,10 @@ function(add_swift_target target)
 
     set(all_sources ${sources})
     list(INSERT all_sources ${i} -primary-file)
+    string(LENGTH all_sources source_length)
+    set(resp_file ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/${name}.resp)
```
compnerd: Traditionally, the extension is `.rsp`.

cmake/modules/SwiftSupport.cmake
```diff
@@ -106,11 +110,22 @@ function(add_swift_target target)
...
+    list(JOIN objs " " objs_resp_text)
+    file(WRITE ${objs_resp_file} ${objs_resp_text})
+
     list(APPEND mods ${mod})
+    set(mods_resp_file ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/modects.resp)
+    list(JOIN mods " " mods_resp_text)
+    file(WRITE ${mods_resp_file} ${mods_resp_text})
+
     list(APPEND docs ${doc})
+    set(docs_resp_file ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/docsects.resp)
```
compnerd: Similar

cmake/modules/SwiftSupport.cmake
```diff
@@ -106,11 +110,22 @@ function(add_swift_target target)
...
-                         ${CMAKE_SWIFT_COMPILER} -frontend ${compile_flags} -emit-module-path ${mod} -emit-module-doc-path ${doc} -o ${obj} -c ${all_sources})
+                         ${CMAKE_SWIFT_COMPILER} -frontend ${compile_flags} -emit-module-path ${mod} -emit-module-doc-path ${doc} -o ${obj} -c "@${resp_file}")
 
     list(APPEND objs ${obj})
+    set(objs_resp_file ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/objects.resp)
+    list(JOIN objs " " objs_resp_text)
+    file(WRITE ${objs_resp_file} ${objs_resp_text})
+
     list(APPEND mods ${mod})
+    set(mods_resp_file ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/modects.resp)
```
compnerd: Similar.  Also, did you mean modsects?  Whats the `ects` suffix supposed to stand for?

cmake/modules/SwiftSupport.cmake
```diff
@@ -106,11 +110,22 @@ function(add_swift_target target)
                          ${source}
                          ${AST_DEPENDS}
                        COMMAND
-                         ${CMAKE_SWIFT_COMPILER} -frontend ${compile_flags} -emit-module-path ${mod} -emit-module-doc-path ${doc} -o ${obj} -c ${all_sources})
+                         ${CMAKE_SWIFT_COMPILER} -frontend ${compile_flags} -emit-module-path ${mod} -emit-module-doc-path ${doc} -o ${obj} -c "@${resp_file}")
 
     list(APPEND objs ${obj})
+    set(objs_resp_file ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/objects.resp)
```
compnerd: Similar

cmake/modules/SwiftSupport.cmake
jmittert: eee, should just be mods, bad search and replace after copying it from the objects one.

jmittert:
I can make this change in dispatch and xctest as well if that's wanted. However, what would be the better way of doing it?

jmittert:
Alright, updated this to use the swiftc driver + output_file_maps to track dependencies. The initial Windows build will be slower until we get a multithreaded swiftc Windows driver in, but the once built it's quick to rebuild.

cmake/modules/SwiftSupport.cmake
```diff
@@ -1,6 +1,36 @@
...
+    set(dia ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/${name}.dia)
+    set(output_entry "\"${source}\": {\n\
+    \"object\": \"${obj}\",\n\
+    \"dependencies\": \"${deps}\",\n\
+    \"swift-dependencies\": \"${swiftdeps}\",\n\
+    \"diagnostics\": \"${dia}\"\n\
+}")
+    list(APPEND output_list ${output_entry})
+  endforeach()
+  set(master_deps ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/main-build-record.swiftdeps)
```
compnerd: I think that the name should be `${target}.swiftdeps`.

cmake/modules/SwiftSupport.cmake
```diff
@@ -84,53 +115,14 @@ function(add_swift_target target)
...
-
-    math(EXPR i "${i}+1")
-  endforeach()
-
-  if(AST_LIBRARY)
-    get_filename_component(module_directory ${AST_MODULE_PATH} DIRECTORY)
-
-    set(module ${AST_MODULE_PATH})
-    set(documentation ${module_directory}/${AST_MODULE_NAME}.swiftdoc)
+  set(output_map_path ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/output_file_map.json)
```
compnerd: I'd prefer `output-file-map.json`.

cmake/modules/SwiftSupport.cmake
```diff
@@ -84,53 +115,14 @@ function(add_swift_target target)
...
-                         ${documentation}
-                       DEPENDS
-                         ${mods}
-                         ${docs}
-                         ${AST_DEPENDS}
-                       COMMAND
-                         ${CMAKE_SWIFT_COMPILER} -frontend ${compile_flags} -sil-merge-partial-modules -emit-module ${mods} -o ${module} -emit-module-doc-path ${documentation})
-  endif()
+  string(LENGTH sources source_length)
+  set(rsp_file ${CMAKE_CURRENT_BINARY_DIR}/${target}.dir/${target}.rsp)
```
compnerd: I think that we should be able to drop the `rsp` file here and let CMake/ninja DTRT.

jmittert: DTRT?

jmittert: Oh, do the right thing. CMake/Ninja don't do the right thing though. They write a batch file, but that's still over the 32k character limit for CreateProcessW.

compnerd: Ugh, right, because of the `add_custom_command` rather than the `add_library` *sigh*

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@jmittert:

20:02:09 CMake Error at cmake/modules/SwiftSupport.cmake:29 (list):
20:02:09   list does not recognize sub-command JOIN

jmittert:
Oh.. list(JOIN...) is cmake 3.12 I'll find an alternative

jmittert:
Should be good now. Switched it to not use an intermediate list.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

millenomi:
@compnerd is this good to go?

compnerd:
Yeap, this is good to go!  Thanks for the ping (wow, so that's what its like on the other side :-p)

PR #1960, opened 2019-03-02
NSAttributedString: Rename NSAttributedStringKey to NSAttributedString.Key

spevans:
- Also add an `@available()` deprecate the old name.

- Matches Darwin for Swift 5.0


spevans:
@swift-ci test

PR #1965, opened 2019-03-04
Decimal: NSDecimalNormalize() - avoid overflow/underflow on exponent difference

spevans:
- The difference of the expoents could exceed Int8.min / Int8.max, so
  convert to Int first.

- This avoids a crash if comparing numbers with a large exponent
  difference.


spevans:
@swift-ci test

PR #1966, opened 2019-03-04
Implement NSURLDirectoryEnumerator for Windows

jmittert:
Uses `FindFirstFileW` repeatedly to enumerate the directories. The unix implementation via fts does a preorder traversal, though the order that the children are visited in seems random. The Windows implementation also shares this behaviour: `FindNextFile` isn't guaranteed to be alphabetical depending on the file system.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -15,6 +15,21 @@ internal func &(left: UInt32, right: mode_t) -> mode_t {
 
 import CoreFoundation
 
+#if os(Windows)
+internal func joinPath(prefix: String, suffix: String) -> String {
+    var pszPath: PWSTR?
+    _ = prefix.withCString(encodedAs: UTF16.self) { prefix in
+        _ = suffix.withCString(encodedAs: UTF16.self) { suffix in
+            PathAllocCombine(prefix, suffix, ULONG(PATHCCH_ALLOW_LONG_PATHS.rawValue), &pszPath)
+        }
+    }
+
+    let path: String = String(decodingCString: pszPath!, as: UTF16.self)
+    LocalFree(pszPath)
+    return path
+}
+#endif
+
 open class FileManager : NSObject {
     
     /* Returns the default singleton instance.
@@ -800,21 +815,6 @@ open class FileManager : NSObject {
         })
         return contents
     }
-
-#if os(Windows)
-    private func joinPath(prefix: String, suffix: String) -> String {
-      var pszPath: PWSTR?
-      _ = prefix.withCString(encodedAs: UTF16.self) { prefix in
-        _ = suffix.withCString(encodedAs: UTF16.self) { suffix in
-          PathAllocCombine(prefix, suffix, ULONG(PATHCCH_ALLOW_LONG_PATHS.rawValue), &pszPath)
-        }
-      }
-
-      let path: String = String(decodingCString: pszPath!, as: UTF16.self)
-      LocalFree(pszPath)
-      return path
-    }
-#endif
 
     /**
     Performs a deep enumeration of the specified directory and returns the paths of all of the contained subdirectories.
@@ -2377,7 +2377,6 @@ extension FileManager {
             self.innerEnumerator = ie
         }
         
-        @available(Windows, deprecated, message: "Not Yet Implemented")
         override func nextObject() -> Any? {
             let o = innerEnumerator.nextObject()
             guard let url = o as? URL else {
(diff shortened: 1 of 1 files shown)


jmittert:
@compnerd @millenomi

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,113 @@ extension FileManager {
...
+                        if let handler = _errorHandler {
+                            keepGoing = handler(URL(fileURLWithPath: url.absoluteString),
+                                                _NSErrorWithWindowsError(GetLastError(), reading: true))
+                        } else {
+                            keepGoing = true
+                        }
+                        if !keepGoing {
+                            return nil
+                        }
+                        continue
```
spevans: Could this be collapsed down to 
```
if let handler = _errorHandler {
    let keepGoing = handler(URL(fileURLWithPath: url.absoluteString),
	          _NSErrorWithWindowsError(GetLastError(), reading: true))
    if !keepGoing { return nil }
}
continue
```

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,113 @@ extension FileManager {
...
+                var files: [URL] = []
+                repeat {
+                    let fileArr = [WCHAR](UnsafeBufferPointer(start: &ffd.cFileName.0,
+                                                              count: MemoryLayout.size(ofValue: ffd.cFileName)))
+                    let file = String(decodingCString: fileArr, as: UTF16.self)
+                    if (file != "."
+                        && file != ".."
+                        && (_options.contains(.skipsHiddenFiles)
+                            && (ffd.dwFileAttributes & DWORD(FILE_ATTRIBUTE_HIDDEN) == 0))) {
+                        files.append(URL(fileReferenceLiteralResourceName: "\(directory.absoluteString)\\\(file)"))
```
pvieito: Did you mean `URL(fileURLWithPath:)`?

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,113 @@ extension FileManager {
...
             return path
-#endif
         }
     }
 
-#if os(Windows)
     internal class NSURLDirectoryEnumerator : DirectoryEnumerator {
-        internal typealias ErrorHandler = /* @escaping */ (URL, Error) -> Bool
+#if os(Windows)
+        var _url : URL
```
spevans: Is `_url` needed? It doesnt look to be used outside of `init()` since it is also stored in `_stack`.

jmittert: Nope, removed.

Foundation/FileManager.swift
jmittert: Yes I did :s

millenomi:
@swift-ci please test

jmittert:
Added a missing FindClose on the FindFirstFile handle.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,107 @@ extension FileManager {
...
-        @available(Windows, deprecated, message: "Not Yet Implemented")
         override func nextObject() -> Any? {
             let o = innerEnumerator.nextObject()
             guard let url = o as? URL else {
                 return nil
             }
 
 #if os(Windows)
-            NSUnimplemented()
+            let path = url.path.replacingOccurrences(of: baseURL.path+"\\", with: "")
```
compnerd: I really would prefer that we use `PathRelativePathToW` here instead.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,107 @@ extension FileManager {
...
     }
 
-#if os(Windows)
     internal class NSURLDirectoryEnumerator : DirectoryEnumerator {
-        internal typealias ErrorHandler = /* @escaping */ (URL, Error) -> Bool
+#if os(Windows)
+        var _options : FileManager.DirectoryEnumerationOptions
+        var _errorHandler : ((URL, Error) -> Bool)?
+        var _stack: [URL]
+        var _current: URL?
```
compnerd: Can we name this `_url` instead?  That way we can make `_options`, `_errorHandler`, and `_url` common to all the implementations.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,107 @@ extension FileManager {
...
+        var _root_depth : Int
```
compnerd: `_rootDepth` to follow the style please (or `_depth` if you so like).

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,107 @@ extension FileManager {
...
 
-        init(url: URL, options: FileManager.DirectoryEnumerationOptions, errorHandler: ErrorHandler?) {
+        init(url: URL, options: FileManager.DirectoryEnumerationOptions, errorHandler: (/* @escaping */ (URL, Error) -> Bool)?) {
+            _options = options
+            _errorHandler = errorHandler
+            _stack = [url]
+            _root_depth = url.pathComponents.count
+        }
+
+
```
compnerd: I believe the style is 1 line spacing.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,107 @@ extension FileManager {
...
+        override func nextObject() -> Any? {
+            func contentsOfDir(directory: URL) -> [URL]? {
+                var ffd: WIN32_FIND_DATAW = WIN32_FIND_DATAW()
+                let dirPath = Array((directory.absoluteString + "\\*\0").utf16)
+                guard let h = dirPath.withContiguousStorageIfAvailable(
+                    {FindFirstFileW($0.baseAddress, &ffd)}),
```
compnerd: The lack of spaces around the closure make this hard to comprehend.  Remove the parenthesis:

```
let h: HANDLE = dirPath.withContiguousStorageIfAvailable {
  FindFirstFileW($0.baseAddress, &ffd)
}
guard h != INVALID_HANDLE_VALUE else { return nil }
```

Is easier to read IMO.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,107 @@ extension FileManager {
...
+                    h != INVALID_HANDLE_VALUE else {
+                    return nil
+                }
+                defer {
+                    FindClose(h)
+                }
+
+                var files: [URL] = []
+                repeat {
+                    let fileArr = [WCHAR](UnsafeBufferPointer(start: &ffd.cFileName.0,
```
compnerd: I think that spelling the array here as `Array<WCHAR>` is a bit more accessible

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,107 @@ extension FileManager {
...
+                                                              count: MemoryLayout.size(ofValue: ffd.cFileName)))
+                    let file = String(decodingCString: fileArr, as: UTF16.self)
+                    if file != "."
+                        && file != ".."
+                        && (_options.contains(.skipsHiddenFiles)
+                            && (ffd.dwFileAttributes & DWORD(FILE_ATTRIBUTE_HIDDEN) == 0)) {
```
compnerd: Am I messing up my boolean logic?  If the file is not `.` and not `..` AND does not contain `.skipsHiddenFiles` (that is, I want the hidden files), it skips the file?

Foundation/FileManager.swift
compnerd: I think that the unwrapped version that @spevans suggested is easier to read.

Foundation/FileManager.swift
jmittert: `_current` here more aligns with what `_current` is doing in the Unix implementation: storing the most recently visited file. While `url` in the Unix implementation is saving the url parameter. I think it would be a bit confusing to use them in different ways.

Foundation/FileManager.swift
jmittert: Yeah, it should be if (not `.` and not `..` and either of (it doesn't skip hidden, or the file is not hidden)

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,122 @@ extension FileManager {
...
             let o = innerEnumerator.nextObject()
             guard let url = o as? URL else {
                 return nil
             }
 
 #if os(Windows)
-            NSUnimplemented()
+            var relativePath: UnsafeMutableBufferPointer<WCHAR> = UnsafeMutableBufferPointer<WCHAR>.allocate(capacity: Int(MAX_PATH))
+            defer { relativePath.deallocate() }
+            guard TRUE == baseURL.absoluteString.withCString(encodedAs: UTF16.self, { pszFrom in
```
compnerd: Hmm, wouldn't we want this to be the FSR?  CC @millenomi

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,122 @@ extension FileManager {
...
+                    guard fromAttrs != INVALID_FILE_ATTRIBUTES,
+                        toAttrs != INVALID_FILE_ATTRIBUTES else { return FALSE }
+                    return PathRelativePathToW(
+                        relativePath.baseAddress,
+                        pszFrom,
+                        fromAttrs,
+                        pszTo,
+                        toAttrs)
+                })
+            }) else { return nil }
```
compnerd: Move the closing `)` to before the completion block.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,122 @@ extension FileManager {
...
+                    let fromAttrs = GetFileAttributesW(pszFrom)
+                    let toAttrs = GetFileAttributesW(pszTo)
+                    guard fromAttrs != INVALID_FILE_ATTRIBUTES,
+                        toAttrs != INVALID_FILE_ATTRIBUTES else { return FALSE }
+                    return PathRelativePathToW(
+                        relativePath.baseAddress,
+                        pszFrom,
+                        fromAttrs,
+                        pszTo,
+                        toAttrs)
```
compnerd: I believe that the arguments should be bin-packed.

Foundation/FileManager.swift
compnerd: Bleh, didn't see `_current` there an thought that it was the `_url` equivalent.

Foundation/FileManager.swift
jmittert: Alright, prettied up  the PathRelativeToPathW call some what and used FSR now.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,123 @@ extension FileManager {
...
+            _options = options
+            _errorHandler = errorHandler
+            _stack = [url]
+            _rootDepth = url.pathComponents.count
+        }
+
+        override func nextObject() -> Any? {
+            func contentsOfDir(directory: URL) -> [URL]? {
+                var ffd: WIN32_FIND_DATAW = WIN32_FIND_DATAW()
+                let dirPath = directory.absoluteString + "\\*"
```
compnerd: This needs a FSR I believe.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,123 @@ extension FileManager {
...
+                repeat {
+                    let fileArr = Array<WCHAR>(
+                      UnsafeBufferPointer(start: &ffd.cFileName.0,
+                                          count: MemoryLayout.size(ofValue: ffd.cFileName)))
+                    let file = String(decodingCString: fileArr, as: UTF16.self)
+                    if file != "."
+                        && file != ".."
+                        && (!_options.contains(.skipsHiddenFiles)
+                            || (ffd.dwFileAttributes & DWORD(FILE_ATTRIBUTE_HIDDEN) == 0)) {
+                        files.append(URL(fileURLWithPath: "\(directory.absoluteString)\\\(file)"))
```
compnerd: No, we have an internal helper `joinPath`, please use that rather than the `"\(directory.absoluteString)\\\(file)"`.  Also, FSR the `directory`.

jmittert:
Removed all the absolute strings and path concatting and used the path functions instead.

Foundation/FileManager.swift
```diff
@@ -802,7 +802,7 @@ open class FileManager : NSObject {
     }
 
 #if os(Windows)
-    private func joinPath(prefix: String, suffix: String) -> String {
+    fileprivate func joinPath(prefix: String, suffix: String) -> String {
```
compnerd: mmm, pretty sure that we want this to remain `private` as `fileprivate` is discouraged now I believe.  There is no need for this to be a member function, just extract it to a standalone function and let it remain `private` or make it `internal`.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,126 @@ extension FileManager {
...
+            _options = options
+            _errorHandler = errorHandler
+            _stack = [url]
+            _rootDepth = url.pathComponents.count
+        }
+
+        override func nextObject() -> Any? {
+            func contentsOfDir(directory: URL) -> [URL]? {
+                var ffd: WIN32_FIND_DATAW = WIN32_FIND_DATAW()
+                guard let dirFSR = directory.withUnsafeFileSystemRepresentation({ $0.flatMap { fsr in String(utf8String: fsr) } })
```
compnerd: The parenthesis around the block for the `withUnsfeFileSystemRepresentation` is unnecessary.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,126 @@ extension FileManager {
...
+                else { return nil }
+                let dirPath = FileManager().joinPath(prefix: dirFSR, suffix: "*")
```
compnerd: Please use `FileManager.default` rather than `FileManager()`.

Foundation/FileManager.swift
```diff
@@ -2377,46 +2377,126 @@ extension FileManager {
...
+                repeat {
+                    let fileArr = Array<WCHAR>(
+                      UnsafeBufferPointer(start: &ffd.cFileName.0,
+                                          count: MemoryLayout.size(ofValue: ffd.cFileName)))
+                    let file = String(decodingCString: fileArr, as: UTF16.self)
+                    if file != "."
+                        && file != ".."
+                        && (!_options.contains(.skipsHiddenFiles)
+                            || (ffd.dwFileAttributes & DWORD(FILE_ATTRIBUTE_HIDDEN) == 0)) {
+                        files.append(URL(fileURLWithPath: FileManager().joinPath(prefix: dirFSR, suffix: file)))
```
compnerd: `FileManager()` -> `FileManager.default`

Foundation/FileManager.swift
jmittert: This one isn't, the guard doesn't properly without it.

Foundation/FileManager.swift
jmittert: Sure, moved it to an internal func.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1969, opened 2019-03-05
FileHandle: create invalid handle properly on Windows

compnerd:
Use the FileHandle(handle:closeOnDealloc:) initializer rather than
passing the fd as `-1` which causes a failure in the C runtime due to an
invalid fd being passed to `_get_ofshandle`.


compnerd:
CC: @jakepetroules @millenomi

compnerd:
@swift-ci please test

PR #1970, opened 2019-03-05
Decimal: Enable bridging between Decimal / NSDecimalNumber for Codable fix.

spevans:
- Decoding Decimal was going via Double / NSNumber but this loses
  information for numbers with lots of digits that cant be represented
  by Double but can be by Decimal.

- TestCodable.swift: Re-enable some tests that were broken due to
  trying to JSON encode non-toplevel types.  Wrap them in an array
  to satisfy the requirement.

### DIFF
--- a/Foundation/Decimal.swift
+++ b/Foundation/Decimal.swift
@@ -397,6 +397,30 @@ extension Decimal : SignedNumeric {
     }
 }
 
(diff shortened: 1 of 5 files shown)


spevans:
@swift-ci test

PR #1971, opened 2019-03-05
[cmake] Skip regex when CMAKE_STATIC_LIBRARY_PREFIX is empty.

drodriguez:
In platforms like Windows, there's no prefix for static libraries.
Interpolating it in the regex will end up with a regex "^", which seems
not to be valid. Since the regex was removing the prefix, we can simply
skip the operation and the result will be the same.

Also do the same for CMAKE_STATIC_LIBRARY_SUFFIX, even if in that case,
Windows will have a value (but if it is empty, it will fail trying to
create the regex).


drodriguez:
Same fix in libdispatch: apple/swift-corelibs-libdispatch#452
Same fix in xctest: apple/swift-corelibs-xctest#257

millenomi:

PR #1972, opened 2019-03-06
[4.2] Fix a bug in format localized number on s390x and ARM

spevans:
(cherry picked from commit f9f75d25acd91741b9833c4697b8a47fcecad457)

This should fix the issue that broke #1929 on ARM


spevans:
@swift-ci test 4.2

spevans:
cc @weissi @futurejones

weissi:
@rjmccall @tkremenek can we take this? From my perspective it's fine but I'd like your input.

weissi:
> @rjmccall @tkremenek can we take this? From my perspective it's fine but I'd like your input.

sorry, please ignore me. I thought this was on the Swift repo :)

PR #1976, opened 2019-03-07
Implement removeItem on Windows

jmittert:
Use FindFirstFileW to "recursively" remove a directory tree in Windows.

It's probably simpler to implement this by creating a NSURLDirectoryEnumerator (https://github.com/apple/swift-corelibs-foundation/pull/1966) and DeleteFileW/RemoveDirectoryW'ing everything it returns. Would that be the preferred way to do it?

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -1441,14 +1441,68 @@ open class FileManager : NSObject {
         }
     }
 
-    @available(Windows, deprecated, message: "Not Yet Implemented")
     private func _removeItem(atPath path: String, isURL: Bool, alreadyConfirmed: Bool = false) throws {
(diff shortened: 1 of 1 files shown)


millenomi:
Either's fine.

millenomi:
@swift-ci please test

jmittert:
Updated to use `.withCString(encodedAs:)` since String.utf16 isn't null terminated.

compnerd:
I think that the DirectoryEnumerator is nicer since we can have the same codepath for all the targets that way.

jmittert:
> 
> 
> I think that the DirectoryEnumerator is nicer since we can have the same codepath for all the targets that way.

See https://github.com/apple/swift-corelibs-foundation/pull/2002

millenomi:
I'm closing this in favor of #2002.

PR #1979, opened 2019-03-07
RunLoop: remove incorrect usage of libdispatch queue

compnerd:
This has the wrong signature and is available directly.  Just use it
rather than create a macro.


compnerd:
@swift-ci please test

compnerd:
CC: @millenomi @phausler @parkera

PR #1980, opened 2019-03-08
Initial attempt to bring up the Foundation tests on Windows

compnerd:


TestFoundation/TestFileHandle.swift
```diff
@@ -8,6 +8,7 @@
 //
 
 import Dispatch
+@testable import Foundation
```
millenomi: This must be guarded with `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT`; see just above.

TestFoundation/TestFileHandle.swift
```diff
@@ -279,46 +280,47 @@ class TestFileHandle : XCTestCase {
     }
 
     func test_truncateFile() {
-        mkstemp(template: "test_truncateFile.XXXXXX") { (fh) in
-            fh.truncateFile(atOffset: 50)
-            XCTAssertEqual(fh.offsetInFile, 50)
+        let (fd, path) = try! _NSCreateTemporaryFile("test_truncate")
```
millenomi: This test depends on internal calls and thus must be guarded with `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT`.

TestFoundation/TestProcess.swift
```diff
@@ -7,6 +7,11 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+@testable import Foundation
```
millenomi: Same.

Since tests with `@testable import` only run for `--debug-foundation`, I’d strongly prefer that we instead find a way to bring in the functionality we need without invoking internal calls.

TestFoundation/TestTimeZone.swift
```diff
@@ -166,13 +166,15 @@ class TestTimeZone: XCTestCase {
     }
 
     func test_systemTimeZoneUsesSystemTime() {
+#if !os(Windows)
```
millenomi: Let’s also bracket this test’s entry in `allTests` with `!os(Windows)`, so that people aren’t misled into thinking this test works there.

TestFoundation/TestURL.swift
```diff
@@ -224,10 +231,19 @@ class TestURL : XCTestCase {
     
     static let gBaseTemporaryDirectoryPath = NSTemporaryDirectory()
     static var gBaseCurrentWorkingDirectoryPath : String {
+#if os(Windows)
```
millenomi: Unless this is testing that method, all of this can be replaced with `FileManager.default.currentDirectoryPath`.

compnerd:
CC: @millenomi @jakepetroules

TestFoundation/TestURL.swift
```diff
@@ -240,6 +256,42 @@ class TestURL : XCTestCase {
     static let gDirectoryDoesNotExistPath = gBaseTemporaryDirectoryPath + gDirectoryDoesNotExistName
 
     static func setup_test_paths() -> Bool {
+#if os(Windows)
```
millenomi: Outside of testing `FileManager`, this should really just use `FileManager` calls now that the class is basically entirely implemented.

TestFoundation/TestFileManager.swift
```diff
@@ -8,7 +8,7 @@
 //
 
 #if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
-    #if (os(Linux) || os(Android))
+    #if os(Linux) || os(Android) || os(Windows)
         @testable import Foundation
     #else
         @testable import SwiftFoundation
```
spevans: Would it be better to use:
```
#if canImport(SwiftFoundation)
#else
  @testable import Foundation
#endif
```

Rather than having an ever growing list of OSes or would this not work?

TestFoundation/TestFileHandle.swift
spevans: Alternatively you could use `NSTemporaryDirectory()` to get a tmp dir name then use something like:
```
let url = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent(ProcessInfo.processInfo.globallyUniqueString)
let fh = try FileHandle(forWritingTo: url)
```

Also, make `func test_truncateFile() throws` and let the throws propogate instead of using `try!`

compnerd:
I don't think that any of this is pertinent any longer.  We have moved far beyond this.

PR #1981, opened 2019-03-08
[4.2] Fix DateFormatter behaviour for nil locale and timeZone. (#1681)

spevans:
The locale and timeZone properties of DateFormatter seems to be "null
resetteable" in macOS, but the Swift Foundation behaviour differs.

Before this patch, setting locale to nil would segfault the program the
next time the formatter is used, while timeZone will incorrectly return
nil. After this patch a nil locale will simply reset to the current
locale and a nil timeZone will reset to the system one.

The tests check that the initial defaults are also correctly set.

(cherry picked from commit 1965003fd062915279353a1afc0a25fae1abbff3)


spevans:
@swift-ci test 4.2

PR #1982, opened 2019-03-08
[4.2] Fix DateFormatter TimeZone Setter

spevans:
Because of the use of 'timeZone' instead of 'newValue' in the setter, setting the timeZone of a DateFormatter doesn't actually work. Instead, it winds up re-setting the old value, which happens to be whatever was originally set, usually what we get from copying the default down in CoreFoundation.

Also adds a unit test to help prevent breaks like this again.

(cherry picked from commit 9f3e050bf26a1c8fe7e94849c8e0183f6ee10b39)

Fix Off-By-One Error With System TimeZone

This one is nasty, since it creates problems for DateFormatter, while appearing to behave correctly. (Mostly)

By copying 1 too many bytes, we wind up with the null terminator in the CFString. This breaks DateFormatter in an annoying way that the system TimeZone is always considered to be GMT by the formatter. You can ask for the zone’s abbreviation, and it will be correct. The name will appear to be correct, except the length is 1 longer than it should be. All sorts of fun. But the moment you ask DateFormatter to give you a string (or parse one), it assumes the TimeZone is GMT and ruins your day.

Included is a test case which can detect this. The downside is that it doesn’t detect the problem if GMT is already the current TimeZone. It is possible a second, more targeted test should also be added here?

This also breaks getting the description on the NSTimeZone/CFTimeZone, but not the TimeZone because of the wrapping at play, because of the terminator that gets inserted into the string description.

(cherry picked from commit 827e84274aae2b195b3278d7efec52e5f85511e0)

Cleanup SystemTimeZone tests

Cleans up the DateFormatter test, improves the information about how it doesn’t catch issues when the system time zone is already GMT, and adds a more focused test for the SystemTimeZone name issue that can catch it when GMT is the system time zone.

(cherry picked from commit 2219bd19850500af788c56852b46509068dbf584)

Make UTC Explicit in test_dateDifferenceComponents

The test assumes that it is running as GMT/UTC. Because of the TimeZone.current fixes, the test is no longer guaranteed to be using GMT/UTC, and will use the system TimeZone and breaks when in PDT for example. By making the UTC assumption explicit, we remove one more variable on this test.

(cherry picked from commit 5b1bfc963b935c916ea568ab913834d963f2fdfb)

Disable test_expectedTimeZone Case 1

This is hitting an issue on GMT time zones where it shows up as UTC when formatted.

(cherry picked from commit e82625234355da7abb5a035c24995d53907e7fb2)

### DIFF
--- a/CoreFoundation/NumberDate.subproj/CFTimeZone.c
+++ b/CoreFoundation/NumberDate.subproj/CFTimeZone.c
@@ -793,7 +793,7 @@ static CFTimeZoneRef __CFTimeZoneCreateSystem(void) {
         size_t zoneInfoDirLen = CFStringGetLength(__tzZoneInfo);
         if (strncmp(linkbuf, tzZoneInfo, zoneInfoDirLen) == 0) {
             name = CFStringCreateWithBytes(kCFAllocatorSystemDefault, (uint8_t *)linkbuf + zoneInfoDirLen,
-                                           strlen(linkbuf) - zoneInfoDirLen + 1, kCFStringEncodingUTF8, false);
+                                           strlen(linkbuf) - zoneInfoDirLen, kCFStringEncodingUTF8, false);
         } else {
             name = CFStringCreateWithBytes(kCFAllocatorSystemDefault, (uint8_t *)linkbuf, strlen(linkbuf), kCFStringEncodingUTF8, false);
         }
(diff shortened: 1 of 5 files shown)


spevans:
@swift-ci test 4.2

weissi:
@spevans this one failed (and does conflict now, possibly because of me merging OOO)

spevans:
@weissi I know the reason for failur,e it should be fixed by #1985. I will rebase.

spevans:
@swift-ci test 4.2

spevans:
@swift-ci test 4.2

spevans:
@weissi looks good now

PR #1983, opened 2019-03-08
[5.0] Link to the ICU61 libraries that are built for Linux

spevans:
- The findICU  was not finding the libicu{uc,i18n}swift libraries so
  libFoundation.so was still being linked to the system ICU libraries.

In the `master` branch, `ICU_UC_LIBRARY_DEBUG` `ICU_UC_LIBRARY_RELEASE`, `ICU_I18N_LIBRARY_DEBUG` and `ICU_I18N_LIBRARY_RELEASE` are passed in by `build-swift-impl`, directly set them in `CMakeLists.txt` here.


spevans:
@sw

spevans:
opened against wrong branch

PR #1985, opened 2019-03-10
[4.2] SR-8078: Some time zone identifiers are not supported by TimeZone on Linux

spevans:
- TimeZone.knownTimeZoneIdentifiers was returning time zone names
  that could not be instantiated by TimeZone(identifier:) due to
  CFTimeZone.c:__nameStringOK() filtering out some names.

- Make CFTimeZoneCopyKnownNames(), used by .knownTimeZoneIdentifiers,
  also filter these names using __nameStringOK().

- __nameStringOK calls ICU function ucal_getCanonicalTimeZoneID
  which ultimately decides if the time zone is valid.

(cherry picked from commit 7972ce4bd18b2b42bfa42bcbbbe76855ebf5cba5)


spevans:
@swift-ci test 4.2

spevans:
cc @weissi

PR #1986, opened 2019-03-12
TestFoundation: give preference to SwiftFoundation, SwiftXCTest

compnerd:
Alter the importing to use `canImport` and prefer the Swift prefixed
variants of the modules.  This avoids us from maintaining an ever
growing list of OSes, particularly when the Swift variants are less
widely available.

### DIFF
--- a/TestFoundation/TestFileManager.swift
+++ b/TestFoundation/TestFileManager.swift
@@ -8,10 +8,10 @@
 //
 
 #if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
-    #if (os(Linux) || os(Android))
-        @testable import Foundation
-    #else
+    #if canImport(SwiftFoundation) && !DEPLOYMENT_RUNTIME_OBJC
         @testable import SwiftFoundation
+    #else
+        @testable import Foundation
     #endif
 #endif
 
--- a/TestFoundation/xdgTestHelper/main.swift
+++ b/TestFoundation/xdgTestHelper/main.swift
@@ -7,10 +7,10 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-#if DEPLOYMENT_RUNTIME_OBJC || os(Linux) || os(Android)
-import Foundation
-#else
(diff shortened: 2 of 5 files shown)


compnerd:
CC: @millenomi @spevans

compnerd:
@swift-ci please test

TestFoundation/TestFileManager.swift
```diff
@@ -8,10 +8,10 @@
...
 #if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
-    #if (os(Linux) || os(Android))
-        @testable import Foundation
-    #else
+    #if canImport(SwiftFoundation)
         @testable import SwiftFoundation
+    #else
+        @testable import Foundation
     #endif
 #endif
```
spevans: Can all of this be moved into `TestImports` with an `#else` for the other bits or does `@testable` need to be in the file that is being tested?

compnerd: I don't know about that, this is mirroring what was there previously.  However, I did have to add it to a couple of files to get the builds to work, so I suspect that we cannot just move these.

TestFoundation/xdgTestHelper/main.swift
```diff
@@ -7,10 +7,10 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-#if DEPLOYMENT_RUNTIME_OBJC || os(Linux) || os(Android)
-import Foundation
-#else
+#if canImport(SwiftFoundation)
```
millenomi: Can all the `canImport…` be turned into `canImport(…) && !DEPLOYMENT_RUNTIME_OBJC`? This allows DarwinCompatibilityTests to keep working.

compnerd: Sure, if I can just point out that this is exactly the problem with `canImport` :-p

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1987, opened 2019-03-12
StringsEncoding: correct bitmap loading

compnerd:
This fixes the bitmap loading on Windows.  Windows uses the "Unicode"
(UTF-16) file paths as that is the native representation.  However, we
were using `strcmp` which would not yield the correct check.  This
allows us to load the bitmap data on Windows as well.


compnerd:
CC: @millenomi @phausler @parkera @jakepetroules

compnerd:
@swift-ci please test

PR #1989, opened 2019-03-12
TestNSArray: use FSR on the temporary file paths

compnerd:
The FSR conversion is important for Windows paths to be treated
correctly and the file operations to succeed.  Adjust the test utility
to do so.

### DIFF
--- a/TestFoundation/TestNSArray.swift
+++ b/TestFoundation/TestNSArray.swift
@@ -806,13 +806,18 @@ class TestNSArray : XCTestCase {
     }
 
     private func createTestFile(_ path: String, _contents: Data) -> String? {
-        let tempDir = NSTemporaryDirectory() + "TestFoundation_Playground_" + NSUUID().uuidString + "/"
+        let tempDir: URL = URL(fileURLWithPath: "TestFoundation_Playground_" + NSUUID().uuidString, isDirectory: true,
+                               relativeTo: URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true))
         do {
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @millenomi @gwynne @jakepetroules

compnerd:
@swift-ci please test

spevans:
Shouldn't the `FileManager` methods be doing the FSR? Is it missing from some of them?

compnerd:
Hmm, should it?  If so, there is a lot of changes that are needed in FileManager, since most of the interfaces do not seem to be done the FSR conversions.

spevans:
But isnt it just the system calls that need the FSR wrap? Up to that point it should just be a `String`. Im pretty sure I fixed `FileManager`s system calls a while ago unless there are any new ones that need fixing.

compnerd:
All of the createItem, removeItem, etc do not seem to be doing the FSR conversion.  I don't think that `String` is right though, unless we are going to be using `\` as the separator everywhere.

spevans:
In the case of `removeItem` the underlying `rmdir` and `unlink` calls need to do the wrapping, wether they are called via the `URL` or `String` interfaces. There may be a few that I missed before but I can send in a PR to fix those. Users of `FileManager` shouldnt / wouldnt know which calls need to be wrapped in any case.

compnerd:
Okay, lets get `FileManager` fixed up then.  I am probably going to start finding more of these cases as I'm trying to get the tests running on Windows, and the difference is actually material there.

spevans:
Yup, no problem, good to know the tests cover it all.

compnerd:
I *think* that this is no longer needed with the changes that @spevans did.  Gonna discard it for now.

PR #1990, opened 2019-03-12
NSPathUtilities: relax the constraint for Windows

compnerd:
The path size changing to a smaller value is acceptable.  Furthermore,
the return value will not account for the trailing null, so the count
was off by one.  This fixes the temporary path construction for Windows.


compnerd:
CC: @millenomi @gwynne

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1991, opened 2019-03-12
TestFoundation: tm_zone does not exist on Windows

compnerd:
There is no equivalent for tm_zone on Windows.  Simply ignore the test.


compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

PR #1994, opened 2019-03-12
Wrap more filesystem calls with fileSystemRepresentation()

spevans:


spevans:
@swift-ci test

spevans:
cc @compnerd

Foundation/FileManager.swift
```diff
@@ -1450,7 +1450,7 @@ open class FileManager : NSObject {
 #if os(Windows)
         NSUnimplemented()
 #else
-        if rmdir(path) == 0 {
+        if _fileSystemRepresentation(withPath: path, { rmdir($0) }) == 0 {
```
compnerd: I think that it would be better to wrap the path conversion at the `_removeItem` call site.

Foundation/FileHandle.swift
```diff
@@ -752,8 +752,10 @@ extension FileHandle {
     }
     
     internal static func _openFileDescriptorForURL(_ url : URL, flags: Int32, reading: Bool) throws -> Int32 {
```
compnerd: Again, I think its better to wrap this at the public API side where the call to `_openFileDescriptorURL` is made.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

Foundation/FileHandle.swift
spevans: The advantage of doing it here is you dont need to fix up every caller or any new callers.

Also the type for the file system representaion is not a `URL` or `String` so it would change the method signature.

spevans:
@swift-ci test and merge

PR #1995, opened 2019-03-13
Fix off-by-one when initializing Data with discontiguous, underestimated Sequences

itaiferber:
Applies apple/swift#23244 to here as well.


itaiferber:
@swift-ci Please test

PR #1997, opened 2019-03-13
CFFileUtilties: correct directory iteration for Windows

compnerd:
We did not pass the file handler the absolute path to the file.  This
would break as it expected a non-NULL absolute path.  This is needed for
the Bundle tests to work.


compnerd:
CC: @millenomi @jakepetroules @gwynne

compnerd:
@swift-ci please test

PR #2000, opened 2019-03-13
[SR-10374] URLCache: init method and first time sqlite database setup implemented

karthikkeyan:
* `init` method implemented
* `URLCache.shared` singleton object created with 4MB of memory space and 20MB of disk space
* Directory and database file Cache.db file created under the local directory
* Sqlite Tables and Indices created in the database
* Unit tests added for `URLCache` to verify directory, file, tables, and indices

Bug Link: https://bugs.swift.org/browse/SR-10374

### DIFF
--- a/Foundation/URLCache.swift
+++ b/Foundation/URLCache.swift
@@ -201,6 +201,7 @@ open class URLCache : NSObject {
         self.memoryCapacity = memoryCapacity
         self.diskCapacity = diskCapacity
 
+        // As per the function of URLCache, `diskCapacity` of `0` is assumed as no-persistence.
         if diskCapacity > 0, let _path = path {
             self.persistence = _CachePersistence(path: _path)
         }
@@ -322,17 +323,13 @@ fileprivate struct _CachePersistence {
     }
 
     func setupCacheDirectory() {
-        do {
-            try FileManager.default.createDirectory(atPath: path, withIntermediateDirectories: true)
-        } catch {
-            fatalError("Unable to create directories for URLCache: \(error.localizedDescription)")
-        }
+        try? FileManager.default.createDirectory(atPath: path, withIntermediateDirectories: true)
     }
 
     func saveCachedResponse(_ response: CachedURLResponse, for request: URLRequest) {
-        let archivedData = NSKeyedArchiver.archivedData(withRootObject: response)
         do {
-            if let fileIdentifier = request.cacheFileIdentifier {
+            if let archivedData = try? NSKeyedArchiver.archivedData(withRootObject: response, requiringSecureCoding: true),
+                let fileIdentifier = request.cacheFileIdentifier {
                 let cacheFileURL = urlForFileIdentifier(fileIdentifier)
                 try archivedData.write(to: cacheFileURL)
             }
@@ -348,7 +345,7 @@ fileprivate struct _CachePersistence {
 
         let url = urlForFileIdentifier(fileIdentifier)
         guard let data = try? Data(contentsOf: url),
-            let response = NSKeyedUnarchiver.unarchiveObject(with: data) as? CachedURLResponse else {
+            let response = try? NSKeyedUnarchiver.unarchivedObject(ofClasses:[CachedURLResponse.self], from: data) as? CachedURLResponse else {
             return nil
         }
 
@@ -367,23 +364,28 @@ extension URLRequest {
         guard let scheme = self.url?.scheme, scheme == "http" || scheme == "https",
             let method = httpMethod, !method.isEmpty,
             let urlString = url?.absoluteString else {
-            return nil
+                return nil
         }
 
-        var hashString = "\(scheme)_\(method)_\(urlString)"
+        var hashString = "\(abs(method.hashValue))-\(abs(urlString.hashValue))"
+
         if let userAgent = self.allHTTPHeaderFields?["User-Agent"], !userAgent.isEmpty {
-            hashString.append(contentsOf: "_\(userAgent)")
+            hashString.append("\(abs(userAgent.hashValue))")
         }
 
         if let acceptLanguage = self.allHTTPHeaderFields?["Accept-Language"], !acceptLanguage.isEmpty {
-            hashString.append(contentsOf: "_\(acceptLanguage)")
+            hashString.append("-\(abs(acceptLanguage.hashValue))")
         }
 
         if let range = self.allHTTPHeaderFields?["Range"], !range.isEmpty {
-            hashString.append(contentsOf: "_\(range)")
+            hashString.append("-\(abs(range.hashValue))")
         }
 
-        return String(format: "%X", hashString.hashValue)
+        if let data = self.httpBody, !data.isEmpty {
+            hashString.append("-\(abs(data.hashValue))")
+        }
+
+        return hashString
     }
 
 }
--- a/Foundation/URLCache.swift
+++ b/Foundation/URLCache.swift
@@ -12,6 +12,7 @@ import SwiftFoundation
 #else
 import Foundation
 #endif
+import SQLite3
 
 /*!
     @enum URLCache.StoragePolicy
@@ -128,6 +129,26 @@ open class CachedURLResponse : NSObject, NSSecureCoding, NSCopying {
 
 open class URLCache : NSObject {
     
+    private static let sharedSyncQ = DispatchQueue(label: "org.swift.URLCache.sharedSyncQ")
+    
+    private static var sharedCache: URLCache? {
+        willSet {
+            URLCache.sharedCache?.syncQ.sync {
+                URLCache.sharedCache?._databaseClient?.close()
+                URLCache.sharedCache?.flushDatabase()
+            }
+        }
+        didSet {
+            URLCache.sharedCache?.syncQ.sync {
+                URLCache.sharedCache?.setupCacheDatabaseIfNotExist()
+            }
+        }
+    }
+    
+    private let syncQ = DispatchQueue(label: "org.swift.URLCache.syncQ")
+    private let _baseDiskPath: String?
+    private var _databaseClient: _CacheSQLiteClient?
+    
     /*! 
         @method sharedURLCache
         @abstract Returns the shared URLCache instance.
@@ -147,10 +168,22 @@ open class URLCache : NSObject {
     */
     open class var shared: URLCache {
         get {
-            NSUnimplemented()
+            return sharedSyncQ.sync {
+                if let cache = sharedCache {
+                    return cache
+                } else {
+                    let fourMegaByte = 4 * 1024 * 1024
+                    let twentyMegaByte = 20 * 1024 * 1024
+                    let cacheDirectoryPath = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first?.path ?? "\(NSHomeDirectory())/Library/Caches/"
+                    let path = "\(cacheDirectoryPath)\(Bundle.main.bundleIdentifier ?? UUID().uuidString)"
+                    let cache = URLCache(memoryCapacity: fourMegaByte, diskCapacity: twentyMegaByte, diskPath: path)
+                    sharedCache = cache
+                    return cache
+                }
+            }
         }
         set {
-            NSUnimplemented()
+            sharedSyncQ.sync { sharedCache = newValue }
         }
     }
 
@@ -167,7 +200,13 @@ open class URLCache : NSObject {
         @result an initialized URLCache, with the given capacity, backed
         by disk.
     */
-    public init(memoryCapacity: Int, diskCapacity: Int, diskPath path: String?) { NSUnimplemented() }
+    public init(memoryCapacity: Int, diskCapacity: Int, diskPath path: String?) {
+        self.memoryCapacity = memoryCapacity
+        self.diskCapacity = diskCapacity
+        self._baseDiskPath = path
+        
+        super.init()
+    }
     
     /*! 
         @method cachedResponseForRequest:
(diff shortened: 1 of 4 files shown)


pushkarnk:
@karthikkeyan Apologies for the very late response on this PR. I think implementations of URLCache and URLCredentialStorage have been long pending and this looks like a great start! However, I've a few questions about this pull request:
1. Is SQLite available on Linux and Windows? I guess not.
2. Have we considered other storage alternatives? 
I came across this [old discussion on URLCache ](https://forums.swift.org/t/implementing-urlcache/13398). You might find it relevant.

karthikkeyan:
> @karthikkeyan Apologies for the very late response on this PR. I think implementations of URLCache and URLCredentialStorage have been long pending and this looks like a great start! However, I've a few questions about this pull request:
> 
> 1. Is SQLite available on Linux and Windows? I guess not.
> 2. Have we considered other storage alternatives?
>    I came across this [old discussion on URLCache ](https://forums.swift.org/t/implementing-urlcache/13398). You might find it relevant.

I have failed to give a fair bit of a thought process about the Linux platform. Thanks for reminding of it, I will follow the discussion and see how best we can move this PR forward.

karthikkeyan:
@pushkarnk I have reimplemented store and retrieve logic using `NSKeyedArchiver` and `NSKeyedUnarchiver`. This implementation is not perfect but I think it would be a good start.

spevans:
@swift-ci test

Foundation/URLCache.swift
```diff
@@ -142,13 +147,37 @@ open class URLCache : NSObject {
...
+                }
+            }
         }
         set {
-            NSUnimplemented()
+            guard newValue !== sharedCache else { return }
+            
+            sharedSyncQ.sync {
+                // Remove previous data before resetting new URLCache instance
+                URLCache.sharedCache?.persistence?.deleteAllCaches()
```
drodriguez: Does the original implementation do this? Feels weird removing “user” data (even if it is a cache), just by assigning another value.

Foundation/URLCache.swift
```diff
@@ -243,11 +283,91 @@ open class URLCache : NSObject {
...
     public func storeCachedResponse(_ cachedResponse: CachedURLResponse, for dataTask: URLSessionDataTask) { NSUnimplemented() }
     public func getCachedResponse(for dataTask: URLSessionDataTask, completionHandler: (CachedURLResponse?) -> Void) { NSUnimplemented() }
     public func removeCachedResponse(for dataTask: URLSessionDataTask) { NSUnimplemented() }
 }
+
+fileprivate struct _CachePersistence {
+
+    let path: String
+
+    var cacheSizeInDesk: Int {
```
drodriguez: Typo: `cacheSizeInDisk`

Foundation/URLCache.swift
```diff
@@ -243,11 +283,91 @@ open class URLCache : NSObject {
...
+        do {
+            let subFiles = try FileManager.default.subpathsOfDirectory(atPath: path)
+            var total: Int = 0
+            for fileName in subFiles {
+                let attributes = try FileManager.default.attributesOfItem(atPath: path.appending(fileName))
+                total += (attributes[.size] as? Int) ?? 0
+            }
+            return total
```
drodriguez: Is there plan to keep this value updated as cached responses are added and evicted? Feels like with a big cache and a slow disk, each of these calls might take a while.

Foundation/URLCache.swift
```diff
@@ -243,11 +283,91 @@ open class URLCache : NSObject {
...
+        } catch {
+            return 0
+        }
+    }
+
+    func setupCacheDirectory() {
+        do {
+            if FileManager.default.fileExists(atPath: path) {
+                try FileManager.default.removeItem(atPath: path)
```
drodriguez: Wouldn’t this delete the previous cache contents?

Foundation/URLCache.swift
```diff
@@ -243,11 +283,91 @@ open class URLCache : NSObject {
...
+    func deleteAllCaches() {
+        do {
+            try FileManager.default.removeItem(atPath: path)
+        } catch {
+            fatalError("Unable to flush database for URLCache: \(error.localizedDescription)")
+        }
+    }
+
+    private func urlForFileIdentifier(_ identifier: String) -> URL {
+        return URL(fileURLWithPath: path + "/" + identifier)
```
drodriguez: It would be good to use `URL(fileURLWithPath: path, isDirectory: true).appendingPathComponent(identifier)` to handle platforms with other directory separators.

Foundation/URLCache.swift
```diff
@@ -243,11 +283,91 @@ open class URLCache : NSObject {
...
+
+fileprivate extension URLRequest {
+
+    var cacheFileIdentifier: String {
+        guard let urlString = url?.absoluteString else {
+            fatalError("Unable to create cache identifier for request: \(self)")
+        }
+
+        let method = httpMethod ?? "GET"
+        return urlString + method
```
drodriguez: Since this is used as part of the path, shoulnd’t the output be escaped or hashed somehow?

karthikkeyan:
> Full disclosure: I didn’t know about this PR, and I started working on my own `URLCache` (that's the reason for #2245). I have some parts that are present in this version (with some differences), but I focused in the memory cache first. I have ideas for the disk cache, but that's more complicated. I can also have a lot more feedback about things I know that will need to be handle (for example skipping some protocols from being cached). Are you still planning in working on this?

@drodriguez  Sorry for the delay response, yes I will be working on this PR. I was just waiting for someone to review this. Please give me your suggestions.

Foundation/URLCache.swift
```diff
@@ -142,13 +147,37 @@ open class URLCache : NSObject {
...
     open class var shared: URLCache {
         get {
-            NSUnimplemented()
+            return sharedSyncQ.sync {
+                if let cache = sharedCache {
+                    return cache
+                } else {
+                    let fourMegaByte = 4 * 1024 * 1024
+                    let twentyMegaByte = 20 * 1024 * 1024
+                    let bundleIdentifier = Bundle.main.bundleIdentifier ?? UUID().uuidString
```
drodriguez: That `UUID` at the end is a little bit worrying. The app will fill caches, and will never reuse the cached results, filling the disk of garbage. I can think at least of `ProcessInfo.processInfo.processName` as a good fallback, but it is not unique (but the bundle identifier doesn’t have to be either). Maybe if no bundle identifier is found the best thing is not to have a disk cache at all.

Foundation/URLCache.swift
```diff
@@ -142,13 +147,37 @@ open class URLCache : NSObject {
...
+                    let cacheDirectoryPath = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first?.path ?? "\(NSHomeDirectory())/Library/Caches"
```
drodriguez: Same idea here. If we don’t find a good URL for `.cachesDirectory` maybe it is better to not have disk cache.

The fallback has the problem of being very Darwin specific. For example `NSHomeDirectory()` in Android is the root directory, where one cannot write at all. Also, it might want to use `appendingPathComponent(…).appendingPathComponent(…)` to be nicer to Windows.

(There's another case below where the POSIX path separator is used instead of a higher level function).

Foundation/URLCache.swift
```diff
@@ -162,7 +191,16 @@ open class URLCache : NSObject {
         @result an initialized URLCache, with the given capacity, backed
         by disk.
     */
-    public init(memoryCapacity: Int, diskCapacity: Int, diskPath path: String?) { NSUnimplemented() }
+    public init(memoryCapacity: Int, diskCapacity: Int, diskPath path: String?) {
+        self.memoryCapacity = memoryCapacity
+        self.diskCapacity = diskCapacity
+
+        if let _path = path {
```
drodriguez: You might also want to check that `diskCapacity` is more than zero.

Foundation/URLCache.swift
```diff
@@ -243,11 +283,91 @@ open class URLCache : NSObject {
...
+
+    private func urlForFileIdentifier(_ identifier: String) -> URL {
+        return URL(fileURLWithPath: path + "/" + identifier)
+    }
+
+}
+
+fileprivate extension URLRequest {
+
+    var cacheFileIdentifier: String {
```
drodriguez: I have a slightly different design here (I use a `URLCacheKey` type, instead of `String`), but one important thing is that generating a key from the `URLRequest` depends on some details from the request beside the URL itself. So my equivalent method returns an optional.

The first thing is the protocol. Protocols like `data:` or `file:` should not be cached for obvious reasons. I also think unknown protocols should not be cached either.

Also, for known protocols like `http:` or `https:` (and maybe for `ftp:`, but I don’t know that protocol that well), we should do a couple more things. For example:
- I don’t think you can do a HTTP request without a method, so that's a `return nil` in the case the protocol is HTTP.
- I think some headers should be taken into account. The spec refers to `User-Agent`, but others like `Accept-Language` and `Range` might be interesting.
- I think the hash of the body contents should also be used.
- I also think one should not try to cache request with body streams, because that will consume it, which is not good.
- Finally I would hash everything and return an hexadecimal string, to have a valid file name.

Foundation/URLCache.swift
millenomi: The home directory should be in a persistent writable path everywhere, much as possible. I know that Android is _really_ weird about this, but that looks like incorrect mapping to me.

Foundation/URLCache.swift
millenomi: Can this point to the temporary directory with a fixed name? (A sanitized process name does sound good to me!)

Foundation/URLCache.swift
millenomi: Yeah, please do not fall back like this. If `FileManager` does not return a caches URL, consider not caching at all.

Foundation/URLCache.swift
millenomi: It does not.

Foundation/URLCache.swift
millenomi: Darwin indeed does accept 0 arguments. My assumption is that memory capacity 0 means that as much as possible no data is kept in memory, and disk capacity 0 that no data is written to disk.

Foundation/URLCache.swift
millenomi: Indeed.

Foundation/URLCache.swift
millenomi: Indeed.

millenomi:
I want to shepherd this patch into the repository. Do you need help with changes here?

Foundation/URLCache.swift
karthikkeyan: Let me add a `FIXME` comment, I will create a separate PR for optimizing this. My main focus in this PR is to create an initial setup and kick start the implementation. Hope that's ok.

Foundation/URLCache.swift
karthikkeyan: Sure, I will remove the fallback if `FileManager` for `cachesDirectory` returned nill.

karthikkeyan:
> I want to shepherd this patch into the repository. Do you need help with changes here?

@millenomi I am addressing all the comments added by you and Danial. I will let you know if I need any help. Thanks Lily.

karthikkeyan:
@millenomi @drodriguez I have addressed all the comments. I still have some doubts about the implementation of the property `cacheFileIdentifier`, please let me know if there is any changes needed.

millenomi:
@swift-ci please test

Foundation/URLCache.swift
```diff
@@ -248,11 +289,101 @@ open class URLCache : NSObject {
...
+extension URLRequest {
+
+    fileprivate var cacheFileIdentifier: String? {
+        guard let scheme = self.url?.scheme, scheme == "http" || scheme == "https",
+            let method = httpMethod, !method.isEmpty,
+            let urlString = url?.absoluteString else {
+            return nil
+        }
+
+        var hashString = "\(scheme)_\(method)_\(urlString)"
```
millenomi: The collision space of the default hash is smaller than you probably think. Can we at least check when we load a request that this is actually the request we were looking for?

Foundation/URLCache.swift
```diff
@@ -248,11 +289,101 @@ open class URLCache : NSObject {
...
+    }
+
+    func cachedResponse(for request: URLRequest) -> CachedURLResponse? {
+        guard let fileIdentifier = request.cacheFileIdentifier else {
+            return nil
+        }
+
+        let url = urlForFileIdentifier(fileIdentifier)
+        guard let data = try? Data(contentsOf: url),
+            let response = NSKeyedUnarchiver.unarchiveObject(with: data) as? CachedURLResponse else {
```
millenomi: Please use secure coding here.

Foundation/URLCache.swift
```diff
@@ -248,11 +289,101 @@ open class URLCache : NSObject {
...
+    func setupCacheDirectory() {
+        do {
+            try FileManager.default.createDirectory(atPath: path, withIntermediateDirectories: true)
+        } catch {
+            fatalError("Unable to create directories for URLCache: \(error.localizedDescription)")
+        }
+    }
+
+    func saveCachedResponse(_ response: CachedURLResponse, for request: URLRequest) {
+        let archivedData = NSKeyedArchiver.archivedData(withRootObject: response)
```
millenomi: This is slightly less urgent, but I'd like you to use secure coding wherever you use coding.

Foundation/URLCache.swift
```diff
@@ -248,11 +289,101 @@ open class URLCache : NSObject {
...
+        } catch {
+            return 0
+        }
+    }
+
+    func setupCacheDirectory() {
+        do {
+            try FileManager.default.createDirectory(atPath: path, withIntermediateDirectories: true)
+        } catch {
+            fatalError("Unable to create directories for URLCache: \(error.localizedDescription)")
```
millenomi: Do not introduce fatal errors for things the user may not be able to control. As I asked elsewhere: consider not loading if we do not have a suitable directory.

Foundation/URLCache.swift
millenomi: To use secure coding, create a `NSKeyedUnarchiver`, set its `requiresSecureCoding` to `true`, and use its method to encode an object for key `NSKeyedArchiveRootObjectKey`.

millenomi:
I'm going to commandeer this patch; I may drop the SQLite dependency, though.

karthikkeyan:
> I'm going to commandeer this patch; I may drop the SQLite dependency, though.

@millenomi  I have pushed one more commit addressing the comments related to secure archiving. Also, I have updated the hashing mechanism for generating `cacheFileIdentifier`. 

The one thing I am struggling though is to write a unit test for the internal property `cacheFileIdentifier`

karthikkeyan:
Closing this PR as `URLCache` is implemented in this PR https://github.com/apple/swift-corelibs-foundation/pull/2401

PR #2002, opened 2019-03-14
Implement _removeItem in terms of DirectoryEnumerator

jmittert:
Alternate implementation to https://github.com/apple/swift-corelibs-foundation/pull/1976 that uses DirectoryEnumerator

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -12,6 +12,10 @@ internal func &(left: UInt32, right: mode_t) -> mode_t {
     return mode_t(left) & right
 }
 #endif
+#if os(Windows)
+fileprivate let rmdir = _NS_rmdir
+fileprivate let unlink = _NS_unlink
+#endif
 
 import CoreFoundation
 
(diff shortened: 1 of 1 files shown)


millenomi:
@swift-ci please test

Foundation/FileManager.swift
```diff
@@ -1441,73 +1441,63 @@ open class FileManager : NSObject {
...
                 }
-                return fts_open(ps, FTS_PHYSICAL | FTS_XDEV | FTS_NOCHDIR, nil)
-            }
-            
-            if stream != nil {
-                defer {
-                    fts_close(stream)
+                if (URL(fileURLWithPath: itemPath).hasDirectoryPath) {
+                  #if os(Windows)
+                  let result = _NS_rmdir(itemPath)
```
millenomi: Do we want to just fileprivate let _NS_rmdir = rmdir on non-Windows?

millenomi: And similarly for `_NS_unlink`.

compnerd: Or `fileprivate let rmdir = _NS_rundir` on Windows?

jmittert: Sure!

Foundation/FileManager.swift
```diff
@@ -1441,73 +1445,53 @@ open class FileManager : NSObject {
...
-        } else if errno != ENOTDIR {
-            throw _NSErrorWithErrno(errno, reading: false, path: path)
-        } else if _fileSystemRepresentation(withPath: path, { unlink($0) != 0 }) {
-            throw _NSErrorWithErrno(errno, reading: false, path: path)
+        } else if _fileSystemRepresentation(withPath: path, {path in
+                                            #if os(Windows)
+                                            return _NS_unlink(path) != 0
+                                            #else
+                                            return unlink(path) != 0
+                                            #endif
```
compnerd: This can also be `unlink(path)` now right?

jmittert: eeeyup

spevans:
Wont this spend a lot of effort to convert every file entry from file system representation into a URL then convert it back again to remove it? Im not sure of the benefit of this change.

jmittert:
> 
> 
> Wont this spend a lot of effort to convert every file entry from file system representation into a URL then convert it back again to remove it? Im not sure of the benefit of this change.

It does add extra URL calls, however, it's one less use of fts and means the same codepath works across Windows/Darwin/Linux. How expensive is converting to a URL? I'm still okay with https://github.com/apple/swift-corelibs-foundation/pull/1976 if adding an extra conversion is too expensive to be worth it.

compnerd:
Could also do the `.standardizingPath` to avoid the `URL` conversions.

millenomi:
@swift-ci please test

jmittert:
> 
> 
> Could also do the `.standardizingPath` to avoid the `URL` conversions.

I can remove all the URL conversions in _removeItem, but the issue is still that NSURLDirectoryEnumerator gives back URLs rather than Strings so it will still be doing the conversion there and back.

compnerd:
Yes, but that is a **feature** since the paths need to be converted to FSRs prior to the `unlink`.

millenomi:
@swift-ci please test

jmittert:
Okay, I've removed the unneeded URL conversions in _removeItem, the only one remaining is the needed one to create the Enumerator.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

jmittert:
It fails when creating a symlink for swiftpm... I don't think it's related?
```
12:45:08 [37/38] Compiling Swift Module 'SwiftPMPackageTests' (1 sources)
12:45:08 error: symlink error: File exists (17): /home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/buildbot_linux/swiftpm-linux-x86_64/release -> x86_64-unknown-linux/release
12:45:08 --- bootstrap: error: build failed with exit status 1
12:45:08 Building the standard library for: swift-stdlib-linux-x86_64
12:45:08 Running Swift tests for: check-swift-all-linux-x86_64 check-swift-all-optimize-linux-x86_64
12:45:08 /home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift/utils/build-script: fatal error: command terminated with a non-zero exit status 1, aborting
```

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

jmittert:
> @swift-ci please test
Ah hold up a bit, this is actually broken on Linux, I've repro'd it locally. Just having issues setting up a Linux environment where I can test it properly.

spevans:
You are calling the `unlink` and `rmdir` without using the FSR of the file name. 
Also every `_fileExists` call does at least one `lstat` call and possibly a second `stat` call making this quite system call heavy.

Another point regarding the FSR, is that the `fts_read` already returns the filename in FSR which can be passed straight to `unlink` and `rmdir` instead of converting the name to a `String` and then back again.

I think this approach in general is wrong, for an internal helper like this it should just use the best per-OS solution to delete a file or directory structure.

jmittert:
>
> I think this approach in general is wrong, for an internal helper like this it should just use the best per-OS solution to delete a file or directory structure.

I actually agree, and I think I'd rather do #1976 instead. @compnerd if we really care about a single codepath, one other alternative might be to have an fts shim and implement that on Windows and call through to the real thing on everything else, but that might be more effort than it's worth.

compnerd:
I don't think that we want to re-implement FTS on top of the Windows APIs.  Would it not be possible to actually have an internal method to avoid the overheads of the FTS translation and the stating?  If not, then I suppose that we can split the file into two and have a `FileManager+POSIX.swift` and `FileManager+Win32.swift` and start duplicating the interfaces.

jmittert:
Closing in favour of https://github.com/apple/swift-corelibs-foundation/pull/2117

PR #2004, opened 2019-03-14
build: use `add_subdirectory` instead of `add_external_project`

compnerd:
This switches the CoreFoundation sub-build to use `add_subdirectory`.
This has a few **MAJOR** benefits:
- CoreFoundation changes get tracked and cause compile/link triggers
- CoreFoundation warnings are surfaced in the build
- CoreFoundation dependencies are reflected in the right location
- Foundation dependencies are reflected in the right location
- The overall build is simpler

Eventually, it should be possible to further simplify this.

Unfortunately, a couple of hacks are needed to make this work.  The
sub-projects properties must be cached, but they were just passthrough
previously, so this is not too terrible.  Additionally, the library
control know needs to be saved/restored around it.  The worst thing here
is the need to manually collect the dependency as we do not have a
proper library target for the Foundation (Swift) library.  When CMake
gains proper Swift support, it should be possible to use `add_library`
and get a property target that we can use `target_link_libraries` with
permitting the dependency tracking to allow us to get that without
manual intervention.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,34 +27,14 @@ include(ExternalProject)
 string(TOLOWER ${CMAKE_SYSTEM_NAME} swift_os)
 get_swift_host_arch(swift_arch)
 
-ExternalProject_Add(CoreFoundation
-                    SOURCE_DIR
-                      ${CMAKE_CURRENT_SOURCE_DIR}/CoreFoundation
-                    CMAKE_COMMAND
-                      ${CMAKE_COMMAND}
-                    CMAKE_ARGS
-                      -DBUILD_SHARED_LIBS=NO
-                      -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
-                      -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-                      -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
-                      -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
-                      -DCMAKE_INSTALL_LIBDIR=usr/lib
-                      -DCMAKE_SYSTEM_NAME=${CMAKE_SYSTEM_NAME}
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @millenomi @gwynne

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
apple/swift-corelibs-xctest#261

@swift-ci please test

compnerd:
apple/swift-corelibs-xctest#261
apple/swift-llbuild#460

@swift-ci please test

compnerd:
apple/swift-corelibs-xctest#261
apple/swift-llbuild#460
apple/swift-package-manager#2054

@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
apple/swift-corelibs-xctest#261
apple/swift-llbuild#460
apple/swift-package-manager#2054

@swift-ci please test

compnerd:
apple/swift#23775
apple/swift-package-manager#2054

@swift-ci please test

compnerd:

apple/swift#23869
apple/swift-package-manager#2054

@swift-ci please test

adrian-prantl:
I think this may have broken the LLDB tests. They can't locate Foundation swift module any more. I think we may need to update swift/utils/build-script-impl ?

`DOTEST_EXTRA="-Xcc -F${FOUNDATION_BUILD_DIR}/CoreFoundation-prefix/System/Library/Frameworks"`

compnerd:
@adrian-prantl pointed me to https://ci.swift.org/view/LLDB/job/oss-lldb-swift-5.1-incremental-linux-ubuntu-18_04/378/consoleFull#-3714861073122a513-f36a-4c87-8ed7-cbc36a1ec144 which is a 5.1 build.  Seems that we need to cross-port the migration path to 5.1.  https://github.com/apple/swift/pull/23906 should take care of that.

PR #2005, opened 2019-03-14
TestFoundation: use Thread.sleep instead of usleep/sleep

compnerd:


compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

PR #2009, opened 2019-03-14
TestFileHandle: make the pipe construction work on Windows

compnerd:
Windows does not provide `pipe` but has an equivalent `CreatePipe` API.
Adjust the test for this.

### DIFF
--- a/TestFoundation/TestFileHandle.swift
+++ b/TestFoundation/TestFileHandle.swift
@@ -66,8 +66,22 @@ class TestFileHandle : XCTestCase {
         allTemporaryFileURLs.append(url)
         return fh!
     }
-    
+
+#if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
     func createFileHandleForSeekErrors() -> FileHandle {
+#if os(Windows)
+        var hReadPipe: HANDLE? = INVALID_HANDLE_VALUE
+        var hWritePipe: HANDLE? = INVALID_HANDLE_VALUE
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Foundation/FileHandle.swift
```diff
@@ -377,6 +377,11 @@ open class FileHandle : NSObject, NSSecureCoding {
       _closeOnDealloc = closeopt
     }
 
+    public init(_handle: HANDLE, _closeOnDealloc: Bool) {
```
spevans: Should this not be `internal` and the test wrapped in `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT` ? Otherwise its adding new API.

millenomi: This patch will be hanging for a bit as I resolve this. (I am okay having experiemental (underscored) API for the use of platforms where there is no alternative. On Windows, where `fd`s are `HANDLE`s, there's little choice.)

A nit: the underscore is only on the first parameter.

compnerd: Since I put this up, a `FileHandle(handle:closeOnDealloc:)` has been added as `internal`.  This can go away.

compnerd:
@swift-ci please test

compnerd:

compnerd:

compnerd:

PR #2010, opened 2019-03-15
TestFoundation: use `.standardizingPath` for normalization

compnerd:
Rather than use `realpath` which is not universally available, use
`.standardizingPath` which should normalize the path as `realpath` on
all the targets since we are testing `Process` rather than
`FileManager`.

### DIFF
--- a/TestFoundation/TestProcess.swift
+++ b/TestFoundation/TestProcess.swift
@@ -281,27 +281,8 @@ class TestProcess : XCTestCase {
         }
     }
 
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @millenomi @spevans

compnerd:
@swift-ci please test

spevans:
Agreed, we shouldn't use any per-platform system calls in the tests, better to use something provided by Foundation itself to ensure it gets extra testing.

millenomi:

PR #2012, opened 2019-03-15
TestFoundation: excise tests that use private APIs

compnerd:
Since we do not expose the CoreFoundation APIs fully these tests are not
possible to build since the symbols cannot be resolved.  Rather than
adding additional exports to CoreFoundation for testing only, just drop
the tests.

### DIFF
--- a/TestFoundation/TestBundle.swift
+++ b/TestFoundation/TestBundle.swift
@@ -252,7 +252,7 @@ class TestBundle : XCTestCase {
             ("test_bundlePreflight", test_bundlePreflight),
             ("test_bundleFindExecutable", test_bundleFindExecutable),
             ("test_bundleFindAuxiliaryExecutables", test_bundleFindAuxiliaryExecutables),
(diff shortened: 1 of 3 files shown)


compnerd:
@millenomi - was this the solution that you preferred rather than having a new macro to expose the functions?

spevans:
Could `internal` methods be added to wrap these then testing enabled inside `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT` blocks?  I don't like the idea of disabling so many tests

millenomi:
Or: use `@_cdecl` to pull in strictly SPI we need in the test target.

compnerd:
Using `@_cdecl` wont work, since the symbols are not available through `Foundation.dll`.  Wrapping them in an SPI is probably nicer.

compnerd:
The only test which needed to be removed is handled elsewhere.  The issues that this was tracking are no longer a problem.

PR #2022, opened 2019-03-19
[SR-9887] HTTPCookie on Linux does not accept Substring values

saiHemak:
The HTTPCookie  initializer takes values as ```Any ```  but fails to create cookie when Substring is passed to it. Casting from ```Substring``` to String returns ```nil``` due to which the initializer is failing silently.

### DIFF
--- a/Foundation/HTTPCookie.swift
+++ b/Foundation/HTTPCookie.swift
@@ -255,10 +255,16 @@ open class HTTPCookie : NSObject {
     /// - Experiment: This is a draft API currently under consideration for official import into Foundation as a suitable alternative
     /// - Note: Since this API is under consideration it may be either removed or revised in the near future
     public init?(properties: [HTTPCookiePropertyKey : Any]) {
+        func stringValue(_ strVal: Any?) -> String? {
+            if let subStr = strVal as? Substring {
+                return String(subStr)
+            }
+            return strVal as? String
+        }
         guard
-            let path = properties[.path] as? String,
-            let name = properties[.name] as? String,
-            let value = properties[.value] as? String
+            let path = stringValue(properties[.path]),
+            let name = stringValue(properties[.name]),
+            let value = stringValue(properties[.value])
         else {
             return nil
         }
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

Foundation/HTTPCookie.swift
```diff
@@ -377,6 +377,15 @@ open class HTTPCookie : NSObject {
         }
     }
 
+    private class func  getStringValue(strVal: Any?) -> String? {
```
pushkarnk: Three suggestions here:
1. Though this function name serves the purpose, we could shorten it, may be something like `stringValue`. Just a suggestion :)
2. Could this function be nested inside the initialiser where it is being used?
2. Because we only seem to be adding support for Substrings, do we really need the `strVal as? String` check?

spevans:
@swift-ci test

pushkarnk:
@swift-ci test

Foundation/HTTPCookie.swift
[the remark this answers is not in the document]
karthikkeyan: the words StringValue in the method name and the `strVal` parameter label reads redundant.

Foundation/HTTPCookie.swift
```diff
@@ -255,10 +255,16 @@ open class HTTPCookie : NSObject {
     /// - Experiment: This is a draft API currently under consideration for official import into Foundation as a suitable alternative
     /// - Note: Since this API is under consideration it may be either removed or revised in the near future
     public init?(properties: [HTTPCookiePropertyKey : Any]) {
+        func  stringValue(strVal: Any?) -> String? {
```
ianpartridge: nit: extra space after `func`.

ianpartridge: I think the parameter name is unnecessary.

Foundation/HTTPCookie.swift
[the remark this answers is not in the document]
saiHemak: done ..thanks

spevans:
@swift-ci test

PR #2023, opened 2019-03-19
[4.2] JSONSerialization: Improve parsing of numbers

spevans:
- Check the number looks like a JSON number and exit early if not.

- Use the native Int64(), UInt64(), Double() parsers to avoid creating
  a C string and passing to strtol()/strtod(). This also eliminates a
  memcpy() and removes the 63 digit restriction which would fail to
  parse numbers expressible by Double's full exponent.

- For numbers with a leading '-' sign, parse using Int64() falling
  back to Double(), otherwise parse using UInt64() failling back to
  Double().

(cherry picked from commit ff5f81e42402bd425810466a6e5d8756006e99b5)

- Fully validate that the number conforms to the JSON number
  specification.

- Determine if the number should be parsed as a UInt64, Int64 or
  Decimal before falling back to Decimal.

(cherry picked from commit 1e32417299c577be1fb94d4ef9da92e0b3f99bd8)

Update TestJSONEncoder to account for new error message.

(cherry picked from commit c69ec857c704e31b4ed90332843fd1e4c754e5ce)

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -816,52 +816,149 @@ private struct JSONReader {
     }
     
     //MARK: - Number parsing
-    static let numberCodePoints: [UInt8] = [
-        0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, // 0...9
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test 4.2

PR #2025, opened 2019-03-20
TestFoundation: emulate `setenv` using `putenv` on Win32

compnerd:
Windows does not have `setenv`, but it does have a `putenv`, which we
can use in its stead.


compnerd:
@swift-ci please test

spevans:
LGTM

PR #2026, opened 2019-03-20
Foundation: initialize CFSocket on startup

compnerd:
Windows requires the networking subsystem (WinSock) to be initialized.
However, the user expects to be able to use the networking APIs in
Foundation without performing any one-time initialization.  Furthermore,
due to the macro usage in the initialization path, it is easier to
perform the initialization in C than in Swift.  CoreFoundation has a SPI
for this (`__CFSocketInitializeWinSock`).

### DIFF
--- a/Foundation/NSSwiftRuntime.swift
+++ b/Foundation/NSSwiftRuntime.swift
@@ -150,7 +150,10 @@ internal func __CFSwiftGetBaseClass() -> UnsafeRawPointer {
 @usableFromInline
 @_cdecl("__CFInitializeSwift")
 internal func __CFInitializeSwift() {
-    
+#if os(Windows)
+    __CFInitializeWinSock()
+#endif
+
     _CFRuntimeBridgeTypeToClass(CFStringGetTypeID(), unsafeBitCast(_NSCFString.self, to: UnsafeRawPointer.self))
     _CFRuntimeBridgeTypeToClass(CFArrayGetTypeID(), unsafeBitCast(_NSCFArray.self, to: UnsafeRawPointer.self))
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

Foundation/NSSwiftRuntime.swift
```diff
@@ -147,10 +147,18 @@ internal func __CFSwiftGetBaseClass() -> UnsafeRawPointer {
     return unsafeBitCast(__NSCFType.self, to:UnsafeRawPointer.self)
 }
 
+#if os(Windows)
+@_silgen_name("__CFSocketInitializeWinSock")
```
spevans: I doubt `@_silgen_name` will be allowed outside of stdlib. I think you need to use a `@_cdecl` instead.

compnerd: Can you use `@_cdecl` with a declaration and use it?  I don't want to put it into `__CFInitialize` since it will change the semantics of CoreFoundation outside of Foundation.

spevans: Thinking about it, cant you just declare it in `ForSwiftFoundationOnly.h` ?
Alternatively when `__CFInitializeSwift` is called it is already gated on being inside of Foundation:
```
#if DEPLOYMENT_RUNTIME_SWIFT
        extern void __CFInitializeSwift(void);
        __CFInitializeSwift();
#endif
```
So could it just be added inside of this block?

compnerd: Yeah, I can declare it in `ForSwiftFoundationOnly.h`, that's perfectly fine by me.

compnerd:
@swift-ci please test

spevans:
@swift-ci please test

PR #1540, opened 2018-04-29
Always use FS representation of String when calling FS syscalls

spevans:
- Update calls to stat(), lstat(), access(), link(), unlink(),
  symlink(), readlink(), rename(), chmod(),  mkdir() and chdir().

- Fix deallocating the fs representation in FileManager._lstatFile(withPath:)
  and URL.withUnsafeFileSystemRepresentation().

- Add FileManager._copySymlink(atPath:toPath:) to copy a symlink
  without the roundtrip via String.

PR #2027, opened 2019-03-20
[4.2] JSONSerialization: fix a crash when an escape sequence contains low-s…

spevans:
…urrogate code point without the corresponding high-surrogate code point

(cherry picked from commit 9eca3c07f22160f2b6625a818302b9936ae887cb)

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -738,19 +738,40 @@ private struct JSONReader {
             return nil
         }
 
-        if !UTF16.isLeadSurrogate(codeUnit) {
+        let isLeadSurrogate = UTF16.isLeadSurrogate(codeUnit)
+        let isTrailSurrogate = UTF16.isTrailSurrogate(codeUnit)
+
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test 4.2

millenomi:
I don't know that we have any more 4.2 release vehicles. @weissi?

@spevans: can this be cherry-picked to 5.0 if it isn't there?

weissi:
@millenomi/@spevans I think that one is quite important so I think we could make a release for this soonish, let’s see if there are other important issues like this one coming in.

We should definitely also propose this PR for Swift 5. We’re starting to gear up the monthlies for Swift 5 🙌

spevans:
@millenomi This one is already in 5.0, for all my cherry-picks to 4.2 I made sure they came from 5.0

@weissi I have a use after free to backport to 5.0 and potentially 4.2 #2038 and a leak from #1540 which might also be worth doing if there is going to be one last 4.2 release.

millenomi:
@spevans Please send 'em.

PR #2028, opened 2019-03-20
[CoreFoundation] Fix enum name in function body.

drodriguez:
The enum name was probably autocompleted wrong. There should be no
functional change. LLVM 8 seems to raise a new warning here.


spevans:
@swift-ci test

PR #2029, opened 2019-03-21
[SR-9979] FileHandle class used to implement FileManager._compareFiles

karthikkeyan:
`FileHandler` APIs used to replace the curret implementation of the private method
`_compareFiles(withFileSystemRepresentation:andFileSystemRepresentation:size:bufSize:)`

Bug Link: https://bugs.swift.org/browse/SR-9979

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -342,6 +342,14 @@ open class FileHandle : NSObject, NSSecureCoding {
         }
 #endif
     }
+    
+    internal func _readBytes(into buffer: UnsafeMutablePointer<UInt8>, length: Int) throws -> Int {
+        let amtRead = _read(_fd, buffer, length)
+        if amtRead < 0 {
+            throw _NSErrorWithErrno(errno, reading: true)
+        }
+        return amtRead
+    }
 
     internal func _writeBytes(buf: UnsafeRawPointer, length: Int) throws {
 #if os(Windows)
@@ -425,6 +433,15 @@ open class FileHandle : NSObject, NSSecureCoding {
         }
     }
 #endif
+    
+    internal init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
+        _fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
+        _closeOnDealloc = true
+        super.init()
+        if _fd < 0 {
+            return nil
+        }
+    }
 
     deinit {
         // .close() tries to wait after operations in flight on the handle queue, if one exists, and then close. It does so by sending .sync { … } work to it.
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -433,14 +433,16 @@ open class FileHandle : NSObject, NSSecureCoding {
         }
     }
 #endif
-    
-    internal init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
-        _fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
-        _closeOnDealloc = true
-        super.init()
-        if _fd < 0 {
-            return nil
-        }
+
+    internal convenience init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
+        let fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
+        guard fd > 0 else { return nil }
+        
+#if os(Windows)
+        self.init(handle: HANDLE(bitPattern: _get_osfhandle(fd))!, closeOnDealloc: true)
+#else
+        self.init(fileDescriptor: fd, closeOnDealloc: true)
+#endif
     }
 
     deinit {
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -344,11 +344,20 @@ open class FileHandle : NSObject, NSSecureCoding {
     }
     
     internal func _readBytes(into buffer: UnsafeMutablePointer<UInt8>, length: Int) throws -> Int {
+#if os(Windows)
+        var BytesRead: DWORD = 0
+        let BytesToRead: DWORD = DWORD(length)
+        if ReadFile(_handle, buffer, BytesToRead, &BytesRead, nil) == FALSE {
+            throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+        }
+        return Int(BytesRead)
+#else
         let amtRead = _read(_fd, buffer, length)
         if amtRead < 0 {
             throw _NSErrorWithErrno(errno, reading: true)
         }
         return amtRead
+#endif
     }
 
     internal func _writeBytes(buf: UnsafeRawPointer, length: Int) throws {
@@ -438,11 +447,7 @@ open class FileHandle : NSObject, NSSecureCoding {
         let fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
         guard fd > 0 else { return nil }
         
-#if os(Windows)
-        self.init(handle: HANDLE(bitPattern: _get_osfhandle(fd))!, closeOnDealloc: true)
-#else
         self.init(fileDescriptor: fd, closeOnDealloc: true)
-#endif
     }
 
     deinit {
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -444,9 +444,15 @@ open class FileHandle : NSObject, NSSecureCoding {
 #endif
 
     internal convenience init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
+#if os(Windows)
+        var fd: Int = -1
+        if let path = String(cString: fileSystemRepresentation).cString(using: .utf16) {
+            fd = _CFOpenFileWithMode(path, flags, mode_t(createMode))
+        }
+#else
         let fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
+#endif
         guard fd > 0 else { return nil }
-        
         self.init(fileDescriptor: fd, closeOnDealloc: true)
     }
 
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -1729,7 +1729,6 @@ open class FileManager : NSObject {
             buffer1.deallocate()
             buffer2.deallocate()
         }
-        
         var bytesLeft = size
         while bytesLeft > 0 {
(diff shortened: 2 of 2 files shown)


compnerd:
@swift-ci please test

spevans:
Thanks for this, I just have a few comments:

Since the filenames are already in FSR (File System Representation), converting them back to a `String` is unnecessary, especially since that `String` will then be converted back to an FSR to use in the  `open` call to actually open the file.

I would suggest adding a new `init` method to `FileHandle` along the lines of `internal init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int)` which takes an FSR instead of a `String`  and defaults to `_closeOnDealloc = true`.

For the read loop, `FileHandle.readData(ofLength:)` is actually quite heavy since it will involve calling `fstat()` and a `malloc()/free()`  for every call in the loop.

I would add another `FileHandle` method, something like:
`internal func _readBytes(into: UnsafePointer<UInt8>, count: Int) throws -> Int`  which just wraps the OS `read` system call and also refactor the inner read loop of `_readDataOfLength` to use it.

This could make the `_compareFiles` look something like:

```
guard let file1 = FileHandle(fileSystemRepresentation: file1Rep, ...) else { return false }
guard let file2 = FileHandle(fileSystemRepresentation: file2Rep, ...) else { return false }

let buffer1 = UnsafeMutablePointer<UInt8>.allocate(capacity: bufSize)
let buffer2 = UnsafeMutablePointer<UInt8>.allocate(capacity: bufSize)
defer {
    buffer1.deallocate()
    buffer2.deallocate()
}


var bytesLeft = size
while bytesLeft > 0 {
    let bytesToRead = Int(min(Int64(bufSize), bytesLeft))
    guard let bytesRead = try? file1._readBytes(into: buffer1, count: bytesToRead),
        bytesRead == bytesToRead else {
	return false
    }
    guard let bytesRead = try? file2._readBytes(into: buffer2, count: bytesToRead),
        bytesRead == bytesToRead else {
	return false
    }
    guard memcmp(buffer1, buffer2, bytesToRead) == 0 else {
	return false
    }
    bytesLeft -= Int64(bytesToRead)
}
    
return true
```

karthikkeyan:
@spevans  Thanks for reviewing my PR Simon. I have updated the code as per your suggestion. Now it is ready for the next round of review.

Foundation/FileHandle.swift
```diff
@@ -342,6 +342,14 @@ open class FileHandle : NSObject, NSSecureCoding {
         }
 #endif
     }
+    
+    internal func _readBytes(into buffer: UnsafeMutablePointer<UInt8>, length: Int) throws -> Int {
+        let amtRead = _read(_fd, buffer, length)
```
compnerd: Hmm, is `_read` the internal method?  Shouldn't that just throw if `read` returned `< 0`?  `_read` is also a UCRT function on Windows, so please use `self._read` to make it more obvious.

Foundation/FileHandle.swift
```diff
@@ -425,6 +433,15 @@ open class FileHandle : NSObject, NSSecureCoding {
         }
     }
 #endif
+    
+    internal init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
```
compnerd: I don't think that we should do this.  The problem is that on Windows, you need to use UTF16 (or Unicode) for the filenames for the FSR otherwise you will not be able to open all files.

Foundation/FileManager.swift
```diff
@@ -1720,39 +1720,35 @@ open class FileManager : NSObject {
...
+        guard let file2 = FileHandle(fileSystemRepresentation: file2Rep, flags: O_RDONLY, createMode: 0) else { return false }
+        
+        var buffer1 = UnsafeMutablePointer<UInt8>.allocate(capacity: bufSize)
+        var buffer2 = UnsafeMutablePointer<UInt8>.allocate(capacity: bufSize)
         defer {
             buffer1.deallocate()
             buffer2.deallocate()
         }
-
+
```
compnerd: Bleeding whitespace

Foundation/FileHandle.swift
karthikkeyan: The methods `_compareFiles()` and `contentEquals()` does not support Windows platform, yet. We will cross the bridge when we get to it. As for as this PR concern it seems to me, it is safe to wrap this `init` method under `#if !os(Windows)`. What do you think @compnerd?

compnerd: The SR was filed specifically to enable Windows support, so I think that it should be part of this change.  Using the `FileHandle` was the way around the FSR conversions.

Foundation/FileManager.swift
karthikkeyan: The space between `defer` and the next likes seems necessary, but If you feel strong about this I will update it.

Also while we are in the topic, I have noticed a couple more unnecessary whitespaces as will in the same method; I will remove them too.

Foundation/FileHandle.swift
karthikkeyan: This `_read` is not seems to be an internal method, it is a reference of either Darwin.read(_:_:_:) or Glibc.read(_:_:_:) method.

Foundation/FileHandle.swift
karthikkeyan: Sounds good. I will update the code to support Windows as well.

karthikkeyan: @compnerd  Thanks for the suggestions. I have updated the init method. Now it supports windows as well. Please do a round of review when you get a chance.

Foundation/FileHandle.swift
karthikkeyan: Since I am do not have experience in Windows, I just want to run by you the following implementation for `_readBytes()` method,

```swift
    internal func _readBytes(into buffer: UnsafeMutablePointer<UInt8>, length: Int) throws -> Int {
#if os(Windows)
        var BytesRead: DWORD = 0
        let BytesToRead: DWORD = DWORD(length)
        if ReadFile(_handle, buffer, BytesToRead, &BytesRead, nil) == FALSE {
        }
        return Int(BytesRead)
#else
        let amtRead = _read(_fd, buffer, length)
        if amtRead < 0 {
            throw _NSErrorWithErrno(errno, reading: true)
        }
        return amtRead
#endif
    }
```

karthikkeyan: If the above code implementation works, I will update the method as well.

compnerd: Yeah, the use of `ReadFile` is fine.

Foundation/FileHandle.swift
```diff
@@ -426,6 +434,17 @@ open class FileHandle : NSObject, NSSecureCoding {
     }
 #endif
 
+    internal convenience init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
+        let fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
+        guard fd > 0 else { return nil }
+        
+#if os(Windows)
+        self.init(handle: HANDLE(bitPattern: _get_osfhandle(fd))!, closeOnDealloc: true)
```
compnerd: Why can't you use the `init(fileDescriptor:closeOnDealloc:)`?

Foundation/FileHandle.swift
compnerd: Yeah, that implementation is fine.

Foundation/FileHandle.swift
spevans: Maybe a better solution would be to add an internal typealias for FSR
```
#if os(windows)
internal typealias _FileSystemRepresentation = UnsafePointer<UInt16>
#else
internal typealias _FileSystemRepresentation = UnsafePointer<Int8>
#endif
```

And use that in the signatures for these internal functions. I guess `FileManager.fileSystemRepresentation(withPath path: String) -> UnsafePointer<Int8>` is technically broken on Windows anyway but can probably be ignored for now.

Foundation/FileManager.swift
spevans: Having an empty line is not an issue, its the trailing whitespace which is the problem (the large green rectangle).

Foundation/FileHandle.swift
compnerd: Hmm, we don't want to break the public interfaces, I think that we should convert across the encodings.

Foundation/FileHandle.swift
karthikkeyan: I did not notice that `init` method; I will update the code. Thank you.

karthikkeyan:
Thank you @spevans @compnerd, I think I have addressed all the comments, except the new `typealias _FileSystemRepresentation` which I could not quite figure out whats the expectation. Please give it another round of review when you get a chance.

Foundation/FileHandle.swift
```diff
@@ -426,6 +443,13 @@ open class FileHandle : NSObject, NSSecureCoding {
     }
 #endif
 
+    internal convenience init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
+        let fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
+        guard fd > 0 else { return nil }
+        
+        self.init(fileDescriptor: fd, closeOnDealloc: true)
+    }
```
compnerd: I think that instead of the FSR typealias, we should do something like this:

```
internal convenience init?(fileSystemRepresentation: UnsafePointer<Int8>, flags: Int32, createMode: Int) {
#if os(Windows)
  var fd: Int -1
  if let path = String(cString: fileSystemRepresentation).withCString(encodedAs: UTF16.self) {
    fd = _CFOpenFileWithMode($0, flags, mode_t(createMode))
  }
#else
  let fd = _CFOpenFileWithMode(fileSystemRepresentation, flags, mode_t(createMode))
#endif
  guard fd > 0 else { return nil }
  self.init(fileDescriptor: fd, closeOnDealloc: true)
}
```

karthikkeyan: I think you mean to use `.cString(using: .utf16)` method?

karthikkeyan: It is done.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

karthikkeyan:
@compnerd Looks like the compilation failed. I have built and ran the test cases using Xcode-Beta 10.2 beta 4 (10P107d) using Swift Development Snapshot 2019-03-17 (a) snapshot, successfully. Not sure if it is CI setup issue.

spevans:
@swift-ci please test

spevans:
@karthikkeyan Don't worry too much about my `typealias _FileSystemRepresentation` idea, it was just a vague notation, I may try prototyping it in a separate PR to see if its any benefit but its not a blocker for this PR.

PR #2030, opened 2019-03-22
TestNSNumber: Disable testNSNumberToBool() that breaks on Darwin.

spevans:
- This still works on Linux so disable it conditionally.


spevans:
@swift-ci test

PR #2031, opened 2019-03-22
[4.2] disable test_currencyPluralMinimumIntegerDigits

weissi:
test_currencyPluralMinimumIntegerDigits does not and did never work on
Ubuntu 14.04. In Swift 5 it works because a newer ICU is vendored which
fixes the issue. But in Swift 4.2.x we don't vendor ICU so we can't get
this test to work on Ubuntu 14.04, hence removing as it's no worse than
Swift 4.2.3.

### DIFF
--- a/TestFoundation/TestNumberFormatter.swift
+++ b/TestFoundation/TestNumberFormatter.swift
@@ -30,7 +30,6 @@ class TestNumberFormatter: XCTestCase {
             ("test_scientificMinimumIntegerDigits", test_scientificMinimumIntegerDigits),
             ("test_spellOutMinimumIntegerDigits", test_spellOutMinimumIntegerDigits),
             ("test_ordinalMinimumIntegerDigits", test_ordinalMinimumIntegerDigits),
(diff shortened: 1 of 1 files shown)


weissi:
@swift-ci please test 4.2

kevints:
Is there any way to make this conditional on the Ubuntu flavor in use? Maybe use ICU version as a proxy? It seems like we're likely to restore this regression without this test.

weissi:
> Is there any way to make this conditional on the Ubuntu flavor in use? Maybe use ICU version as a proxy? It seems like we're likely to restore this regression without this test.

Linux flavours and versions are not covered by the `#if os(...)` conditionals.

weissi:
@swift-ci please test linux

spevans:
@kevints Pre 5.0 a lot of tests were simply commented out because they failed on 14.04. Unfortunately the CI doesn't include 14.04 so these sort of failures are not picked up pre-merge.

PR #2035, opened 2019-03-23
Extend URLResponse tests

matis-schotte:
This is my first rather small contribution to the swift project.
I tried to start with updating/adding a few tests for URLResponse.
Since I am still on High Sierra I was working on the 4.2 branch.

Please let me know if there is anything wrong or things I can do better next time.

PR #2036, opened 2019-03-23
Implement Host.isEqual and Test

khashiguchi:
This is my first rather small contribution to the swift project.
I implemented Host.isEqual method and test.

### DIFF
--- a/Foundation/Host.swift
+++ b/Foundation/Host.swift
@@ -76,7 +76,7 @@ open class Host: NSObject {
     
     open func isEqual(to aHost: Host) -> Bool {
         if self === aHost { return true }
-        return _addresses.firstIndex { aHost._addresses.contains($0) } != nil
+        return addresses.firstIndex { aHost.addresses.contains($0) } != nil
     }
     
     internal func _resolveCurrent() {
--- a/Foundation/Host.swift
+++ b/Foundation/Host.swift
@@ -75,7 +75,8 @@ open class Host: NSObject {
     }
     
     open func isEqual(to aHost: Host) -> Bool {
-        return aHost._addresses == _addresses && aHost._info == _info && aHost._names == _names && aHost._resolved == _resolved && aHost._type == _type
+        if self === aHost { return true }
+        return _addresses.firstIndex { aHost._addresses.contains($0) } != nil
     }
     
     internal func _resolveCurrent() {
--- a/Foundation/Host.swift
+++ b/Foundation/Host.swift
@@ -75,7 +75,7 @@ open class Host: NSObject {
     }
     
     open func isEqual(to aHost: Host) -> Bool {
-        return false
+        return aHost._addresses == _addresses && aHost._info == _info && aHost._names == _names && aHost._resolved == _resolved && aHost._type == _type
     }
     
     internal func _resolveCurrent() {
--- a/TestFoundation/TestHost.swift
+++ b/TestFoundation/TestHost.swift
@@ -12,8 +12,7 @@ class TestHost: XCTestCase {
     static var allTests: [(String, (TestHost) -> () throws -> Void)] {
         return [
             ("test_addressesDoNotGrow", test_addressesDoNotGrow),
-            ("test_isEqual_positive", test_isEqual_positive),
-            ("test_isEqual_negative", test_isEqual_negative)
+            ("test_isEqual", test_isEqual)
         ]
     }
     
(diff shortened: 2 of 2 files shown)


khashiguchi:
@ikesyo Open Source Swift workshopに参加しました。初めてのオープンソースです。
よろしくお願いします！

Foundation/Host.swift
```diff
@@ -75,7 +75,7 @@ open class Host: NSObject {
     }
     
     open func isEqual(to aHost: Host) -> Bool {
-        return false
+        return aHost._addresses == _addresses && aHost._info == _info && aHost._names == _names && aHost._resolved == _resolved && aHost._type == _type
```
ikesyo: `NSHost.isEqual(to:)` is documented as follows:

https://developer.apple.com/documentation/foundation/host/1414786-isequal

> true when the receiver and host share at least one network address; false otherwise.

So this does not seem true. Maybe a naive implementation look like the following?

```swift
if self === other { return true }
return _addresses.firstIndex { aHost._addresses.contains($0) } != nil
```

TestFoundation/TestHost.swift
```diff
@@ -32,5 +34,17 @@ class TestHost: XCTestCase {
...
+        let host1 = Host(address: "8.8.8.8")
+        XCTAssertTrue(host0.isEqual(to: host1))
+    }
+
+    func test_isEqual_negative() {
+        let host0 = Host(address: "8.8.8.8")
+        let host1 = Host(address: "8.8.8.9")
+        XCTAssertFalse(host0.isEqual(to: host1))
+    }
 }
```
spevans: I would just write these as one test (`test_isEqual`) with the  `XCTAssertTrue` and `XCTAssertFalse`. Thanks

Foundation/Host.swift
khashiguchi: Thank you for watching.
I did not read the document enough.
However, for some reason the test fails or succeeds.
I don't know the cause, I will look a little more closely.

TestFoundation/TestHost.swift
khashiguchi: Thank you for watching.
I fixed that.

ikesyo:
@swift-ci test

TestFoundation/TestHost.swift
matis-schotte: Is there any guideline on how to structure the tests?

I usually prefer to make them smaller with only one goal in mind. So in this case one is the positive (isEqual), the other the negative test (not isEqual).

spevans: In general I would use 1 test function per method that is being tested, testing both positive and negative values as required. Seeing the whole test function fail to pass will indicate that there is an issue with the method implementation.

Using 1 function per test will just lead to an excess of functions. Bear in mind that some types/classes do not have much functionality at all so may only require a few test functions, whereas large complex types, eg `JSONSerialization` require lots of tests for each method.

spevans:
@swift-ci test

TestFoundation/TestHost.swift
matis-schotte: Thank you very much for your answer. Gonna update the PR #2035 then accordingly.

matis-schotte:
Build failed because of "/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux@2/swift-corelibs-libdispatch/src/swift/Block.swift:14:8: error: no such module '_SwiftDispatchOverlayShims'", even though there is nothing changed in that file in this PR.
I guess its referencing the most current commit on master for libdispatch?

ikesyo:
@swift-ci test Linux platform

Foundation/Host.swift
matis-schotte: The internal `_addresses` is empty before the Host is being resolved, this can be seen in the `open var addresses` which calls `_resolve()` first.

That's why the test fails, @khashiguchi, replacing `_addresses` with `addresses` fixes the problem:
```
open func isEqual(to aHost: Host) -> Bool {
    if self === aHost { return true }
    return addresses.firstIndex { aHost.addresses.contains($0) } != nil
}
```

matis-schotte: Also the test could be extended to also check for hostnames:
```
func test_isEqual() {
    let host0 = Host(address: "8.8.8.8")
    let host1 = Host(address: "8.8.8.8")
    XCTAssertTrue(host0.isEqual(to: host1))
        
    let hostname0 = Host(name: "google.com")
    let hostname1 = Host(name: "google.com")
    XCTAssertTrue(hostname0.isEqual(to: hostname1))

    let host2 = Host(address: "8.8.8.9")
    XCTAssertFalse(host0.isEqual(to: host2))
    let hostname2 = Host(name: "apple.com")
    XCTAssertFalse(hostname0.isEqual(to: hostname2))
}
```

khashiguchi: I could not confirm it easily.
Thank you very much, @matis-schotte .
I will fix it early!

ikesyo:
@swift-ci test

millenomi:
@swift-ci please test

millenomi:
@ikesyo it looks like your concerns were fixed, can you confirm

PR #2038, opened 2019-03-23
_BodyFileSource: Fix use after free in init()

spevans:
- init(fileURL:workQueue:dataAvailableHandler:) was using the closure
  argument for the buffer pointer outside of the closure, after
  .deallocate() was called.


spevans:
@swift-ci test

pushkarnk:
@swift-ci test and merge

PR #2040, opened 2019-03-24
[SR-10156] Fix test_current_working_directory on macOS

nafu:
Resolve: https://bugs.swift.org/browse/SR-10156

## Motivation

TestProcess.test_current_working_directory fails on macOS. This fixes the test failure on macOS.

```bash
Test Case 'TestProcess.test_current_working_directory' started at 2019-03-23 12:09:37.998
/Users/nafu/Projects/swift-source/swift-corelibs-foundation/TestFoundation/TestProcess.swift:294: error: TestProcess.test_current_working_directory : XCTAssertEqual failed: ("/private/tmp") is not equal to ("/tmp") - 
Test Case 'TestProcess.test_current_working_directory' failed (0.623 seconds)
```

### DIFF
--- a/TestFoundation/TestProcess.swift
+++ b/TestFoundation/TestProcess.swift
@@ -282,7 +282,11 @@ class TestProcess : XCTestCase {
     }
 
     func test_current_working_directory() {
+        #if os(macOS)
+        let tmpDir = "/private/tmp"
+        #else
         let tmpDir = "/tmp".standardizingPath
+        #endif
 
         let fm = FileManager.default
         let previousWorkingDirectory = fm.currentDirectoryPath


TestFoundation/TestProcess.swift
```diff
@@ -282,7 +282,11 @@ class TestProcess : XCTestCase {
     }
 
     func test_current_working_directory() {
+        #if os(macOS)
```
millenomi: We should just replace all of this with `NSTemporaryDirectory()`.

nafu: @millenomi Thank you for the comment!

If we use,

```swift
let tmpDir = NSTemporaryDirectory()
```

we will get these errors.

```swift
Test Case 'TestProcess.test_current_working_directory' started at 2019-04-01 01:22:45.245
/Users/nafu/Projects/swift-source/swift-corelibs-foundation/TestFoundation/TestProcess.swift:294: error: TestProcess.test_current_working_directory : XCTAssertEqual failed: ("/private/var/folders/gw/nv0tyl695m91txy7thrdppq40000gn/T") is not equal to ("/var/folders/gw/nv0tyl695m91txy7thrdppq40000gn/T/") - 
/Users/nafu/Projects/swift-source/swift-corelibs-foundation/TestFoundation/TestProcess.swift:303: error: TestProcess.test_current_working_directory : XCTAssertEqual failed: ("/var/folders/gw/nv0tyl695m91txy7thrdppq40000gn/T") is not equal to ("/var/folders/gw/nv0tyl695m91txy7thrdppq40000gn/T/") - 
```

[Q] Do you think it would be better we manipulate for two cases?

1. Add `/private/` and remove last `/` for first failure case
2. Remove last `/` for second failure case

shahmishal:

PR #2043, opened 2019-03-25
TestProcess: Improve reading from a pipe with a sub-process

spevans:
- TestProcess.runTask() uses a pipe to read the stdout of a subprocess
  during tests however sometimes the data read from the pipe is
  truncated causing issues with the testing of the response.

- Use .readabilityHandler to read the data whilst the sub-process is
  running and then drain the pipe after the process has terminated.

### DIFF
--- a/TestFoundation/TestProcess.swift
+++ b/TestFoundation/TestProcess.swift
@@ -609,19 +609,39 @@ internal func runTask(_ arguments: [String], environment: [String: String]? = ni
     let stderrPipe = Pipe()
     process.standardOutput = stdoutPipe
     process.standardError = stderrPipe
+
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

millenomi:
I'm so glad this works now.

spevans:
@millenomi One thing I found when using the new `FileHandle` API is that there is no `throw`ing equivalent to `.availableData`, ie nothing that ends up calling `_readDataOfLength(count, untilEOF: false)` which is a shame. One possible fix might be to modify the signature of `read(upToCount count: Int)` to `read(upToCount count: Int, untilEOF: Bool = true)` or some such.

PR #2046, opened 2019-03-27
Fix: existing attributes are not replaced

hironytic:
According to [the document of Darwin Foundation](https://developer.apple.com/documentation/foundation/xmlelement/1388336-addattribute)  of `XMLElement.addAttribute(_:)`,

> If the receiver already has an attribute with the same name, `anAttribute` replaces the old attribute.

and actually it replaces the old attribute on Darwin.

However in swift-corelibs-foundation, `addAttribute(_:)` do nothing when the attribute with the same name exists.

This patch fixes the behavior and improves compatibility with Darwin Foundation.

### DIFF
--- a/Foundation/XMLElement.swift
+++ b/Foundation/XMLElement.swift
@@ -82,17 +82,15 @@ open class XMLElement: XMLNode {
 
     /*!
         @method addAttribute:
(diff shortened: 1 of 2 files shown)


millenomi:
Good catch; thank you.

millenomi:
@swift-ci please test

PR #2047, opened 2019-03-27
Fix FileHandle Compilation on Windows

jmittert:
fileDescriptor should be an Int32, not an Int


jmittert:
@compnerd

compnerd:
*sigh*

compnerd:

PR #2051, opened 2019-03-30
HTTPCookieStorage: rewrite access to shared storage

Alexander-Ignition:
Removed access to `sharedStorage` through `sharedSyncQ` because the static constants are initialized lazily and in a thread-safe.

### DIFF
--- a/Foundation/HTTPCookieStorage.swift
+++ b/Foundation/HTTPCookieStorage.swift
@@ -36,8 +36,7 @@ extension HTTPCookie {
 */
 open class HTTPCookieStorage: NSObject {
 
-    /* both sharedStorage and sharedCookieStorages are synchronized on sharedSyncQ */
-    private static var sharedStorage: HTTPCookieStorage?
+    private static let sharedStorage = HTTPCookieStorage(cookieStorageName: "shared")
     private static var sharedCookieStorages: [String: HTTPCookieStorage] = [:] //for group storage containers
     private static let sharedSyncQ = DispatchQueue(label: "org.swift.HTTPCookieStorage.sharedSyncQ")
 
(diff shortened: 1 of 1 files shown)


Alexander-Ignition:
@swift-ci please test

spevans:
@swift-ci please test

pushkarnk:
I guess we have `shared` as a `class var` because the[ API doc](https://developer.apple.com/documentation/foundation/httpcookiestorage/1416095-shared) says so. And hence the synchronisation. However, I can see that `HTTPCookieStorage.shared` on Darwin is get-only, but it could have an `internal` or `private` setter. In short, I can't be sure if `shared` should be made a constant.

@millenomi @spevans Any thoughts?

spevans:
`private(set) var` should work to make the public access read-only.

pushkarnk:
@Alexander-Ignition Eager to know your thoughts too.

Alexander-Ignition:
Actually we don’t need sync code because of static is thread-safe by default. I checked no one change shared property (internally), so we don’t need `public var` or `private(set) var`. This approach with `private static let` and computed property for interface is common style in Foundation. `NotificationCenter `, `FileManager `,  `URLSession `

pushkarnk:
Right, if we agree to make it a constant (`class let shared`), then there's no need for the sync. But the question is: given that the API doc defines it as a `class var shared` (though publicly read-only), should we change it to a constant? I'm not sure if the Darwin's Foundation has internal/private setters defined. I just want to be careful about the parity.

pushkarnk:
cc @ianpartridge @weissi

spevans:
The public API only defines a getter so I don't think its a problem to use a constant internally since the API it still correct.

pushkarnk:
@swift-ci test and merge

pushkarnk:
@swift-ci test and merge

PR #2053, opened 2019-04-01
SR-10240: Dont try to write an empty Data() if it has a nil baseAddress.

spevans:
This doesnt actually crash in `master` as the region passed has a non-nil `baseAddress` and `bytes.count` = 0, but eliminate the force unwrap anyway incase the `baseAddress` ever becomes `nil` in the future.


spevans:
@swift-ci test

millenomi:

PR #2054, opened 2019-04-01
[5.0] SR-10240: Dont try to write an empty Data() as it has a nil baseAddress.

spevans:
5.0 version of #2053. This crashes in 5.0 as the region has `baseAddress = nil`

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -126,7 +126,9 @@ open class FileHandle : NSObject, NSSecureCoding {
         guard _fd >= 0 else { return }
         data.enumerateBytes() { (bytes, range, stop) in
             do {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

spevans:
cc @millenomi @weissi

Foundation/FileHandle.swift
```diff
@@ -126,7 +126,9 @@ open class FileHandle : NSObject, NSSecureCoding {
         guard _fd >= 0 else { return }
         data.enumerateBytes() { (bytes, range, stop) in
```
karthikkeyan: Just mentioning, looks like the method `enumerateBytes()` is deprecated in Swift 5.0. I am unsure if this PR is an appropriate place to make the changes, if not, I will update that the code in separate PR.

spevans: I was trying to keep the fix as minimal as possible, I may update to it in a separate PR.

millenomi:
@swift-ci please test

PR #2056, opened 2019-04-01
[4.2] _BodyFileSource: Fix use after free in init()

spevans:
- init(fileURL:workQueue:dataAvailableHandler:) was using the closure
  argument for the buffer pointer outside of the closure, after
  .deallocate() was called.

(cherry picked from commit 21549ff34a9cc0137ee48d842bcc28cbb675deb5)


spevans:
@swift-ci test 4.2

PR #2057, opened 2019-04-02
Split FileManager into Platform Dependent Files

jmittert:
Splits the Windows and POSIX specific File Manager implementations into their own files. See https://github.com/apple/swift-corelibs-foundation/pull/2002#issuecomment-477242876

Not sure if there's a way to make the diff more readable, but in general, what I've done is copy over the function to win32/POSIX, prefixed the name with an underscore, removed the non applicable #if branch, and have the FileManager version pass through to it.

### DIFF
--- a/Foundation/FileManager+POSIX.swift
+++ b/Foundation/FileManager+POSIX.swift
@@ -1,0 +1,1157 @@
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+#if !os(Windows)
+
+#if os(Android) // struct stat.st_mode is UInt32
+internal func &(left: UInt32, right: mode_t) -> mode_t {
+    return mode_t(left) & right
+}
+#endif
+
+import CoreFoundation
+
+extension FileManager {
+    internal func _mountedVolumeURLs(includingResourceValuesForKeys propertyKeys: [URLResourceKey]?, options: VolumeEnumerationOptions = []) -> [URL]? {
+        var urls: [URL] = []
+
+#if os(Linux)
+        guard let procMounts = try? String(contentsOfFile: "/proc/mounts", encoding: .utf8) else {
+            return nil
+        }
+        urls = []
+        for line in procMounts.components(separatedBy: "\n") {
+            let mountPoint = line.components(separatedBy: " ")
+            if mountPoint.count > 2 {
+                urls.append(URL(fileURLWithPath: mountPoint[1], isDirectory: true))
+            }
+        }
+#elseif canImport(Darwin)
+        func mountPoints(_ statBufs: UnsafePointer<statfs>, _ fsCount: Int) -> [URL] {
+            var urls: [URL] = []
+
+            for fsIndex in 0..<fsCount {
+                var fs = statBufs.advanced(by: fsIndex).pointee
+
+                if options.contains(.skipHiddenVolumes) && fs.f_flags & UInt32(MNT_DONTBROWSE) != 0 {
+                    continue
+                }
+
+                let mountPoint = withUnsafePointer(to: &fs.f_mntonname.0) { (ptr: UnsafePointer<Int8>) -> String in
+                    return string(withFileSystemRepresentation: ptr, length: strlen(ptr))
+                }
+                urls.append(URL(fileURLWithPath: mountPoint, isDirectory: true))
+            }
+            return urls
+        }
+
+        if #available(OSX 10.13, *) {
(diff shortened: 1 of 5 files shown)


spevans:
Might be worth moving `internal class NSURLDirectoryEnumerator : DirectoryEnumerator` to the per OS files as well as there looks to be very little shared there.

spevans:
@swift-ci test

Foundation/FileManager+POSIX.swift
```diff
@@ -0,0 +1,778 @@
...
+                return nil
+            }
+            urls = mountPoints(statBuf, Int(fsCount))
+        }
+#else
+#error("Requires a platform-specific implementation")
+#endif
+        return urls
+    }
+ internal func _urls(for directory: SearchPathDirectory, in domainMask: SearchPathDomainMask) -> [URL] {
```
spevans: I think the per-OS helper functions should be `private` instead of `internal` where possible`. These are probably a holdover from when they were first written but may as well fix them now.

Foundation/FileManager+POSIX.swift
```diff
@@ -0,0 +1,778 @@
...
+            if useDarwinPaths {
+                urls.append(contentsOf: darwinURLs(for: directory, in: domain))
+            } else {
+                urls.append(contentsOf: xdgURLs(for: directory, in: domain))
+            }
+        }
+
+        return urls
+    }
+    internal func _createDirectory(atPath path: String, withIntermediateDirectories createIntermediates: Bool, attributes: [FileAttributeKey : Any]? = [:]) throws {
```
spevans: Add a separating blank line between functions

Foundation/FileManager+POSIX.swift
```diff
@@ -0,0 +1,778 @@
...
+        return result
+    }
+
+ internal func _createSymbolicLink(atPath path: String, withDestinationPath destPath: String) throws {
+     try _fileSystemRepresentation(withPath: path, andPath: destPath, {
+         guard symlink($1, $0) == 0 else {
+             throw _NSErrorWithErrno(errno, reading: false, path: path)
+         }
+     })
+ }
```
spevans: This needs indentation.

Foundation/FileManager+POSIX.swift
```diff
@@ -0,0 +1,778 @@
...
+            }
+        }
+    }
+
+    internal func _removeItem(atPath path: String, isURL: Bool, alreadyConfirmed: Bool = false) throws {
+        guard alreadyConfirmed || shouldRemoveItemAtPath(path, isURL: isURL) else {
+            return
+        }
+
+        if rmdir(path) == 0 {
```
spevans: This seems to have lost its `_fileSystemRepresentation` wrapper from #1994. Did you base this split on the latest version from `master`?

Foundation/FileManager+POSIX.swift
jmittert: I believe most of these need to be internal instead of private since they are used in a file they aren't defined in.

jmittert:
- Did some formatting
- Fixed the missing fsr, After going through, I pretty sure that it was the only one that got dropped...
- Also split out the iterators

spevans:
@swift-ci test

Foundation/FileManager+POSIX.swift
spevans: Ah yeah, good point.

spevans:
I think the following can be moved into `FileManager+POSIX.swift` as well

```
darwinPathURLs
darwinURLS
xdgHomeDirectory
xdgURLs
```

and wrapped inside a `#if canImport(Darwin) / #else / #endif` block

jmittert:
> I think the following can be moved into `FileManager+POSIX.swift` as well
> 
> ```
> darwinPathURLs
> darwinURLS
> xdgHomeDirectory
> xdgURLs
> ```
> and wrapped inside a `#if canImport(Darwin) / #else / #endif` block

Moved, though not in an `#if` since they're used in places not guarded by an `#if`

millenomi:
@swift-ci please test

compnerd:
@swift-ci please test

millenomi:
@jmittert I'm okay taking this if you fix the conflicts with the things that are landing.

jmittert:
Alright, rebased!

spevans:
@swift-ci test

compnerd:
@swift-ci please test

jmittert:
These both failed on the Swift test suite
```
11:11:25 ********************
11:11:25 Failing Tests (2):
11:11:25     Swift(macosx-x86_64) :: IRGen/autorelease_optimized_armv7.sil
11:11:25     Swift(macosx-x86_64) :: IRGen/autorelease_optimized_aarch64.sil
```
These don't somehow rely on foundation at all do they?

spevans:
@jmittert The tests have been failing for most of the PRs for unrelated reasons

compnerd:
@swift-ci please test

PR #2016, opened 2019-03-16
Windows: Implement attributesOfItem with _lstat

jmittert:
Requires https://github.com/apple/swift-corelibs-foundation/pull/2015

PR #2059, opened 2019-04-03
FileManager: support .type & .posixPermissions on Windows

GunGraveKoga:


GunGraveKoga:
#2016 already fix this

PR #2065, opened 2019-04-05
[SR-464] _cfTypeID property's access modifier changed to internal as intended

karthikkeyan:
Bug Link: https://bugs.swift.org/browse/SR-464


millenomi:
@swift-ci please test

PR #2066, opened 2019-04-07
SR-10277: FileManger: Handle invalid filenames including "" more gracefully

spevans:
- Add __fileSystemRepresentation(withPath:) which is a throwing version
  to allow validating filenames and throw if invalid or cannot be converted
  to FSR.

- Update _fileSystemRepresentation<ResultType>(withPath:_:) and
  _fileSystemRepresentation<ResultType>(withPath:andPath:_:) to throw,
  and update callers to catch or propagate the error as required.

- .attributesOfFileSystem: Call statfs()/statvfs() using the FSR.

- NSString.deletingLastPathComponent: Fix crash if string is empty.

- Add tests for calling FileManager methods with empty filename and
  handle the response more gracefully by throwing an exception.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -656,25 +656,25 @@ open class FileManager : NSObject {
         This method replaces changeFileAttributes:atPath:.
      */
     open func setAttributes(_ attributes: [FileAttributeKey : Any], ofItemAtPath path: String) throws {
-        for attribute in attributes.keys {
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test

millenomi:
@swift-ci please test

spevans:
@swift-ci test

spevans:
@swift-ci test

PR #2067, opened 2019-04-07
SR-10333: Calendar.current.dateComponents([.weekday], from: Date()).weekday segfaults

spevans:
- Override _NSCopyOnWriteCalendar._cfObject to return the backingCalendar created
  from CFCalendarCopyCurrent().

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -1242,7 +1242,11 @@ internal class _NSCopyOnWriteCalendar: NSCalendar {
     private let lock = NSLock()
     private var needsLocking_isMutated: Bool
     private var needsLocking_backingCalendar: NSCalendar
-    
+    override var _cfObject: CFCalendar {
+        copyBackingCalendarIfNeededWithMutation { _ in }
+        return self.backingCalendar._cfObject
+    }
+
     var backingCalendar: NSCalendar {
         lock.lock()
         let it = needsLocking_backingCalendar
--- a/TestFoundation/TestCalendar.swift
+++ b/TestFoundation/TestCalendar.swift
@@ -16,8 +16,7 @@ class TestCalendar: XCTestCase {
             ("test_gettingDatesOnHebrewCalendar", test_gettingDatesOnHebrewCalendar ),
             ("test_gettingDatesOnChineseCalendar", test_gettingDatesOnChineseCalendar),
             ("test_gettingDatesOnISO8601Calendar", test_gettingDatesOnISO8601Calendar),
-            ("test_gettingDatesOnPersianCalendar",
-                test_gettingDatesOnPersianCalendar),
+            ("test_gettingDatesOnPersianCalendar", test_gettingDatesOnPersianCalendar),
             ("test_copy",test_copy),
             ("test_addingDates", test_addingDates),
             ("test_datesNotOnWeekend", test_datesNotOnWeekend),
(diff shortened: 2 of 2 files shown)


spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
cc @millenomi @weissi 

Not sure if this is the correct fix or even complete but it seemed to make sense.

weissi:
Wow thanks Simon! As you say, looks like it makes sense, deferring to @millenomi for a proper review :)

spevans:
Hmm.. This was actually removed in https://github.com/apple/swift-corelibs-foundation/pull/1755/commits/662f631444922c70c2ecbf873c2fb42fbef7560e, I wonder if that was a mistake or there was a good reason for it.

spevans:
@swift-ci test

millenomi:
This isn't consistent with the semantics of Darwin's copy on write. If you get a COW calendar and you pass it to CF calls that alter the calendar, the backing calendar must _not_ be altered if it hasn't been copied already.

We should treat `._cfObject` as a copying operation here if we take this patch.

I also have an additional objection here: bridging should work with this object without need to alter `._cfObject`. What is the underlying bug here exactly?

millenomi:
i.e.: if I pass NSCOW to CF functions, `._cfObject` should be `self`, and they should bridge into `_CFSwiftCalendar…`, and they should invoke NSCOW's methods.

spevans:
What would be the correct way to retrieve `needsLocking_backingCalendar`? It is protected by an `NSLock()` so I dont see how just an `unsafeBitCast()` would be correct?

Note the error is coming from the call:
```
return _CFCalendarDecomposeAbsoluteTimeV(_cfObject, date.timeIntervalSinceReferenceDate, compDesc, vecBuffer.baseAddress!, Int32(compDesc.count - 1))
```

where `_cfObject` ends up being a cast of an empty `NSCalendar` otherwise and ultimately `_locale` is set to `nil` which causes the bad access.

millenomi:
Here's what's happening:

1. `return _CFCalendarDecomposeAbsoluteTimeV(_cfObject /* (1) this call needs to execute */, date.timeIntervalSinceReferenceDate, compDesc, vecBuffer.baseAddress!, Int32(compDesc.count - 1))`

2. `NSCalendar._cfCalendar` is invoked with self == the NSCOW calendar. It returns `return unsafeBitCast(self, to: CFCalendar.self)` (this is correct)

3. In a perfect world, this would bridge to an appropriate Swift API. But that call isn't bridging; which is somewhat correct — there is no method for it to bridge to. :(

So this needs to perhaps be figured out longer-term, but: your patch is correct _if and only if_ you add a call like this:

```
var _cfObject: … {
     copyBackingCalendarIfNeededWithMutation { _ in }
     return self.backingCalendar._cfObject
}
```

so that any mutation method does not change the original calendar.

spevans:
Ive added the copy, but just thinking when calling not-mutating `_CFCalendar...()` functions inside `NSCalendar` could we add an extra  `private var _cfObjectRO...` which doesn't do the copy to use for those specific calls?

spevans:
@swift-ci test

spevans:
@swift-ci test

millenomi:
@spevans Yeah, that works.

spevans:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #2068, opened 2019-04-08
SR-10157: XMLParser shouldn't swap the parts of the element name.

YOCKOW:
XMLParser swaps the prefix with the local name. This behavior is apparently a bug. This PR fixes it.

Resolves [SR-10157](https://bugs.swift.org/browse/SR-10157).


spevans:
@swift-ci test

PR #2069, opened 2019-04-08
SR-9439: Calendar.dateComponetns: Enable .nanoseconds

spevans:
- Nanoseconds were originally disable in #1658 to fix SR-7011,
  however the CF/NSCalendar rewrite in #1755 fixed them to
  work correctly so they can now be re-enabled.


spevans:
@swift-ci test

PR #2070, opened 2019-04-08
Locale: support building with older ICU

compnerd:
The ICU header was not present on the older versions of ICU, but is sufficient
to build against.  Guard the inclusion.

### DIFF
--- a/CoreFoundation/Locale.subproj/CFLocale.c
+++ b/CoreFoundation/Locale.subproj/CFLocale.c
@@ -32,7 +32,9 @@
 #include <unicode/putil.h>          // ICU low-level utilities
 #include <unicode/umsg.h>           // ICU message formatting
 #include <unicode/ucol.h>
+#if __has_include(<unicode/unumsys.h>)
 #include <unicode/unumsys.h>        // ICU numbering systems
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @mili

compnerd:
@swift-ci please test

compnerd:
CC: @millenomi

millenomi:
Why not use the bundled Swift ICU?

compnerd:
@millenomi - its just much more convenient to do this for quickly testing changes.  I'm not convinced that the include is needed at all.

spevans:
I don't think we really want to build against older versions, in fact I'd rather add a guard against using them if possible. I didnt have a PR for that but I need to update the lldb CI build to add a `--libicu` before I can test merge it.

compnerd:
For build tests, it is very helpful to avoid doing a full build of ICU.

millenomi:
I really would like to avoid maintaining a test configuration.

Can you locally update to ICU 61+ that you can then link without needing `--libicu`?

compnerd:
Its just not available on CentOS :-(.

PR #2073, opened 2019-04-09
Parity: Bundle.classNamed(_:) and Bundle.principalClass.

millenomi:
`.classNamed(_:)` can now be used, though with a limitation noted in the comments. `.principalClass` is now implemented in terms of `.classNamed(_:)`.

Fixes [SR-10362](https://bugs.swift.org/browse/SR-10362).

### DIFF
--- a/Foundation/Bundle.swift
+++ b/Foundation/Bundle.swift
@@ -292,8 +292,18 @@ open class Bundle: NSObject {
         }
     }
     
-    open func classNamed(_ className: String) -> AnyClass? { NSUnimplemented() }
-    open var principalClass: AnyClass? { NSUnimplemented() }
+    open func classNamed(_ className: String) -> AnyClass? {
+        // FIXME: This will return a class that may not be associated with the receiver. https://bugs.swift.org/browse/SR-10347.
+        guard isLoaded || load() else { return nil }
+        return NSClassFromString(className)
+    }
+    
+    open var principalClass: AnyClass? {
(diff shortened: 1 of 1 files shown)


millenomi:
@swift-ci please test

Foundation/Bundle.swift
```diff
@@ -292,8 +292,18 @@ open class Bundle: NSObject {
...
-    open var principalClass: AnyClass? { NSUnimplemented() }
+    open func classNamed(_ className: String) -> AnyClass? {
+        // FIXME: This will return a class that may not be associated with the receiver. https://bugs.swift.org/browse/SR-10347.
+        guard isLoaded || load() else { return nil }
+        return NSClassFromString(className)
+    }
+    
+    open var principalClass: AnyClass? {
+        // NB: Cross-platform Swift doesn't have a notion of 'the first class in the ObjC segment' that ObjC platforms have. For swift-corelibs-foundation, if a bundle doesn't have a principal class named, the principal class is nil.
+        guard let name = infoDictionary?["NSPrincipalClass"] as? String else { return nil }
```
karthikkeyan: > If the principal class is not specified in the information dictionary, the bundle identifies the first class loaded as the principal class. When several classes are linked into a dynamically loadable file, the default principal class is the first one listed on the `ld` command line.

Reading the documentation looks like the principle-class-not-found case is not handle here.

millenomi: That is the case the comment notes, and the comment notes why we don't handle it.

karthikkeyan: Thanks for confirming.

PR #2075, opened 2019-04-10
build: fix installation of files on Windows

compnerd:
This fixes the installation for the static build on Linux as well as the
Windows installation to ensure that the import library is also
installed and that the DLL is located correctly.


compnerd:
@swift-ci please test

PR #2077, opened 2019-04-10
NSSet: Add missing init(set: NSSet) and .customMirror

spevans:
- Remove unnecessary `extension` blocks that break up the main
  class.

- Group the init() methods together.


spevans:
@swift-ci test

spevans:
cc @millenomi

PR #2078, opened 2019-04-10
TestFoundation: use `nil` for the allocator

compnerd:
Rather than explicitly requesting the default allocator or the default system
allocator, use `nil` which will result in the use of the default system
allocator.  This avoids the need for use of the internal symbols of
CoreFoundation which are not exposed through Foundation.


compnerd:
@swift-ci please test

compnerd:

compnerd:
CC: @millenomi

PR #2080, opened 2019-04-10
Parity: TimeZone.local, TimeZone.timeZoneDataVersion

millenomi:
Add `__NSLocalTimeZone`; the name is chosen so as to be `NSCoding`-compatible with Darwin. Make sure `NSTimeZone.local` and `TimeZone.autoupdatingCurrent` return it.

Implement `.timeZoneDataVersion` the same way as it is in Darwin.

Re-enable the setter for `.abbreviationDictionary`. We'll fix its bugs rather than crashing.

Fixes https://bugs.swift.org/browse/SR-10348, https://bugs.swift.org/browse/SR-10349

### DIFF
--- a/Foundation/NSTimeZone.swift
+++ b/Foundation/NSTimeZone.swift
@@ -178,9 +178,6 @@ open class NSTimeZone : NSObject, NSCopying, NSSecureCoding, NSCoding {
         }
         return Date(timeIntervalSinceReferenceDate: CFTimeZoneGetNextDaylightSavingTimeTransition(_cfObject, aDate.timeIntervalSinceReferenceDate))
     }
-}
-
-extension NSTimeZone {
 
     open class var system: TimeZone {
         return CFTimeZoneCopySystem()._swiftObject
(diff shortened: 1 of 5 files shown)


millenomi:
@swift-ci please test

Foundation/NSTimeZone.swift
```diff
@@ -212,12 +211,17 @@ extension NSTimeZone {
...
-            // CFTimeZoneSetAbbreviationDictionary(newValue._cfObject)
-            NSUnimplemented()
+            CFTimeZoneSetAbbreviationDictionary(newValue._cfObject)
         }
     }
 
-    open class var timeZoneDataVersion: String { NSUnimplemented() }
+    open class var timeZoneDataVersion: String {
+        #if os(Windows)
+            return "" // We do not source timezone data from ICU. The empty string is what Darwin would return if ICU isn't able to give a time zone data version.
```
compnerd: Hmm, we should be able to do this on Windows I think.  We do have ICU available and will link against it.

millenomi: Ah, okay. Do you want me to just return the ICU version?

compnerd: Yes please!

millenomi:
@swift-ci please test

compnerd:
@swift-ci please test

PR #2083, opened 2019-04-10
Extended Bundle layout testing

pvieito:
- Added Windows support
- Added both executable and library main executable layouts
- Added test for reverse bundle lookup (required for all implicitly loaded bundles, eg. Bundle.main)

### DIFF
--- a/Foundation/Bundle.swift
+++ b/Foundation/Bundle.swift
@@ -11,13 +11,21 @@ import CoreFoundation
 
 open class Bundle: NSObject {
     private var _bundle : CFBundle!
-
-    internal static var _supportsFHSStyle: Bool {
-    #if DEPLOYMENT_RUNTIME_OBJC
-      return false
-    #else
-      return _CFBundleSupportsFHSBundles()
-    #endif
+    
+    public static var _supportsFHSBundles: Bool {
+        #if DEPLOYMENT_RUNTIME_OBJC
+        return false
+        #else
+        return _CFBundleSupportsFHSBundles()
+        #endif
+    }
+    
+    public static var _supportsFreestandingBundles: Bool {
+        #if DEPLOYMENT_RUNTIME_OBJC
+        return false
+        #else
+        return _CFBundleSupportsFreestandingBundles()
+        #endif
     }
 
     private static var _mainBundle : Bundle = {
@@ -104,6 +112,14 @@ open class Bundle: NSObject {
         }
         
         _bundle = result
+    }
+    
+    public convenience init?(_executableURL: URL) {
+        guard let bundleURL = _CFBundleCopyBundleURLForExecutableURL(_executableURL._cfObject)?.takeRetainedValue() else {
+            return nil
+        }
+        
+        self.init(url: bundleURL._swiftObject)
     }
     
     override open var description: String {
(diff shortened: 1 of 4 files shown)


pvieito:
@compnerd @millenomi

TestFoundation/TestBundle.swift
```diff
@@ -36,13 +36,70 @@ internal func xdgTestHelperURL() -> URL {
...
+                #else
+                return "so"
+                #endif
+            case .executable:
+                #if os(Windows)
+                return "exe"
+                #else
+                return ""
+                #endif
+            }
```
compnerd: I wonder if we want to have some constants for this in `FileManager` as we keep wanting to do this.

millenomi: Yea :\ probably a SPI picking up that information from CF (like: `Bundle._platformLibraryPrefix/Suffix`, `Bundle._platformExecutablePrefix/Suffix` or similar).

millenomi:
I'm not blocking this on the prefix/suffix SPI.

millenomi:

millenomi:
@pvieito This is good to merge after resolving conflicts.

pvieito:
@millenomi I have merged @compnerd changes and added SPI on CF to check if Freestanding Bundles are supported.

spevans:

pvieito:
@spevans I think Jenkins has ignored the build request, could you request another one? Thanks!

spevans:

pvieito:
@compnerd this [Swift CI failure](https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/2411/) is resolved?

spevans:

compnerd:
@swift-ci please test

compnerd:

pvieito:
@millenomi Is the CI broken? I don't understand the error: https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/2451/console

spevans:
@pvieito CI has been breaking all day for issues unrelated to the s-c-f PRs being tested

pvieito:
@spevans It seems to be working now, could you request a test-and-merge? Thanks!

spevans:

Foundation/Bundle.swift
```diff
@@ -106,6 +114,14 @@ open class Bundle: NSObject {
         _bundle = result
     }
     
+    public convenience init?(_executableURL: URL) {
```
spevans: @pvieito @millenomi Should this method be public? also `_supportsFHSBundles` and `_supportsFreestandingBundle` above? If its just for testing shouldn't they be `internal`?

pvieito: @spevans If we mark it `internal` we would have to guard the tests with `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT`, which means that they would not be executed on the Linux CI (and I would like them to be executed on the Linux CI, as I can only test them locally on macOS).

spevans: I thought `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT` works for Linux CI because it uses the `debug-foundation` flag. @millenomi Is that not the case?

PR #2084, opened 2019-04-10
test: use bridging rather than CoreFoundation

compnerd:
Bridge the return CFString object to NSObject via the private
`_nsObject` and use the `length` property to get the length rather than
using the `CFStringGetLength` method in the tests.  This reduces the use
of CoreFoundation private interfaces in the tests, which is needed for
Windows.

### DIFF
--- a/TestFoundation/TestTimeZone.swift
+++ b/TestFoundation/TestTimeZone.swift
@@ -198,13 +198,12 @@ class TestTimeZone: XCTestCase {
         // Ensure that the system time zone creates names the same way as creating them with an identifier.
         // If it isn't the same, bugs in DateFormat can result, but in this specific case, the bad length
         // is only visible to CoreFoundation APIs, and the Swift versions hide it, making it hard to detect.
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

millenomi:
In tests, you can just use `as NSString`.

compnerd:
Oh, well, in that case, I can probably just make this unconditional then!  Thanks!

compnerd:
@swift-ci please test

compnerd:
@swift-ci please testx

spevans:
@swift-ci test linux

compnerd:
I wonder if the forced cast as failing?  That doesn't make sense to me though

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

PR #2085, opened 2019-04-11
appendingPathComponent(:isDirectory:) should account for isDirectory

Coeur:
Same as https://github.com/apple/swift/pull/23926.

When `appendingPathComponent(:isDirectory:)` is falling back from an **NSURL**.appendingPathComponent() to an **NSString**.appendingPathComponent(), it's forgetting to take into account the `isDirectory` parameter.


millenomi:

PR #2086, opened 2019-04-11
ignoring AppCode artifacts

Coeur:
AppCode is a popular IDE from https://www.jetbrains.com/objc/ and it uses the `.idea` folder like many other JetBrains products.


compnerd:

PR #2089, opened 2019-04-11
NSOrderedSet: Implement more functionality

spevans:
- Add copy() and mutableCopy().

- Add .set and .array to return immumtable views of the current
  contents.

- Implement index(of:inSortedRange:options:usingComparator cmp:)
  by calling NSArray()'s version on ._orderedStorage.

- Implement description(withLocale:), description(withLocale:indent:),
  .description and .customMirror.

- NSCoding still needs to be implemented.

### DIFF
--- a/Foundation/NSOrderedSet.swift
+++ b/Foundation/NSOrderedSet.swift
@@ -8,16 +8,21 @@
 //
 
 /****************       Immutable Ordered Set   ****************/
-open class NSOrderedSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, ExpressibleByArrayLiteral {
-    internal var _storage: NSSet
-    internal var _orderedStorage: NSArray
+open class NSOrderedSet: NSObject, NSCopying, NSMutableCopying, NSSecureCoding, ExpressibleByArrayLiteral {
+
+    fileprivate var _storage: NSSet
+    fileprivate var _orderedStorage: NSArray
     
     open override func copy() -> Any {
         return copy(with: nil)
     }
     
     open func copy(with zone: NSZone? = nil) -> Any {
-        NSUnimplemented()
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

spevans:
Note: Need to add some tests.

millenomi:
Fixes https://bugs.swift.org/browse/SR-10405, https://bugs.swift.org/browse/SR-10406, https://bugs.swift.org/browse/SR-10407, https://bugs.swift.org/browse/SR-10408

millenomi:
Can you use `storage._swiftObject` instead?

spevans:
I looked at how `NSArray` and `NSSet` implemented `copy` and `mutableCopy` and didn't see the use of `_swiftObject` there. Is it needed even though the types of the `Set` and `Array` are Swift not CF Types?

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

millenomi:
@swift-ci please test

millenomi:
The idea here is that we're hitting a compiler bug around casting, which is often problematic while developing s-c-f because s-c-f is part of the implementation of casting itself (and also has to coexist with the bridging setup on Darwin.) The guideline going forward is to avoid bridging casts and use `_swiftObject`, `_cfObject` and `_nsObject` where they exist, and `__SwiftValue.store(…)/.fetch(…)` otherwise if all you have is an `Any` or `AnyObject` (which will do the right thing wrt interacting with bridged values coming from the ObjC Foundation onDarwin). This only applies to s-c-f's codebase; outside, including in TestFoundation, it is okay to use bridging as usual.

spevans:
@swift-ci  test

spevans:
@swift-ci test

PR #2090, opened 2019-04-11
NSString: implement `NSString.StringTransform`

compnerd:
Foundation provides a string enumeration as `StringTransform` which
affords the developer to access the ICU string transformations.  Add
this missing enumeration.


compnerd:
@swift-ci please test

PR #2092, opened 2019-04-11
TestFoundation: replace `sleep` with `Thread.sleep(forTimeInterval:)`

compnerd:
Windows does not have a `sleep` function but does have `Sleep`.  Rather
than conditionalise the path, use `Thread.sleep(forTimeInterval:)` which
makes the path uniform for all the targets.


compnerd:
@swift-ci please test

spevans:
@swift-ci test linux

compnerd:
apple/swift#23974

PR #2097, opened 2019-04-12
Parity: NSKeyedArchiver: Consider response to setting an output format other than .xml or .binary

millenomi:
`NSUnimplemented()` is a marker for yet-to-be-implemented functionality. We do not support `.openStep` as a format at all; move this to `fatalError()` instead, since setting `.openStep` is a programmer error on Darwin.

https://bugs.swift.org/browse/SR-10416


millenomi:
@swift-ci please test

compnerd:
@swift-ci please test

spevans:
@swift-ci test macos

PR #2098, opened 2019-04-12
build: adjust flags for TestFoundation on Windows

compnerd:
We need to link against the WinSock library on Windows since we use the
WinSock library for the HTTPServer.  Silence some warnings on Windows
for the linker to make errors easier to spot.


compnerd:
@swift-ci please smoke test and merge

compnerd:

PR #2108, opened 2019-04-14
[URLSession] Fix handling of compressed responses.

drodriguez:
When the server answers with a compressed response (with
Content-Encoding header), cURL is configured to automatically decompress
the response. However, there was some inconsistencies with the
Foundation implementation in Darwin platforms, and some bugs in the
handling of compressed responses.

- In Darwin Foundation, the expected number of bytes is -1 for
  compressed responses. This is because Content-Lenght will report the
  length of the compressed content, but the total bytes written reports
  uncompressed lengths. The code is changed to set the expected number
  of bytes to -1 if the response includes a Content-Encoding header
  different to "identity".
- Since the expected number of bytes is unknown, the data received
  callback cannot check for the total bytes received to be equal to
  close the file handler and provide the temporary file URL to the upper
  level. That responsability has been moved into the complete task
  callback, where it was already happening for tasks with completion
  blocks.
- Added two tests (one for data tasks, one for download tasks) with
  gzipped data.
- Since gzipped data cannot be represented by UTF-8 strings, the
  HTTPServer code has to be modified to allow providing raw data as part
  of the HTTP response. There's a lot of changes so the body is raw
  data, and the previously provided strings are transformed into data
  using UTF-8 instead.
- There was a small bug in the HTTPServer code where the setUp will wait
  for a flag to be different of -2 to indicate the server is ready.
  However, the flag should be checked against -1, which is the initial
  state, while -2 is the final state. I found this when the server port
  that my test wanted to use was uninitialized, because the server is
  started asynchronously in another queue, and the value wasn't valid
  yet.

### DIFF
--- a/Foundation/URLSession/NativeProtocol.swift
+++ b/Foundation/URLSession/NativeProtocol.swift
@@ -136,10 +136,6 @@ internal class _NativeProtocol: URLProtocol, _EasyHandleDelegate {
                 downloadDelegate.urlSession(s, downloadTask: task, didWriteData: Int64(data.count), totalBytesWritten: task.countOfBytesReceived,
                                             totalBytesExpectedToWrite: task.countOfBytesExpectedToReceive)
             }
-            if task.countOfBytesExpectedToReceive == task.countOfBytesReceived {
-                fileHandle.closeFile()
-                self.properties[.temporaryFileURL] = self.tempFileURL
-            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


spevans:
@swift-ci test

drodriguez:
Changed an incorrect test assert message.

spevans:
@swift-ci test

pushkarnk:
@swift-ci test

drodriguez:
Fixed the indentation. I hope I fixed every case, but don’t hesitate to point out if you find any more.

spevans:
@swift-ci test

pushkarnk:
@swift-ci test

pushkarnk:
@drodriguez Thanks! Looks good to me now.

spevans:
@swift-ci test and merge

pushkarnk:
@swift-ci test and merge

PR #2110, opened 2019-04-14
TestFoundation: correct `WSARecv` invocation

compnerd:
The `lpFlags` parameter is not optional, ensure it is set.


compnerd:

PR #2112, opened 2019-04-14
[android] Provide correct temporary directory path.

drodriguez:
Android Bionic uses /data/local/tmp as the temporary writable directory.
TMPDIR is rarely set and /tmp doesn't exist.

There's some usage of /tmp/ in the tests that might be wrong and might
need to be fixed in the future, but I have to check one by one and right
now I don't have a device to test on.

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -57,7 +57,13 @@ public func NSTemporaryDirectory() -> String {
             return tmpdir
         }
     }
+#if os(Android)
+    // Bionic uses /data/local/tmp/ as temporary directory. TMPDIR is rarely
+    // defined.
+    return "/data/local/tmp/"
+#else
     return "/tmp/"
+#endif
 #endif
 }
 


Foundation/NSPathUtilities.swift
```diff
@@ -57,7 +57,13 @@ public func NSTemporaryDirectory() -> String {
...
     }
-    return "/tmp/"
+    #if os(Android)
+      // Bionic uses /data/local/tmp/ as temporary directory. TMPDIR is rarely
+      // defined.
+      return "/data/local/tmp/"
+    #else
+      return "/tmp/"
+    #endif
 #endif
```
spevans: Could you fix the indentation please, the `#if/#else/#endif` should be on the left edge, thanks

compnerd: It would be preferable, but, Foundation is so horribly inconsistent with that, that I don't think that any formatting would make this correct :-p.

drodriguez: I will fix it, because I realized that I'm using 2 spaces, instead of 4.

spevans:
@swift-ci test

PR #2114, opened 2019-04-14
NSPathUtilities: Implement NSString.fileSystemRepresentation

spevans:
Fixes https://bugs.swift.org/browse/SR-10364

### DIFF
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -537,8 +537,8 @@ extension NSString {
         return path + "/"
     }
     
-    public var fileSystemRepresentation : UnsafePointer<Int8> {
-        NSUnimplemented()
+    public var fileSystemRepresentation: UnsafePointer<Int8> {
+        return FileManager.default.fileSystemRepresentation(withPath: self._swiftObject)
     }
     
     public func getFileSystemRepresentation(_ cname: UnsafeMutablePointer<Int8>, maxLength max: Int) -> Bool {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

Foundation/NSPathUtilities.swift
```diff
@@ -537,8 +537,8 @@ extension NSString {
         return path + "/"
     }
     
-    public var fileSystemRepresentation : UnsafePointer<Int8> {
-        NSUnimplemented()
+    public var fileSystemRepresentation: UnsafePointer<Int8> {
+        return FileManager.default.fileSystemRepresentation(withPath: self._swiftObject)
```
compnerd: Isn't this introducing a memory leak?  @millenomi

spevans: Indeed it is, its the same as `FileManager.fileSystemRepresentation(withPath:)` and after use `.deallocate()` needs to be on Linux.

On macOS this is normally an autoreleasing pointer. Hopefully at some point a new method will be added which uses a closure.

compnerd: Can we at least have some way to indicate that this introduces a memory leak?

millenomi: While I'm okay taking a patch that brings this to parity with `.fSR(wP:)`, `.fSR(wP:)` is not at parity with Darwin, and I'll be looking at introducing a block-taking experimental version.

PR #2115, opened 2019-04-14
Support Swift 4.2 Hashable

colemancda:
Updates implementation of `Hashable` in Foundation types to properly support Swift 4.2 and 5.

### DIFF
--- a/Foundation/Date.swift
+++ b/Foundation/Date.swift
@@ -137,9 +137,15 @@ public struct Date : ReferenceConvertible, Comparable, Equatable {
      */
     public static let distantPast = Date(timeIntervalSinceReferenceDate: -63114076800.0)
     
+    #if swift(>=4.2)
+    public func hash(into hasher: inout Hasher) {
+        _time.hash(into: &hasher)
+    }
+    #else
     public var hashValue: Int {
         return Int(bitPattern: __CFHashDouble(_time))
     }
+    #endif
     
     /// Compare two `Date` values.
     public func compare(_ other: Date) -> ComparisonResult {
--- a/Foundation/Measurement.swift
+++ b/Foundation/Measurement.swift
@@ -35,10 +35,17 @@ public struct Measurement<UnitType : Unit> : ReferenceConvertible, Comparable, E
         self.value = value
         self.unit = unit
     }
-
+    
+    #if swift(>=4.2)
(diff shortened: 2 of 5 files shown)


colemancda:
@spevans Can you test this please?

Foundation/Date.swift
```diff
@@ -137,9 +137,15 @@ public struct Date : ReferenceConvertible, Comparable, Equatable {
      */
     public static let distantPast = Date(timeIntervalSinceReferenceDate: -63114076800.0)
     
+    #if swift(>=4.2)
```
spevans: The `swift` version check should not be needed as there is a branch per version.

Foundation/Measurement.swift
```diff
@@ -35,10 +35,17 @@ public struct Measurement<UnitType : Unit> : ReferenceConvertible, Comparable, E
         self.value = value
         self.unit = unit
     }
-
+    
+    #if swift(>=4.2)
+    public func hash(into hasher: inout Hasher) {
+        unit.hash(into: &hasher)
+        value.hash(into: &hasher)
+    }
```
spevans: I dont see this method on Darwin's Foundation (unless its in the latest beta?)

spevans:
Could you add a testcase just calling the method for each of the types are modifying please

Foundation/Measurement.swift
colemancda: If you don't implement `func hash(into hasher: inout Hasher)` it uses `hashValue` as a fallback implementation, which on Darwin is just the `NSObject` hash.

millenomi:
cc @lorentey who was already working on this, or similar.

lorentey:
I suggest to hold this; this is somewhat trickier than it looks, and it has to be synced with the Foundation overlay. In https://github.com/apple/swift/pull/23832, I submitted a draft PR for the overlay parts; I'm preparing a similar PR for corelibs-foundation, too.

lorentey:
Update: My corelibs-foundation PR is up at https://github.com/apple/swift-corelibs-foundation/pull/2118.

millenomi:
@colemancda thank you for the contribution — I'm gonna close this because it has been superseded by #2118.

colemancda:
Ok, if this is done better somewhere else, that's great.

PR #2116, opened 2019-04-14
Disable toll-free bridging for CFRunLoopTimer

tokyovigilante:
Foundation.Timer is not currently implemented with a compatible memory layout
so attempt to bridge these classes will fail. Disabling bridging allows
code using CFRunLoopTimer to work until Foundation.Timer is reimplemented.
Fixes https://bugs.swift.org/browse/SR-10465

### DIFF
--- a/Foundation/NSSwiftRuntime.swift
+++ b/Foundation/NSSwiftRuntime.swift
@@ -173,8 +173,8 @@ internal func __CFInitializeSwift() {
     _CFRuntimeBridgeTypeToClass(CFAttributedStringGetTypeID(), unsafeBitCast(NSMutableAttributedString.self, to: UnsafeRawPointer.self))
 //    _CFRuntimeBridgeTypeToClass(CFReadStreamGetTypeID(), unsafeBitCast(InputStream.self, UnsafeRawPointer.self))
 //    _CFRuntimeBridgeTypeToClass(CFWriteStreamGetTypeID(), unsafeBitCast(OutputStream.self, UnsafeRawPointer.self))
-   _CFRuntimeBridgeTypeToClass(CFRunLoopTimerGetTypeID(), unsafeBitCast(Timer.self, to: UnsafeRawPointer.self))
-    
+   //_CFRuntimeBridgeTypeToClass(CFRunLoopTimerGetTypeID(), unsafeBitCast(Timer.self, to: UnsafeRawPointer.self))
+
     __CFSwiftBridge.NSObject.isEqual = _CFSwiftIsEqual
     __CFSwiftBridge.NSObject.hash = _CFSwiftGetHash
     __CFSwiftBridge.NSObject._cfTypeID = _CFSwiftGetTypeID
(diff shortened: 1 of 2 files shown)


CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -4306,7 +4306,7 @@ CFIndex CFRunLoopTimerGetOrder(CFRunLoopTimerRef rlt) {
 
 void CFRunLoopTimerInvalidate(CFRunLoopTimerRef rlt) {	/* DOES CALLOUT */
     CHECK_FOR_FORK();
-    CF_OBJC_FUNCDISPATCHV(CFRunLoopTimerGetTypeID(), void, (NSTimer *)rlt, invalidate);
+    //CF_OBJC_FUNCDISPATCHV(CFRunLoopTimerGetTypeID(), void, (NSTimer *)rlt, invalidate);
```
millenomi: Please do not disable CF_OBJC_… functions. As the name implies, these are for ObjC (and are no-ops in Swift.)

CoreFoundation/RunLoop.subproj/CFRunLoop.h
```diff
@@ -29,7 +29,7 @@ typedef struct CF_BRIDGED_MUTABLE_TYPE(id) __CFRunLoopSource * CFRunLoopSourceRe
 
 typedef struct CF_BRIDGED_MUTABLE_TYPE(id) __CFRunLoopObserver * CFRunLoopObserverRef;
 
-typedef struct CF_BRIDGED_MUTABLE_TYPE(NSTimer) __CFRunLoopTimer * CFRunLoopTimerRef;
+typedef struct /*CF_BRIDGED_MUTABLE_TYPE(id)*/ __CFRunLoopTimer * CFRunLoopTimerRef;
```
millenomi: Please do not disable this. This does not bridge to `NSTimer`; it bridges to `id`, meaning that it allows `CFRunLoopTimer` to be treated as an `AnyObject` (and thus its API be imported without using `Unmanaged<…>`.

millenomi:
In general, changing public API headers is not allowed.

CoreFoundation/RunLoop.subproj/CFRunLoop.c
tokyovigilante: OK great, I didn't realise that. Specifically that one however does appear to be casting `CFRunLoopTimer` to `NSTimer *`, is that ok to keep in?

CoreFoundation/RunLoop.subproj/CFRunLoop.h
tokyovigilante: OK makes sense, I did note CFRunLoop itself etc are also  `CF_BRIDGED_MUTABLE_TYPE` there. Should the type remain `NSTimer` however or be `id` like the rest?

tokyovigilante:
Noted, thanks. Thanks for the feedback!

tokyovigilante:
This one should do it, works on my system without changing the API. Thanks again for the feedback!

millenomi:

millenomi:

PR #2118, opened 2019-04-15
Modernize hashing in Foundation's Swift-only types

lorentey:
This is the corelibs-Foundation counterpart to https://github.com/apple/swift/pull/23832 for the Foundation SDK overlay. See there for a full description and rationale. 

This PR attempts to bring hashing roughly in sync in the two Foundation flavors, although some inconsistencies remain. Like in https://github.com/apple/swift/pull/23832, I'll post self-review notes below to highlight some points.

### DIFF
--- a/Foundation/Boxing.swift
+++ b/Foundation/Boxing.swift
@@ -75,7 +75,9 @@ internal protocol _SwiftNativeFoundationType : class {
     
     func mutableCopy(with zone : NSZone) -> Any
     
+    func hash(into hasher: inout Hasher)
     var hashValue: Int { get }
+
     var description: String { get }
     var debugDescription: String { get }
     
@@ -115,6 +117,10 @@ extension _SwiftNativeFoundationType {
         return _mapUnmanaged { ($0 as NSObject).mutableCopy() }
     }
     
+    func hash(into hasher: inout Hasher) {
+        _mapUnmanaged { hasher.combine($0) }
+    }
+
     var hashValue: Int {
         return _mapUnmanaged { return $0.hashValue }
     }
--- a/Foundation/Calendar.swift
+++ b/Foundation/Calendar.swift
@@ -862,8 +862,9 @@ public struct Calendar : Hashable, Equatable, ReferenceConvertible, _MutableBoxi
     public func hash(into hasher: inout Hasher) {
         // We implement hash ourselves, because we need to make sure autoupdating calendars have the same hash
         if _autoupdating {
-            hasher.combine(1 as Int8)
+            hasher.combine(false)
         } else {
+            hasher.combine(true)
             hasher.combine(_handle.map { $0 })
         }
     }
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -828,11 +828,11 @@ public struct FileAttributeKey : RawRepresentable, Equatable, Hashable {
     public init(rawValue: String) {
         self.rawValue = rawValue
     }
-    
-    public var hashValue: Int {
-        return self.rawValue.hashValue
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(rawValue)
     }
-    
+
     public static func ==(_ lhs: FileAttributeKey, _ rhs: FileAttributeKey) -> Bool {
         return lhs.rawValue == rhs.rawValue
     }
@@ -873,8 +873,8 @@ public struct FileAttributeType : RawRepresentable, Equatable, Hashable {
         self.rawValue = rawValue
     }
 
-    public var hashValue: Int {
-        return self.rawValue.hashValue
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(rawValue)
     }
 
     public static func ==(_ lhs: FileAttributeType, _ rhs: FileAttributeType) -> Bool {
--- a/Foundation/NSCharacterSet.swift
+++ b/Foundation/NSCharacterSet.swift
@@ -46,7 +46,7 @@ open class NSCharacterSet : NSObject, NSCopying, NSMutableCopying, NSCoding {
     }
     
     open override var hash: Int {
-        return Int(bitPattern: CFHash(_cfObject))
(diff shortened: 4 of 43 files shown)


Foundation/Boxing.swift
```diff
@@ -75,7 +75,9 @@ internal protocol _SwiftNativeFoundationType : class {
     
     func mutableCopy(with zone : NSZone) -> Any
     
+    func hash(into hasher: inout Hasher)
```
lorentey: In protocol extensions that forward hashing to some core value, it is generally a good idea to implement both `hash(into:)` and `hashValue`, so that hashing always matches the behavior of the underlying value. (But if we have to choose one, I'd go with just `hash(into:)`.

millenomi:
@swift-ci please test

Foundation/Calendar.swift
```diff
@@ -862,8 +862,9 @@ public struct Calendar : Hashable, Equatable, ReferenceConvertible, _MutableBoxi
     public func hash(into hasher: inout Hasher) {
         // We implement hash ourselves, because we need to make sure autoupdating calendars have the same hash
         if _autoupdating {
-            hasher.combine(1 as Int8)
+            hasher.combine(false)
         } else {
+            hasher.combine(true)
```
lorentey: The original implementation here was ever so slightly wrong -- it had a variable-length hash encoding without discriminators to mark its boundaries, which could (theoretically) lead to collisions when multiple `Calendar` values are fed into a hasher.

Granted, I doubt this would be a problem in practice; however, this change makes `Calendar`'s hash encoding match that of the Foundation overlay, which is nice. I made similar changes for `TimeZone` and `Locale`.

Foundation/FileManager.swift
```diff
@@ -828,11 +828,11 @@ public struct FileAttributeKey : RawRepresentable, Equatable, Hashable {
     public init(rawValue: String) {
         self.rawValue = rawValue
     }
-    
-    public var hashValue: Int {
-        return self.rawValue.hashValue
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(rawValue)
```
lorentey: This `hash(into:)` and the `==` below are identical to the definitions that the compiler would synthesize on its own. It may be worth considering removing them here (as well as all other RawRepresentable enums).

Foundation/NSCharacterSet.swift
```diff
@@ -46,7 +46,7 @@ open class NSCharacterSet : NSObject, NSCopying, NSMutableCopying, NSCoding {
     }
     
     open override var hash: Int {
-        return Int(bitPattern: CFHash(_cfObject))
+        return Int(bitPattern: _CFNonObjCHash(_cfObject))
```
lorentey: This fixes an infinite recursion in `NSCharacterSet.hash`.

Foundation/NSGeometry.swift
```diff
@@ -74,9 +74,10 @@ extension CGPoint: NSSpecialValueCoding {
             return false
         }
     }
-    
-    var hash: Int {
-        return self.x.hashValue &+ self.y.hashValue
+
+    func hash(into hasher: inout Hasher) {
```
lorentey: The Hashable conformances for CGPoint, CGSize and CGRect are inconsistent with Apple platforms. (See https://bugs.swift.org/browse/SR-10476.)

lorentey:
```
15:15:48 
/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift-corelibs-foundation/TestFoundation/main.swift:37:14: error: use of unresolved identifier 'TestDateInterval'
15:15:48     testCase(TestDateInterval.allTests),
15:15:48              ^~~~~~~~~~~~~~~~
15:15:48 Foundation.NSDateInterval:1:12: note: did you mean 'NSDateInterval'?
```

Ah, I missed some CMakeLists.txt additions. 

The macOS failure is different -- perhaps it just needs a clean build.

lorentey:
@swift-ci please clean test

millenomi:
@swiftix please test clean

millenomi:
@swift-ci please test

lorentey:
@swift-ci please test

spevans:
@swift-ci please test

Foundation/FileManager.swift
millenomi: I agree, esp. given that on Darwin it's the compiler that synthesizes those conformances, right?

spevans:
@swift-ci test macos

Foundation/FileManager.swift
lorentey: I looked into it in more detail: on Darwin the hash implementations come from extensions on `_SwiftNewtypeWrapper` that are identical to similar extensions provided on `RawRepresentable`. The stdlib also defines a `==` that’s generic over `RawRepresentable` types; this gets picked up for Equality conformance.

So if we removed these, then the same definitions would be found on `RawRepresentable` instead of compiler synthesis. But the end result is effectively the same; we should do it!

TestFoundation/Utilities.swift
```diff
@@ -134,50 +132,313 @@ extension Optional {
...
     }
 }
+
+/// Test that the elements of `instances` satisfy the semantic
+/// requirements of `Equatable`, using `oracle` to generate equality
+/// expectations from pairs of positions in `instances`.
+///
+/// - Note: `oracle` is also checked for conformance to the
+///   laws.
+public func checkEquatable<Instances: Collection>(
```
millenomi: Are we okay having a duplicate implementation of this between SCF and the stdlib tests, or is there something we can do?

This does not block the patch, but it's a point of concern going forward.

millenomi: cc @lorentey

lorentey: Oh yes! The XCTest-StdlibUnittest dichotomy is super annoying, we should do something about that. Ideally we would have a single unified testing infrastructure.

The high-level behavioral tests should be shared between all platforms, too.

PR #2102, opened 2019-04-12
Parity: NSKeyedArchiver: Secure initializers

millenomi:
- Added missing API: `NSDebugDescriptionErrorKey`
- Added secure initializers and one-shot factory methods to `NSKeyed{A,Una}rchiver`
- Deprecated existing initializers and factory methods

PR #2119, opened 2019-04-15
Parity: NSCoder miscellanea

millenomi:
- Added missing API: `DecodingFailurePolicy.raiseException`. It is now the default like on Darwin.

- Match Darwin behavior while decoding if you decode without raising exceptions.

- `decodeTopLevel…` variants are noted unavailable on swift-corelibs-foundation, with rationale and replacement directions.

Swift doesn’t have exceptions, and it used to define `.setErrorAndReturn` as the defaults. However, until very recently `failWithError(_:)` was unimplemented and trapped with a fatal error, effectively meaning that all coding failures were behaving as if `.raiseException` was set in a Darwin Swift-only app: by crashing.

Since code written to 4.x-5.0 may be relying on this behavior, we make it explicit and consistent with Darwin: there is now a `.raiseException` constant, it’s the default, and it fails with a fatal error in `failWithError(_:)`. The new secure initializers (at https://github.com/apple/swift-corelibs-foundation/pull/2102) set `.setErrorAndReturn` just like on Darwin, and thus won’t raise and instead just thrown an error.

Tests will be added.

### DIFF
--- a/Foundation/NSCoder.swift
+++ b/Foundation/NSCoder.swift
@@ -14,6 +14,7 @@ extension NSCoder {
     /// supports exceptions here, and there may be other approaches supported
     /// in the future, so its included for completeness.
     public enum DecodingFailurePolicy : Int {
+        case raiseException
         case setErrorAndReturn
     }
 }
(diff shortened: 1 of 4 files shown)


millenomi:
@swift-ci please test

millenomi:
Fies https://bugs.swift.org/browse/SR-10413, https://bugs.swift.org/browse/SR-10418

millenomi:

millenomi:

PR #2120, opened 2019-04-15
[android] Fix some parts of Bundle in Android.

drodriguez:
While Android does not follow FHS, and it doesn't seem to have something
similar published, it follows some conventions of Unix. One of them
is that libraries are normally named prefixed with "lib" and suffixed
with ".so" like Linux. Executables are also normally named without any
prefix nor suffix.

The changes in this commit lift up restrictions that were in place for
Android, and allow some of the Bundle tests to pass on Android (and lift
another restriction in TestObjCRuntime).

While bundles might never be a common construct in Android, these
changes at least allow to test some of the scenarios, in case someone
needs this behaviour.

This allows `TestBundle`, `TestNSError`, `TestPropertyListSerialization`, `TestObjCRuntime`, and `TestNSKeyedUnarchiver` to pass on Android devices.

### DIFF
--- a/TestFoundation/TestObjCRuntime.swift
+++ b/TestFoundation/TestObjCRuntime.swift
@@ -43,10 +43,10 @@ class TestObjCRuntime: XCTestCase {
         let name = testBundleName()
         XCTAssertEqual(NSStringFromClass(NSObject.self), "NSObject")
         XCTAssertEqual(NSStringFromClass(SwiftClass.self), "\(name).SwiftClass")
-#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
-        XCTAssertEqual(NSStringFromClass(XCTestCase.self), "XCTest.XCTestCase");
-#else
+#if canImport(SwiftXCTest) && !DEPLOYMENT_RUNTIME_OBJC
         XCTAssertEqual(NSStringFromClass(XCTestCase.self), "SwiftXCTest.XCTestCase");
+#else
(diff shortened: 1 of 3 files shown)


compnerd:
The library/executable naming is not part of the FHS, its part of the Unix heritage.

compnerd:
@swift-ci please test

TestFoundation/TestObjCRuntime.swift
```diff
@@ -43,7 +43,7 @@ class TestObjCRuntime: XCTestCase {
         let name = testBundleName()
         XCTAssertEqual(NSStringFromClass(NSObject.self), "NSObject")
         XCTAssertEqual(NSStringFromClass(SwiftClass.self), "\(name).SwiftClass")
-#if DEPLOYMENT_RUNTIME_OBJC || os(Linux)
+#if DEPLOYMENT_RUNTIME_OBJC || os(Linux) || os(Android)
```
millenomi: This should really just switch to the condition we have in place now for importing `SwiftFoundation` vs `Foundation`:

```
#if canImport(SwiftXCTest) && !DEPLOYMENT_RUNTIME_OBJC
… // use SwiftXCTest
#else
… // use XCTest
#endif
```

drodriguez:
Incorporated the feedback from Lily and reworded the commit/PR message to not imply that libraries or executable naming are related to FHS (from the comments in the code I thought they were).

spevans:
@swift-ci test

compnerd:

PR #2121, opened 2019-04-16
[Windows] Now FileManager.isExecutableFile(atPath:) checks if the file exists and is readable

pvieito:


pvieito:
@compnerd

Foundation/FileManager+Win32.swift
```diff
@@ -497,8 +502,7 @@ extension FileManager {
     }
 
     internal func _isExecutableFile(atPath path: String) -> Bool {
-        // FIXME(compnerd) is there some test that we can perform here?
-        return true
+        return _nonDirectoryFileExists(atPath: path) && _isReadableFile(atPath: path)
```
compnerd: Is there a reason to create the `_nonDirectoryFileExists`?  I think that we can just inline it (unless you have another use that you have planned?).

pvieito: No, mainly for potential future usage and better readability. If you prefer it, I could inline it.

compnerd: Yeah, I think that the inline is going to be better for now.  We can extract it when needed.

pvieito: Done!

compnerd:
Could you please squash the changes @pvieito?

pvieito:
@compnerd Squashed at #2133

PR #2128, opened 2019-04-17
Parity: objectValue(…)

millenomi:
swift-corelibs-foundation’s Formatter subclasses do not have working funnel methods at the moment. Rather than advertise an experimental API that has no implementation, retire the experimental API for now and provide source compatibility stubs that aid in porting code at compile time.

Fixes https://bugs.swift.org/browse/SR-10353, https://bugs.swift.org/browse/SR-10354


millenomi:
@swift-ci please test

PR #2130, opened 2019-04-17
Foundation: remove a number of non-portable POSIX codes

compnerd:
These are not available on Windows, do not use them.

### DIFF
--- a/Foundation/NSError.swift
+++ b/Foundation/NSError.swift
@@ -966,8 +966,10 @@ extension POSIXError {
     /// Bad address.
     public static var EFAULT: POSIXError.Code { return .EFAULT }
 
+    #if !os(Windows)
     /// Block device required.
     public static var ENOTBLK: POSIXError.Code { return .ENOTBLK }
+    #endif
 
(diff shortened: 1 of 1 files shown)


compnerd:
@colemancda not sure if this is what you had in mind

millenomi:
LGTM, but @colemancda let me know if this is good by you.

Foundation/NSError.swift
```diff
@@ -966,8 +966,10 @@ extension POSIXError {
     /// Bad address.
     public static var EFAULT: POSIXError.Code { return .EFAULT }
 
+    #if !os(Windows)
```
colemancda: I was doing
```
#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS) || os(Linux) || os(Android)
    /// Block device required.
    public static var ENOTBLK: POSIXError.Code { return .ENOTBLK }
    #endif
```

But yours is probably better xD

colemancda:
@compnerd @millenomi Looks good!

compnerd:

PR #2132, opened 2019-04-17
Parity: MeasurementFormatter, DateComponentFormatter

millenomi:
ICU currently doesn’t have a C header or implementation for these. Mark as unavailable until we do.

Fixes https://bugs.swift.org/browse/SR-10352, https://bugs.swift.org/browse/SR-10357


millenomi:
@swift-ci please test

PR #2134, opened 2019-04-17
Remove explicit ==, hashing definitions from RawRepresentable types

lorentey:
This implements @millenomi's review note in https://github.com/apple/swift-corelibs-foundation/pull/2118#discussion_r275510020.

Remove explicit definitions for == and `hash(into:)` from Foundation's enum-like RawRepresentable types. The stdlib provides identical definitions in conditional extensions of RawRepresentable.

Note: This would be an ABI-breaking change in Darwin's Foundation overlays because it removes these symbols; but it seems a reasonable simplification for corelibs-foundation, where AFAIK we don't have to be compatible with Swift 5's ABI.

Note also that RawRepresentable's definitions are inlinable, while the original definitions weren't. The corresponding definitions on Darwin are already inlinable, so this is unlikely to cause *new* issues.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -829,14 +829,6 @@ public struct FileAttributeKey : RawRepresentable, Equatable, Hashable {
         self.rawValue = rawValue
     }
 
-    public func hash(into hasher: inout Hasher) {
-        hasher.combine(rawValue)
-    }
-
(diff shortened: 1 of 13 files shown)


lorentey:
@swift-ci test

millenomi:

PR #2136, opened 2019-04-17
TestFoundation: limit `test_arrayUsedAsCFArrayInvokesArrayMethods`

compnerd:
This test uses CoreFoundation methods.  However, CoreFoundation is only
available on Darwin and Linux.  Limit the test to those environments
specifically.


compnerd:
@swift-ci please test

PR #2140, opened 2019-04-18
Match Darwin implementation of `NSError.localizedDescription`

colemancda:
Fixes the difference between what NSError on Darwin prints and what we are currently printing.

### DIFF
--- a/Foundation/NSError.swift
+++ b/Foundation/NSError.swift
@@ -127,13 +127,30 @@ open class NSError : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
     
     open var localizedDescription: String {
-        let desc = userInfo[NSLocalizedDescriptionKey] as? String
-        
-        return desc ?? "The operation could not be completed"
+        if let localizedDescription = userInfo[NSLocalizedDescriptionKey] as? String {
+            return localizedDescription
+        } else {
+            // placeholder values
+            return "The operation could not be completed." + " " + (self.localizedFailureReason ?? "(\(domain) error \(code).)")
+        }
     }
     
     open var localizedFailureReason: String? {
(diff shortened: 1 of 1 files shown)


colemancda:
@compnerd please test

Foundation/NSError.swift
```diff
@@ -481,7 +498,7 @@ public protocol _BridgedNSError : __BridgedNSError, RawRepresentable, _Objective
 
 /// Describes a bridged error that stores the underlying NSError, so
 /// it can be queried.
-public protocol _BridgedStoredNSError : __BridgedNSError, _ObjectiveCBridgeableError, CustomNSError, Hashable {
+public protocol _BridgedStoredNSError : __BridgedNSError, _ObjectiveCBridgeableError, CustomNSError, Hashable, CustomStringConvertible {
```
compnerd: Mind moving the `CustomStringConvertible` to the extension below?

compnerd:
@swift-ci please test

Foundation/NSError.swift
colemancda: @compnerd That is what I initially attempted, but "protocol extensions cannot have inheritance clauses"

colemancda: A conditional conformance would not work either, because at some point each concrete type (`POSIXError`, `CocoaError`, `URLError`) would need to declare `CustomStringConvertible` conformance.

compnerd: Hmm, I wonder if that is something special here, because I've done extensions for conformances previously.

colemancda:
@compnerd Passed CI, ready to merge.

compnerd:
I would like @millenomi to take a look at this due to the change in the conformance.

millenomi:
Since this is an internal class, I think we have that flexibility.

PR #2142, opened 2019-04-18
[BUILD AND TEST FIX] Mark several tests as XFAIL

millenomi:
- Add infrastructure to mark a test as XFAIL. You can set the `NS_FOUNDATION_ATTEMPT_XFAIL_TESTS` environment variable to `YES` to attempt them.

 - TestFileManager fails to find a creationDate on kernel >= 4.11.0 in Docker. Mark this as XFAIL;

 - test_substringFromCFString is crashing on Ubuntu 14.04. Mark this as XFAIL.

 - test_FromContentsOfURL is failing in CI occasionally. Mark this as XFAIL.

PR #2143, opened 2019-04-18
Fix StringTransform usage of forced casts.

drodriguez:
In Linux, a (constant?) CFString cannot be force casted to String
directly. The code generated for those values with optimizations enabled
was simply a crash. The changes uses `_swiftObject` to correctly
transform to a `Swift.String`.

There some other cases in Foundation where other enums are initialized
from CF values, but in those cases the literals of the CF values are
used, which avoids this problem.

Should fix the problems introduced in #2087 in Linux.

PD: I still think that #2087 changes the meaning of the test. It should be reverted or removed.


millenomi:
… Oh, yep, that makes sense.

I have disabled a test that likely tripped on this issue in https://github.com/apple/swift-corelibs-foundation/pull/2142/files — can you revert the `testExpectedToFail` line about `test_substringFromCFString` and see if this fixes that?

drodriguez:
I re-enabled the test (I didn’t pull from last night, so in my working directory it wasn’t disabled). Should be ready to test.

millenomi:
@swift-ci please test

drodriguez:
I think I pushed the new version before the test finished, and CI didn’t start a new test run.

millenomi:
@swift-ci please test

PR #2144, opened 2019-04-18
[android] Fix tilde/home directory tests.

drodriguez:
In Android, the home directory is the root directory. Foundation returns
the right one, but the tests where checking against having a final
trailing slash, which is not true if the home directory is the root
directory.

The change adds an extra check to avoid the failure in case the
resulting tilde expansion is just the root directory. The comment
explains why the change is needed.

Additionally, instead of using `+` to add two strings,
`appendingPathComponent` is used to build an expected value from the
result of `NSHomeDirectory()`, in order to deal with cases where the
`NSHomeDirectory()` has a trailing slash.

### DIFF
--- a/TestFoundation/TestNSString.swift
+++ b/TestFoundation/TestNSString.swift
@@ -1014,25 +1014,30 @@ class TestNSString: LoopbackServerTest {
     }
     
     func test_expandingTildeInPath() {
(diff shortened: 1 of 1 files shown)


millenomi:
@swift-ci please test

PR #2145, opened 2019-04-18
[android] Replace URL tests that use hardcoded /tmp

drodriguez:
In Android /tmp doesn't exist, and the temporary directory is normally
/data/local/tmp. It is not a symlink. The tests were mostly working, but
because `resolvingSymlinksInPath` returns a trailing slash if the
directory actually exists, in Android it was returning without the final
trailing slash in some of the cases.

The changes split the large test into smaller pieces that test one of
the behaviours of `resolvingSymlinksInPath`, and tries not to use the
temporarl directory by itself or suppose anything about it.

There are, however, a couple of tests that check behaviours that are
only applicable to Darwin, so those tests are only performed in Darwin
platforms.

### DIFF
--- a/TestFoundation/TestURL.swift
+++ b/TestFoundation/TestURL.swift
@@ -438,60 +438,103 @@ class TestURL : XCTestCase {
         XCTAssertTrue(strncmp(TestURL.gFileDoesNotExistName, relativePath, lengthOfRelativePath) == 0, "fileSystemRepresentation of file path is wrong")
     }
     
-    func test_URLByResolvingSymlinksInPath() {
-        let files = [
-            NSTemporaryDirectory() + "ABC/test_URLByResolvingSymlinksInPath"
-        ]
-        
-        guard ensureFiles(files) else {
-            XCTAssert(false, "Could create files for testing.")
-            return
-        }
-        
-        // tmp is special because it is symlinked to /private/tmp and this /private prefix should be dropped,
-        // so tmp is tmp. On Linux tmp is not symlinked so it would be the same.
-        do {
-            let url = URL(fileURLWithPath: "/.//tmp/ABC/..")
-            let result = url.resolvingSymlinksInPath().absoluteString
-            XCTAssertEqual(result, "file:///tmp/", "URLByResolvingSymlinksInPath removes extraneous path components and resolve symlinks.")
-        }
-        
-        do {
-            let url = URL(fileURLWithPath: "~")
-            let result = url.resolvingSymlinksInPath().absoluteString
-            let expected = "file://" + FileManager.default.currentDirectoryPath + "/~"
-            XCTAssertEqual(result, expected, "URLByResolvingSymlinksInPath resolves relative paths using current working directory.")
-        }
-
-        do {
-            let url = URL(fileURLWithPath: "anysite.com/search")
-            let result = url.resolvingSymlinksInPath().absoluteString
-            let expected = "file://" + FileManager.default.currentDirectoryPath + "/anysite.com/search"
-            XCTAssertEqual(result, expected)
-        }
-
-        // tmp is symlinked on macOS only
-        #if os(macOS)
-        do {
-            let url = URL(fileURLWithPath: "/tmp/..")
-            let result = url.resolvingSymlinksInPath().absoluteString
-            XCTAssertEqual(result, "file:///private/")
-        }
-        #else
-        do {
-            let url = URL(fileURLWithPath: "/tmp/ABC/test_URLByResolvingSymlinksInPath")
-            let result = url.resolvingSymlinksInPath().absoluteString
-            XCTAssertEqual(result, "file:///tmp/ABC/test_URLByResolvingSymlinksInPath", "URLByResolvingSymlinksInPath appends trailing slash for existing directories only")
-        }
-        #endif
-
-        do {
-            let url = URL(fileURLWithPath: "/tmp/ABC/..")
-            let result = url.resolvingSymlinksInPath().absoluteString
-            XCTAssertEqual(result, "file:///tmp/")
-        }
+    func test_URLByResolvingSymlinksInPathShouldRemoveDuplicatedPathSeparators() {
+        let url = URL(fileURLWithPath: "//foo///bar////baz/")
+        let result = url.resolvingSymlinksInPath()
+        XCTAssertEqual(result, URL(fileURLWithPath: "/foo/bar/baz"))
     }
-    
+        
+    func test_URLByResolvingSymlinksInPathShouldRemoveSingleDotsBetweenSeparators() {
+        let url = URL(fileURLWithPath: "/./foo/./.bar/./baz/./")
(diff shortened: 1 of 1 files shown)


TestFoundation/TestURL.swift
```diff
@@ -394,11 +394,20 @@ class TestURL : XCTestCase {
         }
         
         // tmp is special because it is symlinked to /private/tmp and this /private prefix should be dropped,
-        // so tmp is tmp. On Linux tmp is not symlinked so it would be the same.
+        // so tmp is tmp. On Linux tmp is not symlinked so it would be the same. In Android it doesn't exist
+        // so we have to use NSTemporaryDirectory to be safe.
+        #if !os(Android)
```
compnerd: This is going to be needed for Windows as well.  I think that this is safe to do always (as in always use `NSTemporaryDirectory`.  @millenomi - thoughts?

spevans: I agree we should remove the hardcoded `/tmp` and just use `NSTemporaryDirectory` for all platforms. I think the test just want a directory that exists, it doesn't actually write to it anyway.

drodriguez: In Darwin, at least, I got a `NSTemporaryDirectory` which was `/var/<letters and numbers>`. So the checks against resolving symlinks are going to be wrong if we use `NSTemporaryDirectory`.

A more resilient idea will be creating the directories ourselves, but it might also be complicated in Windows (there's no symlinks until recent versions of Windows, and you need special permissions). In the end `resolvingSymlinksInPath` does only a couple of things: removes empty components, removes `.`, resolves `..`, resolves symlinks, and removes the `/private` prefix. Most of the checks can be done against non-existing paths, except the ones that resolve symlinks (which can be done in any temporary directory), the rest should not access the FS.

If everyone agrees, I can do the changes, and I think it will work on Windows too.

millenomi: I have some PRs coming where `TestFileManager` will create its own directories.

I recommend creating your own directories, for instance: `NSTemporaryDirectory()` appending `./org.swift.Foundation.TestFoundation.\(pid)/TestURL/etc. etc.`

millenomi: @drodriguez See the test code in https://github.com/apple/swift-corelibs-foundation/pull/2149

drodriguez:
Replaced the test implementation for smaller tests testing each of the parts of `resolvingSymlinksInPath`. I’m building macOS still, but the test pass in Linux and Android.

From my understanding, even in Windows, the URL path separator is still `/`, so I think the tests should work unchanged in Windows, but maybe the symlink test has to be disabled.

drodriguez:
Small fix for macOS (a problem using temporary directory with unexisting paths not removing the `/private` part). Should be ready to review and tests.

spevans:
@swift-ci test

drodriguez:
I’m puzzled. From reading https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/2672/console (which seems to be the tests for this PR), it seems to have tested the previous version (c49f9fa) instead of the last version (f3b0874).

spevans:
@swift-ci test

drodriguez:
@swift-ci please test

drodriguez:
Has anyone any more feedback about these changes? Otherwise it would be great to merge. Thanks.

drodriguez:
Ready for review. Rebased with master and removed the conflicts with the recent changes for Windows.

@swift-ci please test

drodriguez:
Rebased on top of a more current master and check it still works on Linux and Android. 

@spevans, @millenomi, @compnerd: Does someone has something to say in favor or against this PR? I would love to get it merged to improve the Android support.

TestFoundation/TestURL.swift
```diff
@@ -434,60 +434,103 @@ class TestURL : XCTestCase {
...
+    func test_resolvingSymlinksInPathShouldNotRemovePrivatePrefixIfOnlyComponent() {
+        // NOTE: this test only works on Darwin, since only there /tmp is
+        // symlinked to /private/tmp.
+        let url = URL(fileURLWithPath: "/tmp/..")
+        let result = url.resolvingSymlinksInPath()
+        XCTAssertEqual(result, URL(fileURLWithPath: "/private"))
+    }
+
+    func test_resolvingSymlinksInPathShouldNotChangeNonFileURLs() {
+        let url = URL(string: "myscheme://server/foo/bar/baz")!
```
spevans: Best to avoid force unwrapping if at all possible to avoid crashing out and stopping other tests from running. This should be: 
```
func test_resolvingSymlinksInPathShouldNotChangeNonFileURLs() throws {
    let url = URL(string: "myscheme://server/foo/bar/baz").unwrapped()
```

or, since only one value is being tested you could use:
```
let url = URL(string: "myscheme://server/foo/bar/baz")
let result = url?.resolvingSymlinksInPath().absoluteString
```

TestFoundation/TestURL.swift
```diff
@@ -434,60 +434,103 @@ class TestURL : XCTestCase {
...
+        let result = url.resolvingSymlinksInPath().absoluteString
+        XCTAssertEqual(result, "myscheme://server/foo/bar/baz")
+    }
+
+    func test_resolvingSymlinksInPathShouldNotChangePathlessURLs() {
+        let url = URL(string: "file://")!
```
spevans: Same as above re: force unwrapping

TestFoundation/TestURL.swift
drodriguez: I will change it. Thanks for pointing it out.

In this case, since it was a literal string, I knew it couldn’t fail, that's why I decided to use the forced unwrapping. But the truth is that `URL(string:)` is our own implementation, so it might actually fail if someone changes the implementation.

drodriguez:

drodriguez:

buttaface:
Can we finally get this in? I tried it when compiling master natively on Android and it gets all the tests working, along with this small tweak:
```
diff --git a/TestFoundation/TestURL.swift b/TestFoundation/TestURL.swift
index 21f9ca88..e6239bde 100644
--- a/TestFoundation/TestURL.swift
+++ b/TestFoundation/TestURL.swift
@@ -324,7 +324,7 @@ class TestURL : XCTestCase {
             }
         }
 
-        #if os(Android)
+        #if os(Linux)
         FileManager.default.changeCurrentDirectoryPath("/data/local/tmp")
         #endif
 
```
I know everyone is on winter break now, but would be good to finally get this in next month.

buttaface:
This pull needs to be rebased. It fixes the last remaining failing test when compiling and running TestFoundation natively on an Android host.

drodriguez:
I was waiting for some reviewer accepting the changes (/cc @millenomi, @spevans, @compnerd). I might try to rebase during the weekend, but I cannot longer work as much as before. You are free to copy this changes, rebase them, and open another if you want to take over.

buttaface:
Did what you suggested, this is in now, can be closed.

PR #2148, opened 2019-04-19
Added `CocoaError` messages

colemancda:
Add missing error messages for known CocoaError codes.

Generated from [playground](https://gist.github.com/colemancda/9f80ff08505d96821380c0b74622e58c) on macOS 10.14.4

### DIFF
--- a/Foundation/NSError.swift
+++ b/Foundation/NSError.swift
@@ -144,7 +144,7 @@ open class NSError : NSObject, NSCopying, NSSecureCoding, NSCoding {
             case NSPOSIXErrorDomain:
                 return String(cString: strerror(Int32(code)), encoding: .ascii)
             case NSCocoaErrorDomain:
-                return nil
+                return CocoaError.errorMessages[CocoaError.Code(rawValue: code)]
             case NSURLErrorDomain:
                 return nil
             default:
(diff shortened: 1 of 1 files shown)


colemancda:
CC @millenomi @compnerd

Foundation/NSError.swift
```diff
@@ -670,6 +670,43 @@ public struct CocoaError : _BridgedStoredNSError {
     }
 }
 
+internal extension CocoaError {
+    static let errorMessages = [
+        Code(rawValue: 4): "The file doesn’t exist.",
```
millenomi: Can you open a bug on https://bugs.swift.org to localize these?

millenomi:
@swift-ci please test

colemancda:
@millenomi Ok, but to be clear, at the moment there is no support for `NSLocalizedString()` on non-Darwin platforms right?

Foundation/NSError.swift
colemancda: Sure

colemancda: @millenomi https://bugs.swift.org/browse/SR-10523

millenomi:
`NSLocalizedString(…)` should work, and bundle resources work as well. We just don't have the string files, or a bundle for the Foundation module.

PR #2150, opened 2019-04-20
TestFoundation: add `-D_DLL` to the build

compnerd:
Declare `_DLL` when building TestFoundation to ensure that the system
functions are imported correctly with the proper DLL storage.


compnerd:

PR #2154, opened 2019-04-22
Parity: NSCoding: DateIntervalFormatter

millenomi:
- Implement NSCoding for DateIntervalFormatter through new hooks in CFDateIntervalFormatter. This requires fixing several bugs uncovered in the process:

 - Correctly initialize CFDateIntervalFormatter’s date and time style.

 - Fix __NSSwiftData to never try to encode itself, matching what the Darwin counterpart does.

 - NSCalendar needs to check if a coder contains values for a specific key before trying to fetch that key.

 - Add tests with fixtures. Since time zone data can come from different ICU versions, only compare time zone names.

 - Prevent fixtures with duplicate identifiers.


millenomi:
Fixes https://bugs.swift.org/browse/SR-10355

millenomi:
@swift-ci please test

Foundation/NSTimeZone.swift
```diff
@@ -42,8 +42,14 @@ open class NSTimeZone : NSObject, NSCopying, NSSecureCoding, NSCoding {
 #endif
         }
 #endif
+        
         super.init()
-        if !_CFTimeZoneInit(_cfObject, tzName._cfObject, aData?._cfObject) {
+        
+        /* From https://developer.apple.com/documentation/foundation/nstimezone/1387250-init:
+         "Discussion
+         As of macOS 10.6, the underlying implementation of this method has been changed to ignore the specified data parameter."
```
millenomi: cc @compnerd Is this going to be a problem for your implementation?

compnerd: Actually for Windows, it is the inverse - that is more inline with the current implementation.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

PR #2155, opened 2019-04-22
SR-10525: Decimal(string:locale:) different results macOS vs Linux

spevans:
- Decimal: Add support for the Locale in init(string:locale:)

- Update the Decimal scanner to match the Scanner helper functions.

- Add tests for compatibility with Darwin.


spevans:
@swift-ci test

PR #2160, opened 2019-04-23
SR-10516: TimeZone.nextDaylightSavingTimeTransition does not return nil

spevans:
- Check the result of CFTimeZoneGetNextDaylightSavingTimeTransition()
  as 0 indicates no time transition will occur.


spevans:
@swift-ci test

PR #2162, opened 2019-04-23
SR-10515: Calendar.date(from:) unexpectedly mutates Calendar instance

spevans:
- date(from:) was setting the timezone before calling
  _CFCalendarComposeAbsoluteTimeV() but was not restoring it afterwards.


spevans:
@swift-ci test

PR #2163, opened 2019-04-23
Foundation/CoreFoundation: correctly link resources

compnerd:
If CoreFoundation is built statically, the resources will be dropped.
Instead, create an OBJECT library for CoreFoundationResources for the
Windows target.  This allows the resources to be incorporated into an
external consumer of CoreFoundation (i.e. Foundation).  This corrects
the resources handling for the timezone tests.


millenomi:
@swift-ci please test

millenomi:
In general, working with Foundation as a library is going to be less and less feasible in the future, as the need for a framework bundle gets greater.

compnerd:
@millenomi - until we get to the point where we are loading the resources from a bundle, I think this tides us over rather well.  But, Im happy to see things evolve towards the framework model.

compnerd:

PR #2164, opened 2019-04-23
[Windows] Modify What Counts as a Regular File

gmittert:
- Reparse points should be checked to ensure they are actually symlinks
- A regular file may not have FILE_ATTRIBUTE_REGULAR set. It may have
  alternative attributes like FILE_ATTRIBUTE_ARCHIVE or
  FILE_ATTRIUBTE_HIDDDEN and should still count as a regular file.


millenomi:

PR #2166, opened 2019-04-23
Un-XFAIL TestURL

millenomi:
- Ensure that the temporary directory has a unique name per run.
 - Fix error checking in `setup_test_paths()`


millenomi:

PR #2168, opened 2019-04-24
[android] Support for time zones in Android.

drodriguez:
The time zone database in Android sits in a non-standard place, and it
can be overriden by a second location. Besides not sitting in a standard
place, it differs from other Unixes in that the database is bundled into
one binary file, and there are no folders with small files for each time
zone. Luckily the database is generated from Olson, so most of the
CFTimeZone code already handles the details.

The code implements a generic enumeration of the time zones contained in
the database (checking for both the override and the standard location).
The enumeration code is used for both enumerating the known time zones,
as well as creating time zones by name.

At many points there was code blocking the usage of time zones in
Android that has been removed.

This allows the tests in TestDateFormatter, TestCalendar and
TestTimeZone to pass on Android.

### DIFF
--- a/CoreFoundation/NumberDate.subproj/CFTimeZone.c
+++ b/CoreFoundation/NumberDate.subproj/CFTimeZone.c
@@ -30,6 +30,8 @@
 #include <unistd.h>
 #if !TARGET_OS_ANDROID
 #include <sys/fcntl.h>
+#else
+#include <sys/endian.h>
 #endif
 #endif
 #if TARGET_OS_WIN32
@@ -70,7 +72,7 @@ struct tzhead {
 
 #include <time.h>
 
-#if !TARGET_OS_WIN32
+#if !TARGET_OS_WIN32 && !TARGET_OS_ANDROID
 static CFStringRef __tzZoneInfo = NULL;
 static char *__tzDir = NULL;
 static void __InitTZStrings(void);
@@ -117,9 +119,10 @@ CF_INLINE void __CFTimeZoneUnlockCompatibilityMapping(void) {
     __CFUnlock(&__CFTimeZoneCompatibilityMappingLock);
 }
 
+#define COUNT_OF(array) (sizeof((array)) / sizeof((array)[0]))
+
 #if TARGET_OS_WIN32
 #define CF_TIME_ZONES_KEY L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Time Zones"
-#define COUNT_OF(array) (sizeof((array)) / sizeof((array)[0]))
 
 /* This function should be used for WIN32 instead of
  * __CFCopyRecursiveDirectoryList function.
@@ -183,11 +186,178 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
 
     RegCloseKey(hKey);
 }
+#elif TARGET_OS_ANDROID
+/*
+ * Android does not ship with the standard Unix Olsen files, with the directory
+ * structure, and the files with the same name as the time zone.
+ * Instead all the information is in one file, where all the time zone
+ * information for all the zones is held. Also, to allow upgrades to the time
+ * zone database without an update to the system, the database can be overriden
+ * by a secondary location.
+ * - /data/misc/zoneinfo/current/tzdata overrides for the time zone information
+ *   from the system.
+ * - /system/usr/share/zoneinfo/tzdata system time zone information.
+ * The format of these files is slightly documented in Bionic's source file
+ * libc/tzcode/bionic.cpp.
+ * The file start with a header which is 24 bytes long.
+ * - 6 bytes should be the ASCII string "tzdata"
+ * - 6 bytes of the Olson database version in ASCII. For example 2018a. Includes
+ *   a final nul character.
+ * - 4 bytes MSB of the offset of the index inside the file.
+ * - 4 bytes MSB of the offset of the start of the data inside the file.
+ * - 4 bytes MSB of the zonetab (unused in this code).
+ * The index sits between the offset for the index and the offset for the data.
+ * Each index entry is 52 bytes long.
+ * - 40 bytes for the name of the zone. Seems to be nul terminated.
+ * - 4 bytes MSB of the offset to this time zone data. Notice that this offset
+ *   is relative to the data offset from the header.
+ * - 4 bytes MSB of the data length.
+ * - 4 bytes unused.
+ */
+
+#define ANDROID_TZ_HEADER_SIZE 24
+#define ANDROID_TZ_HEADER_TAG_SIZE 6
+#define ANDROID_TZ_HEADER_INDEX_OFFSET 12
+#define ANDROID_TZ_HEADER_DATA_OFFSET 16
+#define ANDROID_TZ_ENTRY_SIZE 52
+#define ANDROID_TZ_ENTRY_NAME_LENGTH 40
+#define ANDROID_TZ_ENTRY_START_OFFSET 40
+#define ANDROID_TZ_ENTRY_LENGTH_OFFSET 44
+
+/**
+ * Callback invoked for each of the time zones in the timezone file.
+ * - name: The name of the time zone.
+ * - offset: The final offset inside the file of the data for the time zone.
+ * - length: The length of the data for the time zone.
+ * - fp: The file pointer to read the data from. The file might not be in the
+ *       right offset, so if you want to read the right time zone data, you
+ *       probably want to fseek into offset.
+ * - context1: The argument passed into __CFAndroidTimeZoneListEnumerate.
+ * - context2: The argument passed into __CFAndroidTimeZoneListEnumerate.
+ */
+typedef Boolean (*__CFAndroidTimeZoneListEnumerateCallback)(const char name[ANDROID_TZ_ENTRY_NAME_LENGTH], int32_t offset, int32_t length, FILE *fp, void *context1, void *context2);
+
+static void __CFAndroidTimeZoneParse(FILE *fp, __CFAndroidTimeZoneListEnumerateCallback callback, void *context1, void *context2) {
+    if (!fp) {
+        return;
+    }
+
+    char header[ANDROID_TZ_HEADER_SIZE];
+    if (fread(header, 1, sizeof(header), fp) != sizeof(header)) {
+        return;
+    }
+    if (strncmp(header, "tzdata", ANDROID_TZ_HEADER_TAG_SIZE) != 0) {
+        return;
+    }
+
+    int32_t indexOffset;
+    memcpy(&indexOffset, &header[ANDROID_TZ_HEADER_INDEX_OFFSET], sizeof(int32_t));
+    indexOffset = betoh32(indexOffset);
+
+    int32_t dataOffset;
+    memcpy(&dataOffset, &header[ANDROID_TZ_HEADER_DATA_OFFSET], sizeof(int32_t));
+    dataOffset = betoh32(dataOffset);
+
+    if (indexOffset < 0 || dataOffset < indexOffset) {
+        return;
+    }
+    if (fseek(fp, indexOffset, SEEK_SET) != 0) {
+        return;
+    }
+
+    char entry[52];
+    size_t indexSize = dataOffset - indexOffset;
+    size_t zoneCount = indexSize / sizeof(entry);
+    if (zoneCount * sizeof(entry) != indexSize) {
+        return;
+    }
+    for (size_t idx = 0; idx < zoneCount; idx++) {
+        if (fread(entry, 1, sizeof(entry), fp) != sizeof(entry)) {
+            break;
+        }
+
+        int32_t start;
+        memcpy(&start, &entry[ANDROID_TZ_ENTRY_START_OFFSET], sizeof(int32_t));
+        start = betoh32(start);
+        start += dataOffset;
+
+        int32_t length;
+        memcpy(&length, &entry[ANDROID_TZ_ENTRY_LENGTH_OFFSET], sizeof(int32_t));
+        length = betoh32(length);
+
+        if (start < 0 || length < 0) {
+            break;
+        }
+
+        long pos = ftell(fp);
+        Boolean done = callback(entry, start, length, fp, context1, context2);
+        if (done || fseek(fp, pos, SEEK_SET) != 0) {
+            break;
+        }
+    }
+}
+
+static void __CFAndroidTimeZoneListEnumerate(__CFAndroidTimeZoneListEnumerateCallback callback, void *context1, void *context2) {
+    // The best reference should be Android Bionic's libc/tzcode/bionic.cpp
+    static const char *tzDataFiles[] = {
+        "/data/misc/zoneinfo/current/tzdata",
+        "/system/usr/share/zoneinfo/tzdata"
+    };
+
+    for (int idx = 0; idx < COUNT_OF(tzDataFiles); idx++) {
+        FILE *fp = fopen(tzDataFiles[idx], "rb");
+        __CFAndroidTimeZoneParse(fp, callback, context1, context2);
+        if (fp) {
+            fclose(fp);
+        }
+    }
+}
+
+static Boolean __CFCopyAndroidTimeZoneListCallback(const char name[ANDROID_TZ_ENTRY_NAME_LENGTH], int32_t start, int32_t length, FILE *fp, void *context1, void *context2) {
+    CFMutableArrayRef result = (CFMutableArrayRef)context1;
+    CFStringRef timeZoneName = CFStringCreateWithCString(kCFAllocatorSystemDefault, name, kCFStringEncodingASCII);
+    CFArrayAppendValue(result, timeZoneName);
+    CFRelease(timeZoneName);
+    return FALSE;
+}
+
+static Boolean __CFTimeZoneDataCreateCallback(const char name[ANDROID_TZ_ENTRY_NAME_LENGTH], int32_t start, int32_t length, FILE *fp, void *context1, void *context2) {
+    char *tzNameCstr = (char *)context1;
+    CFDataRef *dataPtr = (CFDataRef *)context2;
+
+    if (strncmp(tzNameCstr, name, ANDROID_TZ_ENTRY_NAME_LENGTH) == 0) {
+        if (fseek(fp, start, SEEK_SET) != 0) {
+            return TRUE;
+        }
+        uint8_t *bytes = malloc(length);
+        if (!bytes) {
+            return TRUE;
+        }
+        if (fread(bytes, 1, length, fp) != length) {
+            free(bytes);
+            return TRUE;
+        }
+        *dataPtr = CFDataCreate(kCFAllocatorSystemDefault, bytes, length);
+        free(bytes);
+        return TRUE;
+    }
+
+    return FALSE;
+}
+
+static CFMutableArrayRef __CFCopyAndroidTimeZoneList() {
+    CFMutableArrayRef result = CFArrayCreateMutable(kCFAllocatorSystemDefault, 0, &kCFTypeArrayCallBacks);
+    __CFAndroidTimeZoneListEnumerate(__CFCopyAndroidTimeZoneListCallback, result, NULL);
+    return result;
+}
+
 #elif TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_BSD
 static CFMutableArrayRef __CFCopyRecursiveDirectoryList() {
     CFMutableArrayRef result = CFArrayCreateMutable(kCFAllocatorSystemDefault, 0, &kCFTypeArrayCallBacks);
+#if !TARGET_OS_ANDROID
     if (!__tzDir) __InitTZStrings();
     if (!__tzDir) return result;
+#endif
     int fd = open(__tzDir, O_RDONLY);
 
     for (; 0 <= fd;) {
@@ -637,6 +807,8 @@ static void __InitTZStrings(void) {
     });
 }
 
+#elif TARGET_OS_ANDROID
+// Nothing
 #elif TARGET_OS_LINUX || TARGET_OS_BSD
 static void __InitTZStrings(void) {
     __tzZoneInfo = CFSTR(TZDIR);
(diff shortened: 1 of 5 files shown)


CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1058,6 +1175,53 @@ Boolean _CFTimeZoneInitInternal(CFTimeZoneRef timezone, CFStringRef name, CFData
...
+        buffer = malloc(maxSize);
+        if (!buffer) {
+            return NULL;
+        }
+        if (CFStringGetCString(tzName, buffer, maxSize, kCFStringEncodingASCII)) {
+            tzNameCstr = buffer;
+        }
+    }
+    if (!tzNameCstr) {
+        return NULL;
```
spevans: Does this need a `if (buffer) { free(buffer); }` before the return because the `CFStringGetCString()`  could fail and `tzNameCStr` could still be `NULL` ?

spevans:
@swift-ci test

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
 
     RegCloseKey(hKey);
 }
+#elif TARGET_OS_ANDROID
+typedef Boolean (*__CFAndroidTimeZoneListEnumerateCallback)(char name[40], int32_t start, int32_t length, FILE *fp, void *context);
```
compnerd: Why the limit of the name?  If you want a nullability attribute instead, just spell it as `_Nonnull`.  Also, the name should be `const` no?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+            continue;
+        }
+        if (strncmp(header, "tzdata", 6) != 0) {
+            fclose(fp);
+            continue;
+        }
+        int32_t indexOffset;
+        memcpy(&indexOffset, &header[12], sizeof(int32_t));
+        indexOffset = betoh32(indexOffset);
+        int32_t dataOffset;
```
compnerd: Newline here would be nice

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+        char header[24];
+        if (fread(header, 1, sizeof(header), fp) != sizeof(header)) {
+            fclose(fp);
+            continue;
+        }
+        if (strncmp(header, "tzdata", 6) != 0) {
+            fclose(fp);
+            continue;
+        }
+        int32_t indexOffset;
```
compnerd: Newline here would be nice

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+            }
+        }
+
+        fclose(fp);
+    }
+}
+
+static Boolean __CFCopyAndroidTimeZoneListCallback(char name[40], int32_t start, int32_t length, FILE *fp, void *context) {
+    CFMutableArrayRef result = (CFMutableArrayRef)context;
+    size_t lenght = strnlen(name, 40);
```
compnerd: Nit: `lenght` -> `length`.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -637,6 +754,13 @@ static void __InitTZStrings(void) {
     });
 }
 
+#elif TARGET_OS_ANDROID
+static void __InitTZStrings(void) {
```
compnerd: Why not handle this as Windows and just not use them?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+    if (strncmp(tzNameCstr, name, 40) == 0) {
+        if (fseek(fp, start, SEEK_SET) != 0) {
+            return TRUE;
+        }
+        uint8_t *bytes = malloc(length);
+        if (!bytes) {
+            return TRUE;
+        }
+        if (fread(bytes, 1, length, fp) != length) {
+            return TRUE;
```
compnerd: You are leaking `bytes`.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+        }
+        *dataPtr = CFDataCreate(kCFAllocatorSystemDefault, bytes, length);
```
compnerd: This copies the `data` buffer, so, you are leaking `bytes`.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+        return TRUE;
+    }
+
```
compnerd: You are leaking `bytes`.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,6 +185,121 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+static Boolean __CFCopyAndroidTimeZoneListCallback(char name[40], int32_t start, int32_t length, FILE *fp, void *context) {
+    CFMutableArrayRef result = (CFMutableArrayRef)context;
+    size_t lenght = strnlen(name, 40);
+    CFStringRef timeZoneName = CFStringCreateWithCString(kCFAllocatorSystemDefault, name, kCFStringEncodingASCII);
+    CFArrayAppendValue(result, timeZoneName);
+    CFRelease(timeZoneName);
+    return FALSE;
+}
+
+struct __CFTimeZoneDataCreateContext {
```
compnerd: Why the `struct` shenanigans?  The callback is entirely local, and the caller is local as well.  You could just adjust the signature to pass the `tzname` and `tzdata`.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: No, it doesn't need it.  `free(NULL)` is well specified.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1058,6 +1175,53 @@ Boolean _CFTimeZoneInitInternal(CFTimeZoneRef timezone, CFStringRef name, CFData
...
+    CFDataRef data = NULL;
+    CFURLRef tempURL = CFURLCreateCopyAppendingPathComponent(kCFAllocatorSystemDefault, baseURL, tzName, false);
+    if (NULL != tempURL) {
+        if (_CFReadBytesFromFile(kCFAllocatorSystemDefault, tempURL, &bytes, &length, 0, 0)) {
+            data = CFDataCreateWithBytesNoCopy(kCFAllocatorSystemDefault, bytes, length, kCFAllocatorSystemDefault);
+        }
+        CFRelease(tempURL);
+    }
+    return data;
+#endif
```
compnerd: Can you split out the change to extract this function?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
spevans: I mean wouldnt the `buffer = malloc(maxSize)` be leaked if  `CFStringGetCString` returned 0 as `tzNameCstr` would still be `NULL` and will return on line 1197 ?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: The name can be `const` indeed, it isn’t in the original struct from Bionic, so I didn’t use `const`. I can change it.

The reason for the 40 is because that's the maximum length of the name in the original struct. There's very little documentation about if there will always be a final zero or not, so I preferred to deal with the length.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: When I normally pass function pointers and they need to share data, there's only one `context` pointer, and you have to figure out how to pack all the needed data into one pointer.

For enumerating the known time zones, only one pointer is needed (the `CFMutableArrayRef`), but for finding out about one time zone, I need both the required time zone identifier and a double pointer to where to leave the data.

I can indeed pass many more pointers, but this was a more general solution, and a pattern everyone will recognize.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: I will check the Windows paths again, but I didn’t want to create conditional compilation branches if not needed (Android falls into most of the Linux branches).

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: I never remember that `free(NULL)` is actually valid. I will correct an instance below.

I will add the `free(buffer)` that simons proposes. I don’t know why I thought that I will be fine here without it, but `CFStringGetCString` might still fail because of the encoding (but `__nameStringOK` should take care of most of those cases anyway).

drodriguez:
This might not apply cleanly without #2172. But one can look only at the second commit, if you want the Android specific changes only.

drodriguez:
After fighting a little with the removal of `__InitTZStrings` and the multiple conditional compilation branches it introduces, I don’t think it is a great idea, but that's the code that it is up for review. If we want to back out of that change because it introduces complexity, we are still on time to do it.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: Please add a comment explaining that.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: I think that both patterns are ones that everyone will recognise.  It is very common in C to have custom callbacks with an arbitrary set of parameters.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,11 +185,133 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+
+static void __CFAndroidTimeZoneListEnumerate(__CFAndroidTimeZoneListEnumerateCallback callback, void *context) {
+    // The best reference should be Android Bionic's libc/tzcode/bionic.cpp
+    static const char *tzDataFiles[] = {
+        "/data/misc/zoneinfo/current/tzdata",
+        "/system/usr/share/zoneinfo/tzdata"
+    };
+
+    Boolean done = FALSE;
+    for (int idx = 0; idx < sizeof(tzDataFiles)/sizeof(tzDataFiles[0]) && !done; idx++) {
```
compnerd: I think it would be nice to extract the `ARRAY_SIZE` into a macro.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,11 +185,133 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+        FILE *fp = fopen(tzDataFiles[idx], "rb");
+        if (!fp) {
+            continue;
+        }
+
+        char header[24];
```
compnerd: Can you add a blob explaining the structure to the beginning?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,11 +185,133 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+        dataOffset = betoh32(dataOffset);
+
+        if (indexOffset < 0 || dataOffset < indexOffset) {
+            fclose(fp);
+            continue;
+        }
+        if (fseek(fp, indexOffset, SEEK_SET) != 0) {
+            fclose(fp);
+            continue;
+        }
```
compnerd: Can you extract the parsing into a helper function?  That avoids the `fclose(fp); continue;` all over the place.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -70,7 +72,7 @@ struct tzhead {
 
 #include <time.h>
 
-#if !TARGET_OS_WIN32
+#if !TARGET_OS_WIN32 && !TARGET_OS_ANDROID
```
compnerd: Hmm, I think that this can now be rewritten to `TARGET_OS_LINUX || TARGET_OS_BSD`.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -702,7 +829,9 @@ static CFTimeZoneRef __CFTimeZoneCreateSystem(void) {
         } else {
             name = CFStringCreateWithBytes(kCFAllocatorSystemDefault, (uint8_t *)linkbuf, strlen(linkbuf), kCFStringEncodingUTF8, false);
         }
-    } else {
+    } else
```
compnerd: I think it might be nicer to just end the previous block with the one duplicated line for the return and then avoid the `else` and braces for the remainder.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1058,6 +1180,52 @@ Boolean _CFTimeZoneInitInternal(CFTimeZoneRef timezone, CFStringRef name, CFData
     return success;
 }
 
+CFDataRef _CFTimeZoneDataCreate(CFURLRef baseURL, CFStringRef tzName) {
```
compnerd: Huh?  Isn't this created in a previous patch?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: I will try to write my understanding of the file format.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: I will see if that's possible, but I’m going to remove a lot of `fclose(fp); continue;` and replace them with a lot of `fclose(fp); return FALSE;`.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: I think Github is not showing the right thing here because this PR include the previous patch to allow it to compile. When I rebase, this should not appear as this.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: I think it is not possible here, since there’s a block in the middle that has to be run for both branches. The return cannot be moved here.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: Oh, bleh, I missed that.  Yeah, the negation makes sense I suppose.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: Hmm, did I miss that?  What is the block which needs to be executed?

drodriguez: The `if (name)` block that starts in L890 (new version). The previous branches (the `if` and the `else` branch, only the `else` is valid for Android), are trying to find the `name`, while that last bit is find the information about the time zone with that name.

drodriguez:
@swift-ci please test

drodriguez:
@compnerd, @spevans: thanks for all the feedback already. I was wondering if there's more pieces that anyone thinks need improving, or we can plan to merge this into master. Thanks!

drodriguez:
Fixed some merge errors for the Android code.

@swift-ci please test

buttaface:
Heh, I had to make almost the same changes to the previous commit to get it to compile, though I didn't spot the fclose fix and it would segfault.

Now, several dozen more timezone tests pass on Android AArch64 with this latest commit, nice work.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -183,11 +186,178 @@ static void __CFTimeZoneGetOffset(CFStringRef timezone, int32_t *offset) {
...
+ * - name: The name of the time zone.
+ * - offset: The final offset inside the file of the data for the time zone.
+ * - length: The length of the data for the time zone.
+ * - fp: The file pointer to read the data from. The file might not be in the
+ *       right offset, so if you want to read the right time zone data, you
+ *       probably want to fseek into offset.
+ * - context1: The argument passed into __CFAndroidTimeZoneListEnumerate.
+ * - context2: The argument passed into __CFAndroidTimeZoneListEnumerate.
+ */
+typedef Boolean (*__CFAndroidTimeZoneListEnumerateCallback)(const char name[ANDROID_TZ_ENTRY_NAME_LENGTH], int32_t offset, int32_t length, FILE *fp, void *context1, void *context2);
```
compnerd: I would give them proper names rather than `context1` and `context2`.  (e.g. `name`, `data`)

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: How about folding `!TARGET_OS_ANDROID` into the `if` condition on L868?

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1149,7 +1347,9 @@ Boolean _CFTimeZoneInit(CFTimeZoneRef timeZone, CFStringRef name, CFDataRef data
         CFStringRef mapping = CFDictionaryGetValue(dict, name);
         if (mapping) {
             name = mapping;
-        } else if (CFStringHasPrefix(name, __tzZoneInfo)) {
+        }
+#if !TARGET_OS_ANDROID
```
compnerd: I think it would be nicer to just fold this into the condition below as `} else if (!TARGET_OS_ANDROID && CFStringHasPrefix(name, __tzZoneInfo))`

CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1351,7 +1556,9 @@ CFTimeZoneRef CFTimeZoneCreateWithName(CFAllocatorRef allocator, CFStringRef nam
 	CFStringRef mapping = CFDictionaryGetValue(dict, name);
 	if (mapping) {
 	    name = mapping;
-	} else if (CFStringHasPrefix(name, __tzZoneInfo)) {
+	}
+#if !TARGET_OS_ANDROID
```
compnerd: Similar, folding this into the `if` condition would be nicer.

Foundation/NSTimeZone.swift
```diff
@@ -29,17 +29,13 @@ open class NSTimeZone : NSObject, NSCopying, NSSecureCoding, NSCoding {
     }
 
     public init?(name tzName: String, data aData: Data?) {
-#if os(Android) || os(Windows)
+#if os(Windows)
```
compnerd: :-(

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: They are `name` and `data` in the case you are calling `__CFTimeZoneDataCreateCallback`, but `context1` is actually a `CFMutableArrayRef` in the case of calling `__CFCopyAndroidTimeZoneListCallback`. That's why I used just one `context` before.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: I'm not sure exactly why you mean (please provide the code of what you mean), but `__tzZoneInfo` is used in L866, and that variable doesn't exist in Android, so it has to be compiled out.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: `__tzZoneInfo` doesn't exist in Android, so even if that will work, the references to the variable will make the compilation fail.

PD: `__tzZoneInfo` used to exist in Android in a previous version of this PR, but it was an empty dummy, which removes a lot of conditional compilation from the rest of the code.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
drodriguez: Same explanation as above.

Foundation/NSTimeZone.swift
drodriguez: I tried to not modify the Windows code. I was under the impression that you added support for Windows. Is this a left over?

compnerd: No, this implies that we have more work to do on Windows and android is currently doing better than Windows in the Time Zone department.

CoreFoundation/NumberDate.subproj/CFTimeZone.c
compnerd: Oh, right.  Yeah, that was the desire, to get rid of that path which doesn't really make sense.

PR #2170, opened 2019-04-24
[android] Enable locale test which actually works on Android.

drodriguez:
Maybe the test was disabled because ICU wasn't available then, but
currently the test simply works.

This should not affect Darwin or Linux in any way.


millenomi:

PR #2171, opened 2019-04-24
[android] Enable NumberFormatter test that actually works on Android.

drodriguez:
The original commit doesn't state why the test in particular was
disabled, but in my tests (AArch64), the test seems to work.

/cc @johnno1962


millenomi:

PR #2172, opened 2019-04-24
Move duplicated code into its own function.

drodriguez:
To allow easier adoption of the Android code, this duplicated piece of
code is moved into its own function, since it will need to change
significantly for Android.

As requested in #2168.


drodriguez:
Found two more instances of the same code.

compnerd:
@swift-ci please test

compnerd:

PR #2173, opened 2019-04-24
NSStringAPI: perf: UTF8 String -> Data

ianpartridge:
Currently, str.data(using:allowLossyConversion:) always bridges via
NSString.

As String is now natively UTF-8 we can fastpath this conversion in the
case where the user requests UTF-8 encoding.

Port of apple/swift@a6fbc40 from the overlay.


ianpartridge:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #2175, opened 2019-04-25
[5.0] NSStringAPI: perf: UTF8 String -> Data

ianpartridge:
Currently, str.data(using:allowLossyConversion:) always bridges via
NSString.

As String is now natively UTF-8 we can fastpath this conversion in the
case where the user requests UTF-8 encoding.

Port of
https://github.com/apple/swift/commit/a6fbc40e761f106cae7192288d920a17c7351669
from the overlay.

Backport of https://github.com/apple/swift-corelibs-foundation/commit/7704c94c212e76db4db1a718268e64c171a5e788.


ianpartridge:
@swift-ci test

ianpartridge:
@weissi let's make Linux faster than Darwin :-)

PR #2176, opened 2019-04-25
[android] Enable tests for FileManager in Android.

drodriguez:
Many tests were disabled in Android, but they actually work without
almost changes.

- Android uses /proc/mounts as Linux. Enabling that code path gives good
  results in Android.
- `attributesOfFileSystem` works on Android once glibc.modulemap
  includes the right headers (see apple/swift#24286).
- However, since Process is still not available in Android, the test
  that relies on xdgTestHelper tool will not work, and have been
  disabled (but instead of causing the test to fail, they are simply not
  run in Android, which probably will never conform to XDG).

### DIFF
--- a/TestFoundation/TestFileManager.swift
+++ b/TestFoundation/TestFileManager.swift
@@ -348,7 +348,6 @@ class TestFileManager : XCTestCase {
     }
     
     func test_fileSystemAttributes() {
-#if !os(Android)
         let fm = FileManager.default
         let path = NSTemporaryDirectory()
         
@@ -379,7 +378,6 @@ class TestFileManager : XCTestCase {
         } catch {
             XCTFail("\(error)")
         }
-#endif
     }
     
     func test_setFileAttributes() {
@@ -1084,7 +1082,7 @@ class TestFileManager : XCTestCase {
         try testCopy()
     }
     
-#if !DEPLOYMENT_RUNTIME_OBJC // XDG tests require swift-corelibs-foundation
+#if !DEPLOYMENT_RUNTIME_OBJC && !os(Android) // XDG tests require swift-corelibs-foundation
     
     #if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT // These are white box tests for the internals of XDG parsing:
     func test_xdgStopgapsCoverAllConstants() {
@@ -1201,16 +1199,9 @@ VIDEOS=StopgapVideos
     #endif // NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
     
     // This test below is a black box test, and does not require @testable import.
-    
-    enum TestError: Error {
-        case notImplementedOnThisPlatform
-    }
-    
+
+    #if !os(Android)
     func printPathByRunningHelper(withConfiguration config: String, method: String, identifier: String) throws -> String {
-        #if os(Android)
-            throw TestError.notImplementedOnThisPlatform
-        #endif
-        
         let uuid = UUID().uuidString
         let path = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("org.swift.Foundation.XDGTestHelper").appendingPathComponent(uuid)
         try FileManager.default.createDirectory(at: path, withIntermediateDirectories: true)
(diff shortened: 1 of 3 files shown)


millenomi:
Note that xdgTestHelper is used in other tests and is increasingly misnamed.

Is there a way we can spawn processes programmatically on Android

drodriguez:
I’m looking into it. Some of the POSIX APIs used by Foundation appeared in very recent Androids (ref: https://android.googlesource.com/platform/bionic/+/master/libc/include/spawn.h), so I don’t think they are a possibility. I don’t know if I can do `fork`/`exec` instead. I don’t know if `Process` will be possible to bring to Android at all (and so the tests that depend on it will probably never pass in Android).

TestFoundation/TestFileManager.swift
```diff
@@ -1084,7 +1082,7 @@ class TestFileManager : XCTestCase {
         try testCopy()
     }
     
-#if !DEPLOYMENT_RUNTIME_OBJC // XDG tests require swift-corelibs-foundation
+#if !DEPLOYMENT_RUNTIME_OBJC && !os(Android) // XDG tests require swift-corelibs-foundation
     
     #if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT // These are white box tests for the internals of XDG parsing:
```
buttaface: This isn't set on linux or Android, so none of these XDG tests are compiled or run anyway.

TestFoundation/TestFileManager.swift
```diff
@@ -1565,15 +1557,15 @@ VIDEOS=StopgapVideos
             ("test_displayNames", test_displayNames),
         ]
         
-        #if !DEPLOYMENT_RUNTIME_OBJC && NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
+        #if !DEPLOYMENT_RUNTIME_OBJC && NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT && !os(Android)
```
buttaface: Ditto, NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT not set, so checking for Android is redundant.

buttaface:
> I don’t know if I can do fork/exec instead.

[The posix_spawn added in Pie last year is a wrapper around fork](https://android.googlesource.com/platform/bionic/+/f6e5b582604715729b09db3e36a7aeb8c24b36a4/libc/bionic/spawn.cpp#151), it has been backported to older Android versions too.

spevans:
@swift-ci test

drodriguez:

spevans:
@swift-ci test linux

TestFoundation/TestFileManager.swift
drodriguez: You are right, but the reason for adding the two checks is that `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT` is there because `@testable` is not supported in Linux, while `!os(Android)` is there because Android will probably never support XDG. If we ever solve the first problem, the second doesn’t go away, so having both checks in place make sense to me (otherwise you will have to check each place where `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT` is being used for “phantom” conditions, which sounds impossible).

I can change it if everyone disagrees.

TestFoundation/TestFileManager.swift
drodriguez: Same explanation as above.

drodriguez:
@buttaface: I haven’t had problems with `mode_t` or `TimeInterval`. It seems to compile fine with NDK r18 and r16. There's a workaround at the top of the files for `mode_t`, I wonder if that's what’s creating the problem in your setup.

Re `spawn` vs. `fork/exec`. It is completely possible to replicate `spawn`, but the code also uses actions and other pieces of the API that will have to be branched out or duplicated. I think it is not impossible, but I don’t think it is easy either.

TestFoundation/TestFileManager.swift
buttaface: Sounds good to me, I already approved the pull.

buttaface:
>I haven’t had problems with mode_t or TimeInterval. It seems to compile fine with NDK r18 and r16.

Hmm, I'm using the latest NDK r19, with the Android API set to 21, could be why.

>There's a workaround at the top of the files for mode_t, I wonder if that's what’s creating the problem in your setup.

Yes, that bitwise & function for Android makes other code not compile: I haven't looked into why, just disabled it.

>It is completely possible to replicate spawn

I'll try it with the backported posix_spawn from Bionic and see how it goes.

spevans:
@swift-ci test linux

PR #2178, opened 2019-04-26
[android] Enable Foundation.Host

drodriguez:
Depends on apple/swift#24286

The implementation of Host was disabled for Android, probably because
Glibc didn't include one header (netdb.h) which was needed to compile
this file. But changing Glibc is not enough, because the Android
implementation of that function and other bits differ slightly from
Linux implementation. For those cases private overrides are provide, to
keep the code unchanged.

There's also a little bit of formatting in one of Windows parts, which
was missing two spaces of indentation (probably checking without
whitespace enabled is good for this diff).

### DIFF
--- a/Foundation/Host.swift
+++ b/Foundation/Host.swift
@@ -9,6 +9,23 @@
 
 import CoreFoundation
 
+#if os(Android)
+    // Android Glibc differs a little with respect to the Linux Glibc.
+
+    // IFF_LOOPBACK is part of the enumeration net_device_flags, which needs to
+    // convert to UInt32.
+    private extension UInt32 {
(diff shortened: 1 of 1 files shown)


millenomi:
@swift-ci please test

compnerd:
@drodriguez - mind resolving the conflicts?

drodriguez:
It created a merge commit, sadly. I will rebase and squash when I have access to something else but the web editor.

drodriguez:
It should be good to go now.

drodriguez:
@swift-ci please test

spevans:
@swift-ci test

PR #2179, opened 2019-04-26
DateFormatter: Fix date string parse with Japanese calendar

novi:
DateFormatter with Japanese calendar does not parse Japanese era name and years correctly on Linux environment.
This changes fix it and adds small tests related to Japanese calendar for DateFormatter.

### DIFF
--- a/CoreFoundation/Locale.subproj/CFDateFormatter.c
+++ b/CoreFoundation/Locale.subproj/CFDateFormatter.c
@@ -1557,7 +1557,7 @@ static UDate __CFDateFormatterCorrectTimeToARangeAroundCurrentDate(UCalendar *ca
     return __CFDateFormatterCorrectTimeWithTarget(calendar, at, currEraOrCentury+offset, isEra, status);
 }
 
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

ianpartridge:
Interesting. I guess we will need an updated ICU to support the [Reiwa](https://en.wikipedia.org/wiki/Japanese_calendar_era_bug) era too.

spevans:
It looks to be supported in ICU64-2

novi:
New Japanese era name will be supported by updating libicu to 60.3 on Ubuntu 18.04 (55.2 on Ubuntu 16.04) in my understanding.

ianpartridge:
Swift ships its own ICU and does not depend on the system ICU, so we'll need to update the Swift version.

spevans:
@swift-ci test linux

novi:
Thanks, My understanding is incorrect. I will check them.

spevans:
@swift-ci test linux

spevans:
@swift-ci test and merge

PR #2180, opened 2019-04-26
SR-9900: NSURLRequest .url differs between Darwin and Linux for relative URLs

spevans:
- URLRequest.init(url:cachePolicy:timeoutInterval:): Assign the absolute
  URL to .url


spevans:
@swift-ci test

PR #2181, opened 2019-04-26
[android] Enable Thread tests were possible.

drodriguez:
Enable most of the Thread test for Android. The only tests that cannot
pass right now are the ones related to backtraces, which do not use the
Linux API backtrace and would need an specific implementation.

And while Android did not get pthread_getname_np until API 26, the low
level prctl(PR_GET_NAME) works with the same restriction as in Linux
(only 15 characters). With that change, the test about thread names also
passes in Android.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -36,6 +36,10 @@
 
 #endif
 
+#if TARGET_OS_ANDROID
+#include <sys/prctl.h>
+#endif
+
 #if TARGET_OS_MAC || TARGET_OS_WIN32
 #define kCFPlatformInterfaceStringEncoding	kCFStringEncodingUTF8
 #else
@@ -1495,6 +1499,17 @@ CF_CROSS_PLATFORM_EXPORT int _CFThreadSetName(_CFThreadRef thread, const char *_
 CF_CROSS_PLATFORM_EXPORT int _CFThreadGetName(char *buf, int length) {
 #if TARGET_OS_MAC
     return pthread_getname_np(pthread_self(), buf, length);
+#elif TARGET_OS_ANDROID
+    // Android did not get pthread_getname_np until API 26, but prctl seems to
+    // return at most 15 chars of the name + null terminator.
+    char *buffer[16] = {0};
+    if (prctl(PR_GET_NAME, buffer, 0, 0, 0) != 0) {
+        return -1;
+    }
+    size_t sz = MIN(strnlen(buffer, 15), length - 1);
+    memcpy(buf, buffer, sz);
+    buf[sz] = 0;
+    return 0;
 #elif TARGET_OS_LINUX
     return pthread_getname_np(pthread_self(), buf, length);
 #endif
@@ -1532,4 +1547,3 @@ CF_CROSS_PLATFORM_EXPORT void *_CFReallocf(void *ptr, size_t size) {
 }
 
 #endif
-
--- a/TestFoundation/TestThread.swift
+++ b/TestFoundation/TestThread.swift
@@ -21,17 +21,19 @@
 
 class TestThread : XCTestCase {
     static var allTests: [(String, (TestThread) -> () throws -> Void)] {
-#if os(Android)
-        return []
-#endif
-
         var tests: [(String, (TestThread) -> () throws -> Void)] = [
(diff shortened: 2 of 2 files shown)


spevans:
@swift-ci test

drodriguez:
Removed a preprocessor flag that shouldn’t have been there.

@swift-ci please test

spevans:
@swift-ci test

drodriguez:
@millenomi: I think I have seen `TestFileHandle.test_readWriteHandlers` apparently lock up in several CI runs already. I wonder if the `semaphore.wait(timeout: .distantFuture)` can be reduced to a couple of seconds, to avoid locking up forever (probably breaking when `result` is not `.success`). Do you mind me doing that change? It will still fail, but at least the rest of the test suite will run, and the CI machine will not be hold for that long. Do you have a good timeout? 10 seconds?

drodriguez:

millenomi:
There is an issue in libdispatch that is being tracked, but the wait kinda needs to match Darwin's.

spevans:
Would an `alarm(600)` in `Testfoundation/main.swift` act as a simple timout in that `SIGALRM` should terminate the tests after 10 minutes?

drodriguez:
Is the libdispatch issue in a public tracker that I can follow? I don’t know what you refer with “match Darwin's”, but my proposal is changing the timeout for every platform. If I read the code correctly, this is writing and reading from a pipe, so the delays from reading/writing should be minimal, whatever the platform. I think the change will give us the same signal as we have now, but clearer (it will be the assert about the timeout), instead of a CI build killed because of such timeout. In any case, whatever you think is best.

drodriguez:

drodriguez:
Rebased and applied the changes from `DEPLOYMENT_TARGET_*` to `TARGET_OS_*`.

@swift-ci please test

drodriguez:

drodriguez:
@swift-ci please test

drodriguez:
Rebased to fix the conflicts in the test file.

@swift-ci please test

drodriguez:
Rebasing to a recent master. Still passing tests in Linux and Android.

@spevans, @millenomi, @compnerd: some feedback? positive or negative? I'm trying to clean up my old “stuck” PRs.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1495,6 +1499,15 @@ CF_CROSS_PLATFORM_EXPORT int _CFThreadSetName(_CFThreadRef thread, const char *_
...
 #if TARGET_OS_MAC
     return pthread_getname_np(pthread_self(), buf, length);
+#elif TARGET_OS_ANDROID
+    // Android did not get pthread_getname_np until API 26, but prctl seems to
+    // return at least 15 chars of the name.
+    char *buffer[16] = {0};
+    if (prctl(PR_GET_NAME, buffer, 0, 0, 0) != 0) {
+        return -1;
+    }
+    memcpy(buf, buffer, MIN(strnlen(buffer, 16), length));
```
spevans: Dont you need to copy the `NUL` terminator as well. Should it not be:
```
size_t sz = MIN(strnlen(buffer, 16), length - 1);
memcpy(buf, buffer, sz);
buf[sz] = '\0';
```

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1495,6 +1499,15 @@ CF_CROSS_PLATFORM_EXPORT int _CFThreadSetName(_CFThreadRef thread, const char *_
 CF_CROSS_PLATFORM_EXPORT int _CFThreadGetName(char *buf, int length) {
 #if TARGET_OS_MAC
     return pthread_getname_np(pthread_self(), buf, length);
+#elif TARGET_OS_ANDROID
+    // Android did not get pthread_getname_np until API 26, but prctl seems to
+    // return at least 15 chars of the name.
```
spevans: Should this not be 'return at MOST 15 chars of the name' ?

CoreFoundation/Base.subproj/CFPlatform.c
drodriguez: 👍

drodriguez:

drodriguez:

spevans:
LGTM

PR #2182, opened 2019-04-26
[android] Removing explicit castings for unsigned integers.

drodriguez:
Fixes SR-7469.

Seems that the explicit castings are not longer needed in the current
tree compiling for Android.


millenomi:

PR #2183, opened 2019-04-26
StringTransform is a top level type not nested in NSString.

spevans:


spevans:
@swift-ci test

compnerd:
Hmm, is that so?  I got this from the documentation, I wonder how I got that wrong.  Where is this canonically documented?

compnerd:
CC: @millenomi

spevans:
https://developer.apple.com/documentation/foundation/stringtransform

Also try this in Swift Playgrounds, it gives no error:

```
import Foundation
print(StringTransform.latinToArabic.rawValue)
```

PR #2189, opened 2019-04-27
SR-10565: Process crashes when setting standardOutput = FileHandle.nullDevice

spevans:
- If .standardInput, .standardOutput or .standardError are set to
  FileHandle.nullDevice then dup the filedescriptor to '/dev/null'.

Fixes https://bugs.swift.org/browse/SR-10564 and https://bugs.swift.org/browse/SR-10566

### DIFF
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -387,27 +387,51 @@ open class Process: NSObject {
         var siStartupInfo: STARTUPINFOW = STARTUPINFOW()
         siStartupInfo.cb = DWORD(MemoryLayout<STARTUPINFOW>.size)
 
+        var _devNull: FileHandle?
+        func devNullFd() throws -> HANDLE {
+            _devNull = try _devNull ?? FileHandle(forUpdating: URL(fileURLWithPath: "NUL", isDirectory: false))
+            return _devNull!.handle
+        }
+
         switch standardInput {
         case let pipe as Pipe:
-          siStartupInfo.hStdInput = pipe.fileHandleForReading.handle
+            siStartupInfo.hStdInput = pipe.fileHandleForReading.handle
+
+        // nil or NullDevice maps to NUL
+        case let handle as FileHandle where handle === FileHandle._nulldeviceFileHandle: fallthrough
+        case .none:
+            siStartupInfo.hStdInput = try devNullHnd()
+
         case let handle as FileHandle:
-          siStartupInfo.hStdInput = handle.handle
+            siStartupInfo.hStdInput = handle.handle
         default: break
         }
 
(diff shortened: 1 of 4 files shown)


spevans:
@swift-ci test

spevans:
@compnerd @drodriguez Can you provide a windows version

compnerd:
It is spelt "NUL" on Windows.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

TestFoundation/TestProcess.swift
```diff
@@ -505,6 +512,68 @@ class TestProcess : XCTestCase {
...
+        let task = Process()
+        task.executableURL = url
+        task.standardError = FileHandle.nullDevice
+        XCTAssertNoThrow(try task.run())
+        task.waitUntilExit()
+    }
+
+
+    func test_redirect_all_using_null() throws {
+        let url = URL(fileURLWithPath: "/bin/cat", isDirectory: false)
```
millenomi: None of these tests will run outside of a UNIX system. Can we run xdgTestHelper instead, since it's the only executable whose presence we rely on? (Perhaps with a no-op flag.)

Foundation/Process.swift
```diff
@@ -679,30 +679,66 @@ open class Process: NSObject {
         // otherwise.
         var addclose = Set<Int32>()
 
+        var _devNull: FileHandle?
+        func devNullFd() throws -> Int32 {
+            _devNull = try _devNull ?? FileHandle(forUpdating: URL(fileURLWithPath: "/dev/null", isDirectory: false))
```
millenomi: What's the right way to do this on Windows?

spevans:
@swift-ci test

TestFoundation/TestProcess.swift
spevans: I removed all of the unix specific commands (`/bin/cat`, `/bin/env` etc) and added simple versions of them as options in `xdgTestHelper`.

spevans:
@compnerd Is this the best way to open `NUL`? `FileHandle(forUpdating: URL(fileURLWithPath: "NUL", isDirectory: false))` and then I use the `.handle` property?

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

spevans:
@millenomi does this look ok now?

millenomi:

millenomi:

compnerd:
@millenomi this breaks Windows!

PR #2196, opened 2019-04-29
CoreFoundation: correct an over-release in `CFTimeZone`

compnerd:
We would release a value obtained through a `CFDictionaryGetValue` which
would result in the value being over-released as the returned object is
returned at `+0`.  This fixes the use-after-free on Windows.

Thanks to Gwynne Raskind for identifying the mismanagement of the retain
counts!


compnerd:
CC: @gwynne @millenomi

millenomi:

compnerd:

PR #2198, opened 2019-04-30
CoreFoundation: optimize TimeZone PLIST loading on Windows

compnerd:
Use `CFDataCreateWithBytesNoCopy` to avoid the memory copy for the
construction of the `CFData`.

Thanks to Gwynne Raskind for educating me on the fact that
`CFDataCreate` will actually perform a copy which I was trying to avoid.


compnerd:
CC: @gwynne @millenomi

compnerd:
@swift-ci please test

PR #2205, opened 2019-05-01
Fix __builtin_{smulll,saddll}_overflow() usage

spevans:
- Use a variable of type long long variable not int64_t as they may be
  different sizes.

- Fixes "warning: incompatible pointer types passing 'int64_t *' (aka 'long *')
  to parameter of type 'long long *' [-Wincompatible-pointer-types]"

### DIFF
--- a/CoreFoundation/String.subproj/CFString.c
+++ b/CoreFoundation/String.subproj/CFString.c
@@ -6442,9 +6442,9 @@ reswtch:switch (ch) {
 		return true;
 	    }
 	case '1': case '2': case '3': case '4': case '5': case '6': case '7': case '8': case '9': {
-	    int64_t number = 0;
+            long long number = 0;
 	    do {
-                if (__builtin_smulll_overflow(number, 10, &number) || __builtin_saddll_overflow(number, ch - '0', &number)) {
+                if (__builtin_smulll_overflow(number, 10, &number) || __builtin_saddll_overflow(number, ch - '0', &number) || number > INT64_MAX) {
                     if (errorPtr) *errorPtr = __CFCreateOverflowError();
                     return false;
                 }


spevans:
@swift-ci test

spevans:
@swift-ci test linux

CoreFoundation/String.subproj/CFString.c
```diff
@@ -6442,14 +6442,16 @@ reswtch:switch (ch) {
 		return true;
 	    }
 	case '1': case '2': case '3': case '4': case '5': case '6': case '7': case '8': case '9': {
-	    int64_t number = 0;
+            long long _number = 0;
```
compnerd: Why the renaming of the `number`?  That seems gratuitous.

spevans: `number` is used further down as an `int64_t` but that is probably not needed as it only checks for `INT32_MAX` later on, I will update.

spevans:
@swift-ci test

spevans:
@swift-ci test linux

CoreFoundation/String.subproj/CFString.c
```diff
@@ -6442,9 +6442,9 @@ reswtch:switch (ch) {
 		return true;
 	    }
 	case '1': case '2': case '3': case '4': case '5': case '6': case '7': case '8': case '9': {
-	    int64_t number = 0;
+            long long number = 0;
 	    do {
-                if (__builtin_smulll_overflow(number, 10, &number) || __builtin_saddll_overflow(number, ch - '0', &number)) {
+                if (__builtin_smulll_overflow(number, 10, &number) || __builtin_saddll_overflow(number, ch - '0', &number) || number > INT64_MAX) {
```
compnerd: Isn't this a tautological comparison?  `number` is a `long long` which is signed.  `INT64_MAX` is signed.  So the value of the condition is always false.

spevans: `long long` is only defined to be at least 64 bits but could be more if ever ported to other platforms. This is a bit of a nit-picking change but it should make it technically correct.

compnerd: What platform(s) currently define it to something other than 64-bit?

millenomi: I'm okay taking pedantic correctness fixes, since this is code intended to be portable.

millenomi:

PR #2208, opened 2019-05-01
[gardening] Replace usages of deprecated withUnsafeBytes.

drodriguez:
The old signature for withUnsafeBytes/withUnsafeMutableBytes is
deprecated but it is used in the tests. Remove all the usages of the
deprecated signature and use the newer signature instead.

Should remove a lot of deprecation warnings while compiling the tests.


drodriguez:
@swift-ci please test

PR #2158, opened 2019-04-23
Build Fix: TestURL

millenomi:
The catches _are_ exhaustive, and the cast to `NSError` will never fail, but since it is `SwiftFoundation.NSError` (and not ObjC `NSError`) the macOS build cannot detect that; at least, that's my hypothesis. Fix trivially by catching all errors instead.

PR #2210, opened 2019-05-02
[gardening] Remove conditional casts from Error to NSError.

drodriguez:
Since Swift 3 (IIRC) Error can always be casted to NSError, so the
conditional casts are always going to succeed and they produce warnings.
However because in macOS SwiftFoundation.NSError is not the ObjC
NSError, the compiler misses this, and might fail to mark some catch as
exhaustives.

To avoid one warning per use of NSError, move all the warnings into a
single one in a utility function, used from all the places when Error
needs to be transformed into NSError. In a couple of cases, the code is
simplified because a branch disappears.

This should remove a lot of warnings while compiling the tests.

### DIFF
--- a/TestFoundation/TestFileManager.swift
+++ b/TestFoundation/TestFileManager.swift
@@ -1321,81 +1321,81 @@ VIDEOS=StopgapVideos
         XCTAssertNil(NSHomeDirectoryForUser(""))
 
         XCTAssertThrowsError(try fm.contentsOfDirectory(atPath: "")) {
-            let code = CocoaError.Code(rawValue: ($0 as? NSError)!.code)
+            let code = CocoaError.Code(rawValue: $0.asNSError().code)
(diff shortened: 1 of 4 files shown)


drodriguez:
@swift-ci please test

drodriguez:
I will try to have a look at what you are saying. I might be able to avoid the warnings while keeping the compatibility.

drodriguez:
@millenomi: I think I found what you were referring to. Could it be #2158? I think the current solution will be better than the previous one. It still produces a warning, but it was a clear comment that the warning cannot be removed, but at least removes the multiple warnings and reduces noise a lot.

@swift-ci please test

drodriguez:
Rebased on top of current master. I checked if any other of this warning appeared in the mean time and it looks like it still covers all of them.

@millenomi: do you have some feedback about these changes? If you prefer not to apply, that's fine with me. I can abandon this PR in that case.

drodriguez:

PR #2211, opened 2019-05-02
[gardening] Update tests to non-deprecated Process API.

drodriguez:
Some pieces of the Process API has been deprecated (paths have been
turned into URLs, launch has been renamed run and make throwing,...).
These changes update the tests to use the new API.

Since we also want to test the deprecated APIs, a new set of test is
created that specifically test the API still behaves. This set of test
are marked as deprecated themselves, so they don't produce warnings in
their contents. However, they produce warning when used in `allTests`,
which cannot be avoided, but at least those warnings point to a piece of
code that clearly says `test_deprecated`.

There's also some other clean ups in the file, like removing a lot of
`try?` which might have been masking errors (there's no need for marking
the test method as `throws` if you later do `try?`). I have also removed
a lot of `XCTFail`, since they are better served with either `throws`
test methods (and XCTest will report better about the exception than
XCTFail), or use `XCTAssertThrowsError`.

This should remove a lot of warnings while compiling the tests.

### DIFF
--- a/TestFoundation/TestHTTPCookieStorage.swift
+++ b/TestFoundation/TestHTTPCookieStorage.swift
@@ -307,7 +307,7 @@ class TestHTTPCookieStorage: XCTestCase {
         XCTAssertEqual(storage.cookies(for: superSwiftOrgUrl)!, [])
     }
 
-    func test_cookieInXDGSpecPath() {
+    func test_cookieInXDGSpecPath() throws {
 #if !os(Android) && !DARWIN_COMPATIBILITY_TESTS // No XDG on native Foundation
         //Test without setting the environment variable
         let testCookie = HTTPCookie(properties: [
@@ -341,7 +341,7 @@ class TestHTTPCookieStorage: XCTestCase {
         task.environment = environment
 
         // Launch the task
-        task.launch()
+        try task.run()
         task.waitUntilExit()
         let status = task.terminationStatus
         XCTAssertEqual(status, 0)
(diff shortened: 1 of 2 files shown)


drodriguez:
@swift-ci please test

spevans:
Could you hold off on this one please, It will clash with some fixes I made in https://github.com/apple/swift-corelibs-foundation/pull/2189.

drodriguez:
Yes. No problem. I will keep track of the other PR and then I will rebase this one.

millenomi:
This should be an 'and', not an 'or'. We do want the North Star of this to ensure that all API surface is tested, even the deprecated bits.

drodriguez:
As far as I know there's no way to silence the deprecation warnings, right? I would not mind testing the deprecated bits, but the noise they create while compiling is hiding the errors and other warnings that might be important to fix. I wonder if moving all the deprecated usages to their own test will satisfy everyone (the warnings will still create noise, but at least they will be in a file clearly marked for deprecated methods).

drodriguez:
@millenomi: Check if this solution is good for testing both the new API and the deprecated API. I cannot avoid all the warnings, but at least I can reduce them to a minimum, and make it very obvious that they are intentional.

@swift-ci please test

drodriguez:
@millenomi: I was wondering if you will have some time to look at the latest version. I think using the new API, but at the same time have explicit tests for the deprecated APIs, and reducing the warnings during the build improves the health of the build and allow us to advance faster. Thanks!

drodriguez:
Rebased on top of `master` to fix conflicts. Ready for review.

shahmishal:

PR #2212, opened 2019-05-02
[gardening] Remove some optional usage from the tests.

drodriguez:
In some cases in the tests, even if an API returns an optional, we know
that it should never be optional, and it might create warnings in later
code when the optional is interpolated into a string (for example).
Change those cases into using `.unwrapped()` to get a non-optional type,
and create a big error if a nil is actually received.

This should remove a bunch of warnings while compiling the tests.


drodriguez:
@swift-ci please test

spevans:
@swift-ci test and merge

PR #2213, opened 2019-05-02
[gardening] Fix assign to a weak variable releasing immediately.

drodriguez:
The compiler was warning about this assign to a weak variable, where the
value was going to be released immediately. Modify the code to hold the
value in a strong variable during the block to avoid the warning (and
probably a failing test).

This test is actually disabled, so nothing should change.


drodriguez:
After this change (and all the other gardening ones), the only warnings while compiling tests are the ones that are actually testing deprecated APIs (Some usages of `NSArray.write(to:,atomically:)`, `NSArray.init(contentsOfFile:)`, `Scanner.scanString(_,into:)`,…). I don’t want to remove those, specially in cases where there's no alternative.

@swift-ci please test

PR #2216, opened 2019-05-03
Parity: NSCoding: NSIndexSet

millenomi:
- Bridge NSData fully to CF. Work around the fact that NSData has a mutable subclass.

 - Implement NS{Mutable,}IndexSet encoding and decoding.


millenomi:
Fixes https://bugs.swift.org/browse/SR-10403

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
This needs to be written down somewhere, and here is a good spot to do it at a minimum.

On Darwin, `NSData` and many other classes are implemented as class clusters. This means that the classes you interact with normally, like `NSData`, `NSString`, etc., are actually abstract; when you invoke their `+alloc`/`-init…` methods, a suitable subclass is selected. ([More info on that is available in the documentation.](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/ClassClusters/ClassClusters.html)) Darwin provides a tuned set of private subclasses tuned to the specific needs and performance characteristics of the platform. However, if you subclass the public, abstract class, you get none of these implementations; it acts like a true abstract class in all respects.

You may notice that your closest `NSData.h` only declares `@property… bytes` and `@property… length` in the `@interface NSData` proper; most other NSData methods appear in a category. That's because the declared properties are abstract funnels; if the class is subclassed, these need to be overridden and reimplemented without calling `super`. `NSMutableData` adds the `.length` setter and the `@property… mutableBytes`. Methods in the categories are concrete; if they aren't overridden by a subclass, they have default implementations in terms of the abstract funnels. The Darwin subclasses often override these methods to provide more efficient implementations that are aware of the details of their specific backends.

In swift-corelibs-foundation, we have a single concrete implementation of `NS{Mutable,}Data`: the one in `CFData…`. So far, we haven't thought much about this subclassing situation, but with the advent of full `NSData`-to-`CFData…` bridging — which is required by this patch to allow `__NSSwiftData` to encode correctly — we need to resolve the fact, because, where on Darwin we would hit the default implementation, here the methods were immediately calling the `CFData` implementation, which now detects the subclassing and tries to invoke the potentially overridden Swift method again, which invokes the `CFData` implementation in an infinite loop.

This patch matches the Darwin behavior. We cannot have all-abstract methods in the base class and then the concrete methods in an extension like on Darwin, because the latter methods need to be `open`; we also cannot make a class cluster in Swift, since there is no equivalent to the placeholder/trampoline initializer interception method ("factory initializers") that `NSData` uses in its ObjC implementation; NSData needs to be both a concrete class _and_ an abstract class. Thankfully, it is easy to check at runtime which is which, since it needs to act as an abstract class only when subclassed (and not a `NSMutableData`).

This patch adds branched default implementations to all the `open` methods that would be in a category on Darwin, and makes them use the abstract funnels for subclasses. This makes `NS{Mutable,}Data` behave exactly like its ObjC counterpart wrt. subclassing, at the cost of increasing the amount of memory required per-object for subclass instances when comparing to Darwin (since, unlike Darwin's, `NSData` here cannot for now avoid the need to have all the properties needed for its concrete, non-subclassed implementation). Clients that subclassed this class will need to ensure all funnel methods are overridden without invoking `super` going forward, in a way that matches (and thus should be source-compatible with) what one needs to do on Darwin. If a subclasser needs a concrete, default implementation, they will need to delegate to a `CFData` or `NSData` object which hasn't been subclassed.

millenomi:
tl;dr please do not subclass Foundation model classes unless you _really_ have to.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
The funnels need tests added, but this is blocking at least another patch, and I'd rather do it concurrently.

PR #2217, opened 2019-05-04
[Windows] Ensure files open with FILE_SHARE_DELETE

gmittert:
On Windows, open() doesn't set FILE_SHARE_DELETE by default meaning that
if another handle requests deletion privileges, it will always fail
unless the original handle is closed.

An example of where this happens is
```swift
{
    guard let fh = FileHandle(forWritingAtPath: somePath) else { ... }
    defer { try! FileManager.default.removeItem(atPath: somePath) }
    fh.write()
}
```
Without FILE_SHARE_DELETE, the defer happens before the FileHandle's deinit occurs, causing removeItem to fail because it can't open somePath for deletion.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -885,9 +885,73 @@ CF_EXPORT int _NS_rename(const char *oldName, const char *newName) {
 
 CF_EXPORT int _NS_open(const char *name, int oflag, int pmode) {
     wchar_t *wide = createWideFileSystemRepresentation(name, NULL);
-    int fd;
-    _wsopen_s(&fd, wide, oflag, _SH_DENYNO, _S_IREAD | _S_IWRITE);
+
+    DWORD dwDesiredAccess = 0;
+    switch (oflag & (O_RDONLY | O_WRONLY | O_RDWR)) {
+      case _O_RDONLY:
+        dwDesiredAccess = GENERIC_READ;
+        break;
+      case _O_WRONLY:
+        dwDesiredAccess = GENERIC_WRITE;
+        break;
+      case _O_RDWR:
+        dwDesiredAccess = GENERIC_READ | GENERIC_WRITE;
+        break;
+    }
+
+    DWORD dwCreationDisposition;
+    switch (oflag & (O_CREAT | O_EXCL | O_TRUNC)) {
+      case O_CREAT:
+        dwCreationDisposition = OPEN_ALWAYS;
+        break;
+      case O_EXCL:
+        dwCreationDisposition = CREATE_NEW;
+        break;
+      case O_TRUNC:
+        dwCreationDisposition = TRUNCATE_EXISTING;
+        break;
+      default:
+        dwCreationDisposition = OPEN_EXISTING;
+    }
+
+    // Backup semantics are required to receive a handle to a directory
+    DWORD dwFlagsAndAttributes = FILE_FLAG_BACKUP_SEMANTICS;
+    if (pmode & _S_IREAD) {
+      dwFlagsAndAttributes |= FILE_ATTRIBUTE_READONLY;
+    }
+
+    HANDLE handle = CreateFileW(wide, dwDesiredAccess, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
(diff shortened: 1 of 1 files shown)


gmittert:
@millenomi @compnerd

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -884,11 +884,15 @@ CF_EXPORT int _NS_rename(const char *oldName, const char *newName) {
...
 CF_EXPORT int _NS_open(const char *name, int oflag, int pmode) {
+#if TARGET_OS_WIN32
+    return _CFOpenFileWithMode(name, oflag, pmode);
+#else
     wchar_t *wide = createWideFileSystemRepresentation(name, NULL);
     int fd;
     _wsopen_s(&fd, wide, oflag, _SH_DENYNO, _S_IREAD | _S_IWRITE);
     free(wide);
     return fd;
+#endif
```
compnerd: This code is Windows specific already, so this change doesn't make sense.  Just kill the existing implementation.  Although, I think that we want to inline the definition that you placed in `_CFOpenFileWithMode` here as `open` is being replaced with `_NS_open` on Windows.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1444,10 +1448,85 @@ CF_EXPORT char **_CFEnviron(void) {
...
+        break;
+      case _O_WRONLY:
+        dwDesiredAccess = GENERIC_WRITE;
+        break;
+      case _O_RDWR:
+        dwDesiredAccess = GENERIC_READ | GENERIC_WRITE;
+        break;
+    }
+
+    DWORD dwCreationDisposition = OPEN_EXISTING;
```
compnerd: What happens if `_O_CREAT` was passed?  I think it is better to move L1475-L1477 here.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1444,10 +1448,85 @@ CF_EXPORT char **_CFEnviron(void) {
...
+          errno = ENOENT;
+          break;
+        default:
+          errno = EIO;
+          break;
+      }
+      return -1;
+    }
+
+    // _open_osfhandle handles _O_APPEND and _ORDONLY
```
compnerd: Nit: missing `_` in `_O_RDONLY`.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -885,9 +885,73 @@ CF_EXPORT int _NS_rename(const char *oldName, const char *newName) {
...
+    }
+
+    // Backup semantics are required to receive a handle to a directory
+    DWORD dwFlagsAndAttributes = FILE_FLAG_BACKUP_SEMANTICS;
+    if (pmode & _S_IREAD) {
+      dwFlagsAndAttributes |= FILE_ATTRIBUTE_READONLY;
+    }
+
+    HANDLE handle = CreateFileW(wide, dwDesiredAccess, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+        /* lpSecurityAttributes= */nil, dwCreationDisposition, dwFlagsAndAttributes,
```
compnerd: s/nil/NULL/

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -885,9 +885,73 @@ CF_EXPORT int _NS_rename(const char *oldName, const char *newName) {
...
+        /* hTemplatefile= */ nil);
```
compnerd: s/nil/NULL/

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1446,8 +1510,13 @@ CF_EXPORT char **_CFEnviron(void) {
 CF_CROSS_PLATFORM_EXPORT int _CFOpenFileWithMode(const char *path, int opts, mode_t mode) {
     return open(path, opts, mode);
 }
+
 int _CFOpenFile(const char *path, int opts) {
+#if TARGET_OS_WIN32
+    return open(path, opts, 0);
```
compnerd: Just use the 3 parameter `open` always.  Its equivalent.

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

PR #2218, opened 2019-05-04
Fix TestNSData on Windows

gmittert:
In order to delete a directory on Windows, all handles to its children must be
closed.


gmittert:
@swift-ci please test

PR #2219, opened 2019-05-05
SR-10619: Order in which DateFormatter settings are set up matters on Linux

spevans:
- Ensure the setting of .dateStyle and .timeStyle isnt dependant on the
  current setting of .locale.

### DIFF
--- a/TestFoundation/TestDateFormatter.swift
+++ b/TestFoundation/TestDateFormatter.swift
@@ -28,6 +28,7 @@ class TestDateFormatter: XCTestCase {
             ("test_expectedTimeZone", test_expectedTimeZone),
             ("test_dateFrom", test_dateFrom),
             ("test_dateParseAndFormatWithJapaneseCalendar", test_dateParseAndFormatWithJapaneseCalendar),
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

spevans:
cc @weissi

TestFoundation/TestDateFormatter.swift
```diff
@@ -443,4 +444,64 @@ class TestDateFormatter: XCTestCase {
...
+             (".dateStyle", {
+                $0.dateStyle = .medium
+             }),
+             (".timeStyle", {
+                $0.timeStyle = .medium
+             })
+        ]
+
+        // Test all of the combinations of setting the properties produces the same output
+        let expected = "5. mai 2019, 12:52:10"
```
millenomi: Can this be like other tests in that we do not hardcode an answer, so we avoid ICU changes from causing breakage?

spevans: I tested this against Darwin Foundation and it worked find so I don't think this string is ICU version dependant. The PR actually fixes all of the current tests that are disabled due to not working correctly versus Darwin and will be enabled in a followup PR.

PR #2222, opened 2019-05-06
TestFoundation: make xdgTestHelper build on Windows

compnerd:
Windows does not have `kill` but it does have `TerminateProcess`.  This
utility is poorly named at this point and is used as a test helper
utility rather than XDG helper specifically.


compnerd:
CC: @millenomi @spevans @gmittert

millenomi:
@swift-ci please test

PR #2223, opened 2019-05-06
Foundation: expand out `inet_pton`

compnerd:
Rather than use `inet_pton` which cannot be used inline to initialise
the SOCKADDR, initialise the octets manually.  This should correct the
type constructor.


compnerd:
CC: @gmittert

compnerd:
@swift-ci please test

PR #2224, opened 2019-05-06
Foundation: fix socketpair construction on Windows

compnerd:
Windows does not have `socketpair`.  We have a local implementation
`_socketpair` which emulates that behaviour but we did not catch the
instance as it does not explicitly enumerate non-Unix environments.

### DIFF
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -607,10 +607,16 @@ open class Process: NSObject {
             envp.deallocate()
         }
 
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @gmittert

compnerd:
@swift-ci please test

compnerd:
*sigh* use-less.  That is in a non-Windows path.

PR #2225, opened 2019-05-07
[Windows] Fix socket creation in Process.swift

gmittert:
The socket should be set to be non blocking _after_ we attempt to
connect to it.


gmittert:
@compnerd

gmittert:
@swift-ci please test

PR #2227, opened 2019-05-08
CoreFoundation: correct CreateStringFromFileSystemRepresentationByAdd…

compnerd:
…ingPercentEscapes

On Windows,
`CreateStringFromFileSystemRepresentationByAddingPercentEscapes` would
only partially convert the slash character to %2F.  This would result in
paths being truncated.  Fix this behaviour to canonicalise the separator
properly always.  Unfortunately, this uncovered some new cases of file
path handling being broken on Windows.  Add tests to record those cases
so that we can fix them subsequently.

### DIFF
--- a/TestFoundation/TestURL.swift
+++ b/TestFoundation/TestURL.swift
@@ -47,6 +47,60 @@ private func getTestData() -> [Any]? {
 }
 
 class TestURL : XCTestCase {
+#if os(Windows)
+    func test_WindowsPathSeparator() {
+      // ensure that the mixed slashes are handled properly
+      // e.g. NOT file:///S:/b/u1%2/
+      let u1 = URL(fileURLWithPath: "S:\\b\\u1/")
+      XCTAssertEqual(u1.absoluteString, "file:///S:/b/u1/")
+
+      // ensure that trailing slashes are compressed
+      // e.g. NOT file:///S:/b/u2%2F%2F%2F%/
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

TestFoundation/TestURL.swift
```diff
@@ -619,6 +671,15 @@ class TestURL : XCTestCase {
             ("test_URLResourceValues", testExpectedToFail(test_URLResourceValues,
                 "test_URLResourceValues: Except for .nameKey, we have no testable attributes that work in the environment Swift CI uses, for now. SR-XXXX")),
         ]
+
+#if os(Windows)
+        tests.append(contentsOf: [
+            ("test_pathSeparator", test_pathSeparator),
```
millenomi: Can we add 'Windows' in the test names so it's obvious above who they're for? And/or wrap the tests above in `#if os(Windows)` too, since they don't run anywhere else.

compnerd: yes & yes

compnerd:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

compnerd:

compnerd:

PR #2228, opened 2019-05-08
[android] Implement Process and modify TestProcess to pass.

drodriguez:
Until recent API levels, Android haven't provide posix_spawn and other
functions from that family, so Process was not going to work with the
branches used for other POSIX OS.

Instead of creating an specific path for Android, which will probably be
forgotten by other changes, all the POSIX OS will use private CF
functions that wrap the posix_spawn functions. For Android, we try to
figure out if posix_spawn exist, and use it in that case, but  we
implement a subset of the posix_spawn family of functions for lower API
levels, using fork and exec underneath, and other functions from the
POSIX APIs. The already code in Process is only modified slightly to
adapt to the new names have to be modified.

The tests have to be modified slightly for working on Android. Many
changes are related to the environment needing the key LD_LIBRARY_PATH,
or the xdgTestHelper will not be able to start in some of the tests. The
other change is using NSTemporaryDirectory() instead of a hardcoded
`/tmp`, which doesn't exist in Android (the code also removes the last
slash, to match the original code expectations).

It also includes a small fix for when launch() is used and a error is
thrown which is not related to the current working directory. Before the
error was ignored silently, and now it will fatalError.

Other small improvement is reporting the actual line in the helper for
the returned POSIX error codes, instead of always reporting the line of
the helper itself.

Special thanks to @spevans because without the recent changes to use exclusively xdgTestHelper, the changes in the tests would have been more complicated.

### DIFF
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -7,7 +7,6 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-#if !os(Android) // not available
 import CoreFoundation
 
 extension Process {
@@ -287,6 +286,8 @@ open class Process: NSObject {
                 default:
                     fatalError("Process: The specified working directory cannot be set.")
                 }
+            } else {
+                fatalError(String(describing: nserror))
             }
         } catch {
             fatalError(String(describing: error))
@@ -661,7 +662,7 @@ open class Process: NSObject {
         }
 
         var taskSocketPair : [Int32] = [0, 0]
-#if os(macOS) || os(iOS)
+#if os(macOS) || os(iOS) || os(Android)
         socketpair(AF_UNIX, SOCK_STREAM, 0, &taskSocketPair)
 #else
         socketpair(AF_UNIX, Int32(SOCK_STREAM.rawValue), 0, &taskSocketPair)
@@ -738,14 +739,12 @@ open class Process: NSObject {
         let source = CFSocketCreateRunLoopSource(kCFAllocatorDefault, socket, 0)
         CFRunLoopAddSource(managerThreadRunLoop?._cfRunLoop, source, kCFRunLoopDefaultMode)
 
-        // file_actions
-        #if os(macOS) || os(iOS) || CYGWIN
-            var fileActions: posix_spawn_file_actions_t? = nil
-        #else
-            var fileActions: posix_spawn_file_actions_t = posix_spawn_file_actions_t()
-        #endif
-        posix(posix_spawn_file_actions_init(&fileActions))
-        defer { posix_spawn_file_actions_destroy(&fileActions) }
+        var fileActions = _CFPosixSpawnFileActionsAlloc()
+        posix(_CFPosixSpawnFileActionsInit(fileActions))
+        defer {
+            _CFPosixSpawnFileActionsDestroy(fileActions)
+            _CFPosixSpawnFileActionsDealloc(fileActions)
+        }
 
         // File descriptors to duplicate in the child process. This allows
         // output redirection to NSPipe or NSFileHandle.
@@ -820,10 +819,10 @@ open class Process: NSObject {
         }
 
         for (new, old) in adddup2 {
-            posix(posix_spawn_file_actions_adddup2(&fileActions, old, new))
+            posix(_CFPosixSpawnFileActionsAddDup2(fileActions, old, new))
         }
         for fd in addclose {
-            posix(posix_spawn_file_actions_addclose(&fileActions, fd))
+            posix(_CFPosixSpawnFileActionsAddClose(fileActions, fd))
         }
 
         let fileManager = FileManager()
@@ -839,7 +838,7 @@ open class Process: NSObject {
 
         // Launch
         var pid = pid_t()
-        guard posix_spawn(&pid, launchPath, &fileActions, nil, argv, envp) == 0 else {
+        guard _CFPosixSpawn(&pid, launchPath, fileActions, nil, argv, envp) == 0 else {
             throw _NSErrorWithErrno(errno, reading: true, path: launchPath)
         }
 
@@ -1027,11 +1026,10 @@ extension Process {
     public static let didTerminateNotification = NSNotification.Name(rawValue: "NSTaskDidTerminateNotification")
 }
     
-private func posix(_ code: Int32) {
+private func posix(_ code: Int32, file: StaticString = #file, line: UInt = #line) {
     switch code {
     case 0: return
-    case EBADF: fatalError("POSIX command failed with error: \(code) -- EBADF")
-    default: fatalError("POSIX command failed with error: \(code)")
+    case EBADF: fatalError("POSIX command failed with error: \(code) -- EBADF", file: file, line: line)
+    default: fatalError("POSIX command failed with error: \(code)", file: file, line: line)
     }
 }
-#endif
--- a/TestFoundation/TestProcess.swift
+++ b/TestFoundation/TestProcess.swift
@@ -7,7 +7,6 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-#if !os(Android)
 class TestProcess : XCTestCase {
     
     func test_exit0() throws {
@@ -21,7 +20,7 @@ class TestProcess : XCTestCase {
         }
         XCTAssertEqual(executableURL.path, process.launchPath)
         process.arguments = ["--exit", "0"]
-        try? process.run()
+        try process.run()
         process.waitUntilExit()
         
         XCTAssertEqual(process.terminationStatus, 0)
@@ -33,7 +32,7 @@ class TestProcess : XCTestCase {
         process.executableURL = xdgTestHelperURL()
         process.arguments = ["--exit", "1"]
 
-        try? process.run()
+        try process.run()
         process.waitUntilExit()
         XCTAssertEqual(process.terminationStatus, 1)
         XCTAssertEqual(process.terminationReason, .exit)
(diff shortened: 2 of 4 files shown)


drodriguez:
@swift-ci please test

TestFoundation/TestProcess.swift
```diff
@@ -155,6 +154,13 @@ class TestProcess : XCTestCase {
...
         // set in the subprocess.
         process.environment = [:]
+#if os(Android)
+        // In Android, we have to provide at least an LD_LIBRARY_PATH, or
+        // xdgTestHelper will not be able to find the Swift libraries.
+        if let ldLibraryPath = ProcessInfo.processInfo.environment["LD_LIBRARY_PATH"] {
+            process.environment?["LD_LIBRARY_PATH"] = ldLibraryPath
+        }
+#endif
         process.launch()
```
spevans: Is the `LD_LIBRARY_PATH` still needed? When I updated `xdgTestHelper` to contain all of the utilities I updated the `CMakeLists.txt` to add an `rpath` to the build directory which should apply to both Linux and Android:
`set(XDG_TEST_HELPER_RPATH -Xlinker;-rpath;-Xlinker;${CMAKE_CURRENT_BINARY_DIR})`

This was needed because one of the tests clears the environment before calling `xdgTestHelper --env` and before adding in the `rpath` it was failing as the `LD_LIBRARY_PATH` was not set anymore.

drodriguez: `${CMAKE_CURRENT_BINARY_DIR}` is a path in the build machine, but it doesn’t mean anything when moving the executable to the device for testing. That's also why this pieces are restricted to Android (they should not be needed for other targets).

Foundation CMakeLists is not prepared to test in Android, so the testing has to done manually, and while I always use the same directory, I would prefer not to hardcode anything in the CMakeLists for Android (I have to figure out how to simplify testing for Android).

Foundation/Process.swift
```diff
@@ -978,4 +979,127 @@ private func posix(_ code: Int32) {
     default: fatalError("POSIX command failed with error: \(code)")
     }
 }
+
+#if os(Android)
+// Android doesn't provide posix_spawn APIs until recent API level, so we cannot
```
millenomi: Do you want to delegate to the real API if they exist?

millenomi:
I'm okay with this patch otherwise.

Foundation/Process.swift
drodriguez: I don’t know if one can do that.

The availability of the symbol depends on the Android API level. By default, if nobody sets a different `__ANDROID_API__` while compiling, the latest API level is used. Swift sets API level 21, but Foundation doesn't set anything, so at the moment, Foundation is being compiled with access to every available symbol in the Android NDK (I'm not 100% sure about that, though, because I don’t know if the module cache remembers the setting from Swift). So the symbol will be available at compile time, but it might not be at runtime (and the binary will probably fail to start).

I don’t think `#available/@available` works for Android, which might be a good way, so I think this is the best I can come up now.

Foundation/Process.swift
```diff
@@ -978,4 +979,127 @@ private func posix(_ code: Int32) {
...
+                         _ envp: UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>?) -> Int32 {
+    // TODO: We completely ignore attrp, because at the moment, the only
+    // invocation doesn't pass a value.
+
+    // Block signals during this fork/execv dance.
+    var signalSet = sigset_t()
+    sigfillset(&signalSet)
+    var oldmask = sigset_t()
+    guard sigprocmask(SIG_BLOCK, &signalSet, &oldmask) == 0 else {
+        fatalError("sigprocmask() call failed: \(errno)")
```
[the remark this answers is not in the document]
drodriguez: [sigprocmask](https://pubs.opengroup.org/onlinepubs/7908799/xsh/sigprocmask.html) should not fail in this case, since the only failure case is an incorrect first parameter, which is hardcoded here to a valid value. If the code ends up in the `fatalError`, something has gone really wrong, and aborting is the right thing to do, in my opinion.

Foundation/Process.swift
```diff
@@ -978,4 +979,127 @@ private func posix(_ code: Int32) {
...
+        }
+    }
+
+    // This is the child.
+
+    // Clean up the parent signal handlers
+    for idx in 1..<(NSIG as Int32) {
+        var sigAction = sigaction()
+        guard sigaction(idx, nil, &sigAction) == 0 else {
+            fatalError("sigaction(\(idx), ...) call failed: \(errno)")
```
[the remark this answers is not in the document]
drodriguez: It can be an option. The [standard](http://pubs.opengroup.org/onlinepubs/007904875/functions/posix_spawn.html) is not perfectly clear. I will change it.

I have also found that the first device I was using seem to follow Linux interpretation of the standard and ignores SIGKILL and SIGSTOP, but another OS version decides to return EINVAL for this call. I will also change that.

Foundation/Process.swift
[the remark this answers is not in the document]
spevans: Ah right, I see `sigprocmask` only returns `EINVAL` and `EFAULT` as errors which does imply a programming error.

drodriguez:
Added the feedback from Simon.

@swift-ci please test

buttaface:
Tested this with the 5/12 snapshot, Process tests all passing now, whereas before I had to put in more workarounds just to avoid those tests on Android AArch64:
```
diff --git a/TestFoundation/TestProcess.swift b/TestFoundation/TestProcess.swift
index f3ba630f..b38d8b26 100644
--- a/TestFoundation/TestProcess.swift
+++ b/TestFoundation/TestProcess.swift
@@ -514,6 +514,7 @@ private enum Error: Swift.Error {
     case InvalidEnvironmentVariable(String)
 }
 
+#if !os(Android)
 // Run xdgTestHelper, wait for 'Ready' from the sub-process, then signal a semaphore.
 // Read lines from a pipe and store in a queue.
 class _SignalHelperRunner {
@@ -591,7 +592,7 @@ class _SignalHelperRunner {
         return true
     }
 }
-
+#endif
 #if !os(Android)
 internal func runTask(_ arguments: [String], environment: [String: String]? = nil, currentDirectoryPath: String? = nil) throws -> (String, String) {
     let process = Process()
@@ -669,5 +670,9 @@ private func parseEnv(_ env: String) throws -> [String: String] {
     }
     return result
 }
+#else
+internal func runTask(_ arguments: [String], environment: [String: String]? = nil, currentDirectoryPath: String? = nil) throws -> (String, String) {
+return ("", "")
+}
 #endif
 ```

**Comment by drodriguez** [comment]
Rebased on top of master and removed the conflict. This change do not include any changes to the `RPATH` for Android (it still uses `LD_LIBRARY_PATH` tricks instead which should not affect other platforms).

@swift-ci please test

**Comment by drodriguez** [comment]
@swift-ci please test

File: Foundation/Process.swift (hunk 2 above)
Comment (compnerd) [reply Foundation/Process.swift]: I think that it makes sense to pass along the `__ANDROID_API__` when building for Android.  Using `posix_spawn` is preferable to rolling our own.

File: Foundation/Process.swift [hunk 5]
```diff
@@ -978,4 +979,133 @@ private func posix(_ code: Int32) {
...
+        // This is the parent. fork call might have been successful or not.
+
+        // Unblock signals.
+        guard sigprocmask(SIG_SETMASK, &oldmask, nil) == 0 else {
+            fatalError("sigprocmask() call failed: \(errno)")
+        }
+
+        if pid < 0 {
+            return pid
+        } else {
```
compnerd: The `else` is unnecessary.  Just drop the `else` and unindent the body.

Foundation/Process.swift
```diff
@@ -978,4 +979,133 @@ private func posix(_ code: Int32) {
...
+            }
+            return 0
+        }
+    }
+
+    // This is the child.
+
+    // Clean up the parent signal handlers
+    for idx in 1..<(NSIG as Int32) {
+        // Seems that SIGKILL/SIGSTOP are sometimes silently ignored, and
```
compnerd: Odd, is that a documented issue on android?

Foundation/Process.swift
drodriguez: I don't understand where you exactly want to pass `__ANDROID_API__`. Swift definitions are just booleans, and there's no support for something like `#if __ANDROID_API__ >= 21`, IIRC.

We might be able to do those checks in CoreFoundation, but again `__ANDROID_API__` is not a runtime check, but a "minimum deployment target" compilation check. When compiling you are deciding which is the minimal API you want to deploy to. As the code is written, we are supporting API levels before 26. If I use posix_spawn (and link against it), then we are not supporting anything before level 26.

That's my understanding, unless someone has another theory.

Foundation/Process.swift
drodriguez: Not that I know of. I found it doing trial and error. Some device was fine, some other was crashing here.

In https://pubs.opengroup.org/onlinepubs/009695399/functions/sigaction.html there's discussion that they should be ignored silently, but it also talks about SysV and BSD returning EINVAL. I think the problem is that, in the code I found for Bionic, it seems to delegate to Linux `rt_sigaction`, which is documented to fail for those two signals.

In any case, since `SIGKILL` and `SIGSTOP` cannot be ignored or caught, they are already set to `SIG_DFL`, which is what this code will try to do. Skipping them should not have ill effects.

drodriguez:

drodriguez:
All right. I rewrote all the code in a way I hope matches the requests above. Instead always using a custom `posix_spawn` for Android, if the real one is available during runtime (checking with `dlsym`), it will be used instead. This meant a little bit of wrangling in the code for other POSIX OS, but the code path for them is mostly a thin wrapper around the real functions.

@compnerd: can you review if this is what you wanted?

drodriguez:
Rebased to avoid conflicts with master. Ready for review.

drodriguez:
Failures in LLDB are unrelated.

PR #2231, opened 2019-05-08
Parity: Bundle(for: AnyClass)

millenomi:
- Use dladdr() with a symbol guaranteed to be in the source image of the class.

 - We cannot replace the Bundle object init() returns, but we can implement equality correctly so that at least `==` can compare bundles correctly if they refer to the same CFBundle.

 - Add testExpectedToFailOnWindows() and associated infrastructure.


millenomi:
@swift-ci please test

millenomi:
Fixes https://bugs.swift.org/browse/SR-10361

millenomi:
@swift-ci please test

spevans:
@swift-ci please test

PR #2232, opened 2019-05-08
Parity: NSCoding: NSOrderedSet

millenomi:
- NSOrderedSet uses a subset of the way NSSet encodes things; encode using the existing methods.

- We’re using convenience initializers to mimic “/* not inherited */“ factory initializers from ObjC. Just like those initializers, they aren’t actually inherited. We need to replicate all the convenience initializers of NSOrderedSet in NSMutableOrderedSet for source compatibility, and we do so here.


millenomi:
@swift-ci please test

millenomi:
Fixes https://bugs.swift.org/browse/SR-10404

millenomi:

millenomi:

PR #2235, opened 2019-05-09
Foundation: tweak address resolution on Windows

compnerd:
Rewrite the host address resolution to be easier to follow.  The old
path obfuscated pointers to the point where the invalid memory access
could not be found (probably due to the incorrect handling with the
unsafe pointer).  This now passes the Host tests.


compnerd:
@swift-ci please test

PR #2236, opened 2019-05-09
TestFoundation: fix a double free

compnerd:
No allocation was made to perform the explicit deallocation.  Doing so
causes a double free which the Windows C library caught.


compnerd:
CC: @millenomi

compnerd:
@swift-ci please test

compnerd:

compnerd:
@millenomi and I discussed this offline.  This is correct, the memory is owned by the `NSValue`, not by the consumer, so this is an invalid free.  Going to merge this so that I can make a follow up change - @millenomi caught the fact that we have a lifetime problem here and need to extend the lifetime of the value to ensure that there is no use-after-free.

PR #2237, opened 2019-05-09
TestFoundation: make TestPipe pass on Windows

compnerd:
Windows does not provide `fileDescriptor` for file handles as it uses
HANDLEs underneath and it is not possible to provide a non-owning file
descriptor value.  Make this contingent on testable imports and use the
internal `_isPlatformHandleValid` method to check instead.


compnerd:
CC: @millenomi @spevans

compnerd:
@swift-ci please test

PR #2238, opened 2019-05-09
Duplicate the Handle instead of the fd on Windows

gmittert:
Since getting an fd from a handle transfers ownership, we duplicate the
handle before getting an fd from it.


gmittert:
@swift-ci please test

PR #2239, opened 2019-05-09
Add Support for '\' in Windows Paths

gmittert:
While URLs get converted to forward slashes on Windows, strings are
handled as is and thus need to be able to handle \ separators. Since
many Windows apis also support /, we'll handle that as well, but default
to \ if required.

### DIFF
--- a/Foundation/NSURL.swift
+++ b/Foundation/NSURL.swift
@@ -957,6 +957,8 @@ extension NSURL {
     
     open func appendingPathComponent(_ pathComponent: String) -> URL? {
         var result : URL? = appendingPathComponent(pathComponent, isDirectory: false)
+        // Since we are appending to a URL, path seperators should
+        // always be '/', even if we're on Windows
         if !pathComponent.hasSuffix("/") && isFileURL {
             if let urlWithoutDirectory = result {
                 var isDir: ObjCBool = false
--- a/Foundation/NSPathUtilities.swift
+++ b/Foundation/NSPathUtilities.swift
@@ -9,6 +9,12 @@
 
 import CoreFoundation
 
+#if os(Windows)
+let validPathSeps: [Character] = ["\\", "/"]
+#else
+let validPathSeps: [Character] = ["/"]
+#endif
+
 public func NSTemporaryDirectory() -> String {
 #if os(Windows)
     let cchLength: DWORD = GetTempPathW(0, nil)
@@ -51,7 +57,7 @@ public func NSTemporaryDirectory() -> String {
     }
 #endif
     if let tmpdir = ProcessInfo.processInfo.environment["TMPDIR"] {
-        if !tmpdir.hasSuffix("/") {
+        if !validPathSeps.contains(where: { tmpdir.hasSuffix(String($0)) }) {
             return tmpdir + "/"
         } else {
             return tmpdir
@@ -70,7 +76,7 @@ public func NSTemporaryDirectory() -> String {
 extension String {
     
     internal var _startOfLastPathComponent : String.Index {
-        precondition(!hasSuffix("/") && length > 1)
+        precondition(!validPathSeps.contains(where: { hasSuffix(String($0)) }) && length > 1)
         
         let startPos = startIndex
         var curPos = endIndex
@@ -78,7 +84,7 @@ extension String {
         // Find the beginning of the component
         while curPos > startPos {
             let prevPos = index(before: curPos)
-            if self[prevPos] == "/" {
+            if validPathSeps.contains(self[prevPos]) {
                 break
             }
             curPos = prevPos
@@ -88,7 +94,7 @@ extension String {
     }
 
     internal var _startOfPathExtension : String.Index? {
-        precondition(!hasSuffix("/"))
+        precondition(!validPathSeps.contains(where: { hasSuffix(String($0)) }))
 
         var currentPosition = endIndex
         let startOfLastPathComponent = _startOfLastPathComponent
@@ -97,7 +103,7 @@ extension String {
         while currentPosition > startOfLastPathComponent {
             let previousPosition = index(before: currentPosition)
             let character = self[previousPosition]
-            if character == "/" {
+            if validPathSeps.contains(character) {
                 return nil
             } else if character == "." {
                 if startOfLastPathComponent == previousPosition {
@@ -132,8 +138,8 @@ extension String {
         if isEmpty {
             return str
         }
-        if hasSuffix("/") {
-            return self + str
+        if validPathSeps.contains(where: { hasSuffix(String($0)) }) {
+          return self + str
         }
         return self + "/" + str
     }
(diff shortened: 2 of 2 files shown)


gmittert:
@swift-ci please test

gmittert:

spevans:
@gmittert This seems to allow paths containing both `/` and `\` as seperators at the same time, eg `\foo/bar/file.txt`. Is that valid or should there be a check to ensure only one type is used throughout the path?

gmittert:
> @gmittert This seems to allow paths containing both `/` and `\` as seperators at the same time, eg `\foo/bar/file.txt`. Is that valid or should there be a check to ensure only one type is used throughout the path?

That's intentional. Most Windows APIs are able to handle both / and \ but will canonicalize / to \. It does paint over. e.g. `C:\Users/gmittertreiner\foo/bar` is valid to pass to CreateFileW and will do the right thing. It only becomes an issue if you use the \\?\ paths, which will prevent the canonicalization. However, I think we only do that in URL which any thing passed to it should already get canonicalized to \ on Windows.

It potentially paints over areas that are doing things like concatting paths with forward slashes, but since Windows supports the path style, I think that's for the better.

gmittert:
cc @millenomi

lxbndr:
I am eagerly awaiting for this PR, looks very promising. Maybe I am missing something, but is it ok to allow Windows path separator in `NSURL`? Looks like NSURL should correctly handle _initialization_ with filesystem path, but result URL string should contain separators defined in RFC. I guess `path` property of `NSURL`&`URL` also have to follow RFC rules, because it is not an issue for Windows `file:` URLs to have `/` separators.

gmittert:
@lxbndr  Based on how [path is implemented](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSURL.swift#L551) it'll give you a file system path (e.g. `C:\Users\gwenm\AppData\Local\Temp\` on Windows) as opposed to a `file:\\`). Then since NSURL uses its `.path` internally, it unfortunately does have to handle the `\` separators.

gmittert:
Rebased. @swift-ci please test for me!

gmittert:
Rebased.

@swift-ci please test

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test macos platform

lxbndr:
@gmittert Right, implementation of `path` is important for `lastPathComponent`. Actually I am here because I was looking for pending solution of broken lastPathComponent. What looks wrong to me is that URL API threats URL path as filesystem path. I see two ways of fixing that. 
1. Force `NSURL.path` always use POSIX path separators. This will make `lastPathComponent` work. And follows [RFC](https://tools.ietf.org/html/rfc8089#appendix-E.4).
2. Make path components related code work with both slashes.

I did rough implementation of first solution, and it kinda works. Does it make sense? Still not ready for PR yet, but if it sounds good, then guess I able to finish it - look for side effects, check tests, etc.

gmittert:
@lxbndr  Oh I see, NSURL should return a `file://` style url and it doesn't right now. That does seem very wrong.

The PathUtilities changes need to go in anyways, but I could drop the NSURL changes I have ones and then you can PR your patch.

Alternatively, I can fix it this patch and make sure it fixes the same FileManager tests I was aiming to fix.

Does either work better for you?

lxbndr:
@gmittert I believe it would be faster to implement decent fix with your experience. If your current schedule allows, can you please take a time for this? If not, I always glad to contribute, first approach PR could be ready next days after this #2239 merged.

gmittert:
Alright, removed most of the NSURL changes except for one which has handling input paths.

I've then made URL.path return an RFC1808 in https://github.com/apple/swift-corelibs-foundation/pull/2366 and https://github.com/apple/swift-corelibs-foundation/pull/2367

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

Foundation/NSURL.swift
```diff
@@ -957,7 +957,7 @@ extension NSURL {
     
     open func appendingPathComponent(_ pathComponent: String) -> URL? {
         var result : URL? = appendingPathComponent(pathComponent, isDirectory: false)
-        if !pathComponent.hasSuffix("/") && isFileURL {
+        if !validPathSeps.contains(where: { pathComponent.hasSuffix(String($0)) }) && isFileURL {
```
millenomi: So;

Foundation distinguishes between paths (in `NSString`s) and `URL`s. The latter have harmonized behavior between all platforms, whereas the former can contain a platform path and can thus accommodate specialized platform behaviors.

Trying to append a `\` to an URL is a programmer error regardless of platform. URL paths are specified to use `/` as separators. Thus: I would back off this line and I'm okay taking the rest.

millenomi:
The only exceptions I make to this contract are those already documented in `CFURL`.

Foundation/NSURL.swift
millenomi: The only bit that puts a wrench into this particular getup is that the result of `URL`'s `.path` _must_ be useable as a platform path, which can get complicated.

lxbndr: > The only bit that puts a wrench into this particular getup is that the result of `URL`'s `.path` _must_ be useable as a platform path, which can get complicated.

Looks like `/` path separator will meet this requirement for all platforms, as Windows is (more or less) loyal to both variants. But can you please add more on this "must" part? I always thought that URL/CFURL api clearly distinguish between _file system_ paths and URL paths. Now I feel like I need to reevaluate this for myself.

gmittert: > The only bit that puts a wrench into this particular getup is that the result of `URL`'s `.path` _must_ be useable as a platform path, which can get complicated.

 I see, that is complicated. What are your thoughts on handling absolute URL paths then? On Windows, paths which begin with `/` or `\` are drive relative. That is, if my cwd is `C:\` then `/Users/gwenm` resolves to `C:\Users\gwenm`.

Unfortunately, `/C:/Users/gwenm` is not a valid windows path, so I think there really only two reasonable things to return from `URL(fileURLWithPath: "C:\\Users\\gwenm").path` on Windows:
- `/C:/Users/gwenm` which, while not a valid Windows path, we would at least be able to add support for through out Foundation so that e.g. `"foo".write(to: myURL.path, ...)` would work.
- `C:/Users/gwenm` which isn't a proper RFC 1808 path as specified to be returned by Foundation, as an absolute path _must_ start with `/`,  but would work directly with Windows apis (though would not be able to be copied and pasted into cmd.exe and directly used without first being quoted).

gmittert:
Removed the check for backslashes in NSURL and left a comment on why it's not needed instead.

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

gmittert:
@millenomi One week courtesy ping

gmittert:
@millenomi Courtesy ping :)

compnerd:
I believe that Lily's comments have been addressed.  I am going to merge this to get s-p-m on Windows working.

PR #2241, opened 2019-05-09
TestFoundation: ensure that `NSValue` is preserved

compnerd:
Extend the lifetime of the `NSValue` object as we retrieve a pointer to
the data owned by it.  Since the data is owned by the `NSValue`
instance, we need to ensure that the `NSValue` has its lifetime extended
throughout the usage.


compnerd:
@swift-ci please test

PR #2242, opened 2019-05-09
TestFoundation: construct `FileHandle` properly on Windows

compnerd:
This fixes the `FileHandle` construction on Windows so that we can
continue to execute more of the tests.

### DIFF
--- a/TestFoundation/TestFileHandle.swift
+++ b/TestFoundation/TestFileHandle.swift
@@ -10,6 +10,9 @@
 #if !DARWIN_COMPATIBILITY_TESTS // Disable until Foundation has the new FileHandle API
 
 import Dispatch
+#if os(Windows)
+import WinSDK
+#endif
 
 class TestFileHandle : XCTestCase {
     var allHandles: [FileHandle] = []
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

TestFoundation/TestFileHandle.swift
```diff
@@ -101,9 +104,28 @@ class TestFileHandle : XCTestCase {
...
+                        DWORD(FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE),
+                        nil, DWORD(OPEN_EXISTING),
+                        DWORD(FILE_ATTRIBUTE_NORMAL | FILE_FLAG_BACKUP_SEMANTICS), nil)
+        }
+        if hDirectory == INVALID_HANDLE_VALUE {
+          fatalError("unable to create handle to current directory")
+        }
+        let fd = _open_osfhandle(intptr_t(bitPattern: hDirectory), 0)
+        if fd == -1 {
+          fatalError("unable to associate file descriptor with handle")
```
spevans: Wouldn't it be better to make this function return a `FileHandle?` so this can `return nil` and  be `XCTFail`ed by the caller, rather than stopping all of the tests?

compnerd: I followed what the old path was doing (fatal error'ing).  I do like that, and since you asked, I'll go ahead and do that.

compnerd: Actually, lets do that in a follow up, I have a huge patch stack locally for the `DEPLOYMENT_TARGET_*` to `TARGET_OS_*` and that is making things more complicated.

compnerd: I took a look at this and this is really ugly without some of the internal helpers (i.e. `_NSErrorWithErrno` and `_NSErrorWithWindowsError`).  Do you want to convert this to a testable import test?  CC: @millenomi

compnerd:
ping?

TestFoundation/TestFileHandle.swift
millenomi: Note that `expectTrue()` below doesn't invoke `fatalError()`.

compnerd: I'm not sure if we can do anything else here.  Its an internal error because this implies that the handle that was provided was invalid.  It would indicate an error in the test itself.

millenomi: I'm okay with that.

millenomi: Meaning: I'm okay error'ing out if the text is executing such that '.' is not an accessible directory for reading.

PR #2243, opened 2019-05-09
TestFoundation: make `test_nullDevice` test more portable

compnerd:
This makes test_nullDevice more portable so that we can run the test
on Windows as well.

### DIFF
--- a/TestFoundation/TestFileHandle.swift
+++ b/TestFoundation/TestFileHandle.swift
@@ -281,10 +281,11 @@ class TestFileHandle : XCTestCase {
                        "\(FileHandle.readCompletionNotification.rawValue) is not equal to NSFileHandleReadCompletionNotification")
     }
 
+#if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
     func test_nullDevice() {
         let fh = FileHandle.nullDevice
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:

millenomi:
This needs conflict resolution before I can merge.

spevans:
Doesnt this need a 
```
    #if canImport(SwiftFoundation) && !DEPLOYMENT_RUNTIME_OBJC
    #else
        @testable import Foundation
    #endif
#endif
```

At the top if you are using `NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT` ?

compnerd:
That is already present now.  So, this is ready to go as is.

compnerd:
@swift-ci please test

spevans:
It needs to be in every file that does a `@testable import` - I just tested this PR on xcode and it wont compile otherwise.

compnerd:
@spevans - huh?  TestFIleHandle already has the `@testable import`: https://github.com/apple/swift-corelibs-foundation/blob/master/TestFoundation/TestFileHandle.swift#L12-L18

compnerd:
@swift-ci please test

PR #2244, opened 2019-05-09
TestFoundation: make `test_fileDescriptor` more portable

compnerd:
This adjusts the fileDescriptor test to be more portable so that we can
run it on Windows.


spevans:
@swift-ci test

PR #1789, opened 2018-11-28
plutil: Fix RUNPATH and file permissions of installed binary

spevans:

PR #2246, opened 2019-05-09
[cmake] Add runpath for libBlocksRuntime.so

drodriguez:
Dispatch build leaves artifacts in both the root directory and the src
directory, but only the src directory was being added to the runpath, so
the libBlocksRuntime.so library might not have been found.

Additionally, remove the rpath flags from libdispatch_ldflags, which are
used to compile Foundation, but will point to the build directory
results. Add those flags instead to a list used for the test binaries.

In the end Foundation ends up with a path of `$ORIGIN`; TestFoundation
has paths to Foundation, to Dispatch and BlocksRuntime, and to XCTest;
while xdgTestHelper points to Foundation, and Dispatch and
BlocksRuntime.

Extract a small function for adding rpaths to a list, to make the rpath
easier to read.

Also, remove Android from the rpath flags, since Android will not work
with the build tree rpaths, and will need its own solution. For the time
being, do not set any rpath.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,18 +92,27 @@ set(libdispatch_ldflags)
 if(FOUNDATION_ENABLE_LIBDISPATCH)
   set(libdispatch_cflags -I;${FOUNDATION_PATH_TO_LIBDISPATCH_SOURCE};-I;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src/swift;-Xcc;-fblocks)
   set(libdispatch_ldflags -L;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD};-L;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src;-ldispatch;-lswiftDispatch)
-  if(CMAKE_SYSTEM_NAME STREQUAL Linux OR CMAKE_SYSTEM_NAME STREQUAL Android OR CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
-    file(TO_CMAKE_PATH "${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}" FOUNDATION_PATH_TO_LIBDISPATCH_BUILD)
-    list(APPEND libdispatch_ldflags -Xlinker;-rpath;-Xlinker;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src)
-  endif()
 endif()
 
-set(plutil_rpath)
-if(CMAKE_SYSTEM_NAME STREQUAL Android OR CMAKE_SYSTEM_NAME STREQUAL Linux OR
-    CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
-  set(Foundation_RPATH -Xlinker;-rpath;-Xlinker;"\\\$\$ORIGIN")
-  set(XDG_TEST_HELPER_RPATH -Xlinker;-rpath;-Xlinker;${CMAKE_CURRENT_BINARY_DIR})
-  set(plutil_rpath -Xlinker;-rpath;-Xlinker;"\\\$\$ORIGIN/../lib/swift/${swift_os}")
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux OR CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
+  append_linker_rpath(Foundation_RPATH "\\\$\$ORIGIN")
+  append_linker_rpath(TestFoundation_RPATH ${CMAKE_CURRENT_BINARY_DIR})
+  append_linker_rpath(TestFoundation_RPATH ${FOUNDATION_PATH_TO_XCTEST_BUILD})
+  append_linker_rpath(XDG_TEST_HELPER_RPATH ${CMAKE_CURRENT_BINARY_DIR})
+  append_linker_rpath(plutil_rpath "\\\$\$ORIGIN/../lib/swift/${swift_os}")
+  if (FOUNDATION_ENABLE_LIBDISPATCH)
+    file(TO_CMAKE_PATH "${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}" FOUNDATION_PATH_TO_LIBDISPATCH_BUILD)
+    # NOTE: the following two are for testing LLDB and others, but should be removed
+    append_linker_rpath(Foundation_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src)
+    append_linker_rpath(Foundation_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD})
+    append_linker_rpath(TestFoundation_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src)
+    append_linker_rpath(TestFoundation_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD})
+    append_linker_rpath(XDG_TEST_HELPER_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src)
+    append_linker_rpath(XDG_TEST_HELPER_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD})
+  endif()
+elseif(CMAKE_SYSTEM_NAME STREQUAL Android)
+  append_linker_rpath(Foundation_RPATH "\\\$\$ORIGIN")
 elseif(CMAKE_SYSTEM_NAME STREQUAL Windows)
   # FIXME(SR9138) Silence "locally defined symbol '…' imported in function '…'
   set(WORKAROUND_SR9138 -Xlinker;-ignore:4049;-Xlinker;-ignore:4217)
@@ -535,6 +544,7 @@ if(ENABLE_TESTING)
                          -L${FOUNDATION_PATH_TO_XCTEST_BUILD}
                          -lXCTest
                          ${WORKAROUND_SR9138}
+                         ${TestFoundation_RPATH}
                        RESOURCES
                          ${CMAKE_SOURCE_DIR}/TestFoundation/Resources/Info.plist
                          ${CMAKE_SOURCE_DIR}/TestFoundation/Resources/NSURLTestData.plist
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -273,3 +273,20 @@ function(get_swift_host_arch result_var_name)
     message(FATAL_ERROR "Unrecognized architecture on host system: ${CMAKE_SYSTEM_PROCESSOR}")
   endif()
 endfunction()
+
+
+# Appends an the given path as an rpath prefixed by the -Xlinker arguments for
+# the Swift compiler to pass to the linker.
+#
+# Usage:
+#   append_linker_rpath(MyTarget_RPATH one)
+#   append_linker_rpath(MyTarget_RPATH two)
+#   # MyTarget_RPATH=-Xlinker;-rpath;-Xlinker;one;-Xlinker;-rpath;-Xlinker;two
+#
+function(append_linker_rpath result_var_name rpath)
+  set(tmp ${${result_var_name}})
+
+  list(APPEND tmp -Xlinker -rpath -Xlinker "${rpath}")
+
+  set(${result_var_name} ${tmp} PARENT_SCOPE)
+endfunction()


spevans:
Which binary is this setting the `RPATH` for ?

I don't think its needed because its covered by the existing `LD_LIBRARY_PATH`?

drodriguez:
In my machine `xdgTestHelper` ended up with:

```
$ readelf -d xdgTestHelper
…
0x000000000000001d RUNPATH              /data/users/danielrodriguez/scratch/swift/build/my_linux/swift-linux-x86_64/lib/swift/linux:/data/users/danielrodriguez/scratch/swift/build/my_linux/libdispatch-linux-x86_64/src:/home/danielrodriguez/scratch/swift/build/my_linux/foundation-linux-x86_64
…
```

When I tried to run the `TestProcess` tests, the ones that remove all the environment were failing because they couldn’t find `libBlocksRuntime.so`, which sits in `/data/users/danielrodriguez/scratch/swift/build/my_linux/libdispatch-linux-x86_64` (without the `src/`).

The change allows `xdgTestHelper` to find `libBlocksRuntime.so` while running those tests without environment.

(This is unrelated to Android, this is a Linux machine)

spevans:
Oh I see, it works normally on Linux because it is using the system installed `libBlocksRuntime`

```
$ ldd  build/buildbot_linux/foundation-linux-x86_64/TestFoundation/xdgTestHelper |grep Blocks
	libBlocksRuntime.so => /usr/lib/x86_64-linux-gnu/libBlocksRuntime.so (0x00007f851fdc6000)
$ dpkg -l |grep libblocks
ii  libblocksruntime-dev                  0.4.1-1.1                              amd64        Blocks Runtime (development files)
ii  libblocksruntime0                     0.4.1-1.1                              amd64        Blocks Runtime library
```

I guess it depends if we always want to use the one built as part of libdispatch and ignore the system one.

drodriguez:
Your choice to merge or not. It was something I found in my setup. The Android runpath problem is more convoluted.

compnerd:
This really is going down the wrong path.  We should be reducing `RPATH` dependencies, not increasing them.  They are a security concern and increase difficulty in debugging.  I think that relying more on `$ORIGIN/../lib` is the right approach with all the libraries being co-located.

spevans:
If this PR is just adjusting the `rpath` for the test programs, arent those paths already set in the `LD_LIBRARY_PATH` environment variable? Maybe one thing we should do is adjust the `libdispatch` build to put all of the `.so` files into the same build directory so we can prune one more directory from library paths.

kevints:
@spevans we should never be using the system libBlocksRuntime (see the discussion on https://bugs.swift.org/browse/SR-9347) so anything still relying on it is a bug. Whether we use RPATHs or LD_LIBRARY_PATH to do that doesn't matter too much for a test program, but for binaries that are installed onto the system we should prefer to use RPATH.

We should also not be including build tree RPATHs in the installed binaries. In cmake that should be possible with https://cmake.org/cmake/help/v3.4/variable/CMAKE_INSTALL_RPATH.html, which will cause the cmake `install` or `package` step to update the RPATH field of the binary before copying it.

millenomi:
@kevints I'm okay with RPATHs in this instance because these are test binaries which won't be distributed.

drodriguez:
The variable is also used while compiling `Foundation`, and indeed points to the build directory. It works because during testing the build directory is still there (it happens before installing).

I can (gladly) abandon this PR, but the rest of the RPATHs are still pointing to similar paths.

spevans:
@drodriguez Which binaries are affected by this PR, is it just `TestFoundation` and `xdgTestHelper` or libraries as well?

drodriguez:
This PR modified `libdispatch_ldflags`, because that variable was the one used for setting the RPATH pointing to `libdispatch.so`/`libswiftdispatch.so` (in the build directory).

That variable is used while building `libFoundation.so`, `plutil`, `xdgTestHelper` and `TestFoundation`.

One can move this change to another variable, but that doesn’t change the fact that the variable was applying the RPATH to all the artifacts and pointing to the build directory.

millenomi:
Hmm. @spevans is correct that we shouldn't ship with build machine paths for at least some of the products — `libFoundation.so` and `plutil` should not carry them. They're fine to a point in `TestFoundation` and the helper tool.

spevans:
You could just add the extra directory into `XDG_TEST_HELPER_RPATH` which is used by `xdgTestHelper` to set its `rpath`. This could also be used by `TestFoundation` (you may want to rename this variable to `TEST_BINARY_HELPER_RPATH` or something as well). This would fix the test binaries and leave the libs alone.
In fact this is probably better than what we currently have as on Ubuntu CI we currently run the tests against the system `libBlocksRuntime.so` and not the one that is built as part of `libdispatch` so this should improve the testing. 

Would this change allow the removal of the `LD_LIBRARY_PATH` set in `TestProcess` explicitly for Android?

drodriguez:
I can change this PR to work for Linux at least. Android is a special creature. Maybe '$ORIGIN' works there, but I have to check.

drodriguez:
I think this should work in anybody Linux installation. Only the test binaries have extra rpaths. Also, Android is left without them for the time being. I will do it in another commit.

@swift-ci please test

drodriguez:
@swift-ci please test linux

buttaface:
>We should also not be including build tree RPATHs in the installed binaries.

They are still being included in the official releases for linux, [as I noted to @drodriguez in the comments for a different pull a couple months ago](https://github.com/apple/swift/pull/23208#discussion_r268391141). I see that it's still the case for libXCTest.so and other shared libraries in the 5.0 release, along with binaries like plutil, swift-build, or swift-package:
```
> readelf -d swift-5.0-RELEASE-ubuntu18.04/usr/bin/{plutil,swift-build,swift-package}|grep -E "File:|RUNPATH"
File: swift-5.0-RELEASE-ubuntu18.04/usr/bin/plutil
 0x000000000000001d (RUNPATH)            Library runpath: [/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/libdispatch-linux-x86_64/src:$ORIGIN/../lib/swift/linux]
File: swift-5.0-RELEASE-ubuntu18.04/usr/bin/swift-build
 0x000000000000001d (RUNPATH)            Library runpath: [/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:$ORIGIN:$ORIGIN/../lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/swiftpm-linux-x86_64/.bootstrap/lib/swift/linux:$ORIGIN/../lib/swift/pm/llbuild:/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/llbuild-linux-x86_64/lib]
File: swift-5.0-RELEASE-ubuntu18.04/usr/bin/swift-package
 0x000000000000001d (RUNPATH)            Library runpath: [/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:$ORIGIN:$ORIGIN/../lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/swiftpm-linux-x86_64/.bootstrap/lib/swift/linux:$ORIGIN/../lib/swift/pm/llbuild:/home/buildnode/jenkins/workspace/oss-swift-5.0-package-linux-ubuntu-18_04/build/buildbot_linux/llbuild-linux-x86_64/lib]
```

drodriguez:
Yes. That discussion points to the source of all those invalid runpaths, which come from the compiler itself embedding a full path of the build machine. Very nice when you are running those binaries in the build machine, useless for other cases.

This PR only deals with the test binaries, when run against the non-installed libraries. `libFoundation.so` still has `$ORIGIN` (which is not helpful while not installed, but it is when installed), and `plutil` still points to `$ORIGIN` (which is not correct in either case, and I don’t know in the installed result where it gets `$ORIGIN/../lib/swift/linux`, which is correct).

The new commit just fixes some conflicts, but doesn’t try to do anything more. I think this PR might still be interesting by itself, specially since it deals with testing the right setup.

drodriguez:
@swift-ci please test

spevans:
@drodriguez So does this PR fix the `RUNPATH` for only `TestFoundation` and `xdgTestHelper` to include the build directories and doesnt touch anything else ? If that is the case Im fine with the PR.

Note that currently `plutil` in the snapshot release is broken. It should have a `RUNPATH` of `$ORIGIN/../lib/swift/linux` (or whatever os) and was originally fixed in #1789 but has since been broken by updates to the `CMakeLists.txt`.

```
$ chrpath -l ~/swift-DEVELOPMENT-SNAPSHOT-2019-05-18-a-ubuntu18.04/usr/bin/plutil 
/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-05-18-a-ubuntu18.04/usr/bin/plutil: RUNPATH=/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/libdispatch-linux-x86_64/src:$ORIGIN
$ ~/swift-DEVELOPMENT-SNAPSHOT-2019-05-18-a-ubuntu18.04/usr/bin/plutil 
/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-05-18-a-ubuntu18.04/usr/bin/plutil: error while loading shared libraries: libswiftGlibc.so: cannot open shared object file: No such file or directory
```
 So that also needs fixing.

`libFoundation.so` should only have `$ORIGIN` but has the build paths as well:

```
$ chrpath -l ~/swift-DEVELOPMENT-SNAPSHOT-2019-05-18-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so 
/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-05-18-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so: RUNPATH=/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/libdispatch-linux-x86_64/src:$ORIGIN
```

So we need to scrub those as well.

We should really add a test into the `swift-intergration-tests` repo to check these `RUNPATH`s after install to stop extra directories getting into the shipped binaries.

I did open https://bugs.swift.org/browse/SR-9315 about the `RUNPATH`s but it was closed (incorrectly in my opinion).

drodriguez:
> @drodriguez So does this PR fix the `RUNPATH` for only `TestFoundation` and `xdgTestHelper` to include the build directories and doesnt touch anything else ? If that is the case Im fine with the PR.

Correct. It removes the dispatch paths from `Foundation` and `plutil`, but adds them to `TestFoundation` and `xdgTestHelper`, since those are the only targets that need a clue where dispatch is before being installed. 
 
> Note that currently `plutil` in the snapshot release is broken. It should have a `RUNPATH` of `$ORIGIN/../lib/swift/linux` (or whatever os) and was originally fixed in #1789 but has since been broken by updates to the `CMakeLists.txt`.

That explain why I’m seeing different results from @buttaface above. If you want to create that PR, I will be happy to accept it. Notice that right now, I think `plutil` uses `${Foundation_RPATH}`, which I think it is incorrect.


> `libFoundation.so` should only have `$ORIGIN` but has the build paths as well:

I think that path is added by the compiler itself, and only in Linux. https://github.com/apple/swift/blob/ad79bbaf860a3281b72d9f775b7cbfde57e30639/lib/Driver/UnixToolChains.cpp#L199-L205. It has an obvious `TODO` about the standard library moving around, which is what's happening here. In Android, this piece of code is disabled (because when cross-compiling, whatever absolute rpath you add is wrong). That part of the code might be right once the compiler is installed in some machine, but it is not helpful when the compiler is not in the final location. Darwin has a similar piece of code. 🤷‍♂️

drodriguez:
@swift-ci please test

drodriguez:
@swift-ci please test linux

spevans:
I was just having a re-read through this PR and noticed 2 other changes that need to be made:

For platforms where the `RPATH` is being set in`TestFoundation`:
The following should be removed when runnng `TestFoundation`:
```
  ENVIRONMENT
 LD_LIBRARY_PATH=${CMAKE_CURRENT_BINARY_DIR}:${FOUNDATION_PATH_TO_XCTEST_BUILD}:${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}:${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src
```

Also, `${FOUNDATION_PATH_TO_XCTEST_BUILD}` needs to be added to the `RPATH` for `TestFoundation`

drodriguez:
@swift-ci please test

drodriguez:
I hope I captured all the feedback with the last version (environment variables removed, complete the rpaths with XCTest where needed).

And since we are here, I don’t think `plutil` should use `Foundation_RPATH` (which is just `$ORIGIN`), since `plutil` ends up in the bindir. I think we could remove the rpath from `plutil`, but that will make it difficult to use in the build tree.

(Additionally it has the rpath to the build tree even after installing, but as I have said above, that's something the compiler does).

spevans:
Ive done a fix for `plutil` in #2307 which I will rebase on top of this once it has gone in.

drodriguez:
@swift-ci please test

drodriguez:
@spevans so I think removing `-rpath;-Xlinker;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src` from `libdispatch_ldflags` have the secondary effect of not allowing the LLDB test to load Foundation. Seems that LLDB links against the build tree again, which breaks the loading (I imagine that `TestFoundation` doesn’t fail to load dispatch because the executable itself has the right rpath).

I don’t know if it is worth it to copy the libraries to a temporal directory for testing LLDB or that will become untenable over time.

spevans:
I wouldnt bother about copying the libraries I would just leave the path in for now. In future we might be able to set an `LD_LIBRARY_CACHE` path when running lldb but let's leave that for now and try and get this PR in.

drodriguez:
@swift-ci please test

drodriguez:
@swift-ci please test

spevans:
@drodriguez Do you want to rebase this and get it merged?

drodriguez:

spevans:
@swift-ci test linux

drodriguez:
Rebased to avoid conflicts with `master`. Ready for review.

CMakeLists.txt
```diff
@@ -92,18 +92,27 @@ set(libdispatch_ldflags)
...
+  append_linker_rpath(Foundation_RPATH "\\\$\$ORIGIN")
+  append_linker_rpath(TestFoundation_RPATH ${CMAKE_CURRENT_BINARY_DIR})
+  append_linker_rpath(TestFoundation_RPATH ${FOUNDATION_PATH_TO_XCTEST_BUILD})
+  append_linker_rpath(XDG_TEST_HELPER_RPATH ${CMAKE_CURRENT_BINARY_DIR})
+  append_linker_rpath(plutil_rpath "\\\$\$ORIGIN/../lib/swift/${swift_os}")
+  if (FOUNDATION_ENABLE_LIBDISPATCH)
+    file(TO_CMAKE_PATH "${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}" FOUNDATION_PATH_TO_LIBDISPATCH_BUILD)
+    # NOTE: the following two are for testing LLDB and others, but should be removed
+    append_linker_rpath(Foundation_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src)
+    append_linker_rpath(Foundation_RPATH ${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD})
```
spevans: We shouldn't be adding build paths to `libFoundation.so` we should fix lldb instead

drodriguez: That `NOTE` is the result of our conversation on May 31st: https://github.com/apple/swift-corelibs-foundation/pull/2246#issuecomment-497881768

drodriguez:
For some reason https://ci.swift.org/view/Pull%20Request/job/swift-corelibs-foundation-PR-Linux/3426 haven't reported back here. Firing it up again.

CMakeLists.txt
drodriguez: @spevans: Some decision around this? I can try to fix it in LLDB, but I would l love to not have to modify it in there before merging this. Thanks!

buttaface:
Any interest in getting this in? I made [a similar change to remove the libdispatch build directory runpath](https://github.com/termux/termux-packages/pull/4895/files#diff-3cdf9fd33dc2ba4aa8334a1067babff2) when packaging up the Swift toolchain for Android.

buttaface:
This CMake config has changed so much over the past year that this pull doesn't apply anymore. It will have to be redone, either here or in a separate pull.

shahmishal:

PR #2248, opened 2019-05-09
XFAIL TestXMLDocument.test_validation_failure()

brentdax:
Probably the same cause as https://bugs.swift.org/browse/SR-10643.


brentdax:
Force merging to unblock CI.

PR #2253, opened 2019-05-12
Foundation: correct `Thread.sleep(until:)` on Windows

compnerd:
We were computing the duration incorrectly, using the time since the
Cocoa epoch rather than now resulting in what appeared to be an infinite
sleep.  Correct the duration which is relative to *now* in 100ns
intervals.  This allows the NSLock tests to succeed.


compnerd:
@swift-ci please test

PR #2259, opened 2019-05-13
Place a time limit on TestFileHandle.test_readWriteHandlers()

brentdax:
This test seems to sometimes fail, which can make it hang forever. Give it a still-generous-but-not-infinite timeout so it at least fails gracefully when it fails.

Fixes [SR-10680](https://bugs.swift.org/browse/SR-10680) and rdar://problem/50735595.


brentdax:
@swift-ci test

brentdax:

brentdax:

PR #2260, opened 2019-05-13
build: correct dependencies for Windows

compnerd:
When building Foundation, ensure that we build CoreFoundationResources
as we need to incorporate the build artifacts.


compnerd:
@swift-ci please test

PR #2261, opened 2019-05-13
Process: correct environment block handling on Windows

compnerd:
Adjust the Windows environment creation to ensure that we properly form
the string that represents the child's environment.


compnerd:
@swift-ci please test

PR #2262, opened 2019-05-13
TestFoundation: force `process.run()` to succeed

compnerd:
When debugging the process tests on Windows, the failure to
`Process.launch` would silently be ignored.  The tests do not expect the
`launch()` call to fail so ensure that the child process is launched.
This ensures that we get a proper indication of the failure.

### DIFF
--- a/TestFoundation/TestProcess.swift
+++ b/TestFoundation/TestProcess.swift
@@ -10,7 +10,7 @@
 #if !os(Android)
 class TestProcess : XCTestCase {
     
-    func test_exit0() {
+    func test_exit0() throws {
         let process = Process()
         let executableURL = xdgTestHelperURL()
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

spevans:
This would probably be better as `XCTAssertNoThrow(try task.run())` to avoid the hard fail.

compnerd:
@swift-ci please test

compnerd:
@spevans I think that @millenomi's suggestion here should address that as well; but that was a great point

compnerd:
@swift-ci please test

millenomi:
For the record: the suggestion here is to move to `throws` test function and use `try …run()` instead.

PR #2267, opened 2019-05-15
Foundation: fix `Process.run()` handle handling for Windows

compnerd:
Do not close both sides of the pipe.  The pipes are half-duplex and the
parent side of the pipe need to be valid to permit communication with
the child.


compnerd:
@swift-ci please test

PR #2268, opened 2019-05-15
Foundation: avoid `GetFileInformationByHandle` on a pipe

compnerd:
Ensure that we check the handle type before querying the file
information.  It is not permissible to invoke that with a handle to a
pipe.


compnerd:
@swift-ci please test

PR #2269, opened 2019-05-15
[windows] NSRecursiveLock: missing initialization of mutex/condition variable

lxbndr:
Looks like timeout condition variable and corresponding mutex are not initialized in `RSMRecursiveLock` implementation. Noticed random but often crashes in our test code. Best reproduced with nested lock/unlock calls
```
let lock1 = NSRecursiveLock()
let lock2 = NSRecursiveLock()

lock1.lock()
lock2.lock()
lock2.unlock() // <- crash or hang here
```

Not sure if there is no additional issues though, still getting very rare crashes. But success rate changed from, like, "1 success per 100 failures" to "1 failure per 100 success runs".

### DIFF
--- a/Foundation/NSLock.swift
+++ b/Foundation/NSLock.swift
@@ -234,6 +234,8 @@ open class NSRecursiveLock: NSObject, NSLocking {
         super.init()
 #if os(Windows)
         InitializeCriticalSection(mutex)
+        InitializeConditionVariable(timeoutCond)
+        InitializeSRWLock(timeoutMutex)
 #else
 #if CYGWIN
         var attrib : pthread_mutexattr_t? = nil


compnerd:
Um, no, no one would ever write code like that, this is ... yes, yes, something must be wrong with the checkout.  *sigh*  At least I got the non-recursive case right?  That counts for something right?

compnerd:

Foundation/NSLock.swift
```diff
@@ -234,6 +234,8 @@ open class NSRecursiveLock: NSObject, NSLocking {
         super.init()
 #if os(Windows)
         InitializeCriticalSection(mutex)
+        InitializeConditionVariable(timeoutCond)
```
gottesmm: Test?

lxbndr: Oh. Just realized there is no test for locks. It would be helpful for sure. Should I add test to this particular PR?

gottesmm: Can you add it in a subsequent PR? My suggestion: I would add the test with this reverted, make sure it fails and then add this back in.

lxbndr: Sounds good, will do.

PR #2274, opened 2019-05-16
FileHandle: _readDataOfLength: reduce fstat() calls

spevans:
- Cache the blocksize and a regular file flag to avoid
  repeated calls to fstat().

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -211,7 +211,46 @@ open class FileHandle : NSObject {
             fatalError("\(error)")
         }
     }
-    
+
+    private func _fstat() throws -> stat {
+        var statbuf = stat()
+        if fstat(_fd, &statbuf) < 0 {
+            throw _NSErrorWithErrno(errno, reading: true)
+        }
+        return statbuf
+    }
+
+    private func _fileSize() throws -> Int64 {
+        let statbuf = try _fstat()
+        return Int64(statbuf.st_size)
+    }
+
+    private var _isRegularFile: Bool? = nil
+    private var _blockSize: Int? = nil
+
+    private func _isFileAndBlockSize() throws -> (Bool, Int, Int64?) {
+        var fileSize: Int64?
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

Foundation/FileHandle.swift
```diff
@@ -211,7 +211,46 @@ open class FileHandle : NSObject {
...
+        }
+        return statbuf
+    }
+
+    private func _fileSize() throws -> Int64 {
+        let statbuf = try _fstat()
+        return Int64(statbuf.st_size)
+    }
+
+    private var _isRegularFile: Bool? = nil
```
compnerd: Can this not change?  If I call the method and underneath remove the file and recreate it, the cached value will be wrong.  I don't think it is sufficient to just cache this value.

Foundation/FileHandle.swift
```diff
@@ -211,7 +211,46 @@ open class FileHandle : NSObject {
...
+    private var _blockSize: Int? = nil
```
compnerd: This is unlikely to change unless the volume is changed.  Do we need to consider any side effects from operations like `chroot` altering the volume while the application is executing?

Foundation/FileHandle.swift
spevans: If the file descriptor is already open, deleting/recreating the underlying file wont change what fd is pointing at.

Although I just noticed there is a `closeFile` method which should invalidate the cached values so I will update it for that.

millenomi:
In general, I am skeptical of optimizations without a target, especially when the source of truth is the filesystem. What is this reduction useful for?

Also: you have a time-of-check-time-fo-use race window between the `_isFile…` call on 333 and the `mmap` immediately below, which given how volatile `mmap` can be is concerning.

spevans:
The file size is not cached and always read just before the mmap(). Im pretty sure once an FD has been opened its filetype and block size don't change unless Im mistaken?

shahmishal:

PR #2276, opened 2019-05-16
Initialize pthread mutexes in NSRecursiveLock

lxbndr:
I was going to suggest a test for NSRecursiveLock fix in #2269, as promised. But test turned out extremely useful from first run. Found that pthread-based implementation is also missing some initialization.

I suppose it will be ok to add this new fix and test in one PR, right?

### DIFF
--- a/TestFoundation/TestNSLock.swift
+++ b/TestFoundation/TestNSLock.swift
@@ -13,7 +13,8 @@ class TestNSLock: XCTestCase {
 
             ("test_lockWait", test_lockWait),
             ("test_threadsAndLocks", test_threadsAndLocks),
-
+            ("test_recursiveLock", test_recursiveLock),
+            
         ]
     }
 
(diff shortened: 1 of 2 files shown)


compnerd:
Yeah, having that in a single change is fine.

compnerd:
@swift-ci please test

gottesmm:
Can you make the timeout looser? Maybe 30 seconds. Sometimes these machines get hit really hard.

lxbndr:
> Can you make the timeout looser? Maybe 30 seconds. Sometimes these machines get hit really hard.
Sure. I think I need to add better timeout logic then, because current code waits 2 sec unconditionally. I'll change it to wait until all threads finish but not greater than timeout.

TestFoundation/TestNSLock.swift
```diff
@@ -115,4 +116,75 @@ class TestNSLock: XCTestCase {
...
+        var isTimedOut = false
+        while !isTimedOut && completedThreadCount < threadCount {
+            isTimedOut = !threadCompletedCondition.wait(until: endTime)
+        }
+        
+        countdownValueLock.lock()
+        let resultCountdownValue = countdownValue
+        countdownValueLock.unlock()
+        
+        XCTAssertEqual(completedThreadCount, threadCount, "Some threads are still not finished, could be hang")
```
compnerd: nit: "hang" -> "hung"

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

TestFoundation/TestNSLock.swift
lxbndr: Thanks. Small things matter.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

lxbndr:
I see, Linux build fails. Will look into that, guess I missed something in ifdefs.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #2277, opened 2019-05-16
Fix Up FileManager Tests on Windows

gmittert:
- Windows directory enumerator now doesn't look in a directory it
  returns until nextObject is called again
- Fixed some path handling platform issues
- Properly calculate level

See SR-10738

### DIFF
--- a/CoreFoundation/URL.subproj/CFURL.c
+++ b/CoreFoundation/URL.subproj/CFURL.c
@@ -4292,7 +4292,7 @@ static CFStringRef URLPathToWindowsPath(CFStringRef path, CFAllocatorRef allocat
 static CFStringRef _resolveFileSystemPaths(CFStringRef relativePath, CFStringRef basePath, Boolean baseIsDir, CFURLPathStyle fsType, CFAllocatorRef alloc) CF_RETURNS_RETAINED {
     CFIndex baseLen = CFStringGetLength(basePath);
     CFIndex relLen = CFStringGetLength(relativePath);
-    UniChar pathDelimiter = '/';
+    UniChar pathDelimiter = _CFGetSlash();
     UniChar *buf = (UniChar *)CFAllocatorAllocate(alloc, sizeof(UniChar)*(relLen + baseLen + 2), 0);
     CFStringGetCharacters(basePath, CFRangeMake(0, baseLen), buf);
     if (baseIsDir) {
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -652,64 +652,69 @@ extension FileManager {
         var _options : FileManager.DirectoryEnumerationOptions
         var _errorHandler : ((URL, Error) -> Bool)?
         var _stack: [URL]
-        var _current: URL?
+        var _lastReturned: URL
         var _rootDepth : Int
 
         init(url: URL, options: FileManager.DirectoryEnumerationOptions, errorHandler: (/* @escaping */ (URL, Error) -> Bool)?) {
             _options = options
             _errorHandler = errorHandler
-            _stack = [url]
+            _stack = []
             _rootDepth = url.pathComponents.count
+            _lastReturned = url
         }
 
         override func nextObject() -> Any? {
-            func contentsOfDir(directory: URL) -> [URL]? {
-                var ffd: WIN32_FIND_DATAW = WIN32_FIND_DATAW()
-                guard let dirFSR = directory.withUnsafeFileSystemRepresentation({ $0.flatMap { fsr in String(utf8String: fsr) } })
-                else { return nil }
-                let dirPath = joinPath(prefix: dirFSR, suffix: "*")
-                let h: HANDLE = dirPath.withCString(encodedAs: UTF16.self) {
+            func firstValidItem() -> URL? {
+                while let url = _stack.popLast() {
+                    if !FileManager.default.fileExists(atPath: url.path, isDirectory: nil) {
+                        guard let handler = _errorHandler,
+                              handler(url, _NSErrorWithWindowsError(GetLastError(), reading: true))
+                        else { return nil }
+                        continue
+                    }
+                    _lastReturned = url
+                    return _lastReturned
+                }
+                return nil
+            }
+
+            // If we most recently returned a directory, decend into it
+            var isDir: ObjCBool = false
+            guard FileManager.default.fileExists(atPath: _lastReturned.path, isDirectory: &isDir) else {
+              guard let handler = _errorHandler,
+                    handler(_lastReturned, _NSErrorWithWindowsError(GetLastError(), reading: true))
+              else { return nil }
+              return firstValidItem()
+            }
+
+            if isDir.boolValue && (level == 0 || !_options.contains(.skipsSubdirectoryDescendants)) {
+                var ffd = WIN32_FIND_DATAW()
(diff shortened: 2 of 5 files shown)


gmittert:
@spevans Is there a particular scenario you're testing for with the file names ending with `.` (https://github.com/apple/swift-corelibs-foundation/pull/2277/files#diff-fffbc0319b30adac9f635327a750b82fR509)? Windows isn't capable of create/working with files ending with `.` so I'm trying to figure out what I want to do with those test cases on Windows.

Foundation/NSPathUtilities.swift
```diff
@@ -345,7 +349,11 @@ extension NSString {
         
         let automount = "/var/automount"
         resolved = resolved._tryToRemovePathPrefix(automount) ?? resolved
+#if os(Windows)
+        return String(resolved.map { $0 == "\\" ? "/" : $0 })
```
compnerd: Why change the path separator here?

Foundation/NSURL.swift
```diff
@@ -32,7 +32,11 @@ private func _standardizedPath(_ path: String) -> String {
     if !path.absolutePath {
         return path._nsObject.standardizingPath
     }
+    #if os(Windows)
+    return String(path.map({$0 == "/" ? "\\" : $0}))
```
compnerd: My understanding is that the standardizing should collapse relative components (i.e. `..` should be chased, `.` dropped).  As such, any reason to not just use `PathCchCanonicalizeEx`?

Foundation/NSPathUtilities.swift
```diff
@@ -118,7 +118,11 @@ extension String {
     }
 
     internal var absolutePath: Bool {
```
compnerd: Bleh!  It would be nice to rename this to `isAbsolutePath`.

TestFoundation/TestFileManager.swift
```diff
@@ -447,13 +447,25 @@ class TestFileManager : XCTestCase {
...
+                } else if item == "path2\\item" {
+                    item2FileAttributes = e.fileAttributes
+                }
+#else
                 if item == "item" {
                     item1FileAttributes = e.fileAttributes
                 } else if item == "path2/item" {
                     item2FileAttributes = e.fileAttributes
                 }
+#endif
```
compnerd: Isn't this hunk equivalent to just changing L459 (R) to `} else if item == "path2/item" || item == "path2\\item" {`?

TestFoundation/TestFileManager.swift
```diff
@@ -447,13 +447,25 @@ class TestFileManager : XCTestCase {
...
             }
+#if os(Windows)
+            XCTAssertEqual(foundItems, Set(["item", "path2", "path2\\item"]))
```
compnerd: I wonder if you can span the `Set` over two lines and do an `#if os(Windows)` to swap out the third element.

Foundation/FileManager+Win32.swift
```diff
@@ -689,27 +690,52 @@ extension FileManager {
...
-                               _NSErrorWithWindowsError(GetLastError(), reading: true))
-                           if !keepGoing { return nil }
-                        }
-                        continue
-                    }
-                    _stack.append(contentsOf: dirContents)
+
+            func findFirstValidFile() -> URL? {
+              while let url = _stack.popLast() {
+                let dirFSR = url.withUnsafeFileSystemRepresentation { $0.flatMap { fsr in String(utf8String: fsr) } } ?? ""
```
compnerd: Hmm, this seems pretty expensive.

Foundation/FileManager+Win32.swift
```diff
@@ -689,27 +690,52 @@ extension FileManager {
...
+                if !FileManager.default.fileExists(atPath: dirFSR, isDirectory: nil) {
```
compnerd: Huh?  Should `dirFSR` be renamed?  If `isDirectory` is `nil`, then we do not know if it is a file or directory, so it seems to be misleading.

Foundation/FileManager+Win32.swift
```diff
@@ -689,27 +690,52 @@ extension FileManager {
...
                 _current = url
-                return url
+                return _current
+              }
+              return nil
             }
-            return nil
+
+            var isDir: ObjCBool = false
+            let dirFSR = _current.withUnsafeFileSystemRepresentation { $0.flatMap { fsr in String(utf8String: fsr) } } ?? ""
```
compnerd: Again, the `dirFSR` is really misleading here.  Can we rename this as we do not know what the path points to.

gmittert:
- Cleaned up the directory enumerator code to be less terrible
- Fixed some slash issues and removed the extra conversions

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -4474,7 +4474,7 @@ CFStringRef CFURLCreateStringWithFileSystemPath(CFAllocatorRef allocator, CFURLR
     //	and do a linked-on-or-later check so we don't break third parties.
     //	See <rdar://problem/4003028> Converting volume name from POSIX to HFS form fails and
     //	<rdar://problem/4018895> CF needs to back out 4003028 for icky details.
-    if ( relPath && CFURLHasDirectoryPath(anURL) && CFStringGetLength(relPath) > 1 && CFStringGetCharacterAtIndex(relPath, CFStringGetLength(relPath)-1) == '/') {
+    if ( relPath && CFURLHasDirectoryPath(anURL) && CFStringGetLength(relPath) > 1 && CFStringGetCharacterAtIndex(relPath, CFStringGetLength(relPath)-1) == _CFGetSlash()) {
```
compnerd: Since this is from a URL, is this change needed, and if it is from a URL is this not incorrect?  CC: @millenomi

TestFoundation/TestFileManager.swift
```diff
@@ -492,10 +497,17 @@ class TestFileManager : XCTestCase {
         let fm = FileManager.default
         let basePath = NSTemporaryDirectory() + "testdir\(NSUUID().uuidString)/"
         let subDirs1 = basePath + "subdir1/subdir2/.hiddenDir/subdir3/"
-        let subDirs2 = basePath + "subdir1/subdir2/subdir4.app/subdir5./.subdir6.ext/subdir7.ext./"
         let itemPath1 = basePath + "itemFile1"
+#if os(Windows)
+        // Filenames ending with '.' are not valid on Windows, so don't bother testing them
```
compnerd: @millenomi - what do you think about this?

CoreFoundation/URL.subproj/CFURL.c
gmittert: Seems like it's not, I've removed that change.

TestFoundation/TestFileManager.swift
millenomi: You can't. You can move the set creation elsewhere, or make it mutable and `#if os(Windows)` the addition.

gmittert:
@swift-ci please test

TestFoundation/TestFileManager.swift
millenomi: That's fine.

millenomi:

millenomi:

PR #2279, opened 2019-05-17
Prevent pipe-based FileHandle tests from hanging

brentdax:
Writing to a pipe can block if an internal buffer fills up, and the size of the buffer varies by platform and even kernel version, which may be causing the `TestFileHandle.testWritingWith*()` tests to hang. Address this by ~~writing from a background queue and imposing a timeout on the write~~ writing to a temporary file and then reading it back.

### DIFF
--- a/TestFoundation/TestFileHandle.swift
+++ b/TestFoundation/TestFileHandle.swift
@@ -61,17 +61,24 @@ class TestFileHandle : XCTestCase {
         emperor of Blefuscu, and sue for peace
         """.data(using: .utf8)!
     }()
-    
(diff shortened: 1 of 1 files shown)


brentdax:
@swift-ci please test

brentdax:
Lily asked for something that didn't use expectations, so this just does the simplest possible thing: writes the data to a temporary file and reads it back.

@swift-ci please test

brentdax:
@swift-ci please test

brentdax:
@swift-ci please test

PR #2280, opened 2019-05-17
build: remove unnecessary CMake include

compnerd:
We no longer use `ExternalProject` to build CoreFoundation and opt to
use `add_subdirectory` for it to get better dependency resolution.
Remove the vestigial include.


compnerd:

PR #2281, opened 2019-05-17
build: use the new `-libc` option

compnerd:
This sets up the build to link via the new `-libc` option so that we get
the correct MSVC runtime library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,10 +4,21 @@ cmake_minimum_required(VERSION 3.4.3)
 list(APPEND CMAKE_MODULE_PATH
      "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
+# NOTE(compnerd) enable CMP0091 - select MSVC runtime based on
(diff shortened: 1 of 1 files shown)


compnerd:

CMakeLists.txt
```diff
@@ -4,11 +4,22 @@ cmake_minimum_required(VERSION 3.4.3)
 list(APPEND CMAKE_MODULE_PATH
      "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
+# NOTE(compnerd) enable CMP0091 - select MSVC runtime based on
+# CMAKE_MSVC_RUNTIME_LIBRARY.  Requires CMake 3.15 or newer
+if(POLICY CMP0091)
+  cmake_policy(CMP0091 NEW)
```
buttaface: [Shouldn't this be `SET CMP0091 NEW`](https://cmake.org/cmake/help/latest/command/cmake_policy.html#setting-policies-explicitly)? Only hitting this now after finally upgrading to CMake 3.15.

PR #2283, opened 2019-05-18
CoreFoundatin: remove option to disable libdispatch

compnerd:
CoreFoundation requires libdispatch.  It is available on all targets
currently supported, so remove the option to disable it.  It was only
possible to disable it on the Swift deployment.

### DIFF
--- a/CoreFoundation/Base.subproj/module.private.modulemap
+++ b/CoreFoundation/Base.subproj/module.private.modulemap
@@ -1,0 +1,8 @@
+framework module CoreFoundation_Private [extern_c] [system] {
+  umbrella header "ForSwitFoundationOnly.h"
+  export *
+  module * {
+    export *
+  }
+}
+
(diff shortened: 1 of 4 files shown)


compnerd:
CC: @millenomi @phausler

compnerd:
@swift-ci please test

CoreFoundation/Base.subproj/module.private.modulemap
```diff
@@ -0,0 +1,8 @@
+framework module CoreFoundation_Private [extern_c] [system] {
```
phausler: This might pose a problem for the merge back process

compnerd: I'm gonna split this up into a separate change and try to handle the libdispatch related clean up before attempting to figure out how to deal with the private headers.

phausler: sounds like a good idea to me

PR #2284, opened 2019-05-18
Clean up libdispatch handling for CoreFoundation/Foundation

compnerd:
Clean up the libdispatch handling in CoreFoundation.  All deployments of CoreFoundation have libdispatch and CoreFoundation has required libdispatch for some time now.  Now that libdispatch is available on all the deployments even under the swift deployment, remove the option to disable libdispatch portions.  This enables us to remove the dependency of the prefix header on the `-DDEPLOYMENT_ENABLE_LIBDISPATCH` flag which means that we no longer need to pass that as a CFLAG to the clang importer when building Foundation.  This also reduces the complexity in the build of CoreFoundation.

As a result of the flag no longer being needed by the clang importer, we can now simplify the build rules for Foundation, passing the `DEPLOYMENT_ENABLE_LIBDISPATCH` flag to exactly the location that needs it - the build of Foundation.  This sets the stage for removing the option to disable libdispatch in the Foundation build.  This then enables simplifying the build rules for Foundation.

### DIFF
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -14,17 +14,7 @@
 
 #include <CoreFoundation/CFAvailability.h>
 
-#if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_WIN32
-#if DEPLOYMENT_RUNTIME_SWIFT
-#if DEPLOYMENT_ENABLE_LIBDISPATCH
 #define __HAS_DISPATCH__ 1
-#else
-#define __HAS_DISPATCH__ 0
-#endif
-#else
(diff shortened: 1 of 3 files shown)


compnerd:
CC: @millenomi @phausler

compnerd:
@swift-ci please test

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -14,17 +14,7 @@
 
 #include <CoreFoundation/CFAvailability.h>
 
-#if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_WIN32
```
phausler: @millenomi do you foresee any issue with this on the merge back?

millenomi:

PR #2285, opened 2019-05-19
CoreFoundation: remove names in function pointer (NFC)

compnerd:
`interface` is a problematic name on Windows where the Windows headers
will replace it with `struct` with preprocessors which break parsing of
the header.  Remove the names as they are not needed, but leave them as
comments for documentation.


compnerd:
@swift-ci please test

PR #2286, opened 2019-05-19
CoreFoundation: make CFURLSessionInterface Win32 friendly

compnerd:
The socket type on Windows is `SOCKET`, not `int`.  This was mismatched
with CURL.

The timeout value is an alias for invalid socket which is of type
`SOCKET` not `int`.  Adjust the type accordingly and plumb it through.


compnerd:
@swift-ci please test

PR #2288, opened 2019-05-20
Foundation: correct handle management on URLSession

compnerd:
We would cast the CURL socket type to an Int and back during the
MultiHandle setup.  This does not work properly for Windows.  Given that
the interfaces are internal, keep the type sugar as far as possible
before handing off to libdispatch.

### DIFF
--- a/Foundation/URLSession/libcurl/MultiHandle.swift
+++ b/Foundation/URLSession/libcurl/MultiHandle.swift
@@ -122,7 +122,7 @@ fileprivate extension URLSession._MultiHandle {
             let handler = DispatchWorkItem { [weak self] in
                 self?.performAction(for: socket)
             }
-            ss.createSources(with: action, fileDescriptor: Int(socket), queue: queue, handler: handler)
+            ss.createSources(with: action, socket: socket, queue: queue, handler: handler)
         }
         return 0
     }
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @millenomi

Foundation/URLSession/libcurl/MultiHandle.swift
```diff
@@ -395,17 +395,25 @@ fileprivate class _SocketSources {
     var readSource: DispatchSource?
     var writeSource: DispatchSource?
 
-    func createReadSource(fileDescriptor fd: Int, queue: DispatchQueue, handler: DispatchWorkItem) {
+    func createReadSource(socket: CFURLSession_socket_t, queue: DispatchQueue, handler: DispatchWorkItem) {
         guard readSource == nil else { return }
-        let s = DispatchSource.makeReadSource(fileDescriptor: Int32(fd), queue: queue)
+#if os(Windows)
+        let s = DispatchSource.makeReadSource(handle: unsafeBitCast(socket, to: HANDLE.self), queue: queue)
```
compnerd: This is incorrect, this should be `handle: HANDLE(bitPatterrn: Int(socket))!`

Foundation/URLSession/libcurl/MultiHandle.swift
```diff
@@ -395,17 +395,25 @@ fileprivate class _SocketSources {
...
         readSource = s as? DispatchSource
         s.resume()
     }
 
-    func createWriteSource(fileDescriptor fd: Int, queue: DispatchQueue, handler: DispatchWorkItem) {
+    func createWriteSource(socket: CFURLSession_socket_t, queue: DispatchQueue, handler: DispatchWorkItem) {
         guard writeSource == nil else { return }
-        let s = DispatchSource.makeWriteSource(fileDescriptor: Int32(fd), queue: queue)
+#if os(Windows)
+        let s = DispatchSource.makeWriteSource(handle: unsafeBitCast(socket, to: HANDLE.self), queue: queue)
```
compnerd: This is incorrect, this should be `handle: HANDLE(bitPatterrn: Int(socket))!`

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #2289, opened 2019-05-22
FoundationNetworking: Split networking into its own module

millenomi:
This patch introduces a new module, `FoundationNetworking`, which will include all classes related to loading from remote URLs. Several types have been moved to that module, which downstream clients must import explicitly, e.g.:

```
#if canImport(FoundationNetworking)
import FoundationNetworking
#endif
```

 - Split the `CFURLSessionInterface` files out of Core Foundation and into their own library, `libCFURLSessionInterface`. This is produced by the CF CMakeLists.txt file on non-Darwin platforms. Just like CF, we strongly discourage linking to this library directly.

 - Split `URLSession` and related classes (except for `URL` and `URLComponents`) into a new `FoundationNetworking` module (`SwiftFoundationNetworking` on Darwin). Note:

  1. The features in Boxing.swift are needed to both modules. This file is compiled separately in both modules, and relies on a `NS_BUILDING_FOUNDATION_NETWORKING` compilation condition to compile correctly in `FoundationNetworking`

  2. SPI that was used in `InputStream` is now available through a public struct named `_InputStreamSPIForFoundationNetworkingUseOnly` for FoundationNetworking to use. This isn’t API, much like `_ObjectiveCBridgeable`.

Also:

 - Change some details to make tests fail without a fatal error for issues that aren’t programmer errors.

 - In some circumstances, CURL can set an immediate timeout reentrantly if the timeout callback invokes CURL hooks. Always dispatch the CURL hook call asynchronously from a request for an immediate timeout. 

 - Types that have a `contentsOf url: URL` initializer can still accept non-file: URLs if FoundationNetworking is available in the address space of the current process (if it has been linked or dynamically loaded, for example). Loading non-file: URLs without this will result in a fatal error. Loading file: URLs will always succeed even in the absence of `FoundationNetworking`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -346,27 +346,22 @@ add_swift_library(FoundationNetworking
                     ${CMAKE_C_COMPILER_TARGET}
                   CFLAGS
                     ${MSVCRT_C_FLAGS}
-                    ${deployment_target}
-                    ${deployment_enable_libdispatch}
                     -F${CMAKE_CURRENT_BINARY_DIR}
                   LINK_FLAGS
                     ${MSVCRT_LINK_FLAGS}
                     -L${CMAKE_CURRENT_BINARY_DIR}
-                    ${ICU_UC_LIBRARY} ${ICU_I18N_LIBRARY}
                     ${libdispatch_ldflags}
                     -lFoundation
                     ${Foundation_INTERFACE_LIBRARIES}
-					${CFURLSessionInterface_LIBRARIES}
+                    ${CFURLSessionInterface_LIBRARIES}
                     ${CURL_LIBRARIES}
                     ${Foundation_RPATH}
                     ${WORKAROUND_SR9138}
                     ${WORKAROUND_SR9995}
-                    $<$<PLATFORM_ID:Windows>:$<TARGET_OBJECTS:CoreFoundationResources>>
                   SWIFT_FLAGS
                     -DDEPLOYMENT_RUNTIME_SWIFT
-					-DNS_BUILDING_FOUNDATION_NETWORKING
+                    -DNS_BUILDING_FOUNDATION_NETWORKING
                     ${deployment_enable_libdispatch}
-                    -I;${ICU_INCLUDE_DIR}
                     -I;${CMAKE_CURRENT_BINARY_DIR}/swift
                     ${libdispatch_cflags}
                     ${swift_enable_testing}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -370,7 +370,7 @@ add_swift_library(FoundationNetworking
                     uuid
                     CoreFoundation
                     $<$<PLATFORM_ID:Windows>:CoreFoundationResources>
-				    Foundation)
+                    Foundation)
 
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,10 +117,12 @@ endif()
 # be able to remove this and just use
 # `target_link_libraries(Foundation PRIVATE CoreFoundation)`.
 set(CoreFoundation_LIBRARIES $<TARGET_FILE:CoreFoundation>)
+set(CFURLSessionInterface_LIBRARIES $<TARGET_FILE:CFURLSessionInterface>)
 get_target_property(CoreFoundation_LINK_LIBRARIES CoreFoundation LINK_LIBRARIES)
 foreach(library ${CoreFoundation_LINK_LIBRARIES})
   if(NOT library STREQUAL Threads::Threads)
     list(APPEND CoreFoundation_LIBRARIES -l${library})
+    list(APPEND CFURLSessionInterface_LIBRARIES -l${library})
   endif()
 endforeach()
 
@@ -165,8 +167,6 @@ add_swift_library(Foundation
                     Foundation/Formatter.swift
                     Foundation/FoundationErrors.swift
                     Foundation/Host.swift
-                    Foundation/HTTPCookie.swift
-                    Foundation/HTTPCookieStorage.swift
                     Foundation/IndexPath.swift
                     Foundation/IndexSet.swift
                     Foundation/ISO8601DateFormatter.swift
@@ -235,7 +235,6 @@ add_swift_library(Foundation
                     Foundation/NSTimeZone.swift
                     Foundation/NSURL.swift
                     Foundation/NSURLError.swift
-                    Foundation/NSURLRequest.swift
                     Foundation/NSUUID.swift
                     Foundation/NSValue.swift
                     Foundation/NumberFormatter.swift
@@ -264,31 +263,7 @@ add_swift_library(Foundation
                     Foundation/TimeZone.swift
                     Foundation/Unit.swift
                     Foundation/URL.swift
-                    Foundation/URLAuthenticationChallenge.swift
-                    Foundation/URLCache.swift
                     Foundation/URLComponents.swift
-                    Foundation/URLCredential.swift
-                    Foundation/URLCredentialStorage.swift
-                    Foundation/URLProtectionSpace.swift
-                    Foundation/URLProtocol.swift
-                    Foundation/URLRequest.swift
-                    Foundation/URLResponse.swift
-                    Foundation/URLSession/BodySource.swift
-                    Foundation/URLSession/Configuration.swift
-                    Foundation/URLSession/http/HTTPMessage.swift
-                    Foundation/URLSession/http/HTTPURLProtocol.swift
-                    Foundation/URLSession/libcurl/EasyHandle.swift
-                    Foundation/URLSession/libcurl/libcurlHelpers.swift
-                    Foundation/URLSession/libcurl/MultiHandle.swift
-                    Foundation/URLSession/Message.swift
-                    Foundation/URLSession/NativeProtocol.swift
-                    Foundation/URLSession/ftp/FTPURLProtocol.swift
-                    Foundation/URLSession/TaskRegistry.swift
-                    Foundation/URLSession/TransferState.swift
-                    Foundation/URLSession/URLSession.swift
-                    Foundation/URLSession/URLSessionConfiguration.swift
-                    Foundation/URLSession/URLSessionDelegate.swift
-                    Foundation/URLSession/URLSessionTask.swift
                     Foundation/UserDefaults.swift
                     Foundation/UUID.swift
                     Foundation/XMLDocument.swift
@@ -303,7 +278,6 @@ add_swift_library(Foundation
                     -F${CMAKE_CURRENT_BINARY_DIR}
                   LINK_FLAGS
                     ${CoreFoundation_LIBRARIES}
-                    ${CURL_LIBRARIES}
                     ${ICU_UC_LIBRARY} ${ICU_I18N_LIBRARY}
                     ${LIBXML2_LIBRARIES}
                     ${libdispatch_ldflags}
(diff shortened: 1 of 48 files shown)


millenomi:
@swift-ci please test

CoreFoundation/CMakeLists.txt
```diff
@@ -450,49 +520,87 @@ target_compile_options(CoreFoundation
...
                           icucore)
+  target_link_libraries(CFURLSessionInterface
+                        PRIVATE
+                          icucore)
   set_target_properties(CoreFoundation
                         PROPERTIES LINK_FLAGS
                           -Xlinker;-alias_list;-Xlinker;Base.subproj/DarwinSymbolAliases;-twolevel_namespace;-sectcreate;__UNICODE;__csbitmaps;CharacterSets/CFCharacterSetBitmaps.bitmap;-sectcreate;__UNICODE;__properties;CharacterSets/CFUniCharPropertyDatabase.data;-sectcreate;__UNICODE;__data;CharacterSets/CFUnicodeData-L.mapping;-segprot;__UNICODE;r;r)
+  set_target_properties(CFURLSessionInterface
+                        PROPERTIES LINK_FLAGS
+                          -Xlinker;-alias_list;-Xlinker;Base.subproj/DarwinSymbolAliases;-twolevel_namespace;-sectcreate;__UNICODE;__csbitmaps;CharacterSets/CFCharacterSetBitmaps.bitmap;-sectcreate;__UNICODE;__properties;CharacterSets/CFUniCharPropertyDatabase.data;-sectcreate;__UNICODE;__data;CharacterSets/CFUnicodeData-L.mapping;-segprot;__UNICODE;r;r)
```
millenomi: This shouldn't have CFCharacterSet mappings.

CMakeLists.txt
```diff
@@ -331,6 +305,80 @@ add_swift_library(Foundation
...
+                  Foundation/URLSession/TransferState.swift
+                  Foundation/URLSession/URLSession.swift
+                  Foundation/URLSession/URLSessionConfiguration.swift
+                  Foundation/URLSession/URLSessionDelegate.swift
+                  Foundation/URLSession/URLSessionTask.swift
+                  TARGET
+                    ${CMAKE_C_COMPILER_TARGET}
+                  CFLAGS
+                    ${MSVCRT_C_FLAGS}
+                    ${deployment_target}
```
compnerd: This should no longer be needed, `deployment_target` has been removed in favour of `TARGET_OS_*` from `TargetConditionals.h`

CMakeLists.txt
```diff
@@ -331,6 +305,80 @@ add_swift_library(Foundation
...
+                    ${deployment_enable_libdispatch}
```
compnerd: I believe that I have cleaned this up as well.  This was previously needed as there was a mode for `CoreFoundation` to be built without libdispatch which is no longer supported.

CMakeLists.txt
```diff
@@ -331,6 +305,80 @@ add_swift_library(Foundation
...
+                    -F${CMAKE_CURRENT_BINARY_DIR}
+                  LINK_FLAGS
+                    ${MSVCRT_LINK_FLAGS}
+                    -L${CMAKE_CURRENT_BINARY_DIR}
+                    ${ICU_UC_LIBRARY} ${ICU_I18N_LIBRARY}
+                    ${libdispatch_ldflags}
+                    -lFoundation
+                    ${Foundation_INTERFACE_LIBRARIES}
+					${CFURLSessionInterface_LIBRARIES}
```
compnerd: Indentation seems odd

CMakeLists.txt
```diff
@@ -331,6 +305,80 @@ add_swift_library(Foundation
...
+                    ${CMAKE_C_COMPILER_TARGET}
+                  CFLAGS
+                    ${MSVCRT_C_FLAGS}
+                    ${deployment_target}
+                    ${deployment_enable_libdispatch}
+                    -F${CMAKE_CURRENT_BINARY_DIR}
+                  LINK_FLAGS
+                    ${MSVCRT_LINK_FLAGS}
+                    -L${CMAKE_CURRENT_BINARY_DIR}
+                    ${ICU_UC_LIBRARY} ${ICU_I18N_LIBRARY}
```
compnerd: Does this part actually use ICU?

CMakeLists.txt
```diff
@@ -331,6 +305,80 @@ add_swift_library(Foundation
...
+                    ${libdispatch_ldflags}
+                    -lFoundation
+                    ${Foundation_INTERFACE_LIBRARIES}
+					${CFURLSessionInterface_LIBRARIES}
+                    ${CURL_LIBRARIES}
+                    ${Foundation_RPATH}
+                    ${WORKAROUND_SR9138}
+                    ${WORKAROUND_SR9995}
+                    $<$<PLATFORM_ID:Windows>:$<TARGET_OBJECTS:CoreFoundationResources>>
```
compnerd: Why do you need the CoreFoundationResources?  That is for the TimeZone information

CMakeLists.txt
```diff
@@ -331,6 +305,80 @@ add_swift_library(Foundation
...
+                  SWIFT_FLAGS
+                    -DDEPLOYMENT_RUNTIME_SWIFT
+					-DNS_BUILDING_FOUNDATION_NETWORKING
```
compnerd: Indentation seems odd

CMakeLists.txt
```diff
@@ -331,6 +305,80 @@ add_swift_library(Foundation
...
+                    ${deployment_enable_libdispatch}
+                    -I;${ICU_INCLUDE_DIR}
```
compnerd: Do we really need ICU here?

CoreFoundation/CMakeLists.txt
```diff
@@ -398,27 +438,44 @@ target_include_directories(CoreFoundation
...
+  target_include_directories(CFURLSessionInterface
+                             SYSTEM PRIVATE
+                               ${CF_PATH_TO_LIBDISPATCH_SOURCE}/src/BlocksRuntime)
 endif()
 
 if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
   target_compile_options(CoreFoundation
                          PRIVATE
                            $<$<COMPILE_LANGUAGE:C>:/FI${PROJECT_SOURCE_DIR}/Base.subproj/CoreFoundation_Prefix.h>)
+  target_compile_options(CFURLSessionInterface
```
compnerd: Does CFURLSessionInterface use the CoreFoundation prefix header?  (which seems like a PCH attempt)

CoreFoundation/CMakeLists.txt
```diff
@@ -398,27 +438,44 @@ target_include_directories(CoreFoundation
...
+                         PRIVATE
+                           $<$<COMPILE_LANGUAGE:C>:/FI${PROJECT_SOURCE_DIR}/Base.subproj/CoreFoundation_Prefix.h>)
 else()
   target_compile_options(CoreFoundation
                          PRIVATE
                            $<$<COMPILE_LANGUAGE:C>:-include;${PROJECT_SOURCE_DIR}/Base.subproj/CoreFoundation_Prefix.h>)
+  target_compile_options(CFURLSessionInterface
```
compnerd: Same

CoreFoundation/CMakeLists.txt
```diff
@@ -398,27 +438,44 @@ target_include_directories(CoreFoundation
...
+                         PRIVATE
+                           $<$<COMPILE_LANGUAGE:C>:-include;${PROJECT_SOURCE_DIR}/Base.subproj/CoreFoundation_Prefix.h>)
 endif()
 
 if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
   target_compile_options(CoreFoundation
                          PRIVATE
                            -fblocks
                            /EHsc)
+  target_compile_options(CFURLSessionInterface
```
compnerd: Does `CFURLSessionInterface` use blocks? Or exceptions `/EHsc` enables exceptions

CoreFoundation/CMakeLists.txt
```diff
@@ -450,49 +520,87 @@ target_compile_options(CoreFoundation
...
+                        PRIVATE
+                          ${CURL_LIBRARIES})
 endif()
 if(CMAKE_SYSTEM_NAME STREQUAL Android)
   target_link_libraries(CoreFoundation
                         PRIVATE
                           log)
+  target_link_libraries(CFURLSessionInterface
+                        PRIVATE
+                          log)
```
compnerd: Does `CFURLSessionInterface` use the android log apis?

CoreFoundation/CMakeLists.txt
```diff
@@ -450,49 +520,87 @@ target_compile_options(CoreFoundation
...
 endif()
 target_link_libraries(CoreFoundation
                       PRIVATE
                         Threads::Threads
                         ${CMAKE_DL_LIBS})
+target_link_libraries(CFURLSessionInterface
```
compnerd: Does `CFURLSessionInterface` use threads and `dl*` functions?

CoreFoundation/CMakeLists.txt
```diff
@@ -450,49 +520,87 @@ target_compile_options(CoreFoundation
...
+                        Threads::Threads
+                        ${CMAKE_DL_LIBS})
 if(CMAKE_SYSTEM_NAME STREQUAL Windows)
   target_link_libraries(CoreFoundation
                         PRIVATE
                           AdvAPI32
                           Secur32
                           User32
                           mincore)
+  target_link_libraries(CFURLSessionInterface
```
compnerd: I think that you probably don't need to link against most of these libraries.  `WS2_32` and `mincore` are probably all that are needed?

CoreFoundation/CMakeLists.txt
```diff
@@ -450,49 +520,87 @@ target_compile_options(CoreFoundation
...
   target_link_libraries(CoreFoundation
                         PRIVATE
                           icucore)
+  target_link_libraries(CFURLSessionInterface
+                        PRIVATE
+                          icucore)
   set_target_properties(CoreFoundation
                         PROPERTIES LINK_FLAGS
                           -Xlinker;-alias_list;-Xlinker;Base.subproj/DarwinSymbolAliases;-twolevel_namespace;-sectcreate;__UNICODE;__csbitmaps;CharacterSets/CFCharacterSetBitmaps.bitmap;-sectcreate;__UNICODE;__properties;CharacterSets/CFUniCharPropertyDatabase.data;-sectcreate;__UNICODE;__data;CharacterSets/CFUnicodeData-L.mapping;-segprot;__UNICODE;r;r)
+  set_target_properties(CFURLSessionInterface
```
compnerd: This re-embeds the character set data, why?

millenomi:
Collectively responding: I replicated the existing setup for CF line-for-line. Certain things are flat-out incorrect, like the character data; certain are not useful, like icu. I'll do a pass for all the objections.

CoreFoundation/CMakeLists.txt
millenomi: I'm keeping this, in case.

CoreFoundation/CMakeLists.txt
millenomi: I feel like these two should be a baseline for all CF code, and will keep them as-is.

CoreFoundation/CMakeLists.txt
millenomi: I can't test Windows builds at the moment; is it okay if I leave them in and you can remove them later?

millenomi:
You are correct. Yay, sleep deprivation. Fixing.

millenomi:
@compnerd I addressed everything that I could locally check would compile.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

CoreFoundation/CMakeLists.txt
compnerd: Sounds fine to me - unlike Unix, the semantics on Windows are that they will be dropped if unused anyways.

CMakeLists.txt
```diff
@@ -331,6 +305,75 @@ add_swift_library(Foundation
...
+                    ${deployment_enable_libdispatch}
+                    -I;${CMAKE_CURRENT_BINARY_DIR}/swift
+                    ${libdispatch_cflags}
+                    ${swift_enable_testing}
+                    ${swift_optimization_flags}
+                  DEPENDS
+                    uuid
+                    CoreFoundation
+                    $<$<PLATFORM_ID:Windows>:CoreFoundationResources>
+				    Foundation)
```
compnerd: Indentation is off

millenomi: Missed it; I'll fix it before merge.

Foundation/NSObjCRuntime.swift
```diff
@@ -404,3 +431,77 @@ public func NSClassFromString(_ aClassName: String) -> AnyClass? {
     
     return _typeByName(aClassNameWithPrefix) as? AnyClass
 }
+
+// The following types have been moved to FoundationNetworking. They exist here only to allow appropriate diagnostics to surface in the compiler.
+
+@available(*, unavailable, message: "This type has moved to the FoundationNetworking module. Import that module to use it.",
+    renamed: "FoundationNetworking.CachedURLResponse")
+public typealias CachedURLResponse = AnyObject
```
jrose-apple: Note: these should *not* use `renamed:`, because we want people to continue using the unqualified spelling to refer to them once they've imported FoundationNetworking. (That should still work without ambiguity because unavailable declarations should only be used if there's no valid available one.)

millenomi: We're relying on the latter here.

millenomi: In general, I cannot mandate a spelling be qualified or not; it is my job to make _both_ work.

millenomi: (but I agree we shouldn't be recommending to use the qualified one if possible.)

jrose-apple: Right. I'm specifically talking about `renamed:` because that changes the error message and creates a fix-it.

millenomi:

millenomi:

PR #2290, opened 2019-05-23
Update NSString.path(withComponents:) API

keith:
This update is to match Darwin


spevans:
@swift-ci test

millenomi:
We should have a pathWithComponents… obsoleted method to generate diagnostics.

keith:
Here's a possible version of that https://github.com/apple/swift-corelibs-foundation/pull/2298

PR #2291, opened 2019-05-24
Use var binding for switch where appropriate

ikesyo:


ikesyo:
@swift-ci Please test

PR #2292, opened 2019-05-24
[DateFormatter] Fix symbols property optionality

ikesyo:
Those are not IUO on Darwin. And also addresses "Treating a forced downcast to '[String]' as optional will never produce 'nil'" warnings.


ikesyo:
@swift-ci Please test

PR #2294, opened 2019-05-24
Parity: URLSession.reset(…)/.flush(…)

millenomi:
Implement reset(…) and flush(…) per their contracts. Note that URLCredentialStorage is unimplemented right now; invoking reset(…) will thus crash if a credentials storage is set.

Ensure that URLSessionConfiguration users do not attempt to use init(), leading to non-portable code.


millenomi:
@swift-ci please test

millenomi:
Fixed https://bugs.swift.org/browse/SR-10379

PR #2295, opened 2019-05-24
Parity: URLSessionTask.earliestBeginDate

millenomi:
This property only applies to background session tasks. If it’s being used once it’s made available, it may be in error on Linux. Warn; also ensure that properties are open as they are on Darwin.

Fixes https://bugs.swift.org/browse/SR-10382


millenomi:

PR #2299, opened 2019-05-25
CoreFoundation: repair the build after the URLSessionInterface

compnerd:
CoreFoundation's build broke on Windows due to the missing curl headers.
Linux and macOS have the curl headers available and so they did not
notice it.


compnerd:

PR #2301, opened 2019-05-25
Foundation: change `HTTPCookieStorage` to use `FileManager`

compnerd:
Change the implementation of `HTTPCookieStorage` to use
`FileManager.default.url(for:in:)` instead of the internal
_CFXDGCreateDataHomePath` method.  Now that `HTTPCookieStorage` is moved
off to its own library (FoundationNetworking), it no longer has access
to the internal function.  Use the URL query to get the path to the
storage location.


compnerd:
@swift-ci please test

millenomi:

PR #2302, opened 2019-05-25
CoreFoundation: avoid `CFString` construction in `CFURLSessionINterface`

compnerd:
This causes a reference to `CFStringCreateWithBytes` and
`kCFDefaultSystemAllocator` in `CFURLSessionInterface` which requires
linking against CoreFoundation.  Since the swift deployment uses
CoreFoundation statically, this would require a second instance of
CoreFoundation OR that private CoreFoundation interfaces are re-exposed
through Foundation which is explicitly undesired.  Perform the String
construction in Swift and instead just have a trivial wrapper for
`curl_easy_strerror`.

### DIFF
--- a/CoreFoundation/URL.subproj/CFURLSessionInterface.h
+++ b/CoreFoundation/URL.subproj/CFURLSessionInterface.h
@@ -56,7 +56,7 @@ typedef struct CFURLSessionEasyCode {
     int value;
 } CFURLSessionEasyCode;
 
-CF_EXPORT CFStringRef _Nonnull CFURLSessionCreateErrorDescription(int value);
+CF_EXPORT const char * _Nullable CFURLSessionEasyCodeDescription(CFURLSessionEasyCode code);
 
 CF_EXPORT int const CFURLSessionEasyErrorSize;
 
--- a/Foundation/URLSession/libcurl/MultiHandle.swift
+++ b/Foundation/URLSession/libcurl/MultiHandle.swift
@@ -217,9 +217,14 @@ fileprivate extension URLSession._MultiHandle {
         // Find the NSURLError code
         var error: NSError?
         if let errorCode = easyHandle.urlErrorCode(for: easyCode) {
-            let errorDescription = easyHandle.errorBuffer[0] != 0 ?
-                String(cString: easyHandle.errorBuffer) :
-                unsafeBitCast(CFURLSessionCreateErrorDescription(easyCode.value), to: NSString.self) as String
+            var errorDescription: String = ""
+            if easyHandle.errorBuffer[0] == 0 {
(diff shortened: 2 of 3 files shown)


compnerd:
@swift-ci please test

CoreFoundation/URL.subproj/CFURLSessionInterface.h
```diff
@@ -56,7 +56,7 @@ typedef struct CFURLSessionEasyCode {
     int value;
 } CFURLSessionEasyCode;
 
-CF_EXPORT CFStringRef _Nonnull CFURLSessionCreateErrorDescription(int value);
+CF_EXPORT const char * _Nulable CFURLSessionEasyCodeDescription(int value);
```
millenomi: _Nullable.

Foundation/URLSession/libcurl/MultiHandle.swift
```diff
@@ -217,9 +217,14 @@ fileprivate extension URLSession._MultiHandle {
         // Find the NSURLError code
         var error: NSError?
         if let errorCode = easyHandle.urlErrorCode(for: easyCode) {
-            let errorDescription = easyHandle.errorBuffer[0] != 0 ?
-                String(cString: easyHandle.errorBuffer) :
-                unsafeBitCast(CFURLSessionCreateErrorDescription(easyCode.value), to: NSString.self) as String
+            var errorDescription: String = ""
+            if easyHandle.errorBuffer[0] == 0 {
+              let description = CFURLSessionEasyCodeDescription(easyCode)
+              errorDescription = NSString(bytesNoCopy: UnsafeMutableRawPointer(mutating: description), length: strlen(description), encoding: String.Encoding.utf8.rawValue, freeWhenDone: false)! as String
```
millenomi: This violates the implied contract of bridging, where we reserve the right to keep around the object you’re trying to bridge. It happens to work in SCF but it may or may not going forward, and would’ve crashed on Darwin for sure.

millenomi: (Just drop the `…NoCopy`.)

compnerd: That is a good point, better to be safe and do the copy.  The reason that I had elided the copy is because `curl_easy_strerror` would return a string that is globally allocated and can be preserved.

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

compnerd:

PR #2304, opened 2019-05-26
build: port to new Swift support

compnerd:
This ports the foundation build to use the new builtin Swift support in
CMake.  It allows for a significant cleanup of the build rules.


CoreFoundation/CMakeLists.txt
```diff
@@ -422,6 +426,9 @@ if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
                              PRIVATE
                                ${LIBXML2_INCLUDE_DIR})
   find_package(CURL REQUIRED)
+  target_include_directories(CoreFoundation
```
millenomi: CoreFoundation is not supposed to import or link with libcurl.

compnerd: Yeah, this is based on an older state of the tree.  I will have to redo the work, this was to demonstrate what the build would look like.

compnerd:
Superseded by #2524

PR #2307, opened 2019-05-26
plutil: Add a specific rpath for plutil and use it when linking.

spevans:
- plutil requires an rpath of '$ORIGIN/../lib/swift/${swift_os}'
  so set this specificaly for plutil.

- Add a test for running plutil, which relies on LD_LIBRARY_PATH
  being set when the tests are run to point to the build directories.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,10 +97,12 @@ if(FOUNDATION_ENABLE_LIBDISPATCH)
   endif()
 endif()
 
+set(plutil_rpath)
 if(CMAKE_SYSTEM_NAME STREQUAL Android OR CMAKE_SYSTEM_NAME STREQUAL Linux OR
     CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
   set(Foundation_RPATH -Xlinker;-rpath;-Xlinker;"\\\$\$ORIGIN")
   set(XDG_TEST_HELPER_RPATH -Xlinker;-rpath;-Xlinker;${CMAKE_CURRENT_BINARY_DIR})
+  set(plutil_rpath -Xlinker;-rpath;-Xlinker;"\\\$\$ORIGIN/../lib/swift/${swift_os}")
 elseif(CMAKE_SYSTEM_NAME STREQUAL Windows)
   # FIXME(SR9138) Silence "locally defined symbol '…' imported in function '…'
   set(WORKAROUND_SR9138 -Xlinker;-ignore:4049;-Xlinker;-ignore:4217)
@@ -389,11 +391,11 @@ add_swift_executable(plutil
                      CFLAGS
                        -F${CMAKE_CURRENT_BINARY_DIR}
                      LINK_FLAGS
-                       ${libdispatch_ldflags}
                        -L${CMAKE_CURRENT_BINARY_DIR}
+                       -L;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src
                        -lFoundation
                        ${Foundation_INTERFACE_LIBRARIES}
-                       ${Foundation_RPATH}
+                       ${plutil_rpath}
                      SWIFT_FLAGS
                        -DDEPLOYMENT_RUNTIME_SWIFT
                        -I;${CMAKE_CURRENT_BINARY_DIR}/swift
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test

spevans:
@swift-ci test macos

spevans:
@swift-ci test macos

CMakeLists.txt
```diff
@@ -392,11 +395,11 @@ add_swift_executable(plutil
                      CFLAGS
                        -F${CMAKE_CURRENT_BINARY_DIR}
                      LINK_FLAGS
-                       ${libdispatch_ldflags}
                        -L${CMAKE_CURRENT_BINARY_DIR}
+                       -L;${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src
```
compnerd: This may add an empty `/src` to the library search path.  @millenomi - is building Foundation with Dispatch something which is desired?

spevans: `${FOUNDATION_PATH_TO_LIBDISPATCH_BUILD}/src` was in the `libdispatch_ldflags` as a `RUNPATH`, so this is still needed for the link to succeed, but the `RUNPATH` itself isnt.

spevans:
Please test with following pull request:
https://github.com/apple/swift-integration-tests/pull/57

@swift-ci please test

spevans:
Please test with following pull request:
apple/swift-integration-tests#57

@swift-ci please test macos

spevans:
Please test with following pull request:
apple/swift-integration-tests#57

@swift-ci please test macos

spevans:
Please test with following pull request:
apple/swift-integration-tests#57

@swift-ci please test macos

spevans:
Please test with following pull request:
apple/swift-integration-tests#57

@swift-ci please test macos

millenomi:
I'm _really_ not sure we should ship RPATHs in final executables. @jrose-apple how does the rest of the setup deal with dependencies?

jrose-apple:
:-/ I think this is the right answer on Linux where we don't have an equivalent of Mach-O's `@executable_path`.

jrose-apple:
(`$ORIGIN` is that equivalent but AFAIK you can't link directly to something via `$ORIGIN`; you have to go through `@rpath`.)

Alternately, static-link all the things, but that's not a good answer either.

spevans:
Please test with following pull request:
apple/swift-integration-tests#57

@swift-ci please test macos

spevans:
Please test with following pull request:
apple/swift-integration-tests#57
apple/swift-integration-tests#58

@swift-ci please test

jrose-apple:
I don't think the cross-PR testing supports multiple PRs from the same repo. :-(

spevans:
Yeah, I wanted to test apple/swift-integration-tests#58 as it looks like tests can't be run directly in the `swift-integration-tests` repo (or maybe I dont have access).

spevans:
@millenomi We still need to add `RPATH`s whilst we are installing from a tar since the install could be located anywhere, so paths relative to `$ORIGIN` are needed. The only alternative would be to create proper packages and then add the swift lib directory as a config file in `/etc/ld.so.conf.d/`

Other binaries ship with even worse paths, here is `swift-build`:
```
chrpath  -l ~/swift-DEVELOPMENT-SNAPSHOT-2019-01-15-a-ubuntu18.04/usr/bin/swift-build
/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-01-15-a-ubuntu18.04/usr/bin/swift-build: RUNPATH=/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:$ORIGIN:$ORIGIN/../lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swiftpm-linux-x86_64/.bootstrap/lib/swift/linux:$ORIGIN/../lib/swift/pm/llbuild:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/llbuild-linux-x86_64/lib
```

The only two ways I can think of to clean this up would be
1. Add an option to `swiftc` to disable adding `RPATH`s  by default - A bit heavy if its only binaries built by the swift install that have bad paths but might be useful in other cases.
2. Install `chrpath` and use it to prune the paths after building.

jrose-apple:
That option is coming in #24787, by the way, though it isn't hooked up to Linux yet. (I haven't landed it because it's immediately going to cause merge conflicts with the Apple-internal branches and I need to be ready to resolve them.)

spevans:
That `-no-stdlib-rpath` options look ideal, I will keep an eye on the PR and then update it for Linux if anything needs to be done. I think that will allow us clean up a lot of `RPATH`s in the corelibs.

spevans:
@swift-ci test linux

spevans:
Please test with following pull request:
apple/swift-integration-tests#57

@swift-ci please test

spevans:
@swift-ci test linux

compnerd:
This breaks the android build! https://dev.azure.com/compnerd/windows-swift/_build/results?buildId=3584&view=logs&j=d925d346-119b-5290-d4d9-80cb246f632e

drodriguez:
I don't know what's the configuration of the Linux CI machines, and if it is easy to change them, but if `libBlocksRuntime.so` was not installed in the system, many of these problems would surface earlier when testing Linux.

spevans:
Ive unistalled the `blocksRuntime` package on my test VM now so it should catch anymore of these problems if I make them in future PRs.

PR #2309, opened 2019-05-28
Foundation: correctly read thread name

Lukasa:
Foundation was correctly setting the thread name, but it would only ever
read the name that was stored in the thread object. This meant that if
the thread name was set elsewhere (e.g. by Linux, or by SwiftNIO) that
name would not be reflected in Thread's name property.

### DIFF
--- a/Foundation/Thread.swift
+++ b/Foundation/Thread.swift
@@ -251,9 +251,12 @@ open class Thread : NSObject {
     }
 
     open var name: String? {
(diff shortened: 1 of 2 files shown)


Lukasa:
@swift-ci test

Lukasa:
@swift-ci test macOS

Lukasa:
@swift-ci test macOS

PR #2311, opened 2019-05-29
Parity: FileManager.replaceItem(…)

millenomi:
- Instead of returning nothing from .url(for: .itemReplacementDirectory…), produce a path the same way Darwin does.

 - Implement replaceItems(…) for POSIX-y platforms. We try to use Darwin and Linux swap syscalls to ensure race-free atomicity if available, and if these fail we fall back to an implementation that has different degrees of atomicity depending on what object is replaced with what object.

 - Add tests.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -64,6 +64,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <linux/stat.h>
+#include <linux/fs.h>
 #define AT_STATX_SYNC_AS_STAT   0x0000  /* - Do whatever stat() does */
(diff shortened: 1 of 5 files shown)


millenomi:
@swift-ci please test

millenomi:
I need a wrapper for renameat2, I assume.

millenomi:

millenomi:

spevans:
@pschuh A `#ifdef SYS_renameat2` was added in https://github.com/apple/swift-corelibs-foundation/pull/2337/files#diff-33ef000612768310a8c08cd83b00f95fR598 which should stop the syscall being compiled if it is not defined. There is also a runtime check of the kernel version to check the sys call is actually available. What error are you seeing

pschuh:
Sounds good! I just need to wait for the update-checkout to get up to that version then.

PR #2312, opened 2019-05-30
URLCache sketches. Mainly memory cache.

drodriguez:
@millenomi: just in case you work more in `URLCache`. I was working on the memory cache, and I think I have a LRU memory cache and some sketches of other parts. It is not completely compatible with the code in #2000, but I think many parts of the code can be used (or at least the ideas contained in them).

I can also work on this if you prefer. I have some ideas on how to implement the disk cache (with flat files on disk). I was trying to get `URLSession` and the related classes in a less “unimplemented minefield” state. I also have some work done in the credential storage, if you want it.

### DIFF
--- a/Foundation/URLCache.swift
+++ b/Foundation/URLCache.swift
@@ -129,6 +129,9 @@ open class CachedURLResponse : NSObject, NSSecureCoding, NSCopying {
 open class URLCache : NSObject {
     
     /*! 
+    private static let _sharedCacheLock: NSLock = NSLock()
+    private static var _sharedCache: URLCache!
+
         @method sharedURLCache
         @abstract Returns the shared URLCache instance.
         @discussion Unless set explicitly, this method returns an URLCache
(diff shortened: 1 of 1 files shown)


Foundation/URLCache.swift
```diff
@@ -129,6 +129,9 @@ open class CachedURLResponse : NSObject, NSSecureCoding, NSCopying {
 open class URLCache : NSObject {
     
     /*! 
+    private static let _sharedCacheLock: NSLock = NSLock()
```
millenomi: ?

drodriguez: I was moving commits between branches. The documentation comment should go below the two private variables. I didn’t review the code before pushing, sorry about that.

Foundation/URLCache.swift
```diff
@@ -147,14 +150,27 @@ open class URLCache : NSObject {
...
+            _sharedCacheLock.lock()
+            defer { _sharedCacheLock.unlock() }
+            _sharedCache = newValue
         }
     }
 
     /*! 
+    private let _memoryCache: _URLMemoryCache<_URLCacheKey, CachedURLResponse>
+    private let _diskPath: String
+    private let _diskCache: _URLDiskCache<_URLCacheKey, CachedURLResponse>
```
drodriguez: Also happens here. The variables should be outside the comment, not inside.

PR #2313, opened 2019-05-31
XML: Add namespace support for XML attributes

hironytic:
Implement `XMLElement.attribute(forLocalName:uri:)`.
Also fix some other incompatibilities between Linux and Darwin about attributes with namespace.

In the added test case, all expected values are actual ones on Darwin (macOS 10.14.4).


spevans:
@swift-ci test

millenomi:
How much of https://bugs.swift.org/browse/SR-10424 does this fix?

hironytic:
This PR fixes only for attributes.
AFAIK, at least there are still several namespace related `NSUnimplemented()`s in `XMLElement`.

PR #2200, opened 2019-04-30
[SR-10574] Add tests for new Japanese calendar era name

novi:
- Add Calendar test
- Add DateFormatter test

Test Calendar and DateFormatter with new era name ([SR-10574](https://bugs.swift.org/browse/SR-10574)).
Related https://github.com/apple/swift/pull/24387.

PR #2318, opened 2019-06-06
SR-10638: Change the type of force casting in `NSCalendar`.

YOCKOW:
Despite `CFDateFormatterCopyProperty(_:_:)` returns `CFArrayRef` when the key is `kCFDateFormatterEraSymbolsKey`, the compiler on Linux assume that the type is `_NSCFArray` and fails to cast even though they are compatible. 

Resolves [SR-10638](https://bugs.swift.org/browse/SR-10638).

Note: This bug is blocking [PR#2200](https://github.com/apple/swift-corelibs-foundation/pull/2200).


spevans:
@swift-ci test

PR #2321, opened 2019-06-07
SR-10596: Prevent Float being deserialized into an Int

hungtruong:
Check for the NSNumber's underlying value and throw an error if it is a floating point number when deserializing from JSON into an Int.
https://bugs.swift.org/browse/SR-10596

### DIFF
--- a/Foundation/JSONEncoder.swift
+++ b/Foundation/JSONEncoder.swift
@@ -2109,13 +2109,13 @@ extension _JSONDecoder {
             throw DecodingError._typeMismatch(at: self.codingPath, expectation: type, reality: value)
         }
 
-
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

brentdax:
@swift-ci test

parkera:
I recall some discussion around intentionally allowing this.

hungtruong:
@itaiferber yeah I think that makes sense if JSON itself doesn't have a concept of the difference between `1` and `1.0`. I do agree with the original reporter of the bug that the current inconsistency could also cause issues but I suppose it's not enough of a reason to change the status quo. Thanks for commenting!

PR #2320, opened 2019-06-07
Parity: NSDictionary: Non-keyed archiving marked as NSUnimplemented()

trsxxii:
Move from NSUnimpletented() to NSUnsupported() We do not support non-keyed-capable archiving in swift-corelibs-foundation classes.

https://bugs.swift.org/browse/SR-10396

PR #2322, opened 2019-06-07
TestNSDictionary: Add encoding and decoding test for NSCoding

NinoScript:
Adds missing test for https://github.com/apple/swift-corelibs-foundation/pull/2320


millenomi:
Thank you for the help!

We have a bigger infrastructure for this kind of test that I spent some time documenting today: https://github.com/apple/swift-corelibs-foundation/pull/2333 — can we use that instead?

Let me know if you'd like to know more!

PR #2324, opened 2019-06-09
NSCoder.decodingFailurePolicy should be read-only.

spevans:
- See https://developer.apple.com/documentation/foundation/nscoder/1642984-decodingfailurepolicy

- NSKeyedUnarchiver overrides it to be read-write, but NSKeyedArchiver
  does not.


spevans:
@swift-ci test

spevans:
@swift-ci test macos

PR #2326, opened 2019-06-11
Fix Flags in NS_open

gmittert:
- Handle the O_CREATE | O_TRUNC case
- Read only should be set if _write_ permissions are not set


gmittert:
@swift-ci please test

PR #2328, opened 2019-06-11
Fix up _contentsOfDir on Windows

gmittert:
- FindFirstFileW should be given a directory with \\* else it will only
return the given directory
- . and .. should be excluded from the results


millenomi:
@swift-ci please test linux

millenomi:
cc @compnerd?

PR #2329, opened 2019-06-11
Force file removal in Windows _removeItem

gmittert:
unlink on posix ignores if a file is read only since it only needs write
permissions on the parent directory to remove it. To emulate this on
Windows, if a file is read only, unset the read only bit and then remove
it.


millenomi:
@swift-ci please test linux

PR #2330, opened 2019-06-11
SR-10759: Avoid double encoding in `stringValue` of XMLNode.

YOCKOW:
It is unnecessary to encode entities when the node is comment or text. They will be encoded when XML is put out.

Resolves [SR-10759](https://bugs.swift.org/browse/SR-10759).


spevans:
@swift-ci test

PR #2331, opened 2019-06-11
Refactor Operation, and OperationQueue to be closer to the objc implementation

phausler:
This brings in the new barrier APIs as well as progress reporting to OperationQueue


millenomi:
@swift-ci please test linux

millenomi:
Thank you so much. :)

PR #2332, opened 2019-06-11
Implement ._hidden on Windows

gmittert:
Windows has a specific attribute to mark a file as hidden as opposed to
prefixing ".". This allows setting and getting the attribute. On POSIX,
it will return based on the "." prefix and setting it does nothing.

### DIFF
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -16,6 +16,9 @@ fileprivate let UF_HIDDEN: Int32 = 1
 #endif
 
 import CoreFoundation
+#if os(Windows)
+import MSVCRT
+#endif
 
 open class FileManager : NSObject {
     
@@ -400,7 +403,6 @@ open class FileManager : NSObject {
                         newAccessDate = providedDate
                     }
                     #endif
-                    
                 case .immutable: fallthrough
                 case ._userImmutable:
                     prepareToSetOrUnsetFlag(UF_IMMUTABLE)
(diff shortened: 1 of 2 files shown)


gmittert:
@millenomi Are you okay with adding in attributes, especially ones that are more Windows specific? If not, what would be the better way to add this support?

cc @compnerd

millenomi:
(There is already a private ._hidden attribute, IIRC?)

millenomi:
I can consider a Windows-only attribute that is prefixed if there is a compellingly strong use case other than completeness; but the right way to do this is to ask for a new API via Radar. However, in this case, we already have API that corresponds to this in a more modern API (it's `URLResourceKey.isHiddenKey`), and we tend to not duplicate, so I would advise making sure that that one works correctly instead.

gmittert:
I see, I misunderstood how the private attributes worked. I've changed it to use _hidden via URLResource now.

@swift-ci please test linux

gmittert:
@swift-ci please test linux

millenomi:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

PR #2333, opened 2019-06-12
Add extensive testing documentation

millenomi:
* Copy information critical to producing useful tests from my head onto words on disk. 

 * Simplify the instructions for GenerateTestFixtures by making the version strings used match the expectations of the Fixture extensions, and by making the executable default to its current location for output if not specified.

 * Remove a path that is local to my machine from the repository.

### DIFF
--- a/Docs/Testing.md
+++ b/Docs/Testing.md
@@ -1,0 +1,287 @@
+# Testing swift-corelibs-foundation
+
+swift-corelibs-foundation uses XCTest for its own test suite. This document explains how we use it and how we organize certain kinds of specialized testing. This is both different from the Swift compiler and standard library, which use `lit.py`, and from destkop testing, since the version of XCTest we use is not the Darwin one, but the Swift core library implementation in `swift-corelibs-xctest`, which is pretty close to the original with some significant differences.
+
+## Tests Should Fail, Not Crash
+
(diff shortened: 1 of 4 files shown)


millenomi:
@swift-ci please test linux

Docs/Testing.md
```diff
@@ -0,0 +1,287 @@
...
+}
+```
+
+* In the file you're adding the test, if not present, add the appropriate `@testable import` import prologue at the top. It will look like the one below, but **do not** copy and paste this — instead, search the codebase for `@testable import` for the latest version:
+
+```swift
+// It will look something like this:
+
+#if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
+    #if canImport(SwiftFoundation) && !DEPLOYMENT_RUNTIME_OBJC
```
spevans: I dont think the `!DEPLOYMENT_RUNTIME_OBJC` is actually that useful anymore now that `canImport(SwiftFoundation) ` is used. I cant see a situation where `canImport(SwiftFoundation)` returns true but we DONT import it and instead `import Foundation`

millenomi: True. A patch that fixes this should also fix this documentation in the same swoop.

PR #2339, opened 2019-06-12
Fixing up Windows TestFileManager

gmittert:
- _copyOrLinkDirectoryHelper should throw on errors
- Accounting for that Windows permissions aren't compatible with posix ones
- When given `""` as a file name, most Windows apis return file not found, not invalid filename

### DIFF
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -180,7 +180,7 @@ extension FileManager {
         try path.withCString(encodedAs: UTF16.self) {
             if !CreateDirectoryW($0, psaAttributes) {
                 // FIXME(compnerd) pass along path
-                throw _NSErrorWithWindowsError(GetLastError(), reading: false)
+                throw _NSErrorWithWindowsError(GetLastError(), reading: false, paths: [path])
             }
         }
         if let attr = attributes {
@@ -197,7 +197,7 @@ extension FileManager {
 
             let hDirectory: HANDLE = FindFirstFileW($0, &ffd)
             if hDirectory == INVALID_HANDLE_VALUE {
-                throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+                throw _NSErrorWithWindowsError(GetLastError(), reading: true, paths: [path])
             }
             defer { FindClose(hDirectory) }
 
@@ -208,7 +208,7 @@ extension FileManager {
                     }
                 }
                 if path != "." && path != ".." {
-                    try closure(path, Int32(ffd.dwFileAttributes))
+                    try closure(path.standardizingPath, Int32(ffd.dwFileAttributes))
                 }
             } while FindNextFileW(hDirectory, &ffd)
         }
@@ -222,7 +222,7 @@ extension FileManager {
             if entryType & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY {
                 let subPath: String = joinPath(prefix: path, suffix: entryName)
                 let entries = try subpathsOfDirectory(atPath: subPath)
-                contents.append(contentsOf: entries.map { joinPath(prefix: entryName, suffix: $0) })
+                contents.append(contentsOf: entries.map { joinPath(prefix: entryName, suffix: $0).standardizingPath })
             }
         })
         return contents
@@ -232,7 +232,7 @@ extension FileManager {
         var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = WIN32_FILE_ATTRIBUTE_DATA()
         return try path.withCString(encodedAs: UTF16.self) {
             if !GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes) {
-                throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+                throw _NSErrorWithWindowsError(GetLastError(), reading: true, paths: [path])
             }
             return faAttributes
         }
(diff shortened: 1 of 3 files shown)


gmittert:
cc @compnerd 

@swift-ci please test linux

TestFoundation/TestFileManager.swift
```diff
@@ -1286,89 +1311,137 @@ VIDEOS=StopgapVideos
...
-            let code = CocoaError.Code(rawValue: ($0 as? NSError)!.code)
+            let code = CocoaError.Code(rawValue: ($0 as NSError).code)
             XCTAssertEqual(code, .fileReadInvalidFileName)
         }
 
         XCTAssertThrowsError(try fm.copyItem(atPath: "", toPath: "/tmp/t"))  {
-            let code = CocoaError.Code(rawValue: ($0 as? NSError)!.code)
+            let code = CocoaError.Code(rawValue: ($0 as NSError).code)
+#if os(Windows)
+            XCTAssertEqual(code, .fileReadNoSuchFile)
```
millenomi: I think this is an issue in the error reporting — if a filename is invalid, the API should return `.fileReadInvalidFileName` no matter what the host OS thinks the right error is.

TestFoundation/TestFileManager.swift
```diff
@@ -1378,15 +1451,15 @@ VIDEOS=StopgapVideos
         XCTAssertTrue(fm.isDeletableFile(atPath: ""))
 
         XCTAssertThrowsError(try fm.attributesOfItem(atPath: "")) {
-            let code = CocoaError.Code(rawValue: ($0 as? NSError)!.code)
+            let code = CocoaError.Code(rawValue: ($0 as NSError).code)
```
millenomi: This may not build on macOS.

TestFoundation/TestFileManager.swift
gmittert: We could do that, though it would mean every time we throw because for a Win32 api call failed, we'd have to check if the error is `ERROR_FILE_NOT_FOUND` then do some checks to see if it should actually be invalid path or perhaps some other more specific error. That doesn't seem cheap to me, but it would probably only be in the exceptional case.

gmittert:
Removed the NSError changes, and have Windows follow posix such that "" is an invalid file instead of being an unfound file.

@swift-ci please test

gmittert:
@swift-ci please test

gmittert:
Rebase

@swift-ci please test linux

gmittert:
@swift-ci please test linux

compnerd:
I think that doing the remapping in the `_NSErrorWithWindowsErrno` is better.  The new function is less likely to get used I think.

gmittert:
> I think that doing the remapping in the `_NSErrorWithWindowsErrno` is better. The new function is less likely to get used I think.

Fair, I've added a `[Path]?` parameter to `_NSErrorWithWindowsErrno` instead

gmittert:
@swift-ci please test linux

compnerd:
Thanks, this LGTM.

Foundation/FoundationErrors.swift
```diff
@@ -225,8 +225,12 @@ internal func _NSErrorWithWindowsError(_ windowsError: DWORD, reading: Bool) ->
         default:
             if reading {
                 switch Int32(bitPattern: windowsError) {
-                    case ERROR_FILE_NOT_FOUND: cocoaError = .fileReadNoSuchFile
-                    case ERROR_PATH_NOT_FOUND: cocoaError = .fileReadNoSuchFile
+                    case ERROR_FILE_NOT_FOUND, ERROR_PATH_NOT_FOUND:
+                        // On an empty path, Windows will return FILE/PATH_NOT_FOUND
+                        // rather than invalid path as posix does
+                        cocoaError = paths?.contains("") ?? false
```
spevans: This might be better as `paths?.isEmpty` as `.contains("")` will always return `false`.

gmittert: `paths` is a `[String]?` so that's checking if any path it contains is empty as opposed to checking if is `"" ` itself.

Foundation/FileManager+Win32.swift
```diff
@@ -358,8 +360,7 @@ extension FileManager {
     }
 
     internal func _copyOrLinkDirectoryHelper(atPath srcPath: String, toPath dstPath: String, variant: String = "Copy", _ body: (String, String, FileAttributeType) throws -> ()) throws {
-        var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = WIN32_FILE_ATTRIBUTE_DATA()
-        do { faAttributes = try windowsFileAttributes(atPath: srcPath) } catch { return }
+        let faAttributes = try windowsFileAttributes(atPath: srcPath)
```
compnerd: How come we don't need the exception to propagate here?

Foundation/FileManager+Win32.swift
```diff
@@ -372,7 +373,7 @@ extension FileManager {
             let src = joinPath(prefix: srcPath, suffix: item)
             let dst = joinPath(prefix: dstPath, suffix: item)
 
-            do { faAttributes = try windowsFileAttributes(atPath: src) } catch { return }
+            let faAttributes = try windowsFileAttributes(atPath: src)
```
compnerd: And here

Foundation/FileManager+Win32.swift
gmittert: Previously, if srcPath didn't exist, it would catch the error from windowsFileAttributes and _copyItem would silently not do anything. This way it bubbles up the exception and throws a file not found as it should.

PR #2340, opened 2019-06-12
Fix _destinationOfSymbolicLink behaviour on Windows

gmittert:
_destinationOfSymbolic should return a relative path if the symlink is a
relative symlink. Now instead of returning the resolved symlink path, we
read the ReparsePoint to get the path it contains.

### DIFF
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -299,9 +299,92 @@ extension FileManager {
     }
 
     internal func _destinationOfSymbolicLink(atPath path: String) throws -> String {
-        return try _canonicalizedPath(toFileAtPath: path)
+        let faAttributes = try windowsFileAttributes(atPath: path)
+        guard faAttributes.dwFileAttributes & DWORD(FILE_ATTRIBUTE_REPARSE_POINT) == DWORD(FILE_ATTRIBUTE_REPARSE_POINT) else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_BAD_ARGUMENTS), reading: false)
+        }
+
+        let handle = path.withCString(encodedAs: UTF16.self) { symlink in
+            CreateFileW(symlink, GENERIC_READ, DWORD(FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE),
+                        nil, DWORD(OPEN_EXISTING), DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS),
+                        nil)
+        }
+
+        guard handle != INVALID_HANDLE_VALUE else {
+            throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+        }
+        defer { CloseHandle(handle) }
+
+        // Since REPARSE_DATA_BUFFER ends with an arbitrarily long buffer, we
+        // have to manually get the path buffer out of it since binding it to a
+        // type will truncate the path buffer.
+        //
+        // 20 is the sum of the offsets of:
+        // ULONG ReparseTag
+        // USHORT ReparseDataLength
+        // USHORT Reserved
+        // USHORT SubstituteNameOffset
+        // USHORT SubstituteNameLength
+        // USHORT PrintNameOffset
+        // USHORT PrintNameLength
+        // ULONG Flags (Symlink only)
+        let symLinkPathBufferOffset = 20 // 4 + 2 + 2 + 2 + 2 + 2 + 2 + 4
+        let mountPointPathBufferOffset = 16 // 4 + 2 + 2 + 2 + 2 + 2 + 2
+        let buff = UnsafeMutableRawBufferPointer.allocate(byteCount: Int(MAXIMUM_REPARSE_DATA_BUFFER_SIZE),
+                                                          alignment: 8)
+
+        guard let buffBase = buff.baseAddress else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_INVALID_DATA), reading: false)
+        }
+
+        var bytesWritten: DWORD = 0
+        guard DeviceIoControl(handle, FSCTL_GET_REPARSE_POINT, nil, 0,
+                              buffBase, DWORD(MAXIMUM_REPARSE_DATA_BUFFER_SIZE),
+                              &bytesWritten, nil) else {
+            throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+        }
+
+        guard bytesWritten >= MemoryLayout<REPARSE_DATA_BUFFER>.size else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_INVALID_DATA), reading: false)
+        }
+
+        let bound = buff.bindMemory(to: REPARSE_DATA_BUFFER.self)
+        guard let reparseDataBuffer = bound.first else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_INVALID_DATA), reading: false)
+        }
+
+        guard reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_SYMLINK
+                || reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_MOUNT_POINT else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_BAD_ARGUMENTS), reading: false)
+        }
+
+        let pathBufferPtr: UnsafeMutableRawPointer
+        let substituteNameBytes: Int
+        let substituteNameOffset: Int
+        switch reparseDataBuffer.ReparseTag {
+            case IO_REPARSE_TAG_SYMLINK:
+                pathBufferPtr = buffBase + symLinkPathBufferOffset
+                substituteNameBytes = Int(reparseDataBuffer.SymbolicLinkReparseBuffer.SubstituteNameLength)
+                substituteNameOffset = Int(reparseDataBuffer.SymbolicLinkReparseBuffer.SubstituteNameOffset)
+            case IO_REPARSE_TAG_MOUNT_POINT:
+                pathBufferPtr = buffBase + mountPointPathBufferOffset
+                substituteNameBytes = Int(reparseDataBuffer.MountPointReparseBuffer.SubstituteNameLength)
+                substituteNameOffset = Int(reparseDataBuffer.MountPointReparseBuffer.SubstituteNameOffset)
+            default:
+                throw _NSErrorWithWindowsError(DWORD(ERROR_BAD_ARGUMENTS), reading: false)
+        }
+
+        guard substituteNameBytes + substituteNameOffset <= bytesWritten else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_INVALID_DATA), reading: false)
+        }
+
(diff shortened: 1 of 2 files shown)


Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,59 @@ extension FileManager {
...
+            MemoryLayout<ULONG>.size //ULONG ReparseTag
+            + MemoryLayout<USHORT>.size //ULONG ReparseDataLength
+            + MemoryLayout<USHORT>.size //ULONG Reserved
+            + MemoryLayout<USHORT>.size //ULONG SubstituteNameOffset
+            + MemoryLayout<USHORT>.size //ULONG SubstituteNameLength
+            + MemoryLayout<USHORT>.size //ULONG PrintNameOffset
+            + MemoryLayout<USHORT>.size //ULONG PrintNameLength
+            + MemoryLayout<ULONG>.size //ULONG Flags
+
+        var buff = Array(repeating: 0, count: MemoryLayout<REPARSE_DATA_BUFFER>.size + Int(2 * MAX_PATH))
```
millenomi: This is not okay. The address of an `Array` is already bound to `Int`, and you're in deep undefined behavior territory. Use an `UnsafeRawMutableBuffer` instead.

gmittert: Sure, I'll swap it. However, what do you mean by "The address of an `Array` is already bound to `Int`"?

gmittert: Actually, I should just be able to pass `&buff` to `DeviceIoControl`, that should be fine, right?

gmittert:
Changed to use a UnsafeMutableRawBufferPointer

@swift-ci please test

gmittert:
@swift-ci please test

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,58 @@ extension FileManager {
...
+            MemoryLayout<ULONG>.size //ULONG ReparseTag
+            + MemoryLayout<USHORT>.size //ULONG ReparseDataLength
+            + MemoryLayout<USHORT>.size //ULONG Reserved
+            + MemoryLayout<USHORT>.size //ULONG SubstituteNameOffset
+            + MemoryLayout<USHORT>.size //ULONG SubstituteNameLength
+            + MemoryLayout<USHORT>.size //ULONG PrintNameOffset
+            + MemoryLayout<USHORT>.size //ULONG PrintNameLength
+            + MemoryLayout<ULONG>.size //ULONG Flags
+
+        let buffSize = MemoryLayout<REPARSE_DATA_BUFFER>.size + Int(2 * MAX_PATH) * MemoryLayout<wchar_t>.size
```
compnerd: Why `2 * MAX_PATH`?  Can you use `WCHAR` rather than `wchar_t` please?  `wchar_t` is differently sized on different targets, `WCHAR` is a typedef for `unsigned short`.

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,58 @@ extension FileManager {
...
+        let buff = UnsafeMutableRawBufferPointer.allocate(byteCount: buffSize, alignment: 8)
```
compnerd: Does the `allocate` ensure that the allocation succeeded?

Foundation/FileManager+Win32.swift
adierking: Microsoft actually defines a `MAXIMUM_REPARSE_DATA_BUFFER_SIZE` constant which we should use here. `MAX_PATH` is irrelevant.

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,58 @@ extension FileManager {
...
+
+        var bytesWritten: DWORD = 0
+        guard DeviceIoControl(handle, FSCTL_GET_REPARSE_POINT, nil, 0,
```
adierking: Should we check that the file is actually a reparse point before we issue this ioctl? It might make failures easier to understand.

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,58 @@ extension FileManager {
...
+        }
+
+        let bound = buff.bindMemory(to: REPARSE_DATA_BUFFER.self)
+        guard let reparseDataBuffer = bound.first else {
+            fatalError()
+        }
+
+        let pathBufferPtr = buff.baseAddress! + pathBufferOffset
+        let printNameBytes = Int(reparseDataBuffer.SymbolicLinkReparseBuffer.PrintNameLength)
+        let printNameOffset = Int(reparseDataBuffer.SymbolicLinkReparseBuffer.PrintNameOffset)
```
adierking: 1. This is assuming that the reparse point is a symbolic link. Not all reparse points are symbolic links, or even links to files at all. Please verify the reparse tag before doing anything here. If you want an example of a reparse point which is not a link, connect a OneDrive account on Windows 10 and look in your local OneDrive directory. The files there all use a private reparse tag. (Also, do we want to handle mount points here? They can be links to directories.)

2. Why is this using the print name? The print name is only intended to be used for display purposes.

3. Do we need to validate that the name offset and length actually make sense? We shouldn't assume that reparse data is well-formed. AFAIK, UnsafeMutableRawBufferPointer is not always bounds-checked.

Foundation/FileManager+Win32.swift
gmittert: If it fails, as far as I know, it doesn't give any indication.

gmittert:
- Used MAXIMUM_REPARSE_DATA_BUFFER_SIZE 
- Check if it's actually a reparse point
- Add support for Mount points
- Add extra bounds checking
- Hard coded the offset sizes

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,80 @@ extension FileManager {
...
+            throw _NSErrorWithWindowsError(DWORD(ERROR_INVALID_DATA), reading: false)
+        }
+
+        guard reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_SYMLINK
+                || reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_MOUNT_POINT else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_BAD_ARGUMENTS), reading: false)
+        }
+
+        let (pathBufferPtr, substituteNameBytes, substituteNameOffset) = reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_SYMLINK
+            ? (buff.baseAddress! + symLinkPathBufferOffset,
```
spevans: Maybe better to check if `buff.baseAddress` is `nil` and `throw` and error rather than force-unwrapping?

gmittert: Seems like a good idea to me, changed.

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,84 @@ extension FileManager {
     }
 
     internal func _destinationOfSymbolicLink(atPath path: String) throws -> String {
-        return try _canonicalizedPath(toFileAtPath: path)
+        let faAttributes = try windowsFileAttributes(atPath: path)
```
compnerd: Should the error here not be propagated?

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,84 @@ extension FileManager {
...
+        guard faAttributes.dwFileAttributes & DWORD(FILE_ATTRIBUTE_REPARSE_POINT) != 0 else {
```
compnerd: Can you spell this as:

`guard faAttributes.dwFileAttributes & DWORD(FILE_ATTRIBUTE_REPARSE_POINT) == DWORD(FILE_ATTRIBUTE_REPARSE_POINT) else {`

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,84 @@ extension FileManager {
...
+            throw _NSErrorWithWindowsError(DWORD(ERROR_BAD_ARGUMENTS), reading: false)
+        }
+
+        let handle = path.withCString(encodedAs: UTF16.self) { symlink in
+            CreateFileW(symlink, GENERIC_READ, DWORD(FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE),
+                        nil, DWORD(OPEN_EXISTING), DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS),
+                        nil)
+        }
+
+        guard handle != INVALID_HANDLE_VALUE else {
```
compnerd: Any reason to not make this:

```
  if handle == INVALID_HANDLE_VALUE {
  }
```

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,84 @@ extension FileManager {
...
+        //
+        // 20 is the sum of the offsets of:
+        // ULONG ReparseTag
+        // USHORT ReparseDataLength
+        // USHORT Reserved
+        // USHORT SubstituteNameOffset
+        // USHORT SubstituteNameLength
+        // USHORT PrintNameOffset
+        // USHORT PrintNameLength
+        // ULONG Flags (Symlink only)
```
compnerd: *cry*

Foundation/FileManager+Win32.swift
compnerd: Yeah, that sounds like a good idea.  Although, this is an internal function, so I wonder if that has been verified earlier in the call chain

Foundation/FileManager+Win32.swift
```diff
@@ -299,9 +299,84 @@ extension FileManager {
...
+        guard let reparseDataBuffer = bound.first else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_INVALID_DATA), reading: false)
+        }
+
+        guard reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_SYMLINK
+                || reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_MOUNT_POINT else {
+            throw _NSErrorWithWindowsError(DWORD(ERROR_BAD_ARGUMENTS), reading: false)
+        }
+
+        let (pathBufferPtr, substituteNameBytes, substituteNameOffset) = reparseDataBuffer.ReparseTag == IO_REPARSE_TAG_SYMLINK
```
compnerd: Can we use a `switch` here instead?

Foundation/FileManager+Win32.swift
gmittert: windowsFileAttributes throws if the file is not found which is what we also want to do.

Foundation/FileManager+Win32.swift
gmittert: When I know I'm always going to throw, I prefer using a guard for the semantics of "after this point `handle` is guaranteed to be a valid handle".

Foundation/FileManager+Win32.swift
gmittert: Sure! `Switch`ed it.

gmittert:
@swift-ci please test linux

adierking:
The Windows code looks fine to me now. cc @compnerd

PR #2341, opened 2019-06-13
Parity: URLCache

millenomi:
This strictly implements the class. We need a separate adoption patch to hook it up to `_HTTPProtocol`, `URLSession` and `URLSessionConfiguration`. It has on-disk and in-memory caching. This also fixes a couple issues with archiving.

### DIFF
--- a/Foundation/URLCache.swift
+++ b/Foundation/URLCache.swift
@@ -13,6 +13,13 @@ import SwiftFoundation
 import Foundation
 #endif
 
+internal extension NSLock {
+    func performLocked<T>(_ block: () throws -> T) rethrows -> T {
+        lock(); defer { unlock() }
+        return try block()
+    }
+}
+
 /*!
     @enum URLCache.StoragePolicy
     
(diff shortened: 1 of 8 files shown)


millenomi:
cc @drodriguez @karthikkeyan

millenomi:
@swift-ci please test linux

millenomi:
I'm strongly considering `_FTPProtocol` not employing caching.

millenomi:
I mean obviously `_HTTPURLProtocol`/`_FTPURLProtocol`, whoops.

karthikkeyan:
@millenomi Something like this,

```swift
func evictFromMemoryCacheAssumingLockHeld(maximumSize: Int) {
    var totalSize = 0
    var orderedEntryProperties: [(cost: Int, index: Int)] = []
    for (index, identifier) in inMemoryCacheOrder.enumerated() {
        let cost = inMemoryCacheContents[identifier]!.cost
        orderedEntryProperties.append((cost, index))
        totalSize += cost
    }
    
    guard totalSize > maximumSize else { return }
    
    var i = 0
    for entryProperty in orderedEntryProperties {
        inMemoryCacheContents.removeValue(forKey: inMemoryCacheOrder[entryProperty.index])
        totalSize -= entryProperty.cost
        if totalSize < maximumSize {
            break
        }
        
        i += 1
    }
    inMemoryCacheOrder = inMemoryCacheOrder[(i + 1)..<inMemoryCacheOrder.count]
}
```

This way we can reduce the 6 different loops down to just 2. I am not sure if it is worth doing, but what you think about the above approach?

Foundation/URLCache.swift
```diff
@@ -225,36 +503,91 @@ open class URLCache : NSObject {
...
+            var urlsToRemove: [URL] = []
+            enumerateDiskEntries { (entry, stop) in
+                if entry.date > date {
+                    urlsToRemove.append(entry.url)
+                }
+            }
+            
+            for url in urlsToRemove {
+                try? FileManager.default.removeItem(at: url)
+            }
```
drodriguez: From the implementation of `enumerateDiskEntries`, since the `contentsOfDirectory(at:…)` are generated before invoking the block, I think you can merge both loops into one and forget about `urlsToRemove`.

```swift
enumerateDiskEntries { entry, _ in
  if entry.date > data {
    try? FileManager.default.removeItem(at: url)
  }
}
```

millenomi:
Reviving this.

millenomi:
Actually: I'm going to close this, and post a consolidated PR that includes the HTTP correctness fixes discussed above (but not your suggestions yet, @drodriguez — I'm going to go through them.)

millenomi:
Moved to #2401.

PR #2342, opened 2019-06-13
Disable test_ftpDataTaskDelegate crashing occasionally in CI

benlangmuir:
Looks like the same issue as test_ftpDataTask, so marking with the same cause.

https://bugs.swift.org/browse/SR-10922


benlangmuir:
@swift-ci please test Linux

PR #2344, opened 2019-06-14
Fallback to GetFullPathNameW for Windows _canonicalizedPath

gmittert:
If a file is not found, rather than failing, we should use
GetFullPathNameW which doesn't resolve symlinks, but will work on paths
that don't exist.

### DIFF
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -311,8 +311,16 @@ extension FileManager {
                               nil, DWORD(OPEN_EXISTING), DWORD(FILE_FLAG_BACKUP_SEMANTICS),
                               nil)
         }
-        if hFile == INVALID_HANDLE_VALUE {
-            throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+        guard hFile != INVALID_HANDLE_VALUE else {
+            return try path.withCString(encodedAs: UTF16.self) {
+                var dwLength = GetFullPathNameW($0, 0, nil, nil)
+                var szPath = Array<WCHAR>(repeating: 0, count: Int(dwLength + 1))
+                dwLength = GetFullPathNameW($0, DWORD(szPath.count), &szPath, nil)
+                guard dwLength > 0 && dwLength <= szPath.count else {
+                    throw _NSErrorWithWindowsError(GetLastError(), reading: true)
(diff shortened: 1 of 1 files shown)


gmittert:
@swift-ci please test

Foundation/FileManager+Win32.swift
```diff
@@ -312,7 +312,20 @@ extension FileManager {
                               nil)
         }
         if hFile == INVALID_HANDLE_VALUE {
-            throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+            let error = GetLastError()
+            if error == ERROR_FILE_NOT_FOUND || error == ERROR_PATH_NOT_FOUND {
+                return try path.withCString(encodedAs: UTF16.self) {
+                    let dwLength = GetFullPathNameW($0, 0, nil, nil)
+                    var szPath = Array<WCHAR>(repeating: 0, count: Int(dwLength + 1))
+                    let result = GetFullPathNameW($0, dwLength + 1, &szPath, nil)
```
compnerd: I think I prefer `GetFullPathNameW($0, szPath.count, &szPath, nil)`

Foundation/FileManager+Win32.swift
```diff
@@ -312,7 +312,20 @@ extension FileManager {
...
+                    guard result != 0 && result <= dwLength + 1 else {
```
compnerd: I think that `result <= szPath.count` is nicer, in fact, at that point you could change `let result` into `dwLength` and get something like:

`guard dwLength > 0 && dwLength <= szPath.count`

which feels more obvious

Foundation/FileManager+Win32.swift
```diff
@@ -312,7 +312,20 @@ extension FileManager {
...
+                        throw _NSErrorWithWindowsError(GetLastError(), reading: true)
+                    }
+                    return String(decodingCString: szPath, as: UTF16.self)
+                }
+            } else {
```
compnerd: I think that the else { can be elided. You will always get a result or a failure from the error recovery path.

gmittert:
Updated for feedback :) It does seem nicer to use szPath.count.

@swift-ci please test

gmittert:

gmittert:

PR #2347, opened 2019-06-15
Look for abbreviations when creating time zones.

drodriguez:
In the case the code ends up in the fallback that uses localtime_r (in
Unixes), the returned value in tm_zone is an abbreviation, but the code
that created the CFTimeZoneRef was not looking for abbreviations.

This code was hit in Android all the time (since it doesn't have an
Olson TZ file structure), but I didn't realized because my test machine
were using UTC instead of an actual time zone.


drodriguez:

drodriguez:

drodriguez:
This change fixes the failing `TestCalendar` and `TestTimeZone` tests in Android.

PR #2348, opened 2019-06-15
[android] Use timespec like in Linux for URL c/m/atime.

drodriguez:
Android has timespec like Linux, so it can use the Linux codepaths just
fine. The proposed code path actually fail because those accessors are
actually macros in the Android sysroot.


spevans:
@swift-ci test linux

PR #2349, opened 2019-06-15
[android] Provide implementation for Android Bionic.

drodriguez:
In Android the last element of the sockaddr_in struct is actually called
__pad instead of sin_zero, so a special codepath is needed.

/cc @saiHemak


spevans:
@swift-ci test and merge

spevans:
@swift-ci test linux

PR #2350, opened 2019-06-16
Base: include headers appropriately for android

compnerd:
Android has the newer syscalls available but the declarations were not
being pulled in.  Add the appropriate handling.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -55,7 +55,10 @@
 #include <malloc/malloc.h>
 #endif
 
-#if TARGET_OS_LINUX
+#if TARGET_OS_ANDROID
+#include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @spevans @drodriguez

compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please test

spevans:
LGTM

buttaface:
Yes, for some reason, this same patch was posted twice.

compnerd:
I pushed the wrong patch to #2354, I'll fix that.  As to your comment there about this patch - I think that moving these headers into the source file is beyond the scope of this change, but doing that in a subsequent change is reasonable.  I believe that this header has the additional includes because it was originally modelled around `ForFoundationOnly.h`.  This header serves as the point for bridging additional headers into the Swift side as well as for the local declarations which provide local C implementation needed to support the Swift side.  If there are declarations here which are unnecessary, we could prune them.

PR #2351, opened 2019-06-16
build: use `target_include_directories` (NFC)

compnerd:
Use `target_include_directories` rather than `target_compile_options`
This should be entirely equivalent.  NFCI.


compnerd:

PR #2352, opened 2019-06-16
CoreFoundation: use local uuid library

compnerd:


compnerd:
@swift-ci please test

compnerd:
Because the target wont be created until after that CMakeList is processed.  Furthermore, there is no reason to assume that it has access to uuid - consider the OpenDarwin folks who use the CoreFoundation standalone build to target the non-Swift runtime.

PR #2353, opened 2019-06-16
Base: implement KnownLocations for android

compnerd:
Add support for android to KnownLocations.  Assume that only current
user is supported as all applications in android are isolated and single
user.  For now, place all files in ~/Apple/Library/Preferences.  This is
the last piece needed in CoreFoundation for Android support.

### DIFF
--- a/CoreFoundation/Base.subproj/CFKnownLocations.c
+++ b/CoreFoundation/Base.subproj/CFKnownLocations.c
@@ -91,6 +91,26 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
             break;
     }
 
+#elif TARGET_OS_ANDROID
+
+    switch (user) {
+    case _kCFKnownLocationUserAny:
+    case _kCFKnownLocationUserByName:
+      abort();
+    case _kCFKnownLocationUserCurrent: {
+      const char *buffer = getenv("CFFIXED_USER_HOME");
+      if (buffer == NULL || *buffer = '\0') {
+        CFLog(__kCFLogAssertion, CFSTR("CFFIXED_USER_HOME is unset"));
+        HALT;
+      }
+
+      CFURLRef userdir = CFURLCreateFromFileSystemRepresentation(kCFAllocatorSystemDefault, (const unsigned char *)buffer, strlen(buffer), true);
+      location = CFURLCreateWithFileSystemPathRelativeToBase(kCFAllocatorSystemDefault, CFSTR("/Apple/Library/Preferences"), kCFURLPOSIXPathStyle, true, userdir);
+      CFRelease(userdir);
+      break;
+    }
+    }
+
 #else
     
     #error For this platform, you need to define a preferences path for both 'any user' (i.e. installation-wide preferences) or the current user.


compnerd:
CC: @millenomi @drodriguez

compnerd:
@swift-ci please test

spevans:
@swift-ci test linux

CoreFoundation/Base.subproj/CFKnownLocations.c
```diff
@@ -91,6 +91,20 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
...
+
+    switch (user) {
+    case _kCFKnownLocationUserAny:
+    case _kCFKnownLocationUserByName:
+      abort();
+    case _kCFKnownLocationUserCurrent: {
+      CFURLRef home = CFCopyHomeDirectoryURLForUser(CFGetUserName());
+      location = CFURLCreateWithFileSystemPathRelativeToBase(kCFAllocatorSystemDefault, CFSTR("/Apple/Library/Preferences"), kCFURLPOSIXPathStyle, true, home);
+      CFRelease(home);
+      break;
```
drodriguez: I'm not sure this is going to work on the device. The home directory in Android is actually the root directory, and it is not writable by the user, so you will not be able to create any new directories there. Have you tried this on-device?

compnerd: No, I don't have a dvice on hand right now.  The desire here is to get the path to the application's data space.  What would be the way to get that?

drodriguez: There's no such thing in Android, sadly. Android (Java) has the `SharedPreferences` type, but the physical storage is not public. If you want something that will work, you can use the temporary directory, which is readable/writable by any process. I really don’t think you can give a value here that makes sense in Android.

compnerd: Can we assume that we have the Java runtime initialized and can just JNI to get the value?  Using tmp seems .... less than ideal.  This is expected to be used for preserved data.

drodriguez: You still have the problem that you want to return a path, while `SharedPrefereces` is a database.

One solution use in the HTTP support is checking an environment variable to pick up the certificates from an app known location. Maybe this needs a similar solution for Android. Only the app can decide which file system access they are going to have, and where it is going to happen. Even if you have access to JNI, for many APIs you need a reference to `Context`, which will be difficult to obtain.

```java
// A some moment during the start of your Android app
Context context = … // this is the piece that will be quite difficult to obtain in CF
File filesDir = context.getFilesDir();
android.system.Os.setenv("CFAndroidFileDirPath", filesDir.getAbsolutePath(), true);
```

```c
// In this file, or somewhere in CF
const char *filesDirPtr = getenv("CFAndroidFileDirPath");
CFStringRef filesDirPath = CFStringCreateWithCString(NULL, filesDirPtr, kCFStringEncodingUTF16); // NOTE: I think Java uses UTF-16
CFURLRef filesDirURL = CFURLCreateWithFileSystemPAth(NULL, filesDirPath, kCFURLPOSIXPathStyle, true);
location = CFURLCreateWithFileSystemPathRelativeToBase(kCFAllocatorSystemDefault, CFSTR("/Library/Preferences"), kCFURLPOSIXPathStyle, true, filesDirURL);
```

compnerd: Sure, I think that using an environment variable is plausible, but, what would the developer set that to?  This is the *common* case - I just want persistent storage of files

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

millenomi:
My proposal is:

Let’s revert to the non-abort patch and mandate that people set `CFFIXED_USER_HOME` via `Os.setenv()` before any Swift code is invoked.

millenomi:
(That way, they can source the internal storage path from an activity without us guessing.)

millenomi:
This can be enforced by `abort()`-ing here or in `__CFInitialize`.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

CoreFoundation/Base.subproj/CFKnownLocations.c
```diff
@@ -91,6 +91,21 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
...
 
+#elif TARGET_OS_ANDROID
+
+    switch (user) {
+    case _kCFKnownLocationUserAny:
+    case _kCFKnownLocationUserByName:
+      abort();
+    case _kCFKnownLocationUserCurrent: {
+      const char *buffer = getenv("CFFIXED_USER_HOME");
+      CFURLRef userdir = CFURLCreateFromFileSystemRepresentation(kCFAllocatorSystemDefault, (const unsigned char *)buffer, strlen(buffer), true);
```
millenomi: I think we should ensure the environment variable is none,pry and trap?

compnerd: I was going for the environment variable is unset, the `CFURLCreateFromFSR` will abort

drodriguez:
I also approve the environment variable approach (with the last comment by Lily implemented).

CoreFoundation/Base.subproj/CFKnownLocations.c
compnerd: Made it more explicit

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

millenomi:
While we’re using it directly here, it is also already picked up by CFPlatform.c, so API like NSHomeDirectory(), FileManager etc. will all heed it.

PR #2357, opened 2019-06-17
build: allow passing Swift specific link flags

compnerd:
This mirrors the `CMAKE_SWIFT_FLAGS` allowing control over the flags
passed to the linker.  This is needed until the libraries are built
with a newer CMake which supports Swift properly.  This enables
cross-compiling libdispatch for Android on Windows.


compnerd:
@swift-ci please test

PR #2358, opened 2019-06-17
inline some flag computation

compnerd:


compnerd:
@swift-ci please test

PR #2359, opened 2019-06-17
[test] Remove external URLs from tests.

drodriguez:
Some test were using external URLs. External URLs are not ideal for
testing for serveral reasons: they can change without warning, making
the test start failing, they take longer to resolve, and finally they
will not work if the test machine do not have internal access for
security reasons.

The commit modifies those URL to hit the in-memory HTTP server, setting
each request with a 5 second pause to allow to check the needed
behaviours of pausing and resuming.

/cc @pushkarnk @karthikkeyan

### DIFF
--- a/TestFoundation/HTTPServer.swift
+++ b/TestFoundation/HTTPServer.swift
@@ -52,7 +52,7 @@ class _TCPSocket {
 #endif
     private var listenSocket: SOCKET!
     private var socketAddress = UnsafeMutablePointer<sockaddr_in>.allocate(capacity: 1) 
-    private var connectionSocket: SOCKET!
+    private var connectionSocket: SOCKET?
     
     private func isNotNegative(r: CInt) -> Bool {
         return r != -1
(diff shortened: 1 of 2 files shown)


drodriguez:

gottesmm:
+1!

@drodriguez I wonder is there a way where we could ensure that if we aren't going to 127.0.0.1 in a test process, it crashes. (Maybe some form of sandboxing, not sure). Then we can be sure these never sneak in /ever/ again. = ).

karthikkeyan:
Thanks for make this changes @drodriguez. I wasn't aware of the constraints around external URL, I will be conscious of this henceforth.

pushkarnk:
Thanks @drodriguez I agree. While I was aware of this unwritten rule, I must admit I ignored it while reviewing those tests!

pushkarnk:
@swift-ci please test

drodriguez:
@gottesmm: In Linux one can use `unshare` to do something like you propose (ref: http://manpages.ubuntu.com/manpages/bionic/man1/unshare.1.html). It might be nice to run the test under that, but one needs to also open loopback again, and I don’t know if a normal user can do all that. If someone has some time, it might be worth looking into it. There's another test in `TestHost.swift` that resolves a real URL, and I haven’t been able to find out if it should work without connection or not.

I will review the error in CI (seems that the socket is used after being closed).

drodriguez:
@swift-ci please test

drodriguez:
@swift-ci please test

drodriguez:
Rebased on top of current master. Can I have a thumbs up/approval in this PR if there's no more feedback?

Might fail because of the fix in #2428.

drodriguez:
The failure was because of the missing PR. Re-testing since that PR has been merged.

PR #2361, opened 2019-06-18
Very rough prototype (with some bits that rely on a feature that hasn't landed yet stubbed out) of an alternate design for JSONWriter that could be faster. Untested, mostly just for discussion.

Catfish-Man:
This could, in theory at least, recover some static dispatch optimizations through liberal use of @_specialize. I have no idea if it's better or not since a bunch of stuff it uses isn't in the compiler yet, but it's kinda interesting.

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -28,6 +28,253 @@ extension JSONSerialization {
     }
 }
 
+fileprivate protocol _JSONValue {
+    func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws
+}
+
+//@_specialize(where T == String)
+//extension<T: _JSONValue> [String : T] : _JSONValue {
+extension Dictionary : _JSONValue where Key == String, Value : _JSONValue {
+    
+//    @_specialize(where Visitor == JSONWriter)
+//    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        try visitor.visit(self)
+    }
+}
+
+//@_specialize(where T == String)
+//extension<T: _JSONValue> [T] : _JSONValue {
+extension Array : _JSONValue where Element : _JSONValue {
+    
+//    @_specialize(where Visitor == JSONWriter)
+//    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        try visitor.visit(self)
+    }
+}
+
+extension String : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
+    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        try visitor.visit(self)
+    }
+}
+
+//@_specialize(where T == Int)
+//@_specialize(where T == UInt)
+//@_specialize(where T == Int32)
+//@_specialize(where T == UInt32)
+//@_specialize(where T == Int64)
+//@_specialize(where T == UInt64)
+//extension<T: FixedWidthInteger> T : _JSONValue {
+extension Int : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
+    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        try visitor.visit(self)
+    }
+}
+
+//@_specialize(where T == Double)
+//@_specialize(where T == Float)
+//extension<T: FloatingPoint> T : _JSONValue {
+extension Float : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
+    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        guard isFinite && !isNaN else {
+            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Invalid number value (\(self)) in JSON write"])
+        }
+        try visitor.visit(self)
+    }
+}
+
+//TODO: delete this, use parameterized extensions
+extension Double : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
+    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        guard isFinite && !isNaN else {
+            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Invalid number value (\(self)) in JSON write"])
+        }
+        try visitor.visit(self)
+    }
+}
+
+extension Bool : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
+    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        try visitor.visit(self)
+    }
+}
+
+extension Decimal : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
+    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        guard isFinite else {
+            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Invalid number value (\(self)) in JSON write"])
+        }
+        try visitor.visit(self)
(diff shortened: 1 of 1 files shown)


Foundation/JSONSerialization.swift
```diff
@@ -28,6 +28,253 @@ extension JSONSerialization {
...
+fileprivate protocol _JSONValue {
+    func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws
+}
+
+//@_specialize(where T == String)
+//extension<T: _JSONValue> [String : T] : _JSONValue {
+extension Dictionary : _JSONValue where Key == String, Value : _JSONValue {
+    
+//    @_specialize(where Visitor == JSONWriter)
+//    @_specialize(where Visitor == JSONValidator)
```
Catfish-Man: @_specialize isn't clever enough to recover the Key and Value constraints from the where clause above, and also doesn't support the 'Value : _JSONValue' constraint currently

Foundation/JSONSerialization.swift
```diff
@@ -28,6 +28,253 @@ extension JSONSerialization {
...
+    }
+}
+
+//@_specialize(where T == Int)
+//@_specialize(where T == UInt)
+//@_specialize(where T == Int32)
+//@_specialize(where T == UInt32)
+//@_specialize(where T == Int64)
+//@_specialize(where T == UInt64)
+//extension<T: FixedWidthInteger> T : _JSONValue {
```
Catfish-Man: Unclear if FixedWidthInteger is what I want here, but this could be accomplished with gyb instead

Foundation/JSONSerialization.swift
```diff
@@ -28,6 +28,253 @@ extension JSONSerialization {
...
+//    @_specialize(where Visitor == JSONWriter)
+//    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        try visitor.visit(self)
+    }
+}
+
+extension String : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
```
Catfish-Man: Does this generate a bunch of expensive type checks that outweigh the static dispatch? Who knows!

Foundation/JSONSerialization.swift
```diff
@@ -28,6 +28,253 @@ extension JSONSerialization {
...
+            case CFBooleanGetTypeID():
+                try boolValue.visit(with: &visitor)
+            default:
+                try intValue.visit(with: &visitor)
+            }
+        }
+    }
+}
+
+extension NSArray : _JSONValue {
```
Catfish-Man: This should actually be a big win. No need to bridge NSArrays and NSDictionaries at all when you can just iterate through them directly

Foundation/JSONSerialization.swift
```diff
@@ -28,6 +28,253 @@ extension JSONSerialization {
...
+        try visitor.visit(self) //typecheck lazily as we go
+    }
+}
+
+extension NSString : _JSONValue {
+    
+    @_specialize(where Visitor == JSONWriter)
+    @_specialize(where Visitor == JSONValidator)
+    fileprivate func visit<Visitor: JSONVisitor>(with visitor: inout Visitor) throws {
+        try visitor.visit(self as String)
```
Catfish-Man: Bridging here could be done lazily as well at the cost of having yet another override point on JSONVisitor. Dunno if it's worth it.

Foundation/JSONSerialization.swift
```diff
@@ -28,6 +28,253 @@ extension JSONSerialization {
...
+    mutating func visit(_ cocoaDictionary: NSDictionary) throws
+}
+
+extension JSONVisitor {
+    mutating func visit<T>(topLevelJSON value: T?) throws {
+        guard let value = value else {
+            return
+        }
+        if let dict = value as? [String : _JSONValue] {
+            try visit(dict) //TODO: should be dict.visit(with: &self)
```
Catfish-Man: This is because _JSONValue doesn't conform to _JSONValue. Ugh.

Foundation/JSONSerialization.swift
```diff
@@ -28,6 +28,253 @@ extension JSONSerialization {
...
+                try jsonKey.visit(with: &self)
+                try jsonValue.visit(with: &self)
+            } catch where error is NSError {
+                err = (error as! NSError)
+                stopPtr.pointee = true
+            } catch {
+                fatalError()
+            }
+        }
+        if let error = err {
```
Catfish-Man: This is because enumerateKeysAndObjects doesn't rethrow

Foundation/JSONSerialization.swift
```diff
@@ -285,219 +454,176 @@ internal extension JSONSerialization {
...
 
 //MARK: - JSONSerializer
-private struct JSONWriter {
+private struct JSONWriter : JSONVisitor {
 
     var indent = 0
     let pretty: Bool
     let sortedKeys: Bool
-    let writer: (String?) -> Void
+    var result = ""
```
Catfish-Man: Make the optimizer's job a little easier. Nothing was using this flexibility currently.

Foundation/JSONSerialization.swift
```diff
@@ -285,219 +454,176 @@ internal extension JSONSerialization {
...
-        var toSerialize = object
+        var toSerialize = any
 
         if let number = toSerialize as? _NSNumberCastingWithoutBridging {
             toSerialize = number._swiftValueOfOptimalType
         }
         
         guard let obj = toSerialize else {
-            try serializeNull()
+            try NSNull().visit(with: &self)
```
Catfish-Man: I think NSNull() allocates in Swift where it doesn't in ObjC sadly

Foundation/JSONSerialization.swift
```diff
@@ -285,219 +454,176 @@ internal extension JSONSerialization {
...
-            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Invalid object cannot be serialized"])
+        guard let jsonObj = obj as? _JSONValue else {
+            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue, userInfo: ["NSDebugDescription" : "Invalid value (\(obj)) in JSON write"])
         }
+        
+        try jsonObj.visit(with: &self)
+    }
+    
+    mutating func visit<I: FixedWidthInteger> (_ int: I) {
+        result += int.description //TODO
```
Catfish-Man: Relying on .description just doesn't seem right to me… I think we have some nice dedicated stuff for this now?

Foundation/JSONSerialization.swift
```diff
@@ -285,219 +454,176 @@ internal extension JSONSerialization {
...
+    
+    mutating func visit(_ decimal: Decimal) {
+        result += decimal.description
     }
 
-    func serializeString(_ str: String) throws {
-        writer("\"")
+    mutating func visit(_ str: String) throws {
+        result += "\""
         for scalar in str.unicodeScalars {
```
Catfish-Man: This method could maybe use special handling for ASCII? Profiling will tell

Foundation/JSONSerialization.swift
```diff
@@ -285,219 +454,176 @@ internal extension JSONSerialization {
...
+        result += "["
         if pretty {
-            writer("\n")
+            result += "\n"
             incAndWriteIndent()
         }
         
         var first = true
         for elem in array {
+            guard let jsonElem = elem as? _JSONValue else {
```
Catfish-Man: Have to make sure this gets optimized out when it's called from visit(array) rather than visit(nsarray). It *should*… but it's worth double checking.

Foundation/JSONSerialization.swift
```diff
@@ -285,219 +454,176 @@ internal extension JSONSerialization {
...
     }
 
-    mutating func serializeDictionary(_ dict: Dictionary<AnyHashable, Any?>) throws {
-        writer("{")
+    mutating func visit<J: _JSONValue>(_ array: [J]) throws {
+        try visitArrayLikeSequence(array)
+    }
+    
+    mutating func visit(_ cocoaArray: NSArray) throws {
+        try visitArrayLikeSequence(cocoaArray)
```
Catfish-Man: Should do a visitDictionaryLikeSequence one too to avoid bridging NSDictionary just to write it out

Catfish-Man:
Parameterized extensions is in https://github.com/apple/swift/pull/25263

Foundation/JSONSerialization.swift
millenomi: I don't know if this is worth a radar, but maybe it is?

bendjones:
Just noticed this (apologies @Catfish-Man) I'll take a closer review.

Foundation/JSONSerialization.swift
bendjones: Is there somewhere I could read about this new feature?

Foundation/JSONSerialization.swift
bendjones: Yeah I think we want to move away from description in general.

bendjones: Well in Obj-C I believe but not sure in Swift's case...

Foundation/JSONSerialization.swift
bendjones: We chatted on slack about this

shahmishal:

PR #2362, opened 2019-06-18
[SR-10960] URLSessionConfiguration.background(withIdentifier) method implemented

karthikkeyan:
Bug: https://bugs.swift.org/browse/SR-10960

### DIFF
--- a/Foundation/URLSession/URLSessionConfiguration.swift
+++ b/Foundation/URLSession/URLSessionConfiguration.swift
@@ -156,7 +156,17 @@ open class URLSessionConfiguration : NSObject, NSCopying {
         return ephemeralConfiguration
     }
 
-    open class func background(withIdentifier identifier: String) -> URLSessionConfiguration { NSUnimplemented() }
+    open class func background(withIdentifier identifier: String) -> URLSessionConfiguration {
+        if identifier.isEmpty {
+            // As per document: The unique identifier for the configuration object. This parameter must not be nil or an empty string.
+            // But it is unclear what happens when it is empty, the existing Foundation framework creates an instance anyway.
+            // FIXME: Handle empty string properly
+        }
+        let backgroundConfiguration = URLSessionConfiguration.default.copy() as! URLSessionConfiguration
+        backgroundConfiguration.identifier = identifier
+        backgroundConfiguration.urlCache = nil
+        return backgroundConfiguration
+    }
     
     /* identifier for the background session configuration */
     open var identifier: String?
--- a/Foundation/URLSession/URLSessionConfiguration.swift
+++ b/Foundation/URLSession/URLSessionConfiguration.swift
@@ -156,17 +156,8 @@ open class URLSessionConfiguration : NSObject, NSCopying {
         return ephemeralConfiguration
     }
 
-    open class func background(withIdentifier identifier: String) -> URLSessionConfiguration {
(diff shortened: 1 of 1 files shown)


karthikkeyan:
cc @pushkarnk

pushkarnk:
@karthikkeyan Thanks for taking this up. I understand this is for the sake of API parity. But whether the concept of a background task should to be supported on Linux/Windows is a big question that needs a discussion. I guess, background tasks may be relevant in the Swift on Android context though. 

Before we start implementing bits of this, it may be a good idea a start a discussion and decide on if, and how, we are going to support background tasks on non-Darwin platforms. This is my opinion. I'm eager to know what you, and others, think about this.

P.S: I recall [this discussion](https://forums.swift.org/t/nsurlsession-libcurl/1790/3) from the very early days :)

karthikkeyan:
> @karthikkeyan Thanks for taking this up. I understand this is for the sake of API parity. But whether the concept of a background task should to be supported on Linux/Windows is a big question that needs a discussion. I guess, background tasks may be relevant in the Swift on Android context though.
> 
> Before we start implementing bits of this, it may be a good idea a start a discussion and decide on if, and how, we are going to support background tasks on non-Darwin platforms. This is my opinion. I'm eager to know what you, and others, think about this.
> 
> P.S: I recall [this discussion](https://forums.swift.org/t/nsurlsession-libcurl/1790/3) from the very early days :)

@pushkarnk  Yes, I also had the same doubt, but I thought it wouldn't hurt to give it a try, so I had created this PR. We don't have to review or work on this PR if _now_ is not an appropriate time for it. Thanks for sharing thread link, I have also created a thread as well [here](https://forums.swift.org/t/urlsessionconfiguration-background-withidentifier-implementation/26024). I will follow both of them.

Foundation/URLSession/URLSessionConfiguration.swift
```diff
@@ -156,7 +156,17 @@ open class URLSessionConfiguration : NSObject, NSCopying {
         return ephemeralConfiguration
     }
 
-    open class func background(withIdentifier identifier: String) -> URLSessionConfiguration { NSUnimplemented() }
+    open class func background(withIdentifier identifier: String) -> URLSessionConfiguration {
```
millenomi: We do not have a daemon handling background sessions outside of Darwin. This should be marked `@availability(*, unavailable)` with a message.

millenomi:
If an API does something that has strong guarantees, we strive to implement the same guarantees. `.background(…)` creates sessions that the developer expects to complete after the process is killed, and we cannot guarantee that happens without the infrastructure Darwin provides. We should not attempt to make it available this way; instead, a user who is porting code that requires background sessions can sit down and evaluate if they are using background sessions because they rely on these semantics (and restructure their code to match the fact they aren't available) vs. using them because they benefit from those semantics but could still port this code if this was a regular URLSession.

Foundation/URLSession/URLSessionConfiguration.swift
karthikkeyan: @millenomi  I have reverted the implementation and added `@available(*, unavailable, message: "Not available on non-Darwin platforms")`

millenomi:
@swift-ci please test Linux

millenomi:
@swift-ci please test Linux

PR #2364, opened 2019-06-19
build: add an android workaround for pre r20 NDKs

compnerd:
The android NDK compiler does not support this option yet, so we disable
this at the expensive of CFString layouts not being correct.

### DIFF
--- a/CoreFoundation/CMakeLists.txt
+++ b/CoreFoundation/CMakeLists.txt
@@ -489,20 +489,29 @@ else()
                            -fno-common)
 endif()
 if(CF_DEPLOYMENT_SWIFT)
-  if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
-    target_compile_options(CoreFoundation
-                           PRIVATE
-                             /clang:-fcf-runtime-abi=swift)
-    target_compile_options(CFURLSessionInterface
-                           PRIVATE
-                             /clang:-fcf-runtime-abi=swift)
+  if(CMAKE_C_COMPILER_VERSION VERSION_LESS 8.0.7 AND CMAKE_SYSTEM_NAME STREQUAL Android)
+      # NOTE(compnerd) we assumme Swift 5.0 or newer
+      # Swift 4.2: $S10Foundation19_NSCFConstantStringCN
+      # Swift 4.1: __T010Foundation19_NSCFConstantStringCN
+      target_link_options(CoreFoundation PRIVATE
+        -Xlinker --defsym
+        -Xlinker __CFConstantStringClassReference=$$s10Foundation19_NSCFConstantStringCN)
   else()
-    target_compile_options(CoreFoundation
-                           PRIVATE
-                             -fcf-runtime-abi=swift)
-    target_compile_options(CFURLSessionInterface
-                           PRIVATE
-                             -fcf-runtime-abi=swift)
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @drodriguez @millenomi @vgorloff @gonzalolarralde @colemancda

compnerd:
Okay, this is now way safer.  It makes the assumption that Swift 5.0+ is being built, and if that is safe, I think that this can be merged.

drodriguez:
@swift-ci please test

CoreFoundation/CMakeLists.txt
```diff
@@ -489,20 +489,29 @@ else()
...
 endif()
 if(CF_DEPLOYMENT_SWIFT)
-  if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
-    target_compile_options(CoreFoundation
-                           PRIVATE
-                             /clang:-fcf-runtime-abi=swift)
-    target_compile_options(CFURLSessionInterface
-                           PRIVATE
-                             /clang:-fcf-runtime-abi=swift)
+  if(CMAKE_C_COMPILER_VERSION VERSION_LESS 8.0.0 AND CMAKE_SYSTEM_NAME STREQUAL Android)
```
zayass: Looks like condition is wrong. Android NDK 19 has clang 8.0.2 but it is not enough

compnerd: Hmm, this could be the NDK folks changing the versioning.  https://github.com/llvm-mirror/clang/blob/release_80/lib/CodeGen/CodeGenModule.cpp indicates that my changes are in the release.

drodriguez: I don’t know why the version reports as 8.0.2, but it doesn’t include your changes. They are distributing something based on r339409 according to `--version`, and your changes seemed to happen in r345222 (I miss SVN revision numbers).

Seems that NDKv20 includes "8.0.7" which seems to include your changes (seems like all of them). Maybe we can use `VERSION_LESS 8.0.7` since this only deals with Android.

compnerd: WFM

compnerd:
@swift-ci please test

compnerd:
CC: @drodriguez @spevans @millenomi - do we want to merge this until the NDK is fixed or should we just hold on to this?

ephemer:
@colemancda we have to use CFStrings on Android for various Foundation APIs (like `CFRunLoopRun`) - just checking this won't break that?

spevans:
When is the NDK likely to be fixed?

compnerd:
@spevans probably in r20 which should be released by the end of the year, but then there is the adoption problem as well.  So, I imagine it is a few years out until we can safely assume that you can rely on it being the default.

compnerd:
@ephemer - not any more, I ended up putting more work into it, and I expect that this should be compatible with Swift 4.2+ without CFString issues (though earlier versions wont work).

drodriguez:
@compnerd @spevans r20 is already released as stable (according to https://developer.android.com/ndk/downloads/revision_history). It was released last month.

compnerd:
Ooh, should test to see if it has the necessary fixes :-D  Thanks @drodriguez

compnerd:
Well, for now, I've put in a hack into the CI and it is simple enough that others can replicate it for android.  Lets avoid the workaround for now.

PR #2366, opened 2019-06-19
Implemented _getFileSystemRepresentation for Windows

gmittert:
Modified the helper _getFileSystemRepresentation functions to return a
UTF16 Windows style path.

This makes URL.path return a RFC 1808 path on Windows as opposed to the `C:\...` path before. Since `.path` would no longer be valid to pass to win32 functions, it adds Windows specific file system standardization for `standardizingPath` and `getFileSystemRepresentation`. Paths should now be represented internally as RFC 1808 paths on Windows, and only converted when being passed to win32 functions.

### DIFF
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -637,7 +637,7 @@ extension FileManager {
         var szDirectory: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(dwLength + 1))
 
         GetCurrentDirectoryW(dwLength, &szDirectory)
-        return String(decodingCString: &szDirectory, as: UTF16.self)
+        return String(decodingCString: &szDirectory, as: UTF16.self).standardizingPath
     }
 
     @discardableResult
@@ -701,8 +701,8 @@ extension FileManager {
         return true
     }
 
-    internal func _lstatFile(atPath path: String, withFileSystemRepresentation fsRep: UnsafePointer<Int8>? = nil) throws -> stat {
-        let _fsRep: UnsafePointer<Int8>
+    internal func _lstatFile(atPath path: String, withFileSystemRepresentation fsRep: UnsafePointer<NativeFSRCharType>? = nil) throws -> stat {
+        let _fsRep: UnsafePointer<NativeFSRCharType>
         if fsRep == nil {
             _fsRep = try __fileSystemRepresentation(withPath: path)
         } else {
@@ -714,15 +714,13 @@ extension FileManager {
         }
 
         var statInfo = stat()
-        let h = path.withCString(encodedAs: UTF16.self) {
-            CreateFileW(/*lpFileName=*/$0,
-                        /*dwDesiredAccess=*/DWORD(0),
-                        /*dwShareMode=*/DWORD(FILE_SHARE_READ),
-                        /*lpSecurityAttributes=*/nil,
-                        /*dwCreationDisposition=*/DWORD(OPEN_EXISTING),
-                        /*dwFlagsAndAttributes=*/DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS),
-                        /*hTemplateFile=*/nil)
-        }
+        let h = CreateFileW(_fsRep,
+                            /*dwDesiredAccess=*/DWORD(0),
+                            DWORD(FILE_SHARE_READ),
+                            /*lpSecurityAttributes=*/nil,
+                            DWORD(OPEN_EXISTING),
+                            DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS),
+                            /*hTemplateFile=*/nil)
         if h == INVALID_HANDLE_VALUE {
             throw _NSErrorWithWindowsError(GetLastError(), reading: false, paths: [path])
         }
@@ -858,7 +856,7 @@ extension FileManager {
     }
 
     internal func _updateTimes(atPath path: String,
-                               withFileSystemRepresentation fsr: UnsafePointer<Int8>,
+                               withFileSystemRepresentation fsr: UnsafePointer<NativeFSRCharType>,
                                creationTime: Date? = nil,
                                accessTime: Date? = nil,
                                modificationTime: Date? = nil) throws {
@@ -869,10 +867,7 @@ extension FileManager {
       var mtime: FILETIME =
           FILETIME(from: time_t((modificationTime ?? stat.lastModificationDate).timeIntervalSince1970))
 
-      let hFile: HANDLE = String(utf8String: fsr)!.withCString(encodedAs: UTF16.self) {
-        CreateFileW($0, DWORD(GENERIC_WRITE), DWORD(FILE_SHARE_WRITE),
-                    nil, DWORD(OPEN_EXISTING), 0, nil)
-      }
+      let hFile = CreateFileW(fsr, DWORD(GENERIC_WRITE), DWORD(FILE_SHARE_WRITE), nil, DWORD(OPEN_EXISTING), 0, nil)
       if hFile == INVALID_HANDLE_VALUE {
           throw _NSErrorWithWindowsError(GetLastError(), reading: true, paths: [path])
       }
--- a/Foundation/FileManager.swift
+++ b/Foundation/FileManager.swift
@@ -20,6 +20,14 @@ import CoreFoundation
 import MSVCRT
 #endif
 
+#if os(Windows)
+internal typealias NativeFSRCharType = WCHAR
+let NativeFSREncoding = String.Encoding.utf16LittleEndian.rawValue
+#else
+internal typealias NativeFSRCharType = CChar
+let NativeFSREncoding = String.Encoding.utf8.rawValue
+#endif
+
 open class FileManager : NSObject {
     
     /* Returns the default singleton instance.
@@ -383,7 +391,12 @@ open class FileManager : NSObject {
                     #elseif os(Linux) || os(Android) || os(Windows)
                         let modeT = number.uint32Value
                     #endif
-                    guard chmod(fsRep, mode_t(modeT)) == 0 else {
+#if os(Windows)
+                    let result = _wchmod(fsRep, mode_t(modeT))
+#else
+                    let result = chmod(fsRep, mode_t(modeT))
+#endif
+                    guard result == 0 else {
                         throw _NSErrorWithErrno(errno, reading: false, path: path)
                     }
                 
@@ -1021,15 +1034,29 @@ open class FileManager : NSObject {
      */
     open func fileSystemRepresentation(withPath path: String) -> UnsafePointer<Int8> {
         precondition(path != "", "Empty path argument")
+#if os(Windows)
+        // On Windows, the internal _fileSystemRepresentation returns
+        // UTF16 encoded data, so we need to re-encode the result as
+        // UTF8 before returning.
+        return try! _fileSystemRepresentation(withPath: path) {
+            String(decodingCString: $0, as: UTF16.self).withCString() {
+                let sz = strnlen($0, Int(MAX_PATH))
+                let buf = UnsafeMutablePointer<Int8>.allocate(capacity: sz + 1)
+                buf.initialize(from: $0, count: sz + 1)
+                return UnsafePointer(buf)
+            }
+        }
+#else
         return try! __fileSystemRepresentation(withPath: path)
+#endif
     }
 
-    internal func __fileSystemRepresentation(withPath path: String) throws -> UnsafePointer<Int8> {
+    internal func __fileSystemRepresentation(withPath path: String) throws -> UnsafePointer<NativeFSRCharType> {
         let len = CFStringGetMaximumSizeOfFileSystemRepresentation(path._cfObject)
         if len != kCFNotFound {
-            let buf = UnsafeMutablePointer<Int8>.allocate(capacity: len)
+            let buf = UnsafeMutablePointer<NativeFSRCharType>.allocate(capacity: len)
             buf.initialize(repeating: 0, count: len)
-            if path._nsObject.getFileSystemRepresentation(buf, maxLength: len) {
+            if path._nsObject._getFileSystemRepresentation(buf, maxLength: len) {
                 return UnsafePointer(buf)
             }
             buf.deinitialize(count: len)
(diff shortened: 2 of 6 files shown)


gmittert:
See also: https://github.com/apple/swift-corelibs-foundation/pull/2367

cc @compnerd @lxbndr @millenomi @spevans

Foundation/NSPathUtilities.swift
```diff
@@ -550,13 +564,61 @@ extension NSString {
...
             return false
         }
-        
         return CFStringGetFileSystemRepresentation(self._cfObject, cname, max)
+#endif
     }
 
+#if os(Windows)
+    public func _getFileSystemRepresentation(_ cname: UnsafeMutablePointer<foundation_char_t>, maxLength max: Int) -> Bool {
+        guard self.length > 0 else {
```
gmittert: This is probably what I'm most concerned about, I'm not entirely sure my logic is 100% sound. `CreateFileW` is pretty good about being able to figure out a path, so in general, we just have to get the path to be "good enough" to be parsed by it, and it will properly do the logic for the various types of Windows paths. However, there a few exceptions I've found:
- /c:/foo/bar doesn't parse, the first '/' needs to be removed
  - except `/foo/bar` should parse, since it's a drive relative absolute path
- Trailing slashes must be removed
  - except in the case of `[driveletter]:\` where `C:\` is the `C` root directory, but `C:` is the current directory.

Foundation/FileManager+Win32.swift
```diff
@@ -581,8 +581,8 @@ extension FileManager {
         NSUnimplemented()
     }
 
-    internal func _lstatFile(atPath path: String, withFileSystemRepresentation fsRep: UnsafePointer<Int8>? = nil) throws -> stat {
-        let _fsRep: UnsafePointer<Int8>
+    internal func _lstatFile(atPath path: String, withFileSystemRepresentation fsRep: UnsafePointer<foundation_char_t>? = nil) throws -> stat {
```
compnerd: I don't really like the `foundation_char_t`.  Can we not do something like:

```Swift
#if os(Windows)
internal typealias NativeFileSystemCharacterType = WCHAR
#else
internal typealias NativeFileSystemCharacterType = CChar
#endif
```

I'm happy with alternate names such as `NativeFSRCharType` or something along those lines.

Foundation/FileManager+Win32.swift
```diff
@@ -646,34 +644,22 @@ extension FileManager {
...
     internal func _appendSymlinkDestination(_ dest: String, toPath: String) -> String {
-        var isAbsolutePath: Bool = false
-        dest.withCString(encodedAs: UTF16.self) {
-            isAbsolutePath = !PathIsRelativeW($0)
-        }
-
-        if isAbsolutePath {
-            return dest
-        }
+        if dest.isAbsolutePath { return dest }
```
compnerd: This can be split out with some of the other clean ups into a separate change.

Foundation/FileManager.swift
```diff
@@ -20,6 +20,12 @@ import CoreFoundation
 import MSVCRT
 #endif
 
+#if os(Windows)
+public typealias foundation_char_t = WCHAR
+#else
+public typealias foundation_char_t = Int8
```
compnerd: These should not be `public` but rather `internal`.

Foundation/FileManager.swift
```diff
@@ -383,7 +389,12 @@ open class FileManager : NSObject {
                     #elseif os(Linux) || os(Android) || os(Windows)
                         let modeT = number.uint32Value
                     #endif
-                    guard chmod(fsRep, mode_t(modeT)) == 0 else {
+#if os(Windows)
+                    let result = _wchmod(fsRep, mode_t(modeT))
```
compnerd: *sigh* this is a good fix.

Foundation/FileManager.swift
```diff
@@ -1027,9 +1049,19 @@ open class FileManager : NSObject {
     /* stringWithFileSystemRepresentation:length: returns an NSString created from an array of bytes that are in the filesystem representation.
      */
     open func string(withFileSystemRepresentation str: UnsafePointer<Int8>, length len: Int) -> String {
-        return NSString(bytes: str, length: len, encoding: String.Encoding.utf8.rawValue)!._swiftObject
+        let encoding = String.Encoding.utf8.rawValue
+        return NSString(bytes: str, length: len, encoding: encoding)!._swiftObject
     }
-    
+
+    open func _string(withFileSystemRepresentation str: UnsafePointer<foundation_char_t>, length len: Int) -> String {
```
compnerd: I don't think that this should be `open`.  We want this to be `internal`.

Foundation/FileManager.swift
```diff
@@ -1027,9 +1049,19 @@ open class FileManager : NSObject {
...
+#if os(Windows)
+      let encoding = String.Encoding.utf16LittleEndian.rawValue
+#else
+      let encoding = String.Encoding.utf8.rawValue
+#endif
```
compnerd: Could you pull this out and put it near the character type alias?  We should just have a `NativeFileSystemEncoding` constant.

Foundation/NSPathUtilities.swift
```diff
@@ -344,9 +344,23 @@ extension NSString {
...
     
     public var standardizingPath: String {
+#if os(Windows)
+        // Convert to a posix style '/' separated path
+        var converted = self as String
+        if converted.isAbsolutePath {
+            if converted.starts(with: "\\\\?\\") {
+              converted.removeFirst(3)
+            } else if !(converted.starts(with: "/") || converted.starts(with: "\\")) {
+              converted = "/" + converted
```
compnerd: Am I mistaken or does this break with UNC paths?  `\\host\C$` is not going to get mapped into a valid path I think.

Foundation/NSPathUtilities.swift
```diff
@@ -550,13 +564,61 @@ extension NSString {
     }
     
     public func getFileSystemRepresentation(_ cname: UnsafeMutablePointer<Int8>, maxLength max: Int) -> Bool {
+#if os(Windows)
+        let fsr = UnsafeMutablePointer<WCHAR>.allocate(capacity: max)
+        defer{ fsr.deallocate() }
```
compnerd: Nit: space after `defer`.

Foundation/NSURL.swift
```diff
@@ -29,10 +29,14 @@ internal let kCFURLPlatformPathStyle = kCFURLPOSIXPathStyle
 #endif
 
 private func _standardizedPath(_ path: String) -> String {
+#if os(Windows)
```
compnerd: Hmm, could you add a comment why the absolute path check is not needed on windows.

Foundation/NSPathUtilities.swift
```diff
@@ -550,13 +564,61 @@ extension NSString {
...
 
+#if os(Windows)
+    public func _getFileSystemRepresentation(_ cname: UnsafeMutablePointer<foundation_char_t>, maxLength max: Int) -> Bool {
+        guard self.length > 0 else {
+            return false
+        }
+
+        var fsr = self._swiftObject
+        let idx = fsr.startIndex
+        // If we have an rfc 8089 style path /[drive-letter]:/....., drop the
```
compnerd: Nit: capitalize `RFC`, it is an initialism for "Request For Comments".

Foundation/NSPathUtilities.swift
```diff
@@ -550,13 +564,61 @@ extension NSString {
...
         guard self.length > 0 else {
             return false
         }
-        
         return CFStringGetFileSystemRepresentation(self._cfObject, cname, max)
+#endif
     }
 
+#if os(Windows)
+    public func _getFileSystemRepresentation(_ cname: UnsafeMutablePointer<foundation_char_t>, maxLength max: Int) -> Bool {
```
spevans: This can't be `public` as it would be adding new API

Foundation/FileManager.swift
```diff
@@ -992,15 +1003,26 @@ open class FileManager : NSObject {
...
 
-    internal func __fileSystemRepresentation(withPath path: String) throws -> UnsafePointer<Int8> {
+    internal func __fileSystemRepresentation(withPath path: String) throws -> UnsafePointer<foundation_char_t> {
         let len = CFStringGetMaximumSizeOfFileSystemRepresentation(path._cfObject)
         if len != kCFNotFound {
-            let buf = UnsafeMutablePointer<Int8>.allocate(capacity: len)
+            let buf = UnsafeMutablePointer<foundation_char_t>.allocate(capacity: len)
             buf.initialize(repeating: 0, count: len)
-            if path._nsObject.getFileSystemRepresentation(buf, maxLength: len) {
+            if path._nsObject._getFileSystemRepresentation(buf, maxLength: len) {
```
spevans: It looks like there is only a version of `_getFileSystemRepresentation` for Windows but its also used by non-windows platforms or am I missing something?

gmittert:
- Now uses NativeFSRCharType
- UNC paths should now properly be converted
- Made sure things are internal instead of public/open

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

Foundation/FileManager.swift
```diff
@@ -20,6 +20,14 @@ import CoreFoundation
 import MSVCRT
 #endif
 
+#if os(Windows)
+internal typealias NativeFSRCharType = WCHAR
```
xwu: What is "FSR"? Can you spell it out?

compnerd: `FileSystemRepresentation`

Foundation/NSPathUtilities.swift
```diff
@@ -550,13 +567,61 @@ extension NSString {
...
+            }
+            cname.assign(from: $0, count: chars + 1)
+            return true
+        }
+#else
         guard self.length > 0 else {
             return false
         }
-        
         return CFStringGetFileSystemRepresentation(self._cfObject, cname, max)
```
spevans: Would it not be better to just fix `CFStringGetFileSystemRepresentation` to work for Windows as well, then it should work for other callers inside `CoreFoundation`?

Foundation/FileManager.swift
```diff
@@ -992,15 +1005,31 @@ open class FileManager : NSObject {
...
         if len != kCFNotFound {
-            let buf = UnsafeMutablePointer<Int8>.allocate(capacity: len)
+            let buf = UnsafeMutablePointer<NativeFSRCharType>.allocate(capacity: len)
             buf.initialize(repeating: 0, count: len)
-            if path._nsObject.getFileSystemRepresentation(buf, maxLength: len) {
+#if os(Windows)
+            let result = path._nsObject._getFileSystemRepresentation(buf, maxLength: len)
+#else
+            let result = path._nsObject.getFileSystemRepresentation(buf, maxLength: len)
+#endif
```
spevans: This is not a good way to this as every call site would need to be updated. I would add a wrapper `_getFileSystemRepresentation` for the non-Windows version and just call `_getFileSystemRepresentation` for all  uses.

Foundation/NSPathUtilities.swift
gmittert: Do we think it's appropriate to have the Windows implementation do path normalization? The POSIX implementation doesn't, it converts to a cstring in the appropriate system encoding (UTF8/UTF16), but doesn't do any path normalization itself.

In this case, by path normalization, I mean dropping leading and trailing slashes when appropriate (as below), then optionally passing the path to GetFullPathNameW to fully normalize it.

cc @millenomi

Foundation/FileManager+Win32.swift
```diff
@@ -594,15 +594,13 @@ extension FileManager {
...
-                        /*dwDesiredAccess=*/DWORD(0),
-                        /*dwShareMode=*/DWORD(FILE_SHARE_READ),
-                        /*lpSecurityAttributes=*/nil,
-                        /*dwCreationDisposition=*/DWORD(OPEN_EXISTING),
-                        /*dwFlagsAndAttributes=*/DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS),
-                        /*hTemplateFile=*/nil)
-        }
+        let h = CreateFileW(/*lpFileName=*/_fsRep,
+                            /*dwDesiredAccess=*/DWORD(0),
+                            /*dwShareMode=*/DWORD(FILE_SHARE_READ),
```
compnerd: Dont think that the parameter name comment is needed here.

Foundation/FileManager+Win32.swift
```diff
@@ -594,15 +594,13 @@ extension FileManager {
...
+                            /*lpSecurityAttributes=*/nil,
+                            /*dwCreationDisposition=*/DWORD(OPEN_EXISTING),
+                            /*dwFlagsAndAttributes=*/DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS),
```
compnerd: Nor these two

Foundation/FileManager.swift
```diff
@@ -992,15 +1005,31 @@ open class FileManager : NSObject {
      */
     open func fileSystemRepresentation(withPath path: String) -> UnsafePointer<Int8> {
         precondition(path != "", "Empty path argument")
+#if os(Windows)
+        return try! _fileSystemRepresentation(withPath: path) {
+            String(decodingCString: $0, as: UTF16.self).withCString() {
+                let sz = strnlen($0, Int(MAX_PATH))
```
compnerd: Huh?  This seems wrong.  This will almost always return `1`.  You want `wcslen`

millenomi:
I don't want to bifurcate the API. Also: a buffer of WCHARS can be expressed as a buffer of UInt8s with a certain length requirement, so I don't see why bifurcating types is necessary at all.

gmittert:
> I don't want to bifurcate the API. Also: a buffer of WCHARS can be expressed as a buffer of UInt8s with a certain length requirement, so I don't see why bifurcating types is necessary at all.

I'm pretty uncomfortable with storing UTF16 encoded strings in a UInt8 buffer. I think it's asking to get a lot of invisible unicode errors on Windows. At the minimum, calling `_fileSystemRepresentation` then having to rebind the `UnsafePointer<UInt8>` you're passed to a `UnsafePointer<WCHAR>` is confusing would probably involve calling strlen on the buffer each time.

This also doesn't change the public api, only internal. `getFileSystemRepresentation(:maxLength)` still returns a UTF8 encoded cstring, even on Windows.

gmittert:
@millenomi What would you want to see api wise for Windows File System Representations? Shall I just implement the current UInt8 one to return UTF16 data on Windows?

millenomi:
Correct. I understand the wish for strong typing, but this isn't source-compatible across platform, and as such it is in direct opposition to a project goal.

gmittert:
Alright! Finally cleaned this up a bit. It shouldn't touch any public apis, instead modifying the signatures of the internal `_fileSystemRepresentation` functions. The public `getFileSystemRepresentation(:maxLength)` still returns a UTF8 encoded cstring, even on Windows.

gmittert:
@swift-ci please test linux

millenomi:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

Foundation/NSPathUtilities.swift
```diff
@@ -550,11 +567,61 @@ extension NSString {
...
+        // the path separators anyways.
+        fsr = fsr.replacingOccurrences(of: "/", with: "\\")
+
+        // Drop trailing slashes unless it follows a drive letter. On
+        // Windows the path `C:\` indicates the root directory of the
+        // `C:` drive. The path `C:` indicates the current working
+        // directory on the `C:` drive.
+        while fsr.count > 1
+           && (fsr[fsr.index(before: fsr.endIndex)] == "\\")
+           && !(fsr.count == 3 && fsr[fsr.index(fsr.endIndex, offsetBy: -2)] == ":") {
```
compnerd: I wish there was a more elegant way to write this.

compnerd:
LGTM if @millenomi is okay with the restructuring for the getFSR

gmittert:
To confirm, this PR will mean that that `URL(fileURLWithPath: "C:/foo\bar").path` returns `/C:/foo/bar` on Windows which is not a valid Windows path, and if you want "C:\foo\bar", you'll have to call FileManager/NSString .getFilesystemRepresentation. I can easily change this to be have a different behavior if we want something that isn't truly a URL, perhaps `C:/foo/bar` which is good enough to pass to most Windows APIs, or `C:\foo\bar`.

gmittert:
Accidently closed this while keeping it up to date.

gmittert:
@swift-ci please test linux

PR #2369, opened 2019-06-20
Foundation: repair build for android ARMv7

compnerd:
```
D:/a/1/s/swift-corelibs-foundation/Foundation/FileManager+POSIX.swift:987:41: error: cannot convert value of type 'UInt64' to expected argument type 'dev_t' (aka 'UInt32')
                return _dev_major(file1.st_rdev) == _dev_major(file2.st_rdev)
                                  ~~~~~~^~~~~~~
                                  dev_t(       )
```


compnerd:
CC: @gmittert @spevans

compnerd:
@swift-ci please test

spevans:
Looks ok, are there anymore of these `stat` fields that need fixing?

compnerd:
I need to set up better automation for the builds on Windows, right now, the easiest way for me to build is on Azure.  I think that there is the reference to it in NSURL and FileManager, but both of those should be fine.

PR #2370, opened 2019-06-21
build: collapse two identical cases (NFC)

compnerd:


compnerd:
@swift-ci please test

PR #2372, opened 2019-06-21
Fix TestProcessInfo on Windows

gmittert:
- The Executable name should have .exe
- Windows doesn't have setenv, so make putenv work more like it and accept keys with `=` in them.

### DIFF
--- a/TestFoundation/TestProcessInfo.swift
+++ b/TestFoundation/TestProcessInfo.swift
@@ -38,6 +38,8 @@ class TestProcessInfo : XCTestCase {
         // just test that the initial name is not empty or something?
 #if DARWIN_COMPATIBILITY_TESTS
         let targetName = "xctest"
+#elseif os(Windows)
+        let targetName = "TestFoundation.exe"
 #else
         let targetName = "TestFoundation"
 #endif
@@ -71,7 +73,11 @@ class TestProcessInfo : XCTestCase {
 #if os(Windows)
         func setenv(_ key: String, _ value: String, _ overwrite: Int) -> Int32 {
           assert(overwrite == 1)
-          return putenv("\(key)=\(value)")
+          guard !key.contains("=") else {
+              errno = EINVAL
+              return -1
+          }
+          return _putenv("\(key)=\(value)")
         }
 #endif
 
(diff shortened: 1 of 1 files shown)


gmittert:
@swift-ci please test

TestFoundation/TestProcessInfo.swift
```diff
@@ -104,7 +107,12 @@ class TestProcessInfo : XCTestCase {
         XCTAssertNil(env["bad3"])
         XCTAssertNil(env["bad3="])
 
+#if os(Windows)
+        // On Windows, adding an empty environment variable removes it from the environment
+        XCTAssertNil(env["var1"])
```
compnerd: Could we not emulate this in the subscript operator?  Setting it to nothing would set it to the empty string.

TestFoundation/TestProcessInfo.swift
```diff
@@ -71,7 +73,8 @@ class TestProcessInfo : XCTestCase {
 #if os(Windows)
         func setenv(_ key: String, _ value: String, _ overwrite: Int) -> Int32 {
           assert(overwrite == 1)
-          return putenv("\(key)=\(value)")
+          guard !key.contains("=") else { return -1 }
```
compnerd: Can you have `=` in the key on Unix?

gmittert: No, it will return EINVAL.

TestFoundation/TestProcessInfo.swift
gmittert: As far as I'm aware, you can't have an environment variable set to `""` or be unset on Windows. Doing so simply removes it from the environment. Since it's not in the environment, I think it makes sense to have it return nil there.

gmittert:
rebased

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

TestFoundation/TestProcessInfo.swift
compnerd: Should `setenv` in that case behave similar to Unix and return `EINVAL`?

TestFoundation/TestProcessInfo.swift
compnerd: Bleh, yeah, `putenv` behaves like set.  Okay, that seems reasonable.

TestFoundation/TestProcessInfo.swift
gmittert: I've made it match Unix and return return -1 and set `errno` to `EINVAL`.

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test linux

PR #2373, opened 2019-06-21
HTTPCookie: parse domain according to RFC

drodriguez:
When the domain field is specified, the cookie is intended for the
domain and subdomains, so the value of domain has to be prefixed by a
dot.

Added tests to check for this behaviour and modified the incorrect
tests.

NOTE: I checked this behaviour in macOS 13.14.5, and it matches Foundation there.

PR #2374, opened 2019-06-22
Improve cookie parsing algorithm.

drodriguez:
While the previous parser dealt with commas, some cases were not
correctly handled, like when a comma was in the last attribute before a
second cookie.

Also trim the names of the cookies to allow spaces before the equal.

### DIFF
--- a/Foundation/HTTPCookie.swift
+++ b/Foundation/HTTPCookie.swift
@@ -69,6 +69,28 @@ extension HTTPCookiePropertyKey {
     internal static let created = HTTPCookiePropertyKey(rawValue: "Created")
 }
 
+internal extension HTTPCookiePropertyKey {
+    internal static let httpOnly = HTTPCookiePropertyKey(rawValue: "HttpOnly")
+
+    static private let _setCookieAttributes: [String: HTTPCookiePropertyKey] = {
+        // Only some attributes are valid in the Set-Cookie header.
+        let validProperties: [HTTPCookiePropertyKey] = [
+            .expires, .maximumAge, .domain, .path, .secure, .comment,
+            .commentURL, .discard, .port, .version, .httpOnly
+        ]
+        let canonicalNames = validProperties.map { $0.rawValue.lowercased() }
+        return Dictionary(uniqueKeysWithValues: zip(canonicalNames, validProperties))
+    }()
+
+    init?(attributeName: String) {
+        let canonical = attributeName.lowercased()
+        switch HTTPCookiePropertyKey._setCookieAttributes[canonical] {
+        case let property?: self = property
+        case nil: return nil
+        }
+    }
+}
+
 /// `HTTPCookie` represents an http cookie.
 ///
 /// An `HTTPCookie` instance represents a single http cookie. It is
@@ -122,11 +144,6 @@ open class HTTPCookie : NSObject {
 
     static let _allFormatters: [DateFormatter]
         = [_formatter1, _formatter2, _formatter3]
-
-    static let _attributes: [HTTPCookiePropertyKey]
-        = [.name, .value, .originURL, .version, .domain,
-           .path, .secure, .expires, .comment, .commentURL,
-           .discard, .maximumAge, .port]
 
     /// Initialize a HTTPCookie object with a dictionary of parameters
     ///
@@ -348,8 +365,12 @@ open class HTTPCookie : NSObject {
         } else {
             _commentURL = nil
         }
-        _HTTPOnly = false
 
+        if let httpOnlyString = properties[.httpOnly] as? String {
+            _HTTPOnly = httpOnlyString == "TRUE"
+        } else {
+            _HTTPOnly = false
+        }
 
         _properties = [
             .created : Date().timeIntervalSinceReferenceDate, // Cocoa Compatibility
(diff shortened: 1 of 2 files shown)


pushkarnk:
@swift-ci please test

drodriguez:
I have seen those failures in macOS all the time recently. I don’t think they are related to the PR (neither this, neither #2373).

Foundation/HTTPCookie.swift
```diff
@@ -444,14 +488,16 @@ open class HTTPCookie : NSObject {
...
 
     //Bake a cookie
-    private class func createHttpCookie(url: URL, pairs: ArraySlice<String>) -> HTTPCookie? {
-        var properties: [HTTPCookiePropertyKey : Any] = [:]
+    private class func createHttpCookie(url: URL, cookie: Substring) -> HTTPCookie? {
+        let pairs = cookie.components(separatedBy: ";") // split the name/value and attribute/value pairs
+                          .map({$0.trim()})             // trim whitespaces
+                          .filter({$0.caseInsensitiveCompare("HTTPOnly") != .orderedSame})  // we don't use HTTPOnly, do we?
+                          .flatMap(HTTPCookie.createNameValuePair(pair:)) // create Name and Value properties
+
```
pushkarnk: Nit: this can be
```
let pairs = cookie.components(separatedBy: ";") // split the name/value and attribute/value pairs
	                     .map { $0.trim() }            // trim whitespaces
	                     .filter { $0.caseInsensitiveCompare("HTTPOnly") != .orderedSame }   // we don't use HTTPOnly ...
	                     .flatMap(HTTPCookie.createNameValuePair) // create Name and Value properties
```

drodriguez: I will change it. I tried to keep the old code, just in case the diff was simpler, but since it is not, there's no reason to keep the parenthesis.

drodriguez:

drodriguez:

Foundation/HTTPCookie.swift
```diff
@@ -654,12 +692,27 @@ fileprivate extension String {
...
+        // CTL, 0-31 and DEL (127)
+        if asciiValue <= 31 || asciiValue >= 127 {
+            return false
+        }
+
+        return self != "(" && self != ")" && self != "<" && self != ">" &&
+               self != "@" && self != "," && self != ";" && self != ":" &&
+               self != "\\" && self != "\"" && self != "/" && self != "[" &&
+               self != "]" && self != "?" && self != "=" && self != "{" &&
+               self != "}" && self != " " && self != "\t"
```
spevans: This might be better expressed as:
```
var notTokens = "()<>@,;:\\\"/[]?={} \t"
return !notTokens.contains(self)
```

drodriguez:

drodriguez:

Foundation/HTTPCookie.swift
```diff
@@ -444,14 +488,16 @@ open class HTTPCookie : NSObject {
...
+                          .flatMap(HTTPCookie.createNameValuePair) // create Name and Value properties
+
+        var properties: [HTTPCookiePropertyKey : String] = [:]
         for pair in pairs {
             let name = pair.components(separatedBy: "=")[0]
-            var value = pair.components(separatedBy: "\(name)=")[1]  //a value can have an "="
-            if canonicalize(name) == .expires {
-                value = value.unmaskCommas()    //re-insert the comma
-            }
+            let value = pair.components(separatedBy: "\(name)=")[1]  //a value can have an "="
```
spevans: Is the string guaranteed to have an `=` in it? If not this will cause an array out of bounds error. It might be safer to write this as:
```
if let idx1 = pair.firstIndex(of: "=")  {
        let idx2 = pair.index(after: idx1)
        let name = String(pair[..<idx1])
        let value = String(pair[idx2...])
        properties[canonicalize(name)] = value
}
```

With checks for `name.isEmpty` as appropriate. This should also be faster than calling `.components(separatedBy:)` twice.

Foundation/HTTPCookie.swift
```diff
@@ -497,7 +535,7 @@ open class HTTPCookie : NSObject {
         let name = pair.components(separatedBy: "=")[0]
         let value = pair.components(separatedBy: "\(name)=")[1]
```
spevans: See comment above

TestFoundation/TestHTTPCookie.swift
```diff
@@ -171,6 +173,35 @@ class TestHTTPCookie: XCTestCase {
...
+        do {
+            let cookies = HTTPCookie.cookies(withResponseHeaderFields: [
+                "Set-Cookie": " AAA=1; path=/; domain=.example.com; expires=Sun, 16-Aug-2025 22:39:54 GMT, BBB=2; path=/; domain=.example.com; HttpOnly; expires=Sat, 15-Feb-2014 22:39:54 GMT"
+            ], for: try URL(string: "http://www.example.com/").unwrapped())
+            XCTAssertEqual(cookies.count, 2)
+            XCTAssertEqual(cookies[0].name, "AAA")
+            XCTAssertEqual(cookies[1].name, "BBB")
+        }
+    }
+
```
spevans: Probably worth adding in some broken cookies and ones where the value part has `=` in it to test the parsing. I would also add in `XCTAssertEqual(cookies[0].value...` checks to validate the values.

Foundation/HTTPCookie.swift
drodriguez: I think since these is strings that are sent from the server, one cannot have any guarantee.

My intention with these changes was fixing the cookie split algorithm, but you are right that the cookie parsing algorith also needs some love (for example you point out that the name/value split seems to be happening in two points). I think much of the previous algorithm relied on crashing on badly formatted input (for example if there's no `=`, there's no `[1]` to get the value).

I was planning to improve the cookie parsing in a later commit, making it closer to the grammar in the RFC, but I was planning to do it in a separate PR. Would that be good?

spevans: If you want to do it in a follow-up PR and merge the two places thats fine.

drodriguez:
I will try to improve the cookie parsing with the recommendations and add more tests with broken cookies in a later PR.

drodriguez:
I finally reused this PR, but I haven't add new tests right now. The current one pass, which is quite good already. I will write better ones tomorrow.

drodriguez:

drodriguez:
Added the tests I promised for yesterday. Fixed some pieces to match better macOS Foundation.

drodriguez:

PR #2377, opened 2019-06-24
Fix two Decimal members with respect to NaN

xwu:
_This PR corresponds to apple/swift#25698._

It fixes `Decimal.magnitude` so that `Decimal.nan.magnitude` is no longer incorrectly `0`.

The corresponding PR in the apple/swift repository also incorporates changes from #1759; here, minor edits (NFC), including one suggested on review, are back-propagated so as to help the two versions of `Foundation.Decimal` remain in sync.

Resolves [SR-10938](https://bugs.swift.org/browse/SR-10938).

### DIFF
--- a/Foundation/Decimal.swift
+++ b/Foundation/Decimal.swift
@@ -202,16 +202,12 @@ extension Decimal {
 extension Decimal : Hashable, Comparable {
     internal var doubleValue: Double {
         if _length == 0 {
-            if _isNegative == 1 {
-                return Double.nan
-            } else {
-                return 0
-            }
+            return _isNegative == 1 ? Double.nan : 0
         }
 
         var d = 0.0
-        for idx in stride(from: min(_length, 8), to: 0, by: -1) {
-            d = d * 65536 + Double(self[idx - 1])
+        for idx in (0..<min(_length, 8)).reversed() {
+            d = d * 65536 + Double(self[idx])
         }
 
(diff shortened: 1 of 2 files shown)


xwu:
@swift-ci Please test

xwu:
@spevans There's at least one other fix that isn't synchronized between here and the overlay. One thing I'd like to do (or rather, have already done but would like to contribute in a PR) is to reorganize `Decimal.swift` and bring the two implementations back into alignment where it makes sense to do so. Even the organization of two files has fallen out of sync, making it increasingly difficult to ensure parity.

spevans:
This looks fine but might be worth adding a few tests for `.magnitude` of various numbers, eg the `Decimal` constants `Decimal.leastFiniteMagnitude`, `Decimal.greatestFiniteMagnitude` etc and maybe `Decimal(1)`, `Decimal(-1)`

xwu:
@swift-ci Please test Linux platform

xwu:
@swift-ci Please test Linux platform

spevans:
Yes, definitely a good idea to resync the two versions. I think the scl-f version may be slightly ahead in bug fixes but there should be enough tests in it to cover any slight regressions. Probably worth doing a couple of PRs to resync them.

xwu:
@swift-ci Please test Linux

xwu:
@swift-ci Please test Linux

PR #2378, opened 2019-06-25
Synchronize Decimal with Darwin overlay

xwu:
Reorganize Decimal.swift, bringing the two implementations back into alignment where it makes sense to do so. Even the organization of the two files has fallen out of sync, making it increasingly difficult to ensure parity.

A [corresponding PR](https://github.com/apple/swift/pull/25745) has been drafted for the Darwin overlay.

The major functional change here is incorporation of the function `pow(_: Decimal, _: Int)`, which is present in the Darwin overlay but not found in swift-corelibs-foundation.

### DIFF
--- a/Foundation/Decimal.swift
+++ b/Foundation/Decimal.swift
@@ -196,6 +196,37 @@ extension Decimal {
 }
 
 extension Decimal : Hashable, Comparable {
+    // (Used by VariableLengthNumber and doubleValue.)
+    fileprivate subscript(index: UInt32) -> UInt16 {
+        get {
+            switch index {
+            case 0: return _mantissa.0
+            case 1: return _mantissa.1
+            case 2: return _mantissa.2
+            case 3: return _mantissa.3
+            case 4: return _mantissa.4
+            case 5: return _mantissa.5
+            case 6: return _mantissa.6
+            case 7: return _mantissa.7
+            default: fatalError("Invalid index \(index) for _mantissa")
+            }
+        }
+        set {
+            switch index {
+            case 0: _mantissa.0 = newValue
+            case 1: _mantissa.1 = newValue
+            case 2: _mantissa.2 = newValue
+            case 3: _mantissa.3 = newValue
+            case 4: _mantissa.4 = newValue
+            case 5: _mantissa.5 = newValue
+            case 6: _mantissa.6 = newValue
+            case 7: _mantissa.7 = newValue
+            default: fatalError("Invalid index \(index) for _mantissa")
+            }
+        }
+    }
+
+    // (Used by NSDecimalNumber and hash(into:).)
     internal var doubleValue: Double {
         if _length == 0 {
             return _isNegative == 1 ? Double.nan : 0
@@ -218,13 +249,12 @@ extension Decimal : Hashable, Comparable {
         return _isNegative != 0 ? -d : d
     }
 
-    // Return the low 64bits of the integer part
+    // The low 64 bits of the integer part. (Used by uint64Value and int64Value.)
     private var _unsignedInt64Value: UInt64 {
         if _exponent < -20 || _exponent > 20 {
             return 0
         }
-
-        if _length == 0 || isZero || magnitude < Decimal(0) {
+        if _length == 0 || isZero || magnitude < (0 as Decimal) {
             return 0
         }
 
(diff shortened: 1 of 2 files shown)


xwu:
@swift-ci Please test Linux

Foundation/Decimal.swift
```diff
@@ -661,17 +661,19 @@ extension Decimal {
...
         var theDecimal = Decimal()
+        scan.locale = locale
         if !scan.scanDecimal(&theDecimal) {
             return nil
         }
         self = theDecimal
     }
 
+    // Note: In the Darwin overlay, `description` is implemented in terms of
+    // `NSDecimalString(_:_:)`; here, it's the other way around.
```
spevans: We should do the same here and implement the conversion in `NSDecimalString(_:_:)` since it takes a `Locale`. Then `NSDecimalNumber.description(withLocale:)`, `NSDecimalNumber.description` and `Decimal.description` can all call `NSDecimalString(_:_:)` to match the Darwin overlay and have localisation work correctly as well. Do you want to split out a PR to do this change? You can ignore the localisation part for now as it can be done in a followup to keep it as NFC.

Foundation/Decimal.swift
```diff
@@ -661,17 +661,19 @@ extension Decimal {
     public mutating func formTruncatingRemainder(dividingBy other: Decimal) { fatalError("Decimal does not yet fully adopt FloatingPoint") }
 }
 
-extension Decimal: CustomStringConvertible {
+extension Decimal : CustomStringConvertible {
```
spevans: Are we consistent in the spacing in the `extension` declarations? It seemed to be a mix, not sure if you just want to move everything to be one or other.

Foundation/Decimal.swift
```diff
@@ -340,56 +340,57 @@ extension Decimal : SignedNumeric {
         fatalError()
     }
 
-    public static func +=(_ lhs: inout Decimal, _ rhs: Decimal) {
-        var leftOp = lhs
-        var rightOp = rhs
-        _ = NSDecimalAdd(&lhs, &leftOp, &rightOp, .plain)
+    public static func +=(lhs: inout Decimal, rhs: Decimal) {
+        var rhs = rhs
+        _ = withUnsafeMutablePointer(to: &lhs) {
```
spevans: Is there a benefit to using `withUnsafeMutablePointer(to:)` here? Seems better to not requite `unsafe` functions if we can avoid using them.

Foundation/Decimal.swift
```diff
@@ -714,107 +677,131 @@ extension Decimal {
...
 
-    public var isSignMinus: Bool {
-        return _isNegative != 0
-    }
+    public var isCanonical: Bool { true }
 
-    public var isNormal: Bool {
-        return !isZero && !isInfinite && !isNaN
-    }
+    /// `true` iff `self` is negative.
```
spevans: Is `iff` a typo here (and below)?

Foundation/Decimal.swift
xwu: Yes, I'm moving the formatting to be the same in the portion of the file that's common between sclf and the overlay.

I'm going to do (have already done) a gardening pass for formatting on the remainder of this file that is unique to sclf, but I left it out of this PR to keep things focused here on synchronizing the two files.

Foundation/Decimal.swift
xwu: `NSDecimalAdd`, etc., are explicitly noted in both the Obj-C and Swift versions to permit the result pointer to be the same as either the left or right operand. Calling the method directly was possible (and indeed what we did in both versions of `Decimal`) until enforcement of the Law of Exclusivity.

The `withUnsafeMutablePointer(to:)` version is safe because, well, that's the documented behavior of the API we call. By using it, we tell the compiler not to deny us based the Law of Exclusivity (a change in this case made by me and reviewed by @phausler, I believe, over in the Darwin overlay). The alternative is an additional copy which may or may not be elided on compilation, which is less optimal for a basic operation.

Foundation/Decimal.swift
xwu: No; it's the shorthand for "if and only if" synchronized verbatim from the documentation in the Darwin overlay.

Foundation/Decimal.swift
xwu: Yes, that would be an optimal arrangement.

I have some additional gardening planned for the portions specific to the sclf version of this file. After that, I think the way to go would be some PRs to implement `NSDecimalString` and other `NSDecimal*` functions without using overlay methods, then to switch over the overlay implementations to call the `NSDecimal*` methods. There are several other divergences (`compare(to:)` versus `NSDecimalCompare`) that could benefit from this approach.

xwu:
@swift-ci Please test Linux

xwu:
@swift-ci Please test Linux

xwu:
@swift-ci Please test Linux

Foundation/Decimal.swift
spevans: FYI Ive done an implementation of `NSDecimalString` that also takes the `locale` into account so once this PR is merged I will put it up.

xwu: That's great news!

xwu:
Unrelated SwiftPM failure; let's see if second time's the charm.

xwu:
@swift-ci Please test Linux

PR #2381, opened 2019-06-25
Interpret Process Exit Codes on Windows

gmittert:
Attempt to interpret NTSTATUS and signal errors on Windows so that we
can indicate when a process crashed.


compnerd:
@swift-ci please test

PR #2385, opened 2019-06-27
Also pass Path to the Child Process on Windows

gmittert:
Since child processes use `Path` for their dll search path, `Path`
should always be passed to children if not already specified.

### DIFF
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -445,6 +445,12 @@ open class Process: NSObject {
           environment["PWD"] = currentDirectoryURL.path
         }
 
+        // On Windows, the PATH is required in order to locate dlls needed by
+        // the process so we should also pass that to the child
+        if environment["Path"] == nil, let path = ProcessInfo.processInfo.environment["Path"] {
+            environment["Path"] = path
+        }
+
         // NOTE(compnerd) the environment string must be terminated by a double
         // null-terminator.  Otherwise, CreateProcess will fail with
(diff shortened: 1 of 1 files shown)


gmittert:
@swift-ci please test linux

Foundation/Process.swift
```diff
@@ -445,6 +445,12 @@ open class Process: NSObject {
           environment["PWD"] = currentDirectoryURL.path
         }
 
+        // On Windows, the PATH is required in order to locate dlls needed by
+        // the process so we should also pass that to the child
+        if environment["Path"] == nil, let path = ProcessInfo.processInfo.environment["Path"] {
```
compnerd: Is the variable not `PATH` rather than `Path`?

gmittert: It is (surprisingly) actually `Path`. cmd.exe/powershell are case insensitive to it so `PATH` works, but the actual environment variable is `Path` if you print it out/open the environment variable editor.

compnerd: *blink* I hate windows.

compnerd:
CC: @millenomi - I don't think that there is a better option (without this, all the libraries must be either in the system directory or in the current directory.

Foundation/Process.swift
millenomi: Shouldn't this be `#if os(Windows)`?

compnerd: L381 :-)

PR #2386, opened 2019-06-27
Always Handle Broken Pipes on Windows

gmittert:
Getting an ERROR_BROKEN_PIPE shouldn't be an error in any case, not just
when we're reading untilEOF.


gmittert:

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

PR #1837, opened 2019-01-20
FileManager.attributesOfItem(atPath:) - use statx() on Linux if available.

spevans:
- Linux kernel >= 4.11.0 has a statx() call for extended file stat
   information and this can be used to set .creationDate

PR #2388, opened 2019-06-28
Fallback if statx syscall is not permitted

djones6:
## Summary

This patch adds a fallback in the `_stat_with_btime` function to `lstat`, in the case where the `statx` system call is available, but is denied (`EPERM`).

## Motivation / context

In https://github.com/apple/swift-corelibs-foundation/pull/1837 the `statx` system call was introduced to be used - when available - to get file attributes including creation time.  If the call isn't available (eg. on older kernels) then the `_stat_with_btime` function falls back to `lstat`.

However, the call may be available but fail for permission reasons: if running under Docker, both Docker's security profile and `libseccomp` need to know about (and permit) the new system call. `libseccomp` added support [in version `2.3.3`](https://github.com/seccomp/libseccomp/commit/5cbecad4e885d63ed6fd174b846b6aee09749cdc), and Docker [in version `18.04`](https://github.com/docker/for-linux/issues/208).

It happens that, in a Travis environment, you get an older version of libseccomp that doesn't know about (and denies) statx, and this results in a Swift Package Manager failure (trying to verify the presence of `swiftc` by querying its file attributes).
...I was able to work around this our CI by upgrading the `libseccomp` package before invoking Docker, but this patch will ensure it functions if statx exists, but is denied.

### DIFF
--- a/Foundation/FileManager+POSIX.swift
+++ b/Foundation/FileManager+POSIX.swift
@@ -927,13 +927,24 @@ extension FileManager {
     // structure and optional creation date.
 
     internal func _statxFile(atPath path: String) throws -> (stat, Date?) {
-        return try _fileSystemRepresentation(withPath: path) { fsRep in
+        // Fallback if statx() is unavailable or fails
+        func _statxFallback(atPath path: String, withFileSystemRepresentation fsRep: UnsafePointer<Int8>?) throws -> (stat, Date?) {
+            let statInfo = try _lstatFile(atPath: path, withFileSystemRepresentation: fsRep)
+            return (statInfo, nil)
+        }
 
+        return try _fileSystemRepresentation(withPath: path) { fsRep in
             if supportsStatx {
                 var statInfo = stat()
                 var btime = timespec()
-                guard _stat_with_btime(fsRep, &statInfo, &btime) == 0 else {
-                    throw _NSErrorWithErrno(errno, reading: true, path: path)
+                let statxErrno = _stat_with_btime(fsRep, &statInfo, &btime)
+                guard statxErrno == 0 else {
+                    switch statxErrno {
+                    case EPERM:
+                      return try _statxFallback(atPath: path, withFileSystemRepresentation: fsRep)
+                    default:
+                      throw _NSErrorWithErrno(statxErrno, reading: true, path: path)
+                    }
                 }
 
                 let sec = btime.tv_sec
@@ -947,9 +958,7 @@ extension FileManager {
                 }
                 return (statInfo, creationDate)
             } else {
-                // fallback if statx() is unavailable or fails
-                let statInfo = try _lstatFile(atPath: path, withFileSystemRepresentation: fsRep)
-                return (statInfo, nil)
+                return try _statxFallback(atPath: path, withFileSystemRepresentation: fsRep)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Foundation/FileManager+POSIX.swift
```diff
@@ -932,8 +932,14 @@ extension FileManager {
             if supportsStatx {
                 var statInfo = stat()
                 var btime = timespec()
-                guard _stat_with_btime(fsRep, &statInfo, &btime) == 0 else {
-                    throw _NSErrorWithErrno(errno, reading: true, path: path)
+                let statxResult = _stat_with_btime(fsRep, &statInfo, &btime)
```
spevans: I dont think we need the `_SyscallResult` struct since `errno` as used in `_NSErrorWithErrono` should have the correct value anyway.

Foundation/FileManager+POSIX.swift
```diff
@@ -932,8 +932,14 @@ extension FileManager {
             if supportsStatx {
```
spevans: Can we set `supportsStatx = false` after the `EPERM` to avoid later `_statx` calls that will fail.

Foundation/FileManager+POSIX.swift
```diff
@@ -947,12 +953,16 @@ extension FileManager {
...
+                return try _statxFallback(atPath: path, withFileSystemRepresentation: fsRep)
             }
         }
     }
+
+    // Fallback if statx() is unavailable or fails
+    private func _statxFallback(atPath path: String, withFileSystemRepresentation fsRep: UnsafePointer<Int8>?) throws -> (stat, Date?) {
+        let statInfo = try _lstatFile(atPath: path, withFileSystemRepresentation: fsRep)
+        return (statInfo, nil)
+    }
```
spevans: I would move this function to be inside `_statxFile(atPath:)` since that is the only place it is called from.

spevans:
@swift-ci test linux

Foundation/FileManager+POSIX.swift
djones6: Possibly.  But it's not clear (to me) whether statx could return `EPERM` for other reasons besides the syscall being denied.  From the docs, it seems like it would return `EACCES` if the user supplies an inaccessible path, so maybe it would be okay?

spevans: Ok lets leave it for now, it can always be added as an enhancement later if necessary.

ianpartridge:
@swift-ci test

ianpartridge:
@swift-ci test

djones6:
@spevans I've addressed your comments re. return types and squashed.  This should be ready for a re-test.

ianpartridge:
@swift-ci test

ianpartridge:
@swift-ci test

Foundation/FileManager+POSIX.swift
```diff
@@ -927,13 +927,24 @@ extension FileManager {
...
+        return try _fileSystemRepresentation(withPath: path) { fsRep in
             if supportsStatx {
                 var statInfo = stat()
                 var btime = timespec()
-                guard _stat_with_btime(fsRep, &statInfo, &btime) == 0 else {
-                    throw _NSErrorWithErrno(errno, reading: true, path: path)
+                let statxErrno = _stat_with_btime(fsRep, &statInfo, &btime)
+                guard statxErrno == 0 else {
+                    switch statxErrno {
+                    case 1:  // EPERM
```
millenomi: Please use EPERM here as a constant.

djones6: Thanks, I've pushed that change.

ianpartridge:
@swift-ci test

ianpartridge:
@swift-ci test

ianpartridge:
@swift-ci test

PR #2389, opened 2019-07-01
refactored NotificationCenter to use hash based storage for faster filtering

Cyberbeni:
I'm rewriting the NotificationCenter API for myself to use associated types and my performance tests have to be boosted on Linux because of the low performance. This should increase performance when NotificationCenter has a lot of observations.

You can find the performance tests in this repo: https://github.com/Cyberbeni/TypedNotificationCenter


millenomi:
@phausler does it look good?

millenomi:
@swift-ci please test

millenomi:
Looks like macOS is still problematic during testing. Ignore that failure.

PR #2390, opened 2019-07-01
build: prevent CoreFoundation from being installed

compnerd:
CoreFoundation is an internal implementation detail of Foundation.
Prevent CoreFoundation from being installed.


compnerd:
@swift-ci please test

PR #2392, opened 2019-07-02
build: remove RUNPATH from Foundation/FoundationNetworking

compnerd:
The executable's RUNPATH/RPATH will be used for the library search path.
Don't bother adding one to Foundation/FoundationNetworking.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,7 +100,6 @@ endif()
 set(plutil_rpath)
 if(CMAKE_SYSTEM_NAME STREQUAL Android OR CMAKE_SYSTEM_NAME STREQUAL Linux OR
     CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
-  set(Foundation_RPATH -Xlinker;-rpath;-Xlinker;"\\\$\$ORIGIN")
   set(XDG_TEST_HELPER_RPATH -Xlinker;-rpath;-Xlinker;${CMAKE_CURRENT_BINARY_DIR})
   set(plutil_rpath -Xlinker;-rpath;-Xlinker;"\\\$\$ORIGIN/../lib/swift/${swift_os}")
 elseif(CMAKE_SYSTEM_NAME STREQUAL Windows)
@@ -279,7 +278,6 @@ add_swift_library(Foundation
                     ${LIBXML2_LIBRARIES}
                     ${libdispatch_ldflags}
                     $<TARGET_FILE:uuid>
-                    ${Foundation_RPATH}
                     ${WORKAROUND_SR9138}
                     $<$<PLATFORM_ID:Windows>:-lDbgHelp>
                     $<$<PLATFORM_ID:Windows>:-lOle32>
@@ -352,7 +350,6 @@ add_swift_library(FoundationNetworking
                     ${Foundation_INTERFACE_LIBRARIES}
                     ${CFURLSessionInterface_LIBRARIES}
                     ${CURL_LIBRARIES}
-                    ${Foundation_RPATH}
                     ${WORKAROUND_SR9138}
                     ${WORKAROUND_SR9995}
                   SWIFT_FLAGS


compnerd:
@swift-ci please test

compnerd:
CC: @drodriguez

spevans:
I think this will break things. `$ORIGIN` was added to fix https://bugs.swift.org/browse/SR-9251 and I think the only reason the tests work without it is because the swift compiler is adding an `RPATH` to the build directory so even with the missing `$ORIGIN` things still work. Unfortunately I think the only way to verify this PR is to build a full linux dev toolchain then try to use with a test that using `Foundation` on a host without the build directory.

spevans:
Test using latest snapshot:
```
$ cat test.swift 
import Foundation
print(ProcessInfo.processInfo.operatingSystemVersion)
$ ~/swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04/usr/bin/swiftc -o test test.swift 
$ chrpath  -l ~/swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so 
/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so: RUNPATH=/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/libdispatch-linux-x86_64/src:$ORIGIN
$ ./test 
OperatingSystemVersion(majorVersion: 4, minorVersion: 18, patchVersion: 0)

# Remove the $ORIGIN
$ chrpath -r /home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/libdispatch-linux-x86_64/src ~/swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so
/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so: RUNPATH=/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/libdispatch-linux-x86_64/src:$ORIGIN
/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04/usr/lib/swift/linux/libFoundation.so: new RUNPATH: /home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux:/home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/build/buildbot_linux/libdispatch-linux-x86_64/src
$ ./test 
./test: error while loading shared libraries: libicuucswift.so.61: cannot open shared object file: No such file or directory

```

compnerd:
@spevans what happens if you add the `$ORIGIN` to the `DT_RUNPATH` on the executable?

spevans:
It works:
```
$ ln -s swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04 swift
$ chrpath  -r '/home/spse/swift/usr/lib/swift/linux:$ORIGIN' ./test
./test: RUNPATH=/home/spse/swift-DEVELOPMENT-SNAPSHOT-2019-07-01-a-ubuntu18.04/usr/lib/swift/linux
./test: new RUNPATH: /home/spse/swift/usr/lib/swift/linux:$ORIGIN
$ ./test 
OperatingSystemVersion(majorVersion: 0, minorVersion: 0, patchVersion: 24)
$ 
```

compnerd:
Okay, thats what I expect and believe is the right thing to do.  The library does not dictate the runpath that the loader consults, so we should just ensure that we alter the run path for the executable if the runtime is bundled.  Otherwise, we can assume that the runtime components are in the expected location.  Does that sound reasonable to you @spevans?

drodriguez:
If I understand the above output correctly, the `$ORIGIN` embedded in the executable means the path of the executable when loading libraries for the executable, but the path for the libraries when loading second level libraries for those first level libraries? That seems weird. If it works, perfect, but it is weird (unless I'm misunderstanding the layout of the files in the examples above). From https://refspecs.linuxfoundation.org/elf/gabi4+/ch5.dynamic.html#shobj_dependencies it seems to imply that `DT_RUNPATH` is only supposed to be used for first level dependencies. I wonder if `chrpath` is only creating `DT_RPATH` (and maybe removing `DT_RUNPATH`?).

I only think it is weird. If it works for everyone, it is fine for me.

spevans:
Apologies, I retested and my original test was broken (bad symlink). It doesnt work if the `$ORIGIN` is not in `libFoundation.so`. Without a `RUNPATH`, `libFoundation.so` simply uses the default `ld.so` `RUNPATH`, so this PR wont work.

PR #2393, opened 2019-07-05
[perf] JSON encoding can be faster by skipping string building.

drodriguez:
I realized that one of the test related to CharacterSet was very slow in
constrained devices, but also relatively slow in beefy Linux machines
(taking a couple of seconds to perform back and forth JSON
encoding/decoding trips in memory). The test in question have to embed
relatively large Data in the JSON string (around 128KB, if I remember
correctly), which are serialized as Base64.

When serializing the data, the serializer asks for the String
representation of each object, and then appends all of them into a
String, to later transform into an UTF-8 C string copied into a Data.
One can improve performance by appending the UTF-8 representation of
each piece of the JSON serialization instead of the String building
code, since appending [UInt8] without doing all the checking that String
will do is faster for this case.

### DIFF
--- a/Foundation/JSONSerialization.swift
+++ b/Foundation/JSONSerialization.swift
@@ -123,14 +123,14 @@ open class JSONSerialization : NSObject {
     /* Generate JSON data from a Foundation object. If the object will not produce valid JSON then an exception will be thrown. Setting the NSJSONWritingPrettyPrinted option will generate JSON with whitespace designed to make the output more readable. If that option is not set, the most compact possible JSON will be generated. If an error occurs, the error parameter will be set and the return value will be nil. The resulting data is a encoded in UTF-8.
      */
     internal class func _data(withJSONObject value: Any, options opt: WritingOptions, stream: Bool) throws -> Data {
-        var jsonStr = String()
+        var jsonStr = [UInt8]()
         
         var writer = JSONWriter(
             pretty: opt.contains(.prettyPrinted),
             sortedKeys: opt.contains(.sortedKeys),
             writer: { (str: String?) in
                 if let str = str {
-                    jsonStr.append(str)
+                    jsonStr.append(contentsOf: Array(str.utf8))
                 }
             }
         )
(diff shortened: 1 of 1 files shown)


drodriguez:
@swift-ci please test

drodriguez:
macOS seems to still fail with some Xcode error.

ianpartridge:
Cool. How much faster is it?

drodriguez:
IIRC the constrained device I was using reduced from 42 seconds to 32 seconds. I think this was all the `TestCodable/test_CharacterSet_JSON`. Those times were building without stdlib assertions enabled. With those, the test seemed not to finish in the same device (that was why I noticed). In a beefy Linux machine, but with the stdlib assertions, I think the time changed from 7 to 6 seconds, and it was not really significant without the stdlib assertions enabled.

millenomi:
cc @bendjones — can you take a quick look here?

millenomi:
And: is it worth it to port it to the overlay?

millenomi:
Per offline discussion with @bendjones — yup; for the overlay, the perf characteristics of `String`'s `.utf8` may not make this a win, depending on what goes on.

bendjones:
Seems fine for SCF to me and a reasonable change. As to the porting to the overlay question I’d just want to make sure the UTF8 guarantees are the same for bridged types (I think it’s fine but want to make sure and measure)

Foundation/JSONSerialization.swift
```diff
@@ -123,14 +123,14 @@ open class JSONSerialization : NSObject {
...
-        var jsonStr = String()
+        var jsonStr = [UInt8]()
         
         var writer = JSONWriter(
             pretty: opt.contains(.prettyPrinted),
             sortedKeys: opt.contains(.sortedKeys),
             writer: { (str: String?) in
                 if let str = str {
-                    jsonStr.append(str)
+                    jsonStr.append(contentsOf: Array(str.utf8))
```
bendjones: I missed this when I scanned this earlier but you could be able to avoid the array allocation by just doing `jsonStr.append(contentsOf: str.utf8)` right?

drodriguez: Yes, it seems it should work. I will prepare a follow up. Thanks for pointing it out!

PR #2394, opened 2019-07-07
build: remove CFLAGS from `add_swift_library`

compnerd:
This is in preparation for migrating to Swift support in CMake.  That
will allow us to use exported targets to track the dependencies properly
across projects.


compnerd:
@swift-ci please test

PR #2396, opened 2019-07-07
Use NSDecimalString(_:_) for Decimal -> String conversion.

spevans:
- This allows using an optional Locale to localise the output.

- Update Decimal.description, NSDecimalNumber.description and
  NSDecimalNumber.description(withLocale) to call NSDecimalString().

### DIFF
--- a/Foundation/Decimal.swift
+++ b/Foundation/Decimal.swift
@@ -267,61 +267,9 @@ extension Decimal : CustomStringConvertible {
         self = theDecimal
     }
 
-    // Note: In the Darwin overlay, `description` is implemented in terms of
-    // `NSDecimalString(_:_:)`; here, it's the other way around.
     public var description: String {
-        if self.isNaN {
-            return "NaN"
-        }
-        if _length == 0 {
-            return "0"
-        }
-        var copy = self
-        let ZERO: CChar = 0x30 // ASCII '0' == 0x30
-        let DECIMALPOINT: CChar = 0x2e // ASCII '.' == 0x2e
-        let MINUS: CChar = 0x2d // ASCII '-' == 0x2d
-
-        let bufferSize = 200 // Max value: 39 + 128 + sign + decimal point
-        var buffer = Array<CChar>(repeating: 0, count: bufferSize)
-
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test linux

Foundation/Decimal.swift
```diff
@@ -1691,10 +1639,71 @@ public func NSDecimalMultiplyByPowerOf10(_ result: UnsafeMutablePointer<Decimal>
...
+    if copy._isNegative != 0 {
+        i -= 1
+        buffer[i] = MINUS
+    }
+    let result = String(cString: Array(buffer.suffix(from: i)))
+
+    var decimalSeparator: String?
+    if let locale = locale as? Locale {
+        decimalSeparator = locale.decimalSeparator
+    } else if let dictionary = locale as? Dictionary<NSLocale.Key, String> {
```
millenomi: Should this be `[NSLocale.Key: Any]`?

Note that unlike Darwin, `NSLocale.Key` will not automatically bridge from `String` or `NSString`. Should we have a branch for that too?

TestFoundation/TestDecimal.swift
```diff
@@ -9,35 +9,6 @@
 
 class TestDecimal: XCTestCase {
 
-    static var allTests : [(String, (TestDecimal) -> () throws -> Void)] {
```
millenomi: Thanks for moving this.

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

Foundation/Decimal.swift
spevans: I added a branch for `[String: String]`. 

For the `[NSLocale.Key: Any]` case wouldn't that then mean `as!` casting the value to `String`? It that what you want it changed to?

millenomi: No, the current patch is good as-is.

millenomi: (After adding the branch.)

PR #2397, opened 2019-07-09
[android] Add errno.h header to the Android branch.

drodriguez:
The inline code in this file has started using the errno symbol, which
was included correctly in Linux, but not included in Android. This
commit adds the include to allow compiling the code.

@compnerd: this breakage haven't reach yet the Azure CI, but it will tonight, I think.


drodriguez:

spevans:
@swift-ci test linux

PR #2400, opened 2019-07-09
FileHandle: fix readInBackgroundAndNotify for Win32

compnerd:
`DispatchIO.read(fromFileDescriptor:maxLength:runningHandlerOn:)` is not
viable on Windows as the fileDescriptor is not available as we cannot
convert the handle to a file descriptor.  Extract the operation into a
local closure and dispatch from the right function in DispatchIO.


compnerd:
@swift-ci please test

PR #2401, opened 2019-07-10
Parity: URLCache adoption for HTTP and FTP invocations.

millenomi:
⚠️ Needs extensive testing and adding many of the fixes discussed in the comments of the PR linked below.

 - Includes the changes from previous PR: <https://github.com/apple/swift-corelibs-foundation/pull/2341>. They introduce an implementation of `URLCache`.

 - Store and provide the storage date for cached responses, for internal use only.

 - Implement protocol properties.

 - `URLSessionConfiguration` now configures the shared and ephemeral sessions with caches of appropriate size and storage setup. Each ephemeral session has a new in-memory-only cache, and the shared session used the shared `URLCache .

 - The `URLProtocolClient` that drives `URLSession` tasks now asks its delegate for caching, and if a cacheable response is determined, uses the `URLSession`’s cache to store it. It will also fetch from the cache if appropriate, and will initialize the protocol with that response.

 - Add hooks to `_NativeProtocol` to affect whether the transfer can be satisfied by the cached response it was initialized with. If it can, the response will be replayed instead.

 - `_HTTPURLProtocol` implements those hooks to make the `URLCache` behave as a RFC 7234 private cache. (Note that `_FTPURLProtocol` uses the default implementation of those hooks, which will cause related responses to be evicted from the cache and no response to be fulfilled from the cache. This is intentional.)

### DIFF
--- a/Foundation/URLCache.swift
+++ b/Foundation/URLCache.swift
@@ -53,7 +53,7 @@ class StoredCachedURLResponse: NSObject, NSSecureCoding {
     func encode(with aCoder: NSCoder) {
         aCoder.encode(cachedURLResponse.response, forKey: "response")
         aCoder.encode(cachedURLResponse.data as NSData, forKey: "data")
-        aCoder.encode(cachedURLResponse.storagePolicy.rawValue, forKey: "storagePolicy")
+        aCoder.encode(Int(bitPattern: cachedURLResponse.storagePolicy.rawValue), forKey: "storagePolicy")
         aCoder.encode(cachedURLResponse.userInfo as NSDictionary?, forKey: "userInfo")
         aCoder.encode(cachedURLResponse.date as NSDate, forKey: "date")
     }
@@ -61,7 +61,7 @@ class StoredCachedURLResponse: NSObject, NSSecureCoding {
     required init?(coder aDecoder: NSCoder) {
         guard let response = aDecoder.decodeObject(of: URLResponse.self, forKey: "response"),
               let data = aDecoder.decodeObject(of: NSData.self, forKey: "data"),
-              let storagePolicy = URLCache.StoragePolicy(rawValue: UInt(aDecoder.decodeInt64(forKey: "storagePolicy"))),
+            let storagePolicy = URLCache.StoragePolicy(rawValue: UInt(bitPattern: aDecoder.decodeInteger(forKey: "storagePolicy"))),
               let date = aDecoder.decodeObject(of: NSDate.self, forKey: "date") else {
                 return nil
         }
@@ -196,7 +196,7 @@ open class CachedURLResponse : NSObject, NSCopying {
 open class URLCache : NSObject {
     
     private static let sharedLock = NSLock()
-    private static var _shared = URLCache(memoryCapacity: 4 * 1024 * 1024, diskCapacity: 20 * 1024 * 1024, diskPath: nil)
+    private static var _shared: URLCache?
     
     /*! 
         @method sharedURLCache
@@ -218,7 +218,13 @@ open class URLCache : NSObject {
     open class var shared: URLCache {
         get {
             return sharedLock.performLocked {
-                return _shared
+                if let shared = _shared {
+                    return shared
+                }
+                
+                let shared = URLCache(memoryCapacity: 4 * 1024 * 1024, diskCapacity: 20 * 1024 * 1024, diskPath: nil)
+                _shared = shared
+                return shared
             }
         }
         set {
@@ -258,58 +264,38 @@ open class URLCache : NSObject {
     private var inMemoryCacheContents: [String: CacheEntry] = [:]
     
     func evictFromMemoryCacheAssumingLockHeld(maximumSize: Int) {
-        let sizes: [Int] = inMemoryCacheOrder.map {
-            inMemoryCacheContents[$0]!.cost
-        }
+        var totalSize = inMemoryCacheContents.values.reduce(0) { $0 + $1.cost }
         
-        var totalSize = sizes.reduce(0, +)
-        
-        guard totalSize > maximumSize else { return }
-        
-        var identifiersToRemove: Set<String> = []
-        for (index, identifier) in inMemoryCacheOrder.enumerated() {
-            identifiersToRemove.insert(identifier)
-            totalSize -= sizes[index]
-            if totalSize < maximumSize {
+        var countEvicted = 0
+        for identifier in inMemoryCacheOrder {
+            if totalSize > maximumSize {
+                countEvicted += 1
+                let entry = inMemoryCacheContents.removeValue(forKey: identifier)!
+                totalSize -= entry.cost
+            } else {
                 break
             }
         }
         
-        for identifier in identifiersToRemove {
-            inMemoryCacheContents.removeValue(forKey: identifier)
-        }
-        inMemoryCacheOrder.removeAll(where: { identifiersToRemove.contains($0) })
+        inMemoryCacheOrder.removeSubrange(0 ..< countEvicted)
     }
     
     func evictFromDiskCache(maximumSize: Int) {
-        var entries: [DiskEntry] = []
-        enumerateDiskEntries(includingPropertiesForKeys: [.fileSizeKey]) { (entry, stop) in
-            entries.append(entry)
+        let entries = diskEntries(includingPropertiesForKeys: [.fileSizeKey]).sorted {
+            $0.date < $1.date
         }
         
-        entries.sort { (a, b) -> Bool in
-            a.date < b.date
+        let sizes = entries.map { (entry) in
+            (try? entry.url.resourceValues(forKeys: [.fileSizeKey]).fileSize) ?? 0
         }
-        
-        let sizes: [Int] = entries.map {
-            return (try? $0.url.resourceValues(forKeys: [.fileSizeKey]).fileSize) ?? 0
-        }
-        
+
         var totalSize = sizes.reduce(0, +)
         
-        guard totalSize > maximumSize else { return }
-        
-        var urlsToRemove: [URL] = []
         for (index, entry) in entries.enumerated() {
-            urlsToRemove.append(entry.url)
-            totalSize -= sizes[index]
-            if totalSize < maximumSize {
-                break
+            if totalSize > maximumSize {
+                try? FileManager.default.removeItem(at: entry.url)
+                totalSize -= sizes[index]
             }
-        }
-        
-        for url in urlsToRemove {
-            try? FileManager.default.removeItem(at: url)
         }
     }
     
@@ -330,8 +316,10 @@ open class URLCache : NSObject {
         self.memoryCapacity = memoryCapacity
         self.diskCapacity = diskCapacity
         
+        let url: URL?
+        
         if let path = path {
-            cacheDirectory = URL(fileURLWithPath: path)
+            url = URL(fileURLWithPath: path)
         } else {
             do {
                 let caches = try FileManager.default.url(for: .cachesDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
@@ -342,15 +330,23 @@ open class URLCache : NSObject {
 
                 // We append a Swift Foundation identifier to avoid clobbering a Darwin cache that may exist at the same path;
                 // the two on-disk cache formats aren't compatible.
-                let url = caches
-                    .appendingPathComponent("org.swift.Foundation.URLCache", isDirectory: true)
+                url = caches
+                    .appendingPathComponent("org.swift.foundation.URLCache", isDirectory: true)
                     .appendingPathComponent(directoryName, isDirectory: true)
+            } catch {
+                url = nil
+            }
+        }
+        
+        if let url = url {
+            do {
                 try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
-                
                 cacheDirectory = url
             } catch {
                 cacheDirectory = nil
             }
+        } else {
+            cacheDirectory = nil
         }
     }
     
@@ -381,7 +377,7 @@ open class URLCache : NSObject {
         var identifier: String
         
         init?(_ url: URL) {
-            if url.pathExtension.localizedCompare(DiskEntry.pathExtension) != .orderedSame {
+            if url.pathExtension.caseInsensitiveCompare(DiskEntry.pathExtension) != .orderedSame {
                 return nil
             }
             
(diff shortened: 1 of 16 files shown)


millenomi:
cc @drodriguez

millenomi:
cc @ianpartridge

millenomi:
@swift-ci please test linux

Foundation/URLCache.swift
```diff
@@ -40,6 +47,38 @@ extension URLCache {
...
+    func encode(with aCoder: NSCoder) {
+        aCoder.encode(cachedURLResponse.response, forKey: "response")
+        aCoder.encode(cachedURLResponse.data as NSData, forKey: "data")
+        aCoder.encode(cachedURLResponse.storagePolicy.rawValue, forKey: "storagePolicy")
+        aCoder.encode(cachedURLResponse.userInfo as NSDictionary?, forKey: "userInfo")
+        aCoder.encode(cachedURLResponse.date as NSDate, forKey: "date")
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        guard let response = aDecoder.decodeObject(of: URLResponse.self, forKey: "response"),
```
millenomi: A variant I have re indentation style is that the conditions in multiline-expression `if` and `guard` statements align visually and have thus different indentation than both the `guard`/`if` and the body of the block.

Foundation/URLCache.swift
```diff
@@ -13,6 +13,13 @@ import SwiftFoundation
 import Foundation
 #endif
 
+internal extension NSLock {
+    func performLocked<T>(_ block: () throws -> T) rethrows -> T {
```
millenomi: `.synchronized` is an internal method in Foundation, and this code is in FoundationNetworking. — @karthikkeyan

Foundation/URLCache.swift
```diff
@@ -193,7 +326,123 @@ open class URLCache : NSObject {
...
+                try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
+                
+                cacheDirectory = url
+            } catch {
+                cacheDirectory = nil
+            }
+        }
+    }
+    
+    private func identifier(for request: URLRequest) -> String? {
```
millenomi: The identifier now uses only the `host`, `port` and `path` components of a URL to generate a key. — @karthikkeyan @drodriguez 

Note that this is not a HTTP cache, but a generic cache usable by any `URLProtocol` implementation — HTTP concerns go in the appropriate protocol implementation (and are included below).

Foundation/URLCache.swift
```diff
@@ -173,13 +217,102 @@ open class URLCache : NSObject {
...
+        static func ==(_ lhs: CacheEntry, _ rhs: CacheEntry) -> Bool {
+            return lhs.identifier == rhs.identifier
+        }
+    }
+    
+    private let inMemoryCacheLock = NSLock()
+    private var inMemoryCacheOrder: [String] = []
+    private var inMemoryCacheContents: [String: CacheEntry] = [:]
+    
+    func evictFromMemoryCacheAssumingLockHeld(maximumSize: Int) {
```
millenomi: @karthikkeyan I like the approach you had in the other patch; I'll have a follow-up commit here to adopt it.

millenomi:
@drodriguez I'm also concerned by the absence of locking, but we are both cross-platform (and thus can't rely on the _platform_ having locking) and not all volumes may support locking even on the same OS.

I'm thinking `flock`ing may be an okay thing to do tentatively, but I'd love the current implementation to not have any obvious incorrectness without filesystem locking (assuming that this is a cache and, as all caches do, it services its clients best-effort — I'm okay with a race evicting a cache value that may have been otherwise served.)

millenomi:
@swift-ci please test linux

TestFoundation/TestURLCache.swift
```diff
@@ -0,0 +1,271 @@
...
+        cache.storeCachedResponse(response, for: request)
+        
+        XCTAssertEqual(try FileManager.default.contentsOfDirectory(atPath: writableTestDirectoryURL.path).count, 1)
+        XCTAssertNotNil(cache.cachedResponse(for: request))
+        
+        // Ensure that the fulfillment doesn't come from memory:
+        try FileManager.default.removeItem(at: writableTestDirectoryURL)
+        try FileManager.default.createDirectory(at: writableTestDirectoryURL, withIntermediateDirectories: true)
+        
+        XCTAssertNil(cache.cachedResponse(for: request))
```
millenomi: @drodriguez The test here is explicitly about checking that the one `URLCache` object did use disk but didn't use memory if configured with `memoryCapacity: 0`.

TestFoundation/TestURLCache.swift
```diff
@@ -0,0 +1,271 @@
...
+        return (request, CachedURLResponse(response: response, data: Data(count: size), storagePolicy: storagePolicy))
+    }
+    
+    var writableTestDirectoryURL: URL!
+    
+    override func setUp() {
+        super.setUp()
+        
+        let pid = ProcessInfo.processInfo.processIdentifier
+        writableTestDirectoryURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("org.swift.TestFoundation.TestURLCache.\(pid)")
```
millenomi: @drodriguez re: PID — it is understood that there's no change; the attempt here is to avoid a proliferation of temporary directories in case we have a bug with the logic here.

Foundation/URLResponse.swift
```diff
@@ -32,21 +32,20 @@ open class URLResponse : NSObject, NSSecureCoding, NSCopying {
...
-            self.url = encodedUrl as URL
-        }
+        guard let nsurl = aDecoder.decodeObject(of: NSURL.self, forKey: "NS.url") else { return nil }
+        self.url = nsurl as URL
         
-        if let encodedMimeType = aDecoder.decodeObject(forKey: "NS.mimeType") as? NSString {
-            self.mimeType = encodedMimeType as String
-        }
+        
+        let nsmimetype = aDecoder.decodeObject(of: NSString.self, forKey: "NS.mimeType")
```
millenomi: @drodriguez re: differing patterns: you're right, I'll harmonize.

Foundation/URLCache.swift
```diff
@@ -40,6 +47,38 @@ extension URLCache {
...
+        aCoder.encode(cachedURLResponse.data as NSData, forKey: "data")
+        aCoder.encode(cachedURLResponse.storagePolicy.rawValue, forKey: "storagePolicy")
+        aCoder.encode(cachedURLResponse.userInfo as NSDictionary?, forKey: "userInfo")
+        aCoder.encode(cachedURLResponse.date as NSDate, forKey: "date")
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        guard let response = aDecoder.decodeObject(of: URLResponse.self, forKey: "response"),
+              let data = aDecoder.decodeObject(of: NSData.self, forKey: "data"),
+              let storagePolicy = URLCache.StoragePolicy(rawValue: UInt(aDecoder.decodeInt64(forKey: "storagePolicy"))),
```
millenomi: @drodriguez — Re: bitPattern usage: IIRC that will not compile on 32-bit platforms.

Foundation/URLCache.swift
```diff
@@ -154,6 +195,9 @@ open class CachedURLResponse : NSObject, NSCopying {
 
 open class URLCache : NSObject {
     
+    private static let sharedLock = NSLock()
+    private static var _shared = URLCache(memoryCapacity: 4 * 1024 * 1024, diskCapacity: 20 * 1024 * 1024, diskPath: nil)
```
millenomi: @drodriguez IIRC the language guarantees that, for static and global variables, their initializers aren't run until first access anyway.

Foundation/URLCache.swift
```diff
@@ -173,13 +217,102 @@ open class URLCache : NSObject {
...
+                _shared = newValue
+            }
         }
     }
-
+    
+    private let cacheDirectory: URL?
+    
+    private struct CacheEntry: Hashable {
+        var identifier: String
```
millenomi: @drodriguez Generally, I like keeping struct entries `var` unless there's a strict need; value type semantics prevent unintended mutation at usage sites anyway, and more complex mutation scenarios are aided by the ability to copy a value and edit the fields.

Foundation/URLCache.swift
```diff
@@ -173,13 +217,102 @@ open class URLCache : NSObject {
...
+    private let inMemoryCacheLock = NSLock()
+    private var inMemoryCacheOrder: [String] = []
+    private var inMemoryCacheContents: [String: CacheEntry] = [:]
+    
+    func evictFromMemoryCacheAssumingLockHeld(maximumSize: Int) {
+        let sizes: [Int] = inMemoryCacheOrder.map {
+            inMemoryCacheContents[$0]!.cost
+        }
+        
+        var totalSize = sizes.reduce(0, +)
```
millenomi: @drodriguez I'm avoiding duplication of logic here; I'd rather reason about clearer coed in this first pass, and address performance as needed.

Foundation/URLCache.swift
```diff
@@ -173,13 +217,102 @@ open class URLCache : NSObject {
...
+            totalSize -= sizes[index]
+            if totalSize < maximumSize {
+                break
+            }
+        }
+        
+        for identifier in identifiersToRemove {
+            inMemoryCacheContents.removeValue(forKey: identifier)
+        }
+        inMemoryCacheOrder.removeAll(where: { identifiersToRemove.contains($0) })
```
millenomi: @drodriguez I completely forgot about `dropFirst`, which obviously is always correct here. I'll adopt your suggestion.

Foundation/URLCache.swift
```diff
@@ -193,7 +326,123 @@ open class URLCache : NSObject {
...
+                    .replacingOccurrences(of: "/", with: "_")
+                    .replacingOccurrences(of: "\\", with: "_")
+                    .replacingOccurrences(of: ":", with: "_")
+
+                // We append a Swift Foundation identifier to avoid clobbering a Darwin cache that may exist at the same path;
+                // the two on-disk cache formats aren't compatible.
+                let url = caches
+                    .appendingPathComponent("org.swift.Foundation.URLCache", isDirectory: true)
+                    .appendingPathComponent(directoryName, isDirectory: true)
+                try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true)
```
millenomi: @drodriguez IIRC on Darwin the path, if passed, must exist, hence the logic here. I'll double-check that this is the case.

Foundation/URLCache.swift
```diff
@@ -193,7 +326,123 @@ open class URLCache : NSObject {
...
+    
+    private struct DiskEntry {
+        static let pathExtension = "storedcachedurlresponse"
+        
+        var url: URL
+        var date: Date
+        var identifier: String
+        
+        init?(_ url: URL) {
+            if url.pathExtension.localizedCompare(DiskEntry.pathExtension) != .orderedSame {
```
millenomi: @drodriguez good catch here — I intended to use `caseInsensitiveCompare`.

Foundation/URLCache.swift
```diff
@@ -193,7 +326,123 @@ open class URLCache : NSObject {
...
+                if stop {
+                    return
+                }
+            }
+        }
+    }
+    
+    private func diskContentsURL(for request: URLRequest, forCreationAt date: Date? = nil) -> URL? {
+        guard let identifier = self.identifier(for: request) else { return nil }
+        guard let directory = cacheDirectory else { return nil }
```
millenomi: @drodriguez also good catch here — I'll swap these two.

Foundation/URLCache.swift
```diff
@@ -193,7 +326,123 @@ open class URLCache : NSObject {
...
+            if entry.identifier == identifier {
+                foundURL = entry.url
+                stop = true
+            }
+        }
+        
+        if let date = date {
+            // If we're trying to _create_ an entry and it already exists, then we can't -- we should evict the old one first.
+            if foundURL != nil {
+                return nil
```
millenomi: @drodriguez I cannot _guarantee_ that this check will not be executed even around L491, because we may be racing. You're correct that a behavior guarantee would be easier to achieve with filesystem locking, but as noted we may be operating without and not have a choice in the matter.

I need to sit down and think more deeply about the date being part of the filename, since it may ruin what little atomicity we have in `store…` and lead to duplicate entries for the same URL. It may be that we can just blindly store and then do an eviction pass at the end for all URLs that have the same key and are not _the highest timestamp we find on disk_ (not the one we just stored!), but I need to think through the racing implications of that.

Foundation/URLCache.swift
drodriguez: Yes, that's true, but that initialization seems to happen even if the static variable is a lvalue, like in line 226.

You can type the following in an Xcode playground and check:

```swift
var globalCounter = 0

class TestValue {
    init() {
        print("TestValue.init")
        globalCounter += 1
    }
}

class TestClass {
    private static var _static: TestValue = TestValue()
    
    class var accessor: TestValue {
        get {
            print("TestClass.accessor.get: before return")
            return _static
        }
        set {
            print("TestClass.accessor.set: before assign")
            _static = newValue
            print("TestClass.accessor.set: after assign")
        }
    }
}

let t = TestValue()
TestClass.accessor = t
print(globalCounter)

// Output:
// TestValue.init
// TestClass.accessor.set: before assign
// TestValue.init
// TestClass.accessor.set: after assign
// 2
```

TestFoundation/TestURLCache.swift
drodriguez: I proposed it because using the same value in the different test might mean that a later test is affected by the previous one during the same run. It could lead to hard to debug problems where the test work (or doesn’t) when run in a suite, but fails (or works) when run individually.

It is not a big deal anyway.

Foundation/URLCache.swift
drodriguez: Yes, my comments about locking the file storage wasn't about locking it against two processes accessing the same cache, but about the same process accessing its own cache from two threads. My original design serialized many of the access to disk to avoid those kind of problems. Knowing that the access was serial simplified many of the checks a lot.

millenomi: The problem with that reasoning is, given that `URLCache.shared` is the default, it is not useful to do that anyway — the moment a second process runs on the same node it will use the same path. So we need to be flexible enough to work in that case anyway.

drodriguez: Two processes of the same binary will create problems with the file system storage. That's unavoidable without file level locking. However I don’t think the situation happens that often for consumer software (different story for server software, though). My intention providing in-process serialization was to honor the documentation that states that `URLCache` is thread safe.

Foundation/URLCache.swift
millenomi: It was indeed not the case.

Foundation/URLCache.swift
millenomi: The moment we have to reason about safety at an interprocess level without locking, if we get it right, we by definition have safety at an intra-process, inter-thread level without locking. At that point, adding locking is at best a performance issue and at worst hides interprocess issues that we would've caught otherwise.

millenomi:
cc @drodriguez many fixes.

Foundation/URLCache.swift
millenomi: Done. Note that it had to be reworked: side effects in `.prefix` are suspect, and I inlined the loops a little better.

Foundation/URLCache.swift
millenomi: Fixed.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test Linux

millenomi:
@drodriguez merging today if you see no issue; I want to make sure that our existing URLSession tests are hitting the cache before merging, but I would be satsified if there are no regressions there.

drodriguez:
@millenomi if you need to merge, please do. I haven't had time to go over the new code, but we can always fix them after it lands. This has a lot more code than the original PR. I have been only answering to the comments brought from the previous PR.

millenomi:
I will merge as soon as merging is re-enabled after the build fix.

millenomi:
I have verified that the `URLSession` tests are using `URLSession.shared`, meaning they're exercising the `URLCache.shared`.

Foundation/URLSession/URLSessionTask.swift
```diff
@@ -252,8 +336,10 @@ open class URLSessionTask : NSObject, NSCopying {
             self.updateTaskState()
             
             if self.suspendCount == 1 {
-                self.workQueue.async {
-                    self._protocol?.stopLoading()
+                self._getProtocol { (urlProtocol) in
+                    self.workQueue.async {
```
drodriguez: It is complicated to determine, but with the `workQueue` and the protocol being invalidated inside `_protocolLock`, some of these pieces of code might not have the same behaviour and might not be expected. If `_getProtocol` here returns a non-`nil` protocol, it might have already been invalidated by the time the `workQueue` executes the code. That didn't happen in the previous version.

I don't know if that's important or not, however.

Foundation/URLSession/http/HTTPURLProtocol.swift
```diff
@@ -60,6 +60,202 @@ internal class _HTTPURLProtocol: _NativeProtocol {
...
+        }
+        
+        // Cache-Control: https://tools.ietf.org/html/rfc7234#section-5.2
+        var hasCacheControl = false
+        var hasMaxAge = false
+        if let cacheControl = httpResponse.allHeaderFields["Cache-Control"] as? String {
+            let directives = CacheControlDirectives(headerValue: cacheControl)
+            
+            if directives.noCache || directives.noStore {
+                return false
```
drodriguez: My understanding of `no-cache` is that the caches should revalidate with the origin server, but the cached content can still be used if the server answers “Not modified”.

PR #2402, opened 2019-07-11
[android] Mark link test in TestFileManager as failing for Android.

drodriguez:
Implement support for marking test as expected to fail in Android
similar to what Windows has already. Use the helper function to mark
the tests that were already disabled in Android (there are some others
with extra checks that should be done more carefully).

Additionally, mark one of the tests of TestFileManager as failing in
Android. Android doesn't allow normal programs to use the link syscall,
and it always fails with insuficient permissions. The test will never
work. Disable the test in Android completely. Users should deal with the
always failing API, since errors are always possible for other reason.

Also, remove a disabled test in TestURLSession because I cannot make it
fail in my setup. It might have fail before, but it not longer fails.


drodriguez:

compnerd:
Sounds like a good idea to me.  @millenomi any concerns?

PR #2406, opened 2019-07-12
build: don't link C++ runtime on android

compnerd:


millenomi:

millenomi:
Er, it'll fail because Mac. Oh well.

compnerd:
Lets hold off on this for just a tad, I think I have a much better generic solution that I can add to swift and have it cover all the targets.

PR #2407, opened 2019-07-12
build: add DT_SONAME for ELF libraries

compnerd:
Unfortunately, because we do not use CMake for generating the libraries,
we need to explicitly control the library names for the time being.  Add
a DT_SONAME entry to the ELF binary that we generate to ensure that we
are able to properly link to the libraries at load time.


compnerd:

PR #2408, opened 2019-07-12
[Windows] Overwrite the destination if it exists in _copySymlink

gmittert:
The POSIX implementation will overwrite the destination so the Windows
version should do the same.

In addition, since we are copying a symlink, we know the type of the
target, so we can specify this to createSymlink. This is useful when
copying directories and you copy a symlink before you copy its target.
This allows you to create a (temporarily) broken symlink of the right
file type rather than having to copy the target first.

### DIFF
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -276,19 +276,30 @@ extension FileManager {
         return result
     }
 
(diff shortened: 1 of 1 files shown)


gmittert:
@swift-ci please test linux

PR #2409, opened 2019-07-12
[Windows] Fix _fileExists for relative links

gmittert:
A relative link should be determined from the path that it is in, not
the current working directory


gmittert:
@swift-ci please test linux

PR #2411, opened 2019-07-12
Implement _contentsEqual on Windows

gmittert:
Compared to the POSIX implementation, this:
- Uses the same _compareFile and _compareDirectory utilities.
- Implements Symlink comparison with `_destinationOfSymbolicLink`
- Doesn't handle devices since there's no reasonable way to check for
  that equivalence on Windows (as far as I'm aware). If given a device, it
  will instead always return false. This is probably fine since devices
  tend to be less file like on Windows anyways.

### DIFF
--- a/Foundation/FileManager+POSIX.swift
+++ b/Foundation/FileManager+POSIX.swift
@@ -853,34 +853,6 @@ extension FileManager {
         }
     }
 
-    private func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64, bufSize: Int) -> Bool {
-        guard let file1 = FileHandle(fileSystemRepresentation: file1Rep, flags: O_RDONLY, createMode: 0) else { return false }
-        guard let file2 = FileHandle(fileSystemRepresentation: file2Rep, flags: O_RDONLY, createMode: 0) else { return false }
-
-        var buffer1 = UnsafeMutablePointer<UInt8>.allocate(capacity: bufSize)
-        var buffer2 = UnsafeMutablePointer<UInt8>.allocate(capacity: bufSize)
-        defer {
-            buffer1.deallocate()
-            buffer2.deallocate()
-        }
-        var bytesLeft = size
-        while bytesLeft > 0 {
-            let bytesToRead = Int(min(Int64(bufSize), bytesLeft))
-
-            guard let file1BytesRead = try? file1._readBytes(into: buffer1, length: bytesToRead), file1BytesRead == bytesToRead else {
-                return false
-            }
-            guard let file2BytesRead = try? file2._readBytes(into: buffer2, length: bytesToRead), file2BytesRead == bytesToRead else {
-                return false
-            }
-            guard memcmp(buffer1, buffer2, bytesToRead) == 0 else {
-                return false
-            }
-            bytesLeft -= Int64(bytesToRead)
-        }
-        return true
-    }
-
     private func _compareSymlinks(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64) -> Bool {
         let bufSize = Int(size)
         let buffer1 = UnsafeMutablePointer<CChar>.allocate(capacity: bufSize)
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -589,7 +589,7 @@ extension FileManager {
 
                     if ffd.dwFileAttributes & DWORD(FILE_ATTRIBUTE_READONLY) == FILE_ATTRIBUTE_READONLY {
                         let readableAttributes = ffd.dwFileAttributes & DWORD(bitPattern: ~FILE_ATTRIBUTE_READONLY)
-                        guard file.withCString(encodedAs: UTF16.self, { SetFileAttributesW($0, readableAttributes) }) else {
+                        guard itemPath.withCString(encodedAs: UTF16.self, { SetFileAttributesW($0, readableAttributes) }) else {
                             throw _NSErrorWithWindowsError(GetLastError(), reading: false, paths: [file])
                         }
                     }
@@ -684,10 +684,6 @@ extension FileManager {
         return true
     }
 
-    internal func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64, bufSize: Int) -> Bool {
-        NSUnimplemented()
-    }
-
     internal func _lstatFile(atPath path: String, withFileSystemRepresentation fsRep: UnsafePointer<Int8>? = nil) throws -> stat {
         let _fsRep: UnsafePointer<Int8>
         if fsRep == nil {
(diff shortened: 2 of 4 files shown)


Foundation/FileManager+POSIX.swift
```diff
@@ -853,34 +853,6 @@ extension FileManager {
         }
     }
 
-    private func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<Int8>, andFileSystemRepresentation file2Rep: UnsafePointer<Int8>, size: Int64, bufSize: Int) -> Bool {
```
gmittert: This is moved to FileManager since it also works for Windows.

Foundation/FileManager.swift
```diff
@@ -825,6 +825,8 @@ open class FileManager : NSObject {
         guard let enumerator2 = enumerator(atPath: path2) else {
             return false
         }
+        enumerator1.skipDescendants()
+        enumerator2.skipDescendants()
```
gmittert: In both the Windows and POSIX implementations, `_compareDirectories` and `_contentsEqual` are mutually recursive. Not skipping descendants causes `_contentsEqual` to all `_compareDirectories` on every subdirectory which will each then call `_contentsEqual` on every one of their children causing excessive comparisons. This should fix **a lot** of unneeded recursion.

TestFoundation/TestFileManager.swift
```diff
@@ -1769,6 +1800,7 @@ VIDEOS=StopgapVideos
             ("test_getRelationship", test_getRelationship),
             ("test_displayNames", test_displayNames),
             ("test_getItemReplacementDirectory", test_getItemReplacementDirectory),
+            ("test_contentsEqual", test_contentsEqual),
```
gmittert: I was _very_ excited when the contentsEqual test didn't fail when I ran the suite... until I figured out why.

millenomi:
@swift-ci please test linux

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
     }
 
     internal func _contentsEqual(atPath path1: String, andPath path2: String) -> Bool {
-        NSUnimplemented()
+        guard let path1Handle = path1.withCString(encodedAs: UTF16.self, {
+            CreateFileW($0, GENERIC_READ, DWORD(FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE), nil,
+                        DWORD(OPEN_EXISTING), DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS), nil)
+        }) else {
```
compnerd: Does this evaluate properly for `INVALID_HANDLE_VALUE`?  The return value in case of error is not 0.

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+            return false
+        }
+        defer { CloseHandle(path1Handle) }
+
+        guard let path2Handle = path2.withCString(encodedAs: UTF16.self, {
+            CreateFileW($0, GENERIC_READ, DWORD(FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE), nil,
+                        DWORD(OPEN_EXISTING), DWORD(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS), nil)
+        }) else {
```
compnerd: Similar

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+            return false
+        }
+        defer { CloseHandle(path2Handle) }
+
+        guard path1Handle != INVALID_HANDLE_VALUE, path2Handle != INVALID_HANDLE_VALUE else {
```
compnerd: Should these not be before the `defer`?

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+        let file1Type = GetFileType(path1Handle)
+        guard GetLastError() == NO_ERROR else {
+            return false
+        }
+        let file2Type = GetFileType(path2Handle)
+        guard GetLastError() == NO_ERROR else {
+            return false
+        }
+
+        guard file1Type == file2Type, file1Type == FILE_TYPE_DISK else {
```
compnerd: I think that writing this as `file1Type == FILE_TYPE_DISK, file2Type == FILE_TYPE_DISK` is nicer.

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+        var path1FileInfo = BY_HANDLE_FILE_INFORMATION()
+        var path2FileInfo = BY_HANDLE_FILE_INFORMATION()
+        guard GetFileInformationByHandle(path1Handle, &path1FileInfo),
+              GetFileInformationByHandle(path2Handle, &path2FileInfo) else {
+            return false
+        }
+
+        // If both paths point to the same volume/filenumber or they are both zero length
+        // then they are considered equal
+        let sameFile = path1FileInfo.nFileIndexHigh == path2FileInfo.nFileIndexHigh
```
compnerd: Just inline the single use variable.

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+              && path1FileInfo.dwVolumeSerialNumber == path2FileInfo.dwVolumeSerialNumber
+
+        if sameFile {
+            return true
+        }
+
+        let path1Attrs = path1FileInfo.dwFileAttributes
+        let path2Attrs = path2FileInfo.dwFileAttributes
+        if DWORD(FILE_ATTRIBUTE_REPARSE_POINT) & path1Attrs == DWORD(FILE_ATTRIBUTE_REPARSE_POINT)
+             || DWORD(FILE_ATTRIBUTE_REPARSE_POINT) & path2Attrs == FILE_ATTRIBUTE_REPARSE_POINT {
```
compnerd: Why the inverted check? (i.e. why not `path1Attr & DWORD(FILE_ATTRIBUTE_REPARSE_POINT) == DWORD(FILE_ATTRIBUTE_REPARSE_POINT)`?)

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+            guard DWORD(FILE_ATTRIBUTE_REPARSE_POINT) & path1Attrs == DWORD(FILE_ATTRIBUTE_REPARSE_POINT)
+                    && DWORD(FILE_ATTRIBUTE_REPARSE_POINT) & path2Attrs == FILE_ATTRIBUTE_REPARSE_POINT else {
+                return false
+            }
```
compnerd: Huh?  What am I missing?  Isn't this the condition that is in the `if`?

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+                  let pathDest2 = try? _destinationOfSymbolicLink(atPath: path2) else {
+                return false
+            }
+            return pathDest1 == pathDest2
+        } else if DWORD(FILE_ATTRIBUTE_DIRECTORY) & path1Attrs == DWORD(FILE_ATTRIBUTE_DIRECTORY)
+                    || DWORD(FILE_ATTRIBUTE_DIRECTORY) & path2Attrs == DWORD(FILE_ATTRIBUTE_DIRECTORY) {
+            guard DWORD(FILE_ATTRIBUTE_DIRECTORY) & path1Attrs == DWORD(FILE_ATTRIBUTE_DIRECTORY)
+                    && DWORD(FILE_ATTRIBUTE_DIRECTORY) & path2Attrs == FILE_ATTRIBUTE_DIRECTORY else {
+                return false
+            }
```
compnerd: Similar

Foundation/FileManager+Win32.swift
```diff
@@ -648,7 +644,95 @@ extension FileManager {
...
+            return _compareDirectories(atPath: path1, andPath: path2)
+        } else {
+            let bothEmpty = path1FileInfo.nFileSizeHigh == 0
+              && path1FileInfo.nFileSizeLow == 0
+              && path2FileInfo.nFileSizeHigh == 0
+              && path2FileInfo.nFileSizeLow == 0
+
+            if bothEmpty {
```
compnerd: Inline the single use variable

Foundation/FileManager+Win32.swift
gmittert: The inner one is `&&`, the outer is `||`. If either is a reparse point, go into the `if`, but if one is not a reparse point, the immediately bail out. Sadly there's no logical XOR which might make that clearer.

Foundation/FileManager+Win32.swift
gmittert: I swore I read somewhere it was safe to call CloseHandle on an `INVALID_HANDLE_VALUE` but the docs definitely say it's not...

Foundation/FileManager+Win32.swift
compnerd: Hmm, does `^` not work here?

compnerd:
@millenomi - any additional comments from you?

millenomi:
@swift-ci please test Linux

gmittert:
@swift-ci please test linux

spevans:
@swift-ci please test linux

compnerd:
I think that all the review comments have been addressed, lets get this merged.

PR #2412, opened 2019-07-12
[test] Remove external networking in TestHost

drodriguez:
Ideally the test suite should not depend on external factors. TestHost
was doing DNS requests to turn Host names into Host addresses.

To try to avoid external factors, one can use localhost, which should be
resolved locally all the time, and the check should work the same.

There's an usage of another external name in the test. Sadly there's no
guarantees of another name being defined locally. Some machines have an
IPv6 specific name, but it is not as standard as localhost. Luckily, the
test will work with an external name, even if it cannot be resolved,
since it should not be the same as localhost for most people.


spevans:
@swift-ci test linux

PR #2413, opened 2019-07-13
[Windows] Fix TestURL.swift crash

gmittert:
On Windows, `|` is a valid character in a URL (as a replacement to `:`,
e.g. `/C|/Users/gwenm`. This causes a test testing
`http://test.com/unescaped|pipe` which is expected  to not parse to
parse successfully on Windows. The test then tries to parse the `<null
url>` in the expected result and crashes.

This changes the test to check to make sure it can parse the expected
result instead of force casting so that the test fails instead of
crashing.

This allows TestURL to run to completion in the mean time while working
on URL for Windows.

### DIFF
--- a/TestFoundation/TestURL.swift
+++ b/TestFoundation/TestURL.swift
@@ -253,8 +253,12 @@ class TestURL : XCTestCase {
             } else {
                 if let url = url {
                         let results = generateResults(url, pathComponent: inPathComponent, pathExtension: inPathExtension)
-                        let (isEqual, differences) = compareResults(url, expected: expectedNSResult as! [String: Any], got: results)
-                        XCTAssertTrue(isEqual, "\(title): \(differences.joined(separator: "\n"))")
+                        if let expected = expectedNSResult as? [String: Any] {
+                            let (isEqual, differences) = compareResults(url, expected: expected, got: results)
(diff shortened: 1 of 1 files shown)


gmittert:
@swift-ci please test

millenomi:
I need to look closely at this — if `|` is not allowed in a URL path, that doesn't change between platforms.

millenomi:
@gmittert Why is URL parsing different on Windows?

millenomi:
(URL parsing is anchored to a standard; what is the use case for deviating for it? Am I missing something re: it being already the case in CF?)

gmittert:
Yeah, it's already accepted in CF on Windows: https://github.com/apple/swift-corelibs-foundation/blob/master/CoreFoundation/URL.subproj/CFURL.c#L1980

From the Standard: https://tools.ietf.org/html/rfc8089#appendix-E.2.2

```
   Historically, some usages of file URIs have included a vertical line
   character "|" instead of a colon ":" in the drive letter construct.
   [RFC3986] forbids the use of the vertical line; however, it may be
   necessary to interpret or update old URIs.
...
 This is intended to support regular DOS or Windows file URIs with
   vertical line characters in the drive letter construct.  For example:

   o  "file:///c|/path/to/file"
```

millenomi:
Approved.

PR #2415, opened 2019-07-15
build: correct the `DT_SONAME`

compnerd:
When ninja is invoked from a subdirectory it will convert the path to a
relative path, and that resulted in incorrect test results.  This
ensures that we get the correct soname.


millenomi:
@swift-ci please test Linux

aciidb0mb3r:
I've tested that this patch fixes the issue.

millenomi:
The failure appears unrelated.

millenomi:
@swift-ci please test Linux

millenomi:
If we have another unrelated error, I'll ask @shahmishal to force-merge.

PR #2418, opened 2019-07-17
SR-11150: NSDecimalNumber .multiplying(byPowerOf10:) produces incorrect results.

spevans:
- Call NSDecimalMultiplyByPowerOf10() instead of NSDecimalPower() to
  multiply the number by 10 raised to the given power, not raise
  the number to a given power.


spevans:
@swift-ci test linux

PR #2420, opened 2019-07-17
Remove unnecessary Array copy.

drodriguez:
Array.append(contentsOf:) supports any Sequence, and String.UTF8View is
a sequence, so there's no need to create an Array from it to append it.

From a post-merge feedback in #2393.


drodriguez:

PR #2421, opened 2019-07-18
Parity: URLCredentialStorage

millenomi:
- Do not crash when a synchronizable credential is added. Instead, act as Darwin does in similar situations (do not store). To make sure the difference is visible, diagnose uses of `.synchronized`.

 - `URLSession` is now correctly initialized with appropriate credential storages via its configuration.

 - Credentials are fetched appropriately from storage, and when a delegate is consulted about authentication fetched credentials are proposed as part of the auth challenge correctly. (If multiple credentials match, always propose the first in lexicographical order.)

 - Interpret the disposition of a delegate if queried. If `.performDefaultHandling` is used, or if the task has no delegate, use the default proposed credential once automatically before giving up.

 - When a task finishes, if a credential was used, save it in the credential storage if appropriate.

### DIFF
--- a/Foundation/URLSession/URLSession.swift
+++ b/Foundation/URLSession/URLSession.swift
@@ -212,8 +212,6 @@ open class URLSession : NSObject {
     fileprivate static let _shared: URLSession = {
         var configuration = URLSessionConfiguration.default
         configuration.httpCookieStorage = HTTPCookieStorage.shared
-        //TODO: Set urlCache to URLCache.shared. Needs implementation of URLCache.
-        //TODO: Set urlCredentialStorage to `URLCredentialStorage.shared`. Needs implementation of URLCredentialStorage.
         configuration.protocolClasses = URLProtocol.getProtocols()
         return URLSession(configuration: configuration, delegate: nil, delegateQueue: nil)
     }()
--- a/Foundation/URLSession/URLSessionConfiguration.swift
+++ b/Foundation/URLSession/URLSessionConfiguration.swift
@@ -72,7 +72,7 @@ open class URLSessionConfiguration : NSObject, NSCopying {
                   httpAdditionalHeaders: nil,
                   httpMaximumConnectionsPerHost: 6,
                   httpCookieStorage: .shared,
-                  urlCredentialStorage: nil, // Should be .shared once implemented.
+                  urlCredentialStorage: .shared,
                   urlCache: .shared,
                   shouldUseExtendedBackgroundIdleMode: false,
                   protocolClasses: [_HTTPURLProtocol.self, _FTPURLProtocol.self])
@@ -148,11 +148,9 @@ open class URLSessionConfiguration : NSObject, NSCopying {
     }
 
     open class var ephemeral: URLSessionConfiguration {
-        // Return a new ephemeral URLSessionConfiguration every time this property is invoked
-        // TODO: urlCredentialStorage should also be ephemeral/in-memory
-        // URLCredentialStorage is still unimplemented
         let ephemeralConfiguration = URLSessionConfiguration.default.copy() as! URLSessionConfiguration
         ephemeralConfiguration.httpCookieStorage = .ephemeralStorage()
+        ephemeralConfiguration.urlCredentialStorage = URLCredentialStorage() // All credential storage in s-c-f are ephemeral.
         ephemeralConfiguration.urlCache = URLCache(memoryCapacity: 4 * 1024 * 1024, diskCapacity: 0, diskPath: nil)
         return ephemeralConfiguration
     }
--- a/Foundation/URLSession/URLSessionConfiguration.swift
+++ b/Foundation/URLSession/URLSessionConfiguration.swift
@@ -150,7 +150,7 @@ open class URLSessionConfiguration : NSObject, NSCopying {
     open class var ephemeral: URLSessionConfiguration {
         let ephemeralConfiguration = URLSessionConfiguration.default.copy() as! URLSessionConfiguration
         ephemeralConfiguration.httpCookieStorage = .ephemeralStorage()
-        ephemeralConfiguration.urlCredentialStorage = URLCredentialStorage() // All credential storage in s-c-f are ephemeral.
+        ephemeralConfiguration.urlCredentialStorage = URLCredentialStorage(ephemeral: true)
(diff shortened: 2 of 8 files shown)


millenomi:
The `URLSession` tests will exercise this code, but not specific uses of `URLCredentialStorage`.

millenomi:
@swift-ci please test Linux

millenomi:
cc @drodriguez

Foundation/URLSession/URLSession.swift
```diff
@@ -212,8 +212,6 @@ open class URLSession : NSObject {
     fileprivate static let _shared: URLSession = {
         var configuration = URLSessionConfiguration.default
         configuration.httpCookieStorage = HTTPCookieStorage.shared
```
drodriguez: I think the cookie storage line can also be removed.

Foundation/URLSession/URLSessionConfiguration.swift
```diff
@@ -148,11 +148,9 @@ open class URLSessionConfiguration : NSObject, NSCopying {
     }
 
     open class var ephemeral: URLSessionConfiguration {
-        // Return a new ephemeral URLSessionConfiguration every time this property is invoked
-        // TODO: urlCredentialStorage should also be ephemeral/in-memory
-        // URLCredentialStorage is still unimplemented
         let ephemeralConfiguration = URLSessionConfiguration.default.copy() as! URLSessionConfiguration
         ephemeralConfiguration.httpCookieStorage = .ephemeralStorage()
+        ephemeralConfiguration.urlCredentialStorage = URLCredentialStorage() // All credential storage in s-c-f are ephemeral.
```
drodriguez: Maybe change to `NOTE: all credential storage in s-c-f is ephemeral at the moment`, since someone might implement `.permanent` at some point for each platform.

Foundation/URLSession/URLSessionTask.swift
```diff
@@ -682,16 +683,49 @@ extension _ProtocolClient : URLProtocolClient {
...
+                    proceed(proposing: nil)
+                }
+                
                 return
             }
         }
         
+        if let storage = session.configuration.urlCredentialStorage,
+           let last = task._protocolLock.performLocked({ task._lastCredentialUsedFromStorageDuringAuthentication }),
+           last.credential.persistence != .none && last.credential.persistence != .forSession {
```
drodriguez: `URLCredentialStorage.set(_, for:, task:)` should take care of `.none`. I don't understand completely why `.forSession` is also discarded here. A session is suppose to last until the credential storage stops being used, if I understand correctly.

Foundation/URLSession/URLSessionTask.swift
```diff
@@ -758,25 +792,65 @@ extension _ProtocolClient : URLProtocolClient {
...
         guard let session = task.session as? URLSession else { fatalError("Task not associated with URLSession.") }
+        
+        func proceed(using credential: URLCredential?) {
+            let protectionSpace = challenge.protectionSpace
+            let authScheme = protectionSpace.authenticationMethod
+
+            task.suspend()
+            
+            guard let handler = URLSessionTask.authHandler(for: authScheme) else {
+                fatalError("\(authScheme) is not supported")
```
drodriguez: From the docs of `urlSession(_:didReceive:completionHandler:)`: "This method handles only the NSURLAuthenticationMethodNTLM, NSURLAuthenticationMethodNegotiate, NSURLAuthenticationMethodClientCertificate, and NSURLAuthenticationMethodServerTrust authentication types. For all other authentication schemes, the session calls only the urlSession(_:task:didReceive:completionHandler:) method."

I don't know exactly, but this might be better as a `task.resume()` and `return` to end up in that other delegate call.

Foundation/URLSession/URLSessionConfiguration.swift
millenomi: Maybe I should just have a marker `init(ephemeral:)` initializer, so it's clear to a future maintainer that some storages must be ephemeral.

Foundation/URLSession/URLSessionTask.swift
millenomi: Strategic implementation choice: all storage is best effort, so we can choose not to support `.forSession` storage (because we do not have a safe way to associate a credential with a session right now).

Foundation/URLSession/URLSessionTask.swift
millenomi: Good catch. This is replicating existing behavior, so perhaps that merits a separate patch so we can test regression separately.

millenomi: Thanks, good catch. I'll amend.

millenomi: I need to sit down and evaluate what the impact is of making that change, because we are changing current s-c-f behavior (to realign, but I may want to make a separate patch for this just so I can run tests on only that part).

Foundation/URLSession/URLSessionTask.swift
drodriguez: I will have to double check the documentation/Darwin implementation, but I always interpreted .forSession as “only in memory”. I also think that this check might be better in the cache, since someone can implement externally  a cache that does .forSession, but not if this check is done here. I think the checks should be in the cache and keep the responsibilities separated.

millenomi: Actually, let me see if I can retrofit it without putting the rest of the patch train too much behind.

millenomi:
I made a tactical mistake — `.forSession` means "for the duration of the current process (or login session)", _not_ for the lifetime of any particular `URLSession`. Naming is a hard problem, etc.

millenomi:
@swift-ci please test Linux

Foundation/URLSession/URLSession.swift
millenomi: Yeah, it's redundant now. Still, one thing at a time.

millenomi:
@swift-ci please test Linux

Foundation/URLSession/URLSessionTask.swift
millenomi: See comment below — I mistook what `.forSession` meant. We can just allow `forSession` storage.

millenomi: … the same way as we allow `persistent`.

millenomi:
This fixed https://bugs.swift.org/browse/SR-10375

PR #2423, opened 2019-07-19
[URLSession] Avoid calling completion callbacks/delegates twice.

drodriguez:
A callback/delegate call could have been called twice for the same task
if the task error happened during the task initialization and the task
was cancelled shortly after. The error was delivered first, but the
state of the task do not change to completed until after the completion
block has been called, which allows the cancellation to happen and
enqueue another failure callback.

The solution is checking for the task not being already completed before
the completion callbacks are invoked again. Basically the state machine
was allowing a transition from .completed to .completed, when it should
not allow it.

Includes a test that uses a bogus protocol to fail the task quickly and
being able to check for the double callback. The test will have fail
before this change is applied, and passes after it.

### DIFF
--- a/Foundation/URLSession/URLSessionTask.swift
+++ b/Foundation/URLSession/URLSessionTask.swift
@@ -720,6 +720,7 @@ extension _ProtocolClient : URLProtocolClient {
                 }
             }
             session.delegateQueue.addOperation {
+                guard task.state != .completed else { return }
                 delegate.urlSession(session, task: task, didCompleteWithError: nil)
                 task.state = .completed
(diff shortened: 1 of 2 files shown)


drodriguez:

drodriguez:

pushkarnk:

pushkarnk:
@drodriguez With reference to your test, the first cancellation was being delivered by your `FailFastProtocol.startLoading()` and the second by `session.invalidateAndCancelTasks()`. And we basically made sure the second one isn't delivered. Is that right? If yes, I agree with your reasoning.

PR #2424, opened 2019-07-20
[URL Session] Avoid Closing Sockets Before Cancelling Event Sources

gmittert:
When you create an event source for a handle in dispatch, dispatch
assumes it owns it, however, when curl finishes with the socket, it
closes it, which leaves dispatch with an invalid handle.

This sets the curl close callback to do nothing which allows us to close
the socket after cancelling the dispatch sources.


gmittert:
@swift-ci please test linux

compnerd:
CC: @millenomi

PR #2426, opened 2019-07-22
Add Conformance of 'DateInterval' to protocol 'Decodable'

damuellen:
Add Conformance of 'DateInterval' to protocol 'Decodable'  to match Darwin.

### DIFF
--- a/Foundation/DateInterval.swift
+++ b/Foundation/DateInterval.swift
@@ -218,3 +218,23 @@ extension DateInterval : _ObjectTypeBridgeable {
         return result!
     }
 }
+
+extension DateInterval : Codable {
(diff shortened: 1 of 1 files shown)


millenomi:
Can you add tests?

Good catch — I need to figure out a good way to API diff to finish catching these cases.

millenomi:
Actually. I would like to write those tests, I think (I want to extend fixtures to add Codable support).

millenomi:
@swift-ci please test Linux

xwu:
Hmm, in the Swift overlay, the `Codable` conformance is synthesized. Why does this diverge from that?

PR #2427, opened 2019-07-22
CoreFoundation: enforce 8-byte alignment on CFRuntimeBase

compnerd:
The atomic typed reference count is a 64-bit value.  We need to ensure
that any statically allocated structure is 8-byte aligned.  This comes
up in the case of CFString where CFConstantString types are emitted into
`__cfstring` on ELF, which can end up 4-byte aligned on ARMv7.  When
accessed at runtime, the offset becomes underaligned, resulting in a bus
error due to the unaligned access.

Although the attribution of `CFRuntimeBase` to 8-byte alignment should
enforce the alignment of `__CFString` due to the member alignment
requirement, just be extra cautious and mark both as requiring 8-byte
alignment.


compnerd:

PR #2428, opened 2019-07-22
[URLSession] Do not allow resuming cancelled/completed tasks.

drodriguez:
A cancelling or completed task might have been resumed (or suspended)
because the implementation didn't check for the state of the task to be
valid.

Includes a small test that reproduces the problem when this patch is not
applied and that doesn't fail when it is.


drodriguez:

PR #2430, opened 2019-07-24
Parity: Networking: URLAuthenticationChallenge

millenomi:
The class is mostly implemented, but there are several `NSUnimplemented()` that mark methods that, according to URLSession’s contract, are not meant to be invoked. Replace them with more verbose fatalErrors().


millenomi:
@swift-ci please test

millenomi:
Fixes https://bugs.swift.org/browse/SR-10372

PR #2433, opened 2019-07-26
[android] Fix symlink comparison in Android.

drodriguez:
The size reported by stat.st_size might not be the size of the contents
of the symlink, since it might include metadata. In Android at least the
size of a symlink seems to start at 18 bytes, even if the contents are
smaller.

The commit disables the st_size comparison against the contents
comparison for Android, and rewrites a little the function to use Swift
constructs like defer and guards.

This fixes `TestFileManager.test_contentsEqual` in Android.

PD: I don't know if the check is correct in many other file systems, but I don't think the test should be done at all. I only disabled it for Android because it is the only system where the symlinks doesn't seem to follow the pattern.

### DIFF
--- a/Foundation/FileManager+POSIX.swift
+++ b/Foundation/FileManager+POSIX.swift
@@ -853,24 +853,27 @@ extension FileManager {
         }
     }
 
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

spevans:
@swift-ci test linux

PR #2438, opened 2019-07-30
build: correct invocation of `cmake_policy`

compnerd:
The first parameter needs to be either `SET` or `GET`.  In this case, we
are trying to set the policy to NEW, so pass the correct parameter.


compnerd:

PR #2439, opened 2019-07-30
Make Host NSUnimplemented for Android

ephemer:
It is currently impossible to load Foundation on Android 5.0 because `getifaddrs` is not available there.

In the absence of `#if available(Android 24, *)` (which would be the ideal solution here), we need to avoid compiling any mention of `getifaddrs`, which is what this PR aims to achieve.

### DIFF
--- a/Foundation/Host.swift
+++ b/Foundation/Host.swift
@@ -140,6 +140,11 @@ open class Host: NSObject {
           pAdapter = pAdapter!.pointee.Next
         }
         _resolved = true
+#elseif os(Android)
+        // Android 5.0 doesn't contain an implementation of `getifaddrs`
+        // and therefore crashes when trying to load libFoundation.so.
+        // The existing code, below, may work on higher Android versions.
+        NSUnimplemented()
 #else
         var ifaddr: UnsafeMutablePointer<ifaddrs>? = nil
         if getifaddrs(&ifaddr) != 0 {


spevans:
If `ResolveType.current` cant be supported on Android its probably better to mark the `.current` value as unavailable using `@available`. Then the call to `_resolveCurrent()` can be removed on Android as well. Might also be worth making `_resolveCurrent()` `private` since it isn't used outside of this file.

spevans:
@swift-ci test

spevans:
cc @drodriguez

millenomi:
I hate to regress. Can we provide a dynamic shim in CF for getifaddrs for Android use only that errors out?

millenomi:
By that I mean: that errors out _if you're running on an old-enough Android OS_.

drodriguez:
It would be better to not disable it completely. I was aware of this usage, and there should be a way of fixing it. Lily proposal might be a way (even if the alternative for pre-5.0 is just something like `NSUnimplemented`.

In any case, if this is all the implementation, it also should include disabling the tests, or they will fail. `Host` functionality is not that interesting in any case (also not being available in iOS makes it a less used type than others in Foundation).

millenomi:
For context, I've provided some guidance on what the preferred way to deal with these situations [on the forums](https://forums.swift.org/t/lets-add-if-available-android/27439/5?u=millenomi).

ephemer:
Hi @drodriguez and @millenomi I agree it doesn't feel good to regress on available functionality. But I don't see a better way around this. As I mentioned [here](https://forums.swift.org/t/lets-add-if-available-android/27439/6?u=geordie_j) I don't think it's possible to make a shim for this on Android: I'm pretty sure we need a compile-time API Level check. Otherwise we have *no* Foundation at all on Android 5.0, which is a much worse regression than not having functionality.

I have no idea how often this functionality would be used on Android (have never used the API myself) but if it's not available on iOS I think it's reasonable for it not to be available on Android either.

We'd need to disable the Host tests on Android, yes.

And I agree that `NSUnsupported()` would be better than `NSUnimplemented()`. Even better than either of those would be an `@available` annotation so people don't think it'd be a good idea to try it for Android. Frrom what I understand `@available` isn't available for Android yet (and alone it also wouldn't suffice to fix the error we're having here - we'd still need to ensure that `getifaddrs` doesn't enter the symbol table for Android by using a conditional compilation).

ephemer:
@spevans sorry, that last message about NSUnimplemented was directed at you too. I appreciate your thoughts on this. I agree, it'd probably be ideal to mark `Host` as not `@available` on Android, but I don't think we have this feature in Swift yet (for Android). Without that, I'd rather only hit `NSUnsupported()` in the cases where it really is not supported. That way we'd make it clear to other s-c-f devs where the problem lies, and decrease the amount of unavailable API to a minimum. But I'm happy to go with your guidance on this if you have a strong opinion- I personally have never used the `Host` API- I just want to get Foundation working on Android 5.0 again.

ephemer:
@millenomi looks like I can take your suggestion after all using the approach from https://github.com/apple/swift-corelibs-foundation/pull/2228.

What do you think the best way of implementing the `NSUnsupported()` behaviour would be? Is something available in C along these lines?

spevans:
Could `getifaddrs` be implemented as a `weak` function that just returns failure? Then  shouldn't it be replaced by the version in the SDK if it exists otherwise the weak version would be used as a fallback?

millenomi:
I'm not quite sure what the support for that is on Android. @ephemer, thank you for going a different route.

ephemer:
@millenomi I'm just getting to making a PR now (I have very little idea what I'm doing though- as I mentioned I've never used this API before). I am just dynamically checking for `getifaddrs` and `freeifaddrs` which I should be able to figure out.

What I'm missing is a way to fatal error out from C in the `NSUnsupported` case. Do I just `NSLog` and `abort`, or is there a preferred way to do this in `CFFoundation`?

spevans:
Could you make the shim function return `errno = EOPNOTSUPP`if the functions don't exist and check this result in the swift code?

Molanda:
I found an Android implementation of `getifaddrs`/`freeifaddrs` that might be helpful...

https://github.com/morristech/android-ifaddrs

However, for myself, I have just used the `NSUnimplemented` solution, as Android 5.0 support is much more important to me than a working address lookup in `Host`.

shahmishal:

PR #2440, opened 2019-07-30
Use GetThreadId instead of Storing Handles Directly

gmittert:
CoreFoundation keeps a Dictionary of threads to their runloops. On
Windows, it previously used the threads' Handle as keys into the
dictionary. However, this doesn't work, as if you are attempting to look
up or set the current thread's RunLoop using GetCurrentThread.
GetCurrentThread returns a psuedo handle (FFFFFFFE) rather than a real
handle to a thread which causes it to index incorrectly into the
dictionary.

This was causing the creation of multiple runloops per thread and then
causing dispatch queues not to be pumped leaving blocks to never be run.

The fix here is to use the thread id instead of a handle which is unique
on the system until the thread terminates.


gmittert:
@swift-ci please test linux

gmittert:
cc @compnerd

gmittert:
@swift-ci please test linux

PR #2441, opened 2019-07-30
Handle Windows Errors in FileHandle Tests

gmittert:
- Errors on Windows are in the WindowsError domain
- When a file handle for a directory is attemptted to be read, instead
of `ERROR_ACCESS_DENIED` as triggered by windows, intercept it and
return `ERROR_DIRECTORY_NOT_SUPPORTED` to emulate POSIX's `EISDIR`
instead.


gmittert:
@swift-ci please test linux

PR #2443, opened 2019-07-31
Made methods in CFSocketStream consistent in style

pi1024e:
In CFSocketStream, there are two methods, _CFStreamErrorFromError and _CFErrorFromStreamError. Both methods are similar, except one has multiple return statements while the other only has one. I changed both methods to have multiple return statements while keeping the original structure the same.

### DIFF
--- a/CoreFoundation/Stream.subproj/CFSocketStream.c
+++ b/CoreFoundation/Stream.subproj/CFSocketStream.c
@@ -209,7 +209,7 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
     } else if (CFEqual(domain, kCFErrorDomainOSStatus)) {
         result.domain = kCFStreamErrorDomainMacOSStatus;
     } else if (CFEqual(domain, kCFErrorDomainMach)) {
-         result.domain = 11; // kCFStreamErrorDomainMach, but that symbol is in CFNetwork
+        result.domain = 11; // kCFStreamErrorDomainMach, but that symbol is in CFNetwork
     } else {
         result.domain = kCFStreamErrorDomainCustom;
     }
@@ -242,5 +242,4 @@ CF_PRIVATE CFErrorRef _CFErrorFromStreamError(CFAllocatorRef alloc, CFStreamErro
     CFRelease(value);
     CFRelease(dict);
     return result;
-    }
 }
--- a/CoreFoundation/Stream.subproj/CFSocketStream.c
+++ b/CoreFoundation/Stream.subproj/CFSocketStream.c
@@ -226,7 +226,6 @@ CF_PRIVATE CFErrorRef _CFErrorFromStreamError(CFAllocatorRef alloc, CFStreamErro
     canUpCall = (CFNetworkSupport._CFErrorCreateWithStreamError != NULL);
     __CFUnlock(&(CFNetworkSupport.lock));
 
-    
     if (canUpCall) {
         result = CFNETWORK_CALL(_CFErrorCreateWithStreamError, (alloc, streamError));
     } else {
--- a/CoreFoundation/Stream.subproj/CFSocketStream.c
+++ b/CoreFoundation/Stream.subproj/CFSocketStream.c
@@ -215,6 +215,7 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
     }
     result.error = CFErrorGetCode(error);
     return result;
+    
 }
 
 CF_PRIVATE CFErrorRef _CFErrorFromStreamError(CFAllocatorRef alloc, CFStreamError *streamError) {
--- a/CoreFoundation/Stream.subproj/CFSocketStream.c
+++ b/CoreFoundation/Stream.subproj/CFSocketStream.c
@@ -209,7 +209,7 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
     } else if (CFEqual(domain, kCFErrorDomainOSStatus)) {
         result.domain = kCFStreamErrorDomainMacOSStatus;
     } else if (CFEqual(domain, kCFErrorDomainMach)) {
-        result.domain = 11; // kCFStreamErrorDomainMach, but that symbol is in CFNetwork
+         result.domain = 11; // kCFStreamErrorDomainMach, but that symbol is in CFNetwork
     } else {
         result.domain = kCFStreamErrorDomainCustom;
     }
@@ -242,4 +242,5 @@ CF_PRIVATE CFErrorRef _CFErrorFromStreamError(CFAllocatorRef alloc, CFStreamErro
     CFRelease(value);
     CFRelease(dict);
     return result;
+    }
 }
--- a/CoreFoundation/Stream.subproj/CFSocketStream.c
+++ b/CoreFoundation/Stream.subproj/CFSocketStream.c
@@ -201,7 +201,7 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
     if (canUpCall) {
         return CFNETWORK_CALL(_CFStreamErrorFromCFError, (error));
     } 
-    
+
     CFStreamError result;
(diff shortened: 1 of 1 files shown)


xwu:
I don't think this is an improvement. The purpose of having `return result` at the end is to allow the reader to see at a glance that a result is returned regardless of which branches are taken.

Moreover, you have taken code that's bracketed by a lock and moved it outside, which does not appear to be correct.

pi1024e:
Fixed!

pi1024e:
@swift-ci please test

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -191,14 +191,14 @@ CF_EXPORT void CFStreamCreatePairWithPeerSocketSignature(CFAllocatorRef alloc, c
...
 CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
-    CFStreamError result;
     Boolean canUpCall;
     
     __CFLock(&(CFNetworkSupport.lock));
     if (!__CFBitIsSet(CFNetworkSupport.flags, kTriedToLoad)) initializeCFNetworkSupport();
     canUpCall = (CFNetworkSupport._CFStreamErrorFromCFError != NULL);
     __CFUnlock(&(CFNetworkSupport.lock));
 
+    CFStreamError result;
```
xwu: Please keep the declaration where it is.

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -218,21 +218,21 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
...
-    CFErrorRef result;
     Boolean canUpCall;
-    
+
     __CFLock(&(CFNetworkSupport.lock));
     if (!__CFBitIsSet(CFNetworkSupport.flags, kTriedToLoad)) initializeCFNetworkSupport();
     canUpCall = (CFNetworkSupport._CFErrorCreateWithStreamError != NULL);
     __CFUnlock(&(CFNetworkSupport.lock));
 
+    CFStreamError result;
```
xwu: Likewise here.

xwu:
You don't have permissions to trigger testing. For any changes you submit, you should test them locally on your own computer. See the readme for more details.

pi1024e:
I do test them. They pass. The exception was the one rebase where the extra parenthesis got in there and resulted in failed compilation

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -218,9 +218,9 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
 }
 
 CF_PRIVATE CFErrorRef _CFErrorFromStreamError(CFAllocatorRef alloc, CFStreamError *streamError) {
-    CFErrorRef result;
+    CFStreamError result;
```
xwu: This is not correct. I'd be surprised if it compiles. Please revert to the original.

pi1024e:
I disagree with your statement, because a return statement is a return statement. Developers know when a return statement is hit, nothing else is done; it is the end. At a glance, developers need to check the blocks to see if there will be any changes even in class where that does not happen; only in the else block is a variable changed as it goes along. In the other branches, result is set and only returned at the end.

As for the result variable, only declaring it when there needs to be a variable to be changed is more practical and I doubt any developer would know this. I would understand if the functions are over 30 lines long, but when a function is under 20 lines and fit entirely on screen, the entire function can be seen and developers know each branch has an end.

Those are my two cents on this. You can disagree with them.

CoreFoundation/Stream.subproj/CFSocketStream.c
pi1024e: Another case and point on my newest comment.

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -220,19 +220,20 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
...
     CFErrorRef result;
     Boolean canUpCall;
-    
+
     __CFLock(&(CFNetworkSupport.lock));
     if (!__CFBitIsSet(CFNetworkSupport.flags, kTriedToLoad)) initializeCFNetworkSupport();
     canUpCall = (CFNetworkSupport._CFErrorCreateWithStreamError != NULL);
     __CFUnlock(&(CFNetworkSupport.lock));
 
+
```
xwu: Please revert this whitespace change.

pi1024e:
Perfect!

xwu:
Yes, early return can be an improvement, but for such consistency the code would have had to be refactored differently (and more extensively).

In the absence of a compelling reason, we try not to make stylistic changes to existing code, because that can only introduce additional bugs. We'll see if others find this change worthwhile as it is for the sake of consistency, but honestly, I think the existing code is fine.

pi1024e:
I agree with your proposal. Maybe others will find it worthwhile, maybe other's won't.

xwu:
Did you change the PR again? If you want to use early return, then this refactoring will have to be more extensive. Specifically, there’s no point in using `else` if the other branch always returns. That would improve the clarity to the reader.

But this is a more extensive change, and as I stated, in general, we don’t make stylistic changes like this to long-established code just for the sake of style, because it can introduce new bugs.

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -200,47 +199,49 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
...
+    } 
+
+    CFStreamError result;
+    CFStringRef domain = CFErrorGetDomain(error); 
+    if (CFEqual(domain, kCFErrorDomainPOSIX)) {
+        result.domain = kCFStreamErrorDomainPOSIX;
+    } else if (CFEqual(domain, kCFErrorDomainOSStatus)) {
+        result.domain = kCFStreamErrorDomainMacOSStatus;
+    } else if (CFEqual(domain, kCFErrorDomainMach)) {
+         result.domain = 11; // kCFStreamErrorDomainMach, but that symbol is in CFNetwork
```
xwu: Indenting inconsistency here.

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -200,47 +199,49 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
...
-            result.domain = 11; // kCFStreamErrorDomainMach, but that symbol is in CFNetwork
-        } else {
-            result.domain = kCFStreamErrorDomainCustom;
-        }
-        result.error = CFErrorGetCode(error);
+        result.domain = kCFStreamErrorDomainCustom;
     }
+    result.error = CFErrorGetCode(error);
     return result;
+
```
xwu:
```suggestion
```

CoreFoundation/Stream.subproj/CFSocketStream.c
```diff
@@ -200,47 +199,49 @@ CF_PRIVATE CFStreamError _CFStreamErrorFromError(CFErrorRef error) {
...
+    }
+    
+    CFStringRef key = CFSTR("CFStreamErrorDomainKey");
+    CFNumberRef value = CFNumberCreate(alloc, kCFNumberCFIndexType, &streamError->domain);
+    CFDictionaryRef dict = CFDictionaryCreate(alloc, (const void **)(&key), (const void **)(&value), 1, &kCFTypeDictionaryKeyCallBacks, &kCFTypeDictionaryValueCallBacks);
+    CFErrorRef result = CFErrorCreate(alloc, CFSTR("BogusCFStreamErrorCompatibilityDomain"), streamError->error, dict);
+    CFRelease(value);
+    CFRelease(dict);
     return result;
+    }
```
xwu: Syntax error.

pi1024e:
Fixed!

pi1024e:
Closing this for now

PR #2447, opened 2019-07-31
Parity: URLProtocol

millenomi:
The only issue here was using NSUnimplemented() (“we still need to implement this”) instead of NSRequiresConcreteImplementation() (“this is an abstract method”). Fix it. Fixes https://bugs.swift.org/browse/SR-10377


millenomi:

PR #2448, opened 2019-07-31
Parity: URLProtectionSpace

millenomi:
Remove the last NSUnimplemented() calls from this class. Since we can’t guarantee Darwin interoperability with coding, remove NSSecureCoding conformance from this class in swift-corelibs-foundation only. Note that certain authentication methods require support that’s not available in s-c-f; diagnose relevant symbols when used. Fixes https://bugs.swift.org/browse/SR-10376


millenomi:

PR #2450, opened 2019-08-01
Consistently add #error when compiling on unsupported architectures.

spevans:
- #if arch() guard blocks are used for per architecture code paths, so
  for unsupported architectures add an #else/#error message to ensure
  the correct path is added.

- This fixes some initializers that may have otherwise been empty.

- Ignore blocks with just one architecture as this is usually a specific
  targeted fix.

I think the message is correct for all cases but can be changed if necessary.


spevans:
@swift-ci test

PR #2316, opened 2019-06-04
SR-10776: Fix runtime crash of calling XMLDocument's `var name: String?`

YOCKOW:
Runtime crash occurs when `XMLDocument`'s computed property `var name: String?` is called. This PR fixes the issue.

Resolves [SR-10776](https://bugs.swift.org/browse/SR-10776).

~~Note: This PR would conflict with [PR#2287](https://github.com/apple/swift-corelibs-foundation/pull/2287) and [PR#2313](https://github.com/apple/swift-corelibs-foundation/pull/2313). That's why it is a draft at the moment. This branch will be rebased after they are merged.~~

**EDITED**
#2287 has been merged. This PR seems not to conflict with #2313 after checking again. Sorry.

PR #2425, opened 2019-07-20
Refactored code to remove unnecessary checks

pi1024e:
When creating an immutable funnel, the code checks the value of "hasLengthByte" twice if it ends up being true. Checking for it once is all that is needed, as another branch can check for the other boolean statement in the if statement previously on line 1483. Differentiating the branches allows less redundant checks and does not introduce any side effects.

PR #2453, opened 2019-08-03
Fix crash that occurs when attempting to retrieve the name property of an XMLDocument by making it return nil.

pi1024e:
Runtime crash occurs when one attempts to retrieve XMLDocument's name property. This was due to the lack of checking if Documents were allowed or not to have its name property retrieved.

This came from SR-10776, and a PR has been written, but no tests were run and the PR has not been attended to: #2316.

Unlike the original PR, this PR does not have any changes on XMLDocument in Swift, as there is no need to check for document there, but rather in the _getQName function.

### DIFF
--- a/CoreFoundation/Parsing.subproj/CFXMLInterface.c
+++ b/CoreFoundation/Parsing.subproj/CFXMLInterface.c
@@ -470,7 +470,7 @@ void _CFXMLNodeSetPrivateData(_CFXMLNodePtr node, void* data) {
     if (!node) {
         return;
     }
-	((xmlNodePtr)node)->_private = data;
+    ((xmlNodePtr)node)->_private = data;
 }
 
 void* _Nullable  _CFXMLNodeGetPrivateData(_CFXMLNodePtr node) {
--- a/CoreFoundation/Parsing.subproj/CFXMLInterface.c
+++ b/CoreFoundation/Parsing.subproj/CFXMLInterface.c
@@ -467,9 +467,10 @@ void _CFXMLNodeSetURI(_CFXMLNodePtr node, const unsigned char* URI) {
 }
 
 void _CFXMLNodeSetPrivateData(_CFXMLNodePtr node, void* data) {
-    if (node) {
-        ((xmlNodePtr)node)->_private = data;
+    if (!node) {
+        return;
     }
+	((xmlNodePtr)node)->_private = data;
 }
 
 void* _Nullable  _CFXMLNodeGetPrivateData(_CFXMLNodePtr node) {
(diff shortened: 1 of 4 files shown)


pi1024e:
@swift-ci please test

YOCKOW:
My opinions about #2316 and this PR are:

- The latest tests of #2316 *did* run, but Jenkins just didn't report the results of them.
- #2316 is also the first step to resolve [SR-10764](https://bugs.swift.org/browse/SR-10764) not only SR-10776.
- I thought `name`'s behavior should be defined in SwiftFoudation not in CoreFoundation whose CFXML* are just wrappers of libxml2’s struct.
- This PR has too many commits to resolve the issue.
- Why don’t you review #2316 before creating a new PR.

Thank you.

pi1024e:
The reason why it should be in corefoundation is because that is where, according to the crash report you submitted and the crash that happened on my machine, was. The crash happened in that function, where document was allowed to run unchecked.

pi1024e:
I would like to test these changes on here (they passed locally), before I squash the commits just to be sure these bugs are fixed and no new ones are introduced.

YOCKOW:
> The reason why it should be in corefoundation is because that is where, according to the crash report you submitted and the crash that happened on my machine, was. The crash happened in that function, where document was allowed to run unchecked.

Well, sorry, but I talked about the setter.
Moreover, the setter calls only `_CFXMLNodeSetName`. I think your PR contains unrelated changes.


> I would like to test these changes on here (they passed locally), before I squash the commits just to be sure these bugs are fixed and no new ones are introduced.

You can read @xwu’s comments in your own past PR about that:
- https://github.com/apple/swift-corelibs-foundation/pull/2425#issuecomment-516577735
- https://github.com/apple/swift-corelibs-foundation/pull/2443#issuecomment-516985934

xwu:
@pi1024e:

Again, as a matter of etiquette, it's not cool to take a bug that someone else is clearly still working on with an active pull request; if you're interested in helping to resolve such an issue or disagree with how they've approached the topic, you can review their work and offer to help that person.

@YOCKOW has also pointed out that this PR also contains unrelated changes.

I'm not sure of your level of experience with these things, but these sort of things apply generally to most if not all projects.

pi1024e:
I apologize, as many of these bugs appear to be inactive. I am deeply sorry about this.

millenomi:
As a separate note, I'm less against mixed-content patches so long that all the changes in a patch are either required for the main part of it to function, or strongly needed to improve its ergonomics.

In general, I do not take cleanup patches to Core Foundation code in the summer because it complicates the process of merging that I do every year in the early fall.

pi1024e:
Fixed it so it is not unnecessary changes! @millenomi

millenomi:
I took YOCKOW's patch, per discussion on the patch I closed — if we're targeting 5.0 still, we should cherry-pick that one.

pi1024e:
Got it!

PR #2455, opened 2019-08-03
build: sink sched_getaffinity check into CoreFoundation

compnerd:


compnerd:
@swift-ci please test

PR #2456, opened 2019-08-04
build: remove option to disable GCD support

compnerd:
GCD support is available on all targets, make it a hard dependency.
This was originally for ease of porting, but is no longer needed.


compnerd:
@swift-ci please test

PR #2459, opened 2019-08-04
Foundation: remove unnecessary `-DDEPLOYMENT_ENABLE_LIBDISPATCH`

compnerd:
libdispatch is required by Foundation.  Remove the single case where we
depended on the flag to be set when building Foundation.


compnerd:
@swift-ci please test

compnerd:
CC: @millenomi

PR #2460, opened 2019-08-05
Place read-only data symbols in the .rodata section.

mundaym:
This data was previously put into the default section, .text, which
is generally used for executable data.

This fixes a gold warning on IBM Z: "S/390 code fill of odd length
requested". This warning was triggered because executable
instructions on IBM Z must be 2-byte aligned and the data in one of
these sections is 1-byte aligned.

### DIFF
--- a/CoreFoundation/String.subproj/CFCharacterSetData.S
+++ b/CoreFoundation/String.subproj/CFCharacterSetData.S
@@ -9,6 +9,10 @@
 
 #include <CoreFoundation/CFAsmMacros.h>
 
+#if defined(__ELF__)
+.section .rodata
+#endif
+
     .global _C_LABEL(__CFCharacterSetBitmapData)
 _C_LABEL(__CFCharacterSetBitmapData):
     .incbin CF_CHARACTERSET_BITMAP
--- a/CoreFoundation/String.subproj/CFUniCharPropertyDatabase.S
+++ b/CoreFoundation/String.subproj/CFUniCharPropertyDatabase.S
@@ -9,6 +9,10 @@
 
 #include <CoreFoundation/CFAsmMacros.h>
 
+#if defined(__ELF__)
+.section .rodata
+#endif
+
     .global _C_LABEL(__CFUniCharPropertyDatabase)
 _C_LABEL(__CFUniCharPropertyDatabase):
     .incbin CF_CHARACTERSET_UNICHAR_DB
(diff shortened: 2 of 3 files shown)


mundaym:
@swift-ci Please test.

millenomi:
What is the effect on other platforms here? Is there any platform that would balk at this, and should we conditionalize it?

mundaym:
> What is the effect on other platforms here? Is there any platform that would balk at this, and should we conditionalize it?

I'm not totally sure. It is definitely fine when targeting ELF objects. Windows might need to be `.rdata` and I think `.rodata` will do the right thing on Darwin but I'm not certain since MACH doesn't seem to have that section by default. Given the uncertainty I've modified the PR to only target ELF objects.

We should probably do the other object file formats too though at some point. Dumping the data into the `.text` section isn't ideal.

mundaym:
@swift-ci Please test.

millenomi:
@swift-ci please test

CoreFoundation/Base.subproj/CFAsmMacros.h
```diff
@@ -10,10 +10,22 @@
...
+#include <TargetConditionals.h>
+#endif
+
 #define CONCAT(a,b) a##b
 #define CONCAT_EXPANDED(a,b) CONCAT(a,b)
 #define _C_LABEL(name) CONCAT_EXPANDED(__USER_LABEL_PREFIX__,name)
 
+// Define section name for read-only data section, if available.
+// TODO: equivalent section names for other operating systems.
+#if (TARGET_OS_LINUX || TARGET_OS_ANDROID || TARGET_OS_BSD) && defined(__ELF__)
```
compnerd: This feels wrong.  This section is really a property of the object file, not the OS.  I really prefer that we just conditionalise this on `__ELF__`, and sink this into the data files.

compnerd:
@millenomi since this is a property of the object file, no platform would balk at this.  The flags need to be setup appropriately for the target object file format, `__ELF__` should ensure that the correct section is used for ELF targets.  MachO and PE/COFF can be addressed in a follow up IMO.

CoreFoundation/Base.subproj/CFAsmMacros.h
mundaym: Done, I agree that just `__ELF__` is probably enough. Thanks for taking a look.

mundaym:
@swift-ci Please test.

millenomi:
@swift-ci please test

compnerd:

PR #2462, opened 2019-08-07
[gardening] Migrate from unwrapped() to XCTUnwrap()

drodriguez:
Instead of using an extension on Optional only available in Foundation,
we should start using the recently introduced XCTUnwrap available in
XCTest to perform the same job. The XCTest API would probably be more
known in other code bases, so people will be more willing to use it and
learn it.

The commit removes all the usages of unwrapped and replaces them with
XCTUnwrap. Additionally it marks unwrapped() as deprecated, so people
receive a clear message about the disappearance with a helpful hint of
what to use instead.

### DIFF
--- a/Docs/Testing.md
+++ b/Docs/Testing.md
@@ -7,7 +7,7 @@ swift-corelibs-foundation uses XCTest for its own test suite. This document expl
 ### In brief
 
 * Tests should fail rather than crashing; swift-corelibs-xctest does not implement any crash recovery
-* You should avoid forced optional unwrapping (e.g.: `aValue!`). Use `try aValue.unwrapped()` instead
+* You should avoid forced optional unwrapping (e.g.: `aValue!`). Use `try XCTUnwrap(aValue)` instead
 * You can test code that is expected to crash; you must mark the whole body of the test method with `assertCrashes(within:)`
 * If a test or a portion of a test is giving the build trouble, use `testExpectedToFail` and write a bug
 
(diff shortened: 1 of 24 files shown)


drodriguez:
GenerateTestFixtures has its own implementation and I haven't touch it with this commit (ref: the file as it is with this commit https://github.com/apple/swift-corelibs-foundation/blob/d3df79dc5f2624be38dd7086a2291a51452111a9/Tools/GenerateTestFixtures/GenerateTestFixtures/Utilities.swift)

drodriguez:
All right. I understand what you mean. It can be modified.

I think it flew under my radar because CMake doesn't seem to build that target.

drodriguez:
@swift-ci please test Linux Platform

millenomi:
It is intentional that CMake doesn't build that target; that target is meant to be built on macOS with the shipping version of Xcode, for now.

millenomi:
More to the point, right now we're building that target for Mojave (to generate the macOS-10.14 test fixtures in TestFoundation/Fixtures).

millenomi:
I hadn't thought this through — _of course_ we can build GTF with XCTUnwrap, since you're providing one. Whoops.

millenomi:

millenomi:

millenomi:
I'll need to catch up with new tests.

drodriguez:
Removed the `unwrapped()` calls introduced in `TestSocketPort`.

@swift-ci please test

drodriguez:
@swift-ci please test

drodriguez:

PR #2463, opened 2019-08-08
Parity: Port, PortMessage and SocketPort

millenomi:
⚠️ Needs tests.

 - Implement `Port` as a top-level abstract class, like Darwin’s.
 - Implement `PortMessage`.
 - Implement `SocketPort`, our only cross-platform port. Note that Distributed Objects is not available in s-c-f, and thus `NSConnection`-related functionality is not available nor implemented.

A few differences with Darwin:

 - `PortCoder` is not available. Classes that ‘support coding only through a `NSPortCoder`’ do not conform to `NSCoding` in swift-corelibs-foundation.

 - `MessagePort` is not available in s-c-f. It is noted as a to-avoid API in the documentation and relies on Mach ports, which are not available everywhere. The default implementation of `Port`’s initializer would return a `MessagePort` as a factory initializer on Darwin if invoked directly; this is not the case here, and we diagnose that use.

 - Similarly, `NSMachPort` is also unavailable.

 - `SocketPort` uses a factory initializer on Darwin to return the same instance for the same remote port signature (uniquing). This is not the case in Swift. The return value of `==` and `.isEqual(_:)` will match Darwin’s behavior and return true for uniqued `SocketPort`s, but certain swift-corelibs-foundation uniqued `SocketPort`s will be different instances as reported by the `===` and `!==` operators.

### DIFF
--- a/Foundation/Port.swift
+++ b/Foundation/Port.swift
@@ -7,56 +7,53 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+import CoreFoundation
+
+// MARK: Port and related types
 
 public typealias SocketNativeHandle = Int32
 
 extension Port {
-    public static let didBecomeInvalidNotification  = NSNotification.Name(rawValue:  "NSPortDidBecomeInvalidNotification")
+    public static let didBecomeInvalidNotification  = NSNotification.Name(rawValue: "NSPortDidBecomeInvalidNotification")
 }
 
-open class Port : NSObject, NSCopying, NSCoding {
-    
-    
+open class Port : NSObject, NSCopying {
+    @available(*, deprecated, message: "On Darwin, you can invoke Port() directly to produce a MessagePort. Since MessagePort's functionality is not available in swift-corelibs-foundation, you should not invoke this initializer directly. Subclasses of Port can delegate to this initializer safely.")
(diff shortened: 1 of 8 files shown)


millenomi:
cc @phausler

millenomi:
Fixes https://bugs.swift.org/browse/SR-10368

millenomi:
@swift-ci please test

phausler:
Overall this looks pretty good,  there are a few minor tweaks for performance that can probably be done in the handling of Data but everything here looks salient.

The one consideration that might be worthwhile to look into is the serialization of in6_addr on different platforms since that structure has different members and layouts on different OSes.

Foundation/Port.swift
```diff
@@ -7,136 +7,793 @@
...
+    }
+    
+    open override func invalidate() {
+        guard var core = core else { return }
+        self.core = nil
+        
+        var signatureToRemove: Signature?
+        
+        if core.isUniqued {
+            if isKnownUniquelyReferenced(&core) {
```
phausler: 🤯

millenomi:
@swift-ci please test

millenomi:
I will have a separate patch with more tests.

millenomi:
@swift-ci please test

PR #2464, opened 2019-08-08
[cmake] Create a GenerateTestFixtures targets in CMake.

drodriguez:
The target reproduces the basic structure of the Xcode project and
allows other platforms to build easily and test.

The target is not installed intentionally. It is supposed to be a
development tool.

(This would have avoided a problem that was almost introduced in #2462)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,6 +426,28 @@ add_swift_executable(plutil
                        Foundation
                        CoreFoundation)
 
+add_swift_executable(GenerateTestFixtures
+                     SOURCES
+                       Tools/GenerateTestFixtures/GenerateTestFixtures/main.swift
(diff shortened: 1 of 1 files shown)


millenomi:
I'm pretty sure it doesn't build on s-c-f. It's meant to be built with Darwin Foundation only, for now.

millenomi:
@swift-ci please test

millenomi:
Let's see if I'm wrong.

drodriguez:
From what I found in the documentation, the tool also handles building in Linux, but it is indeed more useful in Darwin. Feel free to merge or not. It would be useful to avoid unintentional breakage, at least.

millenomi:
I wrote Linux support in it, but I have never tested it.

millenomi:
Eh, no time like the present.

PR #2468, opened 2019-08-14
build: workaround hook for zlib dependency

compnerd:
Until CMake 3.15, dependency tracking is not properly possible due to custom
commands for linking.  Provide an escape hatch in the form is a variable that
can be setup to force the linking of zlib.  We do not search for zlib, simply
allow the user specify a magic variable to permit them to inject the linking.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -327,6 +327,7 @@ add_swift_library(FoundationNetworking
                     ${Foundation_INTERFACE_LIBRARIES}
                     ${CFURLSessionInterface_LIBRARIES}
                     ${CURL_LIBRARIES}
+                    ${ZLIB_LIBRARY}
                     ${Foundation_RPATH}
                     ${WORKAROUND_SR9138}
                     ${WORKAROUND_SR9995}


compnerd:
@swift-ci please smoke test

compnerd:
@swift-ci please test

millenomi:
Where is this set?

compnerd:
@millenomi - that is the magic of this horrible workaround.  It never is!  You see, this is effectively a hook for the user to specify the path if necessary (or desired).  We never explicitly setup anything so the build remains identical to what it is currently (because it is unset, it will evaluate to the empty string and become a no-op).

millenomi:
Okay, I get it isn't. I'd like to get on the record why it is there.

millenomi:
i.e. in what case would one set it?

compnerd:
Oh, an example is windows: https://dev.azure.com/compnerd/windows-swift/_build/results?buildId=6266&view=logs&j=b9e62f99-1a98-5ed7-01d2-f4794231ed79&t=c9058fa6-ea7a-556c-1928-b43ca1a9cfca&l=914

The build failure is due to the lack of linking against zlib which is needed curl.

PR #2474, opened 2019-08-15
Parity: XMLDocument: NSCoding, XSLT

millenomi:
We’re out of time to add libxslt as a dependency. Note that XSLT is not available at compile time, for now, and remove the undue reference to coding.

Fixes https://bugs.swift.org/browse/SR-10420, https://bugs.swift.org/browse/SR-10420


millenomi:
@swift-ci please test

PR #2476, opened 2019-08-17
Use Array.isEmpty instead of checking account property

x4snowman:
According to the Xcode documentation, we should use isEmpty property to check whether your collection is empty.
[https://developer.apple.com/documentation/swift/array/1688398-isempty](https://developer.apple.com/documentation/swift/array/1688398-isempty)
> When you need to check whether your collection is empty, use the isEmpty property instead of checking that the count property is equal to zero. For collections that don’t conform to RandomAccessCollection, accessing the count property iterates through the elements of the collection.


ikesyo:
@swift-ci test and merge

PR #2461, opened 2019-08-05
Revert "Parity: Networking: URLUploadTask and URLDownloadTask"

millenomi:
Reverts apple/swift-corelibs-foundation#2293 — temporarily un-fails the race seen in https://ci.swift.org//job/oss-swift-incremental-RA-linux-ubuntu-14_04/7837/consoleFull#1582003869ba62d58e-7248-467b-91e0-c7508d5cf947

PR #2477, opened 2019-08-19
Parity: Networking: URLUploadTask and URLDownloadTask

millenomi:
- Implement downloadTask(withResumeData…). We do not currently support resume data, so act as Darwin does when invalid data is fed to these methods.

 - Implement uploadTask(withStreamedRequest:).

 - Some cleanup, some testing.

This restores the patch after the revert at https://github.com/apple/swift-corelibs-foundation/pull/2461, after the test issues were revealed as unrelated.


millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

PR #2478, opened 2019-08-20
[SR-10764]  XMLNode: Make the behavior of `name` the same as Darwin.

YOCKOW:
As described in [SR-10764](https://bugs.swift.org/browse/SR-10764), some behavior of `var name: String? { get set }` of `XMLNode` differs between DarwinFoundation and SwiftFoundation.
This PR resolves the issue.


Foundation/XMLDTD.swift
```diff
@@ -51,7 +51,12 @@ open class XMLDTD : XMLNode {
             //TODO: throw a generic error?
             fatalError("parsing dtd from data failed")
         }
-
+        
+        // _CFXMLParseDTDFromData assigns "none" to DTD's name when there's no name for DTD.
+        if _CFXMLNodeNameEqual(node, "none") {
+            _CFXMLNodeForceSetName(node, nil)
+        }
```
YOCKOW: This is a simple way but might not the best way.
If we don't approve it, we have to implement our own DTD parser.

spevans:
@swift-ci test

Foundation/XMLDTD.swift
millenomi: Nah, it's okay.

millenomi:
@swift-ci please test

millenomi:
The test was unnecessary, whoops. I'll merge when done.

PR #2480, opened 2019-08-20
Parity: XMLDTD.init()

millenomi:
Implement it to create an empty `.DTDKind` node, as it does on Darwin.


millenomi:
@swift-ci please test

PR #2481, opened 2019-08-20
SR-11323: NSCalendar functions are not thread safe

spevans:
- CFCalendarGetTimeRangeOfUnit() modifies the calendar->_cal object
  so make a copy of the NSCalendar before calling it.

### DIFF
--- a/CoreFoundation/Locale.subproj/CFCalendar.c
+++ b/CoreFoundation/Locale.subproj/CFCalendar.c
@@ -639,6 +639,7 @@ CF_CROSS_PLATFORM_EXPORT bool _CFCalendarInitWithIdentifier(CFCalendarRef calend
     calendar->_userSet_firstWeekday = false;
     calendar->_userSet_minDaysInFirstWeek = false;
     calendar->_userSet_gregorianStart = false;
+    CF_LOCK_INIT_FOR_STRUCTS(calendar->_lock);
     ICU_LOG("                // _CFCalendarInitWithIdentifier exit true\n");
     return true;
 }
--- a/CoreFoundation/Locale.subproj/CFCalendar_Internal.h
+++ b/CoreFoundation/Locale.subproj/CFCalendar_Internal.h
@@ -15,6 +15,7 @@
 #include <CoreFoundation/CFPriv.h>
 #include <CoreFoundation/CFTimeZone.h>
 #include <CoreFoundation/CFString.h>
+#include <CoreFoundation/CFLocking.h>
 
 #include "CFDateComponents.h"
 #include "CFDateInterval.h"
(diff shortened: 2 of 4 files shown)


spevans:
Im not sure if this is the correct fix but if it is I think some of the other `NSCalendar` methods should be audited for thread safety.

spevans:
@swift-ci test

millenomi:
I'd rather have a lock here than allocate a whole new calendar.

spevans:
@swift-ci test

spevans:
I added a `CFLock_t` into `__CFCalendar` and `NSCalendar` but does it need to be destroyed in `__CFCalendarDeallocate`?

 `__CFLocale` also has a `CFLock_t` and I couldn't see a corresponding lock deinit in `__CFLocaleDeallocate`.

Catfish-Man:
The way we dealt with this sort of thing in objc Foundation is to wrap NSCalendar instanced in a subclass that does copy on write and has a lock. The rest of the implementation is not at all expected to be threadsafe.

spevans:
@Catfish-Man the original solution(see below)  did just make a copy of the calendar just before calling `CFCalendarGetTimeRangeOfUnit()` but @millenomi wanted a lock instead because the copy was too heavy. 

```
@@ -462,7 +462,8 @@ open class NSCalendar : NSObject, NSCopying, NSSecureCoding {
         var ti: CFTimeInterval = 0.0
         let res: Bool = withUnsafeMutablePointer(to: &start) { startp in
             withUnsafeMutablePointer(to: &ti) { tip in
-                return CFCalendarGetTimeRangeOfUnit(_cfObject, unit._cfValue, date.timeIntervalSinceReferenceDate, startp, tip)
+                let tmpCopy = self.copy() as! NSCalendar
+                return CFCalendarGetTimeRangeOfUnit(tmpCopy._cfObject, unit._cfValue, date.timeIntervalSinceReferenceDate, startp, tip)
             }
         }
 ```

Catfish-Man:
Yeah, copy on write (if it were working. IIRC it’s not currently, even in objc foundation) would make the copy cheap IFF nobody else mutated that calendar. It’s a tricky situation for sure; “real” calendar copies are horrifyingly expensive.

spevans:
I wonder if its worth using a `trylock()` and create a tmp copy if the lock fails. There looks to be some other `CFCalendar...` calls that need to be made thread-safe . I guest it depends on the relative cost of the individual `CFCalendar..` call versus the copy.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
I got it working but had to use a recrusive mutex due to the copies that can be made in `_setComp`. The biggest problem is actually exposing the `_CFRecursiveMutex` typedef back to Swift.

On alternative is do we still care about bridging the `NSCalendar` to `CFCalendar`? Then `NSCalendar` could just contains a `CFCalendar` struct and an `NSLock`.  If CoreFoundation is not exposed anymore Im not sure the toll free bridging brings any benefits.

PR #2484, opened 2019-08-22
XFAIL test_invalidResumeDataForDownloadTask

aschwaighofer:
It randomly fails with Fatal error: Trying to access a behaviour for a task that in not in the registry

SR-11353
rdar://54555179


aschwaighofer:
@swift-ci test

PR #2488, opened 2019-08-24
Fix Operation.swift Operation addDependency bug

x4snowman:
In this commit, I only fixed the `addDependency` bug.  However, there is a critical bug [https://bugs.swift.org/browse/SR-11333](https://bugs.swift.org/browse/SR-11333), exists from swift5.0 (I didn't check swift version < 5.0).

I created a test case for OperationQueue and Operation which testing the dependency.

### DIFF
--- a/Foundation/Operation.swift
+++ b/Foundation/Operation.swift
@@ -407,7 +407,7 @@ open class Operation : NSObject {
             withExtendedLifetime(op) {
                 var up: Operation?
                 _lock()
-                if __dependencies.first(where: { $0 === op }) != nil {
+                if __dependencies.first(where: { $0 === op }) == nil {
                     __dependencies.append(op)
                     up = op
                 }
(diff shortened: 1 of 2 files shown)


x4snowman:
I really hope some guys could fix the Operation & OperationQueue dependency bug. Everything works fine on the Darwin platform. As the **Operation.swift** changes so much between swift-5.0 and the current master branch. I guess the Foundation team may have a new design for Operation&OperationQueue and everything is in progress.

x4snowman:
@millenomi The test case failed as I only fixed the `addDependency`. For the dependency login in `OperationQueue` there is still a bug here which I haven't fixed. You can take a look at [SR-11333](https://bugs.swift.org/browse/SR-11333).

x4snowman:
@swift-ci please test

x4snowman:
@millenomi I updated the test case to only handle the `addDependency` method. According to [SR-11333] (https://bugs.swift.org/browse/SR-11333), is that properly I could work and fix that? However, I am not sure if the foundation team there has already got a plan and overall design for Operation & OperationQueue. Looking forward to get how do you think this.

PR #2498, opened 2019-08-31
build: stage libdispatch movement

compnerd:
In order to support libdispatch tests on Windows, the libraries must be
colocated with the tests.  This requires adjusting the build tree
layout.  Account for that layout structure.


compnerd:
@swift-ci please test

PR #2504, opened 2019-09-09
[Windows] Fix test_reachable

gmittert:
Windows doesn't have `/data` or `/usr`. Since Windows does not guarantee
a drive letter actually existing, defer to NSHomeDirectory to give us a
directory which should always exist.


gmittert:
@swift-ci please test

gmittert:
cc @compnerd

PR #2505, opened 2019-09-09
[Windows] O_EXCL should override O_TRUNC

gmittert:
Setting `O_CREAT` and `O_EXCL` together should always force the creation
of a new file and fail otherwise, regardless of if O_TRUNC is set.

In addition, `O_EXCL` should only take effect if `O_CREAT` is set,
otherwise, the result is undefined behavior per the specification (in
our case, setting to OPEN_EXISTING).

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -906,14 +906,15 @@ CF_EXPORT int _NS_open(const char *name, int oflag, int pmode) {
 
     DWORD dwCreationDisposition;
     switch (oflag & (O_CREAT | O_EXCL | O_TRUNC)) {
+      case O_CREAT | O_EXCL | O_TRUNC:
+      case O_CREAT | O_EXCL:
+        dwCreationDisposition = CREATE_NEW;
+        break;
       case O_CREAT | O_TRUNC:
         dwCreationDisposition = CREATE_ALWAYS;
         break;
       case O_CREAT:
         dwCreationDisposition = OPEN_ALWAYS;
-        break;
-      case O_EXCL:
-        dwCreationDisposition = CREATE_NEW;
         break;
       case O_TRUNC:
         dwCreationDisposition = TRUNCATE_EXISTING;


gmittert:
@swift-ci please test linux

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -906,15 +906,16 @@ CF_EXPORT int _NS_open(const char *name, int oflag, int pmode) {
 
     DWORD dwCreationDisposition;
     switch (oflag & (O_CREAT | O_EXCL | O_TRUNC)) {
+      case O_CREAT | O_EXCL | O_TRUNC:
+      case O_CREAT | O_EXCL:
+	dwCreationDisposition = CREATE_NEW;
```
compnerd: Indentation is off.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -906,15 +906,16 @@ CF_EXPORT int _NS_open(const char *name, int oflag, int pmode) {
 
     DWORD dwCreationDisposition;
     switch (oflag & (O_CREAT | O_EXCL | O_TRUNC)) {
+      case O_CREAT | O_EXCL | O_TRUNC:
```
compnerd: I think that `O_TRUNC` is interesting, the behaviour between these two is different I believe.

gmittert: I don't think it is different, referencing https://pubs.opengroup.org/onlinepubs/9699919799/:
```
O_EXCL
If O_CREAT and O_EXCL are set, open() shall fail if the file exists. The check for the
existence of the file and the creation of the file if it does not exist shall be atomic
with respect to other threads executing open() naming the same filename in the
same directory with O_EXCL and O_CREAT set. If O_EXCL and O_CREAT are set, and
path names a symbolic link, open() shall fail and set errno to [EEXIST], regardless of
the contents of the symbolic link. If O_EXCL is set and O_CREAT is not set, the result
is undefined.
```
and 
```
O_TRUNC
If the file exists and is a regular file, and the file is successfully opened O_RDWR or
O_WRONLY, its length shall be truncated to 0, and the mode and owner shall be
unchanged. It shall have no effect on FIFO special files or terminal device files.
Its effect on other file types is implementation-defined. The result of using
O_TRUNC without either O_RDWR or O_WRONLY is undefined.
```
In the case of `O_CREAT | O_EXCL | O_TRUNC`, if a file exists, then O_EXCL causes open to fail regardless of `O_TRUNC`. If the file doesn't exist, then O_TRUNC doesn't do anything and is the same as `O_CREAT | O_EXCL`.

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test

PR #2506, opened 2019-09-09
[Windows] Fix TestURL/test_URLStrings

gmittert:
Since '|' is a valid character in a url on Windows as implemented by
CoreFoundation, on Windows, we should expect that `unescaped|pipe`
should be properly parsed.

This is hacky, but adds a test case on Windows, and disables it on other
platforms and vice versa.

### DIFF
--- a/TestFoundation/Resources/NSURLTestData.plist
+++ b/TestFoundation/Resources/NSURLTestData.plist
@@ -3493,13 +3493,67 @@
 		</dict>
 		<dict>
 			<key>In-Title</key>
-			<string>NSURLWithString-parse-absolute-escape-006</string>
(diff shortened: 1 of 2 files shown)


gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test

gmittert:
@swift-ci please test linux

gmittert:
@millenomi Are you okay doing testing like this? Is there a better way to do it?

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

compnerd:
Thank you for cleaning up the logic here!  LGTM

PR #2508, opened 2019-09-10
[Windows] Disable TestHTTPCookieStorage XDG Test

gmittert:
Windows doesn't conform to XDG so the test isn't applicable


gmittert:

PR #2510, opened 2019-09-11
[Windows, CoreFoundation] Construct Preferences Using Known Locations

gmittert:
The current Windows implementation doesn't properly handle
kCFPreferencesAnyUsers, which causes it point to the same location as
kCFPreferencesCurrentUser which causes preferences to be loaded
incorrectly.

Instead, we should defer to CFKnownLocations as the non Windows
implementation already does


compnerd:
@swift-ci please test

PR #2511, opened 2019-09-12
[Windows] Match error behavior of POSIX

gmittert:
On POSIX Foundation, removing  a file is considered writing to it and so
when the file to be removed isn't found, it throws fileNoSuchFile. Since
Windows instead checks for the file attributes first which is considered
a read, it throws there with fileReadNoSuchFile. In removeItem to match
that behavior, we catch the fileReadNoSuchFile and turn it into
fileNoSuchFile.


gmittert:
@swift-ci please test linux

PR #2512, opened 2019-09-13
[CoreFoundation] Use a Windows Timezone if we are given one

gmittert:
Before attempting to check if a timezone is an abbreviation or Olson
format, we should make sure we can't just query Windows for the
timezone. For example, CFTimeZoneCopyKnownNames will return UTC[-+XX]
timezones which should otherwise would fail on.

### DIFF
--- a/CoreFoundation/NumberDate.subproj/CFTimeZone.c
+++ b/CoreFoundation/NumberDate.subproj/CFTimeZone.c
@@ -1307,6 +1307,15 @@ Boolean _CFTimeZoneInit(CFTimeZoneRef timeZone, CFStringRef name, CFDataRef data
     Boolean result = false;
 
 #if TARGET_OS_WIN32
+    // Start by checking if we're just given a timezone Windows knows about
+    int32_t offset;
+    __CFTimeZoneGetOffset(name, &offset);
+    if (offset) {
+        // TODO: handle DST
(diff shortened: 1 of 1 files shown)


CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1319,9 +1328,8 @@ Boolean _CFTimeZoneInit(CFTimeZoneRef timeZone, CFStringRef name, CFDataRef data
     CFRelease(abbrevs);
 
     if (tzName) {
-        int32_t offset;
         __CFTimeZoneGetOffset(tzName, &offset);
-        // TODO(compnerd) handle DST
+        // TODO: handle DST
```
gmittert: I'm thinking here we might not need to care about handling DST. The CF functions which expose DST functionality accept an absolute time which they use to check against. Since we're not given a date here anyways, the correct thing to do might be to ignore it until given a date.

gmittert:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
Please synchronise with @millenomi on merging this

compnerd:
@swift-ci please test

gmittert:
@swift-ci please test linux

PR #1729, opened 2018-10-18
Foundation: synchronize macOS/iOS and Linux paths

compnerd:

PR #2513, opened 2019-09-16
Enable Enum Imports and Synchronize Paths

gmittert:
Windows wasn't going down the 64bit enum path in CoreFoundation which caused CFDateComponentUndefined to be 32bit and NSDateComponentUndefined and thus not match. Since changing this causes the enum behavior to change, I've rebased https://github.com/apple/swift-corelibs-foundation/pull/1729 and included it in this pull.


gmittert:
@swift-ci please test

compnerd:
I'm in favour of this, this looks good to me.  Lets just poke @millenomi and @parkera before merging this.

PR #2515, opened 2019-09-17
CoreFoundation: correct typo in ASM Macros

compnerd:
Correct the else if macro usage, which was resulting in redefinition of macros
incorrectly.


compnerd:
@swift-ci please test

PR #2516, opened 2019-09-18
Fix timezone creation on Windows

gmittert:
Windows returns the bias in minutes, but CoreFoundation expects the
offset in seconds.


gmittert:

gmittert:
cc @compnerd

PR #2517, opened 2019-09-18
Remove old Windows Code

gmittert:
Windows now has a database (the registry) and can handle GMT/UTC, so
this is irrelevant code.


gmittert:
@swift-ci please test

PR #2518, opened 2019-09-19
Include null terminator in length for getCString

gmittert:
Both CFStringGetCString and NSString._getCString expect the buffer size
to include space for a null terminator, so there is no reason to
subtract 1 from the buffer size before handing off to NSString

This was causing the last character to be truncated in certain calls to CFStringGetCString.


gmittert:
@swift-ci please test

parkera:
Hm, on Darwin this boils down to calling the `getBytes` API on `NSString` which does not add a null...

shahmishal:

PR #2520, opened 2019-09-25
[Android] Use CFFIXED_USER_HOME for bundle path

Molanda:
For Android, the main bundle path is set to the executable directory, which is something like `/system/bin/app_process32`.  As a result, things like localized strings do not work.

This PR allows the bundle path to be set by the `CFFIXED_USER_HOME` environmental variable.

In my case, I set this in the Kotlin (or Java) initialization code as follows...

```
Os.setenv("CFFIXED_USER_HOME", mainActivity.filesDir.path, true)
```

I then copy the files from the asset manager into the files directory.  This includes `Info.plist` which configures the bundle identifier and development region, and `en.lproj/Localizable.strings` which contain the English strings.

Once these are in place, I can call `NSLocalizedString` to get the localized string.

### DIFF
--- a/CoreFoundation/PlugIn.subproj/CFBundle_Executable.c
+++ b/CoreFoundation/PlugIn.subproj/CFBundle_Executable.c
@@ -310,6 +310,20 @@ static CFURLRef _CFBundleCopyBundleURLForExecutablePath(CFStringRef str) {
     CFIndex buffLen;
     CFURLRef url = NULL;
     CFStringRef outstr;
+
+#if TARGET_OS_ANDROID
+    const char *fixedUserHome = __CFgetenv("CFFIXED_USER_HOME");
+    if (fixedUserHome) {
+        outstr = CFStringCreateWithCString(kCFAllocatorSystemDefault, fixedUserHome, kCFStringEncodingUTF8);
+        if (outstr) {
+            url = CFURLCreateWithFileSystemPath(kCFAllocatorSystemDefault, outstr, PLATFORM_PATH_STYLE, true);
+            CFRelease(outstr);
+            if (url) {
+                return url;
+            }
+        }
+    }
+#endif
     
     buffLen = CFStringGetLength(str);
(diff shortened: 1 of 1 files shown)


Molanda:
@swift-ci please test

Molanda:
Courtesy ping.

Molanda:
Any comments on this one?  It should be consistent with other areas configured for Android.

drodriguez:
I never seen the `/system/bin/app_process32` when starting executables directly. Does that only happen when the Swift code is embedded into an apk and the code is used through JNI?

@swift-ci please test

Molanda:
Here is a good description from [StackExchange](https://unix.stackexchange.com/questions/534200/full-path-of-running-processes-in-android):

> However, for _all_ the android _apps_, that's just `/system/bin/app_process32` or `/system/bin/app_process64` ie the same executable as the `main`/`zygote` process (itself started by `init`).
>
>That process will just fork children process which will then load the java classes making up an Android app, without `exec()`ing any extra binary.
>
>The command lines you see in the output of `ps` (`/proc/PID/cmdline`) are just "faked" as usual by overwriting the `argv` block -- see [here](https://android.googlesource.com/platform/frameworks/base/+/e6024b34edaa57817149b94da563dddd5ee30764/core/jni/AndroidRuntime.cpp#323).

This commit is written so if there is no wrapper Java/Kotlin code setting `CFFIXED_USER_HOME` and loading Foundation,  the original code path will determine the bundle path.

drodriguez:
Yes, that's what I imagined, that it was the original process for zygote. I don't see anything wrong in the code, and should only affect Android. The failure in macOS seems unrelated. Let's see if it goes away.

Molanda:
Thank you for looking the code over and running the tests.

drodriguez:
I think it is good to merge. I think the problem in macOS is unrelated, or I cannot see how this change can affect SwiftPM at all.

PR #2522, opened 2019-09-27
UBSan: Fix unaligned load/stores found by Undefined Behaviour Sanitizer.

spevans:
- Add unaligned_load32() / unaligned_store32() and
  unaligned_load{16,32,64}be() that load/store using byte access. On
  supported architectures clang should convert these into single
  load / store instructions.

- When allocating buffers on the stack using STACK_BUFFER_DECL() ensure
  the length is > 0, as VLA (Variable Length Arrays) cannot be 0 length.

### DIFF
--- a/CoreFoundation/Parsing.subproj/CFBinaryPList.c
+++ b/CoreFoundation/Parsing.subproj/CFBinaryPList.c
@@ -501,7 +501,7 @@ static void _flattenPlist(CFPropertyListRef plist, CFMutableArrayRef objlist, CF
     CFDictionaryAddValue(objtable, plist, (const void *)(uintptr_t)refnum);
     if (_kCFRuntimeIDCFDictionary == type) {
         CFIndex count = CFDictionaryGetCount((CFDictionaryRef)plist);
-        STACK_BUFFER_DECL(CFPropertyListRef, buffer, count <= 128 ? count * 2 : 1);
+        STACK_BUFFER_DECL(CFPropertyListRef, buffer, (count > 0 && count <= 128) ? count * 2 : 1);
         CFPropertyListRef *list = (count <= 128) ? buffer : (CFPropertyListRef *)CFAllocatorAllocate(kCFAllocatorSystemDefault, 2 * count * sizeof(CFTypeRef), 0);
         CFDictionaryGetKeysAndValues((CFDictionaryRef)plist, list, list + count);
         for (CFIndex idx = 0; idx < 2 * count; idx++) {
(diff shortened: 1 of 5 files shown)


spevans:
@swift-ci test

spevans:
@swift-ci test macos

CoreFoundation/Parsing.subproj/CFBinaryPList.c
```diff
@@ -501,7 +501,7 @@ static void _flattenPlist(CFPropertyListRef plist, CFMutableArrayRef objlist, CF
     CFDictionaryAddValue(objtable, plist, (const void *)(uintptr_t)refnum);
     if (_kCFRuntimeIDCFDictionary == type) {
         CFIndex count = CFDictionaryGetCount((CFDictionaryRef)plist);
-        STACK_BUFFER_DECL(CFPropertyListRef, buffer, count <= 128 ? count * 2 : 1);
+        STACK_BUFFER_DECL(CFPropertyListRef, buffer, (count > 0 && count <= 128) ? count * 2 : 1);
```
parkera: Does this same check need to be repeated on the next line?

CoreFoundation/String.subproj/CFString.c
```diff
@@ -535,6 +535,18 @@ CFStringEncoding __CFStringComputeEightBitStringEncoding(void) {
...
 */
 CF_INLINE Boolean __CFBytesInASCII(const uint8_t *bytes, CFIndex len) {
+#if TARGET_RT_64_BIT
+    uint64_t align_mask = 7;
+#else
+    uint32_t align_mask = 3;
+#endif
+
+    /* Read bytes until the buffer is aligned. */
+    while( ( (uintptr_t)bytes & align_mask ) && len > 0) {
```
parkera: Style nit; let's make it match the rest of the file

CoreFoundation/Parsing.subproj/CFBinaryPList.c
spevans: It shouldn't be needed because if `count == 0` then the stack allocation is used instead of the `malloc()`'d buffer. 1 byte is wasted on the stack but this is also the case if `count > 128` and the heap is used instead of the stack.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@millenomi @phausler would you be able to have a look over this if you have some time?

spevans:
@swift-ci test and merge

PR #2526, opened 2019-09-30
SR-11408: Foundation.Process: can invoke run() again without error

spevans:
- Throw an error if run() is called more than once.


spevans:
@swift-ci test

spevans:
@swift-ci test linux

spevans:
@swift-ci test and merge

PR #2527, opened 2019-09-30
[test] Rewrite URLSession cookie tests to be independent.

drodriguez:
Some cookie tests were depending on previously run tests to be
successful. Rewrite all the tests to be independent from each other.

The first change is that all the tests that deal with cookies try to
clean their cookie storages before starting the testing. This should
avoid previous state leaking into the current test.

Some other changes are made to clarify how the cookie tests are actually
performed. To check that the cookies are set by URLSession, an special
endpoint that echoes back the request header are used. To make it clear,
that endpoint has changed names.

In the cases that we were checking that the response headers were
setting the cookies, we actually check for the right headers to been
set, not some random data in the response (because for some reason, the
request headers were also being echo in another endpoint, which might
have created false positives).

There is also some cleaning of unused variables and var that should have
been let.

I tested this by running all the tests in TestURLSession in the same
run, and also running the individual tests one by one from the command
line.

I recommend reading this diff with whitespace changes disabled (`?w=1` in the address bar).

### DIFF
--- a/TestFoundation/HTTPServer.swift
+++ b/TestFoundation/HTTPServer.swift
@@ -588,17 +588,16 @@ public class TestURLSessionServer {
         }
 
         if uri == "/requestCookies" {
-            let text = request.getCommaSeparatedHeaders()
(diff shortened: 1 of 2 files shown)


drodriguez:
@swift-ci please test

drodriguez:
Found in the review of #2521.

PR #2538, opened 2019-10-15
[Windows] Search for Dlls in Bundles Properly

gmittert:
On Windows, dlls don't begin with "lib" so the tests should be searching
for "MyDLL.dll", not "libMyDLL.dll"


gmittert:

PR #2540, opened 2019-10-16
[Windows] Remove Fallback Path with Loading Bundles

gmittert:
When loading an invalid bundle rather than failing as the Darwin path
does, it instead returns the parent directory of the given bundle as an
unloaded bundle.

This removes the hunk so that the Windows behavior matches the other
platforms.


gmittert:
cc @millenomi @compnerd @parkera 

I don't have context about what this bit was actually added for, it's part of the original implementation commit with no other explanation about why it's needed. Does anyone know what might break if it is removed?

gmittert:
@swift-ci please test linux

PR #2543, opened 2019-10-16
Fix Running Synchronous Operations Without a Queue

Frizlab:
Fixed issue SR-11621.

When calling `start()` on a synchronous operation, the main was not called.


spevans:
@swift-ci test

PR #2545, opened 2019-10-17
Process: Setting stdout or stderr to nil or nullDevice set stdin instead

spevans:
- If .standardOutput or .standardError was set to nil or
  FileHandle.nullDevice, stdin was set to /dev/null instead.


spevans:
@swift-ci test

spevans:
@weissi Probably worth adding this to the next monthly bug release

weissi:
@spevans agreed! Once we open the merge window for next month's release we should do this.

spevans:
@swift-ci test and merge

PR #2546, opened 2019-10-20
statx: fallback to lstat if ENOSYS is returned in errno.

spevans:
- Security mechanisms used in Linux VMs, eg libseccomp and
  Docker's security policy may block certain system calls if they
  are new and unknown to the library and ENOSYS may be returned
  instead of EPERM.

- Update the use of statx() to disable future calls and fallback to
  lstat() if ENOSYS is returned.


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

PR #2550, opened 2019-10-30
Simplify CFString's Get MaximumSizeForEncoding so its program flow is clearer

pi1024e:
CFString's Get MaximumSizeForEncoding method is not clear in that the else branch ALWAYS following the rest of the method due to early returns.

Because the two top branches always return, I removed the else statement, and put encoding & 0xFFF in the switch statement itself instead of modifying the encoding directly.


pi1024e:
@swift-ci test

parkera:
@swift-ci test

parkera:
Thank you!

PR #2552, opened 2019-11-01
SR-11687: Scanner.scanDouble returns Optional(0.0) instead of nil on failure

spevans:
- Fix Scanner._scan(): Make the local result an Optional and dont
  initalise it until a valid number has been found.


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

PR #2554, opened 2019-11-03
Use receiver's timezone in isValidDate() as per documentation

ps2:
The [documentation of DateComponents.isValidDate(in:)](https://developer.apple.com/documentation/foundation/nsdatecomponents/1412707-isvaliddate) indicates that the receiver's timezone is used, if it is set. The code also appears to be attempting to implement this, by setting the `timeZone` property on a local `cal` var, but this var wasn't being used at all later in the function. 

The result is that seemingly valid DateComponents can return false for isValidDate(), when the receiver's timezone differs from the passed in calendar's timezone, which is often system timezone.

### DIFF
--- a/Foundation/NSCalendar.swift
+++ b/Foundation/NSCalendar.swift
@@ -1904,13 +1904,13 @@ open class NSDateComponents : NSObject, NSCopying, NSSecureCoding {
         if ns != NSDateComponentUndefined && 0 < ns {
             nanosecond = 0
         }
-        let d = calendar.date(from: self._swiftObject)
+        let d = cal.date(from: self._swiftObject)
         if ns != NSDateComponentUndefined && 0 < ns {
             nanosecond = ns
         }
         if let date = d {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

spevans:
@ps2 Does this also fix https://bugs.swift.org/browse/SR-11569 ?

ps2:
@spevans yes, the description and the test case match the behavior I saw and fixed in this ticket. Although the bug does reproduce on Mac for me.

ps2:
@shahmishal Is this fixed?  I noticed that https://bugs.swift.org/browse/SR-11569 was marked fixed in Jan, but I don't see any related code changes to this method.

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please rebase on top of the main and we can reopen.

shahmishal:

PR #2555, opened 2019-11-06
Foundation: remove executable permission (NFC)

compnerd:
Remove execute permissions on NSPathUtilties.swift.  This is a source
file, not an executable.


compnerd:
@swift-ci please test

PR #2561, opened 2019-11-12
SR-11766: Fix coefficients for UnitVolume

ole:
This fixes Measurement values expressed in cubic centimeters and cubic millimeters.

I found the errors in the coefficients because I tested one of my libraries on Linux for the first time and a test was failing (see https://github.com/ole/Ampere/issues/25).

I considered adding unit tests that test some conversions to Foundation, but I'm not sure how useful they would be. A test for converting a Measurement from one unit to another would essentially repeat the code that's being tested; it seems it would be almost equally likely for a bug to be in the test as it would be in the code being tested.

### DIFF
--- a/TestFoundation/TestUnitVolume.swift
+++ b/TestFoundation/TestUnitVolume.swift
@@ -1,0 +1,76 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+class TestUnitVolume: XCTestCase {
+    func testMetricVolumeConversions() {
+        let cubicKilometers = Measurement(value: 4, unit: UnitVolume.cubicKilometers)
+        XCTAssertEqual(cubicKilometers, Measurement(value: 4e9, unit: UnitVolume.cubicMeters), "Conversion from cubicKilometers to cubicMeters")
+
+        let megaliters = Measurement(value: 3, unit: UnitVolume.megaliters)
+        XCTAssertEqual(megaliters, Measurement(value: 3_000_000, unit: UnitVolume.liters), "Conversion from megaliters to liters")
+
+        let kiloliters = Measurement(value: 2, unit: UnitVolume.kiloliters)
+        XCTAssertEqual(kiloliters, Measurement(value: 2000, unit: UnitVolume.liters), "Conversion from kiloliters to liters")
+
+        let cubicMeters = Measurement(value: 2, unit: UnitVolume.cubicMeters)
+        XCTAssertEqual(cubicMeters, Measurement(value: 2000, unit: UnitVolume.liters), "Conversion from cubicMeters to liters")
+        XCTAssertEqual(kiloliters, cubicMeters, "Conversion from kiloliters to cubicMeters")
+
+        let liters = Measurement(value: 5, unit: UnitVolume.liters)
+        XCTAssertEqual(liters, Measurement(value: 5, unit: UnitVolume.cubicDecimeters), "Conversion from liters to cubicDecimeters")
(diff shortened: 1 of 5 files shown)


ole:
Please let me know if you'd consider tests of this kind useful:

```swift
func testVolumeConversions() {
    let input = Measurement(value: 5, unit: UnitVolume.liters)
    let megaliters = input.converted(to: .megaliters)
    XCTAssertEqual(megaliters.value, 5e-6, "Conversion from liters to megaliters")
    let kiloliters = input.converted(to: .kiloliters)
    XCTAssertEqual(kiloliters.value, 0.005, "Conversion from liters to kiloliters")
    let deciliters = input.converted(to: .deciliters)
    XCTAssertEqual(deciliters.value, 50, "Conversion from liters to deciliters")
    ...
```

spevans:
@swift-ci test

spevans:
@ole It would definitely be good to get some tests added, currently there dont seem to be very many. If you could add a couple that would validate the constants used for conversion that would be great. Also, don't worry about the macOS test failure, thats an unrelated issue.

spevans:
@swift-ci test macos

ole:
@spevans Thanks. I will add tests, but I don't know when I'll get to it. Hopefully sometime next week.

parkera:
Thank you @ole !

ole:
@spevans @parkera I added some tests for `UnitVolume` conversions. Let me know what you think. I tried to use conversions that readers of the tests would find reasonably intuitive so that mistakes have a chance to be spotted (e.g. it's more likely that readers will know there should be 4 quarts in a gallon than that 1 gallon should be x.xx cubic inches).

TestFoundation/TestUnitVolume.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    func testMetricToImperialVolumeConversion() {
+        let liters = Measurement(value: 10, unit: UnitVolume.liters)
+        XCTAssertEqual(liters.converted(to: .cubicInches).value, 610.236, accuracy: 0.001, "Conversion from liters to cubicInches")
+    }
+
+    func testImperialVolumeConversions() {
+        let cubicMiles = Measurement(value: 1, unit: UnitVolume.cubicMiles)
+        XCTAssertEqual(cubicMiles.converted(to: .cubicYards).value, 1760 * 1760 * 1760, accuracy: 1_000_000, "Conversion from cubicMiles to cubicYards")
```
ole: Admittedly, this assertion with the insanely high accuracy threshold is kind of ridiculous. I wanted the test to be intuitive, so for people who know that 1 mile = 1760 yards, it makes sense that 1 cubic mile = 1760^3 cubic yards.

Unfortunately, that's not really the case with `Measurement` (even after factoring in floating-point rounding errors) because the conversion factors hardcoded by Foundation often don't use the full range offered by `Double`. In this instance, `cubicMiles` is defined like this (in `Unit.swift`):

```swift
static let cubicMiles           = 4.168e+12 // in liters
```

Whereas the "accurate" conversion factor is more like `4.16818182544058e12` (= 16093.44^3; 1 mile is 16093.44 decimeters) (I think), which leads to the huge discrepancy (in absolute terms) between the actual and expected values.

This inaccuracy is not specific to corelibs-foundation, but is the same in Darwin-Foundation. Out of interest, was it a conscious decision to limit conversion factors to x significant digits or something when the Measurement types were introduced?

spevans:
@swift-ci test linux

TestFoundation/TestUnitVolume.swift
spevans: As far as scl-foundation is concerned it just tries to match Darwin's version as closely as possible, I don't know why Foundation has such low accuracy for its conversions.

The tests should hopefully catch any accidental changes in the future so thanks for adding them, Ive kicked off a CI test for them on Linux and will merge if the tests work ok.

spevans:
@swift-ci test linux

spevans:
@ole You'll need to add `TestUnitVolume.swift` into the file list in `TestFoundation/CMakeLists.txt ` as well - this was the cause of the linux test failure.

ole:
@spevans Done. Sorry I missed this.

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2562, opened 2019-11-12
build: add workaround for CMake 3.15

compnerd:
When building with CMake <3.16, the `target_link_directories` would be
ignored and the test binary would fail to link on Windows.  Add a
workaround to ensure that the library path is honoured.


compnerd:

PR #2563, opened 2019-11-12
build: attempt to restore android cross-compilation

compnerd:
When building on Windows, the check would incorrectly fail, resulting in
the inability to cross-compile from Windows.


compnerd:
@swift-ci please test

compnerd:
*sigh* this doesn't actually fix the issue :-(

PR #2564, opened 2019-11-12
build: add workaround for CMake <3.16

compnerd:
This repairs the TestFoundation build for CMake <3.16 which failed to
link against the import libraries on Windows.


compnerd:
@swift-ci please test

PR #2566, opened 2019-11-13
build: use export targets for XCTest

compnerd:
This adjusts the build to use the CMake config and export targets to
ensure that things are rebuilt when appropriate.  This also reduces the
use of custom variables.


compnerd:
apple/swift-corelibs-xctest#290
apple/swift#28232

@swift-ci please test

compnerd:
apple/swift-corelibs-xctest#290

@swift-ci please test

PR #2567, opened 2019-11-14
build: prune some unnecessary CMake files (NFC)

compnerd:
The FindICU routines from CMake are sufficient for our needs now that we
have a minimum version which provides the package.  The FindUUID module
is no longer needed as we build uuid as part of our build.


compnerd:
@swift-ci please test

PR #2572, opened 2019-11-21
[cmake] Fix dependencies for xdgTestHelp

gmittert:
xdgTestHelper imports FoundationNetworking and FoundationXML but doesn't
have an explicit dependency on them which can cause a race condition
while building.


gmittert:
cc @compnerd 

gmittert:
@swift-ci please test macOS

PR #2573, opened 2019-11-21
Work around for CMake 3.15 Resource file linking

gmittert:
CMake 3.15 doesn't properly include resource files when linking by being
passed a target which means Windows doesn't include the Olson timezone
data. Instead as a work around, link against the target object using
target_link_options.


compnerd:
@swift-ci please test

gmittert:
@swift-ci please test macos

PR #2575, opened 2019-12-03
[Windows] Fix Interpretation of Volume Name Array

gmittert:
Taking &wszPathNAmes[...] doesn't properly get the string and leads to
garbled data. Instead, compute the pointer address and pass that.

### DIFF
--- a/Foundation/FileManager+Win32.swift
+++ b/Foundation/FileManager+Win32.swift
@@ -43,15 +43,21 @@ extension FileManager {
                 continue
             }
 
-            var pPath: DWORD = 0
-            repeat {
-                let path: String = String(decodingCString: &wszPathNames[Int(pPath)], as: UTF16.self)
-                if path.length == 0 {
-                    break
+            // GetVolumePathNamesForVolumeNameW writes an array of
+            // null terminated wchar strings followed by an additional
+            // null terminator.
+            // e.g. [ "C", ":", "\\", "\0", "D", ":", "\\", "\0", "\0"]
+            var remaining = wszPathNames[...]
+            while !remaining.isEmpty {
+                let path = remaining.withUnsafeBufferPointer {
+                    String(decodingCString: $0.baseAddress!, as: UTF16.self)
                 }
-                urls.append(URL(fileURLWithPath: path, isDirectory: true))
-                pPath += DWORD(path.length + 1)
-            } while pPath < dwCChReturnLength
+
+                if !path.isEmpty {
+                    urls.append(URL(fileURLWithPath: path, isDirectory: true))
+                }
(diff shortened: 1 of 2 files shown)


gmittert:
@swift-ci please test

gmittert:
@swift-ci please test macos

Foundation/FileManager+Win32.swift
```diff
@@ -39,19 +39,19 @@ extension FileManager {
...
-                let path: String = String(decodingCString: &wszPathNames[Int(pPath)], as: UTF16.self)
-                if path.length == 0 {
-                    break
-                }
-                urls.append(URL(fileURLWithPath: path, isDirectory: true))
-                pPath += DWORD(path.length + 1)
-            } while pPath < dwCChReturnLength
+            var pPath: Int = 0
+            while wszPathNames[pPath] != 0 {
+                guard let (url, pathLength): (URL, Int) = wszPathNames.withUnsafeBufferPointer({
```
compnerd: This is confusing, can we not move the base address computation above the rest of the logic rather than re-computing the offset each time?

Foundation/FileManager+Win32.swift
```diff
@@ -39,19 +39,19 @@ extension FileManager {
...
+                    guard let pathStart = $0.baseAddress else { return nil }
+                    let path = String(decodingCString: pathStart + pPath, as: UTF16.self)
+                    return (URL(fileURLWithPath: path, isDirectory: true), path.length + 1)
+                }) else { return nil }
+                urls.append(url)
```
compnerd: Why not just create the URL in the `append` and return the string.

Foundation/FileManager+Win32.swift
gmittert: Cleared up the logic a bit so it's easier to read. The base address is now computed by the ArraySlice.

Foundation/FileManager+Win32.swift
```diff
@@ -39,19 +39,19 @@ extension FileManager {
 
             var wszPathNames: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(dwCChReturnLength + 1))
             if !GetVolumePathNamesForVolumeNameW(&wszVolumeName, &wszPathNames, DWORD(wszPathNames.count), &dwCChReturnLength) {
-                // TODO(compnerd) handle error
-                continue
+                return nil
```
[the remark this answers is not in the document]
gmittert: Fair, I originally figured it was better to return no answer than an incomplete one, but I can see either way.

gmittert:
@swift-ci please test linux

gmittert:
@swift-ci please test linux

PR #2577, opened 2019-12-05
Fix TestFileManager Crash in test_displayNames

gmittert:
test_displayNames XCTAsserts that there are enough components in the
array, but if the test fails it still continues on, subtracting from
components.count and causing a negative number out of bounds access.
This doesn't prevent the test from failing since the GreaterThanOrEqual
check will still fail, but it does prevent the test from causing the
test suite to crash.


gmittert:
@swift-ci please test linux

gmittert:
cc @compnerd

PR #2578, opened 2019-12-05
[Windows] Fix TestProcess.swift

gmittert:
Modified the error handling of launching a process in a non existing
working directory so that it matches POSIX behaviour. This causes all
the Process tests to pass on Windows now.


gmittert:
@swift-ci please test

PR #2579, opened 2019-12-06
[Windows] Fix precision loss in CFGetSystemUptime()

lxbndr:
Integer division rounds result to whole seconds. This forces floating point division.

Noticed the issue because `XCTest.measure` depends on this value (through `ProcessInfo.systemUptime`) and produces unreliable results.


compnerd:
@swift-ci please test

PR #2580, opened 2019-12-06
Properly Reset Deadlines when Rearming Timers

gmittert:
In the case where Windows schedules timers too early and CoreFoundation
needs to rearm them, the deadlines need to be reset first else timers
will not be rearmed properly.


gmittert:
@swift-ci please test linux

gmittert:
cc @compnerd  @millenomi

PR #2581, opened 2019-12-06
[Windows] Special case ~ as an absolute path

gmittert:
PathIsRelativeW doesn't consider ~ to be an absolute path, so we special
case it in to match posix behaviour


gmittert:
@swift-ci please test linux

PR #2582, opened 2019-12-09
SR-11922: [FileHandle] Make the function signature the same as DarwinFoundation.

YOCKOW:
Since `func truncateFile(atOffset offset: UInt64)` has been deprecated, new function is introduced. However, the signatures are different between DarwinFoundation and SwiftFoundation.
Resolves [SR-11922](https://bugs.swift.org/browse/SR-11922).

### DIFF
--- a/Foundation/FileHandle.swift
+++ b/Foundation/FileHandle.swift
@@ -570,7 +570,7 @@ open class FileHandle : NSObject {
     }
     
     @available(swift 5.0)
-    public func truncate(toOffset offset: UInt64) throws {
+    public func truncate(atOffset offset: UInt64) throws {
         guard self != FileHandle._nulldeviceFileHandle else { return }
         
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test linux

spevans:
@YOCKOW The old methods still needs to be tested so I would suggest restoring  the original `test_truncateFile` and renaming the new one to  `test_throwingTruncateFile` or something.

YOCKOW:
@spevans 
Thank you for pointing that out. I'll fix it and rebase.

YOCKOW:
Rebased.
Now the original tests remain and just new tests for `func truncate(atOffset:)` have been added.

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2586, opened 2019-12-12
don't rely on getdtablesize()

weissi:
Android and possibly other UNIX systems lack getdtablesize() so let's
try to use a better way of estimating the file descriptors we need to
close post-fork, pre-exec by listing /proc/self/fd. If that fails, we
fall back on getdtablesize or as a last resort a hard-coded 4096.

### DIFF
--- a/Foundation/Process.swift
+++ b/Foundation/Process.swift
@@ -60,6 +60,43 @@ extension CFSocketError {
 }
 #endif
 
+#if !canImport(Darwin) && !os(Windows)
+private func findMaximumOpenFromProcSelfFD() -> CInt? {
+    guard let dirPtr = opendir("/proc/self/fd") else {
+        return nil
+    }
+    defer {
+        closedir(dirPtr)
+    }
+    var highestFDSoFar = CInt(0)
+
+    while let dirEntPtr = readdir(dirPtr) {
+        var entryName = dirEntPtr.pointee.d_name
+        let thisFD = withUnsafeBytes(of: &entryName) { entryNamePtr -> CInt? in
+            CInt(String(decoding: entryNamePtr.prefix(while: { $0 != 0 }), as: Unicode.UTF8.self))
+        }
+        highestFDSoFar = max(thisFD ?? -1, highestFDSoFar)
+    }
+
+    return highestFDSoFar
+}
+
+func findMaximumOpenFD() -> CInt {
+    if let maxFD = findMaximumOpenFromProcSelfFD() {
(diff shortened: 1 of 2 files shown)


weissi:
@swift-ci please test

weissi:
@swift-ci please test

Foundation/Process.swift
```diff
@@ -60,6 +60,43 @@ extension CFSocketError {
...
+
+    // We don't have /proc, let's go with the best estimate.
+#if os(Linux)
+    return getdtablesize()
+#else
+    return 4096
+#endif
+}
+#endif
+
```
spevans: I think this all needs to inside a `#if !os(Windows)`

weissi: oh yes, thank you!

weissi: done

weissi:
@swift-ci please test

weissi:
@swift-ci please test

weissi:
@swift-ci please test

weissi:
@swift-ci please test

weissi:
@swift-ci please test

weissi:
@swift-ci please test

weissi:
@swift-ci please test

Foundation/Process.swift
```diff
@@ -60,6 +60,43 @@ extension CFSocketError {
...
+    while let dirEntPtr = readdir(dirPtr) {
+        var entryName = dirEntPtr.pointee.d_name
+        let thisFD = withUnsafeBytes(of: &entryName) { entryNamePtr -> CInt? in
+            CInt(String(decoding: entryNamePtr.prefix(while: { $0 != 0 }), as: Unicode.UTF8.self))
+        }
+        highestFDSoFar = max(thisFD ?? -1, highestFDSoFar)
+    }
+
+    return highestFDSoFar
+}
```
spevans: Would 
```return try? FileManager.default.contentsOfDirectory(atPath: "/proc/self/fd").compactMap { CInt($0) }.max()```

be a bit simpler?

weissi: @spevans I wasn't sure about the guarantees around symlinks as in does it follow symlinks, does it not, what happens if the symlink is a broken link, etc.  Therefore I went for `readdir`, happy to change if we think it's fine.

compnerd:
Android CI: https://dev.azure.com/compnerd/windows-swift/_build/results?buildId=18717&view=results

PR #2587, opened 2019-12-13
Make URLSession and URLSessionTask more source-compatible with Darwin

broadwaylamb:
I'd be happy to write tests, but what should we test here?

cc @parkera

### DIFF
--- a/Foundation/URLSession/URLSession.swift
+++ b/Foundation/URLSession/URLSession.swift
@@ -268,7 +268,7 @@ open class URLSession : NSObject {
     }
     
     open private(set) var delegateQueue: OperationQueue
-    open var delegate: URLSessionDelegate?
+    open private(set) var delegate: URLSessionDelegate?
     open private(set) var configuration: URLSessionConfiguration
     
     /*
@@ -370,8 +370,13 @@ open class URLSession : NSObject {
         }
     }
 
+    @available(*, unavailable, renamed: "getTasksWithCompletionHandler(_:)")
+    open func getTasksWithCompletionHandler(completionHandler: @escaping ([URLSessionDataTask], [URLSessionUploadTask], [URLSessionDownloadTask]) -> Void) {
+        getTasksWithCompletionHandler(completionHandler)
+    }
+
     /* invokes completionHandler with outstanding data, upload and download tasks. */
-    open func getTasksWithCompletionHandler(completionHandler: @escaping ([URLSessionDataTask], [URLSessionUploadTask], [URLSessionDownloadTask]) -> Void)  {
+    open func getTasksWithCompletionHandler(_ completionHandler: @escaping ([URLSessionDataTask], [URLSessionUploadTask], [URLSessionDownloadTask]) -> Void)  {
         workQueue.async {
             self.delegateQueue.addOperation {
                 var dataTasks = [URLSessionDataTask]()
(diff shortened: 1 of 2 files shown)


Foundation/URLSession/URLSession.swift
```diff
@@ -419,6 +424,22 @@ open class URLSession : NSObject {
...
+     * see <Foundation/NSURLError.h>.  The delegate, if any, will still be
+     * called for authentication challenges.
+     */
+    open func dataTask(with request: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask {
+        return dataTask(with: _Request(request), behaviour: .dataCompletionHandler(completionHandler))
+    }
+
+    open func dataTask(with url: URL, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask {
+        return dataTask(with: _Request(url), behaviour: .dataCompletionHandler(completionHandler))
+    }
```
broadwaylamb: Moved these from an extension to the class itself. With Darwin Foundation we're able to override these methods in subclasses.

Foundation/URLSession/URLSession.swift
```diff
@@ -370,8 +370,13 @@ open class URLSession : NSObject {
...
     }
 
+    @available(*, unavailable, renamed: "getTasksWithCompletionHandler(_:)")
+    open func getTasksWithCompletionHandler(completionHandler: @escaping ([URLSessionDataTask], [URLSessionUploadTask], [URLSessionDownloadTask]) -> Void) {
+        getTasksWithCompletionHandler(completionHandler)
+    }
+
     /* invokes completionHandler with outstanding data, upload and download tasks. */
-    open func getTasksWithCompletionHandler(completionHandler: @escaping ([URLSessionDataTask], [URLSessionUploadTask], [URLSessionDownloadTask]) -> Void)  {
+    open func getTasksWithCompletionHandler(_ completionHandler: @escaping ([URLSessionDataTask], [URLSessionUploadTask], [URLSessionDownloadTask]) -> Void)  {
```
broadwaylamb: [Darwin Foundation doesn't use an argument label](https://developer.apple.com/documentation/foundation/urlsession/1411578-gettaskswithcompletionhandler).

Foundation/URLSession/URLSession.swift
```diff
@@ -268,7 +268,7 @@ open class URLSession : NSObject {
     }
     
     open private(set) var delegateQueue: OperationQueue
-    open var delegate: URLSessionDelegate?
+    open private(set) var delegate: URLSessionDelegate?
```
broadwaylamb: [`URLSession.delegate` is get-only in Darwin Foundation](https://developer.apple.com/documentation/foundation/urlsession/1411530-delegate#).

Therefore, we should be able to override it with a get-only property.

Foundation/URLSession/URLSessionTask.swift
```diff
@@ -376,7 +376,7 @@ open class URLSessionTask : NSObject, NSCopying {
     /*
      * The current state of the task within the session.
      */
-    open var state: URLSessionTask.State {
+    open fileprivate(set) var state: URLSessionTask.State {
```
broadwaylamb: [`URLSessionTask.state` is get-only in Darwin Foundation](https://developer.apple.com/documentation/foundation/urlsessiontask/1409888-state#).

broadwaylamb:
Could anyone please verify this patch?

@phausler @parkera

Foundation/URLSession/URLSession.swift
drodriguez: SE-0117 strikes again.

I think you can mark _each_ method in the extension with `open`, and that should reduce the size of this PR and keep changes easier to see.

broadwaylamb: They were already `open`, it’s just that we can’t override methods declared in extensions.

drodriguez: Sorry, I didn't double checked my proposal.

I just tested in a small Xcode (Mac) project that reproduces the problem, and I get this error "Overriding non-`@objc` declarations from extensions is not supported". I tried adding a simple `@objc` in front of the extension in the Swift framework, and it seems to fix the problem.

Do you think that might be a better solution? I don't mind moving the methods into the main class code, but if someone wrote them as an extension for organization, I would prefer to keep them that one.

drodriguez:
@swift-ci please test

Foundation/URLSession/URLSession.swift
broadwaylamb: I’m afraid `@objc` won’t work on non-Darwin platforms.

drodriguez: That's true. What a strange limitation of extensions. I wonder why `open` there doesn't generate a warning, if it cannot be overriden anyway.

broadwaylamb: AFAIK this has to do with method dispatch. We can't retroactively add methods to class vtable, unless we use some completely different dispatching mechanism, like message passing in Objective-C. That's why `@objc` works — it relies on Objective-C runtime.

broadwaylamb:
Can anyone take a look and merge this please? @phausler @parkera @millenomi

drodriguez:
Merged. You might want to ask for a cherry-pick of this to the latest 5.x.y branch going on for Linux.

PR #2588, opened 2019-12-16
[CF] Fix typo in TARGET_RT_64_BIT ifdef.

3405691582:
__CFGetFullRetainCount contains an ifdef with symbol name
TARGET_RT64_BIT. This is likely a typo since otherwise,
__CFDoExternRefOperation is referenced, but is guarded by an ifdef with
name TARGET_RT_64_BIT instead, which then manifests as a missing symbol.
Other calls to __CFDoExternRefOperation are guarded by the same symbol.


spevans:
@swift-ci test

PR #2589, opened 2019-12-16
[CF] DT_WHT is unavailable on OpenBSD.

3405691582:
Here we duplicate this extended conditional to not reference DT_WHT,
which is not available on OpenBSD. It is probably harmless to actually
remove DT_WHT entirely here unconditionally, but would require grokking
more context for a simple buildfix.

It also seems more readable to duplicate the entire line in the ifdef
despite being able to squish the ifdef in the clause.

### DIFF
--- a/CoreFoundation/Base.subproj/CFFileUtilities.c
+++ b/CoreFoundation/Base.subproj/CFFileUtilities.c
@@ -439,7 +439,11 @@ CF_PRIVATE CFMutableArrayRef _CFCreateContentsOfDirectory(CFAllocatorRef alloc, 
             dirURL = CFURLCreateFromFileSystemRepresentation(alloc, (uint8_t *)dirPath, pathLength, true);
             releaseBase = true;
         }
+#if !defined(__OpenBSD__)
         if (dp->d_type == DT_DIR || dp->d_type == DT_UNKNOWN || dp->d_type == DT_LNK || dp->d_type == DT_WHT) {
+#else
+        if (dp->d_type == DT_DIR || dp->d_type == DT_UNKNOWN || dp->d_type == DT_LNK) {
+#endif
             Boolean isDir = (dp->d_type == DT_DIR);
             if (!isDir) {
                 // Ugh; must stat.


3405691582:
Let me know if you'd prefer adding some additional preprocessor symbols in (maybe) CoreFoundation_Prefix.h instead of injecting actual OS preprocessor symbols.

parkera:
Is `TARGET_OS_BSD` from `TargetConditionals.h` not appropriate?

3405691582:
DT_WHT is available on FreeBSD ([source](https://github.com/freebsd/freebsd/blob/master/sys/sys/dirent.h)) and NetBSD ([source](https://github.com/NetBSD/src/blob/trunk/sys/sys/dirent.h)), but not OpenBSD ([source](https://github.com/openbsd/src/blob/master/sys/sys/dirent.h)).

(I'm also wondering what the ideal stylistic choice is for future PRs as well; for example, Linux doesn't have xlocale, FreeBSD does, but the other BSDs do not...)

3405691582:
Perhaps it's better stylistically to leave granular portability decisions to CoreFoundation_Prefix.h: so, for this example, 

```
#define __HAS_DT_WHT__ 0
#if TARGET_OS_BSD && defined(__OpenBSD__)
#define __HAS_DT_WHT__ 1
#endif  
```

and then use __HAS_DT_WHT__ in CFFileUtilities instead? This means macros in the dependent files are focused on features as opposed to operating systems. It does make the prefix file a lot more verbose though (but on the other hand it's already rather verbose to begin with).

parkera:
I tried that approach in the past (focus on features instead of platforms) but was never happy with the result. It just results in a lot of extra macros, most of which are just aliases for "platform X".

3405691582:
Sounds good. I'll keep this commit as-is and use that pattern for future PRs.

compnerd:
@swift-ci please test

compnerd:
I wish that there was a better way to write this, but I suppose it will do for now.

3405691582:
It'll possibly look better if refactored slightly, but I'll think about circling back around to that after I'm done with all the portability tweaks.

PR #2590, opened 2019-12-17
[CF] Add TARGET_OS_BSD in some places.

3405691582:
These particular uses should hopefully be uncontroversial.

* CFPlatform.c: to use the _CFIsMainThread pthread implementation; to
  properly implement __CFTSDSetSpecific and __CFTSDGetSpecific with the
  same pthread implementation as Linux (and Mac)

* CoreFoundation_Prefix.h: to exclude strlcpy and issetugid which is
  normally available in the system; to include sys/param.h for the
  MIN macro

* CFData.c: for the __CFPageSize implementation

* CFBundle_Resources.c: to include dirent.h symbols

* CFXMLPreferencesDomain.c: to define sleep for the platform

* CFBurstTrie.c: for mmap.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -131,7 +131,7 @@ const char *_CFProcessPath(void) {
 }
 #endif
 
-#if TARGET_OS_MAC || TARGET_OS_WIN32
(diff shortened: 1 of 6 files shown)


parkera:
@swift-ci test and merge

PR #2592, opened 2019-12-18
Fix HTTPURLProtocol Pragma: no-cache  header parsing

mateuszmackowiak:
> The "Pragma" header field allows backwards compatibility with
   HTTP/1.0 caches, so that clients can specify a "no-cache" request
   that they will understand (as Cache-Control was not defined until
   HTTP/1.1).  When the Cache-Control header field is also present and
   understood in a request, Pragma is ignored.

The current implementation of `HTTPURLProtocol` duplicates the parsing of `Cache-Control` header and checks for `no-cache` but the `Pragma` header is ignored.

Also `httpResponse.statusCode` check for status codes: `200, 203, 204, 206, 300, 301, 404, 405, 410, 414, 501` was checked after the interpretations of headers wich was unnecessary.

### DIFF
--- a/Foundation/URLSession/http/HTTPURLProtocol.swift
+++ b/Foundation/URLSession/http/HTTPURLProtocol.swift
@@ -113,6 +113,25 @@ internal class _HTTPURLProtocol: _NativeProtocol {
         guard let httpRequest = task?.currentRequest else { return false }
         guard let httpResponse = response.response as? HTTPURLResponse else { return false }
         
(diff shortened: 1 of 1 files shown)


mateuszmackowiak:
@swift-ci please test

zymxxxs:
The `Pragma` Header is necessary，it fix a bug.

This processing is necessary, why is there no next operation?

shahmishal:

PR #2594, opened 2019-12-20
[CF] Add missing DEPLOYMENT_RUNTIME_SWIFT guard.

3405691582:
CFArray.c references __CFSwiftBridge which is defined behind
DEPLOYMENT_RUNTIME_SWIFT, so this needs to be behind that guard
as well when DEPLOYMENT_RUNTIME_SWIFT is not set.

### DIFF
--- a/CoreFoundation/Collections.subproj/CFArray.c
+++ b/CoreFoundation/Collections.subproj/CFArray.c
@@ -971,8 +971,10 @@ void CFArraySortValues(CFMutableArrayRef array, CFRange range, CFComparatorFunct
         result = CF_OBJC_CALLV((NSMutableArray *)array, isKindOfClass:[NSMutableArray class]);
         immutable = !result;
     } else if (CF_IS_SWIFT(CFArrayGetTypeID(), array)) {
+#if DEPLOYMENT_RUNTIME_SWIFT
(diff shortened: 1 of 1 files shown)


CoreFoundation/Collections.subproj/CFArray.c
```diff
@@ -971,8 +971,10 @@ void CFArraySortValues(CFMutableArrayRef array, CFRange range, CFComparatorFunct
         result = CF_OBJC_CALLV((NSMutableArray *)array, isKindOfClass:[NSMutableArray class]);
         immutable = !result;
     } else if (CF_IS_SWIFT(CFArrayGetTypeID(), array)) {
+#if DEPLOYMENT_RUNTIME_SWIFT
```
parkera: Should we move this up one line, such that the `else` is inside the guard too?

3405691582: It's not strictly necessary, because CF_IS_SWIFT becomes (0) in CFinternal.h:733 when DEPLOYMENT_RUNTIME_SWIFT is undefined. 

Let me know if you'd prefer it be moved, though, since it won't really hurt.

compnerd:
@swift-ci please test

PR #2596, opened 2020-01-06
build: cleanup some of the link rules

compnerd:
Many of the options were in place due to the legacy build.  Linking
against the imported libraries will both link and include the paths.
Furthermore, now with proper Swift support, it is possible to have the
library linking be computed transitively.  Avoid the double dependency
checking.  Although libdispatch is also used by Foundation, allow the
autolinking to take care of that for the Foundation library as we do not
explicitly link against dispatch anywhere.


compnerd:
@swift-ci please test

PR #2597, opened 2020-01-06
[Windows] Redirect CFLog to stderr

lxbndr:
As no specific logging is implemented for win platform, CFLog formats log message and then just drops it. Using stderr looks better than nothing.

cc @compnerd


compnerd:
Sounds like a good idea

compnerd:
@swift-ci please test

PR #2598, opened 2020-01-06
[CF] Match some declarations with their headers.

3405691582:
It appears that standardization on CF types versus C99 types was made,
but some changes were missing. As I understand it, on platforms where
Boolean (a.k.a. unsigned char, when !DEPLOYMENT_RUNTIME_SWIFT) is
convertible to bool, this doesn't cause a problem, but on OpenBSD,
bool (a.k.a., _Bool) is an enum, so causes an error.

That being said, it certainly appears intended that these should be
written as Boolean.


compnerd:
@swift-ci please test

PR #2600, opened 2020-01-09
[CF] Go from three sets of OS* decls to two.

3405691582:
In Prefix, declarations for the OSAtomic* family of functions and
OSMemoryBarrier are duplicated for Linux, BSD, and Windows. In this
commit, the declarations for OS* functions are made common, since they
are apparently identical for both targets.

Unifying the declarations for Windows is a separate task, since those
declarations use CF_EXPORT. There might be a nice way to go to just a
single set of OS* declarations, but that can come later.  This way,
this commit is just a structural change rather than a structural and
functional one.


compnerd:
Thanks for the cleanup!

compnerd:
@swift-ci please test

PR #2602, opened 2020-01-09
CoreFoundation: permit CURL's CMake package for configuration

compnerd:
Support both the CMake package provided by CURL's build as well as the
fallback path when detecting CURL.  This should restore the ability to
use zlib for compression on Windows.


compnerd:
@swift-ci please test

PR #2603, opened 2020-01-10
NSString._getBlockStart - loose precondition check to avoid crash in edge case

lxbndr:
I believe it was just a typo. There are other similar checks already, and all of them use `<=` comparison.

This was discovered as the reason of `paragraphRange(for:)` fail on whole string range. E.g, following code is crashing:
```
let s = "abc" as NSString
let range = NSRange(location: 0, length:3)
_ = s.paragraphRange(for: range) // <- crash
```

Test is also included.

Test checks paragraphRange on:
- whole string range
- every subrange in every single paragraph
- every subrange in every two consecutive paragraphs

### DIFF
--- a/Foundation/NSString.swift
+++ b/Foundation/NSString.swift
@@ -727,7 +727,7 @@ extension NSString {
         let len = length
         var ch: unichar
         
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

compnerd:
This does seem like a typo.  Thanks for fixing this.

PR #2604, opened 2020-01-12
[CF] Unify common locale macros.

3405691582:
Linux (except Android or Cygwin) and Windows don't support extended
locales, so a number of preprocessor macros are in Prefix to default to
the standard API that uses the C locale. Both Linux and Windows have
some macros in common and some platform specialized definitions; here,
we split out the common macros and keep the platform-specific
definitions separate.

This will let us more simply extend support for different platforms that
also don't support extended locales in a future commit.

### DIFF
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -180,18 +180,21 @@ bool OSAtomicCompareAndSwap32Barrier( int32_t oldValue, int32_t newValue, volati
 void OSMemoryBarrier();
 #endif // TARGET_OS_LINUX || TARGET_OS_BSD
 
+#if TARGET_OS_LINUX || TARGET_OS_WIN32
+#define strtod_l(a,b,locale) strtod(a,b)
+#define strtoul_l(a,b,c,locale) strtoul(a,b,c)
+#define strtol_l(a,b,c,locale) strtol(a,b,c)
+
+#define fprintf_l(a,locale,b,...) fprintf(a, b, __VA_ARGS__)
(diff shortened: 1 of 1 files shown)


compnerd:
I think that this is incorrect.  It is in fact the opposite I believe.  Linux but not android supports the extended locale variant.  Windows does not. I'm not certain about cygwin, but I imagine that it does as well.  It just Android (below a certain version per interface) and Windows which do not support the extended locales.

3405691582:
Right; I think I was confused with the earlier PR. Adjusted the commit description.

The code change is -- in a mechanical sense -- still OK since the macros aren't more specially controlled than TARGET_OS_LINUX. 

(Of course, this is confused by the fact that Android supports xlocale.h, these macros are defined unconditionally for Android because Android is TARGET_OS_LINUX, and then those macros become irrelevant anyway when xlocale.h is #included elsewhere in CFInternal.h. And indeed Cygwin probably supports xlocale.h but in CFInternal.h is prevented from being #include'd, so maybe it didn't in the past?)

compnerd:
> I think I was confused with the earlier PR. Adjusted the commit description to describe what CF is doing rather than what is the state of the world.

Thanks.

> The code change is -- in a mechanical sense -- still OK since the macros aren't more specially controlled than TARGET_OS_LINUX.

Agreed.

> I'm not sure off the top of my head what the state of locale support is on Linux. CF however thinks Linux doesn't have extended locale APIs, but maybe that's changed. I can look at that later.

Could you please?  That would be a great improvement.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

3405691582:
> Could you please? That would be a great improvement.

Certainly. I'm not very familiar with a lot of the locale stuff, but I'll take a look.

PR #2605, opened 2020-01-12
[CF] Cleanup commented block.

3405691582:
While we're here, remove these commented lines. These were here since
initial repo commit and are probably no longer relevant.


compnerd:

PR #2609, opened 2020-01-14
[CF] OpenBSD doesn't have the xlocale header.

3405691582:
OpenBSD is one of the platforms that doesn't have xlocale.h, so ensure
that the header is not used and the relevant macros are exposed.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -97,9 +97,9 @@ CF_EXTERN_C_BEGIN
 
 #if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_BSD
 
-#if !TARGET_OS_CYGWIN && !defined(__linux__)
+#if TARGET_OS_MAC || (TARGET_OS_BSD && !defined(__OpenBSD__)) || TARGET_OS_ANDROID
 #include <xlocale.h>
-#endif // !TARGET_OS_CYGWIN && !defined(__linux__)
+#endif // TARGET_OS_MAC || (TARGET_OS_BSD && !defined(__OpenBSD__)) || TARGET_OS_ANDROID
 
 #include <sys/time.h>
 #include <signal.h>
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -236,6 +236,10 @@ CF_INLINE uint64_t mach_absolute_time() {
 #define CF_PRIVATE extern __attribute__((visibility("hidden")))
 #define __strong
 #define __weak
+
+#if defined(__OpenBSD__)
+#define strtoll_l(a,b,c,locale) strtoll(a,b,c)
```
compnerd: Could you move this with the rest of the macros ~L186 please?

CoreFoundation/Base.subproj/CFInternal.h
```diff
@@ -97,9 +97,9 @@ CF_EXTERN_C_BEGIN
 
 #if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_BSD
 
-#if !TARGET_OS_CYGWIN && !defined(__linux__)
+#if !TARGET_OS_CYGWIN && !defined(__linux__) && !defined(__OpenBSD__)
```
compnerd: Not related to your change, but this is confusing.  I assume that the condition in conjunction with L98 means that this is for Android, macOS, and *BSD but not OpenBSD.  Would you mind rewriting this as such: `#if defined(ANDROID) || (TARGET_OS_BSD && !defined(__OpenBSD__)) || TARGET_OS_MAC`

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
3405691582: That'd conflict with the definition of strtoll_l for Windows (as _strtoi64), though.

CoreFoundation/Base.subproj/CFInternal.h
3405691582: Done. (Used TARGET_OS_ANDROID instead of defined(ANDROID) and reordered the clauses.)

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
compnerd: Okay, I think that localizing the definitions would be nicer in the future.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #2611, opened 2020-01-15
Revert "[test] Split TestFoundation in smaller tests for CTest reporting."

compnerd:
Reverts apple/swift-corelibs-foundation#2521


compnerd:

drodriguez:
I would rather prefer you making the change to your build script than reverting the change, if possible. I suggested a change in the comments of #2521.

PR #2613, opened 2020-01-16
SR-8918: warning: conditional cast from 'Any' to 'AnyObject' always succeeds

spevans:
- The bridging from Any to AnyObject was fixed a while ago so use 'as'
  instead of 'as?' casting and remove the redundant else case.


spevans:
@swift-ci test linux

spevans:
@swift-ci test and merge

PR #2614, opened 2020-01-16
SR-12036: Incorrect and Inconsistent NumberFormatter currency behavior on Linux

spevans:
- For numberStyle .currency, when .currencyCode was set but
  .currencySymbol was nil, the .locale.currencySymbol was used before
  the .currencyCode.


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

PR #2622, opened 2020-01-21
[5.2] SR-12036: Incorrect and Inconsistent NumberFormatter currency behavior on Linux

spevans:
- For numberStyle .currency, when .currencyCode was set but
  .currencySymbol was nil, the .locale.currencySymbol was used before
  the .currencyCode.

(cherry picked from commit 187d6b37249459b6353079564143200c2a7c0900)


spevans:
@swift-ci test

millenomi:
Approved.

PR #2612, opened 2020-01-16
[CF] Expose mach_absolute_time for BSD.

3405691582:
This is exposed for Linux, but also is needed for linking on BSD.

(Let me know if you think there should be one definition of mach_absolute_time in Prefix and the body ifdef'd between Windows and Linux/BSD. I thought to include that in this PR, but I'm not sure I like the look of it, so I omitted it. If you think it'd be worthwhile though, I'm happy to roll it in this PR.)

PR #2623, opened 2020-01-21
Base: reorder includes for Windows (NFCI)

compnerd:
This reorders the Windows header includes for `LARGE_INTEGER` after
PR #2612.

### DIFF
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -68,6 +68,33 @@ typedef char * Class;
 #include <pthread.h>
 #endif
 
+#if TARGET_OS_WIN32
+#define BOOL WINDOWS_BOOL
+
+#define MAXPATHLEN MAX_PATH
+#undef MAX_PATH
+#undef INVALID_HANDLE_VALUE
+
+#define WIN32_LEAN_AND_MEAN
+
+#ifndef WINVER
+#define WINVER  0x0601
+#endif
+
+#ifndef _WIN32_WINNT
+#define _WIN32_WINNT 0x0601
+#endif
+
+// The order of these includes is important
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @3405691582

compnerd:

compnerd:
Windows CI: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=24734&view=results

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -68,6 +68,33 @@ typedef char * Class;
...
+
+#ifndef _WIN32_WINNT
+#define _WIN32_WINNT 0x0601
+#endif
+
+// The order of these includes is important
+#define FD_SETSIZE 1024
+#include <winsock2.h>
+#include <windows.h>
+#include <time.h>
```
3405691582: Don't forget time.h is included adjacent to mach_absolute_time.

3405691582:
Optionally, it might be nice to bring up the boolean_t changes below into this block, but that might also be better to do in a separate PR sometime later down the line.

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
compnerd: That needed to be shuffled due to the inclusion of `Windows.h`.

3405691582: What I mean is: can you omit it here again, considering now time.h is included a second time below windows.h, or is there some other reason preventing that?

compnerd: oh, I see; I'll do that as a follow up.  The time.h was moved since it was in the block.

PR #2626, opened 2020-01-22
[CF] Avoid using malloc introspection.

3405691582:
malloc introspection, as afforded by `malloc_size`, `malloc_usable_size`, or `_msize`, is nonportable (and thus not guaranteed to be present on all platforms), not guaranteed to be accurate (though certainly an upper bound), and is generally recommended for debugging.

For those reasons, the following commits are introduced in this PR:
* CFBasicHash.c: only expose `CFBasicHashGetSize`'s `total` branch when one or both of the debugging symbols `ENABLE_MEMORY_COUNTERS` or `ENABLE_DTRACE_PROBES` are true
* CFStorage.c: remove apparently unused function `__CFStorageEstimateTotalAllocatedSize`. If this is necessary for something, please let me know what constraints are required to avoid exposing the `malloc_size`call.

### DIFF
--- a/CoreFoundation/Collections.subproj/CFBasicHash.c
+++ b/CoreFoundation/Collections.subproj/CFBasicHash.c
@@ -1442,6 +1442,7 @@ CF_PRIVATE size_t CFBasicHashGetSize(CFConstBasicHashRef ht, Boolean total) {
     if (ht->bits.keys_offset) size += sizeof(CFBasicHashValue *);
     if (ht->bits.counts_offset) size += sizeof(void *);
     if (__CFBasicHashHasHashCache(ht)) size += sizeof(uintptr_t *);
+#if ENABLE_MEMORY_COUNTERS || ENABLE_DTRACE_PROBES
     if (total) {
         CFIndex num_buckets = __CFBasicHashTableSizes[ht->bits.num_buckets_idx];
         if (0 < num_buckets) {
@@ -1451,6 +1452,9 @@ CF_PRIVATE size_t CFBasicHashGetSize(CFConstBasicHashRef ht, Boolean total) {
             if (__CFBasicHashHasHashCache(ht)) size += malloc_size(__CFBasicHashGetHashes(ht));
         }
     }
+#else
+    (void)total;
+#endif
     return size;
 }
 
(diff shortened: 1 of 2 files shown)


spevans:
It seems that the standard library uses wraps `malloc_size()`, `malloc_usable_size()`, `_msize()` so  isnt it assumed that at least one of these is present? Can we not wrap them the same way here?

https://github.com/apple/swift/blob/f1ef08af7cafbec1a98ac21dd89a54764950fcc8/stdlib/public/SwiftShims/LibcShims.h#L103

3405691582:
Yes, the Swift stdlib also uses malloc_size, but that is something of a separate issue[1], because CF ought to be able to be built standalone outside of Swift. 

[1] This is really unfortunate, since it's basically used at the heart of ManagedBuffer.capacity (and somewhere in ContiguousArray also, from memory): it's exposed as part of a public API and thus isn't something that is only exposed during a debugging rebuild like here in this PR. But trying to deal with that in the right way is a PR for another day.

compnerd:
CC: @parkera @millenomi 

I believe that `__CFStorageEstimateTotalAllocatedSize` is a debugging aid, meant to be invoked from the debugger by a developer.  The user-visible aspect of this is that the description of the `CFBasicHash` will now show the same values for `size` and `total_size`, which I don't believe to be something that they should be parsing.

compnerd:
@swift-ci please test

compnerd:
@3405691582 would you mind updating the commit message for the second change to remove the statement that the function is vestigial, as I believe it to be a debugger-only debugging aid function.

compnerd:

3405691582:
Actually, after looking closer at `__CFStorageEstimateTotalAllocatedSize`, this may be more complicated: as it's a static function, this function's code is not necessarily emitted; perhaps some concoction of compiler flags (that I'm not even confident exists) may be necessary to get that function to be emitted in order to be invoked by a debugger.

On the one hand, that's good, because the symbol reference isn't actually emitted here, _but_ it does cause a compiler warning when `malloc_size` isn't present, so let's clean up and remove this function anyway.

I've therefore reworded that commit accordingly.

compnerd:
It should be preserved at `-O0`, its meant for debugging :)  I believe that the commit message is reasonable now.  Thanks for fixing that.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #2630, opened 2020-01-25
Revert "Implemented _getFileSystemRepresentation for Windows"

compnerd:
Reverts apple/swift-corelibs-foundation#2366

@gmittert, I think that this is causing the tests to fail


compnerd:

PR #2631, opened 2020-01-28
SR-6450: Leak in Process on Linux

spevans:
- The runLoopSource was holding a reference to self creating a retain
  cycle. Set to nil after the child process has finished.


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

PR #2632, opened 2020-01-28
[CF] One malloc_default_zone macro for non-macOS.

3405691582:
malloc_default_zone is unavailable outside of macOS so requires
redefining in Prefix. We have a definition for Linux and Windows, but
this is also required for BSD, so move this to the common Linux, BSD,
and Windows block.


spevans:
@swift-ci test

compnerd:
Thanks for cleaning up some of this duplication!

PR #2633, opened 2020-01-29
[CF] Locale symbol aliases required for BSD.

3405691582:
These symbol aliases need to be exposed for TARGET_OS_BSD otherwise
undefined symbol errors will occur when linking.


compnerd:
@swift-ci please test

PR #2635, opened 2020-01-29
[CF] Apply missing includes and cleanups.

3405691582:
Here we apply a few cursory cleanups for porting purposes. These changes
are as follows:

  * move a TARGET_OS_CYGWIN block: it was nested under TARGET_OS_MAC,
    which is probably not intended.

  * include sys/socket.h for TARGET_OS_BSD: which should be provided.

  * include netinet/in.h for !TARGET_OS_WIN32: which is required for
    struct sockaddr_in; from memory, this is included transitively on
    Linux, but not elsewhere.

  * use TARGET_OS_BSD instead of TARGET_OS_UNIX: TARGET_OS_UNIX appears
    no longer used; TARGET_OS_BSD is a synonym for __unix__ anyway.

  * disuse pthread_setname_np for TARGET_OS_BSD: this doesn't appear in
    the FreeBSD manuals, it's not available on OpenBSD, so let's just
    disuse it.

### DIFF
--- a/CoreFoundation/RunLoop.subproj/CFSocket.c
+++ b/CoreFoundation/RunLoop.subproj/CFSocket.c
@@ -942,14 +942,15 @@ Boolean __CFSocketGetBytesAvailable(CFSocketRef s, CFIndex* ctBytesAvailable) {
 #include <sys/un.h>
 #include <libc.h>
 #include <dlfcn.h>
-#if TARGET_OS_CYGWIN
-#include <sys/socket.h>
 #endif
+#if TARGET_OS_CYGWIN || TARGET_OS_BSD
+#include <sys/socket.h>
 #endif
 #if TARGET_OS_WIN32
 #include <WinSock2.h>
 #else
 #include <arpa/inet.h>
+#include <netinet/in.h>
 #endif
 #if !TARGET_OS_WIN32
 #include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

CoreFoundation/RunLoop.subproj/CFSocket.c
```diff
@@ -2174,9 +2175,9 @@ manageSelectError()
 
 static void *__CFSocketManager(void * arg)
 {
-#if (TARGET_OS_LINUX && !TARGET_OS_CYGWIN) || TARGET_OS_BSD
+#if (TARGET_OS_LINUX && !TARGET_OS_CYGWIN)
     pthread_setname_np(pthread_self(), "com.apple.CFSocket.private");
-#elif TARGET_OS_CYGWIN
+#elif TARGET_OS_CYGWIN || TARGET_OS_BSD
```
compnerd: You didn't introduce this, but, would you mind rewriting this as:
```
#elif !TARGET_OS_CYGWIN && !TARGET_OS_BSD
```
and drop the empty clause instead?

CoreFoundation/RunLoop.subproj/CFSocket.c
```diff
@@ -2174,9 +2175,9 @@ manageSelectError()
 
 static void *__CFSocketManager(void * arg)
 {
-#if (TARGET_OS_LINUX && !TARGET_OS_CYGWIN) || TARGET_OS_BSD
+#if (TARGET_OS_LINUX && !TARGET_OS_CYGWIN)
```
compnerd: Would you mind dropping the parenthesis?

CoreFoundation/RunLoop.subproj/CFSocket.c
3405691582: Done.

CoreFoundation/RunLoop.subproj/CFSocket.c
3405691582: Done.

spevans:
@swift-ci test

PR #2637, opened 2020-01-31
[POSIX] readdir_r() deprectated, replace with readdir()

spevans:
- readdir_r() may be unsafe where NAME_MAX is unspecified as the
  buffer length isnt passed in. This may cause NUL termination issues.

- glibc >= 2.24 deprecates readdir_r()

- POSIX.1-2008 doesnt require readdir() to be thread-safe but, in glibc,
  calls with different directory streams are.

- Callers of readdir() using methods need to ensure synchronisation
  if the same directory is read by multiple threads.


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

PR #2644, opened 2020-02-03
SR-8994: DateFormatter returns UTC for GMT timezone

spevans:
- This looks to be fixed by the latest upgrade to ICU.

- Re-enable other TestDateFormatter tests that were disabled on Linux
  by setting the Locale to "en_US_POSIX" rather than it being set to
  the locale where the tests were run.


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

PR #2646, opened 2020-02-04
[CF] Preliminary support for standalone builds.

3405691582:
This commit makes some preliminary fixes for standalone builds using
cmake. The standalone and Swift builds are similar, except for:
  * the CoreFoundation umbrella header,

  * ForSwiftFoundationOnly.h,

  * CFXMLParser.c and CFXMLNode.c, since these have CFTypeIDs that are
    present in CFRuntime_Internal.h behind DEPLOYMENT_RUNTIME_SWIFT, and

  * CFXMLInterface, since it refers to struct _NSCFXMLBridge, and this is
    defined in ForSwiftFoundationOnly.h.

Not present is a mechanism to refer to Block.h or dependent libraries. 

Problems including TargetConditionals.h is dealt with in #2653.

(I'm not well-versed in cmake-speak, so if there's more idiomatic or stylistic ways of doing this, please let me know.)


CoreFoundation/CMakeLists.txt
```diff
@@ -373,7 +377,8 @@ add_framework(CoreFoundation
                 URL.subproj/CFURLAccess.c
                 URL.subproj/CFURL.c
                 URL.subproj/CFURLComponents.c
-                URL.subproj/CFURLComponents_URIParser.c)
+                URL.subproj/CFURLComponents_URIParser.c
+                ${CF_SWIFT_SOURCES})
```
compnerd: Please use `target_sources` for adding the additional source file.

CoreFoundation/CMakeLists.txt
```diff
@@ -512,10 +525,13 @@ if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
...
 endif()
 
+if(CF_DEPLOYMENT_SWIFT)
+  set(CF_SWIFT_TARGETS CFXMLInterface)
+endif()
 install(TARGETS
           CoreFoundation
           CFURLSessionInterface
-          CFXMLInterface
+          ${CF_SWIFT_TARGETS}
```
compnerd: I think it may be nicer to just use a property that you append to, rather than doing the variable expansion here.

3405691582: I may be misunderstanding something. Do you mean use set_property() for something? If so, I'm not quite sure how you mean to apply it here.

For the moment, I've pulled out CFXMLInterface into a separate install(), which avoids the variable expansion. Maybe this is better too?

CoreFoundation/CMakeLists.txt
3405691582: Done.

3405691582:
I've also added some indents appropriately where there were missing before.

spevans:
@swift-ci test linux

CoreFoundation/CMakeLists.txt
compnerd: Im basically trying to avoid the whole variable soup that CMake <3.0 used to suffer from.  Adding a second `install(TARGETS)` in an `if` is another approach.

millenomi:
@swift-ci please test

shahmishal:

PR #2647, opened 2020-02-05
Feedback Assistant wording and link

AustinConlon:
Not sure if the wording is better, but the link should probably be updated to https://feedbackassistant.apple.com instead of https://bugreport.apple.com so that it goes directly to it rather than a page explaining Apple's update of the bug reporting system.


xwu:

xwu:
@swift-ci Test macOS platform

PR #2655, opened 2020-02-07
SR-11917: Using JSONEncoder's .outputFormatting = .sortedKeys has a memory leak on Linux

spevans:
- Any use of NSString.compare(_:options:range:locale:) with a non-nil locale
  would leak on Linux.

- _CFCompareStringsWithLocale was leaking the collator on non-macOS
  platforms.

- Use the collator cache on Linux and Win32 the same way that macOS does
  and for other platforms ensure that any allocated collator is freed.

### DIFF
--- a/CoreFoundation/String.subproj/CFStringUtilities.c
+++ b/CoreFoundation/String.subproj/CFStringUtilities.c
@@ -420,7 +420,7 @@ static UCollator *__CFStringCopyDefaultCollator(CFLocaleRef compareLocale) {
     return collator;
 }
 
-#if TARGET_OS_MAC
+#if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_LINUX
 static void __collatorFinalize(UCollator *collator) {
     CFLocaleRef locale = _CFGetTSD(__CFTSDKeyCollatorLocale);
     _CFSetTSD(__CFTSDKeyCollatorUCollator, NULL, NULL);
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

CoreFoundation/String.subproj/CFStringUtilities.c
```diff
@@ -603,9 +603,8 @@ CF_PRIVATE CFComparisonResult _CFCompareStringsWithLocale(CFStringInlineBuffer *
     if (range2.location > 0) {
 	range2.location = __extendLocationBackward(range2.location - 1, str2, nonBaseBMP, punctBMP);
     }
-    
+
 #if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_LINUX
-#if TARGET_OS_MAC
```
3405691582: This exposes the variable collator at l616, which is still hidden behind an #if tag above (l577), so this doesn't build properly on other platforms.

spevans: Thanks for catching this. Do you think the the `#if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_LINUX` could be removed entirely and the code would work on the BSD platforms? btw which platforms are you testing on?

spevans:
@swift-ci test

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

CoreFoundation/String.subproj/CFStringUtilities.c
3405691582: Thanks, that builds cleanly now.

The platform-specific branches could be removed --  I did a rush check adding TARGET_OS_BSD on every appearance of TARGET_OS_LINUX atop this commit -- but they'd need to be removed in more places than you are changing here, so perhaps that should be done in a separate PR.

However, I haven't run any of the tests; right now I am just checking to ensure things build properly on OpenBSD without error.

spevans: Thanks for checking, yes probably best to support other platforms in a future PR.

spevans:
@compnerd could you test this on windows when you get a chance? thanks

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

spevans:
@swift-ci test and merge

PR #2663, opened 2020-02-14
CoreFoundation: handle `isDirectory` properly in `WindowsPathToURLPath`

compnerd:
In the case that we were given a non-directory path terminating in `\`,
we would preserve the last path arc constructing a directory path
instead of a file path.  This manifested as test failure in the Swift
test suite when converting paths.


compnerd:

compnerd:
Windows CI:
- Facebook: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=26303&view=results
- Google: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=26304&view=results

PR #2664, opened 2020-02-14
TestFoundation: correct assumption on Windows

compnerd:
This corrects the expectation on Windows.  This was obviously incorrect
previously as:

1. `URL(fileURLWithPath: "S:\\b")` would provide a *file* URL
2. `URL(fileURLWithPath: "eh", relativeTo: ...)` would then construct a
   path relative to the file, thus, the final path would be `S:\eh`.

This matches the reference implementation as demonstrated by:
```
// %clang %s -framework Foundation
int main() {
  NSLog(@"result: %@",
        [[NSURL fileURLWithPath:@"eh"
                  relativeToURL:[NSURL fileURLWithPath:@"/tmp/b"]]
         absoluteString]);
  return 0;
}
```


compnerd:

compnerd:
Ugh, this is wrong :-(

PR #2665, opened 2020-02-15
Foundation: slight cleanup of `FileManager+Win32` (NFC)

compnerd:
Use a `try?` with a coalescing operator to avoid a guard + return x 2.


compnerd:

PR #2667, opened 2020-02-15
Foundation: clear up string abuse on Windows

compnerd:
Use the file system representation of the path rather than encoding the
raw path as a UTF16 string.


compnerd:

PR #2668, opened 2020-02-15
SR-4280: Re-enable TestNotificationQueue tests

spevans:
- These tests work correctly on Linux and Darwin now so re-enable.

Reenabling to see if the underlying timeout issue has disappeared


spevans:
@swift-ci test

PR #2669, opened 2020-02-15
FileManager: cleanup string abuse in FileManager+Win32

compnerd:
Use the file system representation instead of a raw path string encoded
as a UTF16 string.  This ensures that we get valid paths.


compnerd:

PR #2671, opened 2020-02-16
build: reapir the bundling of the OLSON database on Windows

compnerd:
We package the OLSON database as a binary resource in Foundation.dll.
Unfortunately, something has caused a regression and OBJECT linking
does not function even with CMake 3.16.  Explicitly include the
objects from the resource.


compnerd:

compnerd:

compnerd:

PR #2674, opened 2020-02-16
Foundation: correct error code for invalid symlink on Windows

compnerd:
Change the path that is used for the `NSError` as the invalid input is
material to the calculation of the proper error code.  When the
unresolved path is empty, the error code should be
`.fileReadInvalidFileName` rather than `.fileReadNoSuchFile`.  This
now matches the expectation.


compnerd:

PR #2675, opened 2020-02-16
Foundation: remove file system assumptions

compnerd:
The tests assumed that `S:\b` would exist, which is not required to
exist.  Explicitly indicate the type of path (directory vs
non-directory) as we cannot stat the path to check if the path is a file
or directory.


compnerd:

PR #2679, opened 2020-02-19
[CF] Unify implementations of _CFProcessPath.

3405691582:
Firstly, `_CFProcessPath` has no implementation for TARGET_OS_BSD. Indeed,
there is no portable implementation of this action across multiple
platforms. As an initial implementation for TARGET_OS_BSD, use the
environment to help us. Some shells set the environment variable "_" to
the path used to invoke the binary. This mirrors the implementation for
TARGET_OS_OSX, so add the same conditionals on `__CFProcessIsRestricted`.

Relying on the shell has drawbacks, including that the shell doesn't set
the full path to the current executable if it isn't part of the PATH
(or, of course, specified explicitly). Perhaps on some other platforms
there is some miscellaneous kernel trickery that could be done, but
let's defer that to the future and keep getenv("_") as a last resort.

Second, given that, we should bring the implementations together (like
we've done elsewhere), not to mention that `_CFProcessPath` is fundamental
to other parts of CF. This highlights the repetition in each
implementation, so here we also factor off some of it and share it
across the implementations to facilitate code reuse.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -113,18 +113,23 @@ const char **_CFGetProcessPath(void) {
     return &__CFProcessPath;
 }
 
-#if TARGET_OS_WIN32
+static inline void _CFSetProgramNameFromPath(const char *path) {
+    __CFProcessPath = strdup(path);
+    __CFprogname = strrchr(__CFProcessPath, PATH_SEP);
+    __CFprogname = (__CFprogname ? __CFprogname + 1 : __CFProcessPath);
+}
+
 const char *_CFProcessPath(void) {
     if (__CFProcessPath) return __CFProcessPath;
+
+#if TARGET_OS_WIN32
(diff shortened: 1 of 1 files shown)


3405691582:
@compnerd This is mainly a mechanical change, but could you double check this on the win32 buildbot please?

compnerd:
https://dev.azure.com/compnerd/swift-build/_build/results?buildId=26671&view=results

millenomi:
@swift-ci please test

3405691582:
macOS test failure looks unrelated, since it seems to be in xctest and that it has been failing in the same fashion since at least [build 2072 on March 30](https://ci.swift.org/job/swift-corelibs-foundation-PR-macOS/2072/).

compnerd:
@3405691582 this needs to be rebased as there is now a conflict.

3405691582:
Rebase completed.

spevans:
@swift-ci please test

spevans:
@swift-ci please test

PR #2680, opened 2020-02-19
[CF] TARGET_OS_BSD for FS representation behavior.

3405691582:
CFURLGetFileSystemRepresentation's implementation has TARGET_OS_MAC and
TARGET_OS_LINUX, but not TARGET_OS_BSD. Nothing here appears macOS or
Linux specific, and things seem to work okay with TARGET_OS_BSD, so
enable it here too.

This also requires CanonicalFileURLStringToFileSystemRepresentation to
be exposed.

### DIFF
--- a/CoreFoundation/URL.subproj/CFURL.c
+++ b/CoreFoundation/URL.subproj/CFURL.c
@@ -4064,7 +4064,7 @@ static CFStringRef URLPathToPOSIXPath(CFStringRef path, CFAllocatorRef allocator
     return result;
 }
 
-#if TARGET_OS_MAC || TARGET_OS_LINUX
+#if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_BSD
 static Boolean CanonicalFileURLStringToFileSystemRepresentation(CFStringRef str, UInt8 *inBuffer, CFIndex inBufferLen)
 {
     size_t fileURLPrefixLength;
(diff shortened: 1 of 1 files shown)


CoreFoundation/URL.subproj/CFURL.c
```diff
@@ -4534,13 +4534,12 @@ CFStringRef CFURLCreateStringWithFileSystemPath(CFAllocatorRef allocator, CFURLR
...
 Boolean CFURLGetFileSystemRepresentation(CFURLRef url, Boolean resolveAgainstBase, uint8_t *buffer, CFIndex bufLen) {
-#if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_WIN32
     CFAllocatorRef alloc = CFGetAllocator(url);
     CFStringRef path;
 
     if (!url) return false;
-#endif
-#if TARGET_OS_MAC || TARGET_OS_LINUX
+
+#if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_BSD
```
spevans: Would it be better to make this an `#if !TARGET_OS_WIN32` else this function will always return `false` when run on something that isn't a listed platform?

3405691582: CF only mentions `kCFURLPOSIXPathStyle` or `kCFURLWindowsPathStyle`(`kCFURLHFSPathStyle` is deprecated). So I think in this case your suggestion makes sense to make this a binary conditional.

However, we should probably invert sense in that case: TARGET_OS_WIN32 comes first, and POSIX is an `#else`.

spevans:
@wift-ci test linux

3405691582:
Looks like you typo'd the @ there :)

compnerd:

3405691582:
Can someone re-run this? The failure in the run looks unrelated.

xedin:

compnerd:
I think that this is ready, lets get it merged.

PR #2681, opened 2020-02-20
GenerateTestFixtures: Fix building on Xcode

spevans:
- Add a dependancy on SwiftFoundation.

- Fix import to use SwiftFoundation if available, not Foundation.


spevans:
@swift-ci test linux

PR #2683, opened 2020-02-21
build: use forced loading for FoundationNetworking

compnerd:
Since FoundationNetworking symbols are lazily resolved at runtime,
no symbol references to it are emitted.  On PE/COFF targets (i.e.
Windows), this will result in FoundationNetworking not being loaded and
linked against.  Use the force load symbol to ensure that the reference
is preserved.


compnerd:

compnerd:

millenomi:

PR #2684, opened 2020-02-22
test: correct test assumption for Windows paths

compnerd:
The Windows file system path representation `\` is a drive letter
relative path.  That is, the same path can be used across multiple
drives (similar to the concept of alternate roots).  We expect that
the path returned is `[[:letter:]]:` or `[[:letter:]]:[\\/]` rather
than the Unix `/`.  Adjust the expectation.


compnerd:

PR #2685, opened 2020-02-22
tests: disable a test on Windows

compnerd:
Windows does not have the `PWD` environment variable.  The closest
analogue is the `%CD%` magic variable which is specific to the shell
rather than the environment.  Skip the test on Windows.


compnerd:

PR #2686, opened 2020-02-22
TimeZone: delegate to ICU on Windows

compnerd:
Unfortunately, Windows does not have full time zone information
available.  For many cases we can make do with the system information,
augmented with alternate names (the Olson-to-Windows conversion
database).  This papers over the inability to deserialize the full
information in the registry by falling back to ICU to get the
information for the DST time conversion.

### DIFF
--- a/CoreFoundation/NumberDate.subproj/CFTimeZone.c
+++ b/CoreFoundation/NumberDate.subproj/CFTimeZone.c
@@ -1648,13 +1648,32 @@ CFTimeInterval CFTimeZoneGetSecondsFromGMT(CFTimeZoneRef tz, CFAbsoluteTime at) 
     return __CFTZPeriodGMTOffset(&(tz->_periods[idx]));
 }
 
+extern UCalendar *__CFCalendarCreateUCalendar(CFStringRef calendarID, CFStringRef localeID, CFTimeZoneRef tz);
+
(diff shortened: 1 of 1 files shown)


CoreFoundation/NumberDate.subproj/CFTimeZone.c
```diff
@@ -1679,18 +1679,34 @@ CFTimeInterval CFTimeZoneGetDaylightSavingTimeOffset(CFTimeZoneRef tz, CFAbsolut
     return 0.0;
 }
 
+extern UCalendar *__CFCalendarCreateUCalendar(CFStringRef calendarID, CFStringRef localeID, CFTimeZoneRef tz);
```
millenomi: should this be CF_PRIVATE? what's the tag on that function?

compnerd: I'm not sure, it was just moved up from below in the same file.  Pretty sure its meant to be private.

millenomi:
Otherwise approved.

millenomi:

CoreFoundation/NumberDate.subproj/CFTimeZone.c
millenomi: My bad, I just saw the removal under it.

PR #2687, opened 2020-02-22
SR-10512: TestFileManager .creationDate failed on Ubuntu 14.04

spevans:
- .creationDate may not be available on all Linux platforms due to
  kernel version, Docker or other environment security settings etc
  so only check its value if it is non-nil.

- Re-enable this test but allow for .creationDate possibly being nil.


spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

PR #2688, opened 2020-02-22
SR-5576: Calendar.dateComponents(_:from:) crashes using .nanosecond or .quarter

spevans:
- .nanosecond is implemented so re-enable tests.

- .quarter doesnt crash but always returns 0 for now so also re-enable
  in tests.


spevans:
@swift-ci test linux

PR #2690, opened 2020-02-22
Port, PortMessage: Rename sendBeforeDate(...) to send(before: Date...)

spevans:
- Update method signatures to match Darwin.


spevans:
@swift-ci test linux

PR #2691, opened 2020-02-22
Remove old alias _$S15SwiftFoundation19_NSCFConstantStringCN

spevans:
- This is an alias for _$s15SwiftFoundation19_NSCFConstantStringCN
  used while the mangling was being switched from $S to $s but is
  not required anymore.


spevans:
@swift-ci test

PR #2695, opened 2020-02-23
[5.1] SR-6450: Leak in Process on Linux

spevans:
- The runLoopSource was holding a reference to self creating a retain
  cycle. Set to nil after the child process has finished.

(cherry picked from commit b15e4f43a8de844c44070a5987727ef1b864a426)

Back port of #2631 5.2 version is #2694


spevans:
@swift-ci test linux

spevans:
cc @weissi

PR #2697, opened 2020-02-23
[5.1] Process: Setting stdout or stderr to nil or nullDevice set stdin instead.

spevans:
- If .standardOutput or .standardError was set to nil or
  FileHandle.nullDevice, stdin was set to /dev/null instead.

(cherry picked from commit dffe77420524dbd980edefeca93a0223274d49d7)

Backport of #2545 

cc @weissi


spevans:
@swift-ci test

PR #2700, opened 2020-02-27
Return success from `FileManager` recursive directory creation even if a directory was created by another thread concurrently.

fumoboy007:
If another thread creates one of the same directories in the target path, the `!fileExists` check may pass while the `mkdir` call fails with `EEXIST`. However, if the existing file is a directory, then that should still be a success.

Fixes [SR-12272](https://bugs.swift.org/browse/SR-12272).

### DIFF
--- a/Sources/Foundation/FileManager+POSIX.swift
+++ b/Sources/Foundation/FileManager+POSIX.swift
@@ -346,8 +346,16 @@ extension FileManager {
                         try createDirectory(atPath: parent, withIntermediateDirectories: true, attributes: attributes)
                     }
                     if mkdir(pathFsRep, S_IRWXU | S_IRWXG | S_IRWXO) != 0 {
-                        throw _NSErrorWithErrno(errno, reading: false, path: path)
-                    } else if let attr = attributes {
+                        let posixError = errno
+                        if posixError == EEXIST && fileExists(atPath: path, isDirectory: &isDir) && isDir.boolValue {
+                            // Continue; if there is an existing file and it is a directory, that is still a success.
+                            // There can be an existing file if another thread or process concurrently creates the
+                            // same file.
+                        } else {
+                            throw _NSErrorWithErrno(posixError, reading: false, path: path)
+                        }
+                    }
+                    if let attr = attributes {
                         try self.setAttributes(attr, ofItemAtPath: path)
                     }
                 } else if isDir.boolValue {
--- a/Tests/Foundation/Tests/TestFileManager.swift
+++ b/Tests/Foundation/Tests/TestFileManager.swift
@@ -15,6 +15,8 @@
     #endif
 #endif
 
+import Dispatch
+
 class TestFileManager : XCTestCase {
 #if os(Windows)
     let pathSep = "\\"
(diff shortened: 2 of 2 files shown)


fumoboy007:
@swift-ci Please test.

Tests/Foundation/Tests/TestFileManager.swift
```diff
@@ -1755,6 +1757,49 @@ VIDEOS=StopgapVideos
...
+
+        for directoryURL in directoryURLs {
+            if let directoryURL = directoryURL {
+                try? fileManager.removeItem(at: directoryURL)
+            }
+        }
+
+        for error in errors {
+            if let error = error {
+                throw error
```
spevans: Probably better to use `XCTFail(error)` here to show the error.

spevans:
@swift-ci test linux

Tests/Foundation/Tests/TestFileManager.swift
fumoboy007: Why?

fumoboy007: Ping. Can you elaborate on why `XCTFail` is preferable to throwing the error?

spevans: For functions that can fail when setting up the test, it is appropriate to just throw to exit the exit the entire test as nothing else would be expected to work. However for the main part of the test an `XCT...` allows a more useful output string and shows other people reading the test what is actually being tested.

A good example is `URL.init?(string:)`. This returns an `Optional` but would not be expected to fail if given, say, `http://www.example.com`. So if a `URL` needs to be created at the start of the test then:
```
let url = try XCTUnwrap(URL(string: "http://example.com/"))
```
which will `throw` is fine as something has gone seriously wrong if this fails and the rest of the test is going to break.

However if we are writing tests for `URL.init?(string:)` itself, then the following would be used:
```XCTAssertNotNil(URL(string: "http://example.com/"))```

As that is the function actually being tested.
 Note this isnt an exact rule as sometimes its easier to `throw` when testing certain parts  but in your test above the errors are being caught anyway so using an `XCTFail()` is no extra work.

File: Sources/Foundation/FileManager+POSIX.swift [hunk 2]
```diff
@@ -346,8 +346,16 @@ extension FileManager {
                         try createDirectory(atPath: parent, withIntermediateDirectories: true, attributes: attributes)
                     }
                     if mkdir(pathFsRep, S_IRWXU | S_IRWXG | S_IRWXO) != 0 {
-                        throw _NSErrorWithErrno(errno, reading: false, path: path)
-                    } else if let attr = attributes {
+                        let posixError = errno
+                        if posixError == EEXIST && fileExists(atPath: path, isDirectory: &isDir) && isDir.boolValue {
```
spevans: I think the test for `EEXIST` is only needed on the last path component to check if is a file.

If an earlier `mkdir()` fails because the existing path is a  file there should be an `ENOTDIR` error returned, which should also handle the edge case of `path` being a directory when `fileExists` is called but being replaced by a file before the next `mkdir()`.

Would need an extra test with a pre-existing path with includes a file as the last component to check this.

fumoboy007: `EEXIST` is returned by `mkdir` if the last path component already exists, regardless of whether it is a file or a directory. `ENOTDIR` is returned if any of the path components aside from the last one is not a directory.

This is supported by the [documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/mkdir.html) and by the sample code in [SR-12272](https://bugs.swift.org/browse/SR-12272).

fumoboy007:
@millenomi Is this test job stuck or something? How come it has not run yet?

millenomi:
Sometimes the traffic on CI is so high that requests get dropped.

millenomi:

millenomi:

spevans:
@swift-ci test linux

xwu:
@swift-ci test macOS

fumoboy007:
Resolved the conflict. Please re-trigger the tests.

spevans:
@swift-ci please test

fumoboy007:
Yay, passed! Can we merge?

spevans:

PR #2704, opened 2020-02-27
Removes redundant buffer zeroing from String

valeriyvan:


spevans:
@swift-ci test

spevans:
@swift-ci test linux

PR #2707, opened 2020-02-27
[5.2] Limit the amount of recursion allowed while JSONSerialization parses.

millenomi:
rdar://59688482


millenomi:

PR #2708, opened 2020-03-02
Speeds up NSDecimalString(_:_:) function

valeriyvan:
Speeds up NSDecimalString(_:_:) function by about 25% reordering operations and avoiding redundant buffer zeroing

### DIFF
--- a/Sources/Foundation/Decimal.swift
+++ b/Sources/Foundation/Decimal.swift
@@ -1678,25 +1678,43 @@ public func NSDecimalMultiplyByPowerOf10(_ result: UnsafeMutablePointer<Decimal>
 }
 
 public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> String {
+    let ZERO: UInt8 = 0x30 // ASCII '0' == 0x30
+    let zeroString = String(Unicode.Scalar(ZERO)) // "0"
 
     var copy = dcm.pointee
     if copy.isNaN {
         return "NaN"
     }
     if copy._length == 0 {
-        return "0"
+        return zeroString
     }
 
-    let ZERO: CChar = 0x30 // ASCII '0' == 0x30
-    let DECIMALPOINT: CChar = 0x2e // ASCII '.' == 0x2e
-    let MINUS: CChar = 0x2d // ASCII '-' == 0x2d
-    let bufferSize = 200 // Max value: 39 + 128 + sign + decimal point
-    var buffer = Array<CChar>(repeating: 0, count: bufferSize)
+    let decimalSeparatorReversed: String = {
+        // Short circuiting for common case that `locale` is `nil`
+        guard locale != nil else { return "." } // Defaulting to decimal point
 
-    var i = bufferSize - 1
+        var decimalSeparator: String? = nil
+        if let locale = locale as? Locale {
+            decimalSeparator = locale.decimalSeparator
+        } else if let dictionary = locale as? [NSLocale.Key: String] {
+            decimalSeparator = dictionary[NSLocale.Key.decimalSeparator]
+        } else if let dictionary = locale as? [String: String] {
+            decimalSeparator = dictionary[NSLocale.Key.decimalSeparator.rawValue]
+        }
+        guard let dc = decimalSeparator else { return "." } // Defaulting to decimal point
+        return String(dc.reversed())
+    }()
+
+    let resultCount =
+        (copy.isNegative ? 1 : 0) + // sign, obviously
+        39 + // mantissa is an 128bit, so log10(2^128) is approx 38.5 giving 39 digits max for the mantissa
+        (copy._exponent > 0 ? Int(copy._exponent) : decimalSeparatorReversed.count) // trailing zeroes or decimal separator
+
+    var result = ""
+    result.reserveCapacity(resultCount)
+
     while copy._exponent > 0 {
-        i -= 1
-        buffer[i] = ZERO
+        result.append(zeroString)
         copy._exponent -= 1
     }
 
@@ -1707,44 +1725,25 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
     while copy._length != 0 {
         var remainder: UInt16 = 0
         if copy._exponent == 0 {
-            i -= 1
-            buffer[i] = DECIMALPOINT
+            result.append(decimalSeparatorReversed)
         }
         copy._exponent += 1
         (remainder, _) = divideByShort(&copy, 10)
-        i -= 1
-        buffer[i] = Int8(remainder) + ZERO
+        result.append(String(Unicode.Scalar(ZERO + UInt8(remainder))))
     }
     if copy._exponent <= 0 {
         while copy._exponent != 0 {
-            i -= 1
-            buffer[i] = ZERO
+            result.append(zeroString)
             copy._exponent += 1
         }
-        i -= 1
-        buffer[i] = DECIMALPOINT
-        i -= 1
-        buffer[i] = ZERO
+        result.append(decimalSeparatorReversed)
+        result.append(zeroString)
     }
     if copy._isNegative != 0 {
-        i -= 1
-        buffer[i] = MINUS
-    }
-    let result = String(cString: Array(buffer.suffix(from: i)))
-
-    var decimalSeparator: String?
-    if let locale = locale as? Locale {
-        decimalSeparator = locale.decimalSeparator
-    } else if let dictionary = locale as? [NSLocale.Key: String] {
-        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator]
-    } else if let dictionary = locale as? [String: String] {
-        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator.rawValue]
+        result.append("-")
     }
 
-    if let dcm = decimalSeparator, dcm != "." {
-        return result.replacingOccurrences(of: ".", with: dcm)
-    }
-    return result
+    return String(result.reversed())
 }
 
 private func multiplyBy10(_ dcm: inout Decimal, andAdd extra:Int) -> NSDecimalNumber.CalculationError {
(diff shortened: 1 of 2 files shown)


Sources/Foundation/Decimal.swift
```diff
@@ -1687,64 +1687,74 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
...
+    var decimalSeparator: String!
+    if let locale = locale as? Locale {
+        decimalSeparator = locale.decimalSeparator
+    } else if let dictionary = locale as? [NSLocale.Key: String] {
+        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator]
+    } else if let dictionary = locale as? [String: String] {
+        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator.rawValue]
+    } else {
+        decimalSeparator = "."
+    }
```
spevans: Is `decimalSeparator` guaranteed to be non-`nil` at this point ? The two `dictionary` look ups could fail and return `nil`

Sources/Foundation/Decimal.swift
```diff
@@ -1687,64 +1687,74 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
...
+    let decimalSeparatorCChar = decimalSeparator.utf8CString // should be accessed with withUnsafeBufferPointer
```
spevans: Wont this cause an extra memory allocation to create the `utf8CString`, could  a `UTF8View` be used and  later  iterated byte by byte rather than using the `memcpy` below?

Sources/Foundation/Decimal.swift
```diff
@@ -1687,64 +1687,74 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
...
+    let decimalSeparatorCCharCount = decimalSeparatorCChar.withUnsafeBufferPointer { $0.count } - 1 // without null terminator
+
     let ZERO: CChar = 0x30 // ASCII '0' == 0x30
-    let DECIMALPOINT: CChar = 0x2e // ASCII '.' == 0x2e
     let MINUS: CChar = 0x2d // ASCII '-' == 0x2d
-    let bufferSize = 200 // Max value: 39 + 128 + sign + decimal point
-    var buffer = Array<CChar>(repeating: 0, count: bufferSize)
+    // FIXME: where magic constants 39 and 128 come from?
+    let bufferSize = 39 + 128 + 1 /* sign */ + decimalSeparatorCCharCount + 1 /* null terminator */
```
spevans: The 39 is from the mantissa being an 128bit, so log10(2^128) is approx 38.5 giving 39 digits max for the mantissa.
128 is the exponent being an `Int8`

Sources/Foundation/Decimal.swift
```diff
@@ -1687,64 +1687,74 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
...
-    while copy._exponent > 0 {
-        i -= 1
-        buffer[i] = ZERO
-        copy._exponent -= 1
-    }
-
-    if copy._exponent == 0 {
-        copy._exponent = 1
-    }
+    let buffer = Array<CChar>(unsafeUninitializedCapacity: bufferSize) { buf, initializedCount in
```
spevans: I wonder if this current algorithm of allocating an array, then adding all of the digits to it backwards and then constructing a string from the result is slightly more than necessary. A better alternative might be to just use a `var result = ""`, append the digits and then return `result.reverse()`.

 If a decent heuristic can be found to estimate the number of output characters then `reserveCapacity()` can be called to just do one allocation. Then in the best case where the result fits into a short string no heap allocation will occur.

Sources/Foundation/Decimal.swift
```diff
@@ -1687,64 +1687,74 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
...
-    while copy._length != 0 {
-        var remainder: UInt16 = 0
         if copy._exponent == 0 {
+            copy._exponent = 1
+        }
+
+        while copy._length != 0 {
+            var remainder: UInt16 = 0
+            if copy._exponent == 0 {
+                i -= decimalSeparatorCCharCount
```
spevans: Since the `decimalSeparator` is not always required, the initial work earlier in the function to find it could be moved into a scoped function and then only called if needed, eg could add a line above this once with:
```
let (decimalSeparatorCChar , decimalSeparatorCCharCount) = decimalSeparator()
```
Although I would probably still use the `UTF8View` as I think it might be quicker.

Sources/Foundation/Decimal.swift
```diff
@@ -1687,64 +1687,74 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
...
-        decimalSeparator = locale.decimalSeparator
-    } else if let dictionary = locale as? [NSLocale.Key: String] {
-        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator]
-    } else if let dictionary = locale as? [String: String] {
-        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator.rawValue]
+        if copy._isNegative != 0 {
+            i -= 1
+            buf[i] = MINUS
+        }
+        memset(buf.baseAddress, 0, i) // FIXME: this could be avoided as well
```
spevans: Unfortunately the shorter the final result the more bytes here are getting initialised but not really needed. Having a heuristic that tries to overestimate the final string length would help reduce the work done here.

Sources/Foundation/Decimal.swift
```diff
@@ -1699,68 +1699,49 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
     if decimalSeparator == nil {
         decimalSeparator = "."
     }
+    let decimalSeparatorReversed = String(decimalSeparator.reversed())
 
-    let decimalSeparatorUtf8View = decimalSeparator.utf8
-
-    let ZERO: CChar = 0x30 // ASCII '0' == 0x30
-    let MINUS: CChar = 0x2d // ASCII '-' == 0x2d
+    let DIGITS: [String] = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"]
```
spevans: I wouldnt use and array as it will take time to construct including allocating memory on the heap. Digit lookup could be replaced with:
```
let ZERO = UInt8(ascii: "0")

// Append '0'
str.append(String(Unicode.Scalar(ZERO)))

// Append '1'
str.append(String(Unicode.Scalar(ZERO +i )))
```

etc

Sources/Foundation/Decimal.swift
```diff
@@ -1699,68 +1699,49 @@ public func NSDecimalString(_ dcm: UnsafePointer<Decimal>, _ locale: Any?) -> St
...
 
     // The 39 is from the mantissa being an 128bit, so log10(2^128) is approx 38.5 giving 39 digits max for the mantissa.
     // 128 is the exponent being an Int8.
     // 1 is for sign.
-    // The last 1 is for null terminator.
-    let bufferSize = 39 + 128 + 1 + decimalSeparatorUtf8View.count + 1
+    let maxResultCount = 39 + 128 + 1 + decimalSeparatorReversed.count
```
spevans: This can be improved slightly. Although the digit count (39) isn't easily calculable
 the `+1` for the minus sign is known if its needed.
 If the exponent is >=0, then the `128` can be replaced directly with exponent since it is just the number of trailing zeros and there is now need for the `decimalSeparatorReversed`

valeriyvan:
fixed

Sources/Foundation/Decimal.swift
```diff
@@ -1678,25 +1678,41 @@ public func NSDecimalMultiplyByPowerOf10(_ result: UnsafeMutablePointer<Decimal>
...
+    } else if let dictionary = locale as? [NSLocale.Key: String] {
+        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator]
+    } else if let dictionary = locale as? [String: String] {
+        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator.rawValue]
+    }
+    // Defaulting to decimal point
+    if decimalSeparator == nil {
+        decimalSeparator = "."
+    }
+    let decimalSeparatorReversed = String(decimalSeparator.reversed())
```
spevans: I think the last fix we need is to optimise for the common case that `locale` is `nil` and also not to bother `reverse()`ing the decimal point, this may be a better way to go:

```
let decimalSeparatorReversed: String = {
    guard locale != nil else { return "." }

    var decimalSeparator: String? = nil
    if let locale = locale as? Locale {
        decimalSeparator = locale.decimalSeparator
    } else if let dictionary = locale as? [NSLocale.Key: String] {
        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator]
    } else if let dictionary = locale as? [String: String] {
        decimalSeparator = dictionary[NSLocale.Key.decimalSeparator.rawValue]
    }
    if let ds = decimalSeparator { return String(ds.reversed()) }
    // Defaulting to decimal point
    return "."
}()
```

Also worth adding a test with a mutli-character `locale` to test it is added to the string correctly, using a locale like the following should be sufficient:
```
let locale: [NSLocale.Key: String] = [NSLocale.Key.decimalSeparator: "()"]
```

valeriyvan:
Fixed decimal point lookup as suggested.
Test for mutli-character decimal separator already exists.

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

spevans:
This looks ok now, could you just rebase and squash all of the commits into one commit. Also, did you run any benchmarks on this final version to see if it is any faster?

valeriyvan:
Squashed.
The last version is still about 25% faster then version before fix.
I can't see the last fix to be simnifically faster then any intermediate version.

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2709, opened 2020-03-02
CoreFoundation: ensure that the bundle path exists

compnerd:
On Windows, we would assume that the bundle path computed will exist.
However, it is possible to have an executable using Foundation without a
bundle.  In such a case, the main bundle would never be setup, and when
SwiftFoundation would attempt to access it, it would assume that it was
setup.  This would cause a failure due to the forced unwrapping.


compnerd:

compnerd:
Windows CI: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=27369&view=results

PR #2710, opened 2020-03-03
Compact number formats for NumberFormatter.

AlanQuatermain:
Implements compact decimal number formatting in short and long forms.

In **en_US**:

| Input                   | Short Output   | Long Output  |
| ----------------------- | -------------- | ------------ |
| 1000                    | 1K             | 1 thousand   |
| 1100                    | 1.1K           | 1.1 thousand |
| 1_100_000               | 1.1M           | 1.1 million  |
| 1_100_000_000           | 1.1B           | 1.1 billion  |
| 1_100_000_000_000       | 1.1T           | 1.1 trillion |

In **ja_JP** (NB: groups by 4):

| Input                   | Short Output   | Long Output |
| ----------------------- | -------------- | ----------- |
| 10_000                  | 1万            | 1万         |
| 11_000                  | 1.1万          | 1.1万       |
| 110_000_000             | 1.1億          | 1.1億       |
| 1_100_000_000_000       | 1.1兆          | 1.1兆       |

This doesn't match macOS/iOS, but it's something of a glaring omission (especially since I've had to implement it by hand for all locales on top of `NSNumberFormatter` recently). The ICU implementation underneath CoreFoundation provides this support, and CLDR includes all the relevant data. Thus I've made a couple of small changes to enable access to this functionality through `CFNumberFormatter` and `NumberFormatter`. It is my sincere hope that someone at Apple will go "hm, that's interesting," and implement it in macOS 10.16 and so on.

This implementation makes use of the `UNUM_DECIMAL_COMPACT_SHORT` and `UNUM_DECIMAL_COMPACT_LONG` from `unicode/unum.h`. Corresponding constants have been defined for CoreFoundation and Foundation:

| ICU                        | CoreFoundation                              | Foundation                          |
| -------------------------  | ------------------------------------------- | ----------------------------------- |
| UNUM_DECIMAL_COMPACT_SHORT | kCFNumberFormatterDecimalCompactStyle       | NumberFormatter.Style.compact       |
| UNUM_DECIMAL_COMPACT_LONG  | kCFNumberFormatterDecimalCompactPluralStyle | NumberFormatter.Style.compactPlural |

Using these options will have some slight effects on default property values in some cases. In particular, default grouping will be adjusted to match the compaction grouping used by the locale: generally 3 digits, but "ko", "ja", "zh", and "zh_Hant" use 4.

This solution isn't ideal, as it doesn't support compact formats for currency values, only decimals. However, the ICU C API in `unum.h` still uses the old `NumberFormat` and `CompactDecimalFormat` classes under the hood, rather than the newer `NumberFormatter` which properly supports compact number formatting for pretty much everything. I'm not sure how to go about getting that support in place, though; I'm mulling sending a pull request to libicu to update the C API to optionally use `NumberFormatter` internally, or to expose a separate C interface to `NumberFormatter`, which could then be used by CoreFoundation.

### DIFF
--- a/CoreFoundation/Locale.subproj/CFNumberFormatter.c
+++ b/CoreFoundation/Locale.subproj/CFNumberFormatter.c
@@ -111,6 +111,10 @@ CFNumberFormatterRef CFNumberFormatterCreate(CFAllocatorRef allocator, CFLocaleR
 #if U_ICU_VERSION_MAJOR_NUM >= 55
     case kCFNumberFormatterCurrencyAccountingStyle: ustyle = UNUM_CURRENCY_ACCOUNTING; break;
 #endif
+#if U_ICU_VERSION_MAJOR_NUM >= 56 || (U_ICU_VERSION_MAJOR_NUM >= 55 && !defined(U_HIDE_DRAFT_API))
+    case kCFNumberFormatterDecimalCompactStyle: ustyle = UNUM_DECIMAL_COMPACT_SHORT; break;
+    case kCFNumberFormatterDecimalCompactPluralStyle: ustyle = UNUM_DECIMAL_COMPACT_LONG; break;
+#endif
     default:
 	CFAssert2(0, __kCFLogAssertion, "%s(): unknown style %ld", __PRETTY_FUNCTION__, style);
 	ustyle = UNUM_DECIMAL;
@@ -200,6 +204,10 @@ static void __substituteFormatStringFromPrefsNF(CFNumberFormatterRef formatter) 
 #if U_ICU_VERSION_MAJOR_NUM >= 55
 	    case kCFNumberFormatterCurrencyAccountingStyle: key = CFSTR("10"); break;
 #endif
+#if U_ICU_VERSION_MAJOR_NUM >= 56 || (U_ICU_VERSION_MAJOR_NUM >= 55 && !defined(U_HIDE_DRAFT_API))
+        case kCFNumberFormatterDecimalCompactStyle: key = CFSTR("11"); break;
+        case kCFNumberFormatterDecimalCompactPluralStyle: key = CFSTR("12"); break;
+#endif
 	    default: key = CFSTR("0"); break;
 	    }
 	    pref = (CFStringRef)CFDictionaryGetValue((CFDictionaryRef)metapref, key);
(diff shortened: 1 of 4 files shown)


millenomi:
"This doesn't match macOS/iOS" kind of dooms this patch.

Can you write a feedback for Darwin so we can track that, or does this have a SR ticket?

AlanQuatermain:
I’m rejoining the mothership in a couple of weeks; I’ll send over a Radar at that time, possibly with a patch attached.

worchyld:
Does this build upon: compactdecimalformat.h file as found in:

https://github.com/apple/swift-corelibs-foundation/blob/master/bootstrap/x86_64-apple-darwin/usr/local/include/unicode/compactdecimalformat.h

With thanks

xwu:
@worchyld I think you'll find that information in the PR description:

> This implementation makes use of the `UNUM_DECIMAL_COMPACT_SHORT` and `UNUM_DECIMAL_COMPACT_LONG` from `unicode/unum.h`.

> the ICU C API in `unum.h` still uses the old `NumberFormat` and `CompactDecimalFormat` classes under the hood

worchyld:
Many thanks.

shahmishal:

kerrmarin:
@AlanQuatermain is there anything I can do to help push this forward?

malhal:
This would be very useful

kerrmarin:
@malhal i created a pod that does this, it's not perfect because it is a large dependency (it includes icu c++ library) but it works well: https://github.com/kerrmarin/nmbr-icu

danpashin:
Any chance this PR will be implemented in Foundation in 2023 or even 2024? I created a [small snippet](https://gist.github.com/danpashin/5951706a6aa25748a7faa1acd5c1db8b) for my own purposes based on how CoreFoundation processes formats but I'd like to see this in Apple's public API

parkera:
I think if we were to implement this, it would be in the new `FormatStyle` API only. @itingliu any thoughts on this option?

AlanQuatermain:
I can look at what that would take and whip up a PR for that. With the Swift-first Foundation approach, would that API end up underlying the NS API after the fact?

Alternatively, with the new Foundation implementation on the way, would it be easier or more difficult to drop this into CF in a manner similar to this PR? I can write up an internal PR/proposal for that if it helps.

parkera:
The format style implementations are (nearly) all written in Swift and call into ICU directly, so it would be a different implementation. We're working in bringing it up now.

itingliu:
Format style implementation already supports this. So when we bring them over, this will be available instantly!

glotcha:
This would be great to have as a format style, any idea if/when it will be available to 3rd party devs?

PR #2714, opened 2020-03-04
Refactors _typeDescription function of DecodingError

valeriyvan:
Refactors _typeDescription function of DecodingError using switch for type casting


spevans:
@swift-ci test linux

parkera:
I don't know if we need to spend a lot of time refactoring if/else statements into switch statements. Is there any material difference? I think it's about equal in terms of readability.

PR #2715, opened 2020-03-04
SR-11720: Thread.name leads to empty String on Linux

spevans:
- Check the return value of _CFThreadGetName correctly

### DIFF
--- a/Tests/Foundation/Tests/TestThread.swift
+++ b/Tests/Foundation/Tests/TestThread.swift
@@ -20,25 +20,6 @@
 #endif
 
 class TestThread : XCTestCase {
-    static var allTests: [(String, (TestThread) -> () throws -> Void)] {
-        var tests: [(String, (TestThread) -> () throws -> Void)] = [
-            ("test_currentThread", test_currentThread),
-            ("test_threadStart", test_threadStart),
-            ("test_mainThread", test_mainThread),
-            ("test_callStackSymbols", testExpectedToFailOnAndroid(test_callStackSymbols, "Android doesn't support backtraces at the moment.")),
-            ("test_callStackReturnAddresses", testExpectedToFailOnAndroid(test_callStackReturnAddresses, "Android doesn't support backtraces at the moment.")),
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

Tests/Foundation/Tests/TestThread.swift
```diff
@@ -64,57 +45,36 @@ class TestThread : XCTestCase {
...
-            condition.signal()
-            condition.unlock()
-        }
-        thread2.name = "Thread2-1"
-        thread2.start()
-
-        // Allow 1 second for thread2 to finish
-        XCTAssertTrue(condition.wait(until: Date(timeIntervalSinceNow: 1)))
-        condition.unlock()
+        testInternalThreadName("mainThread")
```
3405691582: This might be something to address in another pr, but since thread names aren't supported on TARGET_OS_BSD (that is, `pthread_setname_np` nor `pthread_getname_np` are available), these tests will trip expectedly. Maybe the #if clauses can be tweaked?

spevans: I think for targets that dont support `pthread_{get,set}name_np`, the name can just be stored in a `String` in the `Thread` object.

Do the `swift-corelibs-foundation` tests build & run at all on BSD and is this going to break them or is it just another thing to remember to fix when the tests are ported?

3405691582: If it's as simple as a string variable in the other cases, then I think we can leave it as "another thing to remember to fix when the tests are ported". 

(There's still more hurdles to jump through on the porting -- like Dispatch and friends, _especially_ -- but with the WIP Swift BSD port, a decent percentage of Foundation tests appear to be running and passing.)

spevans:
@compnerd Can you run a windows-ci build to check this wont break it, thanks

PR #2716, opened 2020-03-06
Don't try to follow relative links in __CFTimeZoneCreateSystem

drexin:
Discovered this when trying to compile Swift for CentOS 7. I'm not sure if this is the appropriate fix, but it fixes the problem on CentOS.


drexin:
@swift-ci test

drexin:
@swift-ci reset

spevans:
Do you have an example of a symlink on Centos7 for `/etc/localtime` which is relative?

drexin:
@spevans ../usr/share/zoneinfo/UTC

millenomi:
Approved.

PR #2720, opened 2020-03-10
Fixes tests of OperationQueue

valeriyvan:
Fixes tests of OperationQueue: 
result of assignment was immediately deallocated because of weak property


spevans:
@swift-ci test linux

PR #2722, opened 2020-03-10
feat: support three different formats

zymxxxs:
This PR solves a bug and adds a feature
* unable to process Date field in header
* according the RCF 3.3.1 description, support three different formats (https://tools.ietf.org/html/rfc2616#section-3.3.1)

image

### old code screenshot
image

### new code screenshot
image

### DIFF
--- a/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
+++ b/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
@@ -239,34 +239,31 @@ internal class _HTTPURLProtocol: _NativeProtocol {
         return true
     }
     
-    static let dateFormatter: DateFormatter = {
-        let x = DateFormatter()
-        return x
-    }()
-    
-    static func date(from string: String) -> Date? {
+    static let dateFormatters: [DateFormatter] = {
         // https://tools.ietf.org/html/rfc2616#section-3.3.1
         // HTTP applications have historically allowed three different formats
         // for the representation of date/time stamps
         
         // RCF 822 --- Sun, 06 Nov 1994 08:49:37 GMT
-        self.dateFormatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss zzz"
-        if let d1 = self.dateFormatter.date(from: string) {
-            return d1
-        }
+        let d1 = DateFormatter()
+        d1.dateFormat = "EEE, dd MMM yyyy HH:mm:ss zzz"
         
         // RCF 855 --- Sunday, 06-Nov-94 08:49:37 GMT
-        self.dateFormatter.dateFormat = "EEEE, dd-MMM-yy HH:mm:ss zzz"
-        if let d2 = self.dateFormatter.date(from: string) {
-            return d2
-        }
+        let d2 = DateFormatter()
+        d2.dateFormat = "EEEE, dd-MMM-yy HH:mm:ss zzz"
         
         // ANSI C's asctime() format --- Sun Nov  6 08:49:37 1994
-        self.dateFormatter.dateFormat = "EEE MMM dd HH:mm:ss yy"
-        if let d3 = self.dateFormatter.date(from: string) {
-            return d3
+        let d3 = DateFormatter()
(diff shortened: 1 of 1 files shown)


zymxxxs:
@swift-ci test linux

Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
```diff
@@ -241,11 +241,35 @@ internal class _HTTPURLProtocol: _NativeProtocol {
...
         return x
     }()
     
+    static func date(from string: String) -> Date? {
+        // https://tools.ietf.org/html/rfc2616#section-3.3.1
+        // HTTP applications have historically allowed three different formats
+        // for the representation of date/time stamps
+        
+        // RCF 822 --- Sun, 06 Nov 1994 08:49:37 GMT
+        self.dateFormatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss zzz"
```
millenomi: There are costs associated with changing the settings of a formatter, and I’d rather just have three different formatters ready here.

zymxxxs:
```swift
    private var _dateFormat: String? { willSet { _reset() } }
    open var dateFormat: String! {
        get {
            guard let format = _dateFormat else {
                return CFDateFormatterGetFormat(_cfObject)._swiftObject
            }
            return format
        }
        set {
            _dateFormat = newValue
        }
    }
```

looks like anything was cost

millenomi: Sorry, I don’t understand your comment.

I was referring to the resources that CFDateFormatterSet… will destroy and recreate every time it’s called. I’d rather just have three different date formatters so we can amortize that call.

millenomi: (That call is in _reset().)

zymxxxs: OK，I got it

spevans:
Could you add some tests to in the three different date formats to test all three different formatters as well please.

zymxxxs:
@spevans 

The currently modified code is in the canCache method. There is no  test code. I'm not sure where to write the test code? Is it possible to write in a TestURLSession file or do you have any good suggestions?

spevans:
Since the only thing that needs to be tested is actually parsing dates in different formats I would suggest adding some tests into `TestHTTPURLResponse` that directly call the  `_HTTPURLProtocol.date(from: dateString)` method with dates in different formats.

 Since it is `internal` you will need to put the tests inside a
 `#if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT / #endif`
block and also add
 ```#if NS_FOUNDATION_ALLOWS_TESTABLE_IMPORT
    #if canImport(SwiftFoundation) && !DEPLOYMENT_RUNTIME_OBJC
    #else
        @testable import Foundation
    #endif
```
to the top of the test file.

You could setup some tests for `TestURLSession` but that would really be testing the caching validation mechanism and not specifically the date parsing. Also note that at the moment `TestURLSession` is currently disabled. It will be re-enabled once I have fixed a few more `URLSession` bugs and stopped some of the tests being flaky.

millenomi:
Also: I commented previously that changing the `.dateFormat` of an existing formatter is throwing away resources, and that I would much prefer a patch where three date for matters are created and cached. Can we have that?

It’s a tiny thing but we have our work cut out for us in SCF to improve performance, and not adding an obvious time-and-space issue is one way we can.

zymxxxs:
@millenomi [commit](https://github.com/apple/swift-corelibs-foundation/pull/2722/commits/fe3b72c19bdfe3e0f1dc7277161cd8ebd97e2971)


Is this ok?

millenomi:
I hadn’t seen that! That works.

millenomi:
(I still agree with @spevans that we need tests.)

zymxxxs:
sorry, I have been busy recently. I forgot about this and I will add UT as soon as possible

shahmishal:

PR #2723, opened 2020-03-12
HTTPServer: Update the Test server to be concurrent and improve reading requests

spevans:
- _TCPSocket is now only used for 1 socket, either a listener or client
  connection.

- _HTTPServer now listens for a connection and returns a new _HTTPServer with
  the client connection.

- Client _HTTPServer sub-servers are run in a Dispatch async() block to
  provide conncurrency for incoming connections.

- When reading the request from the socket, _SocketDataReader() is now used to
  read, as `Data` not `String`, and look for appropiate HTTP separators (\r\n).

- When reading chunked requests check for the length of each chunk and look for
  the terminating 0 byte chunk.

Note: This doesn't enable any tests and the current tests for `TestURLSession` are still disabled but following PRs will enable them.

### DIFF
--- a/Tests/Foundation/HTTPServer.swift
+++ b/Tests/Foundation/HTTPServer.swift
@@ -1,16 +1,16 @@
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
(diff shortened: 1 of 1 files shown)


compnerd:
https://dev.azure.com/compnerd/swift-build/_build/results?buildId=28046&view=results

PR #2724, opened 2020-03-13
Fix TestURLSession.test_noDoubleCallbackWhenCancellingAndProtocolFailsFast()

spevans:
- Darwin requires implementation of FailFastProtocol.canonicalRequest()
  to override abstract method in URLProtocol

- Fix the error code from in the callback to be equal to NSURLErrorCancelled.

- This test now passes on Darwin and Linux but is currently left disabled.


spevans:
@swift-ci test linux

PR #2725, opened 2020-03-13
URLSession: Fix InputStream usage

spevans:
- When passing a stream to a URLSession delegate, the stream must not
  be open(). It is opened by the underlying URLSession.

- This makes usage of InputStreams compatible with Darwin.


spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2726, opened 2020-03-13
NSString._fastContents: Remove use of deprecated UTF-16 String properties.

spevans:
- String._guts._isContiguousUTF16 is deprecated and always returns
  false.

- String._guts.startUTF16 is deprecated and aborts if called.

- _fastContents will always return nil as Swift Strings do not store
  UTF-16 data so no valid pointer can be returned, only nil. So hardcode
  this method to return nil.


spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2728, opened 2020-03-15
SR-12300: URLSession.dataTask doesn't follow redirects with lowercased location

spevans:
- When initialising the HTTPURLResponse, store the header fields names
  capitalised except for X- headers. This matches Darwin.

- This allows responses with lowercase fields names to be read correctly
  and redirects to works if the field name is 'location' instead of
  'Location'.

- Add missing function HTTPURLResponse.value(forHTTPHeaderField:).


spevans:
@swift-ci test linux

PR #2730, opened 2020-03-15
Removes not used constants

valeriyvan:


3405691582:
These appear to be used? See for example __CFDataBytesInline, which uses __CFRuntimeGetFlag/__CFRuntimeGetValue.

valeriyvan:
I don't see removed constants to be used.
CFData.c has similar declarations and uses them. 
Swift don't.
If C and Swift versions should have all constants in sync, even not used, Swift then should have `    __kCFNeedsZero = 6` which C has.

Another difference. In Swift we have:
```
private let __kCFMutable: CFOptionFlags = 0x01
private let __kCFGrowable: CFOptionFlags = 0x02
```

When in C we have following, which is not semantically the same.

```
    __kCFMutable = 0,
    __kCFGrowable = 1,
```

But we have in C these as well:

```
enum {
    __kCFMutableMask = 0x01,
    __kCFGrowableMask = 0x02,
};
```

Which could mean constants in Swift named wrongly, naming convention doesn't match one in C code.

millenomi:
In general, due to the way we process the code to make it available, I would rather extraneous declarations be kept.

PR #2731, opened 2020-03-15
Moves inside loop left operand of comma operator in do-while loop condition

valeriyvan:
Fix resolves warning of possible misuse of comma operator.
Fix preserves original logic.
But it seems like what was moved inside loop could be removed completely.
Or it might be that condition in do-while before fix was wrong and there should be && instead of comma operator.


compnerd:
Pretty sure that this was an existing bug, but your change simply drops the check which doesn't make sense to move into the loop.  I think it is better to fix this and that the intent was that both of them should be true and need to be concatenated with `&&`.

valeriyvan:
Done

valeriyvan:
Ping

valeriyvan:
ping

compnerd:

compnerd:
CC: @millenomi

millenomi:
I'm purposefully not taking patches that change the layout of CF code right now because I'm working on finishing this year's merge.

millenomi:
I finished this year's merge!

PR #2732, opened 2020-03-16
TestBridging: Use XCTUnwrap instead of force-unwrap

spevans:
- This test crashes on macOS so make the test soft fail until it is
  fixed properly.


spevans:
@swift-ci test linux

PR #2733, opened 2020-03-16
SR-12336: URLSessionTask should allow extras resume()'s

spevans:
- Every .suspend() has to be balanced with a .resume() to resume a task
  but multiple suspend()s or resume()s are allowed, so remove the check
  that the suspend count was not going < 0.

- Enhance the test HTTPServer by adding /get, /post, /put URLs that
  require the HTTP method to match the endpoint and return the headers
  as a JSON body.


spevans:
@swift-ci test linux

zymxxxs:
yeah, It's good.  I was going to fix it

spevans:
@swift-ci test linux

millenomi:
@Swift-ci please test linux

PR #2735, opened 2020-03-18
SR-12211: URLSession doesn't add Content-Length to empty body requests

spevans:
- For all methods except GET, if the request _Body is .none then
  set a length of 0. This will add a "Content-Length: 0" header.

- Fail any GET requests with a body, to match Darwin.


spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2736, opened 2020-03-18
SR-10652: URLProtocolClient.urlProtocolDidFinishLoading crashes with no response

spevans:
- urlProtocolDidFinishLoading may be called without a response being set
  if the protocol needs to early abort, so remove the check for a valid
  response.


spevans:
@swift-ci test linux

PR #2737, opened 2020-03-18
Support Linux variants with older libcurl

millenomi:
- Implement moving transfer progress information from libcurl to URLSessionTask.progress.
 - Remove unused CFURLSessionInterface constants that reference libcurl symbols that may not be available.
 - Remove commented-out constants.
 - Add two preprocessor macros/Swift build settings:
   - `NS_CURL_MISSING_XFERINFOFUNCTION`: if defined, we will not set a transfer info function. It makes the .progress property of URLSessionTask not report bytes as they are transferred. (It is still partly functional: it starts out as indeterminate, can report certain kind of upload progresses, will be set as done correctly when the task terminates, and its .cancel() method will cancel the task as usual.)
   - `NS_CURL_MISSING_MAX_HOST_CONNECTIONS`: if defined, we will not set the maximum number of per-host connections. The URLSessionConfiguration.httpMaximumConnectionsPerHost property will be ignored.
 - Swift symbols that are impacted by reduced functionality because of an older libcurl will diagnose at compile time with a warning.
 - Added CMake version testing for libcurl to add the compile definitions above when libcurl is older than the appropriate thresholds. It can be tested by specifying the `NS_CURL_ASSUME_FEATURES_MISSING` option on systems with newer libcurl.


millenomi:
@swift-ci please test linux

millenomi:
@swift-ci please test linux

Sources/FoundationNetworking/CMakeLists.txt
```diff
@@ -1,3 +1,19 @@
...
+    find_package(CURL REQUIRED)
+  endif()
+endif()
+
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
+  if((NS_CURL_ASSUME_FEATURES_MISSING) OR (CURL_VERSION_STRING VERSION_LESS "7.32.0"))
+    add_compile_definitions(NS_CURL_MISSING_XFERINFOFUNCTION)
+  endif()
+
+  if((NS_CURL_ASSUME_FEATURES_MISSING) OR (CURL_VERSION_STRING VERSION_LESS "7.30.0")) add_compile_definitions(NS_CURL_MISSING_MAX_HOST_CONNECTIONS)
```
drexin: This is missing a line break. The GitHub split view makes this a bit hard to spot, but the `add_compile_definitions` is on the same line as the `if`, which causes a parse error in cmake.

millenomi: Word wrap has defeated me. Fixing shortly.

millenomi:
@swift-ci please test linux

millenomi:
@swift-ci please test linux

PR #2739, opened 2020-03-21
CoreFoundation: adjust the thread name setting on Windows

compnerd:
Use newer Windows APIs (Windows Version 1607, RedStone 1, 10.0.14393) of
`SetThreadDescription` and `GetThreadDescription` rather than the
under-the-table hand-shake between Visual Studio and the runtime to use
a TLS slot for a thread description.  Although this bumps up the version
of the Windows requirements for setting the thread description, the
version requirements are already for an EOL-ed version of Windows.


compnerd:

PR #2741, opened 2020-03-23
Fix warning for matching enum with associated values as tuples.

spevans:
- Fixes "warning: enum case 'statusLine' has 3 associated values;
  matching them as a tuple is deprecated".


spevans:
@swift-ci test linux

PR #2740, opened 2020-03-22
Revert "Test HTTPServer with urls that match the method"

compnerd:
Reverts apple/swift-corelibs-foundation#2734

PR #2743, opened 2020-03-24
Enhance Test HTTP Server

spevans:
- Add support for URL parameters.

- Add status code URLs for /xxx to allow specifiing specific codes
  including setting the Location header for 3xx codes.

This includes a workaround that caused the revert in #2740 to hopefully build on Windows without triggering the SIL bug.


spevans:
@swift-ci test linux

spevans:
@compnerd Could you test this on windows please? I added a workaround that should hopefully stop the SIL bug from triggering.

spevans:
@swift-ci test linux

PR #2745, opened 2020-03-28
build: add a compatibility shim for 3.15

compnerd:
The default flags were not handled properly when using CMake 3.15.  This
adds a shim for CMake 3.15 to provide a targeted fix for a regression
caused by Foundation being built unoptimized even in release mode.


compnerd:
CC: @apple/swift5-branch-managers

drexin:
CC: @apple/swift5-branch-managers

drexin:
@compnerd I just tried this patch and it does not seem to fix the problem.

spevans:
Ive had some success with using CMake 3.17.0, it worked when I rebuilt `master` using it

compnerd:
@spevans  yes, cmake 3.16 and 3.17 don’t exhibit this issue.  That’s the worst case solution if we cannot work around the issue locally for the 5.2 release.

compnerd:
I think that this is no longer needed with the patch to CMake 3.15 on the 5.2 branch.

PR #2746, opened 2020-03-28
Revert "URLSession: Fix redirection response handling"

compnerd:
Reverts apple/swift-corelibs-foundation#2744

PR #2747, opened 2020-03-28
Revert "Revert "URLSession: Fix redirection response handling""

compnerd:
Reverts apple/swift-corelibs-foundation#2746

I think that this change was incorrectly flagged as the cause.  Lets get this un-reverted and tested to prepare for restoration.


compnerd:

PR #2748, opened 2020-03-30
Foundation: ensure the `_NSCFRuntime` metadata is initialized

compnerd:
Lazy initialization of the foreign class VWT requires that we explicitly
initialize the metadata.  This was previously relying on the compiler
not being able to optimize away the empty body.  Change this to call an
empty function instead.  This repairs the Windows build.


millenomi:

compnerd:
https://dev.azure.com/compnerd/swift-build/_build/results?buildId=29159&view=results Windows CI is successful with this change

compnerd:
The macOS failure is due to XCTest and not related to this test.

PR #2750, opened 2020-03-30
Add kilopound per square inch (ksi) and megapound per square inch to UnitPressure

matsoftware:
The goal of this PR is to extend the available measure units for `UnitPressure` by adding `ksi` (kilopound per square inch) and `Mpsi` (megapound per square inch) to the available measure units.


matsoftware:
@swift-ci test

matsoftware:
@swift-ci test linux

millenomi:
I appreciate the intent, but if it’s not in Darwin (https://developer.apple.com/documentation/foundation/unitpressure/1856093-millibars), it’s not something I will be able to merge as-is. What’s your use case?

millenomi:
Redo for incorrect link:

Hi! In general, I appreciate PRs that add useful API, but those cannot be merged as-is unless they’re implementing API that already exists on Darwin. In this case it doesn’t: https://developer.apple.com/documentation/foundation/unitpressure

What’s your use case?

matsoftware:
Hi! Those measure units are used in material science / civil engineering, so I thought it would be beneficial to add them (same for `UnitForce` which is missing as well), but I understand that we need first to have an alignment with Darwin.

I can extract the unit tests for the existing methods for UnitPressure in another PR. What do you think?

millenomi:
To be clear: we don’t have alignment with Darwin _first_, we have alignment with Darwin _in full_. If Darwin doesn’t offer an API surface, we don’t want to offer it here (so that we minimize the amount of incompatible code between platforms for the API surface we use).

You can create a SwiftPM module with an extension that vends these types for the time being!

millenomi:
(On the other hand, if Darwin has something and it’s missing here, it’s a bug that needs fixing.)

millenomi:
Given that I'm not taking this, closing.

PR #2751, opened 2020-03-31
URLSession: Fix bug requiring a new Body Stream

spevans:
- Initial body stream was being passed as nil and not _Body.none when
  creating a data task without a body. This caused the delegate's
  urlSession(_:task:needNewBodyStream:) method to be called
  unnecessarily.

- Add a test case and enhance the TestURLSession SessionDelegate to log
  the callbacks as they are called to check the number and order.

Fixes a bug introduced in #2735


spevans:
@swift-ci test

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2752, opened 2020-04-01
URLSession: Better handling for willPerformHTTPRedirection delegate calls

spevans:
- If the delegate calls the completionHandler with a nil response, then
  return the 3xx redirection response and its body data to the caller as
  the final response.

- If the delegate fails to call the completionHandler before the timeout
  interval for the request has expired, then dont treat is as an error
  and dont call urlSession(_:task:didCompleteWithError:).

- Fix TestURLSession.test_httpRedirectionWithDefaultPort() to work
  correctly now that the HTTPRedirectionDataTask delegate needs to call
  the completion handler with nil to cancel the redirect.

- This handling now matches Darwin.


spevans:
@swift-ci test linux

PR #2754, opened 2020-04-03
URLSession: Dont try to send a body when URLRequest.httpBody is an empty Data()

spevans:
- If the .httpBody is set to an empty Data() instance then treat this
  as if there was no .httpBody set and do not attempt to send a zero
  byte body to the remote server.

- Update TestURLSession.test_requestWithEmptyBody() and
  TestURLSession.test_requestWithNonEmptyBody() to check the delegate
  callbacks and the posted body sent back in the response.


spevans:
@swift-ci test linux

PR #2756, opened 2020-04-03
URLSession: Add a limit on the number of HTTP Redirects that are followed

spevans:
- Avoid infinite redirect chains by having a hard limit of 16 redirects,
  to match Darwin.


spevans:
@swift-ci test linux

PR #2760, opened 2020-04-05
URLSession: Update all task types with the byte count of sent data

spevans:
- urlSession(_:task:didSendBodyData:totalBytesSent:totalBytesExpectedToSend:)
  was only sent to tasks of type URLSesisonUploadTask, but testing showed
  it should be sent to all task types.

- Update tests that now see an extra delegate callback of this method.

- test_simpleUploadWithDelegateProvidingInputStream was flaky due to the
  expectation sometimes being fulfilled multiple times as it was
  fulfilled in the urlSession(_:dataTask:didReceive:) delegate.

- Migrate this test to the SessionDelegate and add extra tests to check
  the response and delegate callbacks for each HTTP method.

- For HTTPUploadDelegate, add the expectation fulfilment to a delegate
  callback for urlSession(_ :task:didCompleteWithError:) so that
  test_simpleUploadWithDelegate() still works.


spevans:
@swift-ci test

PR #2761, opened 2020-04-05
Create URL Protection Space from HTTPURLResponse

lxbndr:
Basic implementation of "Www-Authenticate" parsing. Some real-life responses could crash previous implementation, as it doesn't work with realm in case insensitive way. We decided to modify existing (but unused) internal helper, considering that a good step in iterative improvement of FoundationNetworking.

This also fixes access to "Www-Authenticate" header in some places,
as HTTPURLResponse does capitalize headers after #2728.

### DIFF
--- a/Sources/FoundationNetworking/URLProtectionSpace.swift
+++ b/Sources/FoundationNetworking/URLProtectionSpace.swift
@@ -332,19 +332,32 @@ open class URLProtectionSpace : NSObject, NSCopying {
 
 extension URLProtectionSpace {
     //an internal helper to create a URLProtectionSpace from a HTTPURLResponse 
-    static func create(using response: HTTPURLResponse) -> URLProtectionSpace {
-        let host = response.url?.host ?? ""
-        let port = response.url?.port ?? 80        //HTTP
-        let _protocol = response.url?.scheme
-        guard let wwwAuthHeader = response.allHeaderFields["Www-Authenticate"] as? String else {
-            fatalError("Authentication failed but no Www-Authenticate header in response")
+    static func create(with response: HTTPURLResponse) -> URLProtectionSpace? {
+        // Using first challenge, as we don't support multiple challenges yet
+        guard let challenge = _HTTPURLProtocol._HTTPMessage._Challenge.challenges(from: response).first else {
+            return nil
         }
+        guard let url = response.url, let host = url.host, let proto = url.scheme, proto == "http" || proto == "https" else {
+            return nil
+        }
+        let port = url.port ?? (proto == "http" ? 80 : 443)
+        return URLProtectionSpace(host: host,
+                                  port: port,
+                                  protocol: proto,
+                                  realm: challenge.parameter(withName: "realm")?.value,
+                                  authenticationMethod: challenge.authenticationMethod)
+    }
+}
 
-        //The format of the authentication header is `<auth-scheme> realm="<realm value>"`
-        //Example: `Basic realm="Fake Realm"`
-        let authMethod = wwwAuthHeader.components(separatedBy: " ")[0]
-        let realm = String(String(wwwAuthHeader.components(separatedBy: "realm=")[1].dropFirst()).dropLast())
-        return URLProtectionSpace(host: host, port: port, protocol: _protocol, realm: realm, authenticationMethod: authMethod)
+extension _HTTPURLProtocol._HTTPMessage._Challenge {
+    var authenticationMethod: String? {
+        if authScheme.caseInsensitiveCompare(_HTTPURLProtocol._HTTPMessage._Challenge.AuthSchemeBasic) == .orderedSame {
+            return NSURLAuthenticationMethodHTTPBasic
+        } else if authScheme.caseInsensitiveCompare(_HTTPURLProtocol._HTTPMessage._Challenge.AuthSchemeDigest) == .orderedSame {
+            return NSURLAuthenticationMethodHTTPDigest
+        } else {
+            return nil
+        }
     }
 }
 
--- a/Sources/FoundationNetworking/URLSession/Message.swift
+++ b/Sources/FoundationNetworking/URLSession/Message.swift
@@ -111,6 +111,10 @@ struct _Delimiters {
     static let Space = UnicodeScalar(0x20)
     static let HorizontalTab = UnicodeScalar(0x09)
     static let Colon = UnicodeScalar(0x3a)
+    static let Backslash = UnicodeScalar(0x5c)!
+    static let Comma = UnicodeScalar(0x2c)!
+    static let DoubleQuote = UnicodeScalar(0x22)!
+    static let Equals = UnicodeScalar(0x3d)!
     /// *Separators* according to RFC 2616
     static let Separators = NSCharacterSet(charactersIn: "()<>@,;:\\\"/[]?={} \t")
 }
--- a/Sources/FoundationNetworking/URLSession/HTTP/HTTPMessage.swift
+++ b/Sources/FoundationNetworking/URLSession/HTTP/HTTPMessage.swift
@@ -105,13 +105,209 @@ extension _HTTPURLProtocol._HTTPMessage {
     struct _Version: RawRepresentable {
         let rawValue: String
     }
+    /// An authentication challenge parsed from `WWW-Authenticate` header field.
+    ///
+    /// Only parts necessary for Basic auth scheme are implemented at the moment.
+    /// - SeeAlso: https://tools.ietf.org/html/rfc7235#section-4.1
+    struct _Challenge {
+        static let AuthSchemeBasic = "basic"
+        static let AuthSchemeDigest = "digest"
+        /// A single auth challenge parameter
+        struct _AuthParameter {
+            let name: String
+            let value: String
+        }
+        let authScheme: String
+        let authParameters: [_AuthParameter]
+    }
 }
 extension _HTTPURLProtocol._HTTPMessage._Version {
     init?(versionString: String) {
         rawValue = versionString
     }
 }
-
+extension _HTTPURLProtocol._HTTPMessage._Challenge {
+    /// Case-insensitively searches for auth parameter with specified name
+    func parameter(withName name: String) -> _AuthParameter? {
+        return authParameters.first { $0.name.caseInsensitiveCompare(name) == .orderedSame }
+    }
+}
+extension _HTTPURLProtocol._HTTPMessage._Challenge {
+    /// Creates authentication challenges from provided `HTTPURLResponse`.
+    ///
+    /// The value of `WWW-Authenticate` field is used for parsing authentication challenges
+    /// of supported type.
+    ///
+    /// - note: `Basic` is the only supported scheme at the moment.
+    /// - parameter response: A response to get header value from.
+    /// - returns: An array of supported challenges found in response.
+    /// # Reference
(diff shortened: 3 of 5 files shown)


lxbndr:
CC @compnerd @spevans @millenomi

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

Sources/FoundationNetworking/URLProtectionSpace.swift
```diff
@@ -332,19 +332,93 @@ open class URLProtectionSpace : NSObject, NSCopying {
 
 extension URLProtectionSpace {
     //an internal helper to create a URLProtectionSpace from a HTTPURLResponse 
-    static func create(using response: HTTPURLResponse) -> URLProtectionSpace {
+    static func create(with response: HTTPURLResponse) -> URLProtectionSpace? {
         let host = response.url?.host ?? ""
-        let port = response.url?.port ?? 80        //HTTP
-        let _protocol = response.url?.scheme
-        guard let wwwAuthHeader = response.allHeaderFields["Www-Authenticate"] as? String else {
-            fatalError("Authentication failed but no Www-Authenticate header in response")
```
spevans, line 340: It might be better to be stricter on the input validation, it can alway be loosened later on. Rather than defaulting values I would suggest:

```
guard let url = response.url, let host = url.host, let _protocol = url.scheme, _protocol == "http" || _protocol == "https" else {
    return nil
}
let port = url.port ?? (_protocol == "http" ? 80 : 443)
```

to tighten up the validation.

Sources/FoundationNetworking/URLProtectionSpace.swift
```diff
@@ -332,19 +332,93 @@ open class URLProtectionSpace : NSObject, NSCopying {
...
 
-        //The format of the authentication header is `<auth-scheme> realm="<realm value>"`
-        //Example: `Basic realm="Fake Realm"`
-        let authMethod = wwwAuthHeader.components(separatedBy: " ")[0]
-        let realm = String(String(wwwAuthHeader.components(separatedBy: "realm=")[1].dropFirst()).dropLast())
-        return URLProtectionSpace(host: host, port: port, protocol: _protocol, realm: realm, authenticationMethod: authMethod)
+        // Typical Www-Authenticate header is something like
+        //   Www-Authenticate: Digest realm="test", domain="/HTTP/Digest", nonce="e3d002b9b2080453fdacea2d89f2d102"
+
+        let components = authenticateValue.components(separatedBy: " ")
```
spevans, line 363: Will this split correctly if there are multiple spaces between each part?
 I think `split(separator: Character, maxSplits: Int = Int.max, omittingEmptySubsequences: Bool = true) -> [Substring]` might be a better choice as by default it will not add empty sequences to the output array.

Sources/FoundationNetworking/URLProtectionSpace.swift
```diff
@@ -332,19 +332,93 @@ open class URLProtectionSpace : NSObject, NSCopying {
...
+            switch authMethod.lowercased() {
+            case "basic":
+                return NSURLAuthenticationMethodHTTPBasic
+            case "digest":
+                return NSURLAuthenticationMethodHTTPDigest
+            case "ntlm":
+                return NSURLAuthenticationMethodNTLM
+            case "negotiate":
+                return NSURLAuthenticationMethodNegotiate
+            default:
```
spevans, line 378: Will it cause problems to create a `URLProtectionSpace` for methods that are not currently implemented (ie everything except Basic)? Could always return `nil` for these other ones for now

Sources/FoundationNetworking/URLProtectionSpace.swift
```diff
@@ -332,19 +332,93 @@ open class URLProtectionSpace : NSObject, NSCopying {
...
+
+        let realm: String? = {
+            guard let openingQuoteIndex = authenticateValue.range(of: "realm=\"", options: .caseInsensitive)?.upperBound else {
+                return nil
+            }
+
+            // Basic implementation.
+            // Doesn't follow complete spec of quoted strings (RFC-7230, p. 3.2.6).
+            // Handles escaping of quotes only.
+
```
spevans, line 391: Give that this is concerned with authentication I think that the complete spec should be followed. Maybe it could be broken out into a separate function that fully parses a `String` into its constituent parts. If any percent decoding needs to be done on the input string then there is  `DataURLProtocol._PercentDecoder` that could possibly be used.

Sources/FoundationNetworking/URLProtectionSpace.swift
lxbndr: Makes sense. And less code. I like it. Will do.

Sources/FoundationNetworking/URLProtectionSpace.swift
lxbndr: I actually thought about this helper point as parsing as much info as possible (and reasonable). Hmm, but on the other side it could take responsibility to cut off unsupported things. Agree, this should return nil for everything other than Basic auth.

Sources/FoundationNetworking/URLProtectionSpace.swift
lxbndr: Thanks for hint. I guess I will also look into curl implementation of similar task. AFAIK it is pretty simple. I like the idea of separate parsing, could be also tested separately.

millenomi:
Good by me once the feedback is looked at.

Sources/FoundationNetworking/URLProtectionSpace.swift
```diff
@@ -332,19 +332,93 @@ open class URLProtectionSpace : NSObject, NSCopying {
...
-            fatalError("Authentication failed but no Www-Authenticate header in response")
+
+        let scheme = response.url?.scheme
+
+        let port: Int = {
+            if let port = response.url?.port {
+                return port
+            }
+
+            switch scheme {
```
millenomi, line 345: I forget if `scheme` is normalized?

millenomi: This should probably be scheme.lowercased().

Sources/FoundationNetworking/URLProtectionSpace.swift
millenomi: This is the grammar:

```
   WWW-Authenticate = *( "," OWS ) challenge *( OWS "," [ OWS challenge
    ] )

   auth-param = token BWS "=" BWS ( token / quoted-string )
   auth-scheme = token

   challenge = auth-scheme [ 1*SP ( token68 / [ ( "," / auth-param ) *(
    OWS "," [ OWS auth-param ] ) ] ) ]
   credentials = auth-scheme [ 1*SP ( token68 / [ ( "," / auth-param )
    *( OWS "," [ OWS auth-param ] ) ] ) ]
```

where:

```
     OWS            = *( SP / HTAB )
                    ; optional whitespace
…
     BWS            = OWS
                    ; "bad" whitespace
```

so it looks like we need to split for " " or "\t". Use the CharacterSet variants maybe?

millenomi: https://tools.ietf.org/html/rfc7235#section-4.1 + https://tools.ietf.org/html/rfc7230#section-3.2.3

lxbndr:
I'm so sorry, the solution is really delayed. It is still WIP. Had done about 3/4. I am thinking to convert this to draft, to not to look like stale PR.

millenomi:
@lxbndr are you still unable to finish this patch?

lxbndr:
@millenomi 
I am sorry for delay, again, but yes, I've revisited it lately and now think it is ready to be pushed. Will do tomorrow.
It actually blocks me from finishing another auth-related issue, so not going to postpone it more.

millenomi:
Thank you! I'll be here to facilitate.

lxbndr:
This all started as an attempt to parse "realm" parameter and not to crash. New implementation is RFC-compliant. I think I found perfect place for parsing code, and it is in `_HTTPURLProtocol` extensions, where other HTTP-related parsing already lives.

All necessary string string comparisons are case-insensitive, but internally data stored as is, so `_Challenge` structure is more like raw parsed data.

Some simplifications made assuming we are supporting only "Basic" scheme. But in general this code could be used to adopt other schemes.

I reused as much existing code as seemed reasonable (like token parsing, whitespace trimming, character constants).

It seems like we could refactor existing code a bit to improve consistency and streamline some logic. E.g. existing code references RFC-2616, which is obsoleted by RFCs 7230-7235. To keep this patch focused I used old code as is.

The general flow of WWW-Authenticate parsing was inspired by CURL's approach, as it seems pretty robust.

lxbndr:
@millenomi Could you please take a look? I think I took all comments into account. But as changes are much bigger than first time, I am expecting everything.

millenomi:
@swift-ci please test

millenomi:
Nothing immediately jumps at me. At this point, this looks large and useful enough that we can start taking bug fixes as separate patches if needed.

millenomi:
@swift-ci please test

lxbndr:
I found that one line I've changed was unnecessary. Reverted [HTTPURLProtocol.swift:137](https://github.com/apple/swift-corelibs-foundation/blob/76068b8caf54f250a7be5336a7c6bb97f55469f8/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift#L137) to original.
No other changes were made.

@millenomi Is there anything else I can do here? I am not in a rush, just thought maybe I missed some requirement.

Sources/FoundationNetworking/URLSession/Message.swift
```diff
@@ -111,6 +111,10 @@ struct _Delimiters {
     static let Space = UnicodeScalar(0x20)
     static let HorizontalTab = UnicodeScalar(0x09)
     static let Colon = UnicodeScalar(0x3a)
+    static let Backslash = UnicodeScalar(0x5c)!
+    static let Comma = UnicodeScalar(0x2c)!
+    static let Dquote = UnicodeScalar(0x22)!
```
xwu, line 116: Nit: Mind if we call this `DoubleQuote`? We already spell HTAB `HorizontalTab` instead of `Htab`, and it just reads a little more clearly, I think, in the context where we have comments referring to "dequoted" strings and then code that refers to "dquote".

lxbndr: Sure! Anything to reduce confusion. Change pushed.

MaxDesiatov:
@swift-ci please test

lxbndr:
Build failure on Linux looks unrelated. @MaxDesiatov Could re-run fix that, WDYT?

MaxDesiatov:

lxbndr:
Rebased to latest main.

MaxDesiatov:
@swift-ci please test

lxbndr:
Check failures look unrelated. @MaxDesiatov could you please poke CI again?

MaxDesiatov:
@swift-ci please test

MaxDesiatov:
Would anyone mind if I merge this?

MaxDesiatov:
@swift-ci lease test

spevans:
@MaxDesiatov No I dont mind, lets get this in. I think have some URLSession fixes that rely on this so would be good to progress those.

MaxDesiatov:
Linux failure seems to be unrelated

MaxDesiatov:

spevans:

PR #2763, opened 2020-04-10
REFACTOR: Remove unnecessarily repeated condition

O-O-wl:
Remove unnecessarily repeated condition

```swift
if type == Data.self {
...
} else if type == Data.self {
...
}
```


compnerd:
I think that you may be confusing `NSDate` and `NSData`, they are different.

PR #2764, opened 2020-04-11
[Windows] CFThreadSpecific implementation allows access to deallocated value

lxbndr:
Unlike pthread-based implementation, TLS/FLS on Windows doesn't return NULL when reading value after destructor call.

To avoid that we have to nullify value in destructor callback.
The implementation needs to store key along with user data to perform proper cleanup.

This fixes crashes in `NotificationQueue` and `OperationQueue` on Windows.

Also introduced small delay in helper func of `TestNotificationQueue` to avoid false positive/false negative results in `test_defaultQueue` and `test_notificationQueueLifecycle`.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -1390,8 +1390,22 @@ CF_PRIVATE int asprintf(char **ret, const char *format, ...) {
 extern void swift_retain(void *);
 extern void swift_release(void *);
 
+#if TARGET_OS_WIN32
+typedef struct _CFThreadSpecificData {
+    CFTypeRef value;
+    _CFThreadSpecificKey key;
+} _CFThreadSpecificData;
+#endif
+
 static void _CFThreadSpecificDestructor(void *ctx) {
+#if TARGET_OS_WIN32
+    _CFThreadSpecificData *data = (_CFThreadSpecificData *)ctx;
+    FlsSetValue(data->key, NULL);
+    swift_release(data->value);
+    free(data);
+#else
     swift_release(ctx);
+#endif
 }
 
 _CFThreadSpecificKey _CFThreadSpecificKeyCreate() {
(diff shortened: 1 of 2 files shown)


lxbndr:
CC @compnerd @spevans @millenomi

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1416,8 +1434,15 @@ void _CFThreadSpecificSet(_CFThreadSpecificKey key, CFTypeRef _Nullable value) {
 #if TARGET_OS_WIN32
     if (value != NULL) {
         swift_retain((void *)value);
+
+        _CFThreadSpecificData *data = malloc(sizeof(_CFThreadSpecificData));
```
compnerd, line 1438: This can fail, please ensure that data is valid or abort.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1416,8 +1434,15 @@ void _CFThreadSpecificSet(_CFThreadSpecificKey key, CFTypeRef _Nullable value) {
...
+        data->value = value;
+        data->key = key;
+
+        FlsSetValue(key, data);
+    } else {
+        FlsSetValue(key, NULL);
```
compnerd, line 1444: This may cause a leak, and worse a crash - you may have had an entry and you lost it.

lxbndr: Oh, this is definitely a leak. Thanks!

I wonder how does it work for Swift value we store here? We are not doing `swift_release` for previously stored object when overwriting it. But it not leaks, as far as I see.

lxbndr:
- done cleanup of old value
- HALT on malloc fail
@compnerd

lxbndr:
As for my question about missing `swift_release` on value overwrite/clean. Value actually leaks when stored with Thread Specific API.
But there is (almost) no coincidence of this leak, because of specific usage:
- NotificationQueue uses ThreadSpecific as singleton storage. It not overwrites value once it stored.
- OperationQueue does manual release in several places, like here: [OperationQueue.swift:305](https://github.com/apple/swift-corelibs-foundation/blob/9991f6a98749923856aba566b161abfd9685379d/Sources/Foundation/Operation.swift#L305). Technically could be overrelease, but it balances Thread Specific retain.
I guess there is still a danger zone, as Thread Specific destructor callback could try `swift_release` already released value. But this potential issue is out of scope of this patch.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1415,9 +1433,23 @@ CFTypeRef _Nullable _CFThreadSpecificGet(_CFThreadSpecificKey key) {
 void _CFThreadSpecificSet(_CFThreadSpecificKey key, CFTypeRef _Nullable value) {
 #if TARGET_OS_WIN32
     if (value != NULL) {
+        _CFThreadSpecificData *oldData = (_CFThreadSpecificData *)FlsGetValue(key);
+        if (oldData) {
+            free(oldData);
```
compnerd, line 1438: Probably should add a comment explaining that the value should not be freed here.

CoreFoundation/Base.subproj/CFPlatform.c
compnerd: This isn't resolved.  The problem is that you are now injecting a value that you are de-referencing rather than previously which was passed to `free` or `swift_release` which accept and ignore `NULL`.

lxbndr: You're right, by bad :(

lxbndr: Done

compnerd:

compnerd:

compnerd:
https://dev.azure.com/compnerd/swift-build/_build/results?buildId=30308&view=results

compnerd:

compnerd:

compnerd:
apple/swift#31144

PR #2765, opened 2020-04-12
Fix Process tests that fails when stdin is empty

Frizlab:
When building Swift in a docker via `docker build`, `stdin` returns `EOF` at each read and there is nothing we can do about it.
Because of this, the tests `test_terminate` and `test_preStartEndState` fail because the helper exits w/ code 0 before the swift `Process` can terminate it.

This PR fixes this problem by setting a `Pipe` stdin to the `Process`es.


millenomi:

PR #2769, opened 2020-04-15
Use time_t.init directly when calling timespec

MaxDesiatov:
Currently an attempt to pass an `Int` fails on platforms where `time_t` is 64-bit, such as WASI with its [wasi-libc](https://github.com/WebAssembly/wasi-libc/blob/320054e84f8f2440def3b1c8700cedb8fd697bf8/basics/include/__typedef_time_t.h).

(cc @compnerd)

### DIFF
--- a/Sources/Foundation/NSLock.swift
+++ b/Sources/Foundation/NSLock.swift
@@ -419,7 +419,7 @@ private func timeSpecFrom(date: Date) -> timespec? {
     let interval = date.timeIntervalSince1970
     let intervalNS = Int64(interval * Double(nsecPerSec))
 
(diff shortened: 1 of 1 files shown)


3405691582:
I've been wanting to make this exact change myself.

spevans:
@swift-ci test linux

MaxDesiatov:
@3405691582 unfortunately WASI doesn't support threading yet, so I'm excluding `Thread.swift` completely from WASI builds, that's why I haven't stumbled upon other instances of this.

3405691582:
@MaxDesiatov No worries, I'll get to it myself eventually.

MaxDesiatov:
@spevans can this be merged now as CI passed?

PR #2775, opened 2020-04-24
Fixes nullability annotation of CFPreferencesCopyMultiple func

valeriyvan:


spevans:
@swift-ci test linux

millenomi:
For historical reasons, some API have stricter nullability modifiers than their implementations. As these are API, they are frozen in time — no changing them.

valeriyvan:
> For historical reasons, some API have stricter nullability modifiers than their implementations. As these are API, they are frozen in time — no changing them.

Does it sound like for historical reasons app could crash?

valeriyvan:
@millenomi, should I close this PR?

PR #2777, opened 2020-04-26
Fix OperationQueue counters

lxbndr:
This fixes typos in code responsible for counting operations in queue.

CC @compnerd

### DIFF
--- a/Tests/Foundation/Tests/TestOperationQueue.swift
+++ b/Tests/Foundation/Tests/TestOperationQueue.swift
@@ -13,7 +13,7 @@ class TestOperationQueue : XCTestCase {
     static var allTests: [(String, (TestOperationQueue) -> () throws -> Void)] {
         return [
             ("test_OperationPriorities", test_OperationPriorities),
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test linux

Tests/Foundation/Tests/TestOperationQueue.swift
```diff
@@ -13,7 +13,7 @@ class TestOperationQueue : XCTestCase {
     static var allTests: [(String, (TestOperationQueue) -> () throws -> Void)] {
         return [
             ("test_OperationPriorities", test_OperationPriorities),
-            /* ⚠️ */ ("test_OperationCount", testExpectedToFail(test_OperationCount, "OperationQueue can't count to more than 1")),
+            ("test_OperationCount", test_OperationCount),
```
compnerd, line 16: Might be nice to have a test that ensures that the operations returned are also correct.  I don't think that we need to hold up this particular patch for that, but would be a good addition to the test suite in a follow up change.

PR #2779, opened 2020-04-29
Balance OperationQueue retain/release

lxbndr:
`Operation._queue` internal property does release stored value on every getter call. In most scenarios it is accessed exactly once, and balanced by "leaking" same value to Thread Specific storage. But under some conditions `_queue` is accessed twice, and crash is inevitable.

I strongly believe that property accessor should not release its value. Kind of typo for me. Operation already has the pair of `_adopt(queue:schedule:)`/`_invalidateQueue()` functions intended to manage ownership of queue, and they are used in proper places.

This patch
- prevents releasing of stored value in `Operation._queue` property getter
- balances retain count after current queue removed from thread-specific storage
- enables all available Operation/OperationQueue tests (including lifecycle test to make sure Queue not leaks with this change)
- makes test_OperationWaitUntilFinished stable
- extends test_OperationCount to cover more conditions

cc @compnerd @spevans  @millenomi

### DIFF
--- a/Sources/Foundation/Operation.swift
+++ b/Sources/Foundation/Operation.swift
@@ -125,7 +125,7 @@ open class Operation : NSObject {
     internal var _queue: OperationQueue? {
         _lock()
         defer { _unlock() }
-        return __queue?.takeRetainedValue()
+        return __queue?.takeUnretainedValue()
     }
     
     internal func _adopt(queue: OperationQueue, schedule: DispatchWorkItem) {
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

compnerd:
Windows Run: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=31075&view=results

compnerd:
Windows build looks good as well.  This seems correct to me.

broadwaylamb:
Will this be able to make it into Swift 5.3? Seems like a pretty nasty bug.

spevans:
@broadwaylamb I created #2831 to backport some OperationQueue fixes to 5.3

PR #2781, opened 2020-05-05
SR-12696: Add JSONEncoder.OutputFormatting.withoutEscapingSlashes for Linux

spevans:
- This was missing and is required to match Darwin.

### DIFF
--- a/Sources/Foundation/JSONEncoder.swift
+++ b/Sources/Foundation/JSONEncoder.swift
@@ -57,6 +57,12 @@ open class JSONEncoder {
         /// Produce JSON with dictionary keys sorted in lexicographic order.
         @available(macOS 10.13, iOS 11.0, watchOS 4.0, tvOS 11.0, *)
         public static let sortedKeys    = OutputFormatting(rawValue: 1 << 1)
+
+        /// By default slashes get escaped ("/" → "\/", "http://apple.com/" → "http:\/\/apple.com\/")
+        /// for security reasons, allowing outputted JSON to be safely embedded within HTML/XML.
+        /// In contexts where this escaping is unnecessary, the JSON is known to not be embedded,
+        /// or is intended only for display, this option avoids this escaping.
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test linux

Sources/Foundation/JSONEncoder.swift
```diff
@@ -57,6 +57,13 @@ open class JSONEncoder {
         /// Produce JSON with dictionary keys sorted in lexicographic order.
         @available(macOS 10.13, iOS 11.0, watchOS 4.0, tvOS 11.0, *)
         public static let sortedKeys    = OutputFormatting(rawValue: 1 << 1)
+
+        /// By default slashes get escaped ("/" → "\/", "http://apple.com/" → "http:\/\/apple.com\/")
+        /// for security reasons, allowing outputted JSON to be safely embedded within HTML/XML.
+        /// In contexts where this escaping is unnecessary, the JSON is known to not be embedded,
+        /// or is intended only for display, this option avoids this escaping.
+        @available(OSX 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
millenomi, line 65: There are a bunch of available(… apple platforms …) tags strewn about SCF, but they're not relevant to SCF. (They really should be `@available(swift 5.3…)`, etc.)

millenomi:
@swift-ci please test

millenomi:
Testing again with the Catalina merge.

Sources/Foundation/JSONEncoder.swift
spevans: Yeah I only copied them from the Foundation headers because there are `@available` annotations on other things. But I agree I dont see the point of them in SCF  - even marking them for a specific swift version seems redundant since the code is tied to the specific branch anyway. I will remove these ones.

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2785, opened 2020-05-07
Allow whitespaces after parsed content in DateFormatter.date(from:)

lxbndr:
Noticed the difference between macOS and Windows test results in our codebase (we are heavily dependent on date parsing). I have not found any docs about this, but actual results and deep debugging left no doubt about the mechanics.

Apple implementation checks for unparsed content in source string.
**Any non-whitespace** leftover characters are considered as failure.
On the other hand, **whitespaces are permitted** and should not affect the result.

Related to #1822 (SR-9322).
This patch alters verification logic to be closer to Apple implementation.

Not sure about `preconditionFailure` check, but that's really unexpected condition.

cc @compnerd @spevans

### DIFF
--- a/Sources/Foundation/DateFormatter.swift
+++ b/Sources/Foundation/DateFormatter.swift
@@ -64,8 +64,15 @@ open class DateFormatter : Formatter {
         }
 
         // range.length is updated with the last position of the input string that was parsed
-        guard range.length == string.length else {
-            // The whole string was not parsed
+        guard let swiftRange = Range(NSRange(range), in: string) else {
+            fatalError("Incorrect range \(range) in \(string)")
+        }
+        
+        // Apple DateFormatter implementation returns nil
+        // if non-whitespace sharacters are left after parsed content.
+        let remainder = String(string[swiftRange.upperBound...])
+        let characterSet = CharacterSet(charactersIn: remainder)
+        guard CharacterSet.whitespaces.isSuperset(of: characterSet) else {
             return nil
         }
         return date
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test linux

millenomi:
@swift-ci please test

compnerd:
Windows Run: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=31494&view=results

Sources/Foundation/DateFormatter.swift
```diff
@@ -64,8 +64,15 @@ open class DateFormatter : Formatter {
         }
 
         // range.length is updated with the last position of the input string that was parsed
-        guard range.length == string.length else {
-            // The whole string was not parsed
+        guard let swiftRange = Range(NSRange(range), in: string) else {
+            preconditionFailure("Incorrect range \(range) in \(string)")
```
spevans, line 68: If this is triggered is it due to a bug? should it be `fatalError()`  instead?

lxbndr: Oh, right. Should be `fatalError()`, indeed.

lxbndr: Done. And rebased on latest master.

spevans:
@swift-ci test linux

millenomi:
@swift-ci please test linux

millenomi:
Gotta say please. 😇

millenomi:
By the way, macOS works now.

PR #2789, opened 2020-05-11
[CF] Unify return types in CFPropertyList again.

3405691582:
In #2598, we brought the prototypes and declarations into alignment for
_CFPropertyListCreateFiltered and _CFPropertyListCreateSingleValue,
namely, using a `Boolean` for prototype and declaration, since they were
not matched.

It appears with the Catalina merge they were brought into alignment the
other way around, namely using `bool` instead. Since the Catalina merge
chose `bool`, let's just go with that.


millenomi:
Can't merge.

The issue here is twofold:

 - This isn't just Swift's Foundation. This is code from Foundation proper, and it gets sync'd back and forth. It would make sense to clean this up in a vacuum, but…

- … this is an exported symbol, and must remain in the form it is because if I ever merge your change back, it would alter Darwin's Foundation. When a symbol is CF_EXPORT'd, even if it isn't API, it means something may be relying on it, and we need to be binary- and source-compatible there.

So, sorry, no. I reverted it on purpose.

3405691582:
In #2598, it made the decision to make these both `Boolean`. As you say, that was not the right choice; they should have been `bool`. So this PR chooses `bool` here. 

Ultimately, they should be the same, or convertible. When `DEPLOYMENT_RUNTIME_SWIFT`, `Boolean` is typedef'd to `_Bool`, which is convertible to `bool`, otherwise, it is `unsigned char`, which isn't: I get a "conflicting types" error when trying to build when not `DEPLOYMENT_RUNTIME_SWIFT`.

So, if they cannot be the same type in header and implementation, would it be better to change the typedef instead?

shahmishal:

PR #2782, opened 2020-05-06
The Catalina Merge

millenomi:
This patch brings swift-corelibs-foundation closer to behavior parity with released versions of macOS Catalina and aligned Apple releases. Most of these changes are under-the-hood bug fixes or performance improvements, but a few may affect the behavior of your applications. In particular:

 - The deprecation of NSURL.parameterString is now also effective in Swift for Linux, with similar effects. This brings URL parsing in line with Catalina. Note that Swift for Linux does not support link checks, since there is no binary compatibility guarantee on Linux; this change will be effective to all application code running on Swift for Linux 5.3 or later.

 - Certain CFXML… types that have been deprecated on macOS are now removed, along with their associated implementation: CFXMLInputStream, CFXMLNode, CFXMLParser. Modern code should be using the FoundationXML module (or, on Darwin, the XMLDocument and related classes in Foundation) to parse documents.

PR #2790, opened 2020-05-11
[CF] Rename remaining __kCFAllocatorTypeID_CONST.

3405691582:
PR #2782 references symbol __kCFAllocatorTypeID_CONST in CFRuntime.c but
this was removed in #1708. After interpreting the commit, it looks like
this symbol was renamed to _kCFRuntimeIDCFAllocator. Therefore, make
this rename again.


millenomi:
Why did this compile?

3405691582:
If I'm looking at it right, CFRelease just passes through to swift_release when DEPLOYMENT_RUNTIME_SWIFT, but I've been experimenting with standalone CF builds, which means DEPLOYMENT_RUNTIME_SWIFT is not set and the rest of the body of CFRelease is not compiled.

millenomi:
Ah, yeah, that makes sense.

millenomi:

millenomi:

PR #2792, opened 2020-05-12
Base64: Speedup encoding

spevans:
- Allocate the output buffer with an alignment of 4 bytes. In the main
  encoding loop, data is written out as a UInt32 using .copyMemory()
  instead of .storeBytes() as the output position can become unaligned
  if EOL markers are written as well.

- This allows removal of the appendByteToResult() lexically scoped
  helper function which was called three times and thus being outlined
  to a separate function. The extra state that needed to be passed to it
  caused extra memory traffic and this has now been eliminated.

- When converting the final output buffer into a String, use
  String(decoding:as:) in preference to
  String(bytesNoCopy:length:encoding:freeWhenDone:) as it is slightly
  faster for string conversion, and there is no chance of it returning
  nil since the input is known to be valid ASCII.

- These changes give approximately a 40% speedup using the benchmarking
  tests @ https://github.com/spevans/FoundationBenchmarks

### DIFF
--- a/Sources/Foundation/NSData.swift
+++ b/Sources/Foundation/NSData.swift
@@ -599,11 +599,12 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
         if dataLength == 0 { return "" }
 
         let capacity = estimateBase64Size(length: dataLength)
-        let ptr = UnsafeMutableRawPointer.allocate(byteCount: capacity, alignment: 1)
+        let ptr = UnsafeMutableRawPointer.allocate(byteCount: capacity, alignment: 4)
+        defer { ptr.deallocate() }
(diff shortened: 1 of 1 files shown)


spevans:
@swift-c test

spevans:
@swift-ci test

millenomi:
Rather than directly call into the stdlib, we should make the String API faster in NSStringAPI.swift if possible.

millenomi:
But that can be a separate patch.

spevans:
I was actually thinking of using `String(unsafeUninitializedCapacity:initializingUTF8With:)` in a future change since the buffer has already been allocated and the generated string is known to be ASCII, however I have a few more PRs for encoding so it can come along later.

PR #2794, opened 2020-05-12
Revert "Work around ICU version < 62"

DougGregor:
Reverts apple/swift-corelibs-foundation#2793. We shouldn't need this now.


DougGregor:
@swift-ci test

PR #2796, opened 2020-05-17
Base64: Speedup decoding fix #1

spevans:
- Use a lookup table to convert input ASCII to byte values. This is
  faster and simpler than scanning through multiple ranges in a loop
  for each input byte.

- Padding character '=' and invalid inputs both have bit 6 set in their
  decoded value so that both values can be tested together.

- Fix the estimate of the output buffer size which was estimating it at
  2/3 of input buffer size, not 3/4.

### DIFF
--- a/Sources/Foundation/NSData.swift
+++ b/Sources/Foundation/NSData.swift
@@ -619,90 +619,88 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
         }))
     }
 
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci please test

spevans:
@swift-ci test and merge

spevans:
@swift-ci test macos

spevans:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #2797, opened 2020-05-18
Test fixes for Darwin

spevans:
- lastPathComponent is a NSString method so bridge appropiately.

- JSONEncoder.OutputFormatting.sortedKeys requires macOS 10.13, add an
  #available around it.


spevans:

PR #2799, opened 2020-05-20
Base64: Speedup encoding 2

spevans:
- In the main loop that reads input bytes, where there at least 3 bytes
  left, read them directly instead of using .next() and use a simpler
  processing loop that can avoid checks for end of input. This allows
  removal of any branches inside the main loop. The end of input is
  handled separately since at most it is run once at the end.

- Rename the input buffer index from `index` to `inputIndex` to avoid
  confusion with the newly added `outputIndex`.

There are some benchmarks for the encoding PRs @ http://si.org/posts/2020/05/base64-encoding.html


spevans:
@swift-ci please test

spevans:
@swift-ci test and merge

PR #2800, opened 2020-05-20
Make fileActions constant, not variable in Process.swift

MaxDesiatov:
The pointer value of it is not changed, which with `var` currently leads to this warning:

```
/Sources/Foundation/Process.swift:828:13: 
warning: variable 'fileActions' was never mutated; consider changing to 'let' constant
        var fileActions = _CFPosixSpawnFileActionsAlloc()
```


spevans:
@swift-ci please test

PR #2802, opened 2020-05-21
Base64: Speedup encoding fix 3

spevans:
- Pass UnsafeRawBufferPointers to NSData.base64EncodeBytes() and make
  the Data.base64EncodedString() and Data.base64EncodedData() methods
  call it directly.

- There is considerable overhead in forwarding the methods from Data to
  NSData. Making the Data methods directly call NSData.base64EncodeBytes()
  removes this overhead.


spevans:
@swift-ci test

PR #2810, opened 2020-06-04
Eliminates redundant array creation in convinience NSString.init()

valeriyvan:


Sources/Foundation/NSString.swift
```diff
@@ -231,8 +231,8 @@ open class NSString : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSC
     }
     
     public override convenience init() {
-        let characters = Array<unichar>(repeating: 0, count: 1)
-        self.init(characters: characters, length: 0)
+        var nul: unichar = 0
+        self.init(characters: &nul, length: 0)
     }
```
spevans, line 236: Since this is ultimately an empy string, can the whole function body be reduced to 
```
self.init("")
```
?

valeriyvan: yes, it could.

spevans:
@swift-ci please test

spevans:

PR #2812, opened 2020-06-04
Changes type avoiding undefined behaviour shifting left negative value in CFTimeZone

valeriyvan:


spevans:
@swift-ci test

valeriyvan:
> Does this need a `return (uint32_t)result` to cast the result to the correct return type ?

No. Implicit cast from unsigned to signed of same size in not undefined behaviour. It is implementation-defined. Which means different compilers might legitimately do this differently.

We could add explicit cast to make intention clear. But anyway, it stays implementation-defined.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test macos

spevans:
@swift-ci test and merge

PR #2813, opened 2020-06-04
SR-12450: Add missing String.init(_ cocoaString: NSString) initialiser

spevans:
- https://developer.apple.com/documentation/swift/string/1410075-init


spevans:
@swift-ci please test

spevans:

PR #2814, opened 2020-06-04
Uses var (otherwise not used) as parameter to open file in CFFileUtilities

valeriyvan:


spevans:
@swift-ci please test

spevans:
@swift-ci please test

valeriyvan:
ping

valeriyvan:
ping

millenomi:
cc @niw This is the CF dual of your last patch — should it be 0666?

niw:
Actually, probably get `mode` by `stat(2)` for existing path is unnecessary at least on POSIX system, because `mode` (`oflag`) is used only when it is creating a new file by `O_CREAT`.
Therefore, the original code works as expected (but has unused `mode`,) also this patch works as expected. 

I think it is preferred to not have unused `mode` for sure, so I think this change is reasonable.

PR #2815, opened 2020-06-04
Fix for Foundation on Android 6.0

Molanda:
Foundation's Host.swift uses `getifaddrs` and `freeifaddrs`, which currently causes Android 6.0 or below to fail to link.

This PR changes these functions so they are called using `dlsym`, which allows Android 6.0 to link while providing full support on later versions of Android.

This is a replacement for PR https://github.com/apple/swift-corelibs-foundation/pull/2439.

### DIFF
--- a/Sources/Foundation/Host.swift
+++ b/Sources/Foundation/Host.swift
@@ -25,7 +25,9 @@ import CoreFoundation
         return Glibc.getnameinfo(addr, addrlen, host, Int(hostlen), serv, Int(servlen), flags)
     }
 
-    // getifaddrs and freeifaddrs are not available in Android 6.0 or earlier, so call the functions dynamically.
+    // getifaddrs and freeifaddrs are not available in Android 6.0 or earlier, so call these functions dynamically.
+    // This only happens during the initial lookup of the addresses and then the results are cached and _resolved is marked true.
+    // If this API changes so these functions are called more frequently, it might be beneficial to cache the function pointers.
 
     private typealias GetIfAddrsFunc = @convention(c) (UnsafeMutablePointer<UnsafeMutablePointer<ifaddrs>?>) -> Int32
     private func getifaddrs(_ ifap: UnsafeMutablePointer<UnsafeMutablePointer<ifaddrs>?>) -> Int32 {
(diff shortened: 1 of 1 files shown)


compnerd:
Does this not re-open the library each call and then lookup the symbol?  That isn't the cheapest operation.  Why not make the function a lazy static that will cache the value avoiding the `dlopen`/`dlsym` on subsequent calls?

Molanda:
No, `RTLD_NOLOAD` returns the shared instance.  This is only called once, then `_resolved` is set to `true`.

compnerd:
Ah, good point about the NOLOAD.  I’m still worried that future changes may end up calling the function multiple times, looking up the symbol each time.

Molanda:
Understood. I updated the comment to capture your concern for any future changes.

compnerd:
@swift-ci please test

compnerd:
CC: @drodriguez

Sources/Foundation/Host.swift
```diff
@@ -24,6 +24,32 @@ import CoreFoundation
...
         return Glibc.getnameinfo(addr, addrlen, host, Int(hostlen), serv, Int(servlen), flags)
     }
+
+    // getifaddrs and freeifaddrs are not available in Android 6.0 or earlier, so call these functions dynamically.
+    // This only happens during the initial lookup of the addresses and then the results are cached and _resolved is marked true.
+    // If this API changes so these functions are called more frequently, it might be beneficial to cache the function pointers.
+
+    private typealias GetIfAddrsFunc = @convention(c) (UnsafeMutablePointer<UnsafeMutablePointer<ifaddrs>?>) -> Int32
+    private func getifaddrs(_ ifap: UnsafeMutablePointer<UnsafeMutablePointer<ifaddrs>?>) -> Int32 {
+        var result: Int32 = 0
```
drodriguez, line 34: I would initialize with -1, which seems to be the error condition for the function. That way, the code below will not try to do anything else in case the handle cannot be obtained, or the function is not found.

Molanda: Sure, I made the change as suggested.

drodriguez:
@swift-ci please test

drodriguez:
@swift-ci please test

drodriguez:
I don't know why CI is not triggering. Someone else might need to trigger it.

spevans:
@swift-ci please test

Molanda:
Looks like the Linux build failed for an unrelated reason...

```
/home/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-Linux/swift-corelibs-libdispatch/src/swift/Block.swift:13:8: error: no such module 'CDispatch'
```

spevans:
@swift-ci please test linux

PR #2821, opened 2020-06-09
Added CVarArg support to NSString and String

Molanda:
This adds `CVarArg` protocol support to `NSString` and `String`.

```
print(String(format: "Testing %@ %@ %@", "One", "Two", "Three"))

Testing One Two Three
```

`NSString` support was easy since `%@` already exists for `CFString` in CoreFoundation.

`String` was a little bit more challenging, since `withVaList` doesn't provide an autorelease pool to retain the required `NSString` instance, which is needed for the `CFString`.

I solved this by using `OperationQueue` to schedule the release.  I'm not sure if there are any other options at this point.

This resolves [SR-957](https://bugs.swift.org/browse/SR-957).

### DIFF
--- a/Sources/Foundation/NSString.swift
+++ b/Sources/Foundation/NSString.swift
@@ -1644,3 +1644,23 @@ extension NSString : _StructTypeBridgeable {
         return _StructType._unconditionallyBridgeFromObjectiveC(self)
     }
 }
+
+extension NSString : CVarArg {
+    @inlinable // c-abi
+    public var _cVarArgEncoding: [Int] {
+        return _encodeBitsAsWords(unsafeBitCast(self, to: CFString.self))
+    }
+}
+
+extension String : CVarArg {
+    @inlinable // c-abi
+    public var _cVarArgEncoding: [Int] {
+        // We don't have an autorelease pool to retain the NSString until the withVaList closure is complete.
+        // So add an operation to release on the next cycle of this thread.
+        let ns = Unmanaged.passRetained(NSString(string: self))
+        OperationQueue.current?.addOperation {
+            ns.release()
+        }
+        return ns.takeUnretainedValue()._cVarArgEncoding
+    }
+}
--- a/Sources/Foundation/NSString.swift
+++ b/Sources/Foundation/NSString.swift
@@ -1648,7 +1648,7 @@ extension NSString : _StructTypeBridgeable {
 extension NSString : CVarArg {
     @inlinable // c-abi
     public var _cVarArgEncoding: [Int] {
-        return _encodeBitsAsWords(unsafeBitCast(self, to: CFString.self))
+        return _encodeBitsAsWords(unsafeBitCast(self, to: AnyObject.self))
     }
 }
 
--- a/Sources/Foundation/NSString.swift
+++ b/Sources/Foundation/NSString.swift
@@ -1652,15 +1652,14 @@ extension NSString : CVarArg {
     }
 }
 
-extension String : CVarArg {
+extension String : CVarArg, _CVarArgObject {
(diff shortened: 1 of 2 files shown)


Molanda:
I created [Swift PR #32311](https://github.com/apple/swift/pull/32311) to implement a new `_CVarArgObject` protocol to retain the `NSString`.

This PR depends on that one to be approved and merged first.

millenomi:
Please test with the following:
https://github.com/apple/swift/pull/32311

@swift-ci please test

Molanda:
With [Swift PR #32311](https://github.com/apple/swift/pull/32311) merged into master, we can now proceed on this one.

It is my understanding that this implementation of Foundation expects that there will never by any Objective-C runtime (at least that is what the introductory README states).  However, I did find some cases of `#if !_runtime(_ObjC)` in the code.

I can think of a few options...

1. Leave this as it is currently, as my understanding is correct.
2. `#if` out the `String` extension when using ObjC runtime.  My best guess is this would be to test the build for an Apple platform rather than actually making something usable, since Apple platforms have their own implementation of Foundation.
3. `#if` the `String` extension to provide an autorelease implementation when using ObjC runtime.  Is this even possible?

Molanda:
To play it safe, I picked #2.  This allows it to build in Xcode using stock Swift without the `_CVarArgObject` protocol.

Molanda:
@millenomi Would you mind running the tests when you get a moment?

MaxDesiatov:
@swift-ci please test

millenomi:
@Molanda we support running SCF for debugging/patch writing purposes only on Darwin (via the Xcode workspace/project in the repository), so please make sure you don't break that. Note that it's a bit of a weird environment, but you will not have Swift code importing the 'real' Darwin Foundation in your process, so it will likely be safe to have the extension around — and you may actually need it since you are using SCF and the bundled CF.

millenomi:
@swift-ci please test

Molanda:
Thank you.

The macOS build failed with...

```
/Users/buildnode/jenkins/workspace/swift-corelibs-foundation-PR-macOS/branch-master/swift-corelibs-foundation/Sources/Foundation/NSString.swift:1651:59: error: class 'CFString' cannot be used in an '@inlinable' function because 'CoreFoundation' was imported implementation-only
18:25:01         return _encodeBitsAsWords(unsafeBitCast(self, to: CFString.self))
```

It seems `@_implementationOnly` is a recent change.  Any advise on how this should be handled now?

For `%@` to work, `_cVarArgEncoding` needs to return a `CFString`.

Molanda:
I can build with the  `@_implementationOnly` change if I remove `@inlinable` from the extension functions.

On the Swift stdlib side, lorentey had made this suggestion...

>...and I'm 99% sure we'd also want to remove @inlinable from all builder methods. (Exposing the vararg builder's class members in the ABI was a regrettable step.)

@millenomi do you agree with removing `@inlinable` or do you have a different suggestion?

(An unrelated note... while looking at the recent merges, I noticed RunLoop.swift has `os(Linux)` but not `os(Android)` in the `#if` conditional.)

millenomi:
@Molanda sorry for the on-and-off; yes, please remove `@inlinable` here, or use `unsafeBitCast(…, to: AnyObject.self)` instead.

millenomi:
(CFString and AnyObject have the same representation — a pointer.)

MaxDesiatov:
@swift-ci please test

millenomi:
The failing test is LLDB, not us.

millenomi:
@swift-ci please test linux

Molanda:
@millenomi Looks good. Thank you for the help!

PR #2798, opened 2020-05-18
tests: add a test case that breaks on Windows

compnerd:
This seems to be failing on Windows, identified through YAMS.  The test
case is reduced from a test case originally provided by JP Simard
(@jpsim).

PR #2822, opened 2020-06-14
ICU: Disable a test in preperation for upgrading to ICU67

spevans:
- Converting Double.leastNonzeroMagnitude with a Locale would result
  in the value '0E+00' instead of the correct '5E-324' in some versions
  of ICU (62-66). Disable the test against '0E+00' so that it wont break
  upgrading to ICU67.


spevans:
@swift-ci please test

spevans:
@swift-ci please test

spevans:
@swift-ci please test

compnerd:
This should repair the Windows CI for Foundation which is using ICU 67 now to deal with #2798

compnerd:

compnerd:

spevans:
@swift-ci please test Linux

PR #2824, opened 2020-06-15
Make ProcessInfo.operatingSystemVersionString useful on Linux and Windows

gwynne:
- On Linux, `operatingSystemVersionString` was previously consulting `/proc/version_signature`, and giving up immediately if that failed for any reason. This file is a duplicate of the value of `uname -r` and has not been present for several major releases of Ubuntu or Debian. The Linux case was rewritten to return the raw value of `utsname.release`, which corresponds to `uname -r` _and_ is the value consulted already by the Linux implementation of `operatingSystemVersion`.

- On Windows, `operatingSystemVersionString` was effectively completely unimplemented - falling back to `CFCopySystemVersionString()` on any non-Darwin platform is the same as having no fallback. The only non-deprecated way to retrieve a system version on Windows that I am aware of (at least without getting into the intensely dubious proposition of trying to invoke .NET classes from Swift) is the routine already being used by `operatingSystemVersion`: `RtlGetVersion()`. The boilerplate necessary for correctly loading `ntdll.dll` and retrieving the symbol from it was refactored out into a private accessor invoked by both `operatingSystemVersion` and `operatingSystemVersionString`. The former's behavior remains unchanged. The latter now constructs as reasonable a string as it can, based almost entirely on the contradictory and poorly-proofread reference table provided by Microsoft (the appropriate link is provided in the code).


jakepetroules:
> The only non-deprecated way to retrieve a system version on Windows that I am aware of (at least without getting into the intensely dubious proposition of trying to invoke .NET classes from Swift) is the routine already being used by operatingSystemVersion: RtlGetVersion().

I have an unfortunate amount of experience with this and can confirm this is the right API to call. :D

Sources/Foundation/ProcessInfo.swift
```diff
@@ -80,11 +80,68 @@ open class ProcessInfo: NSObject {
...
 #if os(Linux)
-        let version = try? String(contentsOf: URL(fileURLWithPath: "/proc/version_signature", isDirectory: false), encoding: .utf8)
-        return version ?? fallback
+        var utsNameBuffer = utsname()
+        guard uname(&utsNameBuffer) == 0 else {
+            return fallback
+        }
+        let release = withUnsafePointer(to: &utsNameBuffer.release.0) { String(cString: $0) }
+
+        return release
```
spevans, line 113: Might be worth adding a check that `guard release != "" else { return fallback }` just incase `utsname()` breaks.

spevans:
@swift-ci please test

spevans:
@compnerd can you please test  this on windows when you get the chance, thanks.

Sources/Foundation/ProcessInfo.swift
```diff
@@ -80,11 +80,68 @@ open class ProcessInfo: NSObject {
...
+            case (6, 0) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += "Windows Vista"
+            case (6, 0): versionString += "Windows Server 2008"
+            case (6, 1) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += "Windows 7"
+            case (6, 1): versionString += "Windows Server 2008 R2"
+            case (6, 2) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += "Windows 8"
+            case (6, 2): versionString += "Windows Server 2012"
+            case (6, 3) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += "Windows 8.1"
+            case (6, 3): versionString += "Windows Server 2012 R2" // We assume the "10,0" numbers in the table for this are a typo
+            case (10, 0) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += "Windows 10"
+            case (10, 0): versionString += "Windows Server 2019" // The table gives identical values for 2016 and 2019, so we just assume 2019 here
```
jakepetroules, line 139: 2016 is 10.0.14393 and 10.0.16299, 2019 is 10.0.17763. Versions are also identified by year and month, for example, "Windows Server 2016, version 1709") and IIRC you can get the latter part from the registry key HKLM\Software\Microsoft\Windows\Windows NT\CurrentVersion\ReleaseID for Windows 10 and up.

Sources/Foundation/ProcessInfo.swift
```diff
@@ -80,11 +80,68 @@ open class ProcessInfo: NSObject {
...
     open var operatingSystemVersionString: String {
         let fallback = "Unknown"
 #if os(Linux)
-        let version = try? String(contentsOf: URL(fileURLWithPath: "/proc/version_signature", isDirectory: false), encoding: .utf8)
-        return version ?? fallback
+        var utsNameBuffer = utsname()
+        guard uname(&utsNameBuffer) == 0 else {
+            return fallback
+        }
+        let release = withUnsafePointer(to: &utsNameBuffer.release.0) { String(cString: $0) }
```
jakepetroules, line 111: Is `uname -r` really the right thing for Linux? Just as we return the macOS/iOS version on Apple platforms rather than the Darwin version, perhaps we should return the distribution version instead of the Linux version? (i.e. /etc/lsb_release, /etc/debian_version and so on)

Here's the approach we took in Qt: https://github.com/qt/qtbase/blob/dev/src/corelib/global/qglobal.cpp -- look for findUnixOsVersion and start from there.

Sources/Foundation/ProcessInfo.swift
compnerd: `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion`.

```
> reg query "HKLM\Software\Microsoft\Windows NT\CurrentVersion" /v ReleaseID

HKEY_LOCAL_MACHINE\Software\Microsoft\Windows NT\CurrentVersion
    ReleaseID    REG_SZ    2004
```

That is 10.0.19465 though, so, the mapping seems slightly confusing.

compnerd:
@spevans - https://dev.azure.com/compnerd/swift-build/_build/results?buildId=33744&view=results

Sources/Foundation/ProcessInfo.swift
compnerd: The `/etc/*-release` is far more free form.  Not everyone really bothers with LSB, so, at least Gentoo, exherbo, Fedora, and debian have `/etc/{distro}-release`.  Its also interesting for distributions which are rolling distributions and thus have no real version number associated with them.  So although your stage may be versioned, the state can vary drastically between installations.

gwynne: My primary aim with this change was to provide _some_ kind of remotely usable value - it's not usually meaningful to make decisions based on a Linux environment's version in a generic sense (i.e. when not limited to a specific distro) anyway, so accuracy wasn't a primary concern; it was more about addressing that the `/proc/version_signature` approach yields nothing at all on most current Linuxes. I could certainly write the code to loop over the set of, say, `/etc/os-release`, `/etc/system-release`, `/etc/lsb-release`, `/etc/debian_version`, `/etc/redhat-release`, `/etc/alpine-release`, `/etc/centos-release`, and so forth (probably in more or less that order) - just `os-release` by itself would already cover everything supported by Swift 5. The question to me is, does doing so confer useful enough information to be worth even that much bother versus just letting `uname*(` give a number?

spevans: I dont have a particularly strong view on this but here's some thoughts:

1. I see `operatingSystemVersionString` as being a stringification of `operatingSystemVersion` - which on Linux is the kernel version - so on the one hand it should just be that and say `Version 5.32.0` or whatever. Note that on macOS today it says something like `Version 10.14.6 (Build 18G5033)`. The `operatingSystemVersion` tuple is actually useful on Linux for checking the availability of system calls.

2. However it should probably display something more than just the kernel version and most users would probably be confused by seeing `Linux 5.2.32` rather than `Ubuntu 20.04`. So I think the code should do a best effort to try and find the distribution name/version and fallback to the generic `Linux <kernel_version>` as a last resort (and make sure this is cached in an ivar). I think we want to avoid this being a User-Agent type thing where it gets parsed and programs become reliant on the format.
 
3. I think a better long term solution for Linux, and probably Windows, is to pitch to add something into the new [System](https://developer.apple.com/documentation/system) module that can return a more detailed answer including enums for the distribution etc so that developers move away from this property.

gwynne: @spevans I concur on all counts; I'll add basic best-effort distro detection with the uname fallback for now. The long-term concern is out of scope for this PR, though; this is just hoping to provide parity with yet another long-standing Foundation API (and I **do** mean long-standing - this API sits right next to `operatingSystem()` and `operatingSystemName()`, of all things! 🤣). What is the best path for making such a pitch for the `System` module?

spevans: @gwynne I dont know how the `System` module is going to be used, Ive only seen the File stuff for it in the documentation. Maybe write a standalone package to prototype some ideas especially with enumerating the different versions of Linux and Windows.

Getting back to this PR, are there any tests that can be written to check the functionally taking into account that the tests could be run on various Linux/Windows versions? Maybe doing a simple check that one of the supported OSes (Linux/Ubuntu/Debain/Centos etc) at least appears in the result ?

Sources/Foundation/ProcessInfo.swift
```diff
@@ -80,11 +80,68 @@ open class ProcessInfo: NSObject {
...
+#endif
+
     open var operatingSystemVersionString: String {
         let fallback = "Unknown"
 #if os(Linux)
-        let version = try? String(contentsOf: URL(fileURLWithPath: "/proc/version_signature", isDirectory: false), encoding: .utf8)
-        return version ?? fallback
+        var utsNameBuffer = utsname()
+        guard uname(&utsNameBuffer) == 0 else {
+            return fallback
```
spevans, line 109: I would probably change the fallback from `Unknown` to at least `Linux` or `Windows` depending on the `#if os(..)`

millenomi:
@gmittert this is pending a change still, correct?

millenomi:
My bad — that was intended to be @gwynne.

gwynne:
> My bad — that was intended to be @gwynne.

Yeah, I haven't had a chance to circle back to this to do the fixups @spevans and I discussed yet, though I still plan to.

millenomi:
If you do and this stalls, please @ me.

gwynne:
@spevans @millenomi (Finally) updated this with the changes that were discussed. Please let me know if I missed anything!

@compnerd Please test on Windows when you have an opportunity!

gwynne:
@swift-ci please test

Sources/Foundation/ProcessInfo.swift
```diff
@@ -80,15 +80,91 @@ open class ProcessInfo: NSObject {
...
+            case (10, 0): versionString += " Server 2019" // The table gives identical values for 2016 and 2019, so we just assume 2019 here
+            case let (maj, min): versionString += " \(maj).\(min)" // If all else fails, just give the raw version number
+        }
+        versionString += " (build \(osVersionInfo.dwBuildNumber))"
+        // For now we ignore the `szCSDVersion`, `wServicePackMajor`, and `wServicePackMinor` values.
+        return versionString
 #else
-        return CFCopySystemVersionString()?._swiftObject ?? fallback
+        // On other systems at least return something.
+        return "Unknown"
```
spevans, line 164: Could these be made slightly more specific for  `os(FreeBSD)`, `os(OpenBSD)` and  `os(Android)` since we at least know the os name?

millenomi:
Once you're done, make sure that you un-draft this PR so that I can merge it.

gwynne:
@swift-ci please test

gwynne:
@millenomi @spevans Assuming I haven't missed anything, and that CI passes, I think this is ready.

gwynne:
@swift-ci test

millenomi:
@swift-ci please test

spevans:
@swift-ci test linux

PR #2825, opened 2020-06-16
Flesh out URLError information upon URLSessionTask cancellation

glessard:
The `URLError` received upon `URLSessionTask` cancellation by a `URLSessionTaskDelegate` or completion handler has a blank `userInfo` dictionary. This is an inconsistency when compared with macOS's Foundation and it makes actual error handling harder than it could be.

This PR fills it out to match the fields filled out by macOS Foundation.

I previously reported this as https://bugs.swift.org/browse/SR-13017, and I believe this fixes that (if a fix is even desirable, that is.)

### DIFF
--- a/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
+++ b/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
@@ -364,11 +364,20 @@ open class URLSessionTask : NSObject, NSCopying {
      */
     open func cancel() {
         workQueue.sync {
-            guard self.state == .running || self.state == .suspended else { return }
-            self.state = .canceling
+            let canceled = self.syncQ.sync { () -> Bool in
+                guard self._state == .running || self._state == .suspended else { return true }
+                self._state = .canceling
+                return false
+            }
+            guard !canceled else { return }
             self._getProtocol { (urlProtocol) in
                 self.workQueue.async {
-                    let urlError = URLError(_nsError: NSError(domain: NSURLErrorDomain, code: NSURLErrorCancelled, userInfo: nil))
+                    var info = [NSLocalizedDescriptionKey: "\(URLError.Code.cancelled)" as Any]
+                    if let url = self.originalRequest?.url {
+                        info[NSURLErrorFailingURLErrorKey] = url
+                        info[NSURLErrorFailingURLStringErrorKey] = url.absoluteString
+                    }
+                    let urlError = URLError(_nsError: NSError(domain: NSURLErrorDomain, code: NSURLErrorCancelled, userInfo: info))
                     self.error = urlError
                     if let urlProtocol = urlProtocol {
                         urlProtocol.stopLoading()
--- a/Tests/Foundation/Tests/TestURLSession.swift
+++ b/Tests/Foundation/Tests/TestURLSession.swift
@@ -1443,6 +1443,9 @@ class TestURLSession: LoopbackServerTest {
             XCTAssertNotNil(error as? URLError)
             if let urlError = error as? URLError {
                 XCTAssertEqual(urlError._nsError.code, NSURLErrorCancelled)
+                XCTAssertEqual(urlError.userInfo[NSURLErrorFailingURLErrorKey] as? URL, URL(string: urlString))
+                XCTAssertEqual(urlError.userInfo[NSURLErrorFailingURLStringErrorKey] as? String, urlString)
+                XCTAssertEqual(urlError.localizedDescription, "cancelled")
             }
 
             expect.fulfill()


glessard:
There are 6 other `URLError` constructors in the FoundationNetworking project, and their `userInfo` situation is quite inconsistent. Would it be desirable to change that?

spevans:
The testcase is ok, I ran it under `DarwinCompatibilityTests` and it works correctly against Darwin.

However the test failed on both Linux and macOS. Note that currently the `TestURLSession` tests are disabled and need to be re-enabled in the `Tests/Foundation/main.swift` file. `URLSession` still has some other issues that makes the tests generally flaky. 

Note in this case it looks like `request?.url`  was `nil` so the `failingURL` and `failureURLString` entries werent added.

glessard:
I thought it ran in the TestFoundation target of the Xcode project, but maybe it didn't. (I should have tried to crash the test to make sure!) Building on Linux failed yesterday, and didn't have enough time to fix that so I crossed my fingers and submitted it. I'll try again.

glessard:
@spevans I haven't had time to solve my Linux build, but now I've gotten the test to properly run from Xcode.

spevans:
I think we should fix the `URLError.failingURL` and `.failureURLString` properties that your original tests exposed. The following seems to fix it in my tests:
```
diff --git a/Sources/Foundation/NSError.swift b/Sources/Foundation/NSError.swift
index 0b55deea..ff71279e 100644
--- a/Sources/Foundation/NSError.swift
+++ b/Sources/Foundation/NSError.swift
@@ -919,12 +919,12 @@ extension URLError {
 
     /// The URL which caused a load to fail.
     public var failingURL: URL? {
-        return _nsUserInfo[NSURLErrorFailingURLErrorKey._bridgeToObjectiveC()] as? URL
+        return _nsUserInfo[NSURLErrorFailingURLErrorKey] as? URL
     }
 
     /// The string for the URL which caused a load to fail.
     public var failureURLString: String? {
-        return _nsUserInfo[NSURLErrorFailingURLStringErrorKey._bridgeToObjectiveC()] as? String
+        return _nsUserInfo[NSURLErrorFailingURLStringErrorKey] as? String
     }
 }
```

glessard:
I wasn't sure what `_bridgeToObjectiveC()` was supposed to do, so I didn't go that route. Also: should it be in a different PR?

This being said, there are other things with URLError: none of the properties added in iOS 13 are on it yet. Obviously they're for either darwin-specific or not-yet-supported things, but shouldn't they be there, returning `nil`?

spevans:
Yes I think a separate PR to fix up the properties in `URLError` would be a good start. I think the 
```
private var _nsUserInfo: [AnyHashable : Any] {
        return _nsError.userInfo
    }
```
can be changed to a `[String: Any]` and then the properties fixed up. 

It just requires adding some tests constructing `URLError` and then testing the properties and validating the tests in the `DarwinCompatibilityTests` project which will run the tests against the native Foundation. It would be good to add in the missing properties as well.

Sources/FoundationNetworking/URLSession/URLSessionTask.swift
```diff
@@ -364,11 +364,20 @@ open class URLSessionTask : NSObject, NSCopying {
      */
     open func cancel() {
         workQueue.sync {
-            guard self.state == .running || self.state == .suspended else { return }
-            self.state = .canceling
+            let (canceled, request) = self.syncQ.sync { () -> (Bool, URLRequest?) in
+                guard self._state == .running || self._state == .suspended else { return (true, nil)}
+                _state = .canceling
+                return (false, _currentRequest ?? originalRequest)
```
glessard, line 370: @spevans: Should the URL included in the error just be the original URL? I'm not certain of the expectation in the case of redirections, so I wonder if this is the correct data.

glessard: If you think `_currentRequest ?? originalRequest` is correct, then this PR should be ready.

spevans: Looking at some of the other tests with errors in redirection it looks like you should just use `originalRequest?.url` and dont bother with `_currentRequest`. Im pretty sure that `originalRequest` is never `nil` at the point the task can be cancelled so there is no need for the `??` fallback

glessard: Great. I'll change it to just use `originalRequest`.

spevans:
@swift-ci test

spevans:
I just tested it locally on Linux with the `TestURLSession` tests enabled and it passed ok.

spevans:
@swift-ci test

glessard:
Thanks!

PR #2826, opened 2020-06-18
Fix more warnings.

spevans:
- Use 'let' instead of 'var' on some immutable variables.

- TestJSONEncoder: Change a test struct from Codable to Encodable
  as the contents are constant and cant be used for Decoding.


spevans:
@swift-ci test

spevans:
@swift-ci test macos

spevans:
@swift-ci test

PR #2827, opened 2020-06-19
SR-13015: Decimal calculation incorrect on Linux

buttaface:
Fix typo and logic error in the overflow check in `integerMultiply()`, which was screwing up Decimal calculations under certain conditions.

I was able to reproduce on Android and spent hours tracking this down on there. I'm guessing it only worked on macOS because this code is unused on there.

### DIFF
--- a/Sources/Foundation/Decimal.swift
+++ b/Sources/Foundation/Decimal.swift
@@ -1025,7 +1025,7 @@ fileprivate func integerMultiply<T:VariableLengthNumber>(_ big: inout T,
                 accumulator = UInt32(carry) + bigij + UInt32(right[j]) * UInt32(left[i])
                 carry = UInt16(truncatingIfNeeded:accumulator >> 16)
                 big[i+j] = UInt16(truncatingIfNeeded:accumulator)
-            } else if carry != 0 || (right[j] == 0 && left[j] == 0) {
+            } else if carry != 0 || (right[j] > 0 && left[i] > 0) {
                 return .overflow
             }
         }
--- a/Tests/Foundation/Tests/TestDecimal.swift
+++ b/Tests/Foundation/Tests/TestDecimal.swift
@@ -342,7 +342,7 @@ class TestDecimal: XCTestCase {
         XCTAssertEqual(otherBits + Decimal(1), highBit)
     }
 
-    func test_Misc() {
+    func test_Misc() throws {
         XCTAssertEqual(.minus, Decimal(-5.2).sign)
         XCTAssertEqual(.plus, Decimal(5.2).sign)
         var d = Decimal(5.2)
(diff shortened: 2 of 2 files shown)


Tests/Foundation/Tests/TestDecimal.swift
```diff
@@ -99,6 +99,10 @@ class TestDecimal: XCTestCase {
         XCTAssertEqual(.lossOfPrecision, NSDecimalAdd(&result, &one, &addend, .plain), "1 + 1e-39")
         XCTAssertEqual("1", result.description)
         XCTAssertEqual(.orderedSame, NSDecimalCompare(&one, &result), "1 + 1e-39")
+
+        let a = Decimal(string: "498.7509045")!
+        let b = Decimal(string: "8.453441368210501065891847765109162027")!
+        XCTAssertEqual(a + b, Decimal(string: "507.2043458682105010658918477651091")!)
```
spevans, line 105: Rather than using force unwrap, mark the test function as `throws` and use:
`let a = try XCTUnwrap(Decimal(string: "498.7509045"))` etc.

spevans:
@swift-ci test

buttaface:
Updated with correct overflow check.

Sources/Foundation/Decimal.swift
```diff
@@ -1025,7 +1025,7 @@ fileprivate func integerMultiply<T:VariableLengthNumber>(_ big: inout T,
                 accumulator = UInt32(carry) + bigij + UInt32(right[j]) * UInt32(left[i])
                 carry = UInt16(truncatingIfNeeded:accumulator >> 16)
                 big[i+j] = UInt16(truncatingIfNeeded:accumulator)
-            } else if carry != 0 || (right[j] == 0 && left[j] == 0) {
+            } else if carry != 0 || (right[j] > 0 && left[i] > 0) {
```
buttaface, line 1028: This will overflow if the accumulator above would be non-zero, ie if there's a carry or both right and left elements that are multiplied are non-zero.

spevans:
Could you add in the extra tests in https://gist.githubusercontent.com/spevans/a5c9da89e91538abc688c5498fe4ed28/raw/e23ca9cdc21b87e2bcc1880ddafc5e01624bd9ce/sr13015.diff and squash it into one commit for easier back porting

spevans:
@swift-ci test linux

buttaface:
Sure, will do.

buttaface:
Rather than checking the same Decimal normalization in three different places, I removed my test and added your more extended tests instead.

spevans:
@swift-ci test linux

spevans:
@swift-ci test

buttaface:
You mentioned backporting this, should I submit a pull for the 5.3 branch? I presume 5.2 releases are done, or will that take this too?

spevans:
Lets get a backport for 5.3 going, Im not sure if it will make it into the release or will end up in 5.3.1 but having it ready to go will be useful, thanks.

PR #2829, opened 2020-06-21
[5.3] SR-13015: Decimal calculation incorrect on Linux

buttaface:
From #2827.


spevans:
@swift-ci test

compnerd:
CC: @millenomi

After staring at the overflow condition for a while, I think that this is right.  It corrects an incorrect result and seems relatively localized.  I think that we should go ahead and merge this.

PR #2830, opened 2020-06-21
[CMake] Handle CURL package version defined by find_package CONFIG mode

lxbndr:
`find_package(CURL CONFIG)` always uses CMake modules bundled with CURL. If search is successful, `CURL_VERSION` is defined, but we are expecting `CURL_VERSION_STRING`. This makes Foundation think that it deals with old CURL version, disabling some features.

This change sets `CURL_VERSION_STRING` using `CURL_VERSION` provided by `find_package(CURL CONFIG)` .


lxbndr:
cc @compnerd

CoreFoundation/CMakeLists.txt
```diff
@@ -27,8 +27,11 @@ find_package(Threads REQUIRED)
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
   find_package(LibXml2 REQUIRED)
   find_package(CURL CONFIG)
-  if(NOT CURL_FOUND)
+  if(CURL_FOUND)
+    set(NS_CURL_VERSION ${CURL_VERSION})
+  else()
     find_package(CURL REQUIRED)
+    set(NS_CURL_VERSION ${CURL_VERSION_STRING})
```
compnerd, line 34: I’m not sure I see the value in the NS prefix here.  What’s the motivation for that?

lxbndr: Nothing special actually. Mostly inspired by NS_CURL_ASSUME_FEATURES_MISSING, NS_CURL_MISSING_MAX_HOST_CONNECTIONS, etc. Any suggestions are appreciated.

compnerd: The difference is that `CURL_VERSION` and `CURL_VERSION_STRING` is that one is a string representation I believe.  I think that we invert the case where the `CONFIG` check just does a `set(CURL_VERSION_STRING "${CURL_VERSION}")` would be simpler.

lxbndr: I didn't want to define option originated from external module. But why not. As it really makes everything simpler. Will do.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

spevans:
@swift-ci please test

compnerd:

compnerd:

PR #2832, opened 2020-06-24
Gets rid of using deprecated String.Index(utf16Offset:in:)

valeriyvan:
Gets rid of using deprecated String.Index(utf16Offset:in:) in URLComponents


spevans:
@swift-ci please test

valeriyvan:
ping

valeriyvan:
ping

PR #2833, opened 2020-06-24
[5.3] Return success from `FileManager` recursive directory creation even if a directory was created by another thread concurrently.

spevans:
If another thread creates one of the same directories in the target path, the `!fileExists` check may pass while the `mkdir` call fails with `EEXIST`. However, if the existing file is a directory, then that should still be a success.

Fixes [SR-12272](https://bugs.swift.org/browse/SR-12272).

(cherry picked from commit b6eea02e84ea46e8bc36588f250d36930fe2ee80)


spevans:
@swift-ci test

PR #2836, opened 2020-06-30
DarwinCompatibilityTest: Update working test list and shims

spevans:
- DarwinShims: Remove the extensions to FileHandle and Scanner as the
  missing functions are in the latest macOS 10.15 release.

- TestsToSkip: Update the list of tests to skip when running
  DarwinCompatibilityTests/xcode-build.sh to be more granular and
  also remove tests that now pass.

- FileHandle.readInBackgroundAndNotify(): If the read fails the userInfo
  NSFileHandleNotificationDataItem entry should be an empty Data() not
  nil.


spevans:
@swift-ci test

PR #2837, opened 2020-07-04
NSDecimalNumber: Fix comparison with NSNumber

spevans:
- NSDecimalNumber: When compare(_:) / isEqual(_:) has a rhs which is an
  NSNumber that cannot be cast as an NSDecimalNumber, use
  NSNumber.compare() to test them.

- NSNumber: Add missing isEqual(to:)


spevans:
@swift-ci please test

PR #2838, opened 2020-07-05
SR-13027: Fix URL encoding in redirects

agisboye:
This resolves a number of issues with how redirects are handled.

Query and fragment components were set as part of the path, resulting in them being percent-encoded even though they already are. Now, the path, query and fragment are set separately.

The authority component (username:password) from the original URL was not used in the new URL. Now, the new URL is constructed from the original, which ensures that the authority component is carried over.

Additionally, I've refactored a few things that were implemented in a slightly convoluted way.


agisboye:
@swift-ci please test

spevans:
Can you add some tests that test the fix? The best thing if you have access to macOS is to add them to the `DarwinCompatibility` xcode project to check the tests work correctly against macOS. Unfortunately `TestURLSession` is disabled in the normal tests as some of them break but I can check them manually on Linux, thanks.

agisboye:
Sure thing. I've added `test_httpRedirectionWithEncodedQuery`.

millenomi:
@swift-ci please test

PR #2839, opened 2020-07-05
build: centralise the install logic for the Swift modules

compnerd:
Add support for installation for macOS targets to the CMake
installation.  Centralise the logic for the installation of the files.
This properly handles the installation of the swiftdoc, swiftinterface,
swiftmodule and the runtime libraries across all the platforms in either
static or shared configurations.

### DIFF
--- a/Sources/Foundation/CMakeLists.txt
+++ b/Sources/Foundation/CMakeLists.txt
@@ -161,12 +161,4 @@ endif()
 
 
 set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS Foundation)
-install(TARGETS Foundation
-  ARCHIVE DESTINATION lib/swift$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:_static>/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>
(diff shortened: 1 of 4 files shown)


compnerd:
CC: @drodriguez @spevans

compnerd:
@swift-ci please test

spevans:
Should sclf be installed on macOS?, I could see it breaking things especially if it was installed as `Foundation`. Its only really used on macOS for development and testing.

compnerd:
The logic to install the modules is completely generic.  The SwiftSupport.cmake can be copied across repositories without alteration.

I believe that we setup the output names to use SwiftFoundation, so it shouldn't get installed as Foundation.

lxbndr:
Could this logic potentially become a part of Swift support in CMake itself? I am not familiar with CMake policies related to such things, but it seems natural to handle target installation as part of language support. Would be really cool if we have it out of the box.

compnerd:
@lxbndr - I have avoided putting this into CMake itself intentionally.  I have some work which will unify the Darwin and non-Darwin paths, which again will update this.  At that point, we will have a unified installation style.  At that point, it will be better to actually add support for Swift to the `install` command rather than rely on special handling and then we can completely remove `SwiftSupport.cmake` once we have a new enough CMake.

spevans:
I have no objections to this patch - I just wanted to note that installing on macOS could cause issues. But since its CMake changes Im happy to leave them in @compnerd 's capable hands :)

PR #2849, opened 2020-08-03
Install uuid when statically building Foundation

drexin:
When building the shared libs, this will be included in `libFoundation.so`, but when building the static libs, we have to install it with the other libraries, so it can be found when linking user binaries.


drexin:
@swift-ci test linux

drexin:
@swift-ci test linux

drexin:
@compnerd fixed

compnerd:
@swift-ci please test

PR #2850, opened 2020-08-04
Add modulemaps that work for statically compiled Foundation

drexin:
The static libraries need to link against additional dependencies, which are listed in the newly added modulemaps

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,39 +62,44 @@ if(ENABLE_TESTING)
   add_subdirectory(Tests)
 endif()
 
+set(swift_lib_dir "lib/swift")
+
 if(NOT BUILD_SHARED_LIBS)
+  set(swift_lib_dir "lib/swift_static")
+  
   set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS
     CoreFoundation CFXMLInterface CFURLSessionInterface)
   install(TARGETS CoreFoundation CFXMLInterface CFURLSessionInterface
     DESTINATION lib/swift_static/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>)
 endif()
+
 # TODO(compnerd) install as a Framework as that is how swift actually is built
 install(DIRECTORY
           ${CMAKE_CURRENT_BINARY_DIR}/CoreFoundation.framework/Headers/
         DESTINATION
-          lib/swift/CoreFoundation
+          ${swift_lib_dir}/CoreFoundation
         FILES_MATCHING PATTERN "*.h")
 install(FILES
-          CoreFoundation/Base.subproj/module.map
+          CoreFoundation/Base.subproj/$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:static/>module.map
         DESTINATION
-          lib/swift/CoreFoundation)
+          ${swift_lib_dir}/CoreFoundation)
 install(DIRECTORY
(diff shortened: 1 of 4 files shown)


drexin:
This so far only works for `-static-stdlib`. To make `-static-executable` work, we would need to list all transitive dependencies.

drexin:
@swift-ci test linux

CMakeLists.txt
```diff
@@ -62,39 +62,44 @@ if(ENABLE_TESTING)
...
+set(swift_lib_dir "lib/swift")
+
 if(NOT BUILD_SHARED_LIBS)
+  set(swift_lib_dir "lib/swift_static")
+  
   set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS
     CoreFoundation CFXMLInterface CFURLSessionInterface)
   install(TARGETS CoreFoundation CFXMLInterface CFURLSessionInterface
-    DESTINATION lib/swift_static/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>)
+    DESTINATION ${swift_lib_dir}/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>)
```
compnerd, line 73: This seems incorrect.  `CoreFoundation` is always built statically for Foundation, so this should always be `swift_static` right?

CMakeLists.txt
```diff
@@ -62,39 +62,44 @@ if(ENABLE_TESTING)
...
         DESTINATION
-          lib/swift/CFXMLInterface
+          ${swift_lib_dir}/CFXMLInterface
         FILES_MATCHING PATTERN "*.h")
 install(FILES
-          CoreFoundation/Parsing.subproj/module.map
+          CoreFoundation/Parsing.subproj/$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:static/>module.map
         DESTINATION
-          lib/swift/CFXMLInterface)
+          ${swift_lib_dir}/CFXMLInterface)
```
compnerd, line 103: Do we really need two copies of the headers?  I think that I should revive my work to remove the private headers at least.  We really want to stop propagating the CoreFoundation headers into the toolchain I think.

CC: @millenomi

CMakeLists.txt
drexin: Good catch!

drexin: Oh, actually this is still correct because it will only be installed if we are compiling statically, in which case the variable is already set to `lib/swift_static`. I can change it back if you want.

compnerd: That would be great.

drexin:
@swift-ci test linux

drexin:
@swift-ci test linux

drexin:
@swift-ci test linux

drexin:
@compnerd I'm not sure what you mean by pulling the definition of `swift_lib_dir` out. It's defined pretty much right before it's being used.

drexin:
@swift-ci test linux

drexin:
@swift-ci test linux

drexin:
@swift-ci test

buttaface:
Nice work, I just tried this out with the latest 8/18 official trunk snapshot build for Ubuntu and it works great.

Any plans to backport these pulls to the 5.3 branch? I will do it for the Android builds I put out if not.

drexin:
@buttaface Thanks for the feedback! I am planning to backport this to 5.3 as well.

buttaface:
Great, looking forward to it.

PR #2851, opened 2020-08-10
Create a file with permissions that respects umask

niw:
**Problem**

`Data.write(to:)` is a only method in the Foundation that can create a
regular file.
However, it ignores `umask` and always set 0600 permission unlike
macOS Foundation, which respects process `umask`.

**Solution**

1. With `.atomic` write option

    It uses `mkstemp(3)` in `_NSCreateTemporaryFile`, which is always
    creating a file with 0600 permission, if the system follows
    the latest POSIX specification or the permission is undefined.

    On macOS Foundation, therefore `_NSCreateTemporaryFile` uses
    `mktemp(3)` and `open(2)` instead to respect `umask`.

2. Without `.atomic` write option

    It uses `0o600` even if it uses `open(2)` that respects `umask`.
    Simply gives `0o666` instead.

This is a bug caused by previous commit in
apple/swift-corelibs-foundation#1876.

Swift JIRA is https://bugs.swift.org/browse/SR-13307.

### DIFF
--- a/Sources/Foundation/NSData.swift
+++ b/Sources/Foundation/NSData.swift
@@ -434,8 +434,7 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
         }
 
         let fm = FileManager.default
-        // The destination file path may not exist so provide a default file permissions of RW user only
-        let permissions = (try? fm._permissionsOfItem(atPath: path)) ?? 0o600
+        let permissions = try? fm._permissionsOfItem(atPath: path)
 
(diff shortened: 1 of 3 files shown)


niw:
@swift-ci test linux

spevans:
Could you add a regression test that validates that a file is created with the correct permissions, thanks

spevans:
@swift-ci test

niw:
@spevans Sure! the test need to change process’s `umask` which is global therefore it need to be reentrant.

millenomi:
@swift-ci please test

PR #2853, opened 2020-08-19
NFC: Fix -Welaborated-enum-base error

gmittert:
Recent versions of clang default -Welaborated-enum-base to -Werror.

### DIFF
--- a/CoreFoundation/AppServices.subproj/CFNotificationCenter.h
+++ b/CoreFoundation/AppServices.subproj/CFNotificationCenter.h
@@ -22,7 +22,7 @@ typedef struct CF_BRIDGED_MUTABLE_TYPE(id) __CFNotificationCenter * CFNotificati
 
 typedef void (*CFNotificationCallback)(CFNotificationCenterRef center, void *observer, CFNotificationName name, const void *object, CFDictionaryRef userInfo);
 
-typedef CF_ENUM(CFIndex, CFNotificationSuspensionBehavior) {
+CF_ENUM(CFIndex, CFNotificationSuspensionBehavior) {
     CFNotificationSuspensionBehaviorDrop = 1,
         // The server will not queue any notifications with this name and object while the process/app is in the background.
     CFNotificationSuspensionBehaviorCoalesce = 2,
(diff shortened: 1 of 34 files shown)


gmittert:
@swift-ci please test

millenomi:
No, this is a source-breaking regression of some kind — we support this pattern explicitly for Darwin development and it should not be an error.

millenomi:
Pinging clang.

millenomi:
@gmittert after consulting with clang: we need a patch but not a patch that touches the CF_ENUM use sites. I'll work to have one shortly.

millenomi:
But I cannot take this one. Closing.

gmittert:
@millenomi My bad about the source breaking, I should know not to do that :(. Alternatively something to essentially throw away the extra typedef at the use sites might work and then wouldn't cause source breaking. Something like
```
#if (__cplusplus)
#define __CF_NAMED_ENUM(_type, _name)     int __CF##_name; enum __CF_ENUM_ATTRIBUTES _name : _type
#define __CF_ANON_ENUM(_type)             enum __CF_ENUM_ATTRIBUTES : _type
#define CF_CLOSED_ENUM(_type, _name)      int  __CF##_name; enum __CF_CLOSED_ENUM_ATTRIBUTES _name : _type
#define CF_OPTIONS(_type, _name) _type _name; int __CF##_name; enum __CF_OPTIONS_ATTRIBUTES : _type
#else
#define __CF_NAMED_ENUM(_type, _name)     int __CF##_name; enum __CF_ENUM_ATTRIBUTES _name : _type; typedef enum _name _name; enum _name : _type
#define __CF_ANON_ENUM(_type)             enum __CF_ENUM_ATTRIBUTES : _type
#define CF_CLOSED_ENUM(_type, _name)      int __CF##_name; enum __CF_CLOSED_ENUM_ATTRIBUTES _name : _type enum _name _name; enum _name : type
#define CF_OPTIONS(_type, _name) int __CF##_name; enum __CF_OPTIONS_ATTRIBUTES _name : _type; typedef enum _name _name; enum _name : _type
#endif
```

Though instead redefining an existing typedef that's already typedef'd to int instead of creating a thowaway name that could break source.

millenomi:
Yeah, that's what the clang recommendation was too. I'm having some trouble with the build setup, but once I have a PR that works locally I'll post it up.

gmittert:
@millenomi I'm able to trigger it with just clang 12 on Linux and the invocation:

```
clang -F /path/to/CoreFoundation.framework swift-corelibs-foundation/CoreFoundation/URL.subproj/CFURLSessionInterface.c
```

Adding -Wno-elaborated-enum-base suppresses the error.

PR #2857, opened 2020-08-24
Avoid a crash when the hash value of a shimmed run loop source is negative

millenomi:
`me.hashValue` is `Int`, and `CFHashCode` is `UInt`, so they have the same size.


millenomi:

PR #2860, opened 2020-08-25
Fix authentication through the session task delegate (SR-12676)

millenomi:
- WWW-Authenticate must canonicalize with 'WWW' uppercase.
- The session delegate must be consulted for authentication even for tasks that have completion blocks.

https://bugs.swift.org/browse/SR-12676


millenomi:
@swift-ci please test

PR #2861, opened 2020-08-25
URL.appendingPathComponent("") crashes

millenomi:
Do not attempt to read characters from this string if it's empty.


millenomi:
@swift-ci please test

PR #2865, opened 2020-08-29
[5.3][CMake] fix runpath for ELF platforms

buttaface:
Remove the absolute path to the host toolchain's stdlib from the three Foundation shared libraries.

@millenomi, here you go.


drexin:
@swift-ci test

buttaface:
macOS CI failure must be spurious as this pull does nothing on Darwin platforms.

buttaface:
I think rerunning on the macOS CI should get this passing.

drexin:
No, there seems to be a problem with the build. We are looking into it.

drexin (approved):

buttaface:
Oh, nice, glad to finally see this fixed.

PR #2872, opened 2020-09-07
Add support for WASI in CFInternal.h

MaxDesiatov:
Atomics are not supported on WASI due to SR-12097. `<xlocale.h>` header is also not available as WASI hosts do not ship locale or time zone files and it's a responsibility of applications to bundle that data.


MaxDesiatov:
@swift-ci please test

PR #2873, opened 2020-09-07
Add WASI support in CoreFoundation_Prefix.h

MaxDesiatov:
Dispatch is not available on WASI due to SR-12097. Atomic shims such as `OSAtomicCompareAndSwapPtr` are made available as they would have a no-op implementation for compatibility in a separate PR.

### DIFF
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -20,7 +20,11 @@
 
 #include <CoreFoundation/CFAvailability.h>
 
+#if TARGET_OS_WASI
+#define __HAS_DISPATCH__ 0
+#else
 #define __HAS_DISPATCH__ 1
+#endif
 
 #include <CoreFoundation/CFBase.h>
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci please test

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -20,7 +20,11 @@
 
 #include <CoreFoundation/CFAvailability.h>
 
+#if TARGET_OS_WASI
```
millenomi, line 23: Huh; do things even build correctly without dispatch, or at least shims? I'm pretty sure we depend on `dispatch_*` kind of all over the place, which is why `__HAS_DISPATCH__` was kept fixed at 1.

MaxDesiatov: Yes, at least a certain subset (albeit a useful subset) does. That's why there's more `#if TARGET_OS_WASI` all over the codebase of our fork, more than a 100 files touched, and certain APIs are disabled completely. I'm going to submit those as separate PRs obviously.

millenomi: Is WASI strictly single-threaded?

MaxDesiatov: Yes, multi-threading support and atomics aren't standardised yet, and will be available in some future (unspecified) browser versions. Only Chrome has a working implementation hidden with experimental flags, as as I remember.

PR #2874, opened 2020-09-07
Add failing test for dynamic cast

lxbndr:
I am getting crash when trying to perform dynamic cast like this:
```
let anyArray: Any = [TestClass()]
let maybeObject = anyArray as? NSObject
```

```
Could not cast value of type 'TestFoundation.TestBridging.(unknown context at $7ff798798298).(unknown context at $7ff7987982a0).TestClass' (00007FF7988CC248) to 'Foundation.NSObject' (00007FFD7253A528). 
```

While it is possibly a bug in new dynamic casting in compiler, I think it is good to cover it with test already.

### DIFF
--- a/Tests/Foundation/Tests/TestBridging.swift
+++ b/Tests/Foundation/Tests/TestBridging.swift
@@ -26,6 +26,7 @@ class TestBridging : XCTestCase {
     static var allTests: [(String, (TestBridging) -> () throws -> Void)] {
         return [
             ("testBridgedDescription", testBridgedDescription),
+            ("testDynamicCast", testDynamicCast),
         ]
     }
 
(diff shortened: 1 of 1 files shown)


lxbndr:
cc @compnerd @spevans 
Could you please run a CI for this? I am going to add the test as xcfail, but first it would be good to see is it fails on Linux as well. I am observing this on Windows.

MaxDesiatov:
@swift-ci please test

lxbndr:
So, it fails on Linux too

MaxDesiatov:
Has anything changed substantially since the last push? Would you like me to retrigger CI for it?

lxbndr:
@MaxDesiatov yes, please, that would be nice. Now it is expected-to-fail, so tests should pass. I would look into the reason of this behavior later.

MaxDesiatov:
@swift-ci please test

MaxDesiatov:
@swift-ci please test

compnerd (approved):
Thanks!  If there is an SR, would be good to reference it.

lxbndr:
I was wrong in test expectations. We should expect the value to be _not_ nil. This doesn't the issue itself, as dynamic cast is still crashes.
@compnerd @slavapestov Guys, could you please run CI one more time?

MaxDesiatov:
@swift-ci please test

MaxDesiatov:
@swift-ci please test

spevans:
@swift-ci test linux

lxbndr:
As the fix for this issue is merged, I am going to enable the test

millenomi:
@swift-ci please test

compnerd:
@swift-ci please test

PR #2875, opened 2020-09-08
Fix test fail triggered by multiline environment variables

stevapple:
Fix `TestProcess.test_passthrough_environment` fails when an environment variable has multiple lines.

### DIFF
--- a/Tests/Foundation/Tests/TestProcess.swift
+++ b/Tests/Foundation/Tests/TestProcess.swift
@@ -215,7 +215,12 @@ class TestProcess : XCTestCase {
         do {
             let (output, _) = try runTask([xdgTestHelperURL().path, "--env"], environment: nil)
             let env = try parseEnv(output)
+#if os(Windows)
+            // On Windows, Path is always passed to the sub process
+            XCTAssertGreaterThan(env.count, 1)
+#else
             XCTAssertGreaterThan(env.count, 0)
+#endif
         } catch {
             XCTFail("Test failed: \(error)")
         }
--- a/Tests/Tools/XDGTestHelper/main.swift
+++ b/Tests/Tools/XDGTestHelper/main.swift
@@ -232,7 +232,7 @@ case "--echo-PWD":
     print(ProcessInfo.processInfo.environment["PWD"] ?? "")
 
 case "--env":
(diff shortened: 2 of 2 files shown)


stevapple:
@swift-ci please test

stevapple:
This is going to fix `TestFoundation` failures on [Azure CI](https://dev.azure.com/compnerd/swift-build/_build) for Windows builds.

CC @compnerd

MaxDesiatov:
@swift-ci please test

stevapple:
The Linux test failed, but the failure seems not caused by this change... Could you find the root of it? @MaxDesiatov 😂

MaxDesiatov:
@swift-ci please test

stevapple:
Can anyone merge this? I’d like to see the effect on Azure CI.

@MaxDesiatov

compnerd:
Thanks for looking into this @stevapple.  Can you expand the commit message to explain why multi line values don't work?  https://docs.microsoft.com/en-us/windows/win32/procthread/environment-variables doesn't seem to indicate the restriction.

stevapple:
@compnerd It’s not a problem of Windows actually.

First of all, Azure CI environment sometimes contains multiline environment variables, which is really a rare case. The former handler of `--env` simply `print("\(key)=\(value)")`, and the parser parse every single line of the output as "key=value", so any extra lines in an environment variable will lead to an error, thus failing the test. What I’m going to do is simply removing all the `newlines` characters from `value`, then the problem will go.

compnerd:
I guess the question is why does this not impact other platforms?  I assume that it is an issue only because it is splitting on the newline?

stevapple:
It’s because multiline environment variable is really a rare case. A regular machine won’t have that. Azure CI is triggering it because it stores the latest commit message in an environment variable.

I believe the problem will also occur if we manually set an environment variable with multiple lines on Linux and macOS.

compnerd:
Then, why should this not apply to all targets instead of being windows only?  The problem exists in all of the cases it sounds like,

stevapple:
The patch is applied to all platforms. The `#if os(Windows)` mark here is to make sure the test is meaningful on Windows because whatever subprocess on Windows will have at least one environment variable.

PR #2877, opened 2020-09-10
Adds missing call of endDocument, fixes SR-2301

dhoepfl:
SR-2301 was mostly fixed by pull request #700 (commit 30b17ed) in 2016.
The fix however still does not call the endDocument callback.
This pull request adds the necessary end of document signaling to SAX.

This pull also fixes several minor (rare) bugs like unchecked error conditions, incorrect handling of small chuncks, etc.

Please note the change in CFXMLInterface: I’m not  sure if this is the best place to handle this. The interface is undocumented and returning an error that the end of the document has been reached when you send the last piece of data was unexpected. Even more so since the meaning of the return values is undocumented and the internal (SAX) errors are not mapped to the ErrorCode enum of XMLParser (Leaving this for a different fix).

### DIFF
--- a/Sources/FoundationXML/XMLParser.swift
+++ b/Sources/FoundationXML/XMLParser.swift
@@ -1,6 +1,6 @@
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See http://swift.org/LICENSE.txt for license information
@@ -504,7 +504,7 @@ open class XMLParser : NSObject {
             // If we have not received 4 bytes, save the bomChunk for next pass
             if bomChunk.count < 4 {
                 _bomChunk = bomChunk
-                return false
+                return true
             }
             // Prepare options (substitute entities, recover on errors)
             var options = _kCFXMLInterfaceRecover | _kCFXMLInterfaceNoEnt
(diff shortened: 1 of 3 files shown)


Sources/FoundationXML/XMLParser.swift
```diff
@@ -532,57 +539,54 @@ open class XMLParser : NSObject {
...
         }
+
         return result
     }
     
     // called to start the event-driven parse. Returns YES in the event of a successful parse, and NO in case of error.
     open func parse() -> Bool {
-        return parseFromStream()
+        XMLParser.setCurrentParser(self)
+        defer { XMLParser.setCurrentParser(nil) }
```
millenomi, line 581: If I start another parser from a parser callback, will this be reentrant correctly?

millenomi:
@dhoepfl still interested in shepherding this?

Sources/FoundationXML/XMLParser.swift
dhoepfl: As far as I understand the it: No it will not.

On the other side: That’s no difference from the current code. (old: parse → parseFromStream → setCurrentParser → parseData → callback, new: parse → setCurrentParser [→ parseFromStream] → parseData → callback)

dhoepfl:
> @dhoepfl still interested in shepherding this?

I think the reentrance problem you found is valid and should be tracked as a separate issue, because this might result in a major refactoring of XMLParser (beyond what I’m capable of doing). I will try to find time to file an issue for it tomorrow.

dhoepfl:
Created SR-13546 regarding the reentrance problem.

millenomi:
@swift-ci please rest and merge

dhoepfl:
> @swift-ci please rest and merge

What a great typo. Bots deserve a rest, too.

millenomi:

millenomi:
Both me and the CI bot needed rest, whoops.

PR #2881, opened 2020-09-16
Hash implementation for NSIndexSet

lxbndr:
We noticed that `NSIndexSet` is missing implementation of `hash` property. Not a big deal, as NSObject provides default hash, but there are some cons:
- Equal `NSIndexSet`s would have different hashes based on their object identifier
- `IndexSet` is backed by `NSIndexSet`, deriving same inconvenience
- This behavior is not consistent with Apple Foundation

After some experiments with NSIndexSet from macOS Foundation, these details were observed:
- Calculation complexity is O(1). No matter how many indexes/ranges is stored, `hash` is always fast.
- The only parameters affecting hash are: count of indexes, first and last index value. This explains constant complexity.
- Empty index set has hash equal to 0.

I think the most important detail here is calculation speed, because my first thoughts on implementation were about iterating over all ranges in set. Which is, obviously, slower. But we still could take into account number of ranges to make hash better.

Based on all above, we'd like to suggest hash implementation powered by standard Hasher and based on index count, range count, first and last index.

### DIFF
--- a/Sources/Foundation/NSIndexSet.swift
+++ b/Sources/Foundation/NSIndexSet.swift
@@ -187,6 +187,15 @@ open class NSIndexSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
         return isEqual(to: indexSet._swiftObject)
     }
     
+    open override var hash: Int {
+        var hasher = Hasher()
+        hasher.combine(_count)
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci please test

PR #2882, opened 2020-09-22
Add discardableResult to XMLParser.parse

keith:
On macOS you don't get a warning when discarding the result of this
function. It isn't written in Swift so it doesn't have this attribute,
so I'm assuming it's Objective-C import semantics that imply allowing
discarding it without warnings.

### DIFF
--- a/Sources/FoundationXML/XMLParser.swift
+++ b/Sources/FoundationXML/XMLParser.swift
@@ -576,6 +576,7 @@ open class XMLParser : NSObject {
     }
     
     // called to start the event-driven parse. Returns YES in the event of a successful parse, and NO in case of error.
+    @discardableResult
     open func parse() -> Bool {
         XMLParser.setCurrentParser(self)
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci please test

MaxDesiatov:
Please select the `main` branch as base instead of `master`. This is the new required check for PRs.

keith:
changed, thanks

MaxDesiatov:
@swift-ci please test

MaxDesiatov:
The CI requirement for `main` is stuck for some reason 😢

keith:
should I force push something?

MaxDesiatov:
IDK, I'd let the reviewers decide, they can probably force the check if anything.

keith:
There was a discussion about this in slack, the tl;dr is that Apple folks don't want this unless there's a plan to annotate all APIs that suffer from this issue.

PR #2883, opened 2020-09-26
NSSet: add missing convenience initializer

egorzhdan:
This initializer exists in the macOS version of Foundation, but isn't present on Linux/Windows.
While it doesn't add any new functionality, it simplifies porting existing code to those platforms.

### DIFF
--- a/Sources/Foundation/NSSet.swift
+++ b/Sources/Foundation/NSSet.swift
@@ -86,6 +86,10 @@ open class NSSet : NSObject, NSCopying, NSMutableCopying, NSSecureCoding, NSCodi
         self.init(array: [object])
     }
 
+    public convenience init(objects elements: Any...) {
+        self.init(array: elements)
+    }
+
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci please test

compnerd:
I see https://developer.apple.com/documentation/foundation/nsset/1574825-setwithobjects?language=objc but it doesn't seem to have the corresponding interface in the Swift view.  The change itself LGTM, assuming @millenomi doesn't know of a reason that the interface is missing.  However, if there is a reason, it certainly should be indicated with a comment.

egorzhdan:
It exists for Swift as well: https://developer.apple.com/documentation/foundation/nsset/1407975-init

I assume Swift isn't shown in the "Language" popup on https://developer.apple.com/documentation/foundation/nsset/1574825-setwithobjects?language=objc because this is a static method in ObjC but an initializer in Swift.

compnerd (approved):

compnerd:
Lets wait for tomorrow just to allow @millenomi a chance to look at it as well, but with that last bit, I think that this is perfectly fine.

millenomi:

PR #2885, opened 2020-09-26
Foundation: explicitly convert value to `Date`

compnerd:
There is no implicit conversion from `Int` to `Date`.  The implicit cast
fails, resulting in `nil` being returned for the `creationDate`
property.  Explicitly unwrap the result of the `.creationDate` attribute
and use that to construct the `Date` to return.

Partially addresses: SR-13617

### DIFF
--- a/Sources/Foundation/URL.swift
+++ b/Sources/Foundation/URL.swift
@@ -149,7 +149,10 @@ public struct URLResourceValues {
     
     /// The date the resource was created.
     public var creationDate: Date? {
-        get { return _get(.creationDateKey) }
+        get {
+          guard let timestamp: Int = _get(.creationDateKey) else { return nil }
+          return Date(timeIntervalSince1970: TimeInterval(timestamp))
+        }
         set { _set(.creationDateKey, newValue: newValue) }
     }
     


compnerd:
@swift-ci please test

compnerd:
CC: @spevans @millenomi

Sources/Foundation/URL.swift
```diff
@@ -149,7 +149,10 @@ public struct URLResourceValues {
     
     /// The date the resource was created.
     public var creationDate: Date? {
-        get { return _get(.creationDateKey) }
+        get {
+          guard let timestamp: Int = _get(.creationDateKey) else { return nil }
+          return Date(timeIntervalSince1970: TimeInterval(timestamp))
+        }
```
xwu, line 155: Wait, `set` doesn't have the corresponding issue?

compnerd: Thats actually why Im not 100% certain why this is needed.  There are other `Date` accessors which should also see similar things?  The test suite seems to be passing though, so Im not sure what is special here.

compnerd:
Okay, with the help from @millenomi this is no longer necessary.

PR #2887, opened 2020-09-28
Foundation: map `ECANCELED` properly for Windows

compnerd:
This is generally not a very helpful constant for Windows, however, it
can be passed from Dispatch in some cases.  Lets just provide the
mapping for the Windows version.


compnerd:
@swift-ci please test

PR #2891, opened 2020-10-03
FoundationNetworking: make `CFURLSessionInterface` `@_implementationO…

compnerd:
…nly`

Remove the public exposure to `CFURLSessionInterface` which was never
meant to be public.  This requires removal of some of the `Equatable`
conformances as scoped conformances are not yet a part of the language.
This does the minimal change for to get `FoundationNetworking` to build.


compnerd:
@swift-ci please test

PR #2892, opened 2020-10-04
Foundation: make `-enable-library-evolution` build

compnerd:
When building Foundation with library evolution (resilience), we fail to
build due to `RangeReference` being internal and marked as
`@usableFromInline` but used in an `@inlinable` `public` context,
exposing the type.  Additionally, the constructor was internal and
`@inlinable` which requires the constructor to be delegating.

Additionally, because `Data.init(bytes:length:copy:deallocator)` is used
in a public context, it leaks and results in an undefined reference to
the allocating initializer.

This allows building Foundation with library evolution, though exposes
some of the internal interfaces.


parkera:

PR #2893, opened 2020-10-04
Foundation: extract `cfSource` from `RunLoop._Source`

compnerd:
Making `cfSource` a member within the structure prevents allowing
overriding of the `open` class as the type it returns cannot be
deserialised.  Because the member is not a stored property, extract it
to an extension which removes it from the vtable.  This allows the use
of the private type in an external context.


millenomi:
Please also file separately to figure out why Windows could not handle this computed variable in this context?

millenomi:

PR #2625, opened 2020-01-22
SR-11569: DateComponents.isValidDate() ignores timeZone

spevans:
- Call components(_:from:) using the correct calendar object with
  the correct time zone.

PR #2894, opened 2020-10-07
Use receiver's timezone in isValidDate() as per documentation

ps2:
The [documentation of DateComponents.isValidDate(in:)](https://developer.apple.com/documentation/foundation/nsdatecomponents/1412707-isvaliddate) indicates that the receiver's timezone is used, if it is set. The code also appears to be attempting to implement this, by setting the `timeZone` property on a local `cal` var, but this var wasn't being used at all later in the function. 

The result is that seemingly valid DateComponents can return false for isValidDate(), when the receiver's timezone differs from the passed in calendar's timezone, which is often system timezone.


ps2:
Ah, this *was* fixed in https://github.com/apple/swift-corelibs-foundation/pull/2625

PR #2895, opened 2020-10-08
[cxx-interop] Fix two headers to be valid in C++ mode.

zoecarver:
This is exported as a C header but in C++ mode `ucal.h` will expose templates that aren't valid. This will hopefully fix the Linux build for https://github.com/apple/swift/pull/32721.

### DIFF
--- a/CoreFoundation/Locale.subproj/CFCalendar_Internal.h
+++ b/CoreFoundation/Locale.subproj/CFCalendar_Internal.h
@@ -19,7 +19,7 @@
 #include "CFDateComponents.h"
 #include "CFDateInterval.h"
 
-#if __has_include(<unicode/ucal.h>)
+#if __has_include(<unicode/ucal.h>) && !defined(__cplusplus)
 #include <unicode/ucal.h>
 #else
 typedef void *UCalendar;
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -672,8 +672,12 @@ CF_EXPORT int _CFPosixSpawnFileActionsDestroy(_CFPosixSpawnFileActionsRef file_a
 CF_EXPORT void _CFPosixSpawnFileActionsDealloc(_CFPosixSpawnFileActionsRef file_actions);
 CF_EXPORT int _CFPosixSpawnFileActionsAddDup2(_CFPosixSpawnFileActionsRef file_actions, int filedes, int newfiledes);
 CF_EXPORT int _CFPosixSpawnFileActionsAddClose(_CFPosixSpawnFileActionsRef file_actions, int filedes);
+#ifdef __cplusplus
+CF_EXPORT int _CFPosixSpawn(pid_t *_CF_RESTRICT pid, const char *_CF_RESTRICT path, _CFPosixSpawnFileActionsRef file_actions, _CFPosixSpawnAttrRef _Nullable _CF_RESTRICT attrp, char *const argv[], char *const envp[]);
+#else
 CF_EXPORT int _CFPosixSpawn(pid_t *_CF_RESTRICT pid, const char *_CF_RESTRICT path, _CFPosixSpawnFileActionsRef file_actions, _CFPosixSpawnAttrRef _Nullable _CF_RESTRICT attrp, char *_Nullable const argv[_Nullable _CF_RESTRICT], char *_Nullable const envp[_Nullable _CF_RESTRICT]);
-#endif
+#endif // __cplusplus
+#endif // !TARGET_OS_WIN32
 
 _CF_EXPORT_SCOPE_END
 


spevans:
@swift-ci test

zoecarver:
@swift-ci please test.

zoecarver:
I don't have commit access. Could someone review/merge this for me?

MaxDesiatov (approved):
This seems legit to me, but as there's a way to test the toolchain PR with this without merging, I'll defer to the rest of the requested reviewers to merge it, unless it's urgent.

zoecarver:
> This seems legit to me, but as there's a way to test the toolchain PR with this without merging, I'll defer to the rest of the requested reviewers to merge it, unless it's urgent.

Not urgent. I'm just eager to get that other patch in 😁 I had forgotten that toolchain PR testing works, thanks for running that.

zoecarver:
Friendly ping.

MaxDesiatov:
@compnerd @millenomi would you have a moment to take a look, or would you mind if I merge this otherwise?

compnerd:
Hmm, where is `CFCalendar_Internal.h` used in a C++ context?  I also have the same question for the `CFForSwiftFoundationOnly.h` header.  I think we should make that more explicit in the commit message at the very least.

zoecarver:
> Hmm, where is CFCalendar_Internal.h used in a C++ context? I also have the same question for the CFForSwiftFoundationOnly.h header.

When C++ interop is enabled, any time these are used it's in a C++ context. The same is true for Swift shims (we had to make similar fixes there a while back).

zoecarver:
In other words, it's used in C++ mode any time a Swift program imports `CoreFoundation` and has the flag `-enable-cxx-interop`.

I'll update the commit messages to clarify that this is needed to import these into programs that use the `-enable-cxx-interop` flag.

MaxDesiatov:
@swift-ci please test

zoecarver:
I've updated the commit message. @compnerd any other comments?

MaxDesiatov:
@swift-ci please test

zoecarver:
@swift-ci please test Linux platform.

zoecarver:
@swift-ci please test Linux platform.

zoecarver:
@swift-ci please test Linux platform.

zoecarver:
@swift-ci please test Linux platform.

zoecarver:
Friendly ping.

zoecarver:
@swift-ci please smoke test Windows platform.

zoecarver:

zoecarver:

zoecarver:

zoecarver:
@millenomi friendly ping. Any comments?

millenomi:
I'm a little thick but I'm not sure I get the change for _Internal, which should not be imported outside of the CF build itself, which is in C mode and not C++. Okay on the ForSwift… because that ends up in the CF module, tho.

CoreFoundation/Locale.subproj/CFCalendar_Internal.h
```diff
@@ -19,7 +19,7 @@
 #include "CFDateComponents.h"
 #include "CFDateInterval.h"
 
-#if __has_include(<unicode/ucal.h>)
+#if __has_include(<unicode/ucal.h>) && !defined(__cplusplus)
```
millenomi, line 22: Per comment: please explain here.

millenomi (changes requested):

CoreFoundation/Locale.subproj/CFCalendar_Internal.h
zoecarver: With C++ Interoperability enabled, all headers are imported as C++ headers. So they will all be imported in C++ mode. Even if they're not publically exposed. In other words, all headers in the CF module will be imported as C++ headers in a program that enables C++ interoperability. 

You're right, I think ForSwift may be the only place this is causing issues. To be honest, I know next to nothing about core foundation so, maybe this isn't the correct fix. Feel free to suggest an alternative. 

Anyway, does that answer your question? Would you also like me to put a comment in "CFCalendar_Internal.h" explaining this?

zoecarver: @millenomi friendly ping. I'd really like to get this moving.

millenomi: My bad, I missed your ping back.

millenomi (approved):

millenomi:

millenomi:
In the interest of getting this going, I'm okay with the safety change on the internal header.

CoreFoundation/Locale.subproj/CFCalendar_Internal.h
zoecarver: No worries. I missed your comment the first time ;) Thanks!

zoecarver:
> In the interest of getting this going, I'm okay with the safety change on the internal header.

Great. If you'd like to keep discussing, I'm happy to. (And if we arrive at a better solution, I'm happy to implement it.)

PR #2187, opened 2019-04-27
NSError: added using userInfoProviders

Alexander-Ignition:
In Darwin Foundation, the value for NSError properties is taken not only from userInfo but also from userInfoProvider.

PR #2898, opened 2020-10-10
NSError: added using userInfoProviders

Alexander-Ignition:
In Darwin Foundation, the value for NSError properties is taken not only from userInfo but also from userInfoProvider.

Copy of #2187


MaxDesiatov:
@swift-ci please test

MaxDesiatov:
@swift-ci please test

PR #2900, opened 2020-10-10
Fix memory leak in CFBundle_Executable.c

valeriyvan:


MaxDesiatov:
@swift-ci please test

compnerd:
Please add a better commit message than 'Fix memory leak'.  How was this detected?  What scenarios did it occur under?  Seems like this is the non-Windows Swift path, which means that Darwin would not be impacted?

compnerd (changes requested):
The commit message should be reworked.

valeriyvan:
> Please add a better commit message than 'Fix memory leak'. How was this detected? What scenarios did it occur under? Seems like this is the non-Windows Swift path, which means that Darwin would not be impacted?

Found with Xcode static analyser.
On line 85 call `CFURLCreateWithFileSystemPathRelativeToBase` returns object with +1 retain count according to [Create rule](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-103029).

MaxDesiatov:
> Found with Xcode static analyser.
> On line 85 call `CFURLCreateWithFileSystemPathRelativeToBase` returns object with +1 retain count according to [Create rule](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-103029).

This should be added to the commit message, not a PR comment which will be lost after merge.

Personally, I would prefer it as a comment at the appropriate location in the code.

compnerd:
@valeriyvan yes, I was able to infer that from reading the code (the path that leaked the CFURL) - but that information belongs in the commit message so that someone going through the history doesn't need to reconstruct every bit of detail from the diff.

valeriyvan:
> Personally, I would prefer it as a comment at the appropriate location in the code.

File `CFBundle_Executable.c` has 46 calls of `CFRelease`. No-one is commented. Why exactly this one should be commented? This is trivial bug. Someone just forgotten that retained object should be released. I don't see any reason why this call of CFRelease deserves to be commented.

valeriyvan:
> The commit message should be reworked.

Updated commit message

compnerd (approved):
Might be nice to actually follow git suggestions for formatting commit messages.

valeriyvan:
> Might be nice to actually follow git suggestions for formatting commit messages.

Fixed commit message.

MaxDesiatov:
@swift-ci please test

PR #2901, opened 2020-10-10
Fix crashes in NSMutableAttributedString: modify in an enumeration

andriydruk:
The current implementation of NSMutableAttributedString crashes if an attributed string change length while an enumeration
We expect similar behavior to Darwin Foundation's NSMutableAttributedString: enumeration should be continued

Also implemented NSAttributedString.description (similar to Darwin version)


Sources/Foundation/NSAttributedString.swift
```diff
@@ -288,11 +316,22 @@ private extension NSAttributedString {
             currentIndex = startIndex
         }
         
-        mutating func advance(step: Int = 1) {
+        mutating func advance(range: NSRange, oldLenght: Int, newLenght: Int) {
```
compnerd, line 319: nit: `Lenght` -> `Length`

MaxDesiatov:
@swift-ci please test

compnerd (approved):
LGTM, though @millenomi should probably also take a look.

PR #2903, opened 2020-10-17
Remove an invalid numeric JSON test

tbkka:
The value `2.7976931348623158e+308` is a perfectly valid JSON number
and should be accepted.  This test is only passing today because the Swift
standard library's `Double(_ String:)` initializer currently has a bug that
rejects numbers larger than the maximum Double value.

But we're getting ready to fix that bug in the Swift standard library,
(apple/swift#34339) which will break this test when it lands.

(After that PR lands, we should update this again to verify that this value
parses to `Double.infinity`.)


stephentyrone (approved):

spevans:
@swift-ci test

spevans (approved):

spevans:
@swift-ci test and merge

spevans:
@swift-ci test

PR #2904, opened 2020-10-17
[SR-13300] FileManager _updateTimes (microseconds)

benrimmington:
The `_updateTimes` function is multiplying by `1.0e9` as if for nanoseconds, but it should be using `1.0e6` for microseconds.

The `timeval.tv_usec` value will be too large, hence the `EINVAL` errno on Raspberry Pi OS (SR-13300).

### DIFF
--- a/Tests/Foundation/Tests/TestFileManager.swift
+++ b/Tests/Foundation/Tests/TestFileManager.swift
@@ -408,15 +408,11 @@ class TestFileManager : XCTestCase {
         try? fm.removeItem(atPath: path)
         XCTAssertTrue(fm.createFile(atPath: path, contents: Data(), attributes: nil))
 
-#if !os(Windows)
         let modificationDate = NSDate(timeIntervalSince1970: 1234567890.5) // 2009-02-13T23:31:30.500Z
-#endif
-        
+
         do {
             try fm.setAttributes([.posixPermissions : NSNumber(value: Int16(0o0600))], ofItemAtPath: path)
-#if !os(Windows)
             try fm.setAttributes([.modificationDate: modificationDate], ofItemAtPath: path)
-#endif
         }
         catch { XCTFail("\(error)") }
         
@@ -427,8 +423,8 @@ class TestFileManager : XCTestCase {
             XCTAssert((attributes[.posixPermissions] as? NSNumber)?.int16Value == 0o0700)
 #else
             XCTAssert((attributes[.posixPermissions] as? NSNumber)?.int16Value == 0o0600)
-            XCTAssertEqual((attributes[.modificationDate] as? NSDate)?.timeIntervalSince1970 ?? .nan, modificationDate.timeIntervalSince1970, accuracy: 1.0)
 #endif
+            XCTAssertEqual((attributes[.modificationDate] as? NSDate)?.timeIntervalSince1970 ?? .nan, modificationDate.timeIntervalSince1970, accuracy: 1.0)
         }
         catch { XCTFail("\(error)") }
 
--- a/Tests/Foundation/Tests/TestFileManager.swift
+++ b/Tests/Foundation/Tests/TestFileManager.swift
@@ -1,6 +1,6 @@
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See http://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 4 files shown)


MaxDesiatov (commented):
Please add tests to cover this behaviour.

benrimmington:
@MaxDesiatov I've added tests for macOS only ([1][], [2][]).

[1]: <https://github.com/apple/swift-corelibs-foundation/blob/9ebb4442a45223b252ab714a2c101748a969e2da/Sources/Foundation/FileManager.swift#L406>
[2]: <https://github.com/apple/swift-corelibs-foundation/blob/9ebb4442a45223b252ab714a2c101748a969e2da/Tests/Foundation/Tests/TestURL.swift#L728>

MaxDesiatov:
What prevents adding tests on Linux, as the bug reported on Jira is reproducible on Linux?

benrimmington:
> What prevents adding tests on Linux, as the bug reported on Jira is reproducible on Linux?

Because the test may fail on Linux, depending on the OS and filesystem:

https://github.com/apple/swift-corelibs-foundation/blob/9ebb4442a45223b252ab714a2c101748a969e2da/Tests/Foundation/Tests/TestURL.swift#L728

benrimmington:
Without this bug fix, time interval `1234567890.5` results in `timeval.tv_usec = 0.5e9`. This value is:

* rejected on Raspberry Pi OS (SR-13300);
* accepted on macOS, but the modification date is 500 seconds later than expected.

MaxDesiatov:
Sorry, I'm not sure I completely follow. Is this not supposed to fix the bug on Raspberry too? Or will that part come as a separate PR?

benrimmington:
This pull request should fix the reported bug on Raspberry Pi OS.

It will also fix an unreported bug on other POSIX platforms (e.g. where 0.5 seconds unexpectedly becomes 500 seconds).

I've updated the test for Linux, to assert that the modification date is almost equal, within an accuracy of 1 second.

Tests/Foundation/Tests/TestFileManager.swift
```diff
@@ -410,6 +410,9 @@ class TestFileManager : XCTestCase {
         
         do {
             try fm.setAttributes([.posixPermissions : NSNumber(value: Int16(0o0600))], ofItemAtPath: path)
+#if !os(Windows)
+            try fm.setAttributes([.modificationDate: NSDate(timeIntervalSince1970: 1234567890.5)], ofItemAtPath: path)
```
MaxDesiatov, line 414: Nitpick: could you move `1234567890.5` value to a `let` binding to reuse it in `XCTAssertEqual` below?

MaxDesiatov:
@swift-ci please test

benrimmington:
@MaxDesiatov Thanks for your help.

MaxDesiatov (approved):
Awesome, thanks!

compnerd (changes requested):
Why are the tests marked as not windows?  The file modification time information is available on windows as well IIRC.

Can you create an extension to convert the time rather than having the two sites which use the wrong constant?  That would ensure that they don't ever drift and cases like this are localized.

benrimmington:
> Why are the tests marked as not windows? The file modification time information is available on windows as well IIRC.

An [`_updateTimes`][] function is available for Windows, but I think the tests would currently fail:

https://github.com/apple/swift-corelibs-foundation/blob/cfef71fe9e0df0a5d6430f6fb202896080a17956/Sources/Foundation/FileManager.swift#L403-L408

[`_updateTimes`]: <https://github.com/apple/swift-corelibs-foundation/blob/cfef71fe9e0df0a5d6430f6fb202896080a17956/Sources/Foundation/FileManager%2BWin32.swift#L885-L909>

benrimmington:
@compnerd I've applied your suggestions as separate commits.

compnerd:
@benrimmington thanks!  I don't mind too much about the separation of the commits, and since the intermediate states in this case should be okay, I dont think that they really matter.  The final result of this looks much better, thank you for the additional changes!

compnerd (approved):

compnerd:
@swift-ci please test

compnerd:
Should probably run this on Windows as well to make sure that it doesn't cause a regression, I can kick off a test run once the CI here passes.

benrimmington:
There was a failure in 'lldb-shell :: SwiftREPL/RedirectInput.test' on [Linux](https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/4749/).

I can see the files in my working copy, but I can't find them on [GitHub](https://github.com/apple/llvm-project/tree/apple/main/lldb/test/Shell).

* "llvm-project/lldb/test/Shell/SwiftREPL/RedirectInput.test" contains:

  ```swift
  Foo.foo()
  // CHECK: ${{R0}}: String = "A"
  ```

* "llvm-project/lldb/test/Shell/SwiftREPL/Inputs/A.swift" contains:

  ```swift
  public class Foo {
    public static func foo() -> String  {
      print("A")
      return "A"
    }
  }
  ```

* The expected REPL output is intermixed with the `print("A")` output:

  ```
  $R0: String =A
   "A"
  ```

benrimmington:
I've reported the test failure as SR-13754.

millenomi:
@swift-ci please test Linux

benrimmington:
@compnerd All checks have passed. Please test on Windows (or else commit 007b850342af9c3b64074edcd29cd8040a10afac can be reverted later).

compnerd:
Tested at https://dev.azure.com/compnerd/swift-build/_build/results?buildId=40716&view=results - windows seems happy too

benrimmington:
@millenomi Do you want me to make any changes?

PR #2884, opened 2020-09-26
Foundation: populate the `.creationDate` attribute on Windows

compnerd:
Ensure that we pass along the creation date from `stat` when querying
the attributes of an item.  This was being dropped resulting in the
creation date for files being unavailable.

Partially addresses SR-13617

PR #2905, opened 2020-10-17
[FileManager] _attributesOfItem: use creationDate from the stat extension in NSURL

buttaface:
rather than [repeating that logic](https://github.com/apple/swift-corelibs-foundation/blob/0f06824e7f81393614201fc67d1d740a8f913df7/Sources/Foundation/NSURL.swift#L2109)

The last case in stat.creationDate also works for Android.

Note that it uses `st_birthtimespec` instead on Darwin, let me know if that makes a difference. @compnerd, this pull redoes #2884 because it didn't work on Android and to reuse the existing logic, let me know what you think.

### DIFF
--- a/Sources/Foundation/FileManager.swift
+++ b/Sources/Foundation/FileManager.swift
@@ -544,14 +544,7 @@ open class FileManager : NSObject {
         result[.creationDate] = creationDate
 #else
         let s = try _lstatFile(atPath: path)
-        // Darwin provides a `st_ctimespec` rather than the traditional Unix
-        // `st_ctime` field.  Since `st_ctime` is more traditional, special case
-        // Darwin platforms and convert the timespec to the absolute time.
-#if os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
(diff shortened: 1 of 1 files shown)


spevans:
looking at the manpage for `stat(2)` on macOS, `st_birthtimespec` does look like the better choice:
```
 struct timespec st_ctimespec;     /* time of last status change */
 struct timespec st_birthtimespec; /* time of file creation(birth) */
```

spevans:
@swift-ci test

compnerd:
Could you please ensure that this still works on Windows?  If it does, LGTM.

buttaface:
I don't use Windows so I can't ensure it, but the Windows case being used now exactly matches the else case you added, so it should be fine.

compnerd:
I was more worried about the property not existing and this failing to build.  Seems that the property is present.  LGTM.

compnerd (approved):

compnerd:
Thanks for the cleanup @buttaface!

PR #2907, opened 2020-10-19
[Windows] Make mach_absolute_time implementation consistent with libdispatch

lxbndr:
libdispatch uses `QueryUnbiasedInterruptTimePrecise` as base time source for timers. This changes CoreFoundation to use same source to schedule timers properly.

Current implementation is observed to work in most cases, because `QueryPerformanceCounter` could return same value as `QueryUnbiasedInterruptTimePrecise`. E.g. it provides reliable values on Windows VM hosted on my MacBook, as well as on cloud CI agents (GitHub Actions/Azure Pipelines). Perhaps virtualization affects the result. But on my Windows desktop these two sources return incomparably different values, making CFRunLoop unable to calculate correct timeout deadline.

The issue comes out as hang of `TestNotificationQueue.test_postNowForDefaultRunLoopMode` test. I also had similar problem with s-p-m, and this change fixes it as well.

### DIFF
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -219,10 +219,9 @@ CF_EXPORT void OSMemoryBarrier();
 
 CF_INLINE uint64_t mach_absolute_time() {
 #if TARGET_OS_WIN32
-    LARGE_INTEGER count;
-    QueryPerformanceCounter(&count);
-    // mach_absolute_time is unsigned, but this function returns a signed value.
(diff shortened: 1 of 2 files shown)


lxbndr:
CC @compnerd

compnerd (approved):

compnerd:
@swift-ci please test

spevans:
Is this worth cherry-picking back to 5.3?

lxbndr:
It's pretty fundamental issue which affects a lot of libraries for sure. I'm for cherry-picking.
@compnerd WDYT?

PR #2910, opened 2020-10-24
[Windows] Add trailing slash to NSTemporaryDirectory result

lxbndr:
NSTemporaryDirectory on Darwin returns path with trailing slash, and almost every test in TestFileManager counts on it.

Without slash tests are generating temporary paths like `C:\Users\user\AppData\Temptest_isReadableFile5004BFCF-4356-4000-8000-000000000000`, which is, obviously, out of Temp directory.

Non-Darwin and non-Windows platforms have already implemented similar workaround, so I just adapted existing solution for Windows.


lxbndr:
CC @compnerd

compnerd (approved):

compnerd:
@swift-ci please test

PR #2909, opened 2020-10-24
[Windows] Restore original working directory in TestFileManager.test_windowsPaths

lxbndr:
Staying inside the directory breaks cleanup on tear down. This is part 1 of 2 of FileManager tests cleanup improvements. Second part is not specific to Windows, so goes separately.

PR #2911, opened 2020-10-24
Improve cleanup in FileManager tests

lxbndr:
Most of tests in `TestFileManager` do cleanup of created content, but some are missing it. After several days of re-running Foundation test suite Temp folder is full of abandoned files. While it is not a big deal (this is temp folder after all), I think it is good to be tidy and do some trivial cleanup.

This change is complement to #2909


MaxDesiatov:
@swift-ci please test

compnerd:
Does it make sense to put these into defer blocks instead at the beginning of the function?

lxbndr:
> Does it make sense to put these into defer blocks instead at the beginning of the function?
I believe it would be better with `defer`. I didn't use it just because there are many similar tests with same approach. But I guess there is no reason to keep consistency with that code. Will do.

compnerd (approved):

PR #2912, opened 2020-10-26
Improve Error to NSError conversion

kgominyuk:
Error->NSError conversion in Darwin Foundation fills `domain` and `code` even if provided Error does not conform to CustomNSError. This patch brings same behavior to swift-corelibs-foundation.

### DIFF
--- a/Tests/Foundation/Tests/TestNSError.swift
+++ b/Tests/Foundation/Tests/TestNSError.swift
@@ -22,7 +22,12 @@ class TestNSError : XCTestCase {
             ("test_CustomNSError_errorCode", test_CustomNSError_errorCode),
             ("test_CustomNSError_errorCodeRawInt", test_CustomNSError_errorCodeRawInt),
             ("test_CustomNSError_errorCodeRawUInt", test_CustomNSError_errorCodeRawUInt),
-            ("test_errorConvenience", test_errorConvenience)
+            ("test_errorConvenience", test_errorConvenience),
+            ("test_ConvertErrorToNSError_domain", test_ConvertErrorToNSError_domain),
+            ("test_ConvertErrorToNSError_errorCode", test_ConvertErrorToNSError_errorCode),
+            ("test_ConvertErrorToNSError_errorCodeRawInt", test_ConvertErrorToNSError_errorCodeRawInt),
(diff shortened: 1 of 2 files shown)


Tests/Foundation/Tests/TestNSError.swift
```diff
@@ -106,6 +111,59 @@ class TestNSError : XCTestCase {
...
+            case three(String)
+            case four
+            case five(String)
+        }
+        
+        XCTAssertEqual((SwiftError.one as NSError).code, 2)
+        XCTAssertEqual((SwiftError.two as NSError).code, 3)
+        XCTAssertEqual((SwiftError.three("three") as NSError).code, 0)
+        XCTAssertEqual((SwiftError.four as NSError).code, 4)
+        XCTAssertEqual((SwiftError.five("five") as NSError).code, 1)
```
MaxDesiatov, line 164: It's not quite clear why these values are what they are, especially for those with associated values. Could you add a comment to the test that clarifies this please?

MaxDesiatov (approved):
Thanks!

MaxDesiatov:
@swift-ci please test

Tests/Foundation/Tests/TestNSError.swift
kgominyuk: The test was initially based on actual observed behavior on Darwin platform. In search of details I've found that default error code for enum case is based on enum tag. Tag generation logic is located in EnumImplStrategy::getTagIndex ( https://github.com/apple/swift/blob/d6227dcca01b74b64f078949e69e91324e39f815/lib/IRGen/GenEnum.cpp#L239-L255 ). It starts from cases with payload (.three and .five), and then proceeds to cases without payload (.one , .two , .four ). This is why we are observing such error codes: .three = 0, .five = 1, .one = 2, .two = 3, .four = 4.

MaxDesiatov: I didn't know that, very interesting behavior. Glad that it could become consistent across platforms.

kgominyuk:
@MaxDesiatov Could you please retrigger CI? I squashed commits into one.

MaxDesiatov:
@swift-ci please test

millenomi:
The only reason we didn't go deep on the first attempt was that we didn't have a lot of infrastructure set up yet and much was deferred to when we'd have good stories for conversion, for localization, etc., which now we do. More compatibility is always a good thing. LGTM

PR #2913, opened 2020-10-27
Foundation: drop `acceptConnectionInBackgroundAndNotify` on Windows

compnerd:
`acceptConnectionInBackgroundAndNotify(forModes:)` is already
unavailable on Windows.  This makes the helper functional also
unavailable on Windows as `SOCKET` is not a `FileHandle` on Windows.


compnerd:
@swift-ci please test

PR #2914, opened 2020-10-27
Foundation: expose `_handle` from `FileHandle` on Windows

compnerd:
Add a SPI for Windows to get the raw `HANDLE` to enable use of `Pipe`
where a raw file descriptor may be needed.

### DIFF
--- a/Sources/Foundation/FileHandle.swift
+++ b/Sources/Foundation/FileHandle.swift
@@ -49,11 +49,7 @@ extension NSError {
  
 open class FileHandle : NSObject {
 #if os(Windows)
-    private var _handle: HANDLE
-
-    internal var handle: HANDLE {
-      return _handle
-    }
+    public private(set) var _handle: HANDLE
 
     @available(Windows, unavailable, message: "Cannot perform non-owning handle to fd conversion")
     open var fileDescriptor: Int32 {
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

Sources/Foundation/FileHandle.swift
```diff
@@ -49,10 +49,10 @@ extension NSError {
  
 open class FileHandle : NSObject {
 #if os(Windows)
-    private var _handle: HANDLE
+    internal private(set) var handle: HANDLE
```
spevans, line 53: Cant this be done as `public private(set) var _handle: HANDLE` and then the property getter isnt needed or am I missing something?

compnerd: It can, but would be a much larger change as there are other files that use `handle`.  I don't remember why there was a `_handle` and `handle` accessor.  I just made `handle` the internal interface since its used outside of this file and `_handle` the accessor since that needs to be underscored for the public exposure.

spevans: If there is an alias for the same variable would it be worth the effort to fix this to avoid future confusion?

compnerd: I can live with that, updated the diffs.

compnerd:
@swift-ci please test

PR #2916, opened 2020-11-04
[5.3] Make mach_absolute_time implementation consistent with libdispatch

lxbndr:
Copied from #2907. The diff is exactly same. Considered safe and worth to port to 5.3.


lxbndr:
CC @spevans @compnerd

MaxDesiatov:
@swift-ci please test

compnerd (approved):
This is a good idea to backport - this actually fixes a bug in the runloop scheduling.

PR #2917, opened 2020-11-04
[5.3] Windows, remove obsolete TLS finalization

lxbndr:
Copied from #2902. The diff is exactly same. Considered safe and worth to port to 5.3.


lxbndr:
CC @compnerd @spevans

MaxDesiatov:
@swift-ci please test

compnerd (approved):
This really should have no impact, but makes the code clearer.

lxbndr:
I believe it fixes an elusive crash. We had a stable reproduction rate "in vivo", as our code actively uses dispatch queues. But I could not reproduce it sample app, unfortunately.

PR #2921, opened 2020-11-10
Explicitly import Glibc in files that rely on it

scentini:
These files rely on Glibc declarations in their public interface, even though they don’t import Glibc. Currently, these declarations are discovered through CoreFoundation, but due to implementation details, Swift canonicalizes the declarations and at the end uses declarations from Glibc (which is imported by other files in the module). This confuses `@_implementationOnly` checker to recognize these declarations as visible even though they shouldn't be.

This is fragile as Foundation now relies on implementation details of Glibc, which in fact we are trying to change in https://github.com/apple/swift/pull/32404.

This PR fixes the issue by explicitly importing Glibc in affected files.

### DIFF
--- a/Sources/Foundation/Data.swift
+++ b/Sources/Foundation/Data.swift
@@ -21,6 +21,10 @@ internal func malloc_good_size(_ size: Int) -> Int {
 
 @_implementationOnly import CoreFoundation
 
+#if canImport(Glibc)
+import Glibc
+#endif
+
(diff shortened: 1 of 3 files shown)


scentini:
cc/ @CodaFi @gribozavr

spevans:
@swift-ci test

hlopko:
CC @hlopko

hlopko:
Friendly ping :)

compnerd (approved):
I'm still not convinced by `canImport`, but that is the convention in the repository ... and, when in Rome ..

PR #2925, opened 2020-11-20
SR-7236: NSDecimalNumber `.intValue` reporting 0 on numbers with many decimals

spevans:
- The early check to return 0 if there were too many trailing zeros
  before the decimal point (> UInt64.max) or too many leading zeros
  after the decimal point didnt allow for a large enough negative
  exponent.

- Also resolves SR-2980


spevans:
@swift-ci test

spevans:
@swift-ci test and merge

spevans:
@swift-ci test and merge

PR #2930, opened 2020-11-27
Fix OperationQueue scheduling logic to prevent disruption of operation lists

lxbndr:
OperationQueue would crash or hang if we force "non-natural" execution order. E.g. when first operation depends on second.

The reason is how we work with internal operation lists in `OperationQueue._schedule()`. I believe the original intention was to adjust `__nextPriorityOperation`, but instead we are modifying `__nextOperation`. This has no sense and leads to crosslinking of the main operation list and corresponding priority operation list. Subsequent scheduling would try to read already finished operation and hang. Also, at this point queue doesn't hold operation object anymore, so trying to read dead object is pretty realistic scenario.

Suggested test case checks scenario described above, and ensures correct operation order as a bonus.


lxbndr:
cc @compnerd @millenomi @spevans

spevans:
@swift-ci test

lxbndr:
Build errors look unrelated. @spevans could you please trigger new CI run?

spevans:
@swift-ci test linux

compnerd (approved):

PR #2928, opened 2020-11-24
SR-13861: Process should spawn a child process it its own process group.

spevans:
- Unify use of posix_spawnattr_t between Darwin and other platforms,
  and set the POSIX_SPAWN_SETPGROUP attribute.

- This matches Darwin's behaviour.

PR #2934, opened 2020-12-04
Revert "SR-13861: Process should spawn a child process it its own process group."

compnerd:
Reverts apple/swift-corelibs-foundation#2928

This breaks the Windows build.


compnerd:
@swift-ci please test

spevans:
@compnerd I think I forgot to guard the test from running on Windows. I assume that's what broke?

spevans:
#2935 Should fix it for windows.

PR #2937, opened 2020-12-06
Fix concurrent operations execution in OperationQueue

lxbndr:
To be able to run operations concurrently OperationQueue needs concurrent underlying DispatchQueue. This patch uses `.concurrent` attribute to create DispatchQueues with required behavior.

This matches with Darwin.

### DIFF
--- a/Sources/Foundation/Operation.swift
+++ b/Sources/Foundation/Operation.swift
@@ -938,15 +938,15 @@ open class OperationQueue : NSObject, ProgressReporting {
             let queue: DispatchQueue
             if let qos = _propertyQoS {
                 if let name = __name {
-                    queue = DispatchQueue(label: name, qos: qos.qosClass)
+                    queue = DispatchQueue(label: name, qos: qos.qosClass, attributes: .concurrent)
                 } else {
-                    queue = DispatchQueue(label: "NSOperationQueue \(Unmanaged.passUnretained(self).toOpaque())", qos: qos.qosClass)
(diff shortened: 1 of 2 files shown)


lxbndr:
cc @compnerd @spevans @millenomi

millenomi:
cc @phausler — was this intended?

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
Does this fix https://bugs.swift.org/browse/SR-11854 (OperationQueue not asynchronous) ?

lxbndr:
Oh, yes, exactly this case.

compnerd (approved):
This seems reasonable to me, though it would be nice to have an additional test to validate the state of the dispatch queue if Foundation is built testable.  That said, it would be beyond the scope of the particular fix, so I think that it is okay to do that as a follow-up.

compnerd:
CC: @tomerd - this is probably a good thing to grab for the upcoming release as well

tomerd:
> CC: @tomerd - this is probably a good thing to grab for the upcoming release as well

@compnerd sounds good but we need this merged to main first, then a PR against the 5.3 branch. 

cc @millenomi @spevans

spevans:
This will also need to go to 5.4 as well. I think #2930 may also need to be backported to 5.3 as well as its another `OperationQueue` fix

tomerd:
> This will also need to go to 5.4 as well. I think #2930 may also need to be backported to 5.3 as well as its another OperationQueue fix

sgtm, do you want to merge this to main and then @lxbndr could backport them into 5.3 and 5.4?

PR #2946, opened 2020-12-17
Convert protocol ... : class to protocol ... : AnyObject

spevans:
- Fixes 'warning: using 'class' keyword for protocol inheritance is deprecated; use 'AnyObject' instead'


spevans:
@swift-ci test

PR #2948, opened 2020-12-18
Fix redirects following POST/PUT/PATCH requests

agisboye:
I’ve stumbled upon a problem in `URLSession` when a redirect occurs following a request with a request body (POST/PUT/PATCH).

Say we create a `URLRequest` with a httpBody and an httpMethod of "POST".

When the POST request is made, a [URLSessionTask is initialized with the httpBody from the URLRequest](https://github.com/apple/swift-corelibs-foundation/blob/bb511a3ac4aef1bb0c815f33efa72c38d0692a7a/Sources/FoundationNetworking/URLSession/URLSessionTask.swift#L263).

Later, when the server responds with a redirect to this request, a new request is made but the original task is reused. This involves calling `startNewTransfer`, which in turn [looks up any existing known body](https://github.com/apple/swift-corelibs-foundation/blob/bb511a3ac4aef1bb0c815f33efa72c38d0692a7a/Sources/FoundationNetworking/URLSession/NativeProtocol.swift#L378) in the `knownBody` property. 

The result is that the new request, which is supposed to follow the redirect, also includes the original request data. This trips [the following check](https://github.com/apple/swift-corelibs-foundation/blob/bb511a3ac4aef1bb0c815f33efa72c38d0692a7a/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift#L287).

This PR attempts to work around the issue by unsetting the `knownBody` on `URLSessionTask` before a redirect request is performed. 

I also changed the behavior of how the redirect request is constructed. Previously, the original request (including httpBody) was copied and the httpMethod changed to GET. This is an invalid request.
Even though this turned out not to be the source of the problem (because the httpBody on the redirect request is never used), it doesn’t make sense for Foundation to be producing invalid requests internally.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
+++ b/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
@@ -485,6 +485,7 @@ internal class _HTTPURLProtocol: _NativeProtocol {
         case .noDelegate, .dataCompletionHandler, .downloadCompletionHandler:
             // Follow the redirect. Need to configure new request with cookies, etc.
             let configuredRequest = session._configuration.configure(request: request)
+            task?.knownBody = URLSessionTask._Body.none
             startNewTransfer(with: configuredRequest)
         }
     }
@@ -597,6 +598,7 @@ extension _HTTPURLProtocol {
         // Otherwise, we'll start a new transfer with the passed in request.
         if let r = request {
             lastRedirectBody = nil
+            task?.knownBody = URLSessionTask._Body.none
             startNewTransfer(with: r)
         } else {
             // If the redirect is not followed, return the redirect itself as the response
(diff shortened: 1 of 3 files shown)


spevans:
@swift-ci test

spevans:
I ran the test case against macOS Foundation (using the `DarwinCompatibilityTests` project) and it timed-out. I don't know if this is a bug in macOS or an incorrect test case.

agisboye:
It seems several of the tests for URLSession fail with a timeout (when testing against macOS Foundation). This happens on commits prior to mine as well so I don't think it's related to the changes that I've made. Could it be an issue with the test server implementation?

spevans:
The reason some of the tests hang on macOS is that the delegates (when used) should be calling a completion handler with `.allow` to keep the transaction progressing. There is a bug in `URLSession` on Linux that doesn't require this completion handler call so the buggy tests still work on Linux. Other tests are failing for different reasons.

My current plan for `URLSession` is to re-enable the current tests on CI (except on Windows where there are still some issues) and then fix up each test to work on macOS, while fixing `URLSession` to correctly pass the fixed tests and fail with the broken tests. Once all of the tests are passing on macOS then the other bugs can be fixed up.

So I would suggest re-examining your test to see why it is failing on macOS as it may indicate another bug in the Linux code.

agisboye:
It seems `DataTask` never calls the completion handler [here](https://github.com/apple/swift-corelibs-foundation/blob/bb511a3ac4aef1bb0c815f33efa72c38d0692a7a/Tests/Foundation/Tests/TestURLSession.swift#L2099).
I'd rather not change the behavior of `DataTask` since it's used by so many of the tests. Instead I'm using `HTTPRedirectionDataTask.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans (approved):
This looks fine now, could you just squash the commits down to one please, and then this can be merged in. Thanks

spevans:
@swift-ci test and merge

PR #2958, opened 2020-12-22
TestFoundation: stop internal http server on test case tear down

lxbndr:
Leaving server running consumes one thread from Dispatch global thread pool per one started `LoopbackServerTest`. There are 5 such test cases at the moment.

On non-Darwin platforms thread pool size is extremely small (equals to the number of CPU cores), and TestFoundation test suite quickly becomes unstable on some machines (e.g. cloud VMs with 2-core CPUs).

This has no significant impact on current CMake-generated tests, as they are started as separate process each, but trying to run whole TestFoundation suite would hang or fail some trivial tests just because global queue is stuck.

### DIFF
--- a/Tests/Foundation/HTTPServer.swift
+++ b/Tests/Foundation/HTTPServer.swift
@@ -239,7 +239,7 @@ class _TCPSocket {
         }
     }
 
-    deinit {
+    func closeSocket() throws {
         guard _socket != nil else { return }
 #if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

spevans:
Note that the `TestURLSession` tests are actually disabled in CI but I tested this locally on linux and against native foundation and it passed.

PR #2959, opened 2020-12-23
NSNumber: Forced bridging an invalid value would not trap.

spevans:
- eg `NSNumber(value: 1.2) as! Int` would return 0 instead of trapping.
  Exit with fatalError() if the bridging failed.

- This matches Darwin.

Cant add a test for trapping behaviour but a toolchain was tested v 5.3.1

```
$ cat bridge.swift 
import Foundation

let x = NSNumber(value: 1.2) as? Int
let y = NSNumber(value: 1.2) as! Int
print(x as Any, y)

$ ~/swift-5.3.1-RELEASE-ubuntu20.04/usr/bin/swift bridge.swift 
nil 0

$ ~/swift-build/toolchains/main-test/usr/bin/swift bridge.swift 
Foundation/NSNumber.swift:373: Fatal error: Unable to bridge NSNumber to Int
Please submit a bug report (https://swift.org/contributing/#reporting-bugs) and include the project and the crash backtrace.
Stack dump:
.
.
.
Illegal instruction (core dumped)
```

### DIFF
--- a/Sources/Foundation/NSNumber.swift
+++ b/Sources/Foundation/NSNumber.swift
@@ -49,7 +49,9 @@ extension Int8 : _ObjectiveCBridgeable {
     }
 
     public static func _forceBridgeFromObjectiveC(_ x: NSNumber, result: inout Int8?) {
-        result = _unconditionallyBridgeFromObjectiveC(x)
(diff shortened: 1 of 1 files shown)


spevans:
@swift-ci test

spevans:
@swift-ci test

PR #2960, opened 2020-12-29
Decimal: init?(string:) and scanDecimal() should return nil on error.

spevans:
- If the exponent is too large for Int8 then return nil instead of NaN.

- This matches Darwin's behaviour.


spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2962, opened 2021-01-08
Refactor TestJSONSerialization to fail, but not to crash on wrong JSONSerialization results

fabianfett:
### Motivation

The `JSONSerializationTests` should indicate errors, if the `JSONSerialization` result is wrong, but they shouldn't crash.

### Modifications

- `do` `catch` blocks have been replaced with `XCTAssertNoThrow` and `XCTAssertThrows` to ensure the correct behaviour. 
- accesses to result arrays `result?[0]` have been replaced with `iterator?.next()`. Ensuring no crash if an empty array has been returned by `JSONSerialization`
- **No tests have been added or removed**

Sorry for the PR size.

⚠️ Please squash this pr before merging.


spevans:
@swift-ci test

PR #2963, opened 2021-01-09
TestURLSession: Re-enable tests on Linux

spevans:
- Re-enable the TestURLSession tests which should no longer be flaky
  however specifc tests are still disabled due to known issues.

- Enable the tests on DarwinCompatibilityTests also.

- Dont re-enable these tests on Windows yet as there are still random
  crashes.

- Disable test_concurrentRequests() as this fails about 4% of the time.

- Disable test_getAllTasks() as this causes other tests to fail.

- Disable test_http0_9SimpleResponses() as this fails on Ubuntu20.04.

- Disable test_cancelledTasksCannotBeResumed() as this sometimes fails
  on Ubuntu18.04.


spevans:
@swift-ci test

spevans:
@swift-ci test macos

PR #2964, opened 2021-01-09
macOS test: cast String to NSString to conform to CVarArg

spevans:
- This isnt required on Linux or native macOS, but is just required on
  macOS for test to compile.


spevans:
@swift-ci test

PR #2965, opened 2021-01-10
TestURLSession: Pass .allow in the completionHandler.

spevans:
- Running tests on macOS under DarwinCompatibilityTests would timeout
  and fail as the .urlSession(_:dataTask:didReceive:completionHandler:)
  method would not call the completion handler at all. Update the tests
  to pass .allow.

- The tests pass on scl-f because URLSession is currently ignoring the
  value passed to the completion handler and always continuing the task,
  even though it should not be.

- The URLSession bug is logged as https://bugs.swift.org/browse/SR-14032


spevans:
@swift-ci test

PR #2966, opened 2021-01-11
Parse JSON with value types

fabianfett:
### Motivation

Increase JSON performance by parsing into a `JSONValue` type, instead of using reference types (NSNumber, NSString, ...)

### Changes

- the json parsing has been replaced with a new approach


MaxDesiatov:
@swift-ci please test

spevans:
I will have a deeper looker later but do we currently have any benchmarks for the current `JSONSerialization`? From some testing I was doing I saw that sclf is about 2-5 times slower than native foundation but it would be useful to have some to track how much of an improvement this change is and also how close to native speed it gets.

Ive also been working on an improved solution for the JSON number parsing to better deal with conversion issues between `Decimal` and `Double` with some slight speedups with bridging - I'll probably put it up as a separate PR as it will be easier to review and then look to merge it in this.

drexin:
Hey, I took the liberty to bench these changes against nightly and it's quite an improvement.

Parsing a sizable JSON doc (~8k bytes) 10,000 times takes

Nightly: 1.71179s
This PR: 0.67263s

3a4oT:
Thanks for doing that @fabianfett . I just want to point out that improvements like this are very important for server-side swift. We use [swift-extras-json](https://github.com/swift-extras/swift-extras-json) (previously Pure Swift JSON) in a production environment and it is much faster than the current version of Foundation. I would love to see at least the same numbers for Foundation which come with Swift so we can drop redundant third-party.

MaxDesiatov:
@swift-ci please test

Sources/Foundation/JSONSerialization.swift
```diff
@@ -176,33 +179,94 @@ open class JSONSerialization : NSObject {
...
+                // if utf8 all is good
+                if encoding == .utf8 {
+                    return try JSONParser().parse(bytes: ptr)
+                }
+                
+                #warning("@Fabian: pretty sure we should throw an error here, if this is invalid data")
+                let json = String(bytes: ptr, encoding: encoding!)!
+                return try json.utf8.withContiguousStorageIfAvailable { (utf8) -> JSONValue in
+                    try JSONParser().parse(bytes: utf8)
+                }!
```
fabianfett, line 208: Can this ever fail? Do we need explicit error handling here? @drexin @weissi

fabianfett:
@spevans @millenomi I'm afraid this is ready for review.

Sources/Foundation/JSONSerialization.swift
weissi: On Darwin, this can fail. On Linux it cannot.

You can write

```
var json = String(bytes:...)
json.makeContiguousUTF8()
```

and after that you can `!` it.

weissi: isn't json always utf8 though?

Sources/Foundation/JSONSerialization.swift
```diff
@@ -274,27 +338,29 @@ internal extension JSONSerialization {
...
-                if length >= 4 && bytes[2] == 0 && bytes[3] == 0 {
-                    return (.utf32LittleEndian, 4)
-                }
-                return (.utf16LittleEndian, 2)
-            case (0xFE, 0xFF):
-                return (.utf16BigEndian, 2)
-            default:
-                break
+    static func parseBOM(_ bytes: UnsafeRawBufferPointer) -> (encoding: String.Encoding, skipLength: Int)? {
+        guard bytes.count >= 2 else {
```
weissi, line 342: shouldn't we just write this as

```swift
if bytes.starts(with: [0xEF, 0xBB, 0xBF]) {
    return (.utf8, 3)
} else if bytes.starts(with: [0, 0]) {
    return (.utf32BigEndian, 4)
} else if ...
```

weissi: sadly, the optimiser doesn't get that yet... Would need to pull the arrays into global/static `let utf8BOM = [0xEF, 0xBB, 0xBF]`.

Sources/Foundation/JSONSerialization.swift
spevans: I found a failure input which is UTF8 with BOM and bad data, Here is a test case to add to 
`TestJSONSerialization.test_JSONObjectWithData_encodingDetection`:

`("{} UTF-8 w/BOM w/ trailing illegal data", [0xEF, 0xBB, 0xBF, 0x7B, 0x7D, 0xff, 0x00]),`
(will need a test to confirm the JSON parsing failed and returned `nil`)


I would suggest merging (or at least have one call the other) `parseBOM` and `detectEncoding` - I cant see anywhere else they are used. If the encoding is determined to be UTF-8 then any leading BOM can be skipped as a new `String` does not need to be created. If if looks like UTF-8 a quick check if the first character is ASCII should be sufficient conformation as I dont think any JSON can start with a unicode point > 127.

Then the code could be simplified to something like:
```

guard let (encoding, advanceBy) = parseBOM(ptr) else { throw ... }
let newPtr = ptr[advanceBy..<ptr.count]
if encoding == .utf8 {
     parse it from newPtr...
 } else {
  guard let string = String(newPtr..) else { throw .. }
}
```

I couldnt find anything on json.org to say that JSON was guaranteed to be UTF-8 but its unlikely not to be so just converting to a new `String` as you have done is fine.

I would also avoid any force unwrapping when using `String()` and just `throw` since you are dealing with external

Sources/Foundation/JSONSerialization.swift
```diff
@@ -549,523 +615,913 @@ private struct JSONWriter {
...
+        @usableFromInline /* private(set) */ var value: UInt8?
+
+        @inlinable init<Bytes: Collection>(bytes: Bytes) where Bytes.Element == UInt8 {
+            if let array = bytes as? [UInt8] {
+                self.array = array
             } else {
-                throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.propertyListReadCorrupt.rawValue,
-                              userInfo: [NSDebugDescriptionErrorKey : "Numbers must start with a 1-9 at character \(input)." ])
+                self.array = Array(bytes)
             }
```
spevans, line 1110: Does `DocumentReader` need to take the input as ` <Bytes: Collection>` ? It looks like the input is always an `UnsafeBufferPointer<UInt8>` and simply storing that struct may be faster. Calling `Array(bytes)` or `as? [UIntt8]` may end up allocating another buffer. I think the generics are just going to add overhead here.

Sources/Foundation/JSONSerialization.swift
```diff
@@ -549,523 +615,913 @@ private struct JSONWriter {
...
-        static let QuotationMark: UInt8  = 0x22 // "
-        static let Escape: UInt8         = 0x5C // \
+struct JSONParser {
+    init() {}
+
+    @inlinable
+    func parse<Bytes: Collection>(bytes: Bytes) throws
+        -> JSONValue where Bytes.Element == UInt8
+    {
+        var impl = JSONParserImpl(bytes: bytes)
```
spevans, line 625: Is there a benefit to having `JSONParserImpl`, could its code not just be folded into `JSONParser`?. Then it can be invoked as:
```
let parser = JSONParser(bytes: ...)
try parser.parse()
```

Sources/Foundation/JSONSerialization.swift
```diff
@@ -549,523 +615,913 @@ private struct JSONWriter {
...
-                    return 1
-                case .utf16BigEndian, .utf16LittleEndian:
-                    return 2
-                case .utf32BigEndian, .utf32LittleEndian:
-                    return 4
-                default:
-                    return 1
-                }
-            }()
+@usableFromInline struct JSONParserImpl {
```
spevans, line 630: Did each of these `@usableFromInline` show a benefit in benchmarks? If not I would avoid using them otherwise.

However one thing it is worth doing is tightening up the visibility of these types (`private`, `fileprivate`) since the compiler will use that information for inlining as well. If the compiler knows it doesn't have to make the type/method/ivar visible outside of this file then that gives it more hints for better code generation.

Sources/Foundation/JSONSerialization.swift
weissi: filed as https://bugs.swift.org/browse/SR-14102

Sources/Foundation/JSONSerialization.swift
parkera: from [here](https://tools.ietf.org/html/rfc4627)

```
3.  Encoding

   JSON text SHALL be encoded in Unicode.  The default encoding is
   UTF-8.

   Since the first two characters of a JSON text will always be ASCII
   characters [RFC0020], it is possible to determine whether an octet
   stream is UTF-8, UTF-16 (BE or LE), or UTF-32 (BE or LE) by looking
   at the pattern of nulls in the first four octets.

           00 00 00 xx  UTF-32BE
           00 xx 00 xx  UTF-16BE
           xx 00 00 00  UTF-32LE
           xx 00 xx 00  UTF-16LE
           xx xx xx xx  UTF-8
```

parkera: For what it’s with, Darwin Foundation checks to make sure it’s UTF8 first and if it’s not just converts it to UTF8 for parsing.

Sources/Foundation/JSONSerialization.swift
fabianfett: I'm pretty sure the compiler will create two perfectly optimized methods for this. Which means we won't have any performance hits here. I don't think we should operate on `UnsafeBufferPointer<UInt8>` here, since this comes with Security implications, based on the unsafe memory access in release builds.

Sources/Foundation/JSONSerialization.swift
fabianfett: Fixed.

Sources/Foundation/JSONSerialization.swift
fabianfett: Fixed.

Sources/Foundation/JSONSerialization.swift
fabianfett: @parkera I'm afraid you quoted RFC 4672, which has been obsoleted for quite some time by 8258:

```
   JSON text exchanged between systems that are not part of a closed
   ecosystem MUST be encoded using UTF-8 [RFC3629].

   Previous specifications of JSON have not required the use of UTF-8
   when transmitting JSON text.  However, the vast majority of JSON-
   based software implementations have chosen to use the UTF-8 encoding,
   to the extent that it is the only encoding that achieves
   interoperability.

   Implementations MUST NOT add a byte order mark (U+FEFF) to the
   beginning of a networked-transmitted JSON text.  In the interests of
   interoperability, implementations that parse JSON texts MAY ignore
   the presence of a byte order mark rather than treating it as an
   error.
```

Am I right in assuming that the Darwin implementation allows UTF16 and UTF32 encoding and further allows byte order marks?

spevans: I think the important takeaway from the 2 RFCs is that the input COULD be in any Unicode encoding but the output MUST always be UTF-8. However since the chance of the input being non-UTF8 is low the conversion from other inputs to UTF-8 as a one-time operation at the start is fine. Your code currently does all of this.

Running the test `test_JSONObjectWithData_encodingDetection` on Darwin using `DarwinCompatibilityTests` seems to indicate that Darwin doesn't handle some inputs (UTF16BE/LE witout BOM and UTF32LE w/BOM)

Sources/Foundation/JSONSerialization.swift
```diff
@@ -274,30 +339,37 @@ internal extension JSONSerialization {
...
+            return (.utf32LittleEndian, 4)
+        }
+        if bytes.starts(with: [0xFF, 0xFE]) {
+            return (.utf16LittleEndian, 2)
+        }
+        if bytes.starts(with: [0xFE, 0xFF]) {
+            return (.utf16BigEndian, 2)
+        }
+        
         return nil
```
spevans, line 363: `detectEncoding` and `parseBOM` should be merged into one function returning `(encoding: String.Encoding, skipLength: Int)? ` since they both work on the same inputs.

Sources/Foundation/JSONSerialization.swift
```diff
@@ -549,523 +621,838 @@ private struct JSONWriter {
...
-            case .utf16BigEndian where buffer[input] == 0:
-                index = input + 1
-            case .utf32BigEndian where buffer[input] == 0 && buffer[input+1] == 0 && buffer[input+2] == 0:
-                index = input + 3
-            case .utf16LittleEndian where buffer[input+1] == 0:
-                index = input
-            case .utf32LittleEndian where buffer[input+1] == 0 && buffer[input+2] == 0 && buffer[input+3] == 0:
-                index = input
+            switch extraCharacter {
+            case UInt8(ascii: " "), UInt8(ascii: "\r"), UInt8(ascii: "\n"), UInt8(ascii: "\t"):
```
spevans, line 650: These should be defined as constants to make it easier to read and reduce errors due to a typo with a characters. Eventually we could merge and reuse the constants across other parsers in sclf.

Sources/Foundation/JSONSerialization.swift
```diff
@@ -549,523 +621,838 @@ private struct JSONWriter {
...
-        static let BeginArray: UInt8     = 0x5B // [
-        static let EndArray: UInt8       = 0x5D // ]
-        static let BeginObject: UInt8    = 0x7B // {
-        static let EndObject: UInt8      = 0x7D // }
-        static let NameSeparator: UInt8  = 0x3A // :
-        static let ValueSeparator: UInt8 = 0x2C // ,
-        static let QuotationMark: UInt8  = 0x22 // "
-        static let Escape: UInt8         = 0x5C // \
+private struct JSONParser {
+
```
spevans, line 625: I think given the size of the parser now its worth moving into its own file (`JSONSerialization+Parser.swift`)? Move any associated helper functions into it as well. `JSONSerialization.swift` can be renamed to `JSONSerialization+Writer.swift` in a followup PR.

tomerd:
@swift-ci please test

millenomi (approved):

drexin:
@swift-ci test

spevans:
@swift-ci test linux

tomerd:
@swift-ci test

spevans:
@swift-ci test linux

millenomi:
@swift-ci please test

millenomi:
We will have follow-ups as bugfix PRs.

PR #2974, opened 2021-01-17
SR-13028: URLSession does not call willPerformHTTPRedirection

spevans:
- Check if the session delegate conforms to URLSessionTaskDelegate
  instead of using the task behaviour, as dataTask()s created with
  a completion handler dont have a .callDelegate behaviour. This was
  preventing the delegate method from being called.


spevans:
@swift-ci test

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2975, opened 2021-01-20
fix missing cf.CFLAGS

holymonson:
It seems get lost before 5.2, you may want some backport because it blocks building on MacOSX.

Since all `*FLAGS` are strings, should we make them empty string by default instead of `None`? And we can use `+=` instead of `=` all the time.


spevans:
@holymonson What is this python script used for? If its part of the old build system it will be removed at some point as CMake is the preferred build system for swift-corelibs-foundation

spevans:
@swift-ci test

holymonson:
> @holymonson What is this python script used for? If its part of the old build system it will be removed at some point as CMake is the preferred build system for swift-corelibs-foundation

We are still using the old `./configure`, if CMake is preferred then we will try switching to it. This seems no need to fix because there are some more incorrect in header list and source list.

PR #2978, opened 2021-01-23
[UserDefaults.swift] - Improved readability of some codes.

mustafagunes:
Hello, everybody 🖐   My first time contributing to Apple open source projects. My goal is to make the UserDefaults class more readable. I've reconfigured some codes without breaking the logical fictions.

### DIFF
--- a/Sources/Foundation/UserDefaults.swift
+++ b/Sources/Foundation/UserDefaults.swift
@@ -22,10 +22,8 @@ open class UserDefaults: NSObject {
         let value = bridgeFromNSCFTypeIfNeeded(nonbridgedValue)
         
         if let value = value as? [Any] {
-            for innerValue in value {
-                if !_isValueAllowed(innerValue) {
-                    return false
-                }
+            for innerValue in value where !_isValueAllowed(innerValue) {
+                return false
             }
             
             return true
@@ -50,7 +48,7 @@ open class UserDefaults: NSObject {
             return true
         }
         
-        let isOfCommonTypes =  value is String || value is Data || value is Date || value is Int || value is Bool || value is CGFloat
+        let isOfCommonTypes = value is String || value is Data || value is Date || value is Int || value is Bool || value is CGFloat
         if isOfCommonTypes {
             return true
         }
@@ -60,19 +58,15 @@ open class UserDefaults: NSObject {
     }
     
     static private func _unboxingNSNumbers(_ value: Any?) -> Any? {
-        if value == nil {
+        guard value != nil else {
             return nil
         }
         
         if let number = value as? NSNumber {
             return number._swiftValueOfOptimalType
-        }
-        
-        if let value = value as? [Any] {
+        } else if let value = value as? [Any] {
             return value.map(_unboxingNSNumbers)
-        }
-        
-        if let value = value as? [AnyHashable: Any] {
+        } else if let value = value as? [AnyHashable: Any] {
             return value.mapValues(_unboxingNSNumbers)
         }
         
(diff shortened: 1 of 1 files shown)


Sources/Foundation/UserDefaults.swift
```diff
@@ -50,7 +48,7 @@ open class UserDefaults: NSObject {
             return true
         }
         
-        let isOfCommonTypes =  value is String || value is Data || value is Date || value is Int || value is Bool || value is CGFloat
+        let isOfCommonTypes = value is String || value is Data || value is Date || value is Int || value is Bool || value is CGFloat
```
mustafagunes, line 51: I wanted to reconfigure such long lines of code as follows, but I want to ask first. Are you writing such long lines of code because of performance anxiety? Or is that how it was forgotten?

```suggestion
        let isOfCommonTypes = value is String 
            || value is Data 
            || value is Date 
            || value is Int 
            || value is Bool 
            || value is CGFloat
```

mustafagunes:
@swift-ci please test

mustafagunes:
cc: @stephencelis @phausler

Sources/Foundation/UserDefaults.swift
```diff
@@ -170,62 +163,64 @@ open class UserDefaults: NSObject {
...
-        }
-        if let bVal = aVal as? Double {
+        } else if let bVal = aVal as? Double {
             return bVal != 0
-        }
-        if let bVal = aVal as? String {
+        } else if let bVal = aVal as? String {
             return NSString(string: bVal).boolValue
         }
+
```
stevapple, line 220: In Swift, we have a really nice `case let` syntax:

```swift
switch aVal {
case let bVal as String:
  // the same as `if let bVal = aVal as? String`
…
}
```

Sources/Foundation/UserDefaults.swift
```diff
@@ -319,16 +320,16 @@ open class UserDefaults: NSObject {
         return names
     }
     
-    open func volatileDomain(forName domainName: String) -> [String : Any] {
+    open func volatileDomain(forName domainName: String) -> [String: Any] {
```
stevapple, line 323: I would suggest not do such changes since `[Key : Value]` is more commonly used to express a `Dictionary` type. Its symmetrical look is much neater than the unsymmetrical `[Key: Value]`.

Sources/Foundation/UserDefaults.swift
```diff
@@ -60,19 +58,15 @@ open class UserDefaults: NSObject {
...
-        }
-        
-        if let value = value as? [Any] {
+        } else if let value = value as? [Any] {
             return value.map(_unboxingNSNumbers)
-        }
-        
-        if let value = value as? [AnyHashable: Any] {
+        } else if let value = value as? [AnyHashable: Any] {
             return value.mapValues(_unboxingNSNumbers)
```
stevapple, line 70: As long as we `return` in every `if` body, `else if` would not make codes better to read. IMO this would just introduce unnecessary inconsistency between these parallel `if`s.

If you want to make the codes shorter and more verbose, use `switch` instead.

Sources/Foundation/UserDefaults.swift
stevapple: 🤔IMO leaving it alone is okay because because this line was really simple and easy. Splitting it into multiple lines wouldn’t affect anything (including performance) since it even makes no difference in SIL, but may make the single statement taking up unnecessarily large visual space.

stevapple (changes requested):

AtariDreams:
Any updates on this?

PR #2979, opened 2021-01-27
FoundationXML: convert CFXMLInterface to ImplementationOnly

compnerd:
The CoreFoundation interfaces are not expected to be exposed.  Convert
the import of `CFXMLInterface` to `@_implementationOnly`.


compnerd:
@swift-ci please test

PR #2980, opened 2021-02-04
JSONSerialization: Improve number parsing for JSON

spevans:
This is a potential solution for SR-7054 and SR-12244 involving parsing JSON numbers and then later bridging them to `Decimal` or a floating point type and losing accuray compared to the original input. It will need to be updated after #2966 is merge but that will be minimal changes. The bridging should also be marginally faster as it doesnt need to create a 2nd `NSNumber` for the equality check.

- Add _NSJSONNumber, an internal subclass of NSNumber to provide lazy
  parsing of numbers at time of use. This avoids parsing numbers as
  Decimal when later bridged to Double/Float and vice-versa which
  has accuracy issues if converted from one to the other.

- Add test cases for SR-7054 and SR-12244

### DIFF
--- a/Sources/Foundation/JSONSerialization+Parser.swift
+++ b/Sources/Foundation/JSONSerialization+Parser.swift
@@ -73,7 +73,7 @@ internal struct JSONParser {
             case UInt8(ascii: "-"), UInt8(ascii: "0") ... UInt8(ascii: "9"):
                 reader.moveReaderIndex(forwardBy: whitespace)
                 let number = try self.reader.readNumber()
-                return .number(number)
+                return .inputNumber(number)
             case ._space, ._return, ._newline, ._tab:
                 whitespace += 1
                 continue
@@ -263,7 +263,7 @@ extension JSONParser {
             try self.readUTF8StringTillNextUnescapedQuote()
         }
         
-        mutating func readNumber() throws -> String {
+        mutating func readNumber() throws -> _JSONNumber {
             try self.parseNumber()
         }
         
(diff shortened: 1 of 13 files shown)


spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

Sources/Foundation/JSONSerialization+Parser.swift
```diff
@@ -516,19 +516,42 @@ extension JSONParser {
...
             var pastControlChar: ControlCharacter = .operand
             var numbersSinceControlChar: UInt = 0
             var hasLeadingZero = false
 
+            // Tracking digits for integer/exponent, eg parsing 12.34, exponent = -2 _integerMagnitude = 1234
+            var integerMagnitude: UInt64? = 0
+            var fractionExponent: Int?
+            var exponentValue: Int?
+            var negativeExponent = false
+            var jsonString = ""
```
fabianfett, line 554: Thanks so much for moving your implementation into this `parseNumber` function! 👍 I guess the only thing left from my side here, is the creation of `jsonString`. In my point of view, we should only create the `jsonString` at the very end of the number parsing based on the numbers arraySlice.

```swift
self.makeStringFast(self[numberStartIndex ..< self.readerIndex])
```

Appending to a String especially with `Character(UnicodeScalar(ascii))` is, if I recall correctly, very slow.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

parkera:

PR #2981, opened 2021-02-12
Store the hostname in _names for Host.current()

Lukasa:
Host.current() looks up the hostname for its _info field, but then never
gives it to the user. As a result, .names is always empty for
Host.current().

This patch is a minimal-effort workaround: it simply throws the value of
gethostname() (or its Windows equivalent) into _names. We could probably
do better than that, but for now this is good enough.

Resolves SR-14197.

### DIFF
--- a/Sources/Foundation/Host.swift
+++ b/Sources/Foundation/Host.swift
@@ -118,7 +118,7 @@ open class Host: NSObject {
         return addresses.firstIndex { aHost.addresses.contains($0) } != nil
     }
     
-    internal func _resolveCurrent() {
+    internal func _resolveCurrent(withInfo info: String) {
 #if os(Windows)
         var szAddress: [WCHAR] =
             Array<WCHAR>(repeating: 0, count: Int(NI_MAXHOST))
@@ -165,6 +165,7 @@ open class Host: NSObject {
 
           pAdapter = pAdapter!.pointee.Next
         }
+        _names = [info]
         _resolved = true
 #else
         var ifaddr: UnsafeMutablePointer<ifaddrs>? = nil
@@ -189,6 +190,7 @@ open class Host: NSObject {
             }
             ifa = ifaValue.ifa_next
         }
+        _names = [info]
         _resolved = true
 #endif
     }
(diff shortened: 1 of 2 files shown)


Lukasa:
@swift-ci please smoke test

Lukasa:
@swift-ci test

Lukasa:
@swift-ci test

Lukasa:
I think the macOS failure is unrelated.

Tests/Foundation/Tests/TestHost.swift
```diff
@@ -49,5 +49,14 @@ class TestHost: XCTestCase {
         let google = Host(name: "google.com")
         XCTAssertFalse(swift0.isEqual(to: google))
     }
+
+    // SR-14197
+    func test_localNamesNonEmpty() {
```
spevans, line 54: This test will need to be added to the `allTests` array at the top of the file.

Sources/Foundation/Host.swift
```diff
@@ -165,6 +165,8 @@ open class Host: NSObject {
 
           pAdapter = pAdapter!.pointee.Next
         }
+        // The force-uwrap here is safe, as if _type is .current then _info is always non-nil.
+        _names = [_info!]
```
spevans, line 169: You could change the signature of this function to `_resolveCurrent(withInfo info: String)` and fixup the callers since `_info` is being unwrapped in the `if` statements. Would simplify having to reason about the force unwrap.

spevans:
> I think the macOS failure is unrelated.

Yeah, it doesnt actually build or test Foundation anyway!

Sources/Foundation/Host.swift
Lukasa: Sure, sounds good.

Lukasa:
@swift-ci test

Tests/Foundation/Tests/TestHost.swift
spevans: @Lukasa The test still needs to be added in

Lukasa:
Updated again. :)

Lukasa:
@swift-ci test

PR #2982, opened 2021-02-12
Added missing `TARGET_OS_CYGWIN` defines to TargetConditionals.h

BenLeggiero:
These seemed to be unintentionally undefined in some cases, so I added them


spevans:
@swift-ci test

PR #2984, opened 2021-02-20
[5.4] Judge executables with API on Windows

stevapple:
- Use `GetBinaryTypeW` to identify if the file is an executable file.

(cherry picked from commit 8a87b13)


spevans:
@swift-ci test

PR #2985, opened 2021-03-02
JSON Decoding based on JSONValue

fabianfett:
This is a follow up pr to #2966.

It should significantly improve overall JSON performance, since we save on a lot of roundtrips through `NSArray` and `NSDictionary`.

### DIFF
--- a/Sources/Foundation/JSONEncoder.swift
+++ b/Sources/Foundation/JSONEncoder.swift
@@ -1152,1215 +1152,118 @@ open class JSONDecoder {
     /// - throws: `DecodingError.dataCorrupted` if values requested from the payload are corrupted, or if the given data is not valid JSON.
     /// - throws: An error if any value throws an error during decoding.
     open func decode<T : Decodable>(_ type: T.Type, from data: Data) throws -> T {
-        let topLevel: Any
         do {
-            topLevel = try JSONSerialization.jsonObject(with: data, options: .allowFragments)
-        } catch {
+            var parser = JSONParser(bytes: Array(data))
+            let json = try parser.parse()
+            return try JSONDecoderImpl(userInfo: self.userInfo, from: json, codingPath: [], options: self.options).unwrap(as: T.self)
+        } catch let error as JSONError {
             throw DecodingError.dataCorrupted(DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: error))
+        } catch {
+            throw error
         }
-
-        let decoder = _JSONDecoder(referencing: topLevel, options: self.options)
-        guard let value = try decoder.unbox(topLevel, as: type) else {
-            throw DecodingError.valueNotFound(type, DecodingError.Context(codingPath: [], debugDescription: "The given data did not contain a top-level value."))
-        }
-
-        return value
     }
 }
 
 // MARK: - _JSONDecoder
 
-fileprivate class _JSONDecoder : Decoder {
-    // MARK: Properties
+fileprivate struct JSONDecoderImpl {
+    let codingPath: [CodingKey]
+    let userInfo: [CodingUserInfoKey: Any]
 
-    /// The decoder's storage.
-    fileprivate var storage: _JSONDecodingStorage
+    let json: JSONValue
+    let options: JSONDecoder._Options
 
-    /// Options set on the top-level decoder.
-    fileprivate let options: JSONDecoder._Options
-
-    /// The path to the current point in encoding.
-    fileprivate(set) public var codingPath: [CodingKey]
-
-    /// Contextual user-provided information for use during encoding.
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

Sources/Foundation/JSONSerialization.swift
```diff
@@ -737,3 +701,48 @@ private extension JSONValue {
         }
     }
 }
+
+extension NSNumber {
+    static func fromJSONNumber(_ string: String) -> NSNumber? {
```
millenomi, line 706: This should probably be a constructor (`init?(jsonNumberString string: String)`).

millenomi:
@bendjones Can you take a look? @fabianfett is working on a full reimplementation at the Swift layer of JSONSerialization in SCF, and this patch has performance improvements by accessing the underlying JSONSerialization implementation instead of calling into JSONSerialization and then working on the returned values.

Sources/Foundation/JSONSerialization.swift
fabianfett: In some cases this might return a `NSDecimalNumber`. I don't think that we can make it work with a `convenience` initializer (which is the only option we have in an extension).

drexin:
@swift-ci test

benlangmuir:
https://github.com/apple/sourcekit-lsp/pull/372

benlangmuir:
https://github.com/apple/sourcekit-lsp/pull/372
@swift-ci please test

spevans:
@swift-ci test

drexin:
@swift-ci test

Sources/Foundation/JSONEncoder.swift
```diff
@@ -1152,1359 +1152,801 @@ open class JSONDecoder {
...
 
-        self.decoder.codingPath.append(key)
-        defer { self.decoder.codingPath.removeLast() }
+        func decode<T>(_: T.Type) throws -> T where T: Decodable {
+            try self.impl.unwrap(as: T.self)
+        }
 
-        guard let value = try self.decoder.unbox(entry, as: UInt.self) else {
-            throw DecodingError.valueNotFound(type, DecodingError.Context(codingPath: self.decoder.codingPath, debugDescription: "Expected \(type) value but found null instead."))
+        @inline(__always) private func decodeFixedWidthInteger<T: FixedWidthInteger>() throws -> T {
```
bendjones, line 1568: Isn't the trick for **really really I mean it inline me** `@inlinable @inline(__always)`?

fabianfett: `@inlinable` can only be applied to `public` functions AFAIK.

Sources/Foundation/JSONSerialization.swift
bendjones: Yeah seems to be the case even though `NSDecimalNumber` is a subclass of `NSNumber`

Sources/Foundation/JSONEncoder.swift
bendjones: Ah true that's the time to use that trick...

bendjones (approved):
This all LGTM it's a large patch but I read through it twice to make sure I didn't miss something major

millenomi:
Merging pending test passes.

Tests/Foundation/Tests/PolymorphicCodableTests.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  PolymorphicCodableTests.swift
+//  TestFoundation
+//
+//  Created by Fabian Fett on 02.03.21.
+//  Copyright © 2021 Swift. All rights reserved.
+//
```
tomerd, line 7: Needs a better header? Or maybe file is redundant?

fabianfett: urgh. this file was not intended to be commited.

drexin:
@swift-ci test

Sources/Foundation/JSONEncoder.swift
```diff
@@ -1152,1359 +1152,801 @@ open class JSONDecoder {
...
+        }
+        
+        if let decimal = Decimal(string: numberString) {
+            return decimal
         }
-        self.codingPath = decoder.codingPath
+        
+        if let double = Double(numberString) {
+            return Decimal(double)
+        }
```
spevans, line 1331: It might be best avoid the fallback to `Double` parsing. Converting a `Double` to `Decimal` is lossy and if the string does not parse as a `Decimal` anyway it can be considered an invalid value for the type.

fabianfett: fixed.

spevans:
@swift-ci test

PR #2986, opened 2021-03-02
macOS: Dont import CFXMLInterface as @_implementationOnly as this breaks

spevans:
- This doesnt work on macOS for some reason, so just import it as a
  normal import. This was breaking testing on macOS.

- This doesnt affect any other platform where it is actually required
  to @_implementationOnly import.

This reverts #2979 on macOS


spevans:
@swift-ci test

PR #2987, opened 2021-03-03
JSON Encoding based on JSONValue

fabianfett:
After #2966 and #2985, this is the last step to make JSON Encoding and Decoding much more performant on Linux.

### DIFF
--- a/Sources/Foundation/JSONEncoder.swift
+++ b/Sources/Foundation/JSONEncoder.swift
@@ -237,176 +237,405 @@ open class JSONEncoder {
     /// - returns: A new `Data` value containing the encoded JSON data.
     /// - throws: `EncodingError.invalidValue` if a non-conforming floating-point value is encountered during encoding, and the encoding strategy is `.throw`.
     /// - throws: An error if any value throws an error during encoding.
-    open func encode<T : Encodable>(_ value: T) throws -> Data {
-        let encoder = _JSONEncoder(options: self.options)
- 
-        guard let topLevel = try encoder.box_(value) else {
+    open func encode<T: Encodable>(_ value: T) throws -> Data {
+        let value: JSONValue = try encodeAsJSONValue(value)
+        let writer = JSONValue.Writer(options: self.outputFormatting)
+        let bytes = writer.writeValue(value)
+        
+        return Data(bytes)
+    }
+
+    func encodeAsJSONValue<T: Encodable>(_ value: T) throws -> JSONValue {
+        let encoder = JSONEncoderImpl(options: self.options, codingPath: [])
+        guard let topLevel = try encoder.wrapEncodable(value, for: nil) else {
             throw EncodingError.invalidValue(value, EncodingError.Context(codingPath: [], debugDescription: "Top-level \(T.self) did not encode any values."))
         }
-
-        let writingOptions = JSONSerialization.WritingOptions(rawValue: self.outputFormatting.rawValue).union(.fragmentsAllowed)
-
-        do {
-            return try JSONSerialization.data(withJSONObject: topLevel, options: writingOptions)
-        } catch {
-            throw EncodingError.invalidValue(value, EncodingError.Context(codingPath: [], debugDescription: "Unable to encode the given top-level value to JSON.", underlyingError: error))
-        }
+        
+        return topLevel
     }
 }
 
 // MARK: - _JSONEncoder
 
-fileprivate class _JSONEncoder : Encoder {
-    // MARK: Properties
+private enum JSONFuture {
+    case value(JSONValue)
+    case encoder(JSONEncoderImpl)
+    case nestedArray(RefArray)
+    case nestedObject(RefObject)
+    
+    class RefArray {
+        private(set) var array: [JSONFuture] = []
 
-    /// The encoder's storage.
-    fileprivate var storage: _JSONEncodingStorage
-
-    /// Options set on the top-level encoder.
-    fileprivate let options: JSONEncoder._Options
-
-    /// The path to the current point in encoding.
-    public var codingPath: [CodingKey]
-
-    /// Contextual user-provided information for use during encoding.
-    public var userInfo: [CodingUserInfoKey : Any] {
-        return self.options.userInfo
-    }
-
-    // MARK: - Initialization
-
-    /// Initializes `self` with the given top-level encoder options.
-    fileprivate init(options: JSONEncoder._Options, codingPath: [CodingKey] = []) {
-        self.options = options
-        self.storage = _JSONEncodingStorage()
-        self.codingPath = codingPath
-    }
-
-    /// Returns whether a new element can be encoded at this coding path.
-    ///
-    /// `true` if an element has not yet been encoded at this coding path; `false` otherwise.
-    fileprivate var canEncodeNewValue: Bool {
-        // Every time a new value gets encoded, the key it's encoded for is pushed onto the coding path (even if it's a nil key from an unkeyed container).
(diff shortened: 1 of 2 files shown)


drexin:
@swift-ci test

Sources/Foundation/JSONEncoder.swift
```diff
@@ -237,176 +237,405 @@ open class JSONEncoder {
...
+                return array
+            }
+        }
 
-    /// The path to the current point in encoding.
-    public var codingPath: [CodingKey]
+        @inline(__always) func setObject(for key: String) -> Object {
+            switch self.dict[key] {
+            case .encoder:
+                preconditionFailure("For key \"\(key)\" an encoder has already been created.")
```
tomerd, line 338: is it safe to precondition here? i.e no user action can cause this?

Sources/Foundation/JSONEncoder.swift
```diff
@@ -237,176 +237,405 @@ open class JSONEncoder {
...
+        case .millisecondsSince1970:
+            return .number((date.timeIntervalSince1970 * 1000).description)
 
-    /// Initializes `self` with no containers.
-    fileprivate init() {}
+        case .iso8601:
+            if #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) {
+                return .string(_iso8601Formatter.string(from: date))
+            } else {
+                fatalError("ISO8601DateFormatter is unavailable on this platform.")
```
tomerd, line 549: hmmm I think we can use regular formatter with right pattern to achieve iso8601, no?

Sources/Foundation/JSONSerialization.swift
```diff
@@ -737,3 +737,217 @@ private extension JSONValue {
...
+                while let item = iterator.next() {
+                    bytes.append(contentsOf: [._comma, ._newline])
+                    self.addInset(to: &bytes, depth: depth + 1)
+                    self.writeValuePretty(item, into: &bytes, depth: depth + 1)
+                }
+                bytes.append(._newline)
+                self.addInset(to: &bytes, depth: depth)
+                bytes.append(._closebracket)
+            case .object(let dict):
+                if #available(OSX 10.13, *), options.contains(.sortedKeys) {
```
tomerd, line 849: doesn't linux support this option?

Sources/Foundation/JSONEncoder.swift
tomerd: also afaik, iso8601 dateEncodingStrategy I supported now on linux

Sources/Foundation/JSONSerialization.swift
```diff
@@ -737,3 +737,217 @@ private extension JSONValue {
...
+                if let first = iterator.next() {
+                    self.writeValue(first, into: &bytes)
+                }
+                while let item = iterator.next() {
+                    bytes.append(._comma)
+                    self.writeValue(item, into:&bytes)
+                }
+                bytes.append(._closebracket)
+            case .object(let dict):
+                if #available(OSX 10.13, *), options.contains(.sortedKeys) {
```
tomerd, line 787: ditto, should be supported on linux afaik

Sources/Foundation/JSONEncoder.swift
tomerd: I see it now

Sources/Foundation/JSONSerialization.swift
tomerd: I see it now

Sources/Foundation/JSONEncoder.swift
fabianfett: This shall lead to a crash because of a programmer error and closely mimics the Darwin Foundation behavior. It is not possible to create a json array or a json dictionary at the same time for the same key.

Sources/Foundation/JSONEncoder.swift
```diff
@@ -237,176 +237,405 @@ open class JSONEncoder {
...
-    open func encode<T : Encodable>(_ value: T) throws -> Data {
-        let encoder = _JSONEncoder(options: self.options)
- 
-        guard let topLevel = try encoder.box_(value) else {
+    open func encode<T: Encodable>(_ value: T) throws -> Data {
+        let value: JSONValue = try encodeAsJSONValue(value)
+        let writer = JSONValue.Writer(options: self.outputFormatting)
+        let bytes = writer.writeValue(value)
+        
+        return Data(bytes)
```
spevans, line 245: If `JSONValue.Writer` wrote directly into a `Data` instead of a `[UInt8]` then the `Data(bytes)` which is doing a copy of the `[UInt8]` could be eliminated and `return writer.writeValue(value)` used instead.

Sources/Foundation/JSONEncoder.swift
```diff
@@ -417,570 +646,338 @@ fileprivate struct _JSONKeyedEncodingContainer<K : CodingKey> : KeyedEncodingCon
...
-    public mutating func superEncoder() -> Encoder {
-        return _JSONReferencingEncoder(referencing: self.encoder, at: self.container.count, wrapping: self.container)
+extension JSONKeyedEncodingContainer {
+    @inline(__always) private mutating func encodeFloatingPoint<F: FloatingPoint & CustomStringConvertible>(_ float: F, key: CodingKey) throws {
+        let value = try self.wrapFloat(float, for: key)
+        self.object.set(value, for: key.stringValue)
+    }
+    
+    @inline(__always) private mutating func encodeFixedWidthInteger<N: FixedWidthInteger>(_ value: N, key: CodingKey) throws {
+        self.object.set(.number(value.description), for: key.stringValue)
```
spevans, line 754: All of the callers of this function are already expanded out per type, could this line not just be used at each call size instead?

Sources/Foundation/JSONEncoder.swift
```diff
@@ -237,176 +237,405 @@ open class JSONEncoder {
...
-            return try JSONSerialization.data(withJSONObject: topLevel, options: writingOptions)
-        } catch {
-            throw EncodingError.invalidValue(value, EncodingError.Context(codingPath: [], debugDescription: "Unable to encode the given top-level value to JSON.", underlyingError: error))
+private enum JSONFuture {
+    case value(JSONValue)
+    case encoder(JSONEncoderImpl)
+    case nestedArray(Array)
+    case nestedObject(Object)
+    
+    class Array {
```
bendjones, line 266: Could we name this something more specific? I get it's a private type but that `case nestedArray(Array)` threw me for a second.

Sources/Foundation/JSONEncoder.swift
fabianfett: If it is okay for you, I'd prefer to keep it this way. Since Swift does not support Macros this is the closest to Macros we can get (`@inline(__always)`). In case we ever need to change something in the encoding method for `Int`s we only need to change it at one point.

Sources/Foundation/JSONEncoder.swift
fabianfett: I just run a small performance test
```
JSONValue to [UInt8] to Data | took: 0.48609s
JSONValue to Data            | took: 5.17270s
```

I guess for now we should stay with `[UInt8]`

bendjones: Yeah `[UInt8]` would be preferred I believe

spevans:
@swift-ci test

bendjones (approved):
Gave it another read through LGTM

bendjones:
cc @millenomi

PR #2989, opened 2021-03-08
JSONDecoder: Check that a parsed float does not return infinity.

spevans:
- The behaviour of the floating point string parsing for Float and
  Double was changed to return +/-infinity instead of nil on overflow.
  (https://github.com/apple/swift/pull/34339)

- Add an isFinite check to ensure that the result of a floating point
  conversion still fits in the type, matching current behaviour.


spevans:
@swift-ci test

spevans:
cc @fabianfett

fabianfett (approved):

spevans:
@swift-ci test linux

spevans:
@swift-ci test linux

PR #2990, opened 2021-03-17
Add necessary missing TARGET_OS_BSD directives.

3405691582:
A number of Foundation tests are failing on OpenBSD due to these missing
directives, or are required for other reasons.

There is still more complex work to get full support for Foundation on
this platform but this will be handled in other commits.


spevans:
@swift-ci test

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -1649,7 +1649,7 @@ int _CFOpenFile(const char *path, int opts) {
 }
 
 CF_CROSS_PLATFORM_EXPORT void *_CFReallocf(void *ptr, size_t size) {
-#if TARGET_OS_WIN32 || TARGET_OS_LINUX
+#if TARGET_OS_WIN32 || TARGET_OS_LINUX || defined(__OpenBSD__)
```
spevans, line 1652: Would it be worth adding a `TARGET_OS_OPENBSD`?

3405691582: I don't know if it would be worth it. There shouldn't be too many points of divergence. Here, FreeBSD has reallocf and OpenBSD doesn't.

spevans:
@swift-ci test linux

3405691582:
Test failure appears unrelated, I think.

spevans:
@swift-ci test linux

spevans:
@swift-ci test

3405691582:
Directives added for CFUtilities.c while we are here. @spevans can we retest and merge if everything else is OK?

spevans:
@swift-ci test

spevans (approved):

PR #2994, opened 2021-03-29
Cleanup lifetime management around NSData.bytes.

Lukasa:
Several blocks of code in Foundation use the .bytes member on NSData to
construct an interior pointer. However, NSData can potentially be freed
very early, and in newer Swift releases is more likely to do so, leading
to this pointer dangling.

This patch attempts a minimal change by forcing the NSData object to
live longer using withExtendedLifetime. This should resolve the misuse
fairly effectively. It does not address cases where the code could
potentially be entirely rewritten to be more efficient or more sensible,
because I wasn't here to engage in a complete refactor.


spevans:
@swift-ci test

spevans:
Could you apply the patch in https://gist.github.com/spevans/c2c6a15b4cae1c376564ea434bda3eea
It should fix the compile errors.

Lukasa:
@swift-ci please test

Lukasa:
@swift-ci please test

Lukasa:
@spevans So it looks like we're down to simple pre-existing test failures in XML. I assume we don't have any particular reason to think those XML failures are related to this code. Are you happy to merge as-is or do we need to fix that failure too?

spevans:
Correct, the XML code does have long standing underlying memory issues. Could you apply this patch to yours to disable that XML test:

https://gist.github.com/spevans/a7306641613a6531b1e977f321ea7bbb

And then we can retest and merge. I testest your PR with that patch and it passes on Ubuntus16,18 & 20

I think we should also backport your fixes to 5.4 (without the XML test disable)

Lukasa:
Specifically, it seems like the CFXML function `_searchNamespace` just...does not work.

Lukasa:
@swift-ci please test

spevans:
@swift-ci test

spevans:
@swift-ci test linux

Tests/Foundation/Tests/TestXMLDocument.swift
```diff
@@ -257,45 +257,48 @@ class TestXMLDocument : LoopbackServerTest {
...
-        XCTAssertEqual(element.attribute(forLocalName: "name", uri: uriNs1)?.stringValue, "Tommy", "ns1:name==Tommy")
-        
-        // Remove attributes
-        element.removeAttribute(forName: "name")
-        XCTAssertNil(element.attribute(forLocalName: "name", uri: nil), "name removed")
-        XCTAssertNotNil(element.attribute(forLocalName: "name", uri: uriNs1), "ns1:name not removed")
-        element.removeAttribute(forName: "ns1:name")
-        XCTAssertNil(element.attribute(forLocalName: "name", uri: uriNs1), "ns1:name removed")
+
+        withExtendedLifetime(root) {
```
atrick, line 261: If you don't want to indent a bunch of code, you can do
`defer { withExtendedLifetime(root) {} }`

spevans:
@swift-ci test linux

Lukasa:
@swift-ci test linux

spevans:
I believe there are still some further fixes that will be required given the number of uses of `unsafeBitCast` in sclf to pass objects to `CoreFoundation` functions, however all of these fixes in this PR are valid so am going to merge this.

PR #2997, opened 2021-05-01
[SR-14496] - When Base64 encoding, only add line separators if there …

ogres:
Line separators were added at the end of base64 encoded string even though the result was fitting in the given max-line length limit.

Adjusted base64 encoding to match Darwin's behaviour - now we will only add separators if there are more bytes to parse.

Resolves https://bugs.swift.org/browse/SR-14496


spevans:
@swift-ci test

PR #3000, opened 2021-05-25
Set `isRunning` of `Process` to `false` before calling the termination handler

ahoppen:
Otherwise, we end up in a race condition where `isRunning` might still be `true` when accessed from inside the `terminationHandler`. If this is the case, accessing `terminationReason` or `terminationStatus` inside the termination handler crashes because their precondition that the task has finished are not satisfied.

We are experiencing this flaky behavior in [SourceKit-LSP’s](https://github.com/apple/sourcekit-lsp) crash recovery tests on Linux.

Fixes rdar://78035044

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -830,6 +830,9 @@ open class Process: NSObject {
                 process._terminationReason = .exit
             }
             
+            // Set the running flag to false
+            process.isRunning = false
+
             // If a termination handler has been set, invoke it on a background thread
             
             if let terminationHandler = process.terminationHandler {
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

spevans (approved):
LGTM

spevans:
@ahoppen Might be good to backport to 5.4 as well since this fixes an obvious bug

ahoppen:
@spevans Do you mean to backport to 5.5 or to 5.4 and 5.5? Only back porting to 5.4 doesn’t make sense to me, unless I’m missing something about the corelibs-foundation branching scheme.

spevans:
Sorry, yes I meant the 5.5 branch.

ahoppen:
OK, set up a 5.5 PR here: https://github.com/apple/swift-corelibs-foundation/pull/3002

PR #3002, opened 2021-05-26
[5.5] Set `isRunning` of `Process` to `false` before calling the termination handler

ahoppen:
Cherry-picks https://github.com/apple/swift-corelibs-foundation/pull/3000 to `release/5.5`.

---------------------

Otherwise, we end up in a race condition where `isRunning` might still be `true` when accessed from inside the `terminationHandler`. If this is the case, accessing `terminationReason` or `terminationStatus` inside the termination handler crashes because their precondition that the task has finished are not satisfied.

Fixes rdar://78035044

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -830,6 +830,9 @@ open class Process: NSObject {
                 process._terminationReason = .exit
             }
             
+            // Set the running flag to false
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

spevans (approved):

PR #3003, opened 2021-07-09
Omnibus to get a clean OpenBSD build.

3405691582:
This pr includes a few changes -- not including runloop changes which will be in a different pr -- to get a clean Foundation build on OpenBSD. Both this pr and the runloop changes are required for Foundation to build properly, work, and pass tests on this platform.

Please feel free to request any or all to be split into multiple prs.

Of note:
* a simple routine to search the runtime PATH to avoid having to specify the full binary name in the environment (kernel support to determine the full binary path on this platform is not present)
* supplying an implementation for `FileManager+POSIX`
* XFAIL'ing a few tests that rely on backtrace support
* handling some missing symbols and other minor API differences


spevans:
@swift-ci test

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -175,6 +181,42 @@ const char *_CFProcessPath(void) {
...
+      while ((p = strsep(&paths, ":")) != NULL) {
+         char pp[PATH_MAX];
+         snprintf(pp, PATH_MAX, "%s/%s", p, argv0);
+         if (!__CFProcessIsRestricted() && access(pp, X_OK) == 0) {
+             _CFSetProgramNameFromPath(argv0);
+             return __CFProcessPath;
+         }
+      }
+    }
+
```
spevans, line 218: Would this have issues if the executable was specified using a relative path ,eg `../some_executable` ? It would fail the first check but not necessarily be in the `PATH`

spevans:
@swift-ci test

CoreFoundation/Base.subproj/CFPlatform.c
3405691582: Yes, that would not work if the realpath call failed and we fell back to argv[0]. We can call realpath a second time to ensure we get an absolute path. Added missing free calls as well.

3405691582:
Addressed the relative path issue, also amended a small cmake change that I missed earlier.

spevans:
@swift-ci test

spevans:
@swift-ci test linux

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -175,6 +181,50 @@ const char *_CFProcessPath(void) {
...
+    // Search PATH.
+    if (argv0) {
+      char *paths = getenv("PATH");
+      char *p = NULL;
+      while ((p = strsep(&paths, ":")) != NULL) {
+         char pp[PATH_MAX];
+         snprintf(pp, PATH_MAX, "%s/%s", p, argv0);
+         char *res = realpath(pp, NULL);
```
spevans, line 212: if `snprintf` returns >= `PATH_MAX` the path in `pp` will be truncated and thus invalid.

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -175,6 +181,50 @@ const char *_CFProcessPath(void) {
...
+         if (!__CFProcessIsRestricted() && access(res, X_OK) == 0) {
+             _CFSetProgramNameFromPath(res);
+             free(argv0);
+             free(res);
+             return __CFProcessPath;
+         }
+	 free(res);
+      }
+      free(argv0);
+    }
```
spevans, line 225: Could you just tidy up the formatting in this `if` block so its all consistent at 4 space indent, thanks.

CoreFoundation/Base.subproj/CFPlatform.c
3405691582: That's fine, if the path to the binary is going to be too long, we will truncate and when we try and see if the truncated file exists, we'll get an error and we'll fall out of this case. Although, did you have an alternative suggestion in mind?

CoreFoundation/Base.subproj/CFPlatform.c
3405691582: Oops -- sorry about that. Fixed.

CoreFoundation/Base.subproj/CFPlatform.c
spevans: The truncated name could still be another valid executable on the filesystem, I would suggest just treating the truncation the same way the `realpath` failure is treated.

3405691582: Good point! Fixed.

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans:
@swift-ci test

spevans (approved):

PR #3004, opened 2021-07-09
Omnibus to implement CFRunLoop with kevent.

3405691582:
Dispatch uses kevent for its event loop backend. However, `CFRunLoop` in Foundation uses Mach ports. This means for other kevent-based platforms such as OpenBSD, new runloop functionality is required, which this omnibus pr implements. 

This pr must be considered in concert with apple/swift-corelibs-libdispatch#559 (see below).

Please feel free to request any or all to be split into multiple prs, though all these commits and the aforementioned Dispatch pr are required to properly implement `CFRunLoop` on this platform.

Notes:
* Some of the Mach-based implementation refers to `MACH_PORT_NULL`, for example, when we have symbols already defined like `CFPORT_NULL`. These are brought more into alignment for consistency.
* Each platform necessitates implementing the "backend" set of `__CFPort*` and `mk_timer_*` stubs. It does not make sense if these methods are not implemented for a given platform, so add a default `#error` case. (The code would greatly benefit from reorganizing the necessary stubs to implement together in one `#if` conditional, but this should probably be done separate from this omnibus as it is not critically important to achieve our result. However, we have implemented all the stubs together below in this commit.)
* A tentative kevent-based implementation of these stubs are provided. In apple/swift-corelibs-libdispatch#559, we introduced a runloop implementation using `pipe2` and packing each 32-bit fd in the pipe into a 64-bt integer. Since CFRunLoop fraternizes with Dispatch via `_dispatch_main_queue_callback_4CF`, this means that `dispatch_runloop_handle_t` in Dispatch must match `__CFPort` here. Unfortunately, this means that we have to perform some bit-packing shenanigans to handle kevent timers. There might be nicer ways to do this, but we'll defer that for now.
* We need to fix SR-14288 while we are here to ensure the runloop timers work correctly and the tests pass properly.
* While implementing the runloop, testing has uncovered some strange behavior with `pthread_main_np` and TSD on this platform. Work around this by doing a simple pthread_self comparison.

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -249,7 +249,11 @@ const char *_CFProcessPath(void) {
 
 #if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_BSD
 CF_CROSS_PLATFORM_EXPORT Boolean _CFIsMainThread(void) {
+#if defined(__OpenBSD__)
+    return pthread_equal(pthread_self(), _CFMainPThread) != 0;
+#else
     return pthread_main_np() == 1;
+#endif
 }
 #endif
 
@@ -774,7 +778,7 @@ static void __CFTSDFinalize(void *arg) {
 #if TARGET_OS_WASI
     __CFMainThreadHasExited = true;
 #else
-    if (pthread_main_np() == 1) {
+    if (_CFIsMainThread()) {
         // Important: we need to be sure that the only time we set this flag to true is when we actually can guarentee we ARE the main thread. 
         __CFMainThreadHasExited = true;
     }
--- a/CoreFoundation/RunLoop.subproj/CFRunLoop.c
+++ b/CoreFoundation/RunLoop.subproj/CFRunLoop.c
@@ -48,6 +48,15 @@ extern void objc_terminate(void);
 #define USE_DISPATCH_SOURCE_FOR_TIMERS 0
 #endif
 
+static inline uint64_t __CFNanosecondsToTSR(uint64_t ns) {
+#if TARGET_OS_MAC || TARGET_OS_LINUX
+    return ns;
+#else
+    CFTimeInterval ti = ns / 1.0E9;
+    return __CFTimeIntervalToTSR(ti);
+#endif
+}
+
 #if USE_DISPATCH_SOURCE_FOR_TIMERS
 #if !TARGET_OS_MAC
 typedef uint32_t mach_port_t;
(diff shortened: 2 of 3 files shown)


3405691582:
cc @millenomi -- input or advice on how to proceed?

AtariDreams:
@compnerd any thoughts?

buttaface:
@parkera, mind reviewing?

buttaface:
@phausler, maybe you could review?

buttaface:
@etcwilde, would you review?

CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -48,6 +48,15 @@ extern void objc_terminate(void);
 #define USE_DISPATCH_SOURCE_FOR_TIMERS 0
 #endif
 
+static inline uint64_t __CFNanosecondsToTSR(uint64_t ns) {
+#if TARGET_OS_MAC || TARGET_OS_LINUX
+    return ns;
+#else
```
etcwilde, line 54: @compnerd, is this the appropriate implementation of this function for Windows?

CoreFoundation/Base.subproj/CFPlatform.c
```diff
@@ -249,7 +249,11 @@ const char *_CFProcessPath(void) {
 
 #if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_BSD
 CF_CROSS_PLATFORM_EXPORT Boolean _CFIsMainThread(void) {
+#if !defined(__OpenBSD__)
```
etcwilde, line 252: It's a bit clearer if we have the OpenBSD case come first and not invert the condition.
```
#if defined(__OpenBSD__) // special case for OpenBSD
    return pthread_equal(pthread_self(), _CFMainPThread) != 0;
#else // everything else
    return pthread_main_np() == 1;
#endif
```
I usually think of `else` indicating anything not special-cased, so having the else only catch the OpenBSD case seems a bit odd.
Thoughts?

CoreFoundation/RunLoop.subproj/CFRunLoop.c
```diff
@@ -444,6 +460,269 @@ CF_INLINE kern_return_t __CFPortSetRemove(__CFPort port, __CFPortSet portSet) {
...
+    return KERN_SUCCESS;
+}
+
+static kern_return_t mk_timer_cancel(__CFPort timer, const void *unused) {
+    int id = __CFPORT_TIMER_UNPACK_ID(timer);
+    struct kevent tev;
+    EV_SET(
+        &tev,
+      id,
+      EVFILT_TIMER,
```
etcwilde, line 574: Inconsistent indentation through here.

etcwilde (commented):
This looks fine for the most part. I left a couple comments on style.

I'm unfortunately not familiar enough with this part of the codebase to really identify logic/behavior issues in the implementation of the `__CFPort*` or `mk_timer` function implementations down in `CFRunLoop.c`. It looks fine from what I can tell, but it would be better to get someone more familiar with this code to look over it and verify that it's correct.

As I'm not very familiar with this part of the codebase, nor am I an owner or maintainer of foundation, I don't really feel comfortable just approving it. I'll leave that part up to @parkera.

CoreFoundation/Base.subproj/CFPlatform.c
3405691582: Reordered the conditional.

CoreFoundation/RunLoop.subproj/CFRunLoop.c
3405691582: Fixed.

parkera (approved):
Looks good to me.

parkera:
@swift-ci test

compnerd:
I think that this might be causing the timeout on Windows ...

PR #3005, opened 2021-07-12
Fix temporary filename coda going above the maximum path length in some cases

millenomi:
This fixes an issue that would occur if the filename for a temporary file plus the `.tmp.<coda>` suffix was beyond the maximum path limit, but would be fine without it. This limits the temp filename to a maximum of 31 characters.

Fixes <https://bugs.swift.org/browse/SR-14823>.

- [x] Post 5.4 PR: https://github.com/apple/swift-corelibs-foundation/pull/3006
- [x] Post 5.5 PR: https://github.com/apple/swift-corelibs-foundation/pull/3007

### DIFF
--- a/Sources/Foundation/NSPathUtilities.swift
+++ b/Sources/Foundation/NSPathUtilities.swift
@@ -735,7 +735,6 @@ public func NSFullUserName() -> String {
 }
 
 internal func _NSCreateTemporaryFile(_ filePath: String) throws -> (Int32, String) {
-    let template = filePath + ".tmp.XXXXXX"
 #if os(Windows)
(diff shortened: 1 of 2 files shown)


millenomi:
cc @compnerd is Windows atomic writing broken? The Windows path above this seems to be using the temporary directory always, which may mean that .atomic writes on a volume other than `C:` (or wherever the temp directory ends up being) may fail due to `rename()` failing.

millenomi:
Never mind — it uses `CopyFileW`. It looks like the `.atomic` path is thus not atomic at all >.<

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test macOS

PR #3006, opened 2021-07-12
[5.4] Fix temporary filename coda going above the maximum path length in some cases

millenomi:
This fixes an issue that would occur if the filename for a temporary file plus the .tmp.<coda> suffix was beyond the maximum path limit, but would be fine without it. This limits the temp filename to a maximum of 31 characters.

Fixes <https://bugs.swift.org/browse/SR-14823>. This is a candidate PR cherry-pick of <https://github.com/apple/swift-corelibs-foundation/pull/3005>


millenomi:
cc @tomerd @drexin — nominated for 5.4.x.

millenomi:
@swift-ci please test

drexin (approved):
Thanks!

tomerd (approved):

PR #3008, opened 2021-07-14
SR-14108: Implement `copy()` in `DateFormatter`.

YOCKOW:
`DateFormatter.copy()` returns the same instance not a copied one so far.
This PR adds an implementation of `copy()` in `DateFormatter` to behave expectedly(i.e. The same behavior with DarwinFoundation).

Resolves [SR-14108](https://bugs.swift.org/browse/SR-14108).

PR #3009, opened 2021-07-16
SR-14933: Implement `copy()` in `ISO8601DateFormatter`.

YOCKOW:
As with #3008, `ISO8601DateFormatter` should also implement its own override of `copy()`.

Resolves SR-14933.

### DIFF
--- a/Sources/Foundation/ISO8601DateFormatter.swift
+++ b/Sources/Foundation/ISO8601DateFormatter.swift
@@ -112,6 +112,13 @@ open class ISO8601DateFormatter : Formatter, NSSecureCoding {
             aCoder.encode(timeZone._nsObject, forKey: "NS.timeZone")
         }
     }
+
+    open override func copy(with zone: NSZone? = nil) -> Any {
+        let copied = ISO8601DateFormatter()
+        copied.timeZone = timeZone
+        copied.formatOptions = formatOptions
+        return copied
+    }
     
     public static var supportsSecureCoding: Bool { return true }
     
--- a/Tests/Foundation/Tests/TestISO8601DateFormatter.swift
+++ b/Tests/Foundation/Tests/TestISO8601DateFormatter.swift
@@ -326,6 +326,28 @@ class TestISO8601DateFormatter: XCTestCase {
             try fixture.assertLoadedValuesMatch(areEqual(_:_:))
         }
     }
+
+    func test_copy() throws {
+        let original = ISO8601DateFormatter()
+        original.timeZone = try XCTUnwrap(TimeZone(identifier: "GMT"))
+        original.formatOptions = [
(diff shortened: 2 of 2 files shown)


spevans:
@swift-ci test

spevans:
So interestingly I ran the `test_copy` on BigSur 11.4.4 using `DarwinCompatibilityTests` and it fails, which leads me to believe this is actually a bug in Darwin as well. 

```
$ swift
Welcome to Apple Swift version 5.5 (swiftlang-1300.0.20.104 clang-1300.0.21.1).
Type :help for assistance.
  1> import Foundation 
  2.  
  3. let f1 = ISO8601DateFormatter() 
  4. let f2 = f1.copy() as! ISO8601DateFormatter 
  5. print(f1) 
  6. print(f2) 
  7. print(f1 === f2)
<NSISO8601DateFormatter: 0x10040dbd0>
<NSISO8601DateFormatter: 0x10040dbd0>
true
``` 

@millenomi Is `ISO8601DateFormatter` implementing `NSCopying` correctly on Darwin?

@YOCKOW I think this PR is correct I just want to see Lily's response first. I thinks its also worth adding a `XCTAssertFalse(copied === original)` into your tests for these `copy()` fixing PRs.

YOCKOW:
@spevans 
I'm sorry, but I didn't know `ISO8601DateFormatter` on Darwin has such behavior.
I'm going to modify the test.

spevans:
@YOCKOW I don't think your PR is wrong, it might be that there is a bug in Darwin's Foundation. I just wanted confirmation that the bug exists on macOS as well. So I would hold off on making any changes for now.

YOCKOW:
@spevans
Noted with thanks. I've just modified the test in this PR on local, but I won't push it as of now.

YOCKOW:
[`Date.ISO8601FormatStyle`](https://developer.apple.com/documentation/foundation/date/iso8601formatstyle) is coming, but I believe this PR is not stale yet.

spevans:
@millenomi Could you check my earlier comment in https://github.com/apple/swift-corelibs-foundation/pull/3009#issuecomment-881934600 it looks like Darwin's Foundation doesn't properly implement `ISO8601DateFormatter.copy`. I just want to confirm that Darwin has a bug

millenomi:
It is indeed a bug. Do you happen to ĥ ave a FB number handy for it?

spevans:
The sync system seemed to link https://bugs.swift.org/browse/SR-14108 to rdar://problem/73742491, not sure if that is the best bug to cover it, otherwise could link to https://bugs.swift.org/browse/SR-14933 which covers this fix.

YOCKOW:
Should I file a separate report about that using an other bug tracker such as Feedback Assistant?

PR #3013, opened 2021-07-23
Fix memory leak in PropertyListSerialization

drexin:
rdar://80919378

PR #3020, opened 2021-07-27
Revert "Fix memory leak in PropertyListSerialization"

drexin:
Reverts apple/swift-corelibs-foundation#3013

Reverting while we investigate a segfault that may have been caused by this change.


drexin:
@swift-ci test

PR #3021, opened 2021-07-27
asprintf exists on linux

soloturn:
asprintf exists on linux for quite a while, 2015 at least. do use it.

fixes https://bugs.swift.org/browse/SR-14935

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -1472,7 +1472,7 @@ void _CF_dispatch_once(dispatch_once_t *predicate, void (^block)(void)) {
 #pragma mark -
 #pragma mark Windows and Linux Helpers
 
-#if TARGET_OS_WIN32 || TARGET_OS_LINUX
+#if TARGET_OS_WIN32
 
 #include <stdio.h>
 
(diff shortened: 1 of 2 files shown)


spevans:
@swift-ci test

compnerd (approved):
It would be good to verify that `_GNU_SOURCE` is being defined.  I suspect that this might come back to hurt us a bit as not all libc implementation may provide this extension.  Perhaps a comment at these sites makes sense to indicate that we are relying on the fact that we only support glibc (and bionic) and that they provide `asprintf` (at least under some flags).

drodriguez:
We do for Android in https://github.com/apple/swift/blob/0053526c5d15532ef38fcf72006a9e16840b92a0/lib/ClangImporter/ClangImporter.cpp#L682-L688 . There's a comment for why it is not enabled in Linux, so if you need it for every Linux, maybe it needs to be done manually before any `#include` that end up in glibc.

etcwilde (approved):
I'm okay with merging this. We should probably add something like `|| (TARGET_OS_LINUX && !defined(_GNU_SOURCE))` just to completely cover the bases, but that can come in a subsequent patch.

etcwilde:
Thanks for the contribution @soloturn!

PR #3022, opened 2021-07-31
SR-14953: Data.write(to:) fails when writing to /dev/stdout

spevans:
- FileHandle.synchronize() would throw when calling fsync() on special
  files. Ignore EINVAL and EROFS if fsync() fails, so that the behaviour
  matches Darwin Foundation.


spevans:
@swift-ci test

PR #3028, opened 2021-08-08
Add `#ifdef` conditions to `CFPlatform.c` for WASI

MaxDesiatov:
WASI doesn't support multi-threading and process info or environment variables. Here we `#ifdef` relevant code out, or replace with stubs that allow Foundation to be compiled for WASI.


MaxDesiatov:
@swift-ci please test

spevans (approved):

PR #3032, opened 2021-08-09
Add `CMakeLists.txt` to `Sources/BlocksRuntime`

MaxDesiatov:
Sources in this directory are unused on most platforms, since Blocks are already included in libdispatch. Platforms, such as WASI, that don't have multi-threading support need to build BlocksRuntime bundled with Foundation.


MaxDesiatov:
@swift-ci please test

PR #3033, opened 2021-08-10
Use International Mile for UnitLength operations

nVitius:
I made this PR based on some conversation in [this](https://stackoverflow.com/questions/49373300) SO post. This patch changes the Coefficient for the mile to `1609.344`, to be in-line with the official definition of the "international mile".

From the NIST's _[Conversion Factor's for General Use](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication1038.pdf)_ handbook:
image

The existing Coefficient leads to an incorrect result. This can be seen by trying to convert 1 mile to feet (1 mile should be 5280 feet):
`1 * 1609.34 / 0.3048 = 5279.9868766404199475065616797900262467191601049868766404199475065...`
Using the standard `1609.344` instead gives `5280`.

The Coefficients for inches, feet, and yards are all correct.


amomchilov (approved):

spevans:
@swift-ci test

nVitius:
@spevans @amomchilov 
I believe I fixed the failing test cases. 

I'm not 100% sure about the TestFoundation.TestMeasurement-testLoadedValuesMatch test. 


It uses the `.archive` fixtures. As far as I can tell, those were created with NSKeyedArchiver. I'm not sure how to replicate that, but I think it's just a binary plist. I converted to XML and back to binary after updating the incorrect property (the coefficient for mile). 


Will you please re-run the tests?

spevans:
@swift-ci test

nVitius:
Looks like the changes allowed the tests to pass.


@amomchilov @spevans
Not sure what the protocol is for merging PRs. Just pinging you two to make sure you've seen this. Let me know if you need anything else from me. Thanks!

millenomi:
@nVitius does your patch correspond to Darwin behavior? We try to keep things working 1:1 bug-for-bug where possible.

nVitius:
@millenomi Looking through the docs, it would seem like Darwin also has this issue:
https://developer.apple.com/documentation/foundation/unitlength?changes=latest_minor
image

I sent a report via FeedbackAssistant for now; I'm not sure where the best place to report this would be.

amomchilov:
> We try to keep things working 1:1 bug-for-bug where possible.

Oh jeeze. That's unfortunate, but understandable.

I've had better success piping my bug reports to `/dev/null` than sending them into feedback assistant 😅

millenomi:
*sighs*

Can you post your FB number so I can pass it on internally?

amomchilov:
Pinging @nVitius, he's the one who submitted a bug report.

nVitius:
The FB number is: `FB9542852`

millenomi:
Thank you. I'm going to close this for now, but it was important and I'll make sure the issue is seen so we can ultimately fix it here as well.

Smail:
Has there been any (internal) progress / status update on this problem? The value is still wrong, and I found that the conversion for pounds is also incorrect, as well as a couple of other values (gravity, ounces, squareFeet, squareYards).

nVitius:
My ticket on Feedback Assistant remains open and has not received any updates since it was originally filed in 2021

amomchilov:
@nVitius That sounds about right. At least `/dev/null` doesn't give you false hope 😅 

I suggest working around this with your own custom units

parkera:
It looks like the Darwin version was fixed, but not swift-corelibs-foundation. All we need is a PR we can apply to main.

Smail:
> It looks like the Darwin version was fixed, but not swift-corelibs-foundation. All we need is a PR we can apply to main.

The Darwin version is closed source right? If so, can you check if they by any chance have fixed the other coefficients as well? 

For example pounds to kilograms should be `0.45359237`, but here in the core libs it is still `0.453592`.

The gravity constant should also probably be `9.80665` and not `9.81`.

parkera:
Pounds to kilograms is still 0.453592 and gravity constant is 9.81. If you open a PR here to fix them for swift-corelibs-foundation I'll take care of getting the internal sources for those fixed. We haven't rewritten them in Swift yet.

PR #2996, opened 2021-04-27
Fix autolink mechanism for static libraries on Linux

kateinoigakukun:
Resolve: [[SR-14536] Importing FoundationNetworking with -static-stdlib is broken with missing symbols on Linux - Swift](https://bugs.swift.org/browse/SR-14536)

Discussion: [Autolinking behavior of @_implementationOnly with static linking - Development / Core Libraries - Swift Forums](https://forums.swift.org/t/autolinking-behavior-of-implementationonly-with-static-linking/44393)

~~Blocked by [[Frontend] Add -public-autolink-library option by kateinoigakukun · Pull Request #35936 · apple/swift](https://github.com/apple/swift/pull/35936)~~

## Overview

`Foundation` imports `CoreFoundation` with `@_implementationOnly` as a private dependency, and it doesn't emit `IMPORTED_MODULE` in swiftmodule.
Users of Foundation can link `libCoreFoundation.so` automatically when using dynamic linking because it's already linked into `libFoundation.so`.
However, this automatic linking doesn't work when static linking, so users need to link CoreFoundation and dependent libraries of CoreFoundation explicitly by adding `-lCoreFoundation -lBlocksRuntime -licui18n`.

To avoid forcing users to link private dependencies explicitly, this patch changed to merge the dependencies into libFoundation.a. And also to avoid forcing to link dependent libs of CoreFoundation like `curl` or `icui18n`, add `LINK_LIBRARY` entry in Foundation.swiftmodule using `-public-autolink-library` which is introduced by https://github.com/apple/swift/pull/35936

PR #3035, opened 2021-08-11
Revert "Fix autolink mechanism for static libraries on Linux"

drexin:
Reverts apple/swift-corelibs-foundation#2996

The icui18n library that is linked against is not necessarily the same as the one used to build Foundation and Swift, so this is causing builds to fail.

### DIFF
--- a/Sources/Foundation/CMakeLists.txt
+++ b/Sources/Foundation/CMakeLists.txt
@@ -160,18 +160,6 @@ set_target_properties(Foundation PROPERTIES
   Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_BINARY_DIR}/swift)
 
-if(NOT BUILD_SHARED_LIBS)
-  target_compile_options(Foundation
-    PRIVATE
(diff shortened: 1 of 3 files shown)


drexin:
@swift-ci test

millenomi:

kateinoigakukun:
@drexin Could you let me know a sample failed CI job for investigation?

drexin:
@kateinoigakukun https://ci.swift.org/job/oss-swift-package-centos-7/1112

```
/opt/rh/devtoolset-7/root/usr/lib/gcc/x86_64-redhat-linux/7/../../../../bin/ld.gold: error: cannot find -licui18n
clang-5.0: error: linker command failed with exit code 1 (use -v to see invocation)
<unknown>:0: error: link command failed with exit code 1 (use -v to see invocation)
```

drexin:
If we are building ICU from source, it should link against those libraries, which may have a different name (e.g. icui18nswift in case of the default Linux build). So the correct solution I think would be to determine to actual name of the library that cmake detects and use that here as well.

kateinoigakukun:
Thank you for sharing :) Oh, I forgot `swift` suffix variants. I'll work on this in a few days

PR #3036, opened 2021-08-11
Add AsyncBytes, AsyncLineSequence, AsyncCharacterSequence, and AsyncUnicodeScalarSequence

Catfish-Man:
Only the local-file versions of AsyncBytes for now

### DIFF
--- a/Sources/Foundation/AsyncLineSequence.swift
+++ b/Sources/Foundation/AsyncLineSequence.swift
@@ -1,0 +1,166 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+public struct AsyncLineSequence<Base: AsyncSequence>: AsyncSequence where Base.Element == UInt8 {
+    public typealias Element = String
+    
+    var base: Base
+    
+    @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+    public struct AsyncIterator: AsyncIteratorProtocol {
+        public typealias Element = String
+        
+        var byteSource: Base.AsyncIterator
+        var buffer: Array<UInt8> = []
+        var leftover: UInt8? = nil
+        
+        internal init(underlyingIterator: Base.AsyncIterator) {
+            byteSource = underlyingIterator
+        }
+        
+        // We'd like to reserve flexibility to improve the implementation of
+        // next() in the future, so aren't marking it @inlinable. Manually
+        // specializing for the common source types helps us get back some of
+        // the performance we're leaving on the table.
+        @_specialize(where Base == URL.AsyncBytes)
+        @_specialize(where Base == FileHandle.AsyncBytes)
+        //@_specialize(where Base == URLSession.AsyncBytes)
+        public mutating func next() async rethrows -> String? {
+            /*
+             0D 0A: CR-LF
+             0A | 0B | 0C | 0D: LF, VT, FF, CR
+             E2 80 A8:  U+2028 (LINE SEPARATOR)
+             E2 80 A9:  U+2029 (PARAGRAPH SEPARATOR)
+             */
+            let _CR: UInt8 = 0x0D
+            let _LF: UInt8 = 0x0A
+            let _NEL_PREFIX: UInt8 = 0xC2
+            let _NEL_SUFFIX: UInt8 = 0x85
+            let _SEPARATOR_PREFIX: UInt8 = 0xE2
+            let _SEPARATOR_CONTINUATION: UInt8 = 0x80
+            let _SEPARATOR_SUFFIX_LINE: UInt8 = 0xA8
+            let _SEPARATOR_SUFFIX_PARAGRAPH: UInt8 = 0xA9
+          
+            func yield() -> String? {
+                defer {
+                    buffer.removeAll(keepingCapacity: true)
+                }
+                if buffer.isEmpty {
+                    return nil
+                }
+                return String(decoding: buffer, as: UTF8.self)
+            }
+            
+            func nextByte() async throws -> UInt8? {
+                defer { leftover = nil }
+                if let leftover = leftover {
+                    return leftover
+                }
+                return try await byteSource.next()
+            }
+            
+            while let first = try await nextByte() {
+                switch first {
+                case _CR:
+                    let result = yield()
+                    // Swallow up any subsequent LF
+                    guard let next = try await byteSource.next() else {
+                        return result //if we ran out of bytes, the last byte was a CR
+                    }
+                    if next != _LF {
+                        leftover = next
+                    }
+                    if let result = result {
+                        return result
+                    }
+                    continue
+                case _LF..<_CR:
+                    guard let result = yield() else {
+                        continue
+                    }
+                    return result
+                case _NEL_PREFIX: // this may be used to compose other UTF8 characters
+                    guard let next = try await byteSource.next() else {
+                        // technically invalid UTF8 but it should be repaired to "\u{FFFD}"
+                        buffer.append(first)
+                        return yield()
+                    }
+                    if next != _NEL_SUFFIX {
+                        buffer.append(first)
+                        buffer.append(next)
+                    } else {
+                      guard let result = yield() else {
+                          continue
+                      }
+                      return result
+                    }
+                case _SEPARATOR_PREFIX:
+                    // Try to read: 80 [A8 | A9].
+                    // If we can't, then we put the byte in the buffer for error correction
+                    guard let next = try await byteSource.next() else {
+                        buffer.append(first)
+                        return yield()
+                    }
+                    guard next == _SEPARATOR_CONTINUATION else {
+                        buffer.append(first)
+                        buffer.append(next)
+                        continue
+                    }
+                    guard let fin = try await byteSource.next() else {
+                        buffer.append(first)
+                        buffer.append(next)
+                        return yield()
+                        
+                    }
+                    guard fin == _SEPARATOR_SUFFIX_LINE || fin == _SEPARATOR_SUFFIX_PARAGRAPH else {
+                        buffer.append(first)
+                        buffer.append(next)
+                        buffer.append(fin)
+                        continue
+                    }
+                    if let result = yield() {
+                        return result
+                    }
+                    continue
+                default:
+                    buffer.append(first)
+                }
+            }
+            // Don't emit an empty newline when there is no more content (e.g. end of file)
+            if !buffer.isEmpty {
+                return yield()
+            }
+            return nil
+        }
+        
+    }
+    
+    public func makeAsyncIterator() -> AsyncIterator {
+        return AsyncIterator(underlyingIterator: base.makeAsyncIterator())
+    }
+    
+    internal init(underlyingSequence: Base) {
+        base = underlyingSequence
+    }
+}
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+public extension AsyncSequence where Self.Element == UInt8 {
+    /**
+     A non-blocking sequence of newline-separated `Strings` created by decoding the elements of `self` as UTF8.
+     */
+    @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+    var lines: AsyncLineSequence<Self> {
+        AsyncLineSequence(underlyingSequence: self)
+    }
+}
--- a/Sources/Foundation/FileHandle+Async.swift
+++ b/Sources/Foundation/FileHandle+Async.swift
@@ -1,0 +1,189 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Darwin
+
+final actor IOActor {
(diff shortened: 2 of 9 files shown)


Sources/Foundation/URL+AsyncBytes.swift
```diff
@@ -0,0 +1,60 @@
...
+                return try await buffer.next()
+            }
+            
+            internal init(_ url: URL) {
+                buffer.readFunction = { (buf: inout _AsyncBytesBuffer) -> Int in
+                    if url.isFileURL {
+                        let fh = try await IOActor.default.createFileHandle(reading: url).bytes.makeAsyncIterator()
+                        buf = fh.buffer
+                    } else {
+                        // TODO: add networking support
```
millenomi, line 34: For future reference: use `_NSNonfileURLContentLoader` in SCF to load using URLSession if available. Right now it only returns an in-memory `NSData` after all is done, but it may be sufficient here.

Sources/Foundation/FileHandle+Async.swift
```diff
@@ -0,0 +1,189 @@
...
+    func read(from fd: Int32, into buffer: UnsafeMutableRawBufferPointer) async throws -> Int {
+        while true {
+            let amount = Darwin.read(fd, buffer.baseAddress, buffer.count)
+            if amount >= 0 {
+                return amount
+            }
+            let posixErrno = errno
+            if errno != EINTR {
+                // TODO: get the path of the fd to provide a more informative error
+                throw NSError(domain: NSPOSIXErrorDomain, code: Int(posixErrno), userInfo: [:])
```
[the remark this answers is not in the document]
Catfish-Man: Darwin has an fcntl(…, F_GETPATH) shim written in C here. It didn't seem terrible important to figure out right now.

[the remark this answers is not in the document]
millenomi: I'd still want to use the helper function here to return `CocoaError`s rather than POSIX domain errors unless there's a compelling argument.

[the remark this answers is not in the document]
millenomi: (If Darwin also returns POSIX domain errors, this is okay as-is.)

millenomi:
@swift-ci please test

Sources/Foundation/FileHandle+Async.swift
```diff
@@ -0,0 +1,189 @@
...
+        
+        public func makeAsyncIterator() -> Iterator {
+            return Iterator(file: handle)
+        }
+        
+        @frozen
+        public struct Iterator: AsyncIteratorProtocol {
+            
+            @inline(__always) static var bufferSize: Int {
+                16384
```
Catfish-Man, line 150: It occurs to me this should probably be 16384 minus the size of the object header

Sources/Foundation/FileHandle+Async.swift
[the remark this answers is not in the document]
Catfish-Man: It does

Sources/Foundation/FileHandle+Async.swift
Catfish-Man: Or maybe not so the loop can be aligned. Ugh, I'll need to dig into this at some point.

Sources/Foundation/FileHandle+Async.swift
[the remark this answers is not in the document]
millenomi: I'm good as is on this one then.

Sources/Foundation/FileHandle+Async.swift
```diff
@@ -0,0 +1,189 @@
...
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Darwin
```
millenomi, line 13:
```suggestion
#if canImport(Darwin)
import Darwin
#elseif canImport(Glibc)
import Glibc
#elseif canImport(CRT)
import CRT
#endif
```

This is failing on CI; it needs to pull a platform-appropriate C sodlib.

Sources/Foundation/FileHandle+Async.swift
```diff
@@ -0,0 +1,189 @@
...
+
+final actor IOActor {
+    func read(from fd: Int32, into buffer: UnsafeMutableRawBufferPointer) async throws -> Int {
+        while true {
+            let amount = Darwin.read(fd, buffer.baseAddress, buffer.count)
```
millenomi, line 18: This may need to have an alias; for example, from FileHandle.swift:

```
#if canImport(Darwin)
import Darwin
fileprivate let _read = Darwin.read(_:_:_:)
fileprivate let _write = Darwin.write(_:_:_:)
fileprivate let _close = Darwin.close(_:)
#elseif canImport(Glibc)
import Glibc
fileprivate let _read = Glibc.read(_:_:_:)
fileprivate let _write = Glibc.write(_:_:_:)
fileprivate let _close = Glibc.close(_:)
#endif
```

millenomi: It's probably something like this.

```suggestion
#if canImport(Darwin)
            let read = Darwin.read
#elseif canImport(Glibc)
            let read = Glibc.read
#elseif canImport(CRT)
            let read = CRT._read
#endif
            let amount = Darwin.read(fd, buffer.baseAddress, buffer.count)
```

millenomi: cc @compnerd The docs for CRT say `_read`, is that how it is imported?

compnerd: It would be nice to use `_read` to avoid C4996 (i.e. `-Wdeprecated-declaration`).  The import that you have seems correct.  I can try to kick off a test run of this once this is in a good state.

compnerd:
FYI: this also breaks on Windows: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=53483&view=logs&j=ec47c80d-c7cf-5f2b-7138-13cd268bea3d&t=7a188035-f39c-599a-5e00-254b4ae08c42

millenomi:
@swift-ci please test

millenomi:
I missed the change.

millenomi:
@swift-ci please test

millenomi:
Ah, never mind, this has merge conflicts.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
Waiting for Linux build to pass before correcting the availability annotation for macOS.

Tests/Foundation/Tests/TestFileHandle+Async.swift
```diff
@@ -0,0 +1,426 @@
...
+            ("testNELPrefixButNoSuffix", test_nel_prefix_but_no_suffix),
+            ("testNELPrefixEnd", test_nel_prefix_end),
+            ("testNELPrefixAndSuffix", test_nel_prefix_and_suffix),
+            ("testSeparatorPrefixAndContinuationEnd", test_seperator_prefix_and_continuation_end),
+            ("testSeparatorPrefixButNoSuffix", test_seperator_prefix_but_no_suffix),
+            ("testSeparatorPrefixAndContinuationEnd", test_seperator_prefix_and_continuation_end),
+            ("testSeparatorPrefixAndContinuationWithoutFin", test_seperator_prefix_and_continuation_without_fin)
+        ]
+        
+        return tests
```
millenomi, line 424: @Catfish-Man Is this what you meant here? This method was missing its termination.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
A little worried about this patch as we haven't even gotten to making tests run yet. @Catfish-Man do you have some time to shepherd it actively?

millenomi:
(Hopefully this one is the run that gets to unit tests.)

millenomi:
The tests aren't running because they weren't listed in the main.

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

millenomi:
@swift-ci please test

MaxDesiatov:
@swift-ci please test

Gerzer:
I understand that this pull request is holding up the implementation the new concurrency APIs in `URL` and friends. Is there any update on when it might be merged and when work might begin on those new APIs? Is there any need for additional help with these implementations?

Tests/Foundation/main.swift
```diff
@@ -124,6 +124,7 @@ var allTestCases = [
     testCase(TestUnitVolume.allTests),
     testCase(TestNSLock.allTests),
     testCase(TestNSSortDescriptor.allTests),
+    testCase(TestFileHandleAsync.allTests),
```
YOCKOW, line 127: (Let me leave a comment before the test logs would gone.)
Build failure here:
`error: invalid conversion from 'async' function of type '() async throws -> ()' to synchronous function type '() throws -> Void'`

YOCKOW, line 127: (Let me leave a comment before the test logs would gone.)
Build failure here:
`error: invalid conversion from 'async' function of type '() async throws -> ()' to synchronous function type '() throws -> Void'`

waterskier2007:
@YOCKOW so it looks like there just needs to be a utility function written to wrap an async test case since the currently implementation only takes non-async functions. Is that correct?

YOCKOW:
> Current implementation here https://github.com/apple/swift-corelibs-foundation/blob/a-sink-eye-oh/Tests/Foundation/Utilities.swift#L14

I think, unfortunately, no. It is used only when `DARWIN_COMPATIBILITY_TESTS` is defined.
A function `testCase` is implemented in XCTest, which is open source on Linux:
https://github.com/apple/swift-corelibs-xctest/blob/56501e765303ccac2f104a6e45cfcd6e733cc74e/Sources/XCTest/Public/XCTestCase.swift#L302

waterskier2007:
Alright, so then we need to PR an async version of testCase in XCTest and utilize that in this current PR?

YOCKOW:
> Alright, so then we need to PR an async version of `testCase` in XCTest and utilize that in this current PR?

Or, we can use [`asyncTest`](https://github.com/apple/swift-corelibs-xctest/blob/56501e765303ccac2f104a6e45cfcd6e733cc74e/Sources/XCTest/Public/XCTestCase.swift#L325).

MPLew-is:
Is there anything that a novice community member like me can do to help land this? We're building Lambdas on AWS in Swift and would love to be able to get full AsyncBytes support.

iCharlesHu:
@swift-ci please test

parkera:
> Is there anything that a novice community member like me can do to help land this? We're building Lambdas on AWS in Swift and would love to be able to get full AsyncBytes support.

The Swift Async Algorithms package has an updated version of this as [AsyncBufferedByteIterator](https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift).

MPLew-is:
@parkera thanks for the link, that should help at least bridge the gap a little - however, it would still be incredibly nice to be able to land this `Foundation` version, especially for its integration with `FileHandle`. Right now, doing cross-platform async file I/O is painful since Linux is behind the Mac version.

Sources/Foundation/FileHandle+Async.swift
```diff
@@ -0,0 +1,207 @@
...
+    }
+    
+    @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+    func read(from handle: FileHandle, into buffer: UnsafeMutableRawBufferPointer) async throws -> Int {
+        // this is not incredibly effecient but it is the best we have
+        guard let data = try handle.read(upToCount: buffer.count) else {
+            return 0
+        }
+        data.copyBytes(to: buffer)
+        return data.count
```
Zollerboy1, line 52: Is there a specific reason, `handle.readabilityHandler` isn't used here in conjunction with `withUnsafeThrowingContinuation`? Since this is still not part of Foundation on Linux, I wanted to include this code in my package for the moment and just wondered if it was better to use `readabilityHandler` because this function is blocking otherwise.

Catfish-Man: readabilityHandler is less useful than you'd hope for regular files. They're just always readable.

Sources/Foundation/AsyncLineSequence.swift
```diff
@@ -0,0 +1,166 @@
...
+                    if next != _LF {
+                        leftover = next
+                    }
+                    if let result = result {
+                        return result
+                    }
+                    continue
+                case _LF..<_CR:
+                    guard let result = yield() else {
+                        continue
```
macserv, line 89: Forgive me if I'm oversimplifying... there's a lot going on here. 😊  It looks like this completely removes empty lines, which seems suboptimal, given that...

1. blank lines are meaningful in countless data applications, and
2. it's easy enough to add your own `isEmpty` filter on the sequence if you really do want to strip them out.

Based on this, the default behavior should probably be to *include* empty lines (i.e., change `continue` to `return ""` in the `_CR` and `_LF` cases).

mickeyl: If in doubt, please cross-check with Foundation's behavior. It needs to have the same semantics, else we widen the (already large) gap between APPLE and !APPLE.

macserv: Ah, yeah, good point.  This does match Foundation's behavior, so the issue would be against Apple, not this code.  Thanks for the check!

Sources/Foundation/FileHandle+Async.swift
Higher08: But a FileHandle can also be the output of Process, and it may need to wait for new bytes. So the thread will be locked with `read(upToCount:)`. I think `readabilityHandler` will be a better option that we can wrap into _continuation_

sepbehroozi:
Hey folks,
Any reason why this PR was left alone after 2022? I ran into this PR when dealing with the problem of the `AsyncByte` not being available for linux :( 
Any chance someone might consider working on this PR? If no one's actively on it, I'm volunteering :)
CC: @millenomi @Catfish-Man @YOCKOW

Catfish-Man:
AsyncBytes has a number of limitations that make it an attractive nuisance in server situations. We may be able to fix them by using my IORing changes in the future. The other types it's mostly just due to people being shuffled around organizationally.

mickeyl:
Still… don't you think it's a worthwhile goal to get the API differences as low as possible? After all, classic Foundation is still underpinning a lot of tools that are supposed to run on Apple and !Apple platforms.

PR #3037, opened 2021-08-12
Disable file/process-related code in CF for WASI

MaxDesiatov:
Filesystem (and thus `CFBundle` and `CFPreferences`), sockets,  and processes are unavailable on WASI. We should disable this code if ifdefs.


MaxDesiatov:
@swift-ci please test

PR #3038, opened 2021-08-12
WASI: Make some sources unavailable in CF CMake

MaxDesiatov:
`CFBundle`, plugins, preferences, streams, and run loops are not available when compiling to WASI/Wasm because of the lack of a meaningful filesystem access and multi-threading. These sources files should be excluded from the build in `CoreFoundation/CMakeLists.txt` when compiling for that platform.

### DIFF
--- a/CoreFoundation/CMakeLists.txt
+++ b/CoreFoundation/CMakeLists.txt
@@ -20,17 +20,23 @@ endif()
 
 set(CMAKE_POSITION_INDEPENDENT_CODE YES)
 
-set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci please test

millenomi:
There are a bunch of services tied to e.g. `Bundle.main`; do they need to be excluded from the build?

MaxDesiatov:
I have a stub implementation of `Bundle.main` in the fork that doesn't rely on CoreFoundation, that will come in a separate PR.

millenomi:
@MaxDesiatov It looks like this broke the Windows build: https://ci-external.swift.org/job/oss-swift-windows-toolchain-x86_64-vs2019/130/consoleText

PR #3039, opened 2021-08-12
Add support for CF data structures/utils on WASI

MaxDesiatov:
A lot of data structures, like strings, characters, data, locales, plists just need a proper check for `|| TARGET_OS_WASI` to be enabled on that platform and to use a suitable implementation. Some code related to filesystem access had to be disabled on WASI though.


MaxDesiatov:
@swift-ci please test

PR #3042, opened 2021-08-13
tests: explicitly flush `stderr` on exit

compnerd:
It has been observed in CI that the stderr output is sometimes truncated
on failure.  Explicitly flush the output to ensure that we can see the
error messages.  This has been observed on Windows which does not do a
courtesy flush unlike Linux.


millenomi:

millenomi:

PR #3043, opened 2021-08-13
CoreFoundation: add `US/Pacific` TimeZone mapping for Windows

compnerd:
Windows does not have the same timezone information available as the
other platforms.  We import a subset of the CLDR (due to duplicate
mappings) into Foundation to provide the translation.
`TestCalendar/test_nextDate` usees `US/Pacific` for a test case.
Explicitly add this to the Olson -> Windows mapping.  We cannot provide
a Windows -> Olson mapping entry for this value as that would have
multiple values for the canonical `America/Los_Angeles` key.  This
repairs the `TestCalendar/test_nextDate` test case on Windows.


millenomi:

PR #3044, opened 2021-08-13
tests: repair `TestNSData/test_writeToURLSpecialFile`

compnerd:
This test was added recently and regressed the test suite on Windows.
Windows does not have a devfs mounted at `/dev`.  Use the `CON` special
file which is roughly equivalent to `/dev/stdout` to emulate the desired
test behaviour.


compnerd:
@swift-ci please tet

compnerd:
@swift-ci please test

PR #3045, opened 2021-08-14
FileHandle: improve the compatibility of `synchronize`

compnerd:
TestFileHandle/test_synchronizeOnSpecialFile introduced a failure in the
Windows test suite.  The underlying problem here is that
`FlushFileBuffers` is not permitted on unbuffered file handles (such as
consoles).  Ensure that we do not treat errors on character devices as
flush errors.


millenomi:
@swift-ci test and merge

PR #3048, opened 2021-08-14
Tests: make test_mountedVoumeURLs more robust

compnerd:
Rather than creating a temporary directory, locate the application
support directory.  The temporary directory may be created under a
root that that is not an actual volume, but rather an aliased
mountpoint.  This is the case on the CI hosts where `T:` is a path
substitution rather than an actual disk volume.


compnerd:
@swift-ci please test

PR #3049, opened 2021-08-15
FileManager: correct API misuse on Windows

compnerd:
`GetVolumeInformationW` is documented as taking the root directory of
the volume to describe as `lpRootPathName`. We attempt to compute the
location by determining the volume mount point where the specified
path is mounted. However, `GetVolumePathNameW` does always return the
path to the volume mount point, which is the root of the volume where
the specified path.  With a path substitution that is not rooted at a
volume root, the volume root with a relative path. As we need to get
to the root, force the stripping to the root via `PathCchStripToRoot`.

This should repair the
TestFoundation.TestFileManager/test_filesystemAttributes on Windows.


compnerd:
@swift-ci please test

PR #3052, opened 2021-08-17
Revert "Revert "Fix autolink mechanism for static libraries on Linux""

kateinoigakukun:
Re-land https://github.com/apple/swift-corelibs-foundation/pull/2996

ICU i18n library name is now resolved based on CMake package finding system. It allows to use `swift` suffixed ICU libraries.


kateinoigakukun:
CC: @drexin

MaxDesiatov:
@swift-ci please test

PR #3055, opened 2021-08-19
Introduce AttributedString

iCharlesHu:
This patch introduces struct `AttributedString` and tests. Markdown support is not included in this patch as it will be introduced in a separate package.

See Apple's [documentation](https://developer.apple.com/documentation/foundation/attributedstring/) for more information.

### DIFF
--- a/Sources/Foundation/AttributedString/AttributedString+Locking.swift
+++ b/Sources/Foundation/AttributedString/AttributedString+Locking.swift
@@ -1,0 +1,53 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Darwin)
+import Darwin
+
+@available(macOS 10.12, iOS 10.0, tvOS 10.0, watchOS 3.0, *)
+extension UnsafeMutablePointer where Pointee == os_unfair_lock_s {
+    internal init() {
+        let l = UnsafeMutablePointer.allocate(capacity: 1)
+        l.initialize(to: os_unfair_lock())
+        self = l
+    }
+    
+    internal func cleanupLock() {
+        deinitialize(count: 1)
+        deallocate()
+    }
+    
+    internal func lock() {
+        os_unfair_lock_lock(self)
+    }
+    
+    internal func tryLock() -> Bool {
+        let result = os_unfair_lock_trylock(self)
+        return result
+    }
+    
+    internal func unlock() {
+        os_unfair_lock_unlock(self)
+    }
+}
+
+@available(macOS 10.12, iOS 10.0, tvOS 10.0, watchOS 3.0, *)
+typealias Lock = os_unfair_lock_t
+
+#else
+internal typealias Lock = NSLock
+
+extension NSLock {
+    // Stub to match the API of os_unfair_lock_t
+    func cleanupLock() { /* NOOP */ }
+}
+#endif
--- a/Sources/Foundation/Bridging.swift
+++ b/Sources/Foundation/Bridging.swift
@@ -230,25 +230,23 @@ internal final class __SwiftValue : NSObject, NSCopying {
 // MARK: AnyHashable Bridging
 extension AnyHashable : _ObjectiveCBridgeable {
     public func _bridgeToObjectiveC() -> NSObject {
-        // This is unprincipled, but pretty much any object we'll encounter in
-        // Swift is NSObject-conforming enough to have -hash and -isEqual:.
-        return unsafeBitCast(base as AnyObject, to: NSObject.self)
+        return __SwiftValue.store(base)
     }
-    
+
     public static func _forceBridgeFromObjectiveC(_ x: NSObject, result: inout AnyHashable?) {
-        result = AnyHashable(x)
+        result = (__SwiftValue.fetch(x) as? AnyHashable) ?? AnyHashable(x)
     }
-    
+
     public static func _conditionallyBridgeFromObjectiveC(_ x: NSObject, result: inout AnyHashable?) -> Bool {
         self._forceBridgeFromObjectiveC(x, result: &result)
         return result != nil
     }
-    
+
     @_effects(readonly)
     public static func _unconditionallyBridgeFromObjectiveC(_ source: NSObject?) -> AnyHashable {
         // `nil` has historically been used as a stand-in for an empty
         // string; map it to an empty string.
         if _slowPath(source == nil) { return AnyHashable(String()) }
-        return AnyHashable(source!)
+        return (__SwiftValue.fetch(source) as? AnyHashable) ?? AnyHashable(source!)
     }
 }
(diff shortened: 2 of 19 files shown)


iCharlesHu:
@swift-ci please test

Sources/Foundation/AttributedString/AttributedString+Locking.swift
```diff
@@ -0,0 +1,53 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Darwin)
+import Darwin
+
+@available(macOS 10.12, iOS 10.0, tvOS 10.0, watchOS 3.0, *)
+extension UnsafeMutablePointer where Pointee == os_unfair_lock_s {
```
millenomi, line 17: I thought we had a `Locking.swift` already with this kind of thing defined already, am I misremembering

Tests/Foundation/Tests/TestAttributedStringSupport.swift
```diff
@@ -0,0 +1,144 @@
...
+        static let name = "ForgroundColor"
+    }
+    
+    enum TestBackgroundColorAttribute: CodableAttributedStringKey, MarkdownDecodableAttributedStringKey {
+        typealias Value = Color
+        static let name = "BackgroundColor"
+    }
+
+    #if false
+    @frozen enum MisspelledAttribute : CodableAttributedStringKey, FixableAttribute {
```
millenomi, line 71: We should not have this if we're not compiling this, and Ortography isn't supported by swift-corelibs-foundation.

Tests/Foundation/Tests/TestAttributedStringSupport.swift
```diff
@@ -0,0 +1,144 @@
...
+    
+    struct TestAttributes : AttributeScope {
+        var testInt : TestIntAttribute
+        var testString : TestStringAttribute
+        var testDouble : TestDoubleAttribute
+        var testBool : TestBoolAttribute
+        var foregroundColor: TestForegroundColorAttribute
+        var backgroundColor: TestBackgroundColorAttribute
+    #if false
+        var misspelled : MisspelledAttribute
```
millenomi, line 134: Same as above.

millenomi:
cc @shahmishal @drexin @tomerd This is a large commit that is Delicate in a bunch of ways. I'm going to merge tomorrow after doing a deeper review, but keep an eye out for CI breakage in case we miss something.

Sources/Foundation/Morphology.swift
```diff
@@ -0,0 +1,349 @@
...
+        public var reflexiveForm: String?
+    }
+}
+
+// MARK: -
+// MARK: NSAttributedString Keys
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension NSAttributedString.Key {
+    internal static var morphology = NSAttributedString.Key(rawValue: "NSMorphology")
```
millenomi, line 345: These are all public in Monterey and should probably be declared public here as well since we _do_ have `NSAttributedString` as a usable public class.

Sources/Foundation/Morphology.swift
```diff
@@ -0,0 +1,349 @@
...
+    internal static var inflectionRule = NSAttributedString.Key(rawValue: "NSInflect")
+    internal static var inflectionAlternative = NSAttributedString.Key(rawValue: "NSInflectionAlternative")
+    internal static var language = NSAttributedString.Key(rawValue: "NSLanguage")
```
millenomi, line 348: Also, they should be `let's.

```suggestion
    public static let morphology = NSAttributedString.Key(rawValue: "NSMorphology")
    public static let inflectionRule = NSAttributedString.Key(rawValue: "NSInflect")
    public static let inflectionAlternative = NSAttributedString.Key(rawValue: "NSInflectionAlternative")
    public static let language = NSAttributedString.Key(rawValue: "NSLanguage")
```

Sources/Foundation/Bridging.swift
```diff
@@ -226,3 +226,29 @@ internal final class __SwiftValue : NSObject, NSCopying {
 
     override var description: String { String(describing: value) }
 }
+
+// MARK: AnyHashable Bridging
+extension AnyHashable : _ObjectiveCBridgeable {
+    public func _bridgeToObjectiveC() -> NSObject {
+        // This is unprincipled, but pretty much any object we'll encounter in
+        // Swift is NSObject-conforming enough to have -hash and -isEqual:.
+        return unsafeBitCast(base as AnyObject, to: NSObject.self)
```
millenomi, line 235: This cannot be right. Objects without the ObjC runtime are _not_ NSObjectProtocol-compliant.

millenomi: In particular, they are almost never NSObjects.

millenomi (changes requested):
See suggestion + issue with AnyHashable bridging above. We should perhaps check if the `AnyHashable` is a NSObject-conforming value, and wrap it in a `__SwiftValue` if it isn't.

Sources/Foundation/AttributedString/AttributedString+Locking.swift
millenomi: We do not! Huh! It's good to have it.

Sources/Foundation/Bridging.swift
millenomi, line 235: Possibly something like:

```suggestion
        if let object = __SwiftValue.fetch(base) as? NSObject {
            return object
        }
        
        return __SwiftValue.store(base)
```

millenomi: (`.store` is guaranteed to return NSObjects.)

iCharlesHu:
Addressed @millenomi 's comments.

iCharlesHu:
@swift-ci please test

millenomi:

millenomi (approved):

lxbndr:
Looks like `_loadScopeAttributes(forSymbol:, from:) -> [String : Any.Type]?` in `Conversion.swift` should be wrapped into `#if os(macOS)` as well. Windows build is failing:
```
D:\s\swift-corelibs-foundation\Sources\Foundation\AttributedString\Conversion.swift:295:24: error: cannot find 'dlopen' in scope
    guard let handle = dlopen(path, RTLD_NOLOAD) else {
                       ^~~~~~
D:\s\swift-corelibs-foundation\Sources\Foundation\AttributedString\Conversion.swift:295:37: error: cannot find 'RTLD_NOLOAD' in scope
    guard let handle = dlopen(path, RTLD_NOLOAD) else {
                                    ^~~~~~~~~~~
D:\s\swift-corelibs-foundation\Sources\Foundation\AttributedString\Conversion.swift:298:31: error: cannot find 'dlsym' in scope
    guard let symbolPointer = dlsym(handle, symbol) else {
                              ^~~~~
```
cc @iCharlesHu

xwu:
Hey @iCharlesHu @millenomi, I must be kind of confused, but is it really the case that this project can no longer be run and tested on macOS at all?

Use of `@_spi(Reflection)` in this file means that one has to use a dev toolchain to compile the project, because the default toolchain distributed with Xcode doesn't include any SPI symbols. However, due to SR-12177, one must simultaneously also _not_ use a dev toolchain but rather the default toolchain distributed with Xcode?

millenomi:
Merging this will require that you use a matching toolchain; the main (trunk) snapshot toolchain from https://swift.org/download should in fact work.

In general, even on macOS, you should always be working on this code with a matching toolchain, rather than use Xcode's default. On Linux, this is enforced as `build-script` will build both the toolchain and Foundation for you in one go.

millenomi:
… and I'll be looking at the SR this coming week because whoops.

I _have_ been able to run recently on macOS despite that SR; I'll investigate.

PR #3057, opened 2021-08-23
Fix typo in SingleValueContainer in JSONDecoder.swift

danielbendix:
It is named `SingleValueContainter` instead of `SingleValueContainer`


MaxDesiatov:
@swift-ci please test

PR #3059, opened 2021-08-24
Units: Add UnitInformationStorage

corymosiman12:
# Summary
Adds support for UnitInformationStorage, which seems to be the only Unit to not be implemented per: https://gist.github.com/ianpartridge/9e010553594b9bbf45fa8628af7fb562

## Additional Information
The docs on the [developer page](https://developer.apple.com/documentation/foundation/unitinformationstorage) also need to be corrected. The usage of `e` should really be an exponent (unless lowercase `e` is intended to mean exponent, but that's not very clear). For example:

> Decimal Bits | Coefficient | Binary Bits | Coefficient
> -- | -- | -- | --
> ...
> megabits | 1000e2 | mebibits | 1024e2

should be corrected to

Decimal Bits | Coefficient | Binary Bits | Coefficient
-- | -- | -- | --
kilobits | 1000 | kibibits | 1024
megabits | 1000^2 | mebibits | 1024^2


BradLarson:
@swift-ci Please test.

BradLarson:
@swift-ci Please test Linux platform.

PR #3067, opened 2021-08-28
[SR-6671] Fix `Decimal` implementation of `FloatingPoint` constants

xwu:
This PR is a reprise of #1386, which has long ago gone stale. There are several issues with `Decimal` implementations of `FloatingPoint` constants:

First, as reported in SR-6671,  the `leastNormalMagnitude` and the `leastNonzeroMagnitude` actually aren't. This is corrected in both the overlay and in swift-corelibs, with amendments to the corresponding tests.

Second, as pointed out by @stephentyrone in #1386, the implementation of `leastFiniteMagnitude` is incorrect. (It is also not a constant required by `FloatingPoint`, but with API stability it's too late now to remove from `Decimal`'s API. It is left to the code owners as a separate matter whether they wish to deprecate this API.) Apple's documentation of the API states that it is "the decimal that contains the smallest possible non-infinite magnitude for the underlying representation." Magnitude is never negative, and the correct value here is _zero_. In this PR we make this API an explicit synonym for `.zero`. This is corrected in both the overlay and in swift-corelibs, with amendments to the corresponding tests.


xwu:
@swift-ci test

xwu:
@swift-ci test

xwu:
@swift-ci test

xwu:
@swift-ci test macOS

xwu:
@swift-ci test

xwu:
@swift-ci test

spevans (approved):

stephentyrone:
(Just back from vacation, sorry for delay in reading this)

I'm worried about changing the value of `leastFiniteMagnitude` introducing subtle bugs to existing code that "works" with the current (incorrect) value. My advice would be instead to deprecate the property while keeping its value as is.

xwu:
@stephentyrone Yeah, I as well. On the other hand, the value of the property is _currently_ documented in a way that emphasizes its "magnitudeness" and there may be current code subtly broken relying on that. I think this should be deprecated regardless with a good error message, but it's not clear to me which value of the property is best in its deprecated state.

stephentyrone:
Default assumption should be that already-compiled code is happy with the existing behavior (wrong though it may be), and we shouldn't break them. New code will get a deprecation warning, so we can catch them there. @spevans, how do you want to handle this?

xwu:
#3071

PR #3068, opened 2021-08-29
[SR-15132][SR-15134] Fix `Decimal` implementation of significand APIs

xwu:
This PR fixes <strike>two</strike> three newly discovered issues with `Decimal` implementations of APIs that deliberately mimic those of `FloatingPoint`.

First, as reported in SR-15132, `significand` shouldn't be negative, but `Decimal` returns a negative "significand" for a negative value:

```swift
import Foundation

let x = -42 as Double
x.significand.sign // plus
let y = -42 as Decimal
y.significand.sign // minus (!)
```

Second and not reported in a separate bug, for types conforming to `FloatingPoint`, the initializer `init(sign:exponent:significand:)` returns a value where the sign is determined by the following notional calculation:

```swift
(sign == .minus ? -1 : 1) * significand * radix ** exponent
```

In other words, the sign of the result is not determined by `sign` alone, but rather the result is negative if and only if `sign` and `significand.sign` are of _opposite_ signs. This is not respected by `Decimal`:

```swift
import Foundation

let x = -42 as Double
Double(sign: .plus, exponent: 0, significand: x) // -42
let y = -42 as Decimal
Decimal(sign: .plus, exponent: 0, significand: y) // 42 (!)
```

Third, now reported in SR-15134, `init(sign:exponent:significand:)` currently barfs when exponent is incremented out of range:

```swift
import Foundation
let a = Decimal.leastNonzeroMagnitude
print(Decimal(sign: .plus, exponent: -10, significand: a))
// 100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
let b = Decimal.greatestFiniteMagnitude
print(Decimal(sign: .plus, exponent: 10, significand: b))
// 0.00000000000000000000000000000000000000000000000000000000000000000000000000000000340282366920938463463374607431768211455
```

Instead, the value should properly underflow or overflow (to NaN since `Decimal` has no representation of infinity), respectively, in line with the behavior of binary floating-point types:

```swift
let x = Double.leastNonzeroMagnitude
print(Double(sign: .plus, exponent: -10, significand: x)) // 0.0
let y = Double.greatestFiniteMagnitude
print(Double(sign: .plus, exponent: 10, significand: y)) // inf
```

To fix this, delegate the work of increasing or decreasing the exponent to `NSDecimalMultiplyByPowerOf10`.

These issues are addressed for both the overlay and the swift-corelibs implementation; tests are added.

### DIFF
--- a/Tests/Foundation/Tests/TestDecimal.swift
+++ b/Tests/Foundation/Tests/TestDecimal.swift
@@ -7,6 +7,9 @@
 // See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
+import Foundation
+import XCTest
+
 class TestDecimal: XCTestCase {
 
     func test_NSDecimalNumberInit() {
@@ -809,10 +812,19 @@ class TestDecimal: XCTestCase {
     }
 
     func test_Significand() {
-        let x = -42 as Decimal
+        var x = -42 as Decimal
         XCTAssertEqual(x.significand.sign, .plus)
-        let y = Decimal(sign: .plus, exponent: 0, significand: x)
-        XCTAssertEqual(y.sign, .minus)
+        var y = Decimal(sign: .plus, exponent: 0, significand: x)
+        XCTAssertEqual(y, -42)
+        y = Decimal(sign: .minus, exponent: 0, significand: x)
+        XCTAssertEqual(y, 42)
+
+        x = 42 as Decimal
+        XCTAssertEqual(x.significand.sign, .plus)
+        y = Decimal(sign: .plus, exponent: 0, significand: x)
+        XCTAssertEqual(y, 42)
+        y = Decimal(sign: .minus, exponent: 0, significand: x)
+        XCTAssertEqual(y, -42)
     }
 
     func test_SimpleMultiplication() {
--- a/Tests/Foundation/Tests/TestDecimal.swift
+++ b/Tests/Foundation/Tests/TestDecimal.swift
@@ -825,6 +825,11 @@ class TestDecimal: XCTestCase {
         XCTAssertEqual(y, 42)
         y = Decimal(sign: .minus, exponent: 0, significand: x)
         XCTAssertEqual(y, -42)
+
+        let a = Decimal.leastNonzeroMagnitude
+        XCTAssertEqual(Decimal(sign: .plus, exponent: -10, significand: a), 0)
+        let b = Decimal.greatestFiniteMagnitude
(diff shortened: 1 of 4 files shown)


xwu:
@swift-ci test

Tests/Foundation/Tests/TestDecimal.swift
```diff
@@ -808,6 +808,13 @@ class TestDecimal: XCTestCase {
         XCTAssertEqual(zero3.description, "0")
     }
 
+    func test_Significand() {
+        let x = -42 as Decimal
+        XCTAssertEqual(x.significand.sign, .plus)
+        let y = Decimal(sign: .plus, exponent: 0, significand: x)
+        XCTAssertEqual(y.sign, .minus)
```
spevans, line 815: Could you expand the tests to explicitly test all 4 combinations of `sign`/`significand` being +/-, and test the result of the initialiser against the full value, thanks. I know its a bit cumbersome but I find with number tests its useful to cover all the cases if possible.

xwu:
@swift-ci test

spevans:
I agree your fix is correct in that it makes it compatible with the operation of the types conforming to `FloatingPoint` as documented: https://developer.apple.com/documentation/swift/float/1847179-init however the parameters are listed as being: 
```
sign: The sign to use for the new value.
exponent: The new value’s exponent.
significand: The new value’s significand.
```

Do you think the documentation is assuming the `significand`  is always positive? From my reading of it, the `sign` parameter sets the sign of the resulting value.

I wonder if a better fix would be to trap on a negative `significand` - although that could be considered API breaking. 
I think the documentation may need to be made a bit clearer as it currently seems to contradict itself.

I'd be interested to hear @stephentyrone thoughts on this.

Also, at some point the documentation for https://developer.apple.com/documentation/foundation/decimal/2143030-init needs to be expanded since its single line summary could be interpreted as the `sign` parameter does indeed set the value's sign.

xwu:
@spevans The documentation's description of the arguments assumes (nay, _requires_) that `significand` is a significand (that is, has positive sign and zero exponent). The behavior of the function when `significand` isn't a significand is described later on. The method is meant to be our spelling of the IEEE operation `scaleB`.

xwu:
Note that I'm rolling in a fix for yet another bug in `init(sign:exponent:significand:)`; the PR description has been updated accordingly.

xwu:
@swift-ci test

xwu:
@swift-ci test

spevans (approved):

PR #3069, opened 2021-08-29
Fix orthography of 'orthography'

xwu:
The irony is fun, but particularly in the case of the placeholder API, it only serves its purpose if it's holding the right place.


xwu:
@swift-ci test Linux

PR #3070, opened 2021-08-29
[SR-14384][SR-14974] Fix `Decimal.ulp`, `nextUp`, `nextDown`

xwu:
This PR is a follow-up to #3014, addressing the first issue outlined there and correspondingly addressing SR-14384 regarding `nextDown` and `nextUp`. Namely:

* The implementation of `ulp` does not currently actually return the "unit in the last place" as an IEEE floating-point type would, which should reflect the spacing between two consecutive representable numbers. Instead, the result currently reflects the idiosyncratic internal representation of `Decimal` values, with the consequence that the "unit in the last place" of `1` is allegedly `1`.
* Relatedly, `nextUp` is documented to give "the least representable value that is greater than this decimal," but `(1 as Decimal).nextUp` is allegedly `2`. (Similarly for `nextDown`.)

This behavior is discordant with that of binary floating-point types as well as the documentation. This PR addresses both of these issues by computing the true spacing between two consecutive representable numbers as the `ulp`, and adjusting the implementations of both `nextUp` and `nextDown` to rely on `ulp`. Both the overlay and the swift-corelibs implementations are changed, and appropriate tests are added.

For this implementation of `ulp`, the precomputed array of powers of 10 as well as the shorthand initializer used to create deliberately non-compacted values for that array is used. This code was already present in the swift-corelibs version but not the overlay, so it has been moved up to live next to other code that is shared between the two implementations for ease of keeping them in sync.

### DIFF
--- a/Sources/Foundation/Decimal.swift
+++ b/Sources/Foundation/Decimal.swift
@@ -831,6 +831,23 @@ extension Decimal {
     }
 
     public var nextUp: Decimal {
+        if _isNegative == 1 {
+            if _exponent > -128
+                && (_mantissa.0, _mantissa.1, _mantissa.2, _mantissa.3) == (0x999a, 0x9999, 0x9999, 0x9999)
+                && (_mantissa.4, _mantissa.5, _mantissa.6, _mantissa.7) == (0x9999, 0x9999, 0x9999, 0x1999) {
+                return Decimal(
+                    _exponent: _exponent &- 1, _length: 8, _isNegative: 1, _isCompact: 1,
+                    _reserved: 0, _mantissa: (0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff))
+            }
+        } else {
+            if _exponent < 127
+            && (_mantissa.0, _mantissa.1, _mantissa.2, _mantissa.3) == (0xffff, 0xffff, 0xffff, 0xffff)
+            && (_mantissa.4, _mantissa.5, _mantissa.6, _mantissa.7) == (0xffff, 0xffff, 0xffff, 0xffff) {
+                return Decimal(
+                    _exponent: _exponent &+ 1, _length: 8, _isNegative: 0, _isCompact: 1,
+                    _reserved: 0, _mantissa: (0x999a, 0x9999, 0x9999, 0x9999, 0x9999, 0x9999, 0x9999, 0x1999))
+            }
+        }
         return self + ulp
     }
 
--- a/Sources/Foundation/Decimal.swift
+++ b/Sources/Foundation/Decimal.swift
@@ -492,55 +492,98 @@ extension Decimal : Strideable {
     }
 }
 
-extension Decimal {
-    // (Used by `_powersOfTen` and `ulp`; note that the representation isn't compact.)
-    fileprivate init(_length: UInt32, _mantissa: (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)) {
-        self.init(_exponent: 0, _length: _length, _isNegative: 0, _isCompact: 0,
+private extension Decimal {
+    // Creates a value with zero exponent.
+    // (Used by `_powersOfTen*`.)
+    init(_length: UInt32, _isCompact: UInt32, _mantissa: (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)) {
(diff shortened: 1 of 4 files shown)


xwu:
@swift-ci test Linux

xwu:
@swift-ci test Linux

xwu:
@swift-ci test Linux

xwu:
@swift-ci test macOS

xwu:
@swift-ci test

Tests/Foundation/Tests/TestDecimal.swift
```diff
@@ -860,8 +843,41 @@ class TestDecimal: XCTestCase {
...
+
+        x = 1
+        XCTAssertEqual(x.ulp, Decimal(string: "1e-38")!)
+        XCTAssertEqual(x.nextDown, x - Decimal(string: "1e-38")!)
+        XCTAssertEqual(x.nextUp, x + Decimal(string: "1e-38")!)
+
+        x = .leastNonzeroMagnitude
+        // XCTAssertEqual(x.ulp, x)        // SR-6671
+        // XCTAssertEqual(x.nextDown, 0)   // SR-6671
+        // XCTAssertEqual(x.nextUp, x + x) // SR-6671
```
spevans, line 867: These tests can be enabled now that #3067 has been merged.

Tests/Foundation/Tests/TestDecimal.swift
```diff
@@ -860,8 +843,41 @@ class TestDecimal: XCTestCase {
...
+
+        x = 0
+        XCTAssertEqual(x.ulp, Decimal(string: "1e-128")!)
+        XCTAssertEqual(x.nextDown, -Decimal(string: "1e-128")!)
+        XCTAssertEqual(x.nextUp, Decimal(string: "1e-128")!)
+
+        x = -1
+        XCTAssertEqual(x.ulp, (1 as Decimal).ulp)
+        XCTAssertEqual(x.nextDown, -((1 as Decimal).nextUp))
+        XCTAssertEqual(x.nextUp, -((1 as Decimal).nextDown))
```
spevans, line 880: Could you add a few extra tests for these last 3 to test the result against absolute values eg:
`XCTAssertEqual(x.ulp, Decimal(string: "1e-38")!)`
in case there is an equal and opposite bug in both methods.

spevans (changes requested):
All ok, just requires some changes in the tests.

xwu:
Will revise later this week or over the weekend. There’s a subtle bug owing to the fact that the max supported value of this type is not `10 ** 38 - 1` but rather `2 ** n - 1`.

spevans:
@swift-ci test

xwu:
Thanks :) It’s not quite done, and I’ll keep plugging away at it.

Sources/Foundation/Decimal.swift
```diff
@@ -718,22 +812,47 @@ extension Decimal {
...
-            _exponent: _exponent, _length: 1, _isNegative: 0, _isCompact: 1,
+            _exponent: max(exponent, -128), _length: 1, _isNegative: 0, _isCompact: 1,
             _reserved: 0, _mantissa: (0x0001, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000))
     }
 
     public var nextUp: Decimal {
-        return self + Decimal(
-            _exponent: _exponent, _length: 1, _isNegative: 0, _isCompact: 1,
-            _reserved: 0, _mantissa: (0x0001, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000))
+        if _isNegative == 1 {
```
xwu, line 834: This logic is the `Decimal` counterpart to https://github.com/apple/swift/blob/4c923d6dda6453d3af635f0ac1d64935a21673bb/stdlib/public/core/FloatingPointTypes.swift.gyb#L946.

xwu:
@spevans This should be ready now.

millenomi:
@phausler Would you mind putting a checkmark here if it looks right?

phausler (approved):
There is a question of syncing this but overall the changes look accurate. ulp and such were never fully defined so this seems plausible to bring em in line w/ the expectations.

xwu:
@phausler By syncing you mean the overlay? It’s still being built from the overlay files in this repo, yes?

phausler:
I am talking about the actual overlay, this has never been the real overlay

PR #3071, opened 2021-08-30
Restore previous value of `Decimal.leastFiniteMagnitude` for compatibility and deprecate

xwu:
See discussion in #3067.


xwu:
@swift-ci test

spevans (changes requested):
I agree with reverting to the previous value but the deprecation would require more discussion so that should be removed for now (although the clarifying message is actually really useful).

stephentyrone:
@spevans Let's start that discussion then, since the property is both useless and misleading and should definitely be deprecated.

spevans:
Sorry, I meant it would be up to the Foundation team to decide on the API change, so just removing the deprecation change would allow the PR to go through whilst they come to a decision. Personally I think its a reasonable change.

cc @millenomi

millenomi:
I'm okay with the deprecation. @phausler, any objections?

phausler:
nah, this is fine to deprecate

spevans (approved):

PR #3074, opened 2021-09-04
[SR-15156] Another fix for `Decimal(sign:exponent:significand:)`.

xwu:
This issue has been present all along, and my previous fix in #3068 didn't address it because I didn't know of the problem--and (perhaps unsurprisingly) reproduced the same problem despite a totally new implementation of the initializer. Consider the following:

```swift
import Foundation

let x = Double.greatestFiniteMagnitude
Double(sign: .plus, exponent: 10, significand: x)
// +inf
Double(sign: .plus, exponent: .max, significand: x)
// +inf

let y = Decimal.greatestFiniteMagnitude
Decimal(sign: .plus, exponent: 10, significand: y)
// nan (correct, since `Decimal` has no representation of infinity)
Decimal(sign: .plus, exponent: .max, significand: y)
// runtime error (!)
```

This is because the implementation of `Decimal(sign:exponent:significand:)` converts between `Int` and an integer type of smaller bit width (but still plenty wide enough to represent more than twice the largest possible exponent) using a trapping initializer instead of clamping.

The present PR fixes that oversight now (in both the corelibs and overlay implementations) and adds tests.

### DIFF
--- a/Darwin/Foundation-swiftoverlay-Tests/TestDecimal.swift
+++ b/Darwin/Foundation-swiftoverlay-Tests/TestDecimal.swift
@@ -580,8 +580,10 @@ class TestDecimal : XCTestCase {
 
         let a = Decimal.leastNonzeroMagnitude
         XCTAssertEqual(Decimal(sign: .plus, exponent: -10, significand: a), 0)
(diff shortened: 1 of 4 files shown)


xwu:
@swift-ci test

spevans (approved):

PR #3078, opened 2021-09-11
RunLoop: `typeinfo.h` was removed in VS2019 16.3

compnerd:
`typeinfo.h` is a non-standard header and not always available.  Guard
the inclusion with an availability check.


compnerd:
@swift-ci please test

PR #3080, opened 2021-09-11
URL: correct header inclusion

compnerd:
`assert.h` is universally available and required for the use of
`assert`.


compnerd:
@swift-ci please test

PR #3081, opened 2021-09-16
Plugin: fix calling readlink() with a larger PATH_MAX

buttaface:
The unmodified line built fine for Android with CMake 3.19.7, but recent versions of CMake and the Android NDK enable `-D_FORTIFY_SOURCE=2`, which [errors because `PATH_MAX` on linux is 4096](https://github.com/buttaface/swift-android-sdk/runs/3619410286?check_suite_focus=true#step:7:2574). [Since `CFMaxPathSize` is set much smaller](https://github.com/apple/swift-corelibs-foundation/blob/6634ee3421bcd3e2c1a7502088cb5c26f441106f/CoreFoundation/Base.subproj/CFInternal.h#L1101), that recently enabled compile-time checker errors.

Alternately, we could set `CFMaxPathSize/Length` based on `PATH_MAX` explicitly, rather than implicitly for win32/macOS right now. @millenomi, let me know what you prefer.


buttaface:
Ping, passed CI, just needs to be merged.

PR #3082, opened 2021-09-25
Reapply: Make CF_ENUM compatible with -Welaborated-enum-base

etcwilde:
The syntax relied upon by CF_ENUM does not align with the intent of the C++ feature, and that correction extends to its use in C mode. Add a temporary fix that 'swallows' the user-supplied `typealias` by defining a dummy type, and then does a forward-definition-then-typedef-then-enum-opening dance to apply the correct syntax.

This patch was originally applied in d28b5470f27e961c029e5745ec03d1e10244e0d6, then was undone in d8e8a8b92b3a8af8381a11155328c1bba1c6bd2c. `-Welaborated-enum-base` is taking effect on the swift rebranch branch, so we will need to fix the syntax before that lands.
Failing CI job: https://ci.swift.org/view/Swift%20rebranch/job/oss-swift-rebranch-incremental-RA-linux-ubuntu-18_04/26/console


etcwilde:
@swift-ci please test

etcwilde:
I'm going to go ahead and merge this now.

PR #3086, opened 2021-10-11
CoreFoundation: add a workaround for Linux with the rebranch

compnerd:
This adds an optimization barrier in the loop that is potentially
getting mis-optimized with the rebranch on Linux.  This should allow us
to rebranch and follow up with a proper fix subsequently.

Thanks to @etcwilde for tracking down this issue!

### DIFF
--- a/CoreFoundation/String.subproj/CFAttributedString.c
+++ b/CoreFoundation/String.subproj/CFAttributedString.c
@@ -23,6 +23,10 @@
 @end
 #endif
 
+#if defined(__linux__)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

etcwilde:
> I would however agree that the change should have a comment referencing the SR .... which I swear I put in. :sigh: I don't want to touch this until the results are back from the associated PR. Worst case, a follow up to add that is something I might argue for, just to avoid the CI penalty 

That's fine by me.

PR #3089, opened 2021-10-11
Bundle: avoid second chance bundle loading on Windows

compnerd:
When non-ObjC deployment targets, do not support the second-chance
bundle loading recovery for flat-style bundles on Windows.  No one is
currently using this style and this causes a potentially spurious Bundle
creation rather than a failure to create the bundle as expected with a
Swift runtime.  This repairs the last test failure on Windows after the
last CoreFoundation merge.


millenomi:

millenomi:

PR #3091, opened 2021-10-16
Re-enable string Optimizations

etcwilde:
I think I've found the fix for the failing optimization on the main swift branch.
We should be able to re-enable optimizations once that merges.

The fixing merge is: https://github.com/apple/llvm-project/pull/3439


etcwilde:
https://github.com/apple/llvm-project/pull/3439

@swift-ci please test

PR #3094, opened 2021-10-19
Annotate CFPreferencesCopyMultiple as having a _Nullable return value

AtariDreams:
Especially since the function can return NULL. It seems that the original author forgot to annotate it as such.

### DIFF
--- a/CoreFoundation/Preferences.subproj/CFPreferences.h
+++ b/CoreFoundation/Preferences.subproj/CFPreferences.h
@@ -93,7 +93,7 @@ keysToFetch that are not present in the returned dictionary
 are not present in the domain.  If keysToFetch is NULL, all
 keys are fetched. */
 CF_EXPORT
-CFDictionaryRef CFPreferencesCopyMultiple(_Nullable CFArrayRef keysToFetch, CFStringRef applicationID, CFStringRef userName, CFStringRef hostName);
+_Nullable CFDictionaryRef CFPreferencesCopyMultiple(_Nullable CFArrayRef keysToFetch, CFStringRef applicationID, CFStringRef userName, CFStringRef hostName);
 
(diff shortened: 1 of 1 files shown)


etcwilde:
As @millenomi has already said, we can't take changes to the signature of public API.
This doesn't break ABI, but it has wider source-breaking effects when the API is imported into Swift, for example.

AtariDreams:
Wouldn't that just make it worse if swift programs expect that function to be nonnull when it can return NULL? Especially when someone may force-unwrap the return value thinking there's a guarantee of non-null

etcwilde:
As-is, the API will be imported as an implicitly unwrapped member. And you are right that it will fail if used when it's in a bad state. Unfortunately, changing the nullability affects all of the use-sites, which is a source breaking change. Without a very good understanding of exactly how much code we would be breaking that is in existence today, it isn't something that is acceptable. If the failure starts affecting people and bugs appear in the bug queues for this API, it may be a different story, but as of today, I don't see anyone commenting on unexpected fatal errors due to this API in the bug tracker.

millenomi:
Yeah. There's been some work around these at the time of introduction of nullability — some are intentional, as returning NULL occurs in situations that are developer or unfixable environment errors anyway, such as failing to allocate memory; some are mistakes. Either way, the time has passed, and these cannot be changed.

millenomi:
Also, as usual, a reminder that we have already signaled that CF API are not useful in general in Swift for Linux, are not available at all outside of Linux (e.g. in the Windows build), and we are not interested in changing them — we are more interested in producing useful Swift-level API instead.

AtariDreams:
Alright, but are you sure this doesn't affect macOS or Apple's platforms? If so, I will close this.

PR #2940, opened 2020-12-08
Add support and tests for URLComponents.percentEncodedQueryItems

rauhul:
- rdar://59439703 (URLComponents.percentEncodedQueryItems missing on linux)

part 2 of #2936

PR #2929, opened 2020-11-26
Add support and tests for URLComponents.percentEncodedQueryItems

rauhul:
rdar://59439703 (URLComponents.percentEncodedQueryItems missing on linux)

resubmission of #2870 against `main`

PR #3096, opened 2021-10-20
Add percentEncodedQueryItems

lha:
Some code with queryItems, use existing code in CFNetwork

Fixes SR-12340

This seems to have been attempted a couple of times before (#2929, #2940) by @rauhul


millenomi:
@lha Please ping me when this is fixed and I'll force merge since tests all pass.

PR #3101, opened 2021-10-29
NFC: Fix Xcode prototype warnings

AtariDreams:
Xcode warns that "old style functions have no prototypes." Having (void) as the parameter list fixes these warnings. No functional change.


millenomi:
We should be perhaps adding prototypes to these instead.

AtariDreams:
Where though?

millenomi:
For internal calls, top of the file. For CF_PRIVATE calls, in the corresponding private or internal header.

PR #3103, opened 2021-10-30
Specify variable type in CUserNotifications

AtariDreams:
Specifying a variable without a type is not Standard C, even for sizeof

### DIFF
--- a/CoreFoundation/AppServices.subproj/CFUserNotification.c
+++ b/CoreFoundation/AppServices.subproj/CFUserNotification.c
@@ -190,10 +190,10 @@ static SInt32 _CFUserNotificationSendRequest(CFAllocatorRef allocator, CFStringR
     
 #if TARGET_OS_OSX
     const char *namebuffer = NOTIFICATION_PORT_NAME_MAC;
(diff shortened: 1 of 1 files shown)


CoreFoundation/AppServices.subproj/CFUserNotification.c
```diff
@@ -190,10 +190,10 @@ static SInt32 _CFUserNotificationSendRequest(CFAllocatorRef allocator, CFStringR
     
 #if TARGET_OS_OSX
     const char *namebuffer = NOTIFICATION_PORT_NAME_MAC;
-    const nameLen = sizeof(NOTIFICATION_PORT_NAME_MAC);
+    const CFIndex nameLen = sizeof(NOTIFICATION_PORT_NAME_MAC);
```
parkera, line 193: I think we usually use `size_t` for these, actually.

AtariDreams: Except the arguments they are used in expect CFIndexes, so it makes the most sense in this case. Normally you would be right though.

millenomi:

PR #3107, opened 2021-11-04
Prefer the pthread types from CDispatch

xymus:
The types `pthread_mutex_t` and `pthread_cond_t` were defined via both `SwiftGlibc` and `CDistpatch` but a clang fix should merge these decls and consider `CDispatch` as the true origin.

Let's first see if we still need the explicit reference to `Glibc` when referring to those types.

rdar://84677782

### DIFF
--- a/Sources/Foundation/NSLock.swift
+++ b/Sources/Foundation/NSLock.swift
@@ -22,11 +22,6 @@ public protocol NSLocking {
     func unlock()
 }
 
-#if canImport(Glibc)
-typealias pthread_mutex_t = Glibc.pthread_mutex_t
-typealias pthread_cond_t = Glibc.pthread_cond_t
-#endif
-
 #if os(Windows)
 private typealias _MutexPointer = UnsafeMutablePointer<SRWLOCK>
 private typealias _RecursiveMutexPointer = UnsafeMutablePointer<CRITICAL_SECTION>


xymus:
@swift-ci Please test

xymus:
@swift-ci Please test Linux

xymus:
@MaxDesiatov Do you remember why you added these lines? I'm wondering if you may have run into the same duplicated definition issue we're trying to fix in clang with https://github.com/apple/llvm-project/pull/3497.

MaxDesiatov:
It could be, but these lines no longer have any effect when building for WASI. libc module on that platform is no longer called `Glibc`, it's `WASILibc`, so this merging shouldn't have any effect for WASI.

xymus:
Sounds good then, thank you for the information Max!

The Linux tests are failing in SwiftDocCTests, it looks unrelated.

xymus:
@swift-ci Please test Linux

xymus:
@swift-ci Please test Linux

xymus:
I believe the Linux tests are broken by the DocC tests that rely in networking and unusual temporary directory. (rdar://85055022)

xymus:
@swift-ci Please test Linux

xymus:
One more tweak will be needed but this one should be safe by itself.

MaxDesiatov (approved):

PR #3109, opened 2021-11-11
FileManager: improve compatibility for `setAttributes`

compnerd:
`setAttributes(_:ofItemAtPath:)` and friends should not `fatalError`
when encountering an attribute which cannot be set.  Instead, we should
ignore those to match the behaviour of the reference implementation.
This should allow us to properly enable some test cases in swift-driver.

### DIFF
--- a/Sources/Foundation/FileManager.swift
+++ b/Sources/Foundation/FileManager.swift
@@ -450,7 +450,7 @@ open class FileManager : NSObject {
                     throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.fileWriteUnknown.rawValue)
                     
                 default:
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

Going to run a local test on Windows.

compnerd:
@swift-ci please test

compnerd:
I'll test locally on Windows to verify that it doesn't break with the test changes

compnerd:
@swift-ci please test

compnerd:
Windows looks good: https://gist.github.com/compnerd/b31f1fb994b2a3a3089f74e46cb337a6

millenomi:
For context: Darwin does so as well.

millenomi (approved):

PR #3110, opened 2021-11-12
Fix include logic to only include Objective-C if the target is a Mac, or is explicitly told to have Objective-C

AtariDreams:


millenomi:

parkera:
I think this one is obsolete now.

PR #3111, opened 2021-11-15
CoreFoundation: ensure that we do not overflow the dest

compnerd:
This `memcpy` could potentially overflow if the source was smaller than
the destination.  We now trim this to the minimum of the source and dest
(similar to `strlcpy`).


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #3113, opened 2021-11-16
Allow `Process.waitUntilExit()` to be called from any thread

pcbeard:
This PR relaxes which thread / runloop `Process.waitUntilExit()` can be called from. Before this change, a different thread would return prematurely from `.waitUntilExit()` with `Process.isRunning` still true. This change fixes the problem by observing whether the `.runLoop` property matches `RunLoop.currentRunLoop`, and if it doesn't then it no longer calls `currentRunLoop.run(mode: .default, before: Date(timeIntervalSinceNow: 0.05))` which would always return `false` for this new thread. Instead, it simply calls `currentRunLoop.run(until: Date(timeIntervalSinceNow: 0.05))` and keeps checking the `.isRunning` property.

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -1151,12 +1151,14 @@ open class Process: NSObject {
 
     // poll the runLoop in defaultMode until process completes
     open func waitUntilExit() {
-        //
+        let runInterval = 0.05
         let currentRunLoop = RunLoop.current
         let checkRunLoop : () -> Bool = (currentRunLoop == self.runLoop)
-                ? { currentRunLoop.run(mode: .default, before: Date(timeIntervalSinceNow: 0.05)) }
-                : { currentRunLoop.run(until: Date(timeIntervalSinceNow: 0.05)); return true }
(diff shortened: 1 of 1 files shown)


Sources/Foundation/Process.swift
```diff
@@ -1153,10 +1153,13 @@ open class Process: NSObject {
 
     // poll the runLoop in defaultMode until process completes
     open func waitUntilExit() {
-        
-        repeat {
-            
-        } while( self.isRunning == true && RunLoop.current.run(mode: .default, before: Date(timeIntervalSinceNow: 0.05)) )
+        //
```
compnerd, line 1154: Did you intend to add a comment here that you forgot?

compnerd (approved):
LGTM; Please wait for @millenomi to have a look at this as well.

Sources/Foundation/Process.swift
pcbeard: Remove stray comment, update self.runLoop to enable earlier wakeup from manager thread.

pcbeard:
@swift-ci test

parkera (approved):

PR #3114, opened 2021-11-16
Fix KVO for RawRepresentable types

ffried:
Since the overlay has moved to this repository now, this ports the changes in [PR 20757](https://github.com/apple/swift/pull/20757) over here as well.
I couldn't find any tests, so I created a new test class. However, I was unable to test them locally.

This fixes the `oldValue` / `newValue` properties of `NSKeyValueObservedChange<Value>` being always `nil` when key-value-observing `RawRepresentable` types (typically enums) like `URLSessionTask.State`.

Resolves [SR-5872](https://bugs.swift.org/browse/SR-5872).

### DIFF
--- a/Darwin/Foundation-swiftoverlay-Tests/TestNSObject.swift
+++ b/Darwin/Foundation-swiftoverlay-Tests/TestNSObject.swift
@@ -1,0 +1,85 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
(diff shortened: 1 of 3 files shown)


ffried:
In addition to just porting the changes, I was (likely?) able to solve the problem without adding any new public API. Again, I wasn't able to test this locally. How would local testing work here?

millenomi:
The repository contains the overlay code but isn't set up to CI it. Appropriate people have been pinged.

denizdogan:
This has been an issue since at least 2018 (see https://github.com/apple/swift/pull/20757)

I'm sorry, but why is this taking so long to just fix?

amalenduk:
Still an issue

yanshouwang:
Still not fixed with Swift 5

songgeb:
still not fixed

AMatecki:
Hey, any updates on this issue? This bug is causing serious issues in our app

ffried:
It seems that this is no longer a problem that can be fixed by non-Apple people. The source code for this moved once again and it seems like it's not longer public:

https://github.com/search?q=org%3Aapple+NSKeyValueObservation&type=code
https://github.com/search?q=org%3Aswiftlang+NSKeyValueObservation&type=code

I'm closing this PR.

PR #3116, opened 2021-11-17
add arm64 support for M1 Mac

DanboDuan:


cmake/modules/SwiftSupport.cmake
```diff
@@ -9,8 +9,12 @@
 function(get_swift_host_arch result_var_name)
   if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
     set("${result_var_name}" "x86_64" PARENT_SCOPE)
-  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
-    set("${result_var_name}" "aarch64" PARENT_SCOPE)
+  elseif ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "AArch64|aarch64|arm64")
```
compnerd, line 12: Similar to swift-system, not sure if the `AArch64` case is needed, but this seems fine.

compnerd (approved):

compnerd:
@swift-ci please test

cmake/modules/SwiftSupport.cmake
DanboDuan: I can delete `AArch64 ` if not needed. the code is just copied from swift-atomics/cmake/modules/SwiftSupport.cmake. Better to keep it

tomerd: I think both are needed to cover apple silicon and arm linux. cc @drexin

compnerd: @tomerd I'm not sure - I'm talking about _literally_ `AArch64`, not `aarch64`.  Its a question of is there an extra case being matched against or not.  Either way, it doesn't actually matter,

drexin: I’m not sure about this particular case, but I have seen AMD64 and amd64 on different platforms. I don’t think it hurts to keep it as is.

drexin (approved):

millenomi:
Uh, wait, no. The CMake build isn't supported on Mac. What's up?

DanboDuan:
> Uh, wait, no. The CMake build isn't supported on Mac. What's up?

but the `https://github.com/apple/swift` project needs it. when build the swift project with CMake and Ninjia, it will fail for arm64 not supported

krzyzanowskim:
Hi, looks like this PR is approved for quite some time already. What blocks the merge?

millenomi:
I'm okay with having arm64 support, but this isn't meant to be building _with CMake_ for Macs at all.

millenomi:
Happy to merge if someone explains what's up here.

DanboDuan:
> Happy to merge if someone explains what's up here.

this is for building `https://github.com/apple/swift` project with CMake and Ninja on M1 Mac. it will fail for arm64 not supported if arm64 is not supported @millenomi

PR #3117, opened 2021-11-27
build: allow building on Linux without Swift_FLAGS

compnerd:
When CMAKE_Swift_FLAGS is empty, the string substitution will be empty,
resulting in insufficient parameters to the REPLACE operation.  The
quoting ensures that an empty string does not cause that failure.


compnerd:
@swift-ci please test

PR #3118, opened 2021-12-04
[build] Explicitly link against icudata

Azoy:
In preparation for https://github.com/apple/swift/pull/40340, explicitly link against icudata.

### DIFF
--- a/CoreFoundation/CMakeLists.txt
+++ b/CoreFoundation/CMakeLists.txt
@@ -396,7 +396,8 @@ endif()
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
   target_link_libraries(CoreFoundation PRIVATE
     ICU::uc
-    ICU::i18n)
+    ICU::i18n
+    ICU::data)
 endif()
 
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ if(HAS_LIBDISPATCH_API)
   find_package(dispatch CONFIG REQUIRED)
 endif()
 
-find_package(ICU COMPONENTS uc i18n REQUIRED)
+find_package(ICU COMPONENTS uc i18n data REQUIRED)
 
 include(SwiftSupport)
 include(GNUInstallDirs)


milseman:
@swift-ci please test

milseman:
Do you know if this needs to be merged simultaneously with https://github.com/apple/swift/pull/40340, or can this be merged first?

milseman:
CC @compnerd

CoreFoundation/CMakeLists.txt
```diff
@@ -396,7 +396,8 @@ endif()
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
   target_link_libraries(CoreFoundation PRIVATE
     ICU::uc
-    ICU::i18n)
+    ICU::i18n
+    ICU::data)
```
compnerd, line 400: This seems wrong.  The tests in fact show that this is incorrect - the linking was sufficient.  The problem is that the data is not found at runtime.  This should be handled via a DT_RUNPATH or DT_RPATH in TestFoundation.  This potentially can break the Windows builds.

compnerd (changes requested):

compnerd:
> Do you know if this needs to be merged simultaneously with [apple/swift#40340](https://github.com/apple/swift/pull/40340), or can this be merged first?

Once the change is fixed up, this should be fine to merge first on its own, it will have no detrimental impact.

CoreFoundation/CMakeLists.txt
Azoy: In my testing, the icudataswift library was not being linked correctly at all in CoreFoundation, and TestFoundation is just an example executable that won't be able to run without the correct path for the data library. The data library seems to be a dependency of i18n and uc and running `ldd libFoundation.so` returns the following:
```
libicuucswift.so.65 => /swift/build/Ninja-RelWithDebInfoAssert/swift-linux-x86_64/lib/swift/linux/x86_64/libicuucswift.so.65 (0x00007fc5f7a03000)
libicui18nswift.so.65 => /swift/build/Ninja-RelWithDebInfoAssert/swift-linux-x86_64/lib/swift/linux/x86_64/libicui18nswift.so.65 (0x00007fc5f76d0000)
libicudataswift.so.65 => not found
```

I don't think this is something specific to TestFoundation unless we expect each executable depending on Foundation to pass in `DT_RPATH`.

buttaface: If CoreFoundation itself does not rely on libicudata but only indirectly through libicuuc, I think Saleem is right that we shouldn't add that dependency here. What may fix it instead is to add `$ORIGIN` to the `DT_RPATH` of the libicu libraries that we build for linux, which is also needed [to fix the currently broken `plutil` executable that we ship from this repo](https://bugs.swift.org/browse/SR-12394?focusedCommentId=62344&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-62344).

We should've been doing that already and I think it'll fix both that `plutil` issue and this newly uncovered ICU issue.

Azoy:
Closing because we found a solution that doesn't require this.

PR #3126, opened 2022-01-17
Refactor CFSocket callbacks to call Process.terminateRunLoop()

pcbeard:
This fixes a data race between setting `process.isRunning = false`
and then calling `CFRunLoopSourceInvalidate(process.runLoopSource)`.
Process.waitUntilExit() may wake up and clear .runLoopSource which
results in a fairly common crash on Windows.

Put this logic in a common method to reduce drift between the
Darwin / Linux and Windows code paths.

Thanks to @compnerd for finding the data race.

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -642,18 +642,7 @@ open class Process: NSObject {
                 process._terminationReason = .exit
             }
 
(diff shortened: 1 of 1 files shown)


pcbeard:
@swift-ci please test

PR #3129, opened 2022-01-19
build: allow static linking against ICU data

compnerd:
With the removal of the ICU data dependency in swiftCore, it is now
possible to statically link to the ICU, with static linking to the
ICU data as well.  This would allow us to remove the ICU dependency
from leaking into distributions which was previously required due to
the oversized cost of the ICU data.  With the single use now, it is
possible to statically link the data to avoid the distribution and
packaging.  In order to provide a migration path, this needs to be
optionalised, and as such, we only link the data if it is found (and
can be controlled through `-D ICU_DATA_LIBRARY_RELEASE=`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ if(HAS_LIBDISPATCH_API)
   find_package(dispatch CONFIG REQUIRED)
 endif()
 
-find_package(ICU COMPONENTS uc i18n REQUIRED)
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:

compnerd:
Tested on Windows CI: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=56360&view=results

That test simply indicates that dynamic linking still works after the change.  Actually changing to static linking is something that will need to be done subsequently.

tomerd:
cc @drexin

PR #3132, opened 2022-01-24
Foundation: `final`-ize many bridgeable interfaces

compnerd:
These are not overridden and are private fields which do not need to be
emitted to the vtable.  This allows these members to not impact the
module serialization when `@_implementationOnly import`s are used for
CoreFoundation by removing them entirely from the serialization path.


millenomi:

compnerd:

compnerd:
Hmm, the macOS builder seems to have triggered an assertion:
```
Assertion failed: (inserted && "there should be no duplicate decls"), function getDisplayDecls, file Module.cpp, line 967.
```

millenomi:
https://github.com/apple/swift/pull/41115

@swift-ci please test

PR #3133, opened 2022-01-26
[cmake] Fix ICU static linking

Azoy:
When linking Foundation, we were only passing `-licui18nswift` as a static library, but it depends on `icuucswift` and `icudataswift`, so make sure we static link against those as well.

Resolves: https://bugs.swift.org/browse/SR-15770 and rdar://problem/87991595

### DIFF
--- a/Sources/Foundation/CMakeLists.txt
+++ b/Sources/Foundation/CMakeLists.txt
@@ -176,9 +176,19 @@ if(NOT BUILD_SHARED_LIBS)
   get_filename_component(icu_i18n_dir "${icui18n_path}" DIRECTORY)
   string(REPLACE "lib" "" icu_i18n_basename "${icu_i18n_basename}")
 
+  get_target_property(icuuc_path ICU::uc IMPORTED_LOCATION)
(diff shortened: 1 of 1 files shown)


Azoy:
@swift-ci please build toolchain linux platform

stephentyrone (approved):

Azoy:
@swift-ci please test

Azoy:
@swift-ci please test

Azoy:
@swift-ci please build toolchain linux platform

compnerd (changes requested):
I think that we should be handling these similar to the `icuin` case - this will break builds when used outside of the host.

Azoy:
I've updated the PR to lookup the library name like `icui18n` is doing, but it requires explicitly passing `ICU_DATA_LIBRARY` in the build-script-impl, so I've done that here: https://github.com/apple/swift/pull/41041

Azoy:
@swift-ci please smoke test

compnerd (approved):

MaxDesiatov:
@swift-ci please test

millenomi (approved):

PR #3135, opened 2022-01-28
[5.5] Temporary fix for static linking on Linux

drexin:
When using @_implementationOnly, the dependencies do not get emited into the module files and static linking does not work. There is a proper fix in 5.6, so this is only a temporary solution for 5.5.


drexin:
@swift-ci test

tomerd (commented):
lgtm, thanks @drexin

drexin:
@swift-ci test

tomerd (approved):

millenomi:

PR #3136, opened 2022-02-01
[5.5] Prevent empty dictionary with non-String keys from encoding as a JSON object.

millenomi:

### DIFF
--- a/Tests/Foundation/Tests/TestJSONEncoder.swift
+++ b/Tests/Foundation/Tests/TestJSONEncoder.swift
@@ -869,6 +869,39 @@ class TestJSONEncoder : XCTestCase {
         XCTAssertEqual(JSONEncoder.OutputFormatting.withoutEscapingSlashes.rawValue, 8)
     }
 
+    func test_SR17581_codingEmptyDictionaryWithNonstringKeyDoesRoundtrip() throws {
+        struct Something: Codable {
+            struct Key: Codable, Hashable {
+                var x: String
+            }
+
+            var dict: [Key: String]
+
+            enum CodingKeys: String, CodingKey {
+                case dict
+            }
+
+            init(from decoder: Decoder) throws {
+                let container = try decoder.container(keyedBy: CodingKeys.self)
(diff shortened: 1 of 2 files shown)


millenomi:
@swift-ci please test

parkera (approved):
This looks good to take for 5.5

Tests/Foundation/Tests/TestJSONEncoder.swift
```diff
@@ -869,6 +869,39 @@ class TestJSONEncoder : XCTestCase {
...
+
+        let toEncode = Something(dict: [:])
+        let data = try JSONEncoder().encode(toEncode)
+        let result = try JSONDecoder().decode(Something.self, from: data)
+        XCTAssertEqual(result.dict.count, 0)
+    }
+
     // MARK: - Helper Functions
     private var _jsonEmptyDictionary: Data {
         return "{}".data(using: .utf8)!
```
vlm, line 907: Do we guarantee the empty dictionary is always encoded as `{}` regardless of type? If not, the variable name might be misleading.

millenomi: The empty dictionary is not always encoded as `[]` regardless of type. Specifically, `<T: Encodable> [String: T]` are encoded as JSON objects, and an empty dictionary of that type is encoded as `{}`.

Tests/Foundation/Tests/TestJSONEncoder.swift
```diff
@@ -869,6 +869,39 @@ class TestJSONEncoder : XCTestCase {
...
+
+            init(dict: [Key: String]) {
+                self.dict = dict
+            }
+        }
+
+        let toEncode = Something(dict: [:])
+        let data = try JSONEncoder().encode(toEncode)
+        let result = try JSONDecoder().decode(Something.self, from: data)
+        XCTAssertEqual(result.dict.count, 0)
```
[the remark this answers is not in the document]
millenomi: That makes sense. I'll add it in a follow-up.

vlm (approved):

PR #3137, opened 2022-02-01
[5.6] Prevent empty dictionary with non-String keys from encoding as a JSON object.

millenomi:


millenomi:
@swift-ci please test

PR #3138, opened 2022-02-01
Prevent empty dictionary with non-String keys from encoding as a JSON object.

millenomi:

This is for the `main` branch. Also:
- #3136
- #3137


millenomi:
@swift-ci please test

millenomi:
https://github.com/apple/swift/pull/41115

@swift-ci please test

PR #3142, opened 2022-02-07
[5.6] [cmake] Fix ICU static linking

Azoy:
When static linking Foundation, some of ICU's dependencies were not being explicitly linked against, so static linking failed for those platforms due to undefined references. This patch only affects the build rules for static linking.

Resolves: https://bugs.swift.org/browse/SR-15770 and rdar://87991595

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ if(HAS_LIBDISPATCH_API)
   find_package(dispatch CONFIG REQUIRED)
 endif()
 
-find_package(ICU COMPONENTS uc i18n REQUIRED)
+find_package(ICU COMPONENTS uc i18n REQUIRED OPTIONAL_COMPONENTS data)
 
 include(SwiftSupport)
 include(GNUInstallDirs)
--- a/Sources/Foundation/CMakeLists.txt
+++ b/Sources/Foundation/CMakeLists.txt
@@ -176,9 +176,19 @@ if(NOT BUILD_SHARED_LIBS)
   get_filename_component(icu_i18n_dir "${icui18n_path}" DIRECTORY)
   string(REPLACE "lib" "" icu_i18n_basename "${icu_i18n_basename}")
 
+  get_target_property(icuuc_path ICU::uc IMPORTED_LOCATION)
+  get_filename_component(icu_uc_basename "${icuuc_path}" NAME_WE)
+  string(REPLACE "lib" "" icu_uc_basename "${icu_uc_basename}")
+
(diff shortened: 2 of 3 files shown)


Azoy:
@swift-ci please test

Azoy:
@swift-ci please clean test Linux

Azoy:
@swift-ci please test Linux with the following PRs:
https://github.com/apple/swift/pull/41252

Azoy:
Please test with the following PRs:
https://github.com/apple/swift/pull/41252

@swift-ci please test Linux

Azoy:
@swift-ci please test Linux

Azoy:
@swift-ci please test Linux

Azoy:
@swift-ci please test Linux

CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ if(HAS_LIBDISPATCH_API)
   find_package(dispatch CONFIG REQUIRED)
 endif()
 
-find_package(ICU COMPONENTS uc i18n REQUIRED)
+find_package(ICU COMPONENTS uc i18n REQUIRED OPTIONAL_COMPONENTS data)
```
millenomi, line 50: We need to add data to be found for sure, but can you note here the cases where it'll be optional?

Sources/Foundation/CMakeLists.txt
```diff
@@ -176,9 +176,19 @@ if(NOT BUILD_SHARED_LIBS)
   get_filename_component(icu_i18n_dir "${icui18n_path}" DIRECTORY)
   string(REPLACE "lib" "" icu_i18n_basename "${icu_i18n_basename}")
 
+  get_target_property(icuuc_path ICU::uc IMPORTED_LOCATION)
```
millenomi, line 179: cc @compnerd would these work wrt Windows?

Azoy:
@swift-ci please test macOS

millenomi:
macOS is not required as this is CMake changes only; I won't block on macOS builds.

CMakeLists.txt
Azoy: I think this is a question for @compnerd (as I just cherry-picked his commit).

compnerd: data is optional as we may be dynamically loading the data, in which case, the application does not link against the data, ICU will internally link and load the data as necessary.

Sources/Foundation/CMakeLists.txt
compnerd: I believe that this should work fine with windows.  There is no `lib` prefix so it should be a no-op.  I believe this is already on main?

PR #3145, opened 2022-02-26
[EXPERIMENTAL] Add support for vcpkg

stevapple:
Windows developers have a relatively difficult experience for pulling in dependencies. This patch adds support for [`vcpkg`](https://github.com/microsoft/vcpkg)'s manifest mode, which can seamlessly integrate with CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,8 @@ if(HAS_LIBDISPATCH_API)
   find_package(dispatch CONFIG REQUIRED)
 endif()
 
-find_package(ICU COMPONENTS uc i18n REQUIRED OPTIONAL_COMPONENTS data)
+find_package(ICU COMPONENTS uc i18n REQUIRED)
+find_package(ICU OPTIONAL_COMPONENTS data)
 
 include(SwiftSupport)
 include(GNUInstallDirs)
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -1,0 +1,10 @@
+{
+  "name": "swift-corelibs-foundation",
+  "version-string": "0.1.0",
+  "dependencies": [
+    "curl",
+    "icu",
+    "libxml2",
+    "zlib"
+  ]
+}


stevapple:
Anyone to review?

millenomi:
Pulling the brake on this one because I'd like an explanation of benefits. How does this benefit the Visual Studio story at all? Are you building a VS build solution for Swift itself?

CMake is an internal build detail of Swift itself and I would prefer people do not attempt to maintain their own CMake-based build setups for end user software. (`swift build` should be used for that.)

millenomi:
(Yes, it is clear that SPM has Windows issues aplenty, but the focus should be if possible on those rather than exposing build details of a module that should be picked up with the redistributable instead.)

stevapple:
@millenomi This adds a manifest file of `vcpkg`, allowing developers to have:
```
-DCMAKE_TOOLCHAIN_FILE=[vcpkg-root]/scripts/buildsystems/vcpkg.cmake
```
flag passed on CMake, to let `vcpkg` provide the CMake packages, without having to install them in the system.

`vcpkg` will build everything from source and cache the artifacts across builds. This is not related to Visual Studio — if you wish, you can use it on macOS and Linux too.

Like we’re using CMake as an alternative to SwiftPM somewhere, `vcpkg` support is much of an alternative to `pkgconfig`. It is not meant to be used on official builds, but can largely speed up local development process.

millenomi:
And whoops I shouldn't have assumed this was VS-related, my bad.

stevapple:
> Do you have a use case that works with CMake, but not SPM? Is there anything blocking you from using SPM instead?

I don’t think any of the core libraries can be built with SwiftPM from ever since. We’re either using CMake or custom `build.py` script.

The main purpose of introducing `vcpkg` support is related to missing of `pkgconfig` on Windows, where CMake (and any other build system, including SwiftPM) cannot discover external libraries easily, and users can neither install system-wide dependencies easily. For using `vcpkg` with CMake, check out the documentation: https://github.com/microsoft/vcpkg/blob/master/docs/users/integration.md#cmake-integration

stevapple:
The case of `Foundation` is as follows:

https://github.com/apple/swift/blob/3003b0890509001b24525c0a2522177327958a45/docs/WindowsBuild.md?plain=1#L315-L333

PR #3146, opened 2022-02-27
Inject DLL paths on TestFoundation

stevapple:
Patch DLL paths to environment on building and running the tests.  This eliminates the need to modify `Path` manually and adds to robustness.


stevapple:
Anyone to review?

millenomi:
cc @compnerd ?

compnerd:
I'm not a fan of the use of `?` as a separator, and I would rather that we change that to something more inline with CMake's list operations.  More importantly, I think that this needs to be tested on CI to ensure that this doesn't accidentally cause any regressions.  Likely best to wait until after 5.7.

stevapple:
If we keep `;` as separator we may run into the hell of extra `//`s, and any additional parse may add an extra `//`, making it extremely hard to maintain (since Windows CI is not forced yet, I’m really pessimistic about not breaking this some day.

Taking another separator at least simplifies the implementation and adds to stability — and I picked `?` from forbidden characters in regular Windows path (I knew there’re `\\?\` paths, but I don’t think they’re working & worth being supported).

PR #3147, opened 2022-02-27
CoreFoundation: implement mapping from file descriptor to binary

compnerd:
This avoids an annoying warning in the Windows path that was introduced
with the last update to Foundation.  This implements the functionality
on Windows as well.


compnerd:
@swift-ci please test

stevapple:
I would rather suggest wrapping this in `#elif TARGET_OS_WIN32` to keep the warning for any other new platforms.

compnerd:
@stevapple I don't think that is valuable, this is already under such a condition.

PR #3153, opened 2022-03-15
Fix IndexPath slicing crash

PatrickPijnappel:
(Re-open of accidentally closed #3152)

Subscripting of `IndexPath` using a range currently crashes when removing from the front if the result is a pair or single:
```
let path = IndexPath(indexes: [ 10, 20, 30, 40 ])
print(path[2...]) // Crash
```


PatrickPijnappel:
@swift-ci Please test

PR #3155, opened 2022-03-15
Update CMakeLists.txt

compnerd:
Bump discovery timeout to see if it helps.


stevapple:
This problem should go with #3146.

@compnerd Could you take some time to review that now?

compnerd:
Eh?  Why would that impact this?

stevapple:
With #3146 , `TestFoundation` no longer requires anything to be in the path.

DLL paths will be injected by CMake.

compnerd:
Right, I understand that.  I am not seeing the issue locally, only on the swift.org builder.

compnerd:
@swift-ci please test

PR #3157, opened 2022-03-17
Update CMakeLists.txt

compnerd:
Add an additional search for the threading library.  This allows us to have an imported library definition which is used to statically link the static library `libCoreFoundation.a`.


compnerd:
@swift-ci please test

PR #3158, opened 2022-03-19
Write string external representation in String.write(to: URL)

karwa:
`NSString._getExternalRepresentation()` was not returning the string's external representation 👀.

For some reason, CF decides to only include the BOM in the external representation for encodings using the machine's local endianness (e.g. `.utf16`, `.utf32`) but not if you specify an explicit endianness (e.g. `.utf32BigEndian`). It does the same thing on Darwin Foundation -- I don't know why, and I don't agree with it, and it isn't documented anywhere AFAICT, but there it is.

### DIFF
--- a/Sources/Foundation/NSString.swift
+++ b/Sources/Foundation/NSString.swift
@@ -1254,7 +1254,7 @@ extension NSString {
         let length = self.length
         var numBytes = 0
         let theRange = NSRange(location: 0, length: length)
-        if !getBytes(nil, maxLength: Int.max - 1, usedLength: &numBytes, encoding: enc, options: [], range: theRange, remaining: nil) {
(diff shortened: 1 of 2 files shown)


karwa:
@millenomi Are you able to review this patch, or recommend somebody?

millenomi:

millenomi:
@swift-ci please test

karwa:
@millenomi Does this look good to go, then?

karwa:
@millenomi Not sure what happened to this. CI appears green but it wasn't merged? 🤷

parkera (commented):
@jmschonfeld does this look correct to you?

jmschonfeld (approved):
Yes this looks correct to me. The behavior here which (as you mentioned) is consistent with darwin platforms comes from the Unicode conformance spec which states that the explicit BE/LE encodings should not contain a BOM and the `<00 00 FE FF>` / `<FF FE 00 00>` byte sequences should be interpreted as U+FEFF Zero Width No-Break Space (this applies to both the UTF16 and UTF32 variants).

PR #3159, opened 2022-03-23
Foundation: adjust types for bit-width differences

compnerd:
This adjusts the types to be more explicit about types as appropriate.
These changes are required to enable building a 32-bit variant of
Foundation for Windows.  It also happens to clean up some constants
which were being formed oddly.  This enables building Windows x86
version of Foundation.


compnerd:
@swift-ci please test

millenomi:

PR #3164, opened 2022-04-01
CoreFoundation: correct alignment mis-assumption

compnerd:
The allocator on 32-bit platforms are not guaranteed to provide a
16-byte alignment.  This would cause an over-aligned request which would
be serviced by a different allocation path which needs to be paired with
an appropriate deallocation invocation.  Failure to do so resulted in a
heap corruption on Windows i686.


millenomi:
@swift-ci please test

millenomi:
Is there a way (even if non-portable?) to get the malloc alignment without having to hardcode it this way?

millenomi:
For posterity: no.

PR #4590, opened 2022-05-13
Mitigate an out-of-bounds memory read

glessard:
Mitigate an oob memory read issue due to pointer conversion, by
- adding overloads for `String.init?(cString:encoding:)` and `String.init?(utf8String:)`
- enforcing null-termination precondition when possible
- adding advisory deprecations for the obviously fixable cases
- clarifying the null-termination precondition in doc-comments

A similar issue was resolved in the standard library in https://github.com/apple/swift/pull/42002

addresses rdar://93255079

### DIFF
--- a/Sources/Foundation/NSStringAPI.swift
+++ b/Sources/Foundation/NSStringAPI.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2014 - 2018 Apple Inc. and the Swift project authors
+// Copyright (c) 2014 - 2022 Apple Inc. and the Swift project authors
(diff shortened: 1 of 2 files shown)


Sources/Foundation/NSStringAPI.swift
```diff
@@ -20,7 +20,7 @@
 // This file is shared between two projects:
 //
 // 1. https://github.com/apple/swift/tree/master/stdlib/public/Darwin/Foundation
```
glessard, line 22: This is no longer the case. What should we say instead?

glessard:
@swift-ci please test

glessard:
@swift-ci please test

glessard:
Squashed to a single commit…

glessard:
@swift-ci please test

PR #4592, opened 2022-05-15
Add comments to date types

quinwoods:
Just wanted add some additional documentation to the Date Formatter. Better than going online or having to implement it yourself to see what the result of the enum would be.


YOCKOW:
Welcome, first-time contribution.

I'm just a collaborator, but I think it is better to use documentation comments [adopted by SwiftDocC](https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files) and it is also better that the descriptions are the same with [Apple Developer Documentation](https://developer.apple.com/documentation/foundation/dateformatter/style).

quinwoods:
@YOCKOW You're so right! Will update that to match!

PR #4594, opened 2022-05-21
CoreFoundation: adjust `uuid.h` inclusion

compnerd:
The Swift Runtime path always uses the UUID functions from the
statically linked UUID library.  However, the header was only included
in the non-Windows case.  The latest LLVM update makes implicit
declarations invalid as per the C99 standard.  Adjust the inclusion path
to also include `uuid.h` on Windows when building with the Swift
Runtime.


compnerd:
CC: @bnbarham

compnerd:
@swift-ci please test

bnbarham (approved):
Seems reasonable to me given the current failure on rebranch. @millenomi in case you have any opinions

compnerd:

PR #4598, opened 2022-06-01
[Linux] Add missing glibc import for in_addr declaration

bnbarham:
Rebranch is currently failing with:
```
swift-corelibs-foundation/Sources/FoundationNetworking/HTTPCookie.swift:626:17: error: cannot find 'in_addr' in scope
        var x = in_addr()
                ^~~~~~~
```

Import `Glibc` to grab the declaration.

### DIFF
--- a/Sources/FoundationNetworking/HTTPCookie.swift
+++ b/Sources/FoundationNetworking/HTTPCookie.swift
@@ -13,7 +13,9 @@ import SwiftFoundation
 import Foundation
 #endif
 
-#if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd (approved):
Thanks!

bnbarham:
This isn't the issue - `Foundation` pulls in `Glibc` (and `SwiftGlibc`) so `in_addr` does exist, but it's also in `CoreFoundation`. The improper modularization is what's causing problems.

PR #4599, opened 2022-06-03
[Modularization] Only include cstd headers when building CoreFoundation

bnbarham:
Including these headers results in improper modularization as both
SwiftGlibc and CoreFoundation will include these headers. In particular,
`in_addr` from `netinet/in.h` is currently causing issues when building
on Linux in Swift rebranch.


bnbarham:
@swift-ci please test

compnerd (approved):
Thanks!  This is a good idea - Foundation already does an `@_exported import Glibc` so these should be provided for already via that module and be owned by that module.

Please to get @millenomi to take a look at this as well.

CC: @millenomi

xymus (approved):

PR #4600, opened 2022-06-06
Foundation: correct `Thread.callStackSymbols` on Windows

compnerd:
We would previously print invalid strings due to invalid memory
pointers.  This corrects that and takes the opportunity to simplify some
of the deallocation handling by using a temporary allocation.


compnerd:
@swift-ci please test

PR #4601, opened 2022-06-06
[5.7] Foundation: correct `Thread.callStackSymbols` on Windows

compnerd:
We would previously print invalid strings due to invalid memory usage.
The string name is not guaranteed to be nul-terminated, and the
`String(cString:)` initializer requires that guarantee.  Explicitly do
the alloc + init approach using `strncpy` to initialize the string in
place. This corrects that and takes the opportunity to simplify some of
the deallocation handling by using a temporary allocation.

### DIFF
--- a/Sources/Foundation/Thread.swift
+++ b/Sources/Foundation/Thread.swift
@@ -391,26 +391,31 @@ open class Thread : NSObject {
         return backtraceAddresses { (addresses, count) in
           var symbols: [String] = []
 
-          var buffer: UnsafeMutablePointer<Int8> =
-              UnsafeMutablePointer<Int8>
-                  .allocate(capacity: MemoryLayout<SYMBOL_INFO>.size + 128)
-          defer { buffer.deallocate() }
+          let addresses: UnsafeMutableBufferPointer<PVOID?> =
+              UnsafeMutableBufferPointer<PVOID?>(start: addresses, count: count)
+          withUnsafeTemporaryAllocation(byteCount: MemoryLayout<SYMBOL_INFO>.size + 127,
+                                        alignment: 8) { buffer in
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @millenomi @airspeedswift @parkera

Sources/Foundation/Thread.swift
```diff
@@ -391,26 +391,31 @@ open class Thread : NSObject {
...
-          buffer.withMemoryRebound(to: SYMBOL_INFO.self, capacity: 1) {
-            $0.pointee.SizeOfStruct = ULONG(MemoryLayout<SYMBOL_INFO>.size)
-            $0.pointee.MaxNameLen = 128
-
-            var address = addresses
-            for _ in 1...count {
               var dwDisplacement: DWORD64 = 0
-              if !SymFromAddr(hProcess, unsafeBitCast(address.pointee, to: DWORD64.self), &dwDisplacement, $0) {
-                symbols.append("\($0.pointee)")
+              if SymFromAddr(hProcess, DWORD64(UInt(bitPattern: address)),
```
millenomi, line 407: This may need to be revisited for 32-bit?

millenomi:
@swift-ci please test linux

Sources/Foundation/Thread.swift
compnerd: Oh, to use `UInt64` always ... yes, that is something that we will need to adjust.  I think that for 5.7 that is less critical as I doubt that we will get the last fixes for the 32-bit cleanliness in (it will require reworking SILGen).  Let's do that on main though.

millenomi:
@parkera This needs approval for merging into 5.7.

millenomi:
Per side discussion with @parkera: approved.

PR #4605, opened 2022-06-21
Connect `BUILD_TOOLS` option to build system

etcwilde:
The `BUILD_TOOLS` flag existed, but was never connected into the build
system, so it didn't do anything. Using it to determine whether or not
to dig down into the `Sources/Tools` directory.


etcwilde:
@swift-ci please test

compnerd (approved):
I wonder if this is better or if is better to simply remove the option (since no one noticed that it did nothing).  I think that reducing the options is nicer for maintenance purposes.

parkera:
Ok, thanks!

AtariDreams:
Any updates on this @parkera

PR #4609, opened 2022-06-28
Add `Date.now` static property

dannflor:
I've had builds that succeed on macOS fail on Linux because bizarrely, `Date.now` is only supported on macOS, despite it being equivalent to just calling `init()`. Easy fix.

### DIFF
--- a/Sources/Foundation/Date.swift
+++ b/Sources/Foundation/Date.swift
@@ -28,11 +28,7 @@ public struct Date : ReferenceConvertible, Comparable, Equatable {
     }
     
     /// Returns a `Date` initialized to the current date and time.
-    public static var now: Date {
-        get {
-            init()
-        }
-    }
+    public static var now: Date { Date() }
     
     /// Returns a `Date` initialized to the current date and time.
     public init() {


dannflor:
@swift-ci Please test

parkera (approved):

Sources/Foundation/Date.swift
```diff
@@ -27,6 +27,13 @@ public struct Date : ReferenceConvertible, Comparable, Equatable {
         return CFAbsoluteTimeGetCurrent()
     }
     
+    /// Returns a `Date` initialized to the current date and time.
+    public static var now: Date {
+        get {
+            init()
+        }
+    }
+
```
parkera, line 36: This also works and is perhaps a bit clearer:

```
public static var now : Date { Date() }
```

dannflor: Thanks. Can only members run CI?

parkera:
@swift-ci test

Sources/Foundation/Date.swift
parkera: Sorry I missed this comment from earlier -- yes, people with commit access can run CI.

PR #4610, opened 2022-06-28
Don't redefine asprintf when building for Android

etcwilde:
Bionic includes a definition and implementation of asprintf. We should
not include our own on Android.

See 'include/stdio.h' and 'libc/stdio/stdio.cpp'

https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/include/stdio.h#288
https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/stdio/stdio.cpp#736

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -1510,7 +1510,7 @@ void _CF_dispatch_once(dispatch_once_t *predicate, void (^block)(void)) {
 #pragma mark -
 #pragma mark Windows and Linux Helpers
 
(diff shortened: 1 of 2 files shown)


etcwilde:
@swift-ci please test

buttaface (approved):
LGTM, but maybe the similar #3021 should be merged instead?

etcwilde:
That works for me too.

etcwilde:
I should say that it works for me with caveats.
Digging through the macros. It's already defined if you have `__USE_GNU` enabled, or `STDC_WANT_LIB_EXT2` set, so we shouldn't re-declare it in that case. @compnerd will likely have more comments.

PR #4611, opened 2022-07-08
Foundation: rework handling for `ECANCELED`

compnerd:
Windows does not have a proper POSIX equivalent to `ECANCELED`.  The
closes equivalent appears to be `WSAECANCELLED` which is a WinSock error
code for a cancelled operation.  The Windows Error `ERROR_CANCELLED`
(0x04c7) is the ideal mapping even though the error code resides in a
different domain.  As long as the error code is compared via name, this
should be safe.


compnerd:
This is Windows specific and seems to have passed the cross-repository test.  I'm going to merge this for the time being as this can potentially help resolve some issues in SourceKit-LSP.

PR #4612, opened 2022-07-08
Define `asprintf` when `_GNU_SOURCE` is not set on Linux

etcwilde:
This PR will define `asprintf` when `_GNU_SOURCE` is not set while building for Linux. The build system defines `_GNU_SOURCE` when building for Linux/Android here; https://github.com/apple/swift-corelibs-foundation/blob/main/CoreFoundation/CMakeLists.txt#L69-L70, but being pedantically correct will ensure there are no surprises.


etcwilde:
@swift-ci please test

compnerd (approved):
Thanks!

etcwilde:
swift CI rebooted.

PR #4614, opened 2022-07-13
[5.7] backport some changes regarding `asprintf`

compnerd:
Pull a couple of patches from main which improve handling for asprintf on Linux platforms. asprintf is vended by the system libc for most platforms (including android) when _GNU_SOURCE is defined which we do set. Backporting this ensures that we do not accidentally break the build due to re-definition of the symbol.

These should be low risk for 5.7.  The changes impact only Linux and Android platforms, preventing a duplicate definition of `asprintf`.


compnerd:
CC: @parkera

parkera (approved):

PR #4615, opened 2022-07-15
Foundation: explicitly cast value on Windows for 32-bits

compnerd:
The `ERROR_CANCELLED` constant is imported as an `Int`, but `POSIXError` expects an `Int32`.  Explicitly cast the value to make the 32-bit builds clean.


compnerd:
@swift-ci please test

PR #4620, opened 2022-08-09
Fix typo `ecoding` -> `encoding`

philipturner:
This is present on Apple's developer website: https://developer.apple.com/documentation/swift/string/init(bytes:encoding:)

Screen Shot 2022-08-08 at 8 12 08 PM


philipturner:
Also, I found it in a strange commit from December 2020. It seems that commit never got merged:

https://github.com/apple/swift/commit/be13b470aae8d86163ed175cb476596d321f74f4

parkera (approved):

Issue #4220, opened 2021-02-28
[SR-14288] CFRunLoop assumes mach_absolute_time() is in TSR units


swift-ci:

3405691582:
Fixed in apple/swift-corelibs-foundation#3004.

3405691582:
Unfortunately, this fix may not be complete since we needed to roll the fix back. The conditions which demonstrated this bug no longer hold, unfortunately, so a proper fix may need to be revisited in future by future porters if needed.

PR #4621, opened 2022-08-09
Revert "Omnibus to implement CFRunLoop with kevent."

compnerd:
Reverts apple/swift-corelibs-foundation#3004

The change seems to have caused a regression on Windows between 9bc1a18236f643b73eb60f9633f2687e45dab487...5f57e702b8aaf6136374713967329712071b9eb5

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -249,11 +249,7 @@ const char *_CFProcessPath(void) {
 
 #if TARGET_OS_MAC || TARGET_OS_WIN32 || TARGET_OS_BSD
 CF_CROSS_PLATFORM_EXPORT Boolean _CFIsMainThread(void) {
-#if defined(__OpenBSD__)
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

3405691582:
I'd suggest seeing if setting __CFNanosecondsToTSR to just `return ns` solves the regression on Windows, as the base code is problematic as written anyway (see https://github.com/apple/swift-corelibs-foundation/issues/4220).

3405691582:
Actually, since `__CFTSRRate = 1.0e7` on Windows, it probably should be something like:
```
static inline uint64_t __CFNanosecondsToTSR(uint64_t ns) {
#if TARGET_OS_MAC || TARGET_OS_LINUX
    return ns;
#elif TARGET_OS_WINDOWS
    CFTimeInterval ti = ns / 1.0E7;
    return __CFTimeIntervalToTSR(ti);
#else
    CFTimeInterval ti = ns / 1.0E9;
    return __CFTimeIntervalToTSR(ti);
#endif
}
```
...perhaps those constants should be replaced with `__CFTSRRate`...
```

PR #4622, opened 2022-08-09
Attempt to unbreak test timeouts in Windows.

3405691582:
On Windows, __CFTSRRate is in units of 100 ns. To get a CFTimeInterval
from mach_absolute_time therefore, we should divide by 1.0E7, not 1.0E9.
But this suggests we should be dividing by __CFTSRRate anyway.

(Currently trying to test this now, so making this a draft pr for now to share since #3004 may be causing trouble for Windows.)

### DIFF
--- a/CoreFoundation/RunLoop.subproj/CFRunLoop.c
+++ b/CoreFoundation/RunLoop.subproj/CFRunLoop.c
@@ -48,15 +48,6 @@ extern void objc_terminate(void);
 #define USE_DISPATCH_SOURCE_FOR_TIMERS 0
 #endif
 
-static inline uint64_t __CFNanosecondsToTSR(uint64_t ns) {
-#if TARGET_OS_MAC || TARGET_OS_LINUX
-    return ns;
-#else
-    CFTimeInterval ti = ns / 1.0E9;
-    return __CFTimeIntervalToTSR(ti);
-#endif
-}
-
 #if USE_DISPATCH_SOURCE_FOR_TIMERS
 #if !TARGET_OS_MAC
 typedef uint32_t mach_port_t;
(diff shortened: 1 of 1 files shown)


3405691582:
cc @compnerd

3405691582:
In fact, I don't actually think doing this is correct, I think just dividing by the various units is more correct. The kicker is, of course, `clock_getres` is no longer returning funny clock resolutions on OpenBSD 7.1 and so I can't strictly reproduce the problem. Instead, I think a separate Windows branch that divides by 1.0E7 as a bare constant might make more sense.

3405691582:
What's still confusing me is that while the clock resolution on Windows is different, "fixing" the incommensurability of TSR units and nanoseconds actually broke Windows. To try to avoid spending too much time loading this stuff back into mental cache to try and untangle this, the safest course of action may be to just revert the attempt _just_ to fix https://github.com/apple/swift-corelibs-foundation/issues/4220. 

This shouldn't bite OpenBSD now clock_getres is returning 1 like `TARGET_OS_LINUX` and should probably fix Windows since it was working before. 

@compnerd, please try picking this and seeing if the test timeout troubles you no longer.

compnerd:
Triggered a build, should hopefully know in 90m.

compnerd:
Okay, this does seem to restore the Windows builds.

compnerd:
@swift-ci please test

3405691582:
Just finally confirmed the Foundation RunLoop tests pass on OpenBSD, so this is good to go when ready:
```
Test Suite 'Selected tests' started at 2022-08-09 23:31:52.838
Test Suite 'TestRunLoop' started at 2022-08-09 23:31:52.844
Test Case 'TestRunLoop.test_constants' started at 2022-08-09 23:31:52.844
Test Case 'TestRunLoop.test_constants' passed (0.0 seconds)
Test Case 'TestRunLoop.test_runLoopInit' started at 2022-08-09 23:31:52.844
Test Case 'TestRunLoop.test_runLoopInit' passed (0.001 seconds)
Test Case 'TestRunLoop.test_commonModes' started at 2022-08-09 23:31:52.845
Test Case 'TestRunLoop.test_commonModes' passed (0.004 seconds)
Test Case 'TestRunLoop.test_runLoopRunMode' started at 2022-08-09 23:31:52.849
Test Case 'TestRunLoop.test_runLoopRunMode' passed (0.05 seconds)
Test Case 'TestRunLoop.test_runLoopLimitDate' started at 2022-08-09 23:31:52.900
Test Case 'TestRunLoop.test_runLoopLimitDate' passed (0.0 seconds)
Test Case 'TestRunLoop.test_runLoopPoll' started at 2022-08-09 23:31:52.900
Test Case 'TestRunLoop.test_runLoopPoll' passed (0.0 seconds)
Test Case 'TestRunLoop.test_addingRemovingPorts' started at 2022-08-09 23:31:52.900
Test Case 'TestRunLoop.test_addingRemovingPorts' passed (0.003 seconds)
Test Case 'TestRunLoop.test_mainDispatchQueueCallout' started at 2022-08-09 23:31:52.903
Test Case 'TestRunLoop.test_mainDispatchQueueCallout' passed (2.001 seconds)
Test Suite 'TestRunLoop' passed at 2022-08-09 23:31:54.904
         Executed 8 tests, with 0 failures (0 unexpected) in 2.059 (2.059) seconds
Test Suite 'Selected tests' passed at 2022-08-09 23:31:54.904
         Executed 8 tests, with 0 failures (0 unexpected) in 2.059 (2.059) seconds
```

compnerd:
Thanks!

PR #4628, opened 2022-08-30
[5.7] Write string external representation in String.write(to: URL)

karwa:
Cherry pick of #3158 

Would be nice if this could be included in the 5.7 release.

@jmschonfeld @parkera


parkera:
I think we could consider this for some dot update from 5.7.

shahmishal:
@swift-ci test

PR #4629, opened 2022-08-31
Add definition for _Bool in C++ mode

etcwilde:
`_Bool` is not defined when importing Foundation with C++ interop enabled
resulting in the missing type. This patch defines `_Bool` to `bool` when
compiling for C++.

I'm not entire sure how to test this. The test itself is fairly simple, it's just
```
import Foundation
```
with the invocation `swiftc -Xcc -std=c++17 -enable-experimental-cxx-interop`, but I don't know how to tell XCTest to do that so I would like pointers there.

rdar://99713881

### DIFF
--- a/CoreFoundation/Base.subproj/CFBase.h
+++ b/CoreFoundation/Base.subproj/CFBase.h
@@ -66,8 +66,9 @@
 
 #if defined(__GNUC__) || TARGET_OS_WIN32
 #include <stdint.h>
-#include <stdbool.h>
 #endif
+
+#include <stdbool.h>
 
 #if __BLOCKS__ && (TARGET_OS_OSX || TARGET_OS_IPHONE)
 #include <Block.h>
@@ -80,7 +81,7 @@
 #if !defined(__MACTYPES__)
 #if !defined(_OS_OSTYPES_H)
 #if DEPLOYMENT_RUNTIME_SWIFT
-    typedef _Bool                   Boolean;
+    typedef bool                    Boolean;
 #else
     typedef unsigned char           Boolean;
 #endif
--- a/CoreFoundation/Base.subproj/CFOverflow.h
+++ b/CoreFoundation/Base.subproj/CFOverflow.h
@@ -15,8 +15,8 @@
 #if __has_include(<os/overflow.h>)
 #include <os/overflow.h>
 #else
-    static _Bool __os_warn_unused(_Bool x) __attribute__((__warn_unused_result__));
-    static _Bool __os_warn_unused(_Bool x) { return x; }
+    static bool __os_warn_unused(bool x) __attribute__((__warn_unused_result__));
+    static bool __os_warn_unused(bool x) { return x; }
 
     #if __has_builtin(__builtin_add_overflow) && \
     __has_builtin(__builtin_sub_overflow) && \
--- a/CoreFoundation/Base.subproj/ForFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForFoundationOnly.h
@@ -84,7 +84,7 @@ typedef CF_ENUM(CFOptionFlags, _CFAllocatorHint) {
 #define NSISARGTYPE id _Nullable
 #else
 #define NSISARGTYPE void * _Nullable
-#define BOOL _Bool
+#define BOOL bool
 #endif
 
 CF_EXPORT BOOL _NSIsNSArray(NSISARGTYPE arg);
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -482,10 +482,10 @@ CF_CROSS_PLATFORM_EXPORT void __CFURLComponentsDeallocate(CFTypeRef cf);
 typedef struct {
     void *_Nonnull memory;
     size_t capacity;
-    _Bool onStack;
+    bool onStack;
 } _ConditionalAllocationBuffer;
 
-static inline _Bool _resizeConditionalAllocationBuffer(_ConditionalAllocationBuffer *_Nonnull buffer, size_t amt) {
+static inline bool _resizeConditionalAllocationBuffer(_ConditionalAllocationBuffer *_Nonnull buffer, size_t amt) {
 #if TARGET_OS_MAC
     size_t amount = malloc_good_size(amt);
 #else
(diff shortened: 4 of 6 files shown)


CoreFoundation/Base.subproj/CFBase.h
```diff
@@ -80,6 +80,11 @@
 #if !defined(__MACTYPES__)
 #if !defined(_OS_OSTYPES_H)
 #if DEPLOYMENT_RUNTIME_SWIFT
+
+#ifdef __cplusplus
+#define _Bool bool
+#endif
```
compnerd, line 86: I wonder if we want to use `stdbool.h` here as well as we do on the other platforms.

compnerd: Yeah, that seems like a better approach - using the standard to use the proper spelling irrespective of the language.

parkera:
I agree with @compnerd's approach.

CoreFoundation/Base.subproj/CFBase.h
ian-twilightcoder: Are there any other uses of `_Bool` in CF?

etcwilde: This is the list I have. I just need to find a cycle or two to go through and change them and run the tests.

```
./CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h:static inline _Bool _resizeConditionalAllocationBuffer(_ConditionalAllocationBuffer *_Nonnull buffer, size_t amt) {
./CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h:static inline _Bool _withStackOrHeapBuffer(size_t amount, void (__attribute__((noescape)) ^ _Nonnull applier)(_ConditionalAllocationBuffer *_Nonnull)) {
./CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h:static inline _Bool _withStackOrHeapBuffer(size_t amount, void (__attribute__((noescape)) ^ _Nonnull applier)(_ConditionalAllocationBuffer *_Nonnull)) {
./CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h:static inline _Bool _withStackOrHeapBufferWithResultInArguments(size_t amount, void (__attribute__((noescape)) ^ _Nonnull applier)(void *_Nonnull memory, size_t capacity, _Bool onStack)) {
./CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h:static _Bool const _CFHasRenameat2 = 1;
./CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h:static _Bool const _CFHasRenameat2 = 0;
./CoreFoundation/Base.subproj/CFBase.h:    typedef _Bool                   Boolean;
./CoreFoundation/Base.subproj/ForFoundationOnly.h:#define BOOL _Bool
./CoreFoundation/Base.subproj/CFOverflow.h:    static _Bool __os_warn_unused(_Bool x) __attribute__((__warn_unused_result__));
./CoreFoundation/Base.subproj/CFOverflow.h:    static _Bool __os_warn_unused(_Bool x) { return x; }
./Darwin/shims/FoundationOverlayShims.h:    _Bool onStack;
./Darwin/shims/FoundationOverlayShims.h:static inline _Bool _resizeConditionalAllocationBuffer(_ConditionalAllocationBuffer *_Nonnull buffer, size_t amt) {
./Darwin/shims/FoundationOverlayShims.h:static inline _Bool _withStackOrHeapBuffer(size_t amount, void (__attribute__((noescape)) ^ _Nonnull applier)(_ConditionalAllocationBuffer *_Nonnull)) {
./CoreFoundation/PlugIn.subproj/CFBundle_Grok.c:static inline _Bool _CF_isspace(int c) {
```

etcwilde:
Alright, I've gone though and replaced uses of `_Bool` with `bool` and included `stdbool.h`.

etcwilde:
@swift-ci please test

ian-twilightcoder (approved):

compnerd (approved):
Thanks!

compnerd:
ForFoundationOnly.h L89-102 needs adjustment.

etcwilde:
@swift-ci please test

etcwilde:
> ForFoundationOnly.h L89-102 needs adjustment.

Yeah, the issue is that `BOOL` isn't always `_Bool`, sometimes it's `unsigned char`. I just added `#define BOOL bool` here;

```C
#ifdef __OBJC__
#define NSISARGTYPE id _Nullable
#else
#define NSISARGTYPE void * _Nullable
#define BOOL bool
#endif
```

ian-twilightcoder:
> > ForFoundationOnly.h L89-102 needs adjustment.
> 
> Yeah, the issue is that `BOOL` isn't always `_Bool`, sometimes it's `unsigned char`. I just added `#define BOOL bool` here;
> 
> ```c
> #ifdef __OBJC__
> #define NSISARGTYPE id _Nullable
> #else
> #define NSISARGTYPE void * _Nullable
> #define BOOL bool
> #endif
> ```

That seems really sketchy, why does `BOOL` need to be defined at all in CF?

etcwilde:
> That seems really sketchy, why does BOOL need to be defined at all in CF?

I think it's for ObjC compatibility since we're also building this code without ObjC.
The `ForFoundationOnly.h` only defines `BOOL` when `__OBJC__` isn't defined.

```c
// Arguments to these are id, but this header is non-Objc
#ifdef __OBJC__
#define NSISARGTYPE id _Nullable
#else
#define NSISARGTYPE void * _Nullable
#define BOOL _Bool
#endif
```

etcwilde:
Yeah, the `BOOL` definition came in on the Monterey CoreFoundation sync-up; https://github.com/apple/swift-corelibs-foundation/commit/d8e8a8b92b3a8af8381a11155328c1bba1c6bd2c#diff-32adebf83b3fcbdf66de2d896086ed7702b8cb7083cedb9594f092a11f513618

ian-twilightcoder:
> > That seems really sketchy, why does BOOL need to be defined at all in CF?
> 
> I think it's for ObjC compatibility since we're also building this code without ObjC. The `ForFoundationOnly.h` only defines `BOOL` when `__OBJC__` isn't defined.
> 
> ```c
> // Arguments to these are id, but this header is non-Objc
> #ifdef __OBJC__
> #define NSISARGTYPE id _Nullable
> #else
> #define NSISARGTYPE void * _Nullable
> #define BOOL _Bool
> #endif
> ```

Including <objc/objc.h> would be a lot better, I think it's ok to do that from pure C. Declaring it yourself is likely to eventually cause module problems

etcwilde:
> Including <objc/objc.h> would be a lot better, I think it's ok to do that from pure C. Declaring it yourself is likely to eventually cause module problems

The definition was in the original code, I'm just trying to allow C++ to work, so I think that's maybe for another PR?
Windows and Linux come with `objc/objc.h` and it doesn't look like corelib-foundation is providing it. How are we distributing that header for those cases?

ian-twilightcoder:
> > Including <objc/objc.h> would be a lot better, I think it's ok to do that from pure C. Declaring it yourself is likely to eventually cause module problems
> 
> The definition was in the original code, I'm just trying to allow C++ to work, so I think that's maybe for another PR? Windows and Linux come with `objc/objc.h` and it doesn't look like corelib-foundation is providing it. How are we distributing that header for those cases?

Yeah true, not really related to this PR. I'm not sure what provides the objc runtime on Windows and Linux.

PR #4632, opened 2022-09-11
Remove CFErrorSetCallBackForDomain

AtariDreams:
Since it has been 5 years since iOS 11, we can safely remove the functions that were set to go away after 10.13/iOS 11


AtariDreams:
@parkera Could you please check this

AtariDreams:
@parkera Thoughts?

parkera:
We almost never actually remove things. If any app out there is referencing this (even though it is internal only), then we'd cause them to crash on launch.

PR #4637, opened 2022-09-29
Temporary fix for static linking on Linux

MaxDesiatov:
When using `@_implementationOnly`, the dependencies do not get emitted into the module files and static linking does not work.

Originally submitted by @drexin as https://github.com/apple/swift-corelibs-foundation/pull/3135.

There was another PR that was supposed to fix it, but apparently it currently doesn't work on Linux: https://github.com/apple/swift-corelibs-foundation/pull/3052

Repro test case added in https://github.com/apple/swift/pull/61372

### DIFF
--- a/Sources/FoundationNetworking/URLSession/BodySource.swift
+++ b/Sources/FoundationNetworking/URLSession/BodySource.swift
@@ -23,7 +23,13 @@ import Foundation
 #endif
 
 @_implementationOnly import CoreFoundation
+
+#if os(Linux)
+import CFURLSessionInterface
+#else
 @_implementationOnly import CFURLSessionInterface
+#endif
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci please test

compnerd (changes requested):
This makes `CFURLSessionInterface` "public" which seems like the wrong thing to do.  For the static linking, why not consume the `TARGET_OBJECTS` for the library into the archive for `Foundation`?  Statically linking should DCE the objects appropriately.

MaxDesiatov:
Closing this as we need a different fix for https://github.com/apple/swift-corelibs-foundation/issues/4644.

PR #4650, opened 2022-10-25
Check for invalid URLRequest header values

jrflat:
The URLRequest implementation previously allowed arbitrary CR and LF characters in header values. This change checks header values for invalid characters before setting or adding them to the request. It does allow for header line folding (CRLF followed by HTAB or SP), consistent with URLRequest on Darwin.


parkera:
@swift-ci test

Issue #4642, opened 2022-10-10
Encountering "Constant strings cannot be deallocated" in Linux foundation


dhoepfl:
**Describe the bug**
Using `Bundle.module.url(forResource:, withExtension:,subdirectory:)` results in error message "Foundation/NSCFString.swift:119: Fatal error: Constant strings cannot be deallocated"

**Steps To Reproduce**
1. Create a new package `mkdir swiftbug; cd swiftbug; swift package init`
2. In Package.swift, add a resource to the testTarget:
```swift
        .testTarget(
            name: "swiftbugTests",
            dependencies: ["swiftbug"],
            resources: [
               .copy("Resources")
            ]),
```
3. Create a resource: `mkdir Tests/swiftbugTests/Resources; touch Tests/swiftbugTests/Resources/dummy.png`
4. Modify the test in Tests/swiftbugTests/swiftbugTests.swift to match:
```swift
    func testExample() async throws {
        // This is an example of a functional test case.
        // Use XCTAssert and related functions to verify your tests produce the correct
        // results.
        XCTAssertEqual(swiftbug().text, "Hello, World!")

        _ = Bundle.module.url(forResource: "filename", withExtension: nil, subdirectory: "Resources")
    }
```
5. Run the test: `swift test`

**Expected behavior**
Tests should finish without an error.

**Actual result**
```
Building for debugging...
[20/20] Linking swiftbugPackageTests.xctest
Build complete! (4.82s)
Test Suite 'All tests' started at 2022-10-10 17:57:06.166
Test Suite 'debug.xctest' started at 2022-10-10 17:57:06.171
Test Suite 'swiftbugTests' started at 2022-10-10 17:57:06.171
Test Case 'swiftbugTests.testExample' started at 2022-10-10 17:57:06.172
Foundation/NSCFString.swift:119: Fatal error: Constant strings cannot be deallocated
error: Exited with signal code 4
```

**Environment (please fill out the following information)**
 - OS: Debian Linux
 - Self compiled swiftc: LLVM c205842748e1789, Swift 1f8a5d6045262bb

**Additional context**
Might be related to/regression of #3785

valerianb:
FYI, reported it back in 2021 https://github.com/apple/swift/issues/56730

dhoepfl:
On this day, one year ago, I submitted the pull request.
A few hours ago, it was merged.

[...]

PR #4653, opened 2022-11-03
Fixes issue #4642: Encountering "Constant strings cannot be deallocated" in Linux foundation

dhoepfl:
Fixes issue #4642: Encountering "Constant strings cannot be deallocated" in Linux foundation

The reported error was an CFSTR("") which is later released. While this should not be a problem, the Linux implementation of CFSTR does not ignore dealloc on constant strings. Fixed this by calling CFRetain on the constant string.

Strictly speaking this is only a workaround. Issue #1351 has some hints how this could be fixed but the workaround is used a over the code so I think it is okay to use it here, too.

I found the same problem in CFDateIntervalFormatter.c where it appeared in a error handling code path that should never be called. Fixed anyways.


MaxDesiatov:
@swift-ci please test

notcome:
What can we do to get this one merged? I am starting to play around Swift server side and hit this issue when trying to load a resource from the module Bundle.

parkera:
We can merge this. It does seem like a workaround, because plenty of other places CF will assume that CFSTR is constant.

PR #4656, opened 2022-11-14
Remove unnecessary use of renameat2.

### LINKED ISSUE #4655: FileManager.replaceItemAt(_:withItemAt:) doesn't appear to function
Here's a simple test file:

```swift
import Foundation

func main() {
	print("start")
	let temporaryFolder = FileManager.default
		   .temporaryDirectory
		   .appendingPathComponent(UUID().uuidString, isDirectory: true)
       try! FileManager.default.createDirectory(at: temporaryFolder, withIntermediateDirectories: true)
	print("making temp folder \(temporaryFolder)")

	try! FileManager.default.createDirectory(at: temporaryFolder, withIntermediateDirectories: true)
	print("mde folder \(temporaryFolder)")


	let identityCertificatePath = temporaryFolder.appendingPathComponent("identity.p12")
	let newCertificate = temporaryFolder.appendingPathComponent("identity_new.p12")

	print("copying in files")
	try! String("identity").write(toFile: identityCertificatePath.path, atomically: true, encoding: .utf8)
	try! String("new").write(toFile: newCertificate.path, atomically: true, encoding: .utf8)

	print("Copied in files \(identityCertificatePath) \(newCertificate)")
	try! FileManager.default.replaceItemAt(identityCertificatePath, withItemAt: newCertificate)

	print("done")
}

main()
```

When run in Ubuntu 20.04 with Swift 5.7, this prints the following output:

```
start
making temp folder file:///tmp/4B5A4D7C-4747-468F-B68D-1303C38D8B41/
mde folder file:///tmp/4B5A4D7C-4747-468F-B68D-1303C38D8B41/
copying in files
Copied in files file:///tmp/4B5A4D7C-4747-468F-B68D-1303C38D8B41/identity.p12 file:///tmp/4B5A4D7C-4747-468F-B68D-1303C38D8B41/identity_new.p12
test/test.swift:25: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=4 "The file doesn’t exist."
Current stack trace:
0    libswiftCore.so                    0x0000ffffbaacaf8c _swift_stdlib_reportFatalErrorInFile + 128
1    libswiftCore.so                    0x0000ffffba7ca958 <unavailable> + 1419608
2    libswiftCore.so                    0x0000ffffba7c96e0 _assertionFailure(_:_:file:line:flags:) + 304
3    libswiftCore.so                    0x0000ffffba81d190 Dictionary.init<A>(_:uniquingKeysWith:) + 0
4    test                               0x0000aaaab282268c <unavailable> + 9868
5    test                               0x0000aaaab2821ab0 <unavailable> + 6832
6    libc.so.6                          0x0000ffffba4ccd28 __libc_start_main + 232
7    test                               0x0000aaaab2821874 <unavailable> + 6260
Trace/breakpoint trap
```

On Ubuntu 18.04 with Swift 5.7 this produces:

```
start
making temp folder file:///tmp/E99A55EF-CC74-4008-AE06-51B589950BBB/
mde folder file:///tmp/E99A55EF-CC74-4008-AE06-51B589950BBB/
copying in files
Copied in files file:///tmp/E99A55EF-CC74-4008-AE06-51B589950BBB/identity.p12 file:///tmp/E99A55EF-CC74-4008-AE06-51B589950BBB/identity_new.p12
test/test.swift:25: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=512 "(null)"
Current stack trace:
0    libswiftCore.so                    0x00007f34de75fdd0 _swift_stdlib_reportFatalErrorInFile + 112
1    libswiftCore.so                    0x00007f34de42a91c <unavailable> + 1444124
2    libswiftCore.so                    0x00007f34de42a738 <unavailable> + 1443640
3    libswiftCore.so                    0x00007f34de429220 _assertionFailure(_:_:file:line:flags:) + 419
4    libswiftCore.so                    0x00007f34de47e1c0 swift_unexpectedError + 748
5    test                               0x00005634cad3e9ce <unavailable> + 10702
6    test                               0x00005634cad3da59 <unavailable> + 6745
7    libc.so.6                          0x00007f34ddefaba0 __libc_start_main + 231
8    test                               0x00005634cad3d79a <unavailable> + 6042
```

In both cases the test fails.

This appears to be for different reasons on both platforms. If we find a copy of `renameat2` then we will call `renameat2` with the flag `RENAME_EXCHANGE`. This is a strange choice: no other platform appears to implement `replaceItemAt(_:with:)` with the semantic of exchanging the two files, not even the other code paths in this file. If that's not present, we call `renameat` and then, if the `.usingNewMetadataOnly` option is _not_ present, we will attempt to set the attributes to reapply on the _old_ path, which `renameat` just deleted.

In general there appear to be a number of bugs in rename handling that have slipped through the net here. Given that the exchange semantic of `renameat2` is not maintained in any other code path or platform, I propose deleting that code altogether and falling back to regular old `renameat`, as well as fixing the issues in the implementation.

Lukasa:
On Linux, on some code paths, we call renameat2 with `RENAME_EXCHANGE` in the implementation of replaceItemAt(_:withItemAt:). This is a behaviour utterly unique to that code path: Darwin Foundation doesn't do this, and the non-renameat2 codepaths don't do it either.

As Darwin doesn't do this, we maximise compatibility and minimise complexity by just deleting the path and all supporting code altogether.

Resolves #4655.

### DIFF
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -643,21 +643,6 @@ _stat_with_btime(const char *filename, struct stat *buffer, struct timespec *bti
 }
 #endif // __NR_statx
 
-static unsigned int const _CF_renameat2_RENAME_EXCHANGE = 1 << 1;
-#ifdef SYS_renameat2
-static _Bool const _CFHasRenameat2 = 1;
-static inline int _CF_renameat2(int olddirfd, const char *_Nonnull oldpath,
-                                int newdirfd, const char *_Nonnull newpath, unsigned int flags) {
-    return syscall(SYS_renameat2, olddirfd, oldpath, newdirfd, newpath, flags);
-}
-#else
-static _Bool const _CFHasRenameat2 = 0;
-static inline int _CF_renameat2(int olddirfd, const char *_Nonnull oldpath,
-                                int newdirfd, const char *_Nonnull newpath, unsigned int flags) {
-    return ENOSYS;
-}
-#endif // __SYS_renameat2
-
 
 #endif // TARGET_OS_LINUX
 
--- a/Sources/Foundation/FileManager+POSIX.swift
+++ b/Sources/Foundation/FileManager+POSIX.swift
@@ -1275,16 +1275,6 @@ extension FileManager {
                     return newItemURL.withUnsafeFileSystemRepresentation { (newItemFS) -> Int32? in
                         if let originalFS = originalFS,
                            let newItemFS = newItemFS {
-
-                                #if os(Linux)
-                                if _CFHasRenameat2 && kernelSupportsRenameat2 {
-                                    if _CF_renameat2(AT_FDCWD, originalFS, AT_FDCWD, newItemFS, _CF_renameat2_RENAME_EXCHANGE) == 0 {
-                                        return nil
-                                    } else {
-                                        return errno
-                                    }
-                                }
-                                #endif
                                 if renameat(AT_FDCWD, originalFS, AT_FDCWD, newItemFS) == 0 {
                                     return nil
                                 } else {
--- a/Sources/Foundation/FileManager.swift
+++ b/Sources/Foundation/FileManager.swift
@@ -886,12 +886,6 @@ open class FileManager : NSObject {
         let requiredVersion = OperatingSystemVersion(majorVersion: 4, minorVersion: 11, patchVersion: 0)
         return ProcessInfo.processInfo.isOperatingSystemAtLeast(requiredVersion)
     }()
-
-    // renameat2() is only supported by Linux kernels >= 3.15
-    internal lazy var kernelSupportsRenameat2: Bool = {
-        let requiredVersion = OperatingSystemVersion(majorVersion: 3, minorVersion: 15, patchVersion: 0)
-        return ProcessInfo.processInfo.isOperatingSystemAtLeast(requiredVersion)
-    }()
 #endif
 
     internal func _compareFiles(withFileSystemRepresentation file1Rep: UnsafePointer<NativeFSRCharType>, andFileSystemRepresentation file2Rep: UnsafePointer<NativeFSRCharType>, size: Int64, bufSize: Int) -> Bool {


Lukasa:
@swift-ci test

Lukasa:
So I'm troubled that these tests pass, because it means they're failing to test the actual logic. My next step is to try to work out why they don't fail.

Lukasa:
I've extended the test with an explicit check that the file moved _from_ is actually gone.

Lukasa:
@swift-ci test

Lukasa:
Oh _dang_, I've worked out why the tests are passing: this code is _backwards_. [This function](https://developer.apple.com/documentation/foundation/filemanager/2293212-replaceitemat) is supposed to replace the item at `originalItemURL` with the one at `newItemURL`, removing `newItemURL`. The implementation calls `renameat` in this code: https://github.com/apple/swift-corelibs-foundation/blob/de5f7edc3040057717f7da0ba875df04d3dc3e9e/Sources/Foundation/FileManager%2BPOSIX.swift#L1288

Quoting [the manual page](https://man7.org/linux/man-pages/man2/rename.2.html):

> ```
> int renameat(int olddirfd, const char *oldpath,
>              int newdirfd, const char *newpath);
> ```

That is, this operation renames the file _from_ `oldpath` _to_ `newpath`.

The effect of the call site is that we are moving from `originalFS` (derived from `originalItemURL`) to `newFS` (derived from `newItemURL`). That's backwards to the logic, when we're supposed to move from `newItemURL` to `originalItemURL`. This mistake was masked by the confusing argument names in the original function, which meant that it wasn't entirely clear what the intended semantics were supposed to be.

The fallback path gets it right:

https://github.com/apple/swift-corelibs-foundation/blob/de5f7edc3040057717f7da0ba875df04d3dc3e9e/Sources/Foundation/FileManager%2BPOSIX.swift#L1329-L1334

Moving from `newItemURL` to `originalItemURL`. All the other code paths are backwards, but we never hit them in CI and so the test is also presumably backwards.

Lukasa:
Additionally, the Darwin code is both backwards and replicates the "exchange" mistake from renameat2.

Lukasa:
@swift-ci test

Lukasa:
It seems impossible that the tests are continuing to pass. This strongly suggests that the tests don't actually _exercise_ these code paths, but I don't know why yet.

Lukasa:
Awesome, I worked out why the tests pass: we haven't been running this test since 2019. https://github.com/apple/swift-corelibs-foundation/issues/3327

Lukasa:
@swift-ci test

Lukasa:
@swift-ci test

Lukasa:
@swift-ci test

Lukasa:
Great, ready for review @parkera

MaxDesiatov:
This has conflicts in `_CF_renameat2` definition now.

parkera:
Associated issue comment says this still reproduces on 6.0

WFT:
This still reproduces on nightly docker images & the current 6.2.4. After [some discussion on the forums](https://forums.swift.org/t/is-corelibs-foundation-still-accepting-fixes/85149) it seems like the proper thing to would still be to land this PR here because this API isn't in swift-foundation yet. 

Is there any interest in getting that going on this PR @parkera?

parkera:
We can still get that going but someone needs to pick up this PR and rebase it given the conflicts.

WFT:
Thanks @parkera  - I have made a new PR & fixed some additional test failures in case that makes things easier: https://github.com/swiftlang/swift-corelibs-foundation/pull/5443

PR #4657, opened 2022-11-15
Annotate CFPreferencesCopyMultiple returning nullable

valeriyvan:
Solves two static analyser issues.

Before:
Screenshot 2022-11-15 at 08 52 15

After fix:
Screenshot 2022-11-15 at 08 54 21

### DIFF
--- a/CoreFoundation/Preferences.subproj/CFPreferences.h
+++ b/CoreFoundation/Preferences.subproj/CFPreferences.h
@@ -93,7 +93,7 @@ keysToFetch that are not present in the returned dictionary
 are not present in the domain.  If keysToFetch is NULL, all
 keys are fetched. */
 CF_EXPORT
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci please test

valeriyvan:
merge?

parkera:
This would turn into a source breaking change for Swift users, if they exist, because the result would now be nullable right?

valeriyvan:
> This would turn into a source breaking change for Swift users, if they exist, because the result would now be nullable right?

Correct. It's source breaking change. We have here to choose what is preferred:
* leave everything as is accepting that not nullable function sometime will return nil causing crash;
* accept source breaking change.

parkera:
In general, we do not make source breaking changes. They are disruptive when people update Xcode to a newer SDK.

In other cases like this, we decided to change the behavior instead. For example, return something else like an empty dictionary. It's difficult to tell without much context if certain callers expect one or the other, so it may turn into an ABI break instead. One technique we sometimes use is to look at whatever source code is publicly available and see if we can figure out what would happen if we returned something different.

PR #4667, opened 2022-11-24
Implement SortComparators

ffried:
This adds the `SortComparator` protocol available in Darwin Foundation since macOS 12 (iOS & tvOS 15, watchOS 8), alongside the available implementations for it.

These implementations are all identical to the Darwin ones, with the exception of `SortDescriptor`. The latter is due to the differences in `NSSortDescriptor`. `SortDescriptor` isn't `Codable` (for the same reasons `NSSortDescriptor` doesn't conform to `NSCoding`/`NSSecureCoding`). Apart from that, the public API is the same, and the differences are mostly internal.

Also, some `Sendable` conformances are currently commented out, because the recent addition of `Sendable` to Foundation types hasn't made it into the repository yet.

This PR is still missing unit tests, but I'd like to get some first feedback on this.
Some things were implemented by reverse-engineering Darwin foundation and I'm not sure if it's all completely correct.


ffried:
Since this stuff is now available in swift-foundation, I'm closing this one.

PR #4668, opened 2022-11-26
Use restrict keyword for format and message strings

AtariDreams:
The standard C functions called have undefined behavior if restrict is not met, so we should add them to functions calling vsnprintf and other related functions just in case.


AtariDreams:
@drexin Thoughts on this?

PR #4669, opened 2022-11-26
Remove dead, uncalled function

AtariDreams:
I don't even think it is to be called externally either, so maybe we should remove this entirely


CodaFi (changes requested):
This function is called by GDB to implement their version of `po` https://github.com/bminor/binutils-gdb/blob/94355de7751579b0182bd5821a3223939054f5d7/gdb/objc-lang.c#L1176

PR #4670, opened 2022-11-28
Use Data in JSONParser instead of Array of bytes

ruihe774:
The Data given to JSONDecoder may be memory mapped and convert it to Array can be memory/time inefficient. This commit avoids such coversion and always use Data to represent JSON document in JSONParser.

### DIFF
--- a/Sources/Foundation/JSONSerialization.swift
+++ b/Sources/Foundation/JSONSerialization.swift
@@ -178,7 +178,7 @@ open class JSONSerialization : NSObject {
         return try _data(withJSONObject: value, options: opt, stream: false)
     }
 
-    private static func createParser(with data: Data) throws -> JSONParser {
+    internal static func createParser(with data: Data) throws -> JSONParser {
         let (encoding, advanceBy) = JSONSerialization.detectEncoding(data)
 
         if encoding == .utf8 {
--- a/Sources/Foundation/JSONSerialization.swift
+++ b/Sources/Foundation/JSONSerialization.swift
@@ -177,28 +177,31 @@ open class JSONSerialization : NSObject {
     open class func data(withJSONObject value: Any, options opt: WritingOptions = []) throws -> Data {
         return try _data(withJSONObject: value, options: opt, stream: false)
     }
+
+    private static func createParser(with data: Data) throws -> JSONParser {
+        let (encoding, advanceBy) = JSONSerialization.detectEncoding(data)
+
+        if encoding == .utf8 {
+            // we got utf8... happy path
+            var parser = JSONParser(bytes: data[advanceBy...])
+            return parser
+        }
+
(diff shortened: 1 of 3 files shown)


Sources/Foundation/JSONSerialization.swift
```diff
@@ -183,22 +183,22 @@ open class JSONSerialization : NSObject {
...
+                guard let utf8String = String(bytes: data[advanceBy...], encoding: encoding) else {
                     throw JSONError.cannotConvertInputDataToUTF8
                 }
-                
-                var parser = JSONParser(bytes: Array(utf8String.utf8))
+
+                var parser = JSONParser(bytes: utf8String.data(using: .utf8)!)
                 return try parser.parse()
-            }
+            }()
```
kperryua, line 201: This whole bit might be better suited as a separate function instead of a now lone anonymous closure.

Sources/Foundation/JSONSerialization.swift
```diff
@@ -183,22 +183,22 @@ open class JSONSerialization : NSObject {
...
-                
-                guard let utf8String = String(bytes: ptr[advanceBy..<ptr.count], encoding: encoding) else {
+
+                guard let utf8String = String(bytes: data[advanceBy...], encoding: encoding) else {
                     throw JSONError.cannotConvertInputDataToUTF8
                 }
-                
-                var parser = JSONParser(bytes: Array(utf8String.utf8))
+
+                var parser = JSONParser(bytes: utf8String.data(using: .utf8)!)
```
kperryua, line 199: The force-unwrap makes me wonder about the guarantees here (as they all should). Are we sure this can't fail in any way?

ruihe774:
Here in f9cedca I fixed another issue that `JSONDecoder.decode()` would not detect and convert encoding of the JSON data. Now it is consistent with Darwin Foundation.

Sources/Foundation/JSONSerialization.swift
```diff
@@ -306,7 +309,7 @@ open class JSONSerialization : NSObject {
 
 private extension JSONSerialization {
     /// Detect the encoding format of the NSData contents
-    static func detectEncoding(_ bytes: UnsafeRawBufferPointer) -> (String.Encoding, Int) {
+    static func detectEncoding(_ bytes: Data) -> (String.Encoding, Int) {
```
kperryua, line 312: The `bytes[0]`, etc. subscripts in this function are subtly incorrect for `Data`. A `Data` happens to be its own slice type, and if it's a slice, it can have a non-zero `startIndex`, which will result in a precondition failure for `bytes[0]`, etc.

kperryua: (This is handled properly in the `DocumentReader` implementation, which performs access bounded by `startIndex` and `endIndex`.)

ruihe774:

PR #4671, opened 2022-11-28
Avoid leaking in _NSPrintForDebugger

AtariDreams:
This prevents memory leaks by removing dead, leaking code, and by checking if such a conversion would work before allocating memory in the first place.


kperryua:
Hmmm… shouldn't `CFStringGetCStringPtr` and `CFStringGetBytes` be called with `desc` not `cf`? This code assumes that `cf` is a CFStringRef, when in reality it could be any CFType, right?

(This fact suggests that this is potentially entirely dead code)

AtariDreams:
It is not dead code as this is a library and lldb calls this function

kperryua:
Ostensibly yes, but given the incorrectness of the existing code (calling CFString functions with potentially non CFStrings) we probably want to prove that to ourselves first.

PR #4666, opened 2022-11-18
Ensure libdispatch is always initialized before CoreFoundation

drexin:
This has been causing async functions to not return when using static linking.

rdar://96642321

PR #4672, opened 2022-11-29
[5.7] Ensure libdispatch is always initialized before CoreFoundation

drexin:
Backporting https://github.com/apple/swift-corelibs-foundation/pull/4666

This has been causing async functions to not return when using static linking.

rdar://96642321


drexin:
@swift-ci test

PR #4674, opened 2022-12-06
Fix JSONParser with -Ounchecked

ruihe774:
JSONParser used `reader.read()` in `precondition()`, which would not be evaluated when built with `-Ounchecked`. This commit fixes this and makes the side effect of `read()` always happen.


kperryua (approved):

kperryua:
@swift-ci test

compnerd:
@swift-ci please test

ruihe774:

PR #4679, opened 2022-12-13
JSONParser: Make readBool & readNull throw proper error on unexpected char at EOF

ruihe774:
Previously `DocumentReader.readBool()` and `DocumentReader.readNull()` would throw `JSONError.unexpectedEndOfFile` if the last char in the JSON is an unexpected char. This could happen for JSON document containing a single bool or null with last char unexpected.
This commit fixes it and `unexpectedEscapedCharacter` will be thrown properly.

### DIFF
--- a/Sources/Foundation/JSONSerialization+Parser.swift
+++ b/Sources/Foundation/JSONSerialization+Parser.swift
@@ -270,29 +270,24 @@ extension JSONParser {
         mutating func readBool() throws -> Bool {
             switch self.read() {
             case UInt8(ascii: "t"):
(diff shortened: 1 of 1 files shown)


kperryua:
Thanks! Would you mind adding some tests that demonstrate the fixes for these cases?

ruihe774:
> Thanks! Would you mind adding some tests that demonstrate the fixes for these cases?

@kperryua Sorry but I have trouble writing and running tests on Ventura. Would you please help me to add some tests? Like testing for `"tru"` and `"truf"` using `decode(Bool.self, from:)`.

ruihe774:

PR #4681, opened 2022-12-17
Foundation: attempt a first pass symlink resolution

compnerd:
On Windows, rather than recursively expanding a path component by component, perform a first pass resolution at the given path.  This potentially influences the path resolution as the component traversal would possibly fail to resolve cross-drive references.

### DIFF
--- a/Sources/Foundation/NSURL.swift
+++ b/Sources/Foundation/NSURL.swift
@@ -916,7 +916,35 @@ extension NSURL {
 #endif
         }
 
-        
+#if os(Windows)
+        let hFile: HANDLE = absolutePath.withCString(encodedAs: UTF16.self) {
+          CreateFileW($0, GENERIC_READ,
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @parkera

compnerd:
@swift-ci please test

parkera (approved):
Do we need some kind of windows-specific unit test?

compnerd:
@parkera I think it could be nice to do that, but I didn't want to put in the effort if this code path is unlikely to survive the rewrite of Foundation.  This was something that I ran across when getting the SourceKit-LSP tests resolved, so it is definitely something we hit.  The one thing that I worry about a test case for this is that it will potentially require running with elevated privileges (i.e. as "root") to create the symbolic links to traverse.  However, if we have existing test cases which do that and are passing, we should be able to accommodate that.

compnerd:

parkera:
Ok, that makes sense. LGTM then and thank you.

compnerd:

PR #4683, opened 2023-01-05
Fix XML parsing in WASI environment

kellyo-cricut:
Final chunk of XML document was being skipped when parsing when in a WASI environment

### DIFF
--- a/Sources/FoundationXML/XMLParser.swift
+++ b/Sources/FoundationXML/XMLParser.swift
@@ -601,7 +601,7 @@ open class XMLParser : NSObject {
         var result = true
         var chunkStart = 0
         var chunkEnd = min(_chunkSize, data.count)
-        while result && chunkStart != chunkEnd {
+        while result && chunkStart < chunkEnd {
             let chunk = data[chunkStart..<chunkEnd]
             result = parseData(chunk)
             chunkStart = chunkEnd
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci please test

kateinoigakukun:
CC: @MaxDesiatov

kateinoigakukun:
@swift-ci please test

MaxDesiatov (approved):
Would be great to cover this with a test, but I'm not sure we have testing infrastructure for this. Otherwise seems legit.

MaxDesiatov:
@swift-ci test macOS

kellyo-cricut:
I can't quite tell from the output of the macOS action- Are there legitimate test failures I need to address?

MaxDesiatov:
The test failure seems unrelated, but I requested at least one more review just in case.

MaxDesiatov:

MaxDesiatov:

Sources/FoundationXML/XMLParser.swift
```diff
@@ -601,10 +601,7 @@ open class XMLParser : NSObject {
         var result = true
         var chunkStart = 0
         var chunkEnd = min(_chunkSize, data.count)
-        while result {
-            if chunkStart >= data.count || chunkEnd >= data.count {
-                break
-            }
+        while result && chunkStart != chunkEnd {
```
parkera, line 604: This looks safe based on reading the code below. One reason why I personally stick with < checks in places like this instead of != is that if a later change or some missed path advances beyond the end, we fail safely instead of entering an infinite loop.

PR #4692, opened 2023-01-19
Base.subproject: correct home directory handling on Windows

compnerd:
While we can enumerate the users using `NetUserEnum`, that does not guarantee that we get the full user information.  Perform an explicit `NetUserGetInfo` to query the user information to get the home directory.

Additionally, correct the `CFString` handling within the function where we were improperly using `CFStringCreateWithBytesNoCopy` which requires that the string is 8-bit encoded.  We now use
`CFStringCreateWithCharacters[NoCopy]` to properly create the `CFString` with unicode characters.


compnerd:
@swift-ci please test

compnerd:
CC: @parkera @lorentey

lorentey (approved):

compnerd:
@swift-ci please test

compnerd:

PR #4695, opened 2023-01-20
FileManager: repair `homeDirectoryForCurrentUser` on Windows

compnerd:
The `homeDirectory(forUser:)` is currently not functioning properly for users other than the current user.  Switch to directly using the CoreFoundation API which simplifies the logic and makes the functionality work on Windows.  This uses the forced unwrap for the return value, which was previously the case and resulted in a trap on Windows.

### DIFF
--- a/Sources/Foundation/FileManager.swift
+++ b/Sources/Foundation/FileManager.swift
@@ -1140,7 +1140,7 @@ open class FileManager : NSObject {
     }
 
     open var homeDirectoryForCurrentUser: URL {
-        return homeDirectory(forUser: NSUserName())!
+        CFCopyHomeDirectoryURLForUser(nil)!.takeRetainedValue()._swiftObject
     }
     
     open var temporaryDirectory: URL {
@@ -1149,8 +1149,13 @@ open class FileManager : NSObject {
     
     open func homeDirectory(forUser userName: String) -> URL? {
         guard !userName.isEmpty else { return nil }
-        guard let url = CFCopyHomeDirectoryURLForUser(userName._cfObject) else { return nil }
-        return  url.takeRetainedValue()._swiftObject
+        // Prefer to take the `CFCopyHomeDirectoryURLForUser` path for the
+        // current user.
+        return CFCopyHomeDirectoryURLForUser(userName == NSUserName()
+                                                ? nil
+                                                : userName._cfObject)?
+                    .takeRetainedValue()
+                    ._swiftObject
     }
 }
 


compnerd:
CC: @parkera

compnerd:
@swift-ci please test

etcwilde (commented):
I'd like to understand more about what is failing in `homeDirectory(forUser:)`. It looks like it's going through the same steps as your fix here, with the exception of checking that the username isn't an empty string, and using a guard on the user directory URL (any safety gained by that is immediately lost by the force-unwrap at the upper layer, so I don't consider that to be all that different). Is `NSUserName()` not working right for the current user?

https://github.com/apple/swift-corelibs-foundation/blob/ff83ef2d90e48d9bc3465b13d524ab40cd5f2356/Sources/Foundation/FileManager.swift#L1150-L1154

compnerd:
It isn't `homeDirectory(forUser:)` that is failing, it is that `NetUserGetInfo` will not provide a home directory for the user.

https://github.com/apple/swift-corelibs-foundation/blob/main/CoreFoundation/Base.subproj/CFPlatform.c#L594

By passing in the explicit `nil` for the username, we directly invoke `CFCopyHomeDirectoryURL` which will give you the home directory for the current user.

etcwilde:
We should probably fix `CFCopyHomeDirectoryURLForUser` so that it can properly handle the current username. `homeDirectory(forUser:)` and `homeDirectoryForCurrentUser` should return the same thing when the function is passed the current username (assuming the current user isn't a system account, in which case I expect the computed property to explode and the function to return `nil`).

For performance/inlining/fast-pathing, I'm fine with the direct `nil`-passing in the computed property and a special-case for `userName == GetCurrentUser()` for the function.

Sources/Foundation/FileManager.swift
```diff
@@ -1149,8 +1149,9 @@ open class FileManager : NSObject {
     
     open func homeDirectory(forUser userName: String) -> URL? {
         guard !userName.isEmpty else { return nil }
-        guard let url = CFCopyHomeDirectoryURLForUser(userName._cfObject) else { return nil }
-        return  url.takeRetainedValue()._swiftObject
+        // Prefer to take the `CFCopyHomeDirectoryURLForUser` path for the
+        // current user.
+        return CFCopyHomeDirectoryURLForUser(userName == NSUserName() ? nil : userName._cfObject)?.takeRetainedValue()._swiftObject
```
etcwilde, line 1154: This ternary is hurting my eyes. It's too hard to see where the parameter to `CFCopyHomeDirectoryURLForUser` ends and where the `?.takeRetainedValues()...` begins.

compnerd: Spoke with @etcwilde offline, reflowed text seems to address the concern.

compnerd:
@swift-ci please test

etcwilde (approved):
I'll accept it. It doesn't cause joy, but at least I can read it.

PR #4697, opened 2023-01-25
Base: resort to more drastic means to query profile directory

compnerd:
The profile directory for another user is considered private data and cannot be directly queried without an elevated privilege, the user's password, or a token with impersonation rights.  Resort to trawling through the registry to query the information.


compnerd:
CC: @parkera @etcwilde

compnerd:
> 100% tests passed, 0 tests failed out of 1916

compnerd:
@swift-ci please test

parkera (approved):

PR #4698, opened 2023-01-26
Base: correct TLS handling on Windows x86

compnerd:
The FLS destructors are required to be `__stdcall` calling convention rather than `__cdecl`.  Correct the definitions for them to repair the build on Windows x86 after a newer compiler identified this issue.  This attribute is ignored on non-x86 architectures and thus is safe to apply unconditionally.


compnerd:
@swift-ci please test

PR #4699, opened 2023-01-27
Fix incompatible function pointer type

bnbarham:
libxml2 added `const` to `xmlHashScanner` in v2.9.8. Wrap the call in a version check and cast for the non-const case.


bnbarham:
@shahmishal the other option is to update libxml2 on whatever CI jobs are currently failing with this. Not sure if that's an option, so have put this up just in case.

bnbarham:
@swift-ci please test

bnbarham:
@compnerd and @etcwilde if you have time 🙏

compnerd (approved):
This seems like the best approach here.  I can just be grumpy at libxml2 changing signatures though right?

compnerd:
I think that we should go ahead and merge this for now.  If we update the CI builders, we can remove this, it will be a bit of technical debt that we are carrying around, but it won't break anything.

PR #4705, opened 2023-02-18
Add `data(from:delegate:)` method.

b1ackturtle:
Related issue: https://github.com/apple/swift-corelibs-foundation/issues/3205

### DIFF
--- a/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
+++ b/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
@@ -28,7 +28,7 @@ private class Bag<Element> {
 
 /// A cancelable object that refers to the lifetime
 /// of processing a given request.
-open class URLSessionTask : NSObject, NSCopying {
+open class URLSessionTask : NSObject, NSCopying, @unchecked Sendable {
     
     // These properties aren't heeded in swift-corelibs-foundation, but we may heed them in the future. They exist for source compatibility.
     open var countOfBytesClientExpectsToReceive: Int64 = NSURLSessionTransferSizeUnknown {
@@ -1058,7 +1058,7 @@ extension _ProtocolClient : URLProtocolClient {
                     webSocketDelegate.urlSession(session, webSocketTask: webSocketTask, didOpenWithProtocol: webSocketTask.protocolPicked)
                 }
             }
-        case .noDelegate, .dataCompletionHandler, .downloadCompletionHandler:
+        case .noDelegate, .dataCompletionHandler, .dataCompletionHandlerWithTaskDelegate, .downloadCompletionHandler:
             break
         }
     }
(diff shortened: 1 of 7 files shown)


b1ackturtle:
@swift-ci Please test

b1ackturtle:
@parkera
Could you please give me your feedback?

b1ackturtle:
@jrflat 
Could you please review?

helbertgs (approved):

MaxDesiatov:
@swift-ci test

Sources/FoundationNetworking/URLSession/URLSessionTask.swift
```diff
@@ -28,7 +28,7 @@ private class Bag<Element> {
 
 /// A cancelable object that refers to the lifetime
 /// of processing a given request.
-open class URLSessionTask : NSObject, NSCopying {
+open class URLSessionTask : NSObject, NSCopying, @unchecked Sendable {
```
MaxDesiatov, line 31: Has this class been audited for thread-safety to get this `Sendable` annotation?

b1ackturtle: In Foundation on Apple platforms, `URLSessionTask` has already conformed to `Sendable`. So, I naturally assume that the `URLSessionTask` in Core Foundation could also conform to `Sendable`. Do you think any detailed audits are necessary?

If we need additional audits, I'd like to contribute to it. However, I'm not sure how to audit for thread-safety of this class. Could you please give me your advice?

MaxDesiatov: > In Foundation on Apple platforms, `URLSessionTask` has already conformed to `Sendable`. So, I naturally assume that the `URLSessionTask` in Core Foundation could also conform to `Sendable`. 

`URLSessionTask` is implemented in Foundation proper, it's not present in CoreFoundation. You can confirm that by searching [the `CoreFoundation` directory](https://github.com/apple/swift-corelibs-foundation/tree/main/CoreFoundation).

> Do you think any detailed audits are necessary?

Yes, of course. If something is marked as `Sendable` in Foundation on Darwin platforms, it means that an audit was done and necessary changes that make it thread-safe applied if needed on it. Since these are different codebases, it has to be done separately in this repository too.

> If we need additional audits, I'd like to contribute to it. However, I'm not sure how to audit for thread-safety of this class. Could you please give me your advice?

I don't think there's anything specific to this repository when compared to a general thread-safety audit on an arbitrary codebase. As a bare minimum one would have to confirm that mutable state accessible from different threads is either synchronized or made immutable. If there's a class hierarchy involved, this has to be done to mutable state inherited from superclasses too. In Swift I'd probably start with making a type `Sendable` (not `@unchecked Sendable`) and then auditing and adding matching sendability annotations to types involved in those warnings.

jrflat:
@b1ackturtle Sorry for the slow review here, thanks for your work on this! I built on top of your commit in #4970 to add the remaining async methods and address the Sendability requirement without marking `URLSessionTask` as `@unchecked Sendable`. Any feedback is welcome. Thanks again!

b1ackturtle:
@jrflat 
Thank you for your review.
I also appreciate your additional commits to address the `@unchecked Sendable` conformance.
I hadn't thought of this solution when creating the PR, so your approach has been very educational for me.

DougGregor:
Should we close this PR now that #4970 has landed?

PR #4706, opened 2023-02-19
Make `ISO8601DateFormatter.Options` static properties immutable

### LINKED ISSUE #4703: `ISO8601DateFormatter.Options` should have immutable static properties
[`ISO8601DateFormatter.Options`](https://github.com/apple/swift-corelibs-foundation/blob/7504fdfa4529ac01b77ab83bfee51a5f06e3e6d1/Sources/Foundation/ISO8601DateFormatter.swift#L14) uses `var` for its static properties. The properties should instead be declared using `let` to prevent accidental mutation and to let the compiler know that these properties are concurrency-safe.

Regarding the latter point, the following is an example error caused by this issue:
> error: reference to static property 'withInternetDateTime' is not concurrency-safe because it involves shared mutable state

fumoboy007:
Fixes #4703.

[`ISO8601DateFormatter.Options`](https://github.com/apple/swift-corelibs-foundation/blob/7504fdfa4529ac01b77ab83bfee51a5f06e3e6d1/Sources/Foundation/ISO8601DateFormatter.swift#L14) currently uses `var` for its static properties. The properties should instead be declared using `let` to prevent accidental mutation and to let the compiler know that these properties are concurrency-safe.

Regarding the latter point, the following is an example error caused by this issue:
> error: reference to static property 'withInternetDateTime' is not concurrency-safe because it involves shared mutable state

### DIFF
--- a/Sources/Foundation/ISO8601DateFormatter.swift
+++ b/Sources/Foundation/ISO8601DateFormatter.swift
@@ -17,33 +17,33 @@ extension ISO8601DateFormatter {
         
         public init(rawValue: UInt) { self.rawValue = rawValue }
 
-        public static var withYear = ISO8601DateFormatter.Options(rawValue: 1 << 0)
+        public static let withYear = ISO8601DateFormatter.Options(rawValue: 1 << 0)
 
-        public static var withMonth = ISO8601DateFormatter.Options(rawValue: 1 << 1)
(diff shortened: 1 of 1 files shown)


fumoboy007:
@parkera since you merged #998?

YOCKOW:
@swift-ci Please test

YOCKOW:
@swift-ci Please test Windows platform

PR #4714, opened 2023-03-09
Honor CMAKE_INSTALL_BINDIR

tristanlabelle:
CMake install places binary files under the `CMAKE_INSTALL_BINDIR` subdirectory, but some CMake files hardcode `bin` instead (the default value). This change un-hardcodes these. This has no effect by default, but it will allow redirecting the binaries to a different output directory, which will help with packaging installers for multiple architectures in the Windows build.


compnerd (approved):
https://github.com/apple/swift-corelibs-foundation/blob/main/CMakeLists.txt#L53 includes `GNUInstallDirs` so we should be fine.

compnerd:
@swift-ci please test

compnerd:

tristanlabelle:
The Linux CI build is still hitting the seemingly unrelated:
```
/home/build-user/swift-tools-support-core/Sources/TSCBasic/ByteString.swift:11:8: error: no such module 'Foundation'
import Foundation
       ^

compnerd:

tristanlabelle:
Ah! It looks like in the linux build, `CMAKE_INSTALL_LIBDIR` expands to `lib/x86_64-linux-gnu`

tristanlabelle:
The latest change scopes down to only replacing `bin` with `CMAKE_INSTALL_BINDIR` and leaving lib alone given the variation of the linux build and that for the Windows installer purposes, we only need to override the bin dir. @compnerd , could you re-kick the CI?

PR #4694, opened 2023-01-20
Replace calls to CFStringCreateWithCString with constant CFStrings

AtariDreams:
This would save performance by removing unnecessary calls to CFStringCreateWithCString

PR #4715, opened 2023-03-16
Revert "Replace calls to CFStringCreateWithCString with constant CFStrings"

compnerd:
Reverts apple/swift-corelibs-foundation#4694

### DIFF
--- a/CoreFoundation/Base.subproj/CFPlatform.c
+++ b/CoreFoundation/Base.subproj/CFPlatform.c
@@ -278,11 +278,8 @@ CF_PRIVATE CFStringRef _CFProcessNameString(void) {
     static CFStringRef __CFProcessNameString = NULL;
     if (!__CFProcessNameString) {
         const char *processName = *_CFGetProgname();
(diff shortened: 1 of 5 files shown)


compnerd:
@swift-ci please test

compnerd:
Verified locally, this does repair the regression.  CC: @jckarter

`s10Foundation19_NSCFConstantStringCN` => "type metadata for Foundation._NSCFConstantString"

`_NSCFConstantString` is defined as `internal final`: https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/NSCFString.swift#L62

This means that the type metadata for that is not exported, and there are now references to that in FoundationXML which cannot be fulfilled as the symbol is not visible outside of Foundation.  This is happening as the `CFXMLInterface.c` file was modified to use `CFSTR("")` which generates the `CFString` literal which will require the reference to the type metadata.  I don't think that this is worth making `_NSCFConstantString` public.

CC: @parkera

shahmishal:
@swift-ci test

PR #4718, opened 2023-03-17
Memory optimizations

LumioseSil:
We can replace bzero with memset, and malloc + memset with calloc.

In addition, cases where the size and number arguments of calloc were swapped were corrected.


AtariDreams:
@compnerd Thoughts on this?

AtariDreams:
@compnerd Thoughts on this?

PR #4721, opened 2023-03-21
Update GettingStarted.md so GitHub's 'copy' method works.

JoeMatt:
## What

Use proper shell markdown syntax so github's copy code block function works.

## Why

So copy/pasting into the shell doesn't require an edit first.

## How

1. Changed  '\`\`\`' to ' \`\`\`sh'
2. Removed leading shell '%' symbol.

## Discussion

If this isn't against any style guides, I'd be willing to fix any other docs. I have the regex saved to clear it up globally.


compnerd (approved):

compnerd:
@swift-ci please test

PR #4740, opened 2023-05-05
waitUntilExit: ignore RunLoop.run()'s return value

fischman-bcny:
We've been seeing frequent (but not reproducible in isolation) failures in our builds where waitUntilExit returns but terminationStatus fails its precondition on hasFinished because isRunning is still true. This is a speculative fix, since I haven't been able to reproduce the failure in a self-contained test.

This should be safe since Process.waitUntilExit isn't spec'd to relate to RunLoop. In particular there is no guarantee that "stopping" a RunLoop should cause waitUntilExit to return before the monitored process has exited.

---

Unrelated to the above, the [second commit](https://github.com/apple/swift-corelibs-foundation/pull/4740/commits/5404e239d28e1ab32d7b58b5338980620cb5ea06) in this PR is required to unbreak XCTest's tests, which are run as part of this repo's CI (and so is required to be able to get a green run for this PR).
Absent that commit, XCTest's tests exit -5 on my M1 Pro and -4 on CI (with no output). On my machine, Console reveals:
```
BUG IN CLIENT OF LIBPTHREAD: workgroup functions already installed
Abort Cause 8313026928
```
and running under lldb blames `CFRuntime.c:1185`'s call to `libdispatch_init()`.
Breaking on `libdispatch_init` in lldb confirms `libdispatch_init()` is called twice, first by `swift-corelibs-libdispatch/src/init.c` in an `__attribute__((constructor))` and the second in the site being disabled in this commit.

### DIFF
--- a/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/CoreFoundation/Base.subproj/CFRuntime.c
@@ -1179,7 +1179,7 @@ void __CFInitialize(void) {
     if (!__CFInitialized && !__CFInitializing) {
         __CFInitializing = 1;
 
-#if __HAS_DISPATCH__
+#if __HAS_DISPATCH__ && !TARGET_OS_MAC
     // libdispatch has to be initialized before CoreFoundation, so to avoid
     // issues with static initializer ordering, we are doing it explicitly.
     libdispatch_init();
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -1134,14 +1134,17 @@ open class Process: NSObject {
     open func waitUntilExit() {
         let runInterval = 0.05
         let currentRunLoop = RunLoop.current
-        let checkRunLoop : () -> Bool = (currentRunLoop == self.runLoop)
-                ? { currentRunLoop.run(mode: .default, before: Date(timeIntervalSinceNow: runInterval)) }
-                : { currentRunLoop.run(until: Date(timeIntervalSinceNow: runInterval)); return true }
 
-        // update .runLoop to allow early wakeup.
+        let runRunLoop : () -> Void = (currentRunLoop == self.runLoop)
+                ? { currentRunLoop.run(mode: .default, before: Date(timeIntervalSinceNow: runInterval)) }
+                : { currentRunLoop.run(until: Date(timeIntervalSinceNow: runInterval)) }
+        // update .runLoop to allow early wakeup triggered by terminateRunLoop.
         self.runLoop = currentRunLoop
-        while self.isRunning && checkRunLoop() {}
-        
+
+        while self.isRunning {
+            runRunLoop()
+        } 
+
         self.runLoop = nil
         self.runLoopSource = nil
     }


compnerd:
@swift-ci please test

fischman-bcny:
The reason I believe this PR might help the observed issue is that it's possible to make RunLoop.run(.default) return false with sufficiently esoteric setup, such as:

```swift
import Foundation

let waiter = DispatchGroup()
waiter.enter()
let t = Thread {
    defer { waiter.leave() }
    let runLoop = RunLoop.current
    let customMode = RunLoop.Mode("yo")
    runLoop.add(Timer(timeInterval: 0.1, repeats: false) { _ in
        let innerRunResult = runLoop.run(mode: .default, before: Date(timeIntervalSinceNow: 0.1))
        print("inner runLoop.run(.default) returned \(innerRunResult)")
    }, forMode: customMode)

    let outerRunResult = runLoop.run(mode: customMode, before: Date(timeIntervalSinceNow: 1))
    print("outer runLoop.run(custom) returned: \(outerRunResult)")
}
t.start()
waiter.wait()
```
which emits:
```
inner runLoop.run(.default) returned false
outer runLoop.run(custom) returned: true
```

That `inner runLoop.run(.default) returned false` is what I suspect is similar to what's happening inside `Process.waitUntilExit`'s call to `.run(mode: .default, ...)` and triggering the too-early return.

compnerd:
CC: @parkera @iCharlesHu @pcbeard

compnerd:

compnerd:
@swift-ci please test

compnerd:

CoreFoundation/Base.subproj/CFRuntime.c
```diff
@@ -1179,7 +1179,7 @@ void __CFInitialize(void) {
     if (!__CFInitialized && !__CFInitializing) {
         __CFInitializing = 1;
 
-#if __HAS_DISPATCH__
+#if __HAS_DISPATCH__ && !TARGET_OS_MAC
```
compnerd, line 1182: Ideally we would use the deployment version - but this is still forwards/backwards incompatible.  @parkera - any thoughts?  I'm tempted to say we merge this even with that limitation as this is blocking PRs for other platforms.

Sources/Foundation/Process.swift
```diff
@@ -1134,14 +1134,17 @@ open class Process: NSObject {
     open func waitUntilExit() {
         let runInterval = 0.05
         let currentRunLoop = RunLoop.current
-        let checkRunLoop : () -> Bool = (currentRunLoop == self.runLoop)
-                ? { currentRunLoop.run(mode: .default, before: Date(timeIntervalSinceNow: runInterval)) }
-                : { currentRunLoop.run(until: Date(timeIntervalSinceNow: runInterval)); return true }
 
-        // update .runLoop to allow early wakeup.
+        let runRunLoop : () -> Void = (currentRunLoop == self.runLoop)
+                ? { currentRunLoop.run(mode: .default, before: Date(timeIntervalSinceNow: runInterval)) }
```
compnerd, line 1139: The current run loop may be configured to return upon a source process, and it could be returning `kCFRunLoopRunHandledSource`, terminating the loop before the process completed.  We would now simply pump the runloop while we believe the child is executing, which I believe was the original intention.

compnerd (approved):

CoreFoundation/Base.subproj/CFRuntime.c
parkera: This code is designed to go with the corresponding libdispatch, so if we are making a change to both (or a change to dispatch has already been made), they would go together into a Swift release.

parkera:
LGTM, thanks @fischman-bcny and @compnerd

oO0oO0oO0o0o00:
Hello may I ask when will this fix be included in a release (may be a little off-topic)? The fixed code is present in branch release/5.9, but not in branches release/5.9.0 nor release/5.9.1, meaning it's not fixed in swift-5.9 and swift-5.9.1 binary. What about swift-5.9.2? It has been released but there's no branch release/5.9.2.

PR #4744, opened 2023-05-29
Foundation: implement `Bundle.init(for:)` on Windows

compnerd:
This was disabled due to the use of `dlsym` which has been removed in favour of CoreFoundation APIs.  Enable this code path and tweak the test to allow passing on Windows.  This function is required to support apple/swift-docc on Windows.


compnerd:
@swift-ci please test

PR #4746, opened 2023-06-04
Foundation: support file system representation for long paths

compnerd:
This adjusts the file system representation accessor for a URL to support paths beyond `MAX_PATH`.  We do not know what the use of the file system representation is, so we cannot provide the NT style path (`\\?\`-prefixed).  Rather than using the filled in buffer, resolve the URL and then build the complete path copying that into a newly allocated buffer.  This ensures that we always have sufficient space for the path without having to grow the buffer.

Linux gets away with the current implementation by setting the `PATH_MAX` to page size (4096).  There may be some value in using the single implementation across the different platforms, but this will allow the Windows path to make progress even if it diverges from the reference implementation.

The issue was identified by the DocC test suite which quickly exceeds the length.

### DIFF
--- a/Sources/Foundation/NSURL.swift
+++ b/Sources/Foundation/NSURL.swift
@@ -481,12 +481,16 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
     
     // Memory leak. See https://github.com/apple/swift-corelibs-foundation/blob/master/Docs/Issues.md
     open var fileSystemRepresentation: UnsafePointer<Int8> {
-
 #if os(Windows)
-        let bufSize = Int(MAX_PATH + 1)
+        if let resolved = CFURLCopyAbsoluteURL(_cfObject),
+                let representation = CFURLCopyFileSystemPath(resolved, kCFURLWindowsPathStyle)?._swiftObject {
+            let buffer = UnsafeMutablePointer<Int8>.allocate(capacity: representation.count + 1)
+            representation.withCString { buffer.initialize(from: $0, count: representation.count + 1) }
+            buffer[representation.count] = 0
+            return UnsafePointer(buffer)
+        }
 #else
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/Foundation/NSURL.swift
```diff
@@ -481,19 +481,24 @@ open class NSURL : NSObject, NSSecureCoding, NSCopying {
     
     // Memory leak. See https://github.com/apple/swift-corelibs-foundation/blob/master/Docs/Issues.md
     open var fileSystemRepresentation: UnsafePointer<Int8> {
-
 #if os(Windows)
-        let bufSize = Int(MAX_PATH + 1)
+        if let resolved = CFURLCopyAbsoluteURL(_cfObject),
```
gwynne, line 485: Can we add a comment here explaining why the Windows behavior needs to be different here but not in `_wideFileSystemRepresentation`?

compnerd: I suspect that the `_wideFileSystemRepresentation will need to change.  At that point, I'm more tempted to remove that function.  Having different variants of the functions makes things more difficult to correctly refactor.  That is, this is far more surgical than ideal.  I can add a note on the wide variant that it needs to be updated and ideally removed.

The honest answer is: fixing a bug at a time.  The reality is that most of the file system access routines for Windows need to be adjusted to use the NT path spelling, but that is also a good time to clean up some of the implementation to make it a bit less egregiously pessimistic (e.g. reusing encoded file paths, reusing stat information, etc).

compnerd: Discussed this offline - I'll add the comment in a follow up.  This is the first step towards fixing all the various cases in this file.  The work is going to enable DocC and should also help potentially help make SPM and driver more reliable on Windows with extended paths.

PR #4748, opened 2023-06-04
Foundation: avoid unnecessary disk IO on windows file enumeration

compnerd:
Prefer to use the `hasDirectoryPath` ivar on the URL which should be properly populated when populating the stack.  The one special case is when we first initialise the enumerator, where we can assume that we are given a directory path.


compnerd:
@swift-ci please test

compnerd:

compnerd:

PR #4750, opened 2023-06-04
Windows long paths support

compnerd:
Refactor directory iteration to avoid replicating the conversion to the
NT style paths everywhere.  The iteration logic can be done at a single
site with a callback to handle the iteration.

Improve handle long file paths on Windows when performing a `removeItem`
call.  Furthermore, correct a few areas where we were mishandling
junctions.  This would result in an early termination of the loop in
`removeItem` causing us to fail to clean up directories which we should
have been able to.  This improves the DocC test coverage on Windows.

Rework the file attribute reading on Windows operation to reformulate
the path to the absolute path representation and then into the NT form
before performing the operation.  This technically is a partial repair
as drive relative paths where the current directory is deep enough, the
path evaluation would fail.

### DIFF
--- a/Sources/Foundation/FileManager+Win32.swift
+++ b/Sources/Foundation/FileManager+Win32.swift
@@ -13,6 +13,92 @@
 import let WinSDK.INVALID_FILE_ATTRIBUTES
 import WinSDK
 
+extension URL {
+    fileprivate var NTPath: String {
+        // Use a NT style, device path to avoid the 261-character path
+        // limitation on Windows APIs.  The addition of the prefix will bypass
+        // the Win32 layer for the path handling and thus must be fully resolved
+        // and normalised before being passed in.  This allows us access to the
+        // complete path limit as imposed by the NT kernel rather than the 260
+        // character limit as imposed by Win32.
+        #"\\?\\#(CFURLCopyFileSystemPath(CFURLCopyAbsoluteURL(_cfObject), kCFURLWindowsPathStyle)!._swiftObject)"#
+    }
+
+    fileprivate func withUnsafeNTPath<Result>(_ body: (UnsafePointer<WCHAR>) throws -> Result) rethrows -> Result {
+        try self.NTPath.withCString(encodedAs: UTF16.self, body)
+    }
+}
+
+
+private func withNTPathRepresentation<Result>(of path: String, _ body: (UnsafePointer<WCHAR>) throws -> Result) throws -> Result {
+    guard !path.isEmpty else {
+        throw CocoaError.error(.fileReadInvalidFileName, userInfo: [NSFilePathErrorKey:path])
+    }
+
+    // 1. Normalize the path first.
+
+    var path = path
+
+    // Strip the leading `/` on a RFC8089 path (`/[drive-letter]:/...` ).  A
+    // leading slash indicates a rooted path on the drive for teh current
+    // working directory.
+    var iter = path.makeIterator()
+    if iter.next() == "/", iter.next()?.isLetter ?? false, iter.next() == ":" {
+        path.removeFirst()
+    }
+
+    // Win32 APIs can support `/` for the arc separator. However,
+    // symlinks created with `/` do not resolve properly, so normalize
+    // the path.
+    path = path.replacing("/", with: "\\")
+
+    // Droop trailing slashes unless it follows a drive specification.  The
+    // trailing arc separator after a drive specifier iindicates the root as
+    // opposed to a drive relative path.
+    while path.count > 1, path[path.index(before: path.endIndex)] == "\\",
+            !(path.count == 3 &&
+                path[path.index(path.endIndex, offsetBy: -2)] == ":" &&
+                path[path.index(path.endIndex, offsetBy: -3)].isLetter) {
+        path.removeLast()
+    }
+
+    // 2. Perform the operation on the normalized path.
+
+    return try path.withCString(encodedAs: UTF16.self) { pwszPath in
+        guard !path.hasPrefix(#"\\"#) else { return try body(pwszPath) }
+
+        let dwLength = GetFullPathNameW(pwszPath, 0, nil, nil)
+        let path = withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
+            _ = GetFullPathNameW(pwszPath, DWORD($0.count), $0.baseAddress, nil)
+            return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+        }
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/Foundation/FileManager+Win32.swift
```diff
@@ -13,6 +13,96 @@
...
+
+        let hFind: HANDLE = FindFirstFileW($0, &ffd)
+        if hFind == INVALID_HANDLE_VALUE {
+            throw _NSErrorWithWindowsError(GetLastError(), reading: true, paths: [path.path])
+        }
+
+        defer { FindClose(hFind) }
+
+        repeat {
+            let entry: String = withUnsafePointer(to: ffd.cFileName) {
```
gwynne, line 95: Can you use with `withUnsafeBytes(of:)` here to avoid having to calculate the `capacity` value?

Sources/Foundation/FileManager+Win32.swift
```diff
@@ -198,28 +288,15 @@ extension FileManager {
     }
 
     internal func _contentsOfDir(atPath path: String, _ closure: (String, Int32) throws -> () ) throws {
-        guard path != "" else {
-            throw NSError(domain: NSCocoaErrorDomain, code: CocoaError.fileReadInvalidFileName.rawValue, userInfo: [NSFilePathErrorKey : NSString(path)])
+        guard !path.isEmpty else {
+            throw NSError(domain: NSCocoaErrorDomain,
+                          code: CocoaError.fileReadInvalidFileName.rawValue,
+                          userInfo: [NSFilePathErrorKey:NSString(path)])
```
gwynne, line 294:
```suggestion
            throw CocoaError.error(.fileReadInvalidFileName, userInfo: [NSFilePathErrorKey: path])
```
Either you _don't_ need the `NSString()` conversion here or you _do_ need it above on line 63, which is it? 😕

Sources/Foundation/FileManager+Win32.swift
```diff
@@ -571,94 +648,83 @@ extension FileManager {
...
-                guard alreadyConfirmed || shouldRemoveItemAtPath(itemPath, isURL: isURL) else {
-                    continue
-                }
-
-                if try FileManager.default._fileSystemRepresentation(withPath: itemPath, RemoveDirectoryW) {
-                  continue
-                }
-                guard GetLastError() == ERROR_DIR_NOT_EMPTY else {
-                    throw _NSErrorWithWindowsError(GetLastError(), reading: false, paths: [itemPath])
+            if faAttributes.dwFileAttributes & FILE_ATTRIBUTE_READONLY == FILE_ATTRIBUTE_READONLY {
```
gwynne, line 657: Shame those don't import as an `OptionSet`

gwynne (commented):
Pretty much just a bunch of style nits, functionally this all looks good to me

Sources/Foundation/FileManager+Win32.swift
```diff
@@ -13,6 +13,96 @@
...
+    }
+
+    fileprivate func withUnsafeNTPath<Result>(_ body: (UnsafePointer<WCHAR>) throws -> Result) rethrows -> Result {
+        return try NTPath.withCString(encodedAs: UTF16.self, body)
+    }
+}
+
+
+private func withNTPathRepresentation<Result>(of path: String, _ body: (UnsafePointer<WCHAR>) throws -> Result) throws -> Result {
+    func withNormalizedRepresentation<Result>(of path: String, _ body: (UnsafePointer<WCHAR>) throws -> Result) rethrows -> Result {
```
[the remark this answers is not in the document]
compnerd: I find that the normalization is a bit hairy and its easier to think about it as a separate operation.  We could of course do it inline.  I think that it does make sense to inline it though, especially with some of the suggested tweaks.

Sources/Foundation/FileManager+Win32.swift
compnerd: I don't think that the string is guaranteed to be null terminated as it is a fixed size buffer.  The capacity is `sizeof(cFileName)`.

Sources/Foundation/FileManager+Win32.swift
compnerd: I think that we can get away without doing the `NSString` bridging.  Let's go with that (assuming that tests don't break).

Sources/Foundation/FileManager+Win32.swift
```diff
@@ -13,6 +13,96 @@
...
+
+        // Win32 APIs can support `/` for the arc separator. However,
+        // symlinks created with `/` do not resolve properly, so normalize
+        // the path.
+        path = path.replacing("/", with: "\\")
+
+        // Droop trailing slashes unless it follows a drive specification.  The
+        // trailing arc separator after a drive specifier iindicates the root as
+        // opposed to a drive relative path.
+        while path.count > 1, path[path.index(before: path.endIndex)] == "\\",
```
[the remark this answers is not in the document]
compnerd: I think that I prefer the existing unnecessarily verbose spelling because it matches the following 2 cases and its easier to then realize what is going on.  I do agree with you though that `.last` is nicer.

compnerd:
@swift-ci please test

Sources/Foundation/FileManager+Win32.swift
```diff
@@ -13,6 +13,96 @@
...
+        // working directory.
+        if path.count > 3 {
+            let index0 = path.startIndex
+            let index1 = path.index(path.startIndex, offsetBy: 1)
+            let index2 = path.index(path.startIndex, offsetBy: 2)
+
+            if path[index0] == "/", path[index1].isLetter, path[index2] == ":" {
+                path.removeFirst()
+            }
+        }
```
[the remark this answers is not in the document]
gwynne: Without the trailing `!= nil` you get the effect of `>= 3` instead of `> 3`. If that doesn't matter, then yeah, definitely drop it.

Sources/Foundation/FileManager+Win32.swift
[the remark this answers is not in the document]
gwynne: They're all style nits anyhow, it's always up to you 🙂

Sources/Foundation/FileManager+Win32.swift
gwynne: That's my point - `withUnsafeBytes()` gives you an `UnsafeRawBufferPointer` of the appropriate size (which is going to be the size of the value you're accessing) instead of only an `UnsafePointer` - the former makes a lot more sense for accessing an imported C fixed-length array (which imports as a tuple, sigh).

compnerd:
@swift-ci please test

compnerd:

PR #4751, opened 2023-06-05
Foundation: simplify handler handling

compnerd:
Reduce the conditional logic by unwrapping the handler rather than adding conditional case.  This should make it easier to perform a series of refactorings.


compnerd:
@swift-ci please test

PR #4752, opened 2023-06-05
Foundation: add small extensions for URL

compnerd:
We need to convert the paths to NT style paths to get extended paths to work.  Provide the ability to query the NT style path and a helper to perform a callback with the C-style representation of the NT style path representation.


compnerd:
@swift-ci please test

PR #4753, opened 2023-06-06
Tests: correct spelling of `/dev/null` on Windows

compnerd:
Windows uses a per-directory `NUL` rather than a global `/dev/null`. Adjust the test to use the appropriate spelling rather than potentially seeing a random drive-relative path which may exist as a deletable file.


compnerd:
@swift-ci please test

PR #4760, opened 2023-06-13
[5.9] Do not chdir before calling posix_spawn

MaxDesiatov:
Cherry-pick of https://github.com/apple/swift-corelibs-foundation/pull/4606.

We allow users to set the working directory of the child process when they use the Process API. Unfortunately, this was implemented by calling chdir in the parent process, which has nasty global effects on the process and cannot safely be used concurrently.

glibc has a non-POSIX extension to posix_spawn that can call chdir in the child process, so we add that call instead where it's available, and add a regression test.

Resolves apple/swift#59610


MaxDesiatov:
@swift-ci test

tomerd (approved):

parkera (approved):

Issue #4762, opened 2023-06-14
CFPlatform.c:2214:10: error: call to undeclared function 'posix_spawn_file_actions_addchdir_np'


tshortli:
Foundation is currently failing to build on Ubuntu:

```
/home/build-user/swift-corelibs-foundation/CoreFoundation/Base.subproj/CFPlatform.c:2214:10: error: call to undeclared function 'posix_spawn_file_actions_addchdir_np'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  return posix_spawn_file_actions_addchdir_np((posix_spawn_file_actions_t *)file_actions, path);
         ^
1 error generated.
```

https://ci.swift.org/job/oss-swift-RA-linux-ubuntu-18_04/4169/console

tshortli:
@Lukasa looks like this may be fallout from https://github.com/apple/swift-corelibs-foundation/pull/4606, could you take a look?

Lukasa:
The merged PR requires glibc 2.29, but Ubuntu Bionic ships with 2.27. I don't really know what we should do really, the feature simply doesn't work without it.

Lukasa:
We can revert and keep a fallback to keep the current behaviour if we need to, I suppose.

Lukasa:
The way we'd add the fallback is to add a way to query the support, which can be implemented using `__GLIBC__` and `__GLIBC_MINOR__` macros.

tomerd:
querying + fallback seems reasonable

Lukasa:
@MaxDesiatov Do you have any bandwidth to do this?

MaxDesiatov:
I'll have a look, `__GLIBC_MINOR__` check makes sense to me.

PR #4763, opened 2023-06-15
Process: check for `posix_spawn_file_actions_addchdir_np`

MaxDesiatov:
Before calling `posix_spawn_file_actions_addchdir_np` we should check if it's available. On non-Linux platform where we know it can't be available we're taking a faster path that doesn't check at run time.

Resolves https://github.com/apple/swift-corelibs-foundation/issues/4762.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci smoke test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test macos

jakepetroules:
@MaxDesiatov Note that macOS has `posix_spawn_file_actions_addchdir_np` too.

PR #4765, opened 2023-06-16
Add `NSError.underlyingErrors`

woxtu:
This PR adds `NSError.underlyingErrors` and `NSMultipleUnderlyingErrorsKey`.
https://developer.apple.com/documentation/foundation/nserror/3738169-underlyingerrors
https://developer.apple.com/documentation/foundation/nsmultipleunderlyingerrorskey

### DIFF
--- a/Sources/Foundation/NSError.swift
+++ b/Sources/Foundation/NSError.swift
@@ -33,6 +33,9 @@ public let NSMachErrorDomain: String = "NSMachErrorDomain"
 // Key in userInfo. A recommended standard way to embed NSErrors from underlying calls. The value of this key should be an NSError.
 public let NSUnderlyingErrorKey: String = "NSUnderlyingError"
 
+// Key in userInfo. A recommended standard way to embed a list of several NSErrors from underlying calls. The value of this key should be an NSArray of NSError. This value is independent from the value of `NSUnderlyingErrorKey` - neither, one, or both may be set.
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

Sources/Foundation/NSError.swift
```diff
@@ -738,6 +741,18 @@ public extension CocoaError {
...
 
+    /// Return a list of underlying errors, if any. It includes the values of both NSUnderlyingErrorKey and NSMultipleUnderlyingErrorsKey. If there are no underlying errors, returns an empty array.
+    var underlyingErrors: [Error] {
+        var errors = [Error]()
+        if let underlyingError = _nsUserInfo[NSUnderlyingErrorKey] as? Error {
+            errors.append(underlyingError)
+        }
+        if let underlyingErrors = _nsUserInfo[NSMultipleUnderlyingErrorsKey] as? [Error] {
+            errors.append(contentsOf: underlyingErrors)
+        }
```
compnerd, line 752: Is the ordering well specified?  If not, I think it might be a good idea to swap the order here.  That way we can initialize the array and extend with the single element:

```swift
var errors = (_nsUserInfo[NSMultipleUnderlyingErrors] as? [Error]) ?? []
if let error = _nsUserInfo[NSUnderlyingErrorKey] as? Error {
  errors.append(error)
}
return errors
```

woxtu: I couldn't find about that, so I followed the ordering on macOS for now.

PR #4776, opened 2023-06-21
Foundation: Allow setting `.creationTime` file attribute on Windows

simonomi:
For some reason, the `_updateTimes` method for windows, when calling `SetFileTime`, did not pass an argument for creation time. This resulted in being unable to set the creation time attribute of a file on Windows.

### DIFF
--- a/Sources/Foundation/FileManager+Win32.swift
+++ b/Sources/Foundation/FileManager+Win32.swift
@@ -964,6 +964,8 @@ extension FileManager {
                                modificationTime: Date? = nil) throws {
       let stat = try _lstatFile(atPath: path, withFileSystemRepresentation: fsr)
 
+      var ctime: FILETIME =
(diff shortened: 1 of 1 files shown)


Sources/Foundation/FileManager+Win32.swift
```diff
@@ -977,7 +979,7 @@ extension FileManager {
       }
       defer { CloseHandle(hFile) }
 
-      if !SetFileTime(hFile, nil, &atime, &mtime) {
+      if !SetFileTime(hFile, &ctime, &atime, &mtime) {
```
compnerd, line 982: Hmm, does it make sense to conditionally set only the passed in times?

compnerd (commented):
Would it be possible to test this functionality?  I suspect that we can add a test case for this (or maybe enable an existing one?).

PR #4778, opened 2023-06-22
Implement `URLResponse.init()`

### LINKED ISSUE #4388: [SR-12671] URLResponse.init is not accessible in FoundationNetworking

woxtu:
This PR implements `URLResponse.init()`. Resolves #4388.

### DIFF
--- a/Sources/FoundationNetworking/URLResponse.swift
+++ b/Sources/FoundationNetworking/URLResponse.swift
@@ -87,7 +87,12 @@ open class URLResponse : NSObject, NSSecureCoding, NSCopying {
         let c = url.lastPathComponent
         self.suggestedFilename = c.isEmpty ? "Unknown" : c
     }
-    
+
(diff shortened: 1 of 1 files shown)


compnerd (changes requested):
I don't think that this is correct.  The definition of `NSURLResponse` does not indicate that `-[NSURLResponse init];` is a designated nor convenience initialiser.  This happens to work on macOS due to the ObjC underpinnings.

@parkera - am I way off base here and this should be an available API or should this not be available?

brianmichel:
Just wanted to chime in that I ran into this issue as well recently when trying to bring a library cross platform. While I can understand the desire for correctness, I'd argue that any construction of Swift code that isn't relying on something OS specific feels like fair game for reimplementation on SCF (or the eventual Swift Foundation) since it should be as much of a drop in replacement IMO

PR #4779, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.

### DIFF
--- a/CoreFoundation/Base.subproj/CFUtilities.c
+++ b/CoreFoundation/Base.subproj/CFUtilities.c
@@ -1675,7 +1675,7 @@ CFDictionaryRef __CFGetEnvironment() {
         extern char **environ;
         char **envp = environ;
 #elif TARGET_OS_LINUX
-#if !defined(environ) && !TARGET_OS_ANDROID
+#if !defined(environ) && !TARGET_OS_ANDROID && defined(__GLIBC_PREREQ)
 #define environ __environ
 #endif
         char **envp = environ;
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -189,7 +189,7 @@ static dispatch_queue_t __ ## PREFIX ## Queue(void) {			\
 #define CF_RETAIN_BALANCED_ELSEWHERE(obj, identified_location) do { } while (0)
 #endif
 
-#if (TARGET_OS_LINUX && !TARGET_OS_ANDROID && !TARGET_OS_CYGWIN) || TARGET_OS_WIN32
+#if (TARGET_OS_LINUX && !TARGET_OS_ANDROID && !TARGET_OS_CYGWIN && defined(__GLIBC_PREREQ)) || TARGET_OS_WIN32
 CF_INLINE size_t
 strlcpy(char * dst, const char * src, size_t maxlen) {
     const size_t srclen = strlen(src);
@@ -300,7 +300,7 @@ typedef unsigned long fd_mask;
 #endif
     
 
-#if !TARGET_OS_CYGWIN && !TARGET_OS_BSD
+#if !TARGET_OS_CYGWIN && !TARGET_OS_BSD && defined(__GLIBC_PREREQ)
 #define issetugid() 0
 #endif
 
--- a/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
+++ b/CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
@@ -69,6 +69,7 @@
 #include <features.h>
 #include <termios.h>
 
+#if defined(__GLIBC_PREREQ)
 #if __GLIBC_PREREQ(2, 28) == 0
 // required for statx() system call, glibc >=2.28 wraps the kernel function
 #include <sys/syscall.h>
@@ -78,6 +79,7 @@
 #include <linux/fs.h>
 #define AT_STATX_SYNC_AS_STAT   0x0000  /* - Do whatever stat() does */
 #endif //__GLIBC_PREREQ(2. 28)
+#endif // defined(__GLIBC_PREREQ)
 
 #ifndef __NR_statx
 #include <sys/stat.h>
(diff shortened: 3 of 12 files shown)


MaxDesiatov:
@swift-ci test

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -1675,7 +1675,7 @@ CFDictionaryRef __CFGetEnvironment() {
         extern char **environ;
         char **envp = environ;
 #elif TARGET_OS_LINUX
-#if !defined(environ) && !TARGET_OS_ANDROID
+#if !defined(environ) && !TARGET_OS_ANDROID && defined(__GLIBC_PREREQ)
```
compnerd, line 1678: I don't think that we should use `__GLIBC_PREREQ` to determine this, but rather `__GLIBC__` which has been present since glibc 6 (`__GNU_LIBRARY__` was used prior to that).

MaxDesiatov:
@swift-ci test

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -189,7 +189,7 @@ static dispatch_queue_t __ ## PREFIX ## Queue(void) {			\
 #define CF_RETAIN_BALANCED_ELSEWHERE(obj, identified_location) do { } while (0)
 #endif
 
-#if (TARGET_OS_LINUX && !TARGET_OS_ANDROID && !TARGET_OS_CYGWIN) || TARGET_OS_WIN32
+#if (TARGET_OS_LINUX && !TARGET_OS_ANDROID && !TARGET_OS_CYGWIN && defined(__GLIBC__)) || TARGET_OS_WIN32
```
compnerd, line 192: Seems like `defined(__GLIBC__) || TARGET_OS_WIN32`?

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
```diff
@@ -300,7 +300,7 @@ typedef unsigned long fd_mask;
 #endif
     
 
-#if !TARGET_OS_CYGWIN && !TARGET_OS_BSD
+#if !TARGET_OS_CYGWIN && !TARGET_OS_BSD && defined(__GLIBC__)
```
compnerd, line 303: Seems like `defined(__GLIBC__) || TARGET_OS_WIN32`

compnerd (approved):

CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
MaxDesiatov: Because Swift on Android is using `import Glibc` I'm confused how exactly their platform checks should look, even in C/C++. cc @finagolfin

finagolfin: This C constant works as expected, so it is undefined on non-Glibc platforms like Android. The issue is only in Swift, where [all non-Darwin and non-Windows platforms like Android and the BSDs currently use `import Glibc` to import their C libraries](https://github.com/finagolfin/swift-android-sdk/#porting-swift-packages-to-android) that one has to be careful.

finagolfin (approved):
Should work fine on Android and other platforms, but I agree with Saleem about shortening to `defined(__GLIBC__)` in four of those C conditions.

MaxDesiatov:
@swift-ci test

CoreFoundation/Base.subproj/CFUtilities.c
```diff
@@ -1675,7 +1675,7 @@ CFDictionaryRef __CFGetEnvironment() {
         extern char **environ;
         char **envp = environ;
 #elif TARGET_OS_LINUX
-#if !defined(environ) && !TARGET_OS_ANDROID
+#if !defined(environ) && !TARGET_OS_ANDROID && defined(__GLIBC__)
```
finagolfin, line 1678: Glibc isn't defined on Android, so you can remove the Android check.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
```diff
@@ -562,7 +564,7 @@ CF_CROSS_PLATFORM_EXPORT int _CFOpenFile(const char *path, int opts);
 static inline int _direntNameLength(struct dirent *entry) {
 #ifdef _D_EXACT_NAMLEN  // defined on Linux
     return _D_EXACT_NAMLEN(entry);
-#elif TARGET_OS_ANDROID
+#elif TARGET_OS_ANDROID || !defined(__GLIBC__)
```
finagolfin, line 567: Won't this activate this line for the BSDs and all other libcs? No way to just enable this for Musl alone?

MaxDesiatov: Musl provides no macro we could reliably check for.

MaxDesiatov:
@swift-ci test windows

CoreFoundation/Base.subproj/ForSwiftFoundationOnly.h
finagolfin: Would it make sense to add one in this repo then, something like `TARGET_LIBC_MUSL` [using this SO answer](https://stackoverflow.com/a/70211227)? Because this is technically wrong: glibc uses the `_D_EXACT_NAMLEN` branch above and has nothing to do with this condition.

al45tair: Couldn't we just write `__linux__` instead? Android and Musl are both Linux, after all.

We could even do

```
#elif defined(__linux__) && !defined(_DIRENT_HAVE_D_NAMLEN)
```

which would then mean if Musl or Android added `d_namlen` in the future, things would Just Work.

al45tair: I don't particularly like the attempts at detecting Musl in that Stack Overflow answer; Musl should probably define some macros of its own — it should ideally be possible to test not just for Musl but to check the Musl version as well.

finagolfin: > Couldn't we just write `__linux__` instead?

That would work and be fine here, but I was trying to come up with a more general solution.

al45tair: I think the general solution is to get musl to add `__MUSL__` (and ideally `__MUSL_MINOR__` and `__MUSL_PREREQ()` to go with it) in `<features.h>`. I have a patch for that locally — I'll see if we can upstream it.

al45tair: Hmmm: https://www.openwall.com/lists/musl/2013/03/29/13 doesn't sound good.

PR #4787, opened 2023-06-26
Fix a concurrency bug in the TestFileHandle.test_readToEndOfFileAndNotify() unit test.

grynspan:
Fix a concurrency bug in the `TestFileHandle.test_readToEndOfFileAndNotify()` unit test.

This test fails to build when corelibs-xctest marks the expectation handler as `@Sendable`. The fix is to replace the use of a `Bool` value with a separate expectation.


grynspan:
@swift-ci please test

compnerd (approved):

Catfish-Man (approved):
lgtm

PR #4793, opened 2023-07-10
Foundation: use NT-style paths for `CopyFileW`

compnerd:
When invoking `_copyRegularFile` ensure that we use the NT-style path to enable long file names on all targets.  This allows copying directories that contain a deeply nested path.


compnerd:
@swift-ci please test

PR #4794, opened 2023-07-12
Add `ListFormatter`

### LINKED ISSUE #4223: [SR-14247] ListFormatter missing

woxtu:
This PR adds `ListFormatter`. Resolves #4223.
https://developer.apple.com/documentation/foundation/listformatter

### DIFF
--- a/Sources/Foundation/ListFormatter.swift
+++ b/Sources/Foundation/ListFormatter.swift
@@ -16,7 +16,11 @@ open class ListFormatter: Formatter {
 
     /* Specifies the locale to format the items. Defaults to autoupdatingCurrentLocale. Also resets to autoupdatingCurrentLocale on assignment of nil.
      */
-    open var locale: Locale! = .autoupdatingCurrent
+    private var _locale: Locale = .autoupdatingCurrent
+    open var locale: Locale! {
+        get { return _locale }
+        set { _locale = newValue ?? .autoupdatingCurrent }
+    }
 
     /* Specifies how each object should be formatted. If not set, the object is formatted using its instance method in the following order: -descriptionWithLocale:, -localizedDescription, and -description.
      */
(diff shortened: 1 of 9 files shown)


Tests/Foundation/Tests/TestListFormatter.swift
```diff
@@ -0,0 +1,108 @@
...
+    override func tearDown() {
+        formatter = nil
+
+        super.tearDown()
+    }
+
+    func test_locale() throws {
+        XCTAssertEqual(formatter.locale, Locale.autoupdatingCurrent)
+
+        formatter.locale = Locale(identifier: "en_US_POSIX")
```
compnerd, line 36: This is not portable (`en_US` is fine, `en_US_POSIX` is not a valid locale on Windows).

Sources/Foundation/ListFormatter.swift
```diff
@@ -0,0 +1,99 @@
...
+     */
+    /*@NSCopying*/ open var itemFormatter: Formatter?
+
+    public override init() {
+        self.cfFormatter = _CFListFormatterCreate(kCFAllocatorSystemDefault, CFLocaleCopyCurrent())!
+        super.init()
+    }
+
+    public required init?(coder: NSCoder) {
+        self.cfFormatter = _CFListFormatterCreate(kCFAllocatorSystemDefault, CFLocaleCopyCurrent())!
```
compnerd, line 35: Most of the SCF does not support `initWithCoder`.  Shall we follow precedent and `fatalError()` here?

Sources/Foundation/ListFormatter.swift
```diff
@@ -0,0 +1,99 @@
...
+        super.init(coder: coder)
+    }
+
+    open override func copy(with zone: NSZone? = nil) -> Any {
+        let copied = ListFormatter()
+        copied.locale = locale
```
compnerd, line 41: I believe that we should just use the `copy` method instead

compnerd (commented):
I'm a little on the fence about this as there is new API surface that we need to create and expose from CF.  @parkera what is your take?  The changes to CF are pretty minimal and for a good reason - they are supporting API equality.

parkera:
Our plan of record here is to focus on https://github.com/apple/swift-foundation/, where we actually already have the `ListFormatStyle` API. Why add it here to CF when we can focus on the more modern Swift-first approach?

itingliu:
Closing per @parkera's comment. Also this discussion is quite old and there have been conflicts. Please reopen if you have other thoughts, thanks

PR #4797, opened 2023-07-18
Foundation: improve relative path computation for iteration on Windows

compnerd:
This partially corrects the relative path computation.  We no longer fail to create the file system representation if it exceeds `MAX_PATH` due to the switch to `withUnsafeNTPath`.  We will still fail to properly create the relative path if the paths are unrelated and if the relative path exceeds `MAX_PATH`.  However, this still is an improvement over the status quo.


compnerd:
@swift-ci please test

PR #4799, opened 2023-07-22
Fix `NSDecimalPower(_:_:_:_:)` behavior

Oschly:
`NSDecimalPower` accepts `Int` as `power` parameter, but internally it calls `power(_:roundingMode:)` method, that accepts only `UInt` as power value. That results in overflow error for every call of `NSDecimalPower` with negative `power` parameter.

This PR fixes that behavior by modifying `power` method by making `p` an Int and adds additional code for handling negative `p`.


MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):

Sources/Foundation/Decimal.swift
```diff
@@ -2242,19 +2242,19 @@ extension Decimal {
         _exponent = newExponent
         return .noError
     }
-    fileprivate mutating func power(_ p:UInt, roundingMode:RoundingMode) -> CalculationError {
+    fileprivate mutating func power(_ p:Int, roundingMode:RoundingMode) -> CalculationError {
         if isNaN {
             return .overflow
         }
-        var power = p
+        var power = abs(p)
```
compnerd, line 2249: This does become a bit confusing.  It would've been nicer to use a more verbose name (e.g. `absPower`).

compnerd (approved):

MaxDesiatov:
Either the tests or the implementation need to be updated as this test case fails on CI.

Sources/Foundation/Decimal.swift
Oschly: You're right, I've changed the name to `absPower`.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@compnerd I'm not seeing this macOS CI failure on PRs that you merged recently. Would you agree that it's sporadic and unrelated to this change and that it can be safely merged?

compnerd:

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):

parkera:
This still seems relevant, and we just moved this stuff around too. @iCharlesHu can you take a look please?

Oschly:
It seems like it's still relevant, but now this issue appears in this repo and the swift-foundation. Should I close this PR and apply fix directly to swift-foundation or both PRs should be updated to the latest changes?

parkera:
The idea here is that swift-foundation contains the canonical implementation (shared with Darwin). swift-corelibs-foundation exposes the free-standing function for Linux.

PR #4800, opened 2023-07-22
Foundation: make `_moveItem` resilient to multiple volumes

compnerd:
`MoveFileExW` does not permit moving directories across volumes. Account for this by checking the source and fallback to copy and delete.


compnerd:
@swift-ci please test

PR #4801, opened 2023-07-24
Foundation: adjust `_NSCleanupTemporaryFile` on Windows

compnerd:
Prefer to use `withNTPathRepresentation` over the
`_fileSystemRepresentation` usage as the former will not convert to the NT UNC path representation which may break on long paths.


compnerd:
@swift-ci please test

PR #4802, opened 2023-07-24
Foundation: alter `NSData.contents(of:options:)` on Windows

compnerd:
We would previously use the FSR for the path, however, due to the expectations of FSR, we cannot support long paths with it.  Use the internal `withUnsafeNTPath` extension on `URL` to use a long path representation to support extended paths (>260 characters) on Windows.


compnerd:
@swift-ci please test

PR #4805, opened 2023-07-31
Fix the way known locations are generated on windows

darinf:
Use GetAllUsersProfileDirectoryW and GetProfilesDirectoryW instead of hardcoding paths.

Previously, the known locations were treated as UNC paths, which would typically fail to be read after ~5 second timeout.

### DIFF
--- a/CoreFoundation/Base.subproj/CFKnownLocations.c
+++ b/CoreFoundation/Base.subproj/CFKnownLocations.c
@@ -16,6 +16,10 @@
 
 #include <assert.h>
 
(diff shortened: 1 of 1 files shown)


compnerd:

compnerd:

compnerd:

compnerd:

compnerd:
@swift-ci please clean test macOS platform

compnerd:

compnerd:
This is a windows only path change, and the failure is an infrastructure failure on the macOS side.

Issue #4806, opened 2023-08-03
JSONDecoder parse type differently on Linux


ypli-chn:
Example Code
```swift
import Foundation

class Base: Codable { }
class Child: Base { }

func dynamicType() -> Base.Type {
    return Child.self
}

let data = "{}".data(using: .utf8)!

let classType = dynamicType()
let instance = try JSONDecoder().decode(classType, from: data)
print("type: \(instance) => \(type(of: instance))")
```

on macOS:
```
type: JSONDecoderExample.Child => Child
```
on Linux:
```
type: JSONDecoderExample.Base => Base
```

glessard:
Fixed in https://github.com/apple/swift-corelibs-foundation/pull/4807

PR #4807, opened 2023-08-03
Fix JSONDecoder.decode(_,from:) behavior

ypli-chn:
`JSONDecoder.decode(_,from:)` accepts  a generic parameter, but it does not use any concrete type in practice.
This PR fixes that use parameter `type` instead of `T.self`.

This PR solves #4806 

Example Code
```swift
import Foundation

class Base: Codable { }
class Child: Base { }

func dynamicType() -> Base.Type {
    return Child.self
}

let data = "{}".data(using: .utf8)!

let classType = dynamicType()
let instance = try JSONDecoder().decode(classType, from: data)
print("type: \(instance) => \(type(of: instance))")
```

on macOS:
```
type: JSONDecoderExample.Child => Child
```
on Linux:
```
type: JSONDecoderExample.Base => Base
```

### DIFF
--- a/Sources/Foundation/JSONDecoder.swift
+++ b/Sources/Foundation/JSONDecoder.swift
@@ -197,7 +197,7 @@ open class JSONDecoder {
         do {
             var parser = JSONParser(bytes: Array(data))
             let json = try parser.parse()
-            return try JSONDecoderImpl(userInfo: self.userInfo, from: json, codingPath: [], options: self.options).unwrap(as: T.self)
+            return try JSONDecoderImpl(userInfo: self.userInfo, from: json, codingPath: [], options: self.options).unwrap(as: type)
         } catch let error as JSONError {
             throw DecodingError.dataCorrupted(DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: error))
         } catch {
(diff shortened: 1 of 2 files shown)


lxbndr:
Would you mind adding a test case to the `TestJSONEncoder`? This is a severe regression, and we would like to catch similar issues in future.

ypli-chn:
Sure. I added a method named `test_childTypeDecoder` to test this issue.

glessard (approved):
Thank you for the contribution!

glessard:
@swift-ci please test

glessard:
The discrepancy you describe in https://github.com/apple/swift-corelibs-foundation/issues/4806 has probably been present for a long time. It appears that it is fixed already in https://github.com/apple/swift-foundation, which will become the Foundation implementation for the Linux and Windows at a future point. I will make sure that a test like this exists in swift-foundation.

glessard:
@swift-ci please test windows platform

glessard:
The macOS test failure is a known issue.

glessard:
Circling back, the behaviour fixed here is tested for in the following test of swift-foundation:
https://github.com/apple/swift-foundation/blob/62500a5d7921cc61703c2a1b4c961adb12dc80c3/Tests/FoundationEssentialsTests/JSONEncoderTests.swift#L680

PR #4811, opened 2023-08-17
Added check for existing strlcpy/strlcat on Linux

tachoknight:
The GNU C Library version 2.38 introduced `strlcat()` and `strlcpy()` for the first time (https://lists.gnu.org/archive/html/info-gnu/2023-07/msg00010.html). Because these functions are defined in `CoreFoundation/Base.subproj/CoreFoundation_Prefix.h` there is a compile error in that the `CF_INLINE` expands to include `static`, which the glibc versions do not. Even if the function signatures were the same, there would be link errors due to multiple definitions found of the same functions. 

This patch adds an additional check in the CMakeFiles.txt file under the Linux section for the functions and, if found, defines a flag that prevents the functions from being defined in `CoreFoundation/Base.subproj/CoreFoundation_Prefix.h`.


parkera (approved):

parkera:
@swift-ci test

PR #4820, opened 2023-09-15
build: adjust the build for rebranch

compnerd:
The newer clang compiler prefers `module.modulemap` over `module.map`. This will result in us using the framework form of the module definitions for CoreFoundation. While this is fine for the compilation, the linker directive to link against the framework will also be emitted into the object files. This causes a linker failure on Windows as `-framework (CoreFoundation|CFURLInterface|CFXMLInterface)` is not a valid linker option. Use the swift control over the directives to avoid emitting the framework link directive to repair the build.


compnerd:
@swift-ci please test

compnerd:
I think that we should be able to remove them after the rebranch.  This is going into main for now.

PR #4830, opened 2023-09-29
Optimize _byteCount with compiler intrinsics

AtariDreams:
Use compiler intrinsics to make _byteCount faster


AtariDreams:
@parkera @FranzBusch thoughts?

AtariDreams:
@bnbarham @DougGregor Thoughts on this?

AtariDreams:
@FranzBusch Thoughts? Could you please run please test

compnerd:
There is no guarantee that this is faster.  The loop idiom recognizers will often make the appropriate optimizations.  This simply makes the code less portable - `__builtin_cltz` is not a valid directive for MSVC.

AtariDreams:
> There is no guarantee that this is faster. The loop idiom recognizers will often make the appropriate optimizations. This simply makes the code less portable - `__builtin_ctz` is not a valid directive for MSVC.

I can show you on compiler explorer.

PR #4845, opened 2023-11-24
Mark `Decimal` as `Sendable`

yaglo:
Decimal is a simple value type, so it should be sendable. Currently, using Decimal variables produces warnings when using them with Sendable checking.


jmschonfeld (approved):
Makes sense to me! Thanks for adding that

parkera:
This was done as part of the `Decimal` -> swift-foundation work, [here](https://github.com/apple/swift-foundation/pull/664).

PR #4850, opened 2023-12-01
Android: add better nullability checks for nullability annotations added in NDK 26

finagolfin:
This is needed because [Bionic recently added a bunch of these annotations](https://android.googlesource.com/platform/bionic/+/25725717861813c18189c2699cf663bdb984d1b9%5E%21/#F0). I made sure this pull doesn't break anything by testing most of it with the previous NDK 25c also. I used this patch with others to build the Swift toolchain for my Android CI, finagolfin/swift-android-sdk#122, and [the Termux app for Android](https://github.com/termux/termux-packages/commit/d10fd452662d3c39c0f19a901433bce8711b4688#diff-b8f3b540c354923485cb06eba2d66237536f89e88fd51e170495ddd8f5d32622R44), which now uses NDK 26b.

@drodriguez or @compnerd, please review.

### DIFF
--- a/Sources/Foundation/FileHandle.swift
+++ b/Sources/Foundation/FileHandle.swift
@@ -307,10 +307,15 @@ open class FileHandle : NSObject {
             if options.contains(.alwaysMapped) {
                 // Filesizes are often 64bit even on 32bit systems
                 let mapSize = min(length, Int(clamping: statbuf.st_size))
+              #if os(Android)
+                // Bionic mmap() now returns _Nonnull, so the force unwrap below isn't needed.
                 let data = mmap(nil, mapSize, PROT_READ, MAP_PRIVATE, _fd, 0)
+              #else
+                let data = mmap(nil, mapSize, PROT_READ, MAP_PRIVATE, _fd, 0)!
+              #endif
                 // Swift does not currently expose MAP_FAILURE
                 if data != UnsafeMutableRawPointer(bitPattern: -1) {
-                    return NSData.NSDataReadResult(bytes: data!, length: mapSize) { buffer, length in
+                    return NSData.NSDataReadResult(bytes: data, length: mapSize) { buffer, length in
                         munmap(buffer, length)
                     }
                 }
--- a/Sources/Foundation/FileManager+POSIX.swift
+++ b/Sources/Foundation/FileManager+POSIX.swift
@@ -741,20 +741,26 @@ extension FileManager {
             if rmdir(fsRep) == 0 {
                 return
             } else if errno == ENOTEMPTY {
+                #if os(Android)
+                let ps = UnsafeMutablePointer<UnsafeMutablePointer<Int8>>.allocate(capacity: 2)
+                ps.initialize(to: UnsafeMutablePointer(mutating: fsRep))
+                ps.advanced(by: 1).initialize(to: unsafeBitCast(0, to: UnsafeMutablePointer<Int8>.self))
+                #else
                 let ps = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 2)
                 ps.initialize(to: UnsafeMutablePointer(mutating: fsRep))
                 ps.advanced(by: 1).initialize(to: nil)
+                #endif
                 let stream = fts_open(ps, FTS_PHYSICAL | FTS_XDEV | FTS_NOCHDIR | FTS_NOSTAT, nil)
                 ps.deinitialize(count: 2)
                 ps.deallocate()
 
-                if stream != nil {
+                if let openStream = stream {
                     defer {
-                        fts_close(stream)
+                        fts_close(openStream)
                     }
 
-                    while let current = fts_read(stream)?.pointee {
-                        let itemPath = string(withFileSystemRepresentation: current.fts_path, length: Int(current.fts_pathlen))
+                    while let current = fts_read(openStream)?.pointee, let current_path = current.fts_path {
+                        let itemPath = string(withFileSystemRepresentation: current_path, length: Int(current.fts_pathlen))
                         guard alreadyConfirmed || shouldRemoveItemAtPath(itemPath, isURL: isURL) else {
                             continue
                         }
@@ -762,11 +768,11 @@ extension FileManager {
                         do {
                             switch Int32(current.fts_info) {
                             case FTS_DEFAULT, FTS_F, FTS_NSOK, FTS_SL, FTS_SLNONE:
-                                if unlink(current.fts_path) == -1 {
+                                if unlink(current_path) == -1 {
                                     throw _NSErrorWithErrno(errno, reading: false, path: itemPath)
                                 }
                             case FTS_DP:
-                                if rmdir(current.fts_path) == -1 {
+                                if rmdir(current_path) == -1 {
                                     throw _NSErrorWithErrno(errno, reading: false, path: itemPath)
                                 }
                             case FTS_DNR, FTS_ERR, FTS_NS:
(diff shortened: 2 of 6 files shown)


Sources/Foundation/FileHandle.swift
```diff
@@ -310,9 +310,15 @@ open class FileHandle : NSObject {
                 let data = mmap(nil, mapSize, PROT_READ, MAP_PRIVATE, _fd, 0)
                 // Swift does not currently expose MAP_FAILURE
                 if data != UnsafeMutableRawPointer(bitPattern: -1) {
+                  #if os(Android)
+                    return NSData.NSDataReadResult(bytes: data, length: mapSize) { buffer, length in
+                        munmap(buffer, length)
+                    }
+                  #else
                     return NSData.NSDataReadResult(bytes: data!, length: mapSize) { buffer, length in
```
finagolfin, line 318: As shown in my Bionic link above, `mmap()` now returns `_Nonnull`, so this force-unwrap fails.

parkera: It feels rather unfortunate to ifdef around this here instead of at the place we call `mmap`.

Sources/Foundation/FileHandle.swift
```diff
@@ -307,10 +307,15 @@ open class FileHandle : NSObject {
             if options.contains(.alwaysMapped) {
                 // Filesizes are often 64bit even on 32bit systems
                 let mapSize = min(length, Int(clamping: statbuf.st_size))
+              #if os(Android)
+                // Bionic mmap() now returns _Nonnull, so the force unwrap below isn't needed.
```
finagolfin, line 311: @parkera, is this the change you had in mind?

Sources/Foundation/FileManager+POSIX.swift
```diff
@@ -1085,10 +1091,18 @@ extension FileManager {
             do {
                 guard fm.fileExists(atPath: _url.path) else { throw _NSErrorWithErrno(ENOENT, reading: true, url: url) }
                 _stream = try FileManager.default._fileSystemRepresentation(withPath: _url.path) { fsRep in
+                    #if os(Android)
```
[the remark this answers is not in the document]
finagolfin: Maybe there is, I just don't know how to do it. This is simply the best I came up with after seeing these Swift type-casting issues for the first time, and I wanted to ask if there was a better way.

@glbrntt, you recently added similar code to NIO, wdyt?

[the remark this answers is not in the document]
glbrntt: NIO wraps all of the syscalls and does any platform-specific shenanigans within the wrapper.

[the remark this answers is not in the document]
finagolfin: OK, but that will still require adapting these types for nullability annotations: see [my recent patch of your new NIO code for Android](https://github.com/finagolfin/swift-android-sdk/blob/main/swift-nio-filesystem-ndk26.patch#L56).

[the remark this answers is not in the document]
glbrntt: re: your patch, does using an implicitly unwrapped optional work for both cases? e.g. `[UnsafeMutablePointer<CInterop.PlatformChar>!]`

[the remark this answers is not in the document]
finagolfin: I get the following error with Swift 5.9.2 on Android AArch64 if I add that to the type signature of `libc_fts_open()`:
```
swift-nio/Sources/NIOFileSystem/Internal/System Calls/Syscalls.swift:407:14: error: using '!' is not allowed here; perhaps '?' was intended?
    _ path: [UnsafeMutablePointer<CInterop.PlatformChar>!],
             ^                                          ~
                                                        ?
error: fatalError
```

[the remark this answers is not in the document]
finagolfin: @glbrntt, is that type what you had in mind, or simply unwrapping the arguments when they are passed in inside `ftsOpen()`?

finagolfin:
@parkera, any further feedback here?

Sources/Foundation/FileManager+POSIX.swift
```diff
@@ -741,32 +741,38 @@ extension FileManager {
             if rmdir(fsRep) == 0 {
                 return
             } else if errno == ENOTEMPTY {
+                #if os(Android)
+                let ps = UnsafeMutablePointer<UnsafeMutablePointer<Int8>>.allocate(capacity: 2)
```
compnerd, line 745: Do we really need to duplicate this or can we get away with a rebinding of the memory?

Sources/Foundation/FileManager+POSIX.swift
```diff
@@ -741,32 +741,38 @@ extension FileManager {
...
                 let ps = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: 2)
                 ps.initialize(to: UnsafeMutablePointer(mutating: fsRep))
                 ps.advanced(by: 1).initialize(to: nil)
+                #endif
                 let stream = fts_open(ps, FTS_PHYSICAL | FTS_XDEV | FTS_NOCHDIR | FTS_NOSTAT, nil)
                 ps.deinitialize(count: 2)
                 ps.deallocate()
 
-                if stream != nil {
+                if let openStream = stream {
```
compnerd, line 757: Why not simply `if let stream`?

Sources/Foundation/FileManager+POSIX.swift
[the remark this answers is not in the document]
compnerd: I think that I approached it similarly to what was being suggested - rebinding the memory around the call site.

Sources/Foundation/FileManager+POSIX.swift
```diff
@@ -1315,7 +1329,7 @@ extension FileManager {
             let finalErrno = originalItemURL.withUnsafeFileSystemRepresentation { (originalFS) -> Int32? in
                 return newItemURL.withUnsafeFileSystemRepresentation { (newItemFS) -> Int32? in
                     // This is an atomic operation in many OSes, but is not guaranteed to be atomic by the standard.
-                    if rename(newItemFS, originalFS) == 0 {
+                    if let newFS = newItemFS, let origFS = originalFS, rename(newFS, origFS) == 0 {
```
compnerd, line 1332: Why not simply force unwrap?

Sources/Foundation/Host.swift
```diff
@@ -25,7 +25,8 @@ import WinSDK
 
     // getnameinfo uses size_t for its 4th and 6th arguments.
     private func getnameinfo(_ addr: UnsafePointer<sockaddr>?, _ addrlen: socklen_t, _ host: UnsafeMutablePointer<Int8>?, _ hostlen: socklen_t, _ serv: UnsafeMutablePointer<Int8>?, _ servlen: socklen_t, _ flags: Int32) -> Int32 {
-        return Glibc.getnameinfo(addr, addrlen, host, Int(hostlen), serv, Int(servlen), flags)
+        guard let saddr = addr else { return -1 }
```
compnerd, line 28: I think that I prefer my approach of changing the signature of the private func.

finagolfin: Sorry about the delay, been meaning to get to this, hopefully this week.

finagolfin:
Now that `swift-foundation` was merged, this pull no longer applies cleanly to that rewrite. Instead, I have modified or removed portions of this pull based on the review comments here and applied it to the `release/6.0` branch in #5010 instead, where `swift-foundation` will presumably not be merged also.

Once that's in 6.0, I will rework this pull for `swift-foundation` and make the necessary changes here also. My answers to the review comments above will now address what I changed for #5010.

Sources/Foundation/FileManager+POSIX.swift
finagolfin: No longer needed after your swiftlang/swift#74829, so removed in #5010.

Sources/Foundation/FileManager+POSIX.swift
finagolfin: Good point, removed in #5010, you can add it in a version of #4889 for 6.0.

Sources/Foundation/FileManager+POSIX.swift
finagolfin: That's better, dropped this in #5010, you can do that in a version of #4889 for 6.0.

Sources/Foundation/Host.swift
finagolfin: Changing the signature would work too, dropped this in #5010, you can submit it in a version of #4889 for 6.0.

finagolfin:
Some of these are no longer needed after the re-core and other changes, others have been submitted again in #5024 and my pull for that, hyp#1.

PR #4852, opened 2023-12-11
Make isleap faster

AtariDreams:
The assembler generated is better if we compare directly instead of doing fancy ring buffer-eque logic.

Proof:
https://godbolt.org/z/dferPP9h4


AtariDreams:
@YOCKOW Can we please run the tests?

YOCKOW:
@swift-ci Please test

YOCKOW:
@parkera Can we merge CF-only change for perf here?

YOCKOW:
Tests failed because original `isleap` seems to expect that `year` is the actual year minus `1`.
https://godbolt.org/z/cad119qrG

AtariDreams:
@YOCKOW Fixed! Can we try again?

parkera:
@swift-ci test

YOCKOW:
I wonder why Linux test has passed because I think new implementation has to be changed as I suggested above🤔

Issue #4791, opened 2023-07-03
`URLSession._MultiHandle` doesn't wait for `DispatchSource` cancel handler


lxbndr:
The problem is pretty old and has been [reported](https://github.com/apple/swift-corelibs-libdispatch/issues/609) in Dispatch initially. After the investigation, it appears that it has to be fixed in Foundation. I am opening this issue for tracking and sharing the progress.

### Background
URLSession utilizes cURL for socket management, including reading, writing, and basic protocol operations. cURL, in turn, requires external notification when there is pending data or an event on a socket. This is where `DispatchSource` comes into play. The implementation ties the lifecycle of a `DispatchSource` to the register/unregister requests from cURL.

### The Issue
When cURL requests `_MultiHandle` to unregister a socket, the [`_SocketSources.tearDown`](https://github.com/apple/swift-corelibs-foundation/blob/7fc59bfa50e8679675d19728245d3e9784c34d5f/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift#L452C14-L452C14) function is implicitly called. This function invokes `DispatchSource.cancel()` and immediately returns control to cURL. Right after that cURL closes the socket. The Dispatch [documentation](https://developer.apple.com/documentation/dispatch/dispatchsourceprotocol/1780783-cancel#:~:text=It%20is%20invalid%20to%20close%20a%20file%20descriptor%20or%20deallocate%20a%20mach%20port%20that%20is%20currently%20being%20tracked%20by%20a%20dispatch%20source%20object%20before%20the%20cancellation%20handler%20is%20invoked.) says
> It is invalid to close a file descriptor or deallocate a mach port that is currently being tracked by a dispatch source object before the cancellation handler is invoked.

### Details
Although this violation is common for all platforms, it seems that Windows is the most severely impacted. It is crucial not to close the socket too early, as doing so could result in Windows reusing the freed socket handle for a new connection. This, in turn, may lead to Dispatch handling the same socket simultaneously as both dead and alive. Also, Dispatch proactively checks for invalid sockets in critical locations (like [this](https://github.com/apple/swift-corelibs-libdispatch/blob/9566a131de09a472790f7b69ae661326cdedf306/src/event/event_windows.c#L222C5-L222C5)) and crashes immediately if an invalid socket handle is encountered. Therefore, it is highly important to prevent the socket from being closed until Dispatch has finished its operations.

[...]

PR #4854, opened 2023-12-19
Add tests for `URLSession._MultiHandle` Windows issue

lxbndr:
This adds test cases for #4791.

Test HTTPServer needs to be reconfigured (we need synchronous processing and minimal listening socket backlog) for one of the test scenarios, so there are options now.

- `test_httpTimeout` - former `test_timeoutInterval`, is adjusted to actually test for timeout (it was falsely detecting connection error as expected result)
- `test_connectTimeout` - tests a specific scenario, when the socket is timing out on connection attempt. Seems like with recent cURL library update this now works same way as http timeout, but in the past there was difference, so it is good thing to test
- `test_repeatedRequestsStress` - included only in Windows suite, does a heavy pressure on URLSession/Dispatch using small packs of simultaneous requests, repeated lot of times. There are a bit of randomness, but I find 3000 cycles enough to reproduce crash on my machine. In most of cases it fails on first 300-500 cycles.

macOS system Foundation runs these tests perfectly (btw did we lose Darwin Compatibility Tests target?), but for SwiftFoundation on macOS the `test_repeatedRequestsStress` is surprisingly way slower (I'd say x10 slower) than even on Windows. Anyway, this particular test is intended only for testing against socket reuse on Windows.

### DIFF
--- a/Tests/Foundation/HTTPServer.swift
+++ b/Tests/Foundation/HTTPServer.swift
@@ -99,7 +99,7 @@ class _TCPSocket: CustomStringConvertible {
         listening = false
     }
 
-    init(port: UInt16?) throws {
+    init(port: UInt16?, backlog: Int32) throws {
         listening = true
         self.port = 0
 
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

lxbndr:
```
/Users/ec2-user/jenkins/workspace/swift-corelibs-foundation-PR-macOS/branch-main/swift-corelibs-foundation/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c:151:70: error: initializer element is not a compile-time constant
    static CFStringRef const __CFBundleCommonLanguageNamesArray[] = {CFSTR("English"), CFSTR("French"), CFSTR("German"), CFSTR("Italian"), CFSTR("Dutch"), CFSTR("Spanish"), CFSTR("Japanese")};
```
This part works in Xcode 14.2, but not in Xcode 15. Not related to the changes in this PR.

lxbndr:
Closing, as all relevant code is merged as part of other PR

PR #4857, opened 2023-12-23
Removed unneeded bzero

AreaZR:
This prevents mis-optimization causing problems, particularly rdar://problem/23381916.


glessard:
Note that active development focus is now on [swift-foundation](https://github.com/apple/swift-foundation), rather than this project.
How was this change tested?

PR #4859, opened 2023-12-26
Cancel `DispatchSource` before closing socket (#4791)

lxbndr:
To avoid issues described in #4791 we have to follow `DispatchSource` cancel procedures in the first place. I.e. we have to prevent underlying socket close until `DispatchSource` calls a cancel handler.

Everything becomes complicated a bit, because we have multiple cases for socket and `DispatchSource` lifecycle. Cancelling Dispatch Source is asynchronous operation (with control points at the start and at the end of the process), and other actions, despite being simple and synchronous, become more complicated due to their dependency on cancel operation. Some of basic cases:
- Socket could (but not necessary would) outlive its owning `easy_handle`, as CURL caches connections in `multi_handle`
- 99% of time DispatchSource cancels quickly, but sometimes some other work on the queue squeezes in between the start and the end of cancelling. CURL could easily send a series of register-unregister-register requests, and we have to carefully handle possible overlapping.

This change aims to extend socket life by tying its lifecycle with boxing object (`_SocketReference`). Socket is not closed until `_SocketReference` is alive. While `DispatchSource` cancel process is ongoing, we're keeping such object, sharing it through a storage with the [close socket function](https://curl.se/libcurl/c/CURLOPT_CLOSESOCKETFUNCTION.html). Close socket function implementation marks `_SocketReference` as eligible for closing. If there is no ongoing Dispatch Source cancel, the reference is deinited immediately, effectively closing socket.

Also, this change is trying to be as less invasive as possible. Mostly additive, without structural changes. Major work is done by manipulating the state of `_SocketReference`. I believe there is better way to handle this, but this would probably require more extended rework of how Dispatch Sources are managed and stored.

And the fly in the ointment. This fixes most of crash scenarios on Windows, but not all. I can now say confidently that Dispatch has some flaws related to the socket processing on that platform. During stress testing I observed numerous Dispatch crashes on adding socket handle after it being reused by the system - and that is after graceful and complete cancel of corresponding DispatchSource. Luckily, it appears only on heavy load (test `test_repeatedRequestsStress` - marked as expected to fail), but it definitely affects final product and our users.

Also, this changeset includes tests from #4854 and timer fixes from #4858 (as it makes everything work more predictable). If this whole PR would be unreliable for some reason, I'd like to merge aforementioned changes separately.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
+++ b/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
@@ -45,6 +45,7 @@ extension URLSession {
         let queue: DispatchQueue
         let group = DispatchGroup()
         fileprivate var easyHandles: [_EasyHandle] = []
+        fileprivate var socketReferences: [CFURLSession_socket_t: _SocketReference] = [:]
         fileprivate var timeoutSource: _TimeoutSource? = nil
         private var reentrantInUpdateTimeoutTimer = false
         
@@ -127,13 +128,14 @@ fileprivate extension URLSession._MultiHandle {
             if let opaque = socketSourcePtr {
                 Unmanaged<_SocketSources>.fromOpaque(opaque).release()
             }
+            socketSources?.tearDown(handle: self, socket: socket, queue: queue)
             socketSources = nil
         }
         if let ss = socketSources {
             let handler = DispatchWorkItem { [weak self] in
                 self?.performAction(for: socket)
             }
-            ss.createSources(with: action, socket: socket, queue: queue, handler: handler)
+            ss.createSources(with: action, handle: self, socket: socket, queue: queue, handler: handler)
         }
         return 0
     }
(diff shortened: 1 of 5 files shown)


Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
```diff
@@ -384,13 +481,12 @@ fileprivate extension URLSession._MultiHandle {
             timeoutSource = nil
             queue.async { self.timeoutTimerFired() }
         case .milliseconds(let milliseconds):
-            if (timeoutSource == nil) || timeoutSource!.milliseconds != milliseconds {
-                //TODO: Could simply change the existing timer by using DispatchSourceTimer again.
-                let block = DispatchWorkItem { [weak self] in
-                    self?.timeoutTimerFired()
-                }
-                timeoutSource = _TimeoutSource(queue: queue, milliseconds: milliseconds, handler: block)
+            //TODO: Could simply change the existing timer by using DispatchSourceTimer again.
```
Harry-KNIGHT, line 484: Does it has to be done or one issue will be created ?
```suggestion
            // TODO: Could simply change the existing timer by using DispatchSourceTimer again.
```

Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
```diff
@@ -449,25 +545,56 @@ fileprivate class _SocketSources {
...
+            if action.needsReadSource {
+                self.createReadSource(socket: socket, queue: queue, handler: handler)
+            }
+            if action.needsWriteSource {
+                self.createWriteSource(socket: socket, queue: queue, handler: handler)
+            }
         }
-        if action.needsWriteSource {
-            createWriteSource(socket: socket, queue: queue, handler: handler)
+        if (action.needsReadSource || action.needsWriteSource) {
```
Harry-KNIGHT, line 596: Is `()` is really needed ?
```suggestion
        if action.needsReadSource || action.needsWriteSource {
```

Tests/Foundation/HTTPServer.swift
```diff
@@ -99,7 +99,7 @@ class _TCPSocket: CustomStringConvertible {
         listening = false
     }
 
-    init(port: UInt16?) throws {
+    init(port: UInt16?, backlog: Int32) throws {
```
Harry-KNIGHT, line 102: Why did you choose Int32 ?

Harry-KNIGHT (commented):
Simple review with questions, not blocking for merge.

Tests/Foundation/HTTPServer.swift
lxbndr: I propagated `backlog` parameter from `listen` function signature, and it is `Int32` there. I guess it is imported from C `int`, which is `CInt` in Swift, which, in turn, is the alias for `Int32`. 🤔 Do you think it is better to use `CInt` here instead?

Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
lxbndr: Oh, how did it get here? Should've cleaned this. Thanks!

Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
lxbndr: Can't really say on behalf on this comment author😞 This was here [from the beginning](https://github.com/apple/swift-corelibs-foundation/blob/5fdd36b193cb1f4f97b4e65c0ed5588fafea30c7/Foundation/NSURLSession/MultiHandle.swift#L337). Guess it is better to leave as is to keep the diff less noisy.

Harry-KNIGHT: Yeah, let the Todo, I thought you've added it

Tests/Foundation/HTTPServer.swift
Harry-KNIGHT: Thanks for the explanation 🙌

 No for me Int32 is the good choice, I asked only for know how did you make your choice.

shahmishal:
@swift-ci test

jrflat (approved):
Catching up on this now, LGTM, thanks for all your work on this issue!

PR #4860, opened 2024-01-03
Use CMake to check for strlcat/strlcpy

etcwilde:
Rather than try to hard-code whether or not each version of each platform has strlcat and strlcpy, this patch has CMake check to see if it's available after including string.h. If they are, then we should not redefine them. This comes after trying to build Foundation against glibc 2.38, which finally added these two functions, causing CoreFoundation to fail to compile.


etcwilde:
@swift-ci please test

parkera (approved):

etcwilde:
@swift-ci please test

compnerd (approved):

compnerd (approved):

etcwilde:
@al45tair got his merged https://github.com/apple/swift-corelibs-foundation/pull/4876
Closing this one now.

PR #4861, opened 2024-01-08
Resync URL loading system error codes

dirtyhenry:
As I was developing the `https://github.com/dirtyhenry/swift-blocks` package, I noticed that some error codes from the URL Loading System error codes were out of sync (or missing) from what I could see in Xcode and on the Foundation documentation.

I hope it helps. 🙏

### DIFF
--- a/Sources/Foundation/NSURLError.swift
+++ b/Sources/Foundation/NSURLError.swift
@@ -72,13 +72,13 @@ public var NSURLErrorNoPermissionsToReadFile: Int { return -1102 }
 public var NSURLErrorDataLengthExceedsMaximum: Int { return -1103 }
 
 // SSL errors
-public var NSURLErrorSecureConnectionFailed: Int { return -1201 }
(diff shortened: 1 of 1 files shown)


dirtyhenry:
@swift-ci test

dirtyhenry:
I would love to hear your thoughts on this pull request. 
If there's anything I missed or any concerns you have, I'm all ears!

Thanks.

compnerd:
This would be an ABI break, I'm not sure that we should be doing this ... why do these values have to match?

dirtyhenry:
Easy answer first :)

> Why do these values have to match?

To make error management consistent no matter what OS is running the code. I believe this is very much in line with the overall objective of the project ie: 

> Our **primary goal** is to achieve implementation **parity** with Foundation on Apple platforms. This will help to enable the overall Swift goal of **portability**.

> This would be an ABI break

I don't know how big of an impact this can be. Is an ABI break a deal-breaker for any PR? Or does it mean it has to wait a major version change to be taken into account?

PR #4868, opened 2024-02-07
[wasm][build] Handle `FOUNDATION_BUILD_TOOLS` value set to `OFF`

kateinoigakukun:
This is a follow up to 5e7281b993ca04486ed8f2add39c8b11f132a550, which just adds the `BUILD_TOOLS` option but doesn't actually use it.


MaxDesiatov (approved):

kateinoigakukun:
@swift-ci test

Sources/CMakeLists.txt
```diff
@@ -2,4 +2,6 @@ add_subdirectory(UUID)
 add_subdirectory(Foundation)
 add_subdirectory(FoundationNetworking)
 add_subdirectory(FoundationXML)
-add_subdirectory(Tools)
+if(BUILD_TOOLS)
```
compnerd, line 5: This is not a standard CMake flag. Can you create a proper `option` for this for something like `FOUNDATION_BUILD_TOOLS`? We can control the default for it to disable it by default on WASM targets if desired.

compnerd (changes requested):

Sources/CMakeLists.txt
kateinoigakukun: The option is already declared here as `option` https://github.com/apple/swift-corelibs-foundation/blob/7f62c8c52aeed9a86b503fe7b662c48eda5d0f8a/CMakeLists.txt#L43 but I agree the option name should be prefixed by `FOUNDATION`

kateinoigakukun:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

compnerd (approved):

PR #4866, opened 2024-02-07
[wasm][build] Guard libdispatch and Threads usage with HAS_LIBDISPATCH_API

kateinoigakukun:
This is a follow-up to 5e7281b993ca04486ed8f2add39c8b11f132a550 to make it possible to turn them off when building for that platform.

PR #4869, opened 2024-02-07
[wasm] Port BlocksRuntime for no dlfcn.h platforms

kateinoigakukun:
BlocksRuntime uses dlsym to find objc_destructInstance, which is not available on all platforms. This change adds a check for dlfcn.h and only uses dlsym if it is available and otherwise crashes the program.

The crash will not usually happen, as `_Block_use_RR` is only called from objc-auto, which is not available for such platforms.

### DIFF
--- a/Sources/BlocksRuntime/runtime.c
+++ b/Sources/BlocksRuntime/runtime.c
@@ -15,7 +15,7 @@
 #if TARGET_OS_WIN32
 #include <Windows.h>
 #include <Psapi.h>
-#else
+#elif __has_include(<dlfcn.h>)
 #include <dlfcn.h>
 #endif
 #if __has_include(<os/assumes.h>)
@@ -268,7 +268,11 @@ void _Block_use_RR( void (*retain)(const void *),
         break;
     }
 #else
+# if __has_include(<dlfcn.h>)
     _Block_destructInstance = dlsym(RTLD_DEFAULT, "objc_destructInstance");
+# else
+    _Block_destructInstance = _Block_destructInstance_default;
+# endif
 #endif
 }
 


MaxDesiatov:
@swift-ci test

Sources/BlocksRuntime/runtime.c
```diff
@@ -268,7 +268,11 @@ void _Block_use_RR( void (*retain)(const void *),
         break;
     }
 #else
+# if __has_include(<dlfcn.h>)
     _Block_destructInstance = dlsym(RTLD_DEFAULT, "objc_destructInstance");
+# else
+    assert(0 && "No way to find objc_destructInstance on this platform");
```
compnerd, line 274: Can we just set `_Block_destructInstance` to `NULL`? IIRC, `_Block_destructInstance(...)` should do nothing if the pointer is NULL (this is how weak pointers work).

kateinoigakukun: `_Block_destructInstance` can be called without null guard https://github.com/apple/swift-corelibs-foundation/blob/dbca8c7ddcfd19f7f6f6e1b60fd3ee3f748e263c/Sources/BlocksRuntime/runtime.c#L530, so `_Block_destructInstance_default` might be safer here

compnerd (approved):
Seems reasonable to me. Does this actually get used? We have too many copies of blocks runtime (compiler-rt, libdispatch, and Foundation). I believe on Windows at least, we use the libdispatch copy universally.

kateinoigakukun:
Thanks @compnerd! Right, this change needs to be also applied to those copies eventually.

compnerd:
@swift-ci please test

parkera (approved):

parkera:
```
/Users/ec2-user/jenkins/workspace/swift-corelibs-foundation-PR-macOS/branch-main/swift-corelibs-foundation/CoreFoundation/PlugIn.subproj/CFBundle_Locale.c:151:70: error: initializer element is not a compile-time constant
```

This seems to suggest the CF code is not being compiled with the correct flags on Darwin.

kateinoigakukun:
@parkera Yes, would you mind checking if the upstream gas fixed or not? https://github.com/apple/swift-corelibs-foundation/pull/4866#issuecomment-1932664319

compnerd:
@kateinoigakukun I don't think that `gas` is being used (and it is unlikely to be cctools as as well). This is most likely using the LLVM IAS.

kateinoigakukun:
@compnerd Sorry, I don't understand why you mentioned assemblers here. Would you mind elaborating a little bit more?

compnerd:
Oh, I saw the old version of the comment where `has` was typoed as `gas`.  Whoops, never mind me.

kateinoigakukun:
Ah, that's my bad, sorry for confusion 😓

kateinoigakukun:
@swift-ci test Windows platform

PR #4870, opened 2024-02-07
[wasm] Port thread-related APIs for no thread platform

kateinoigakukun:
This change ports the thread-related APIs for the no thread platform by adding no-op implementations. Note that WASI is going to support threads in the near future, so this does not guard by `TARGET_OS_WASI` but by `!SWIFT_CORELIBS_FOUNDATION_HAS_THREADS` to make it easier for future WASI + threads support.

### DIFF
--- a/CoreFoundation/Base.subproj/CFInternal.h
+++ b/CoreFoundation/Base.subproj/CFInternal.h
@@ -654,6 +654,37 @@ CF_INLINE int _CFRecursiveMutexUnlock(_CFRecursiveMutex *mutex) {
   LeaveCriticalSection(mutex);
   return 0;
 }
+#elif TARGET_OS_WASI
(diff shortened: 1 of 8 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov (approved):

compnerd:
This seems to have caused a regression on macOS builds

kateinoigakukun:
@compnerd I apologize for that breakage. I made a fix here https://github.com/apple/swift-corelibs-foundation/pull/4888

PR #4872, opened 2024-02-14
Stop assuming `CFSTR` is a constant-time expression with Clang 15

kateinoigakukun:
From Clang 15, nested static initializer inside statement-expression is no longer a constant-time expression (See https://reviews.llvm.org/D127201). OSS Foundation defines `CFSTR` as a macro rather than `__builtin___CFStringMakeConstantString` and it uses nested static initializer inside statement-expression, so we can't assume `CFSTR` itself is always a constant-time expression.
This patch removes some `static` qualifiers associated with `CFSTR` to make them acceptable with Clang 15 and later.

I confirmed there is no change at LLVM IR level between before/after this patch with Clang 14 and `-O3`, so I assume Clang 15 and later also can optimize those non-static variables to be static global rodata. So this fix does not cause any performance regression.

This issue was originally found in https://github.com/apple/swift-corelibs-foundation/pull/4866#issuecomment-1932633776

### DIFF
--- a/CoreFoundation/Locale.subproj/CFLocale.c
+++ b/CoreFoundation/Locale.subproj/CFLocale.c
@@ -1274,7 +1274,7 @@ static bool __CFLocaleCopyLocaleID(CFLocaleRef locale, bool user, CFTypeRef *cf,
 
 
 static bool __CFLocaleCopyCodes(CFLocaleRef locale, bool user, CFTypeRef *cf, CFStringRef context) {
-    static CFStringRef const kCFLocaleCodesKey = CFSTR("__kCFLocaleCodes");
+    CFStringRef const kCFLocaleCodesKey = CFSTR("__kCFLocaleCodes");
     
     bool codesWasAllocated = false;
     CFDictionaryRef codes = NULL;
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci Please test

kateinoigakukun:
The original issue seems fixed, but XCTest is still failing due to another issue the version of macOS of the Jenkins node is older than the compiler's default deployment target.

https://ci.swift.org/job/swift-corelibs-foundation-PR-macOS/364/console
```
# command stderr:
dyld[47446]: Symbol not found: _swift_getTypeByMangledNameInContext2
  Referenced from: <7B2821A0-0F0D-3D4D-B6AA-B823A43D0D5E> /Users/ec2-user/jenkins/workspace/swift-corelibs-foundation-PR-macOS/branch-main/swift-corelibs-xctest/Tests/Functional/TestCaseLifecycle/Output/TestCaseLifecycle (built for macOS 14.2 which is newer than running OS)
  Expected in:     <087DADAD-D590-35EA-AC4D-7B0E966804EB> /usr/lib/swift/libswiftCore.dylib

error: command failed with exit status: -6
```

This failures will be fixed by https://github.com/apple/swift-corelibs-foundation/pull/4873

parkera (approved):

kateinoigakukun:
Thanks!

PR #4873, opened 2024-02-14
Bump the macOS deployment target to 10.13 in the Xcode project.

kateinoigakukun:
Xcode 15 no longer recognizes 10.12 as a valid deployment target and just falls back to 14.2, which is too new for our CI. XCTest side has already done a similar change https://github.com/apple/swift-corelibs-xctest/pull/459


MaxDesiatov:
@swift-ci test

kateinoigakukun:
Seems like https://github.com/apple/swift-corelibs-xctest/pull/476 is also needed

PR #4881, opened 2024-02-23
[wasm] Unlock cwd-related APIs

kateinoigakukun:
wasi-libc now has user-land implementations of `getcwd`, so we don't need to guard cwd-related APIs.


MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

PR #4884, opened 2024-02-25
Add required to NSObject

Kyle-Ye:
Close https://github.com/apple/swift/issues/71874


Kyle-Ye:
@swift-ci test

Kyle-Ye:
This patch aligns the behavior on Darwin's ObjectiveC NSObject implementation but it will be source breaking for the downstream user.

Can we add it as part of Swift 6?

PR #4886, opened 2024-02-27
Stream: add missing include for macOS builds

compnerd:
From the swift-corelibs-xctest CI:
```
swift-corelibs-foundation/CoreFoundation/Stream.subproj/CFStream.c:1686:5: error: call to undeclared function '_CFThreadSetName'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    _CFThreadSetName(pthread_self(), "com.apple.CFStream.LegacyThread");
    ^
```

PR #4888, opened 2024-02-28
Repair the macOS build by adding missing `-D` in Xcode project

kateinoigakukun:
macOS build was broken by 3744d4d2df6b43f87fbc075b91296f4d0c94ab0f, which introduced a new constant definition but did not add it to the Xcode project.

### DIFF
--- a/Foundation.xcodeproj/project.pbxproj
+++ b/Foundation.xcodeproj/project.pbxproj
@@ -3966,6 +3966,7 @@
 					"-DCF_BUILDING_CF",
 					"-DDEPLOYMENT_RUNTIME_SWIFT",
 					"-DDEPLOYMENT_ENABLE_LIBDISPATCH",
+					"-DSWIFT_CORELIBS_FOUNDATION_HAS_THREADS",
 					"-DHAVE_STRUCT_TIMESPEC",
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
The macOS job is still failing but the build failure regression has been resolved https://ci.swift.org/job/swift-corelibs-foundation-PR-macOS/391/

compnerd:
I have a similar change in https://github.com/apple/swift-corelibs-foundation/pull/4886. I don't know which of the two is better - I think that I went through all the cases and added the macro (need to add the Swift version). I'm more interested in getting this fixed rather than which PR we merge.

compnerd:
@swift-ci please test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
I think #4886 needs `-Xcc` too to tell the definition to ClangImporter. I've confirmed this change repaired the build https://github.com/apple/swift-corelibs-foundation/pull/4888#issuecomment-1970146173, so let's merge this after the other required status to be green.

compnerd:
Going to merge this to repair the build. We should investigate how to fix the macOS builds.

PR #4890, opened 2024-03-01
[wasm] Port `CoreFoundation/Base.subproj/CFFileUtilities.c`

kateinoigakukun:
This patch makes a part of `CoreFoundation/Base.subproj/CFFileUtilities.c` compilation work with wasi-libc


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #4891, opened 2024-03-01
[wasm] Port ProcessInfo for WASI platform

kateinoigakukun:
This patch makes a part of `PocessInfo` APIs work with wasi-libc. On WASI, it gets program name from argv[0].


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #4892, opened 2024-03-01
[wasm] Port use of `dirent` for wasi-libc

kateinoigakukun:
wasi-libc defines `d_name` as "flexible array member" which is not supported by ClangImporter yet. This patch replaces the use of `d_name` with a helper function `_direntName` to work around the issue.


kateinoigakukun:
@swift-ci test

MaxDesiatov:
This doesn't seem to be exclusive to WASI `#if` branches, so I'll defer to @parkera.

MaxDesiatov (approved):
On a second look this looks like a safe enough change, especially to unblock ongoing WASI support.

PR #4894, opened 2024-03-01
[wasm] Fix definition conflict of `errno`

kateinoigakukun:
WASILibc's `errno` conflicts with Clang-imported `errno` in <errno.h>.


MaxDesiatov:
@swift-ci test

PR #4895, opened 2024-03-01
[wasm] Add objc_retainAutoreleasedReturnValue definition

kateinoigakukun:
The function is usually provided by libdispatch, but for platforms where libdispatch is unavailable at this moment, we need to provide it in CF as we do for BlocksRuntime.

### DIFF
--- a/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/CoreFoundation/Base.subproj/CFRuntime.c
@@ -1787,6 +1787,18 @@ struct _NSCFXMLBridgeUntyped __NSCFXMLBridgeUntyped = {
   &kCFErrorLocalizedDescriptionKey,
 };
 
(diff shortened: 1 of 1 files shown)


al45tair:
I'm not _hugely_ keen on propagating this hack into `swift-corelibs-foundation` or (worse) `swift-foundation`. We've known this was a problem since 2015, and nobody's done anything about it, presumably because "the workaround was good enough"; it needs to be fixed in the compiler.

al45tair:
I think the compiler change needed here is literally a three line fix (I'm just trying it now to see).

al45tair:
Hopefully https://github.com/apple/swift/pull/72011 will fix this problem and then we won't need this (and can delete the corresponding code from `swift-corelibs-libdispatch` too).

kateinoigakukun:
This workaround is no longer necessary!

PR #4896, opened 2024-03-02
[wasm] Provide stub for `_NSCreateTemporaryFile` on WASI

kateinoigakukun:
WASI does not have temp directories and wasi-libc does not provide `mkstemp`, so provide a stub implementation that throws `ENOTSUP`.

### DIFF
--- a/Sources/Foundation/NSPathUtilities.swift
+++ b/Sources/Foundation/NSPathUtilities.swift
@@ -779,6 +779,10 @@ internal func _NSCreateTemporaryFile(_ filePath: String) throws -> (Int32, Strin
     }
     // Don't close h, fd is transferred ownership
     let fd = _open_osfhandle(intptr_t(bitPattern: h), 0)
+    return (fd, pathResult)
+#elseif os(WASI)
+    // WASI does not have temp directories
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

Sources/Foundation/NSPathUtilities.swift
```diff
@@ -814,8 +818,8 @@ internal func _NSCreateTemporaryFile(_ filePath: String) throws -> (Int32, Strin
         close(fd)
         throw _NSErrorWithErrno(_errno, reading: false, path: pathResult)
     }
-#endif
     return (fd, pathResult)
+#endif
```
MaxDesiatov, line 822: this change seems to be impacting non-WASI `#if` branch?

kateinoigakukun: The `return (fd, pathResult)` line was shared among all cases, but it's now cloned into `os(Windows)` and `!os(Windows) && !os(WASI)` cases. So there is no semantic change.

MaxDesiatov (approved):

PR #4897, opened 2024-03-02
[wasm] Port `CoreFoundation/StringEncodings.subproj`

kateinoigakukun:
This commit adds `TARGET_OS_WASI` as ICU supported platform in `#if` conditions under `StringEncodings.subproj`


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #4899, opened 2024-03-02
[wasm] Include syslog.h only if available

kateinoigakukun:
wasi-libc does not provide `syslog.h`, so we should only include it if it is available. syslog APIs are only used for debugging code guarded under `DEBUG_URL_MEMORY_USAGE`, so it is safe to exclude it.


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

PR #4902, opened 2024-03-04
[wasm] Port CoreFoundation/String.subproj

kateinoigakukun:
- LLVM's AsmParser for WebAssembly requires explicit data segment size in `.S`
- Add `TARGET_OS_WASI` branch as Unix-y platform


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #4903, opened 2024-03-04
Fix race condition in CFPreferences

hristost:
In __CFWriteBytesToFileWithAtomicity(), we first write the contents of the plist to an auxiliary copy file, and then move that copy to where the original used to be. Because that file needs to have the same owner as the original, we would use chown() to change ownership as the last step. This allows a race condition where the new file is in its final location, but doesn't have the correct permissions. To fix this, call chown() on the file before moving.

rdar://121597642


hristost:
@swift-ci please test

parkera (approved):

PR #4905, opened 2024-03-07
[wasm] Guard out the use of `libdispatch` types

kateinoigakukun:
To avoid the use of `libdispatch` types in the `CoreFoundation` module when unavailable


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #4906, opened 2024-03-07
[wasm] Grab wasi-libc constants manually

kateinoigakukun:
Unfortunately they are defined in the way ClangImporter can't handle them, so we need to grab them manually through function calls. Those calls are usually inlined, so the performance impact should be minimal.

This workaround can be removed after https://github.com/apple/swift/issues/43102 will be "fixed"


MaxDesiatov:
@swift-ci test

kateinoigakukun:
@swift-ci test

PR #4909, opened 2024-03-07
[wasm] Port CFStream for no-dispatch platforms

kateinoigakukun:
This patch guards the use of RunLoop, which is not available on no-dispatch platforms. `CFStream` is still useful without RunLoop integration to port some APIs like `PropertyListSerialization`.


MaxDesiatov (approved):

kateinoigakukun:
@swift-ci test

PR #4911, opened 2024-03-08
[wasm] Guard RunLoop.subproj sources with `__HAS_DISPATCH__`

kateinoigakukun:
Guard out the whole source code for no-dispatch platforms. Excluding them from the build at all is one of the options, but it introduces some complexities in CMake build script just for the specific target, so I prefer this way.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

kateinoigakukun:
@swift-ci test Linux platform

PR #4912, opened 2024-03-08
[wasm] Port CoreFoundation/CMakeLists.txt

kateinoigakukun:
* CFXMLInterface usually depends on BlocksRuntime through dispatch indirectly, so we need to link it explicitly for no-dispatch builds to include BlocksRuntime headers.
* Enable some wasi-libc emulations for building CF.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #4925, opened 2024-03-15
Use `__attribute__ ((constructor))` instead of `INIT_ROUTINE` on macOS (Xcode)

ikesyo:
I'm not sure why but it looks like `__CFInitialize` is not called on macOS with Xcode.

ref: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Tasks/InitializingFrameworks.html

### DIFF
--- a/CoreFoundation/Base.subproj/CFRuntime.c
+++ b/CoreFoundation/Base.subproj/CFRuntime.c
@@ -1164,7 +1164,7 @@ _CFThreadRef _CF_pthread_main_thread_np(void) {
 
 
 
-#if TARGET_OS_LINUX || TARGET_OS_BSD || TARGET_OS_WASI
+#if TARGET_OS_MAC || TARGET_OS_LINUX || TARGET_OS_BSD || TARGET_OS_WASI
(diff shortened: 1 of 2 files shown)


ikesyo:
@swift-ci test

compnerd:

ikesyo:
https://github.com/apple/swift-corelibs-xctest/pull/476

grynspan:
Static constructors are considered a performance issue on Darwin and are generally avoided. @parkera would this be an acceptable change? Is it even necessary given that corelibs-foundation isn't typically used on Darwin?

compnerd:
It is useful to be able to build and run foundation on macos for testing behavioral issues. I would say that since it didn't normally run in macos, this is likely fine.

ikesyo:
I found another solution, using the classic linker by adding `-Wl,-ld_classic` to `OTHER_LDFLAGS` did solve the problem 🤔

parkera:
CFInitialize simply has to run, and although we don't encourage the use of static initializers in general, we make an exception for this particular one.

I think I would rather give it the source code annotation than force a compatibility mode in the linker.

PR #4931, opened 2024-03-29
Resolve -Wno-int-conversion warnings

jmschonfeld:
This resolves a few warnings from `-Wno-int-conversion` which allows us to remove this flag so that those can become errors by default now that the (valid) issues are resolved.


xedin (approved):

parkera:
Looks good to me. We could actually target this at `main` then keep merging `main` to the `package` branch.

parkera (approved):

PR #4933, opened 2024-03-29
Fix Package test failures for Linux

jmschonfeld:
This change resolves just about all of the current test failures for the SCF package on Linux. This change does the following:

- Removes all `allTests` declarations and moves any conditional logic within those declarations to the tests themselves (replacing conditional compilation in `allTests` with conditional compilation of an `XCTSkip` in the test itself). This resolved a _lot_ of errors caused by tests running when they shouldn't be
- Fixes usages of `Bundle.main` to point correctly to `Bundle.module` when looking for resources (and ensures we still use `Bundle.main` when looking for executables
- Fixes a few minor issues in the implementation (mainly a typo in a string, and a typo in `NSOrderedSet` that wasn't being fully tested)

With these changes, running `swift test` for this package on linux results in only one test failure due to a behavioral difference between SCF and swift-foundation which will be resolved by https://github.com/apple/swift-foundation/pull/515


parkera (approved):
I think the proof for this one is going to be in the tests actually running. :)

PR #4937, opened 2024-04-23
URLSession: Added support for client certificate authentication for non MacOS platforms

oliviermartin:
Swift Foundation Networking does not currently support client certificate authentication which is quite a limitation when integrating with a more complex system.
For MacOS/iOS based platform, the client certificate authentication is done through [URLSessionDelegate](https://developer.apple.com/documentation/foundation/urlsessiondelegate) that handles authentication challenges.

Swift Foundation Networking relies on `libcurl` for `URLSession`. This support does not go through `URLSessionDelegate` for authentication challenge. The approach used by this pull-request is:
- Extend `URLCredential` to also pass client private key and certificate
- Pass the `URLCredential` through `URLSessionConfiguration`. Note: on MacOS/iOS this approach could also be used as the `URLSessionDelegate` is often a copy/paste to pass `URLCredential` back to `URLSession`.
- Configure libCurl to use these client private key and certificate

There is no unittest for this new code as it would require Swift to support server with TLS client certificate authentication support. The code can be locally tested with [`openssl s_server`](https://www.openssl.org/docs/man1.1.1/man1/s_server.html).

Here is an example code to use this API:

```
    func test_clientCertificateAuthentication() {
        let sem = DispatchSemaphore.init(value: 0)

        let privateClientKey: Data
        let privateClientCertificate: Data
        do {
            // TestPrivateKey.key contains the DER format of the private key
            privateClientKey = try Data(contentsOf: URL(fileURLWithPath: "TestPrivateKey.key"))
            // TestClientCertificate.der contains the DER format of the client certificate
            privateClientCertificate = try Data(contentsOf: URL(fileURLWithPath: "TestClientCertificate.der"))
        } catch {
            print("Failed to load file: \(error)")
            return
        }
        let urlCredential = URLCredential(clientKey: privateClientKey, clientCertificate: privateClientCertificate, persistence: .none)
        var urlSessionConfiguration = URLSessionConfiguration()
        urlSessionConfiguration.clientCredential = urlCredential

        let urlString = "https://127.0.0.1:443"
        let url = URL(string: urlString)!
        let session = URLSession(configuration: urlSessionConfiguration)

        let task = session.dataTask(with: url) { data, response, error in
            guard let response = response as? HTTPURLResponse else {
                print("No response from server:\(error)")
                defer { sem.signal() }
                return
            }
            print("Response from server: \(response.statusCode) data:\(data) error:\(error)")
            defer { sem.signal() }
        }
        task.resume()

        sem.wait()
    }
```

Limitations of this support:
- The private key cannot be stored in a secure vault. There is no such API in Swift "Foundation" to handle this generic support.
- As mentioned before, it is a different approach as the one used for MacOS/iOS that is based on `URLSessionDelegate`. But for using the same approach, we would either need to fork `libcurl` in `swift-corelibs-foundation` to handle the authentication challenge in Swift or we would need to a different library "backend" to handle `URLSession`

### DIFF
--- a/Sources/FoundationNetworking/URLSession/Configuration.swift
+++ b/Sources/FoundationNetworking/URLSession/Configuration.swift
@@ -78,6 +78,9 @@ internal extension URLSession {
         let shouldUseExtendedBackgroundIdleMode: Bool
         
         let protocolClasses: [AnyClass]?
+
+        /// The credentials to use for connecting to servers
+        let clientCredential: URLCredential?
     }
 }
 internal extension URLSession._Configuration {
@@ -100,6 +103,7 @@ internal extension URLSession._Configuration {
         urlCache = config.urlCache
         shouldUseExtendedBackgroundIdleMode = config.shouldUseExtendedBackgroundIdleMode
         protocolClasses = config.protocolClasses
+        clientCredential = config.clientCredential
     }
 }
 
--- a/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
+++ b/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
@@ -328,7 +328,14 @@ internal class _HTTPURLProtocol: _NativeProtocol {
         }
         let session = task?.session as! URLSession
         let _config = session._configuration
-        easyHandle.set(sessionConfig: _config)
+        do {
+            try easyHandle.set(sessionConfig: _config)
+        } catch {
+            self.internalState = .transferFailed
+            let nsError = error as? NSError ?? NSError(domain: NSURLErrorDomain, code: NSURLErrorUserAuthenticationRequired)
+            failWith(error: nsError, request: request)
+            return
+        }
         easyHandle.setAllowedProtocolsToHTTPAndHTTPS()
         easyHandle.set(preferredReceiveBufferSize: Int.max)
         do {
--- a/Sources/FoundationNetworking/URLSession/URLSessionConfiguration.swift
+++ b/Sources/FoundationNetworking/URLSession/URLSessionConfiguration.swift
@@ -75,9 +75,10 @@ open class URLSessionConfiguration : NSObject, NSCopying {
                   urlCredentialStorage: .shared,
                   urlCache: .shared,
                   shouldUseExtendedBackgroundIdleMode: false,
-                  protocolClasses: [_HTTPURLProtocol.self, _FTPURLProtocol.self, _WebSocketURLProtocol.self])
+                  protocolClasses: [_HTTPURLProtocol.self, _FTPURLProtocol.self, _WebSocketURLProtocol.self],
+                  clientCredential: nil)
     }
-    
+
     private init(identifier: String?,
                  requestCachePolicy: URLRequest.CachePolicy,
                  timeoutIntervalForRequest: TimeInterval,
@@ -95,7 +96,8 @@ open class URLSessionConfiguration : NSObject, NSCopying {
                  urlCredentialStorage: URLCredentialStorage?,
                  urlCache: URLCache?,
                  shouldUseExtendedBackgroundIdleMode: Bool,
-                 protocolClasses: [AnyClass]?)
+                 protocolClasses: [AnyClass]?,
+                 clientCredential: URLCredential?)
     {
         self.identifier = identifier
         self.requestCachePolicy = requestCachePolicy
@@ -115,6 +117,7 @@ open class URLSessionConfiguration : NSObject, NSCopying {
         self.urlCache = urlCache
         self.shouldUseExtendedBackgroundIdleMode = shouldUseExtendedBackgroundIdleMode
         self.protocolClasses = protocolClasses
+        self.clientCredential = clientCredential
     }
     
     open override func copy() -> Any {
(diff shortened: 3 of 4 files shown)


parkera:
@swift-ci test

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -111,6 +111,14 @@ CFURLSessionEasyCode CFURLSession_easy_setopt_tc(CFURLSessionEasyHandle _Nonnull
     return MakeEasyCode(curl_easy_setopt(curl,  option.value, a));
 }
 
+CFURLSessionEasyCode CFURLSession_easy_setopt_blob(CFURLSessionEasyHandle _Nonnull curl, CFURLSessionOption option, void* data, size_t len) {
+    struct curl_blob stblob;
```
jrflat, line 115: This fails to build in CI because `struct curl_blob` (and `CURLOPT_SSLKEY_BLOB` and friends) are only available in libcurl >= 7.71.0. Let's add a macro to guard this, and do nothing/return `CFURLSessionEasyCodeOK` otherwise. (PR #4877 has a good example with the `NS_CURL_MISSING_CURLINFO_CAINFO` macro)

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -366,6 +374,7 @@ CFURLSessionOption const CFURLSessionOptionCOOKIE = { CURLOPT_COOKIE };
 CFURLSessionOption const CFURLSessionOptionHTTPHEADER = { CURLOPT_HTTPHEADER };
 CFURLSessionOption const CFURLSessionOptionHTTPPOST = { CURLOPT_HTTPPOST };
 CFURLSessionOption const CFURLSessionOptionSSLCERT = { CURLOPT_SSLCERT };
+CFURLSessionOption const CFURLSessionOptionSSLCERT_BLOB = { CURLOPT_SSLCERT_BLOB };
```
jrflat, line 377: Same comment as above, let's guard this option and `CURLOPT_SSLKEY_BLOB` below for libcurl versions that don't support it

CoreFoundation/URL.subproj/CFURLSessionInterface.c
```diff
@@ -111,6 +111,14 @@ CFURLSessionEasyCode CFURLSession_easy_setopt_tc(CFURLSessionEasyHandle _Nonnull
     return MakeEasyCode(curl_easy_setopt(curl,  option.value, a));
 }
 
+CFURLSessionEasyCode CFURLSession_easy_setopt_blob(CFURLSessionEasyHandle _Nonnull curl, CFURLSessionOption option, void* data, size_t len) {
```
jrflat, line 114: Let's mark this `void *_Nonnull data` and only call it with non-nil data to silence a compiler warning

Sources/FoundationNetworking/URLCredential.swift
```diff
@@ -91,9 +114,14 @@ open class URLCredential : NSObject, NSSecureCoding, NSCopying {
         guard aCoder.allowsKeyedCoding else {
             preconditionFailure("Unkeyed coding is unsupported.")
         }
-        
-        aCoder.encode(self._user._bridgeToObjectiveC(), forKey: "NS._user")
-        aCoder.encode(self._password._bridgeToObjectiveC(), forKey: "NS._password")
+
+        if let user = self._user {
+            aCoder.encode(user._bridgeToObjectiveC(), forKey: "NS._user")
```
jrflat, line 119: Let's make sure we also change `init?(coder:)` above

```
let encodedUser = aDecoder.decodeObject(forKey: "NS._user") as! NSString
let encodedPassword = aDecoder.decodeObject(forKey: "NS._password") as! NSString
```

to no longer force-unwrap, since we may no longer be encoding `nil` user or password, e.g. we could do

```
if let encodedUser = aDecoder.decodeObject(forKey: "NS._user") as? NSString {
  self._user = bridgeString(encodedUser)!
}
...
```

jrflat (commented):
Thanks for the work to add this! I believe adding new public API like this would need to go through an API review process, but we might also be able to accomplish this by implementing existing Darwin `URLCredential` API and adding smalls stubs for `SecIdentity`/`SecCertificate`. I'll need to give that a bit more thought, but in the meantime, I've left a few comments that should hopefully help address the build issues/other minor feedback :)

travarin:
> Pass the URLCredential through URLSessionConfiguration
I think it'd be better to set this on the task instead of the session configuration. Generally the client would use the same session + configuration for every request they make, but might not necessarily want to enable client certificate authentication on all of them.

oliviermartin:
Thanks @jrflat and @travarin for the feedback! I have just updated the PR  with your suggestions.

Reading [`URLSession`](https://developer.apple.com/documentation/foundation/urlsession) made me understand a bit better the philosophy behind this design (`URLSession` vs `URLSessionTask`). My understanding came initially from the examples of client certificate authentication I found on Internet that often implement `URLSessionDelegate` instead of `URLSessionTaskDelegate` for their client certificate authentication - and probably create one `URLSession` per URL.

Just on few points you mentioned in your comments:
- With the current implementation (using curl on non-Darwin system), we would need to fork+change+embed curl code in Swift Foundation library to hook `URLSession(Task)Delegate` into libcurl code whether we wanted to follow the same approach as Darwin system (ie: using `func urlSession(URLSession, task: URLSessionTask, didReceive: URLAuthenticationChallenge, completionHandler: (URLSession.AuthChallengeDisposition, URLCredential?) -> Void)` for client authentication).
- An alternative would be to add a new hook into `URLSessionTaskDelegate` (and `URLSessionDelegate()`) to pass URLCredential to URLSession backend before any connection - something like: `func urlSession(URLSession, task: URLSessionTask, getCredential: (URLCredential) -> Void)`

I guess the next step is for me to create a proposal in swift-evolution?

PR #4949, opened 2024-05-05
🍒 [5.10] Check for strlcpy/strlcat

etcwilde:
Some newer builds of glibc have strlcpy and strlcat defined, resulting in build failures on newer Linux distros. Rather than try to hard-code whether or not each version of each platform has strlcat and strlcpy, this patch has CMake check to see if it's available after including string.h. If they are, then we should not redefine them.

This is a targeted change pulling the fix from the patch set merged in 5a1db791ab824e29dc45ce08f729d3541f0517d0.


etcwilde:
@swift-ci please test

al45tair (approved):
LGTM

etcwilde:
@swift-ci please clean test Linux Platform

etcwilde:
@swift-ci please test

etcwilde:
`TARGET_OS_MAC` is _always_ defined. It might be `0` or `1`, but `#ifndef` doesn't work.... Bah!

etcwilde:
@swift-ci please test Windows

PR #4951, opened 2024-05-09
Workaround for https://github.com/apple/swift/issues/73532

hjyamauchi:
Increase the initial size of __CFReadSocketsFds to reduce the chance of resizing it which appears to lead to compiler hangs.


hjyamauchi:
CC @compnerd

tristanlabelle (changes requested):
The issue that @lxbndr pointed to in [this comment](https://github.com/apple/swift/issues/73532#issuecomment-2104277961) means that the CFSocket code is mixing up its handling of `fd_set` on Windows:
- Platform macros FD_SET and FD_ISSET will know that `fd_set` is a struct on Windows,
- CFSocket code will assume it is a bitset for resizing purposes.

The change here will create an allocation big enough so that the CFSocket bitlogic doesn't think it has to resize the allocation, but it will still be doing incorrect bit-level logic on the fd_set. I think we should build on @lxbndr's suggested fix. It would help to avoid defining `fd_mask` on Windows at all to find the places where we are treating an `fd_set` as a bitfield.

hjyamauchi:
Closing in favor of https://github.com/apple/swift-corelibs-foundation/pull/4954

PR #4952, opened 2024-05-09
CoreFoundation modularization fixes

parkera:
The tests do not yet run on Darwin, but it does pass tests for other platforms.

This patch also updates the contents of the CF Unicode data tables, bringing them up to date.

The majority of this patch is probably unreviewable. However I think the Package.swift and other configuration diffs can be reviewed.

Note: This patch does away with the name "SwiftFoundation" on macOS. It's just "Foundation" everywhere now. That means we need to avoid bringing in the system Foundation when we run executables that use this library, including the tests.


Package.swift
```diff
@@ -29,11 +28,26 @@ let buildSettings: [CSetting] = [
...
         "-fno-common",
         "-fcf-runtime-abi=swift",
+        "-include",
+        "\(Context.packageDirectory)/Sources/CoreFoundation/internalInclude/CoreFoundation_Prefix.h",
         // /EHsc for Windows
     ]),
     .unsafeFlags(["-I/usr/lib/swift"], .when(platforms: [.linux, .android])) // dispatch
 ]
 
+#if os(macOS)
```
jmschonfeld, line 38: nit: I think this can also be a `.when(platforms: [.macOS])` on the `.unsafeFlags` below to avoid the ifdefs here if we want

Package.swift
```diff
@@ -63,6 +80,11 @@ let interfaceBuildSettings: [CSetting] = [
     .unsafeFlags(["-I/usr/lib/swift"], .when(platforms: [.linux, .android])) // dispatch
 ]
 
+let swiftBuildSettings: [SwiftSetting] = [
+    .define("DEPLOYMENT_RUNTIME_SWIFT"),
```
jmschonfeld, line 84: It looks like we define this in Swift for all platforms but in C only for Linux - just checking, is that intended?

parkera: It may not actually matter... `CFAvailability.h` defines it now too:

```
#ifndef DEPLOYMENT_RUNTIME_SWIFT
#define DEPLOYMENT_RUNTIME_SWIFT 1
#endif
```

jmschonfeld (approved):

Package.swift
```diff
@@ -63,64 +71,100 @@ let interfaceBuildSettings: [CSetting] = [
...
+let products : [PackageDescription.Product] = [
+    .library(name: "SwiftFoundation", targets: ["Foundation"]),
+    .library(name: "SwiftFoundationXML", targets: ["FoundationXML"]),
+    .library(name: "SwiftFoundationNetworking", targets: ["FoundationNetworking"]),
+    .executable(name: "plutil", targets: ["plutil"]),
+]
+
+let foundationName = "SwiftFoundation"
+let foundationXMLName = "SwiftFoundationXML"
+let foundationNetworkingName = "SwiftFoundationNetworking"
```
iCharlesHu, line 89: Do we still need to do this if we are no longer build on macOS? Mainly if we decide to use these names Cmake needs to match as well.

iCharlesHu (approved):

Package.swift
parkera: No, I think I will revert this part.

parkera:
I've retitled this PR to reflect what it became in the end - fixes for CoreFoundation's modularization.

Package.swift
```diff
@@ -90,37 +94,37 @@ let package = Package(
...
             path: "Sources/Foundation",
-            swiftSettings:  [.define("DEPLOYMENT_RUNTIME_SWIFT"), .define("SWIFT_CORELIBS_FOUNDATION_HAS_THREADS")]
+            swiftSettings: swiftBuildSettings
         ),
         .target(
             name: "FoundationXML",
             dependencies: [
                 .product(name: "FoundationEssentials", package: "swift-foundation"),
-                "Foundation",
+                .targetItem(name: "Foundation", condition: nil),
```
jmschonfeld, line 103: For my context, is there a difference between these two spellings?

Package.swift
```diff
@@ -158,33 +162,36 @@ let package = Package(
...
+                .targetItem(name: "Foundation", condition: nil)
+            ]
         ),
         .executableTarget(
             name: "xdgTestHelper",
             dependencies: [
-                "Foundation",
-                "FoundationNetworking"
+                .targetItem(name: "Foundation", condition: nil),
+                .targetItem(name: "FoundationXML", condition: nil),
```
jmschonfeld, line 173: I think this dependency is new. I don't expect it to be a problem, but was it intentional?

jmschonfeld (approved):

Package.swift
parkera: Oh - I missed this, sorry. It was required when I used a `String` instead of a string literal for the target name. We can revert it back now.

Package.swift
parkera: Yup, it was intentional.

PR #4954, opened 2024-05-12
[Windows] Use `fd_set` according to Winsock2 specifics

lxbndr:
Fixes https://github.com/apple/swift/issues/73532.

On Windows, socket handles in a `fd_set` are not represented as bit flags as in Berkeley sockets. While we have no `fd_set` dynamic growth in this implementation, the `FD_SETSIZE` defined as 1024 in `CoreFoundation_Prefix.h` should be enough for majority of tasks.

This also includes a test for `SocketPort` which uses `CFSocket` internally. This test fails (hangs or crashes with memory corruption error) without the fix.

### DIFF
--- a/CoreFoundation/RunLoop.subproj/CFSocket.c
+++ b/CoreFoundation/RunLoop.subproj/CFSocket.c
@@ -213,13 +213,30 @@ CF_INLINE int __CFSocketLastError(void) {
 }
 
 CF_INLINE CFIndex __CFSocketFdGetSize(CFDataRef fdSet) {
+#if TARGET_OS_WIN32
+    if (CFDataGetLength(fdSet) == 0) {
+        return 0;
+    }
+    return FD_SETSIZE;
+#else
     return NBBY * CFDataGetLength(fdSet);
+#endif
 }
 
 CF_INLINE Boolean __CFSocketFdSet(CFSocketNativeHandle sock, CFMutableDataRef fdSet) {
     /* returns true if a change occurred, false otherwise */
     Boolean retval = false;
     if (INVALID_SOCKET != sock && 0 <= sock) {
+        fd_set *fds;
+#if TARGET_OS_WIN32
+        if (CFDataGetLength(fdSet) == 0) {
+            CFDataIncreaseLength(fdSet, sizeof(fd_set));
+            fds = (fd_set *)CFDataGetMutableBytePtr(fdSet);
+            FD_ZERO(fds);
+        } else {
+            fds = (fd_set *)CFDataGetMutableBytePtr(fdSet);
+        }
+#else
         CFIndex numFds = NBBY * CFDataGetLength(fdSet);
         fd_mask *fds_bits;
         if (sock >= numFds) {
@@ -230,9 +247,11 @@ CF_INLINE Boolean __CFSocketFdSet(CFSocketNativeHandle sock, CFMutableDataRef fd
         } else {
             fds_bits = (fd_mask *)CFDataGetMutableBytePtr(fdSet);
         }
-        if (!FD_ISSET(sock, (fd_set *)fds_bits)) {
+        fds = (fd_set *)fds_bits;
+#endif
+        if (!FD_ISSET(sock, fds)) {
             retval = true;
-            FD_SET(sock, (fd_set *)fds_bits);
+            FD_SET(sock, fds);
         }
     }
     return retval;
@@ -416,6 +435,15 @@ CF_INLINE Boolean __CFSocketFdClr(CFSocketNativeHandle sock, CFMutableDataRef fd
     /* returns true if a change occurred, false otherwise */
     Boolean retval = false;
     if (INVALID_SOCKET != sock && 0 <= sock) {
+#if TARGET_OS_WIN32
+        if (CFDataGetLength(fdSet) > 0) {
+            fd_set *fds = (fd_set *)CFDataGetMutableBytePtr(fdSet);
+            if (FD_ISSET(sock, fds)) {
+                retval = true;
+                FD_CLR(sock, fds);
+            }
+        }
+#else
         CFIndex numFds = NBBY * CFDataGetLength(fdSet);
         fd_mask *fds_bits;
         if (sock < numFds) {
@@ -425,6 +453,7 @@ CF_INLINE Boolean __CFSocketFdClr(CFSocketNativeHandle sock, CFMutableDataRef fd
                 FD_CLR(sock, (fd_set *)fds_bits);
             }
         }
+#endif
     }
     return retval;
 }
@@ -1188,6 +1217,27 @@ static void
 clearInvalidFileDescriptors(CFMutableDataRef d)
 {
     if (d) {
+#if TARGET_OS_WIN32
+        if (CFDataGetLength(d) == 0) {
+            return;
+        }
+
+        fd_set *fds = (fd_set *)CFDataGetMutableBytePtr(d);
+        fd_set invalidFds;
+        FD_ZERO(&invalidFds);
+        // Gather all invalid sockets into invalidFds set
+        for (u_int idx = 0; idx < fds->fd_count; idx++) {
+            SOCKET socket = fds->fd_array[idx];
+            if (! __CFNativeSocketIsValid(socket)) {
+                FD_SET(socket, &invalidFds);
+            }
+        }
+        // Remove invalid sockets from source set
+        for (u_int idx = 0; idx < invalidFds.fd_count; idx++) {
+            SOCKET socket = invalidFds.fd_array[idx];
+            FD_CLR(socket, fds);
+        }
+#else
         SInt32 count = __CFSocketFdGetSize(d);
         fd_set* s = (fd_set*) CFDataGetMutableBytePtr(d);
         for (SInt32 idx = 0;  idx < count;  idx++) {
@@ -1196,14 +1246,13 @@ clearInvalidFileDescriptors(CFMutableDataRef d)
                     FD_CLR(idx, s);
                 }
         }
+#endif
     }
 }
 
 static void
-manageSelectError()
+manageSelectError(SInt32 selectError)
 {
-    SInt32 selectError = __CFSocketLastError();
-
     __CFSOCKETLOG("socket manager received error %ld from select", (long)selectError);
 
     if (EBADF == selectError) {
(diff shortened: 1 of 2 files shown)


lxbndr:
@hjyamauchi this patch has minor differences from that version you checked. Just to be sure, could you please validate original issue with this implementation?

lxbndr:
Force-pushed NFC (removed unused variable `x` in the for-loop in the test code).

compnerd:
@swift-ci please test

hjyamauchi:
> @hjyamauchi this patch has minor differences from that version you checked. Just to be sure, could you please validate original issue with this implementation?

Validated. It worked fine.

hjyamauchi:
CC @tristanlabelle

CoreFoundation/RunLoop.subproj/CFSocket.c
```diff
@@ -213,13 +213,30 @@ CF_INLINE int __CFSocketLastError(void) {
 }
 
 CF_INLINE CFIndex __CFSocketFdGetSize(CFDataRef fdSet) {
+#if TARGET_OS_WIN32
+    if (CFDataGetLength(fdSet) == 0) {
+        return 0;
+    }
+    return FD_SETSIZE;
```
tristanlabelle, line 220: My understanding of the structure on Windows is that it has a default size but it's legit to allocate a bigger buffer and treat the array as bigger than `FD_SETSIZE`. I think a better implementation would calculate the number of entries from `CFDataGetLength` after removing the size of the count field. This would also avoid the implicit dependency where this code is assuming that the data length is at least that of `fd_set`.

CoreFoundation/RunLoop.subproj/CFSocket.c
```diff
@@ -213,13 +213,30 @@ CF_INLINE int __CFSocketLastError(void) {
...
+#endif
 }
 
 CF_INLINE Boolean __CFSocketFdSet(CFSocketNativeHandle sock, CFMutableDataRef fdSet) {
     /* returns true if a change occurred, false otherwise */
     Boolean retval = false;
     if (INVALID_SOCKET != sock && 0 <= sock) {
+        fd_set *fds;
+#if TARGET_OS_WIN32
+        if (CFDataGetLength(fdSet) == 0) {
```
tristanlabelle, line 232: I think this logic should support resizing the buffer past `sizeof(fd_set)` since `fd_set`'s trailing array appears meant to be used as elastic-sized, and we have nothing to detect the error condition if we use a fixed size and can't add a new socket.

CoreFoundation/RunLoop.subproj/CFSocket.c
```diff
@@ -230,9 +247,11 @@ CF_INLINE Boolean __CFSocketFdSet(CFSocketNativeHandle sock, CFMutableDataRef fd
         } else {
             fds_bits = (fd_mask *)CFDataGetMutableBytePtr(fdSet);
         }
-        if (!FD_ISSET(sock, (fd_set *)fds_bits)) {
+        fds = (fd_set *)fds_bits;
+#endif
+        if (!FD_ISSET(sock, fds)) {
             retval = true;
-            FD_SET(sock, (fd_set *)fds_bits);
+            FD_SET(sock, fds);
```
tristanlabelle, line 254: The macOS code made sure to resize so knows that this will succeed. However the Windows code is capping at 64 sockets and does not handle failure.

I think we should avoid using `FD_SET` on Windows because it has to assume an array size of 64, but `fd_set` is designed such that the allocation can make the trailing array bigger than that if it needs to.

CoreFoundation/RunLoop.subproj/CFSocket.c
```diff
@@ -1188,6 +1217,27 @@ static void
 clearInvalidFileDescriptors(CFMutableDataRef d)
 {
     if (d) {
+#if TARGET_OS_WIN32
+        if (CFDataGetLength(d) == 0) {
+            return;
+        }
+
+        fd_set *fds = (fd_set *)CFDataGetMutableBytePtr(d);
+        fd_set invalidFds;
```
tristanlabelle, line 1226: There could be more than 64 invalid sockets if fds->fd_count is bigger than 64.

tristanlabelle (changes requested):
This change is very necessary, thanks! I think there is still a problem with the Windows implementation. The code is treating `fd_set` as fixed-size, which will cause silent failures if we exceed the 64 array size in the structure (not a problem on macOS with bitset resizing). Reading [the docs](https://learn.microsoft.com/en-us/windows/win32/api/winsock/ns-winsock-fd_set), the `fd_set` structure is designed to have a trailing array that can be bigger than the default 64. I think the logic should support resizing the array past 64 as needed, so that we don't have this additional silent failure condition on Windows.

CoreFoundation/RunLoop.subproj/CFSocket.c
lxbndr: Ahh, I think I will update the patch and exclude this function from Win build at all. It is not used in any of Windows code paths except one place where its result is ignored anyway.

lxbndr:
@tristanlabelle so the major concern is the static non-growing structure, right? I agree that it is not comfortable to live with this restriction, and my initial intention was to make it dynamic.

Not sure if it is really ok to manipulate inner structure array, but I think it worth to try. I am suspicious about changing it on the fly, because `FD_SETSIZE` is a define, and it _must_ be defined before including winsock2 header. So it could be baked into some macros like `FD_SET` or `FD_ZERO`.

I initially planned not to grow the array, but create additional `fd_set`s as needed, so we would have growing array of fd_sets instead.

Also, some good news. The default cap is 64, which is pretty low. But look what we got [here](https://github.com/apple/swift-corelibs-foundation/blob/06072877b110c4ba5501ddc012830c1072f7198c/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h#L133C1-L133C24). We already have FD_SETSIZE defined as 1024. That's a plenty of space.

I could try dynamic sizing, but probably not early than this weekend 😔 If that not blocking anything critical for you, we could keep this for now. Or we could land it and I'll do a follow up.

And... Do you think it makes sense to fight 1024 sockets limit, while corelibs are gradually moving away from CoreFoundation as backing implementation? WDYT?

tristanlabelle:
@lxbndr Ah given the 1024 socket override I don't have much of a concern. I'll resolve my comments and file a follow-up issue.

You're right that we wouldn't be able to use `FD_SET` or `FD_ZERO` on Windows with dynamic sizing, they would need their own implementation.

tristanlabelle:
Actually `FD_SET` is the only problematic one for dynamic sizing:

```
#define FD_CLR(fd, set) do { \
    u_int __i; \
    for (__i = 0; __i < ((fd_set FAR *)(set))->fd_count ; __i++) { \
        if (((fd_set FAR *)(set))->fd_array[__i] == fd) { \
            while (__i < ((fd_set FAR *)(set))->fd_count-1) { \
                ((fd_set FAR *)(set))->fd_array[__i] = \
                    ((fd_set FAR *)(set))->fd_array[__i+1]; \
                __i++; \
            } \
            ((fd_set FAR *)(set))->fd_count--; \
            break; \
        } \
    } \
} while(0)

#define FD_SET(fd, set) do { \
    if (((fd_set FAR *)(set))->fd_count < FD_SETSIZE) \
        ((fd_set FAR *)(set))->fd_array[((fd_set FAR *)(set))->fd_count++]=(fd);\
} while(0)

#define FD_ZERO(set) (((fd_set FAR *)(set))->fd_count=0)

#define FD_ISSET(fd, set) __WSAFDIsSet((SOCKET)(fd), (fd_set FAR *)(set))

File: CoreFoundation/RunLoop.subproj/CFSocket.c (hunk 4 above)
Comment (tristanlabelle) [reply CoreFoundation/RunLoop.subproj/CFSocket.c:1226]: Resolving per convo about large FD_SETSIZE override.

File: CoreFoundation/RunLoop.subproj/CFSocket.c (hunk 3 above)
Comment (tristanlabelle) [reply CoreFoundation/RunLoop.subproj/CFSocket.c:254]: Resolved per convo about large FD_SETSIZE override, although we could add an assert.

File: CoreFoundation/RunLoop.subproj/CFSocket.c (hunk 2 above)
Comment (tristanlabelle) [reply CoreFoundation/RunLoop.subproj/CFSocket.c:232]: Resolved per convo about large FD_SETSIZE override.

File: CoreFoundation/RunLoop.subproj/CFSocket.c (hunk 1 above)
Comment (tristanlabelle) [reply CoreFoundation/RunLoop.subproj/CFSocket.c:220]: Resolved per convo about large FD_SETSIZE override

**Comment by compnerd** [comment]

**Comment by compnerd** [comment]

**Comment by hjyamauchi** [comment]

**Comment by hjyamauchi** [comment]
Does it looks like the macos CI was broken on May 11 before this PR was created? The latest run 466 looks broken in the same way as the May 11 run 461.

May 20 (the latest):
https://ci.swift.org/job/swift-corelibs-foundation-PR-macOS/466/

May 11:
https://ci.swift.org/job/swift-corelibs-foundation-PR-macOS/461/

```
Failed Tests (24):
  SwiftXCTestFunctionalTests :: ArgumentParser/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Expectations/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Handler/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Misuse/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Notifications/Expectations/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Notifications/Handler/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Predicates/Expectations/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Predicates/Handler/main.swift
  SwiftXCTestFunctionalTests :: Asynchronous/Use/main.swift
  SwiftXCTestFunctionalTests :: EqualityWithAccuracy/main.swift
  SwiftXCTestFunctionalTests :: ErrorHandling/main.swift
  SwiftXCTestFunctionalTests :: FailingTestSuite/main.swift
  SwiftXCTestFunctionalTests :: FailureMessagesTestCase/main.swift
  SwiftXCTestFunctionalTests :: ListTests/main.swift
  SwiftXCTestFunctionalTests :: Observation/All/main.swift
  SwiftXCTestFunctionalTests :: Observation/MultipleObservers/main.swift
  SwiftXCTestFunctionalTests :: Observation/Selected/main.swift
  SwiftXCTestFunctionalTests :: Performance/Misuse/main.swift
  SwiftXCTestFunctionalTests :: Performance/main.swift
  SwiftXCTestFunctionalTests :: SelectedTest/main.swift
  SwiftXCTestFunctionalTests :: SingleFailingTestCase/main.swift
  SwiftXCTestFunctionalTests :: SkippingTestCase/main.swift
  SwiftXCTestFunctionalTests :: TestCaseLifecycle/Misuse/main.swift
  SwiftXCTestFunctionalTests :: TestCaseLifecycle/main.swift
```

compnerd:
This is really windows specific, the tests haven't regressed, @parkera okay to merge ignoring the macOS issue?

PR #4955, opened 2024-05-13
Remove URL in favor of swift-foundation's URL

jmschonfeld:
This PR removes the `URL` type in favor of the `URL` type provided by swift-foundation. It contains a few minor test adjustments to account for the new URL parser (which can behave slightly differently).

This disables one test which we'll come back to soon to update for the new parsing behavior (cc @jrflat), otherwise all but 2 tests pass and those two failures will be fixed by https://github.com/apple/swift-foundation/pull/602 which contains a bug-fix for `URL`s created from relative paths. I'll update the dependency's commit hash once that PR is merged

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -78,7 +78,7 @@ let package = Package(
             from: "0.0.5"),
         .package(
            url: "https://github.com/apple/swift-foundation",
-           revision: "7eb8b598ad8f77ac743b2decc97d56bf350aedee"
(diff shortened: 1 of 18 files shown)


jmschonfeld:
The latest commit updates the swift-foundation dependency to the commit that resolves the two test failures. Now all tests pass on linux with this change (aside from the one test that is disabled)

parkera (approved):

PR #4957, opened 2024-05-15
Recore FileManager on top of FoundationEssentials.FileManager

jmschonfeld:
This PR removes all of the functionality from `FileManager` that is already surfaced via `FoundationEssentials`. There are still some pieces (namely those that require `URLResourceValues`) that are not available in `FoundationEssentials` yet, and those are left here in an `extension` on `FileManager`

This change depends on https://github.com/apple/swift-foundation/pull/609, I'll update the dependency's commit hash once that PR has been approved and is merged


Tests/Foundation/TestFileManager.swift
```diff
@@ -1385,15 +1393,11 @@ class TestFileManager : XCTestCase {
 
         XCTAssertThrowsError(try fm.attributesOfItem(atPath: "")) {
             let code = ($0 as? CocoaError)?.code
-            XCTAssertEqual(code, .fileReadInvalidFileName)
+            XCTAssertEqual(code, .fileReadNoSuchFile)
```
parkera, line 1396: Do we simply throw a different error now, or is the former API somehow missing?

Tests/Foundation/TestFileManager.swift
```diff
@@ -949,8 +954,8 @@ class TestFileManager : XCTestCase {
 
     func test_homedirectoryForUser() {
         let filemanger = FileManager.default
-        XCTAssertNil(filemanger.homeDirectory(forUser: "someuser"))
-        XCTAssertNil(filemanger.homeDirectory(forUser: ""))
+        XCTAssertNotNil(filemanger.homeDirectory(forUser: "someuser"))
+        XCTAssertNotNil(filemanger.homeDirectory(forUser: ""))
```
parkera, line 958: This is another change in behavior, correct?

Tests/Foundation/TestFileManager.swift
jmschonfeld: We throw an error now. Previously the behavior for these empty paths differed between Darwin and non-Darwin. The swift-foundation implementation takes the behavior from Darwin and so this updates the test expectations accordingly. Previously, getting the file system representation of an empty string unconditionally threw `.fileReadInvalidFileName`, but now we just create en empty cstring and provide it to the underlying system API and let it determine how to handle an empty path (which in most cases results in specifying that the file doesn't exist, unless you're working with a directory in which I believe some cases it can be considered the current directory)

parkera (approved):

PR #4959, opened 2024-05-18
Introduce Cmake support for SwiftCorelibsFoundation

iCharlesHu:
This patch introduces Cmake support for SwiftCorelibFoundation


Sources/CoreFoundation/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+
+target_include_directories(CoreFoundation
+    PUBLIC
+        include
+    PRIVATE
+        internalInclude)
+
+set(_CoreFoundation_build_flags ${_Foundation_common_build_flags})
+list(APPEND _CoreFoundation_build_flags
+    "-DCF_CHARACTERSET_UNICODE_DATA_L=\"${CMAKE_SOURCE_DIR}/Sources/CoreFoundation/CFUnicodeData-L.mapping\""
```
parkera, line 26: Do we need this anymore? I think I removed the need for it with a9ffa779b1?

iCharlesHu: Ahh I must have missed that. I'll remove it.

CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+include(FetchContent)
+if (_SwiftFoundationICU_SourceDIR)
+    FetchContent_Declare(SwiftFoundationICU
+        SOURCE_DIR ${_SwiftFoundationICU_SourceDIR})
+else()
+    FetchContent_Declare(SwiftFoundationICU
+        GIT_REPOSITORY https://github.com/iCharlesHu/swift-foundation-icu.git
+        GIT_TAG charles/cmake-support)
+endif()
+FetchContent_MakeAvailable(SwiftFoundationICU)
```
iCharlesHu, line 59: Merge this one with the one below.

CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+endif()
+FetchContent_MakeAvailable(SwiftFoundation)
+
+# System dependencies (fail fast if dependencies are missing)
+find_package(LibXml2 REQUIRED)
+find_package(CURL REQUIRED)
+find_package(dispatch CONFIG REQUIRED)
+
+# Common build flags (_CFURLSessionInterface, _CFXMLInterface, CoreFoundation)
+set(_Foundation_common_build_flags)
```
etcwilde, line 77: Don't need to declare variables in CMake.

Sources/_CFXMLInterface/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+##===----------------------------------------------------------------------===##
+
+add_library(_CFXMLInterface STATIC CFXMLInterface.c)
+
+target_include_directories(_CFXMLInterface
+    PUBLIC
+        include
+    PRIVATE
+        ../CoreFoundation/internalInclude
+        ../CoreFoundation/include
```
etcwilde, line 22: CoreFoundation lists `CoreFoundation/include` as a public header. Linking against it will pull this in automatically.

Sources/_CFXMLInterface/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+##
+##===----------------------------------------------------------------------===##
+
+add_library(_CFXMLInterface STATIC CFXMLInterface.c)
+
+target_include_directories(_CFXMLInterface
+    PUBLIC
+        include
+    PRIVATE
+        ../CoreFoundation/internalInclude
```
etcwilde, line 21: This include path smells a bit funny though.

Sources/_CFXMLInterface/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+    "SHELL:$<$<COMPILE_LANGUAGE:C>:${_Foundation_common_build_flags}>")
+
+target_link_libraries(_CFXMLInterface PRIVATE
+    CoreFoundation
+    dispatch
+    LibXml2::LibXml2)
+
+if(NOT BUILD_SHARED_LIBS)
+    set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS _CFXMLInterface)
+endif()
```
etcwilde, line 35: missing trailing newline

Sources/_CFURLSessionInterface/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+##===----------------------------------------------------------------------===##
+
+add_library(_CFURLSessionInterface STATIC CFURLSessionInterface.c)
+
+target_include_directories(_CFURLSessionInterface
+    PUBLIC
+        include
+    PRIVATE
+        ../CoreFoundation/internalInclude
+        ../CoreFoundation/include)
```
etcwilde, line 22: Don't need this, linking CoreFoundation will automatically add it.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
...
+#   get_swift_host_os(result_var_name)
+#
+#
+# Sets ${result_var_name} with the converted OS name derived from
+# CMAKE_SYSTEM_NAME.
+function(get_swift_host_os result_var_name)
+  set(${result_var_name} ${SWIFT_SYSTEM_NAME} PARENT_SCOPE)
+endfunction()
+
+function(_install_target module)
```
etcwilde, line 57: lets namespace this `SwiftCorelibsFoundation` or something like that

Sources/_CFURLSessionInterface/CMakeLists.txt
iCharlesHu, line 22: Redundant do not need

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
...
+    install(FILES $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftmodule
+      DESTINATION lib/${swift}/${swift_os}/${module_name}.swiftmodule
+      RENAME ${swift_arch}.swiftmodule)
+  else()
+    install(FILES
+      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftdoc
+      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftmodule
+      DESTINATION lib/${swift}/${swift_os}/${swift_arch})
+  endif()
+endfunction()
```
etcwilde, line 94: missing eof end-of-line.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
+
+# Returns the current architecture name in a variable
+#
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# If the current architecture is supported by Swift, sets ${result_var_name}
+# with the sanitized host architecture name derived from CMAKE_SYSTEM_PROCESSOR.
+function(get_swift_host_arch result_var_name)
```
etcwilde, line 9: We can just ask the compiler for the module triple instead and remove this tower of architectures.

Sources/CoreFoundation/include/CoreFoundation.h
```diff
@@ -76,7 +76,7 @@
 
 #include "ForSwiftFoundationOnly.h"
 
-#if TARGET_OS_OSX || TARGET_OS_IPHONE || TARGET_OS_WIN32
+#if TARGET_OS_OSX || TARGET_OS_IPHONE || TARGET_OS_WIN32 || TARGET_OS_LINUX
```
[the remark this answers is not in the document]
iCharlesHu: Yes. These headers are def needed on Linux and SPM includes them already because they are in the include folder.

Sources/_CFURLSessionInterface/include/CFURLSessionInterface.h
```diff
@@ -30,6 +30,7 @@
 #include "CFTargetConditionals.h"
 #include "CFBase.h"
 #include <stdio.h>
+#include <stdbool.h>
```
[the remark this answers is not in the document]
iCharlesHu: Yes I hit a build failure otherwise saying `bool` is not defined.

Sources/_CFXMLInterface/include/CFXMLInterface.h
```diff
@@ -16,6 +16,7 @@
 
 #include "CFTargetConditionals.h"
 #include "CFBase.h"
+#include "CFError.h"
```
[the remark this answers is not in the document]
iCharlesHu: Yes CFError is also referenced below. I think I worked on SPM because it aggressively includes everything in the `include` folder? I'm not 100% sure.

jmschonfeld (approved):

cmake/modules/FoundationSwiftSupport.cmake
```diff
@@ -0,0 +1,54 @@
+# Returns the os name in a variable
+#
+# Usage:
+#   get_swift_host_os(result_var_name)
+#
+#
+# Sets ${result_var_name} with the converted OS name derived from
+# CMAKE_SYSTEM_NAME.
+function(get_swift_host_os result_var_name)
```
etcwilde, line 9: A function to set the value of a variable to the value of another variable seems a bit excessive.
It looks like this is only getting used below to set `swift_os`. I think you can safely replace the uses of `swift_os` in `_foundation_install_target` with `${SWIFT_SYSTEM_NAME}`.

Sources/CoreFoundation/include/module.modulemap
```diff
@@ -1,4 +1,4 @@
-framework module _CoreFoundation [extern_c] [system] {
+module CoreFoundation {
```
etcwilde, line 1: Can we see if we can just use this modulemap instead of the `module.map` above? If that works, we can delete the `module.map` and stop copying the headers into an intermediate build directory, which would be great. It helps with configure times and makes sure that we can continue working once clang drops support for `module.map` files.

Sources/Foundation/CMakeLists.txt
```diff
@@ -0,0 +1,168 @@
...
+    PUBLIC
+        CoreFoundation
+        FoundationEssentials
+        FoundationInternationalization)
+
+set_target_properties(Foundation PROPERTIES
+    INSTALL_RPATH "$ORIGIN"
+    BUILD_RPATH "$<TARGET_FILE_DIR:swiftDispatch>"
+    Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift
+    INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_BINARY_DIR}/swift)
```
etcwilde, line 162: We may need to massage this a bit. `${CMAKE_BINARY_DIR}` is an absolute path, and that will give written into the install scripts.

Worth reading https://cmake.org/cmake/help/latest/command/target_include_directories.html#creating-relocatable-packages

CMakeLists.txt
```diff
@@ -0,0 +1,116 @@
...
+    # Deduplicate linked libraries where appropriate
+    cmake_policy(SET CMP0156 NEW)
+endif()
+
+if(POLICY CMP0157)
+    # New Swift build model: improved incremental build performance and LSP support
+    cmake_policy(SET CMP0157 NEW)
+endif()
+
+if (NOT DEFINED CMAKE_C_COMPILER)
```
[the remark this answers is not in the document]
iCharlesHu: Yes. When we tested locally sometimes Cmake will pick up `cc` instead of `clang`.

Sources/Foundation/CMakeLists.txt
iCharlesHu: I ended up just removing these.

cmake/modules/FoundationSwiftSupport.cmake
```diff
@@ -0,0 +1,56 @@
...
+  if(type STREQUAL EXECUTABLE)
+    return()
+  endif()
+
+  get_target_property(module_name ${module} Swift_MODULE_NAME)
+  if(NOT module_name)
+    set(module_name ${module})
+  endif()
+
+  if(NOT SwiftFoundation_MODULE_TRIPLE)
```
etcwilde, line 38: We probably want to pull this out of the function and put it at the top-level. Due to scoping, I'm fairly sure that this will get re-computed every time this function gets called if it isn't specified a cmake cache file or directly on the cmake invocation, which is relatively expensive for something that isn't changing.

iCharlesHu:
Merging now to reduce the amount of merge conflicts. We will address any additional comments in followup pushes.

PR #4962, opened 2024-05-21
Update the way we check for libcurl feature availability

parkera:
Since we can't rely on `cmake` to inspect the version of curl when we build as a package, switch to using a different approach for verifying availability of features that rely on certain versions of libcurl.


parkera:
@al45tair this came up when I merged your latest changes from main into our package branch, so including you for review as well.

Sources/FoundationNetworking/URLSession/URLSessionConfiguration.swift
```diff
@@ -216,14 +216,9 @@ open class URLSessionConfiguration : NSObject, NSCopying {
      Note that these headers are added to the request only if not already present. */
     open var httpAdditionalHeaders: [AnyHashable : Any]? = nil
     
-    #if NS_CURL_MISSING_MAX_HOST_CONNECTIONS
     /* The maximum number of simultaneous persistent connections per host */
-    @available(*, deprecated, message: "This platform doles not support selecting the maximum number of simultaneous persistent connections per host. This property is ignored.")
```
parkera, line 221: We lose this part, but that's the consequence until swiftpm supports something like cmake does here.

Sources/_CFURLSessionInterface/include/CFURLSessionInterface.h
```diff
@@ -34,6 +34,27 @@
 #include <winsock2.h>
 #endif
 
+// 7.84.0 or later
+#if LIBCURL_VERSION_MAJOR > 7 || (LIBCURL_VERSION_MAJOR == 7 && LIBCURL_VERSION_MINOR > 84) || (LIBCURL_VERSION_MAJOR == 7 && LIBCURL_VERSION_MINOR == 84 && LIBCURL_VERSION_PATCH >= 0)
+#define NS_CURL_CURLINFO_CAINFO_SUPPORTED 1
```
parkera, line 39: I inverted the sense from "missing" to "supported" because it was a lot easier to understand.

jmschonfeld (approved):

PR #4963, opened 2024-05-23
Mark Bundle as Sendable

parkera:
CFBundle, which this is a wrapper for, is thread safe, and NSBundle is marked as Sendable on Darwin. Since swift package manager inserts references to Bundle when using resources, it is important to make sure this type doesn't produce any warnings in Swift 6 mode.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #4966, opened 2024-05-29
Add upcall for FoundationEssentials to initialize NSNumbers

jmschonfeld:
This allows the changes in https://github.com/apple/swift-foundation/pull/641 to initialize NSNumbers in FoundationEssentials

### DIFF
--- a/Sources/Foundation/NSNumber.swift
+++ b/Sources/Foundation/NSNumber.swift
@@ -9,6 +9,7 @@
 
 
 @_implementationOnly import _CoreFoundation
+@_spi(SwiftCorelibsFoundation) @_exported import FoundationEssentials
 
 internal let kCFNumberSInt8Type = CFNumberType.sInt8Type
 internal let kCFNumberSInt16Type = CFNumberType.sInt16Type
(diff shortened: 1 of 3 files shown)


Sources/Foundation/NSNumber.swift
```diff
@@ -9,6 +9,7 @@
 
 
 @_implementationOnly import _CoreFoundation
+@_spi(SwiftCorelibsFoundation) @_exported import FoundationEssentials
```
itingliu, line 12: Is it `@_exported` so that  `_typeByName("Foundation._FoundationNSNumberInitializer")` is visible to FoundationEssentials?

itingliu (approved):

Sources/Foundation/NSNumber.swift
jmschonfeld: The `@_exported` attribute is the attribute that re-exports all of the symbols from `FoundationEssentials`. All of our imports of `FoundationEssentials` and `FoundationInternationalization` are `@_exported` so that `Foundation.URL == FoundationEssentials.URL` because `Foundation` is re-exporting `FoundationEssentials` and everything within it, so it's not specific to this change it's just something we have to add to every import of those modules

iCharlesHu (approved):

PR #4967, opened 2024-05-29
Remove URLQueryItem from swift-corelibs-foundation in favor of FoundationEssentials.URLQueryItem

iCharlesHu:
`URLQueryItem` is already implemented in `FoundationEssentials`.

### DIFF
--- a/Sources/Foundation/URLQueryItem.swift
+++ b/Sources/Foundation/URLQueryItem.swift
@@ -7,63 +7,12 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 
-
-/// A single name-value pair, for use with `URLComponents`.
-public struct URLQueryItem: ReferenceConvertible, Hashable, Equatable {
-    public typealias ReferenceType = NSURLQueryItem
-
-    fileprivate var _queryItem: NSURLQueryItem
-
-    public init(name: String, value: String?) {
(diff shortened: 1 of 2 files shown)


iCharlesHu:
@swift-ci please test

Sources/Foundation/URLQueryItem.swift
```diff
@@ -7,96 +7,15 @@
...
-}
+// URLQueryItem is defined in FoundationEssentials
+@_exported import FoundationEssentials
 
 extension URLQueryItem: _NSBridgeable {
     typealias NSType = NSURLQueryItem
-    internal var _nsObject: NSType { return _queryItem }
-}
-
-extension URLQueryItem: _ObjectiveCBridgeable {
```
jmschonfeld, line 69: Looks like this change drops the `_ObjectiveCBridgeable` conformance here, was that intentional? I don't think we want to remove it, we'll likely just have to adjust the implementation to call the right initializers (we can use `_nsObject` and `_swiftObject` for convenience)

iCharlesHu: `_ObjectiveCBridgeable` is already implemented in `FoundationEssentials`

iCharlesHu:
Added `_ObjectiveCBridgeable` conformance back.

iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #4968, opened 2024-05-30
NSData options should typealias to Data options

jmschonfeld:
This removes the various options structs defined in an extension on `NSData` in favor of type aliases pointing to the "true" option structs provided by FoundationEssentials


itingliu (approved):
🙌

PR #4969, opened 2024-06-04
[Windows] Dynamic `fd_set` in `CFSocket`

### LINKED ISSUE #4958: CFSocket should use dynamically-sized `fd_set` on Windows
The `fd_set` structure on Windows has a trailing array of socket that defaults to size 64, which we override to 1024 https://github.com/apple/swift-corelibs-foundation/blob/06072877b110c4ba5501ddc012830c1072f7198c/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h#L133C1-L133C24 .

Instead of using a large fixed-sized buffer, we can treat the structure as dynamically-sized and reallocate it when it gets full. Winsock functions read the size of the array from the count field in order to support the array to exceed 64 as needed. The only caveat is that we can't use the `FD_SET` macro because it has to use the compile-time size.

lxbndr:
This makes `fd_set` on Windows growable by adding additional space to the storing buffer, which effectively extends `fd_set.fd_array` capacity.

With this change fixed `FD_SETSIZE 1024` becomes unnecessarily large (every such `fd_set` takes more than 8Kb of memory), so it seems reasonable to reduce it to lower value. I guess `128` will do (which is twice as big as default `64`).

Also, I removed some unused defines left from old bit-field implementation and tossed local vars in `__CFSocketManager` function to make them available according their usage per platform.

Follow-up for #4954. Resolves #4958.

### DIFF
--- a/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
+++ b/CoreFoundation/Base.subproj/CoreFoundation_Prefix.h
@@ -130,7 +130,7 @@ typedef char * Class;
 #endif
 
 // The order of these includes is important
-#define FD_SETSIZE 1024
+#define FD_SETSIZE 128
 #include <winsock2.h>
 #include <windows.h>
 
--- a/CoreFoundation/RunLoop.subproj/CFSocket.c
+++ b/CoreFoundation/RunLoop.subproj/CFSocket.c
@@ -51,10 +51,6 @@
 #include <process.h>
 #endif
 
-#ifndef NBBY
-#define NBBY 8
-#endif
-
 #if TARGET_OS_WIN32
 
 // redefine this to the winsock error in this file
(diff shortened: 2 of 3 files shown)


lxbndr:
cc @tristanlabelle

lxbndr:
Rebased on latest `main`

parkera:
@swift-ci test

lxbndr:
Not sure if I am reading CI output correctly, but
```
/Users/ec2-user/jenkins/workspace/swift-corelibs-foundation-PR-macOS/branch-main/swift-corelibs-xctest/Sources/XCTest/Public/XCTestMain.swift:21:23: error: no such module 'SwiftFoundation'
    @_exported import SwiftFoundation
```
seems to be unrelated to the changes.

parkera:
We need to disable the macOS build here, it's not actually supported in SCL-F any longer.

PR #4970, opened 2024-06-04
Add async URLSession methods

jrflat:
Thanks to @b1ackturtle for providing the foundation of this work in #4705! 

This PR implements the remaining async `URLSession` methods and the `URLSessionTask.delegate` property, allowing a delegate to be set in the async methods (or in a non-async context before resumption). `URLSessionTask.delegate` returns the explicitly set delegate if present, otherwise returns the session delegate `as? URLSessionTaskDelegate`. It's now used wherever we previously casted the session delegate to `URLSessionTaskDelegate`. To accompany this change, functions in `extension URLSessionTaskDelegate` (which are called if not implemented by the task's delegate) now check if the session's delegate implements it and call the session's delegate if that's the case.


jrflat:
@swift-ci please test

parkera (approved):

PR #4971, opened 2024-06-05
Remove additional unnecessary FileManager code

jmschonfeld:
This PR removes more FileManager code that is duplicated from swift-foundation. It removes a handful of functions that are no longer necessary, removes the private `._userImmutable` file attribute since it is the same as the API `.immutable` attribute, and updates the private file attribute fetching/setting functions to call down to the public API functions to handle all API attributes instead of duplicating code for the API attributes.

All tests pass locally on linux with this change

### DIFF
--- a/Tests/Foundation/TestFileManager.swift
+++ b/Tests/Foundation/TestFileManager.swift
@@ -1335,7 +1335,7 @@ class TestFileManager : XCTestCase {
         }
         XCTAssertThrowsError(try fm.copyItem(atPath: "/tmp/t", toPath: "")) {
             let code = ($0 as? CocoaError)?.code
-            XCTAssertEqual(code, .fileNoSuchFile)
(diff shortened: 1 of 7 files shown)


Tests/Foundation/TestFileManager.swift
```diff
@@ -1364,7 +1364,7 @@ class TestFileManager : XCTestCase {
         }
         XCTAssertThrowsError(try fm.linkItem(atPath: "/tmp/t", toPath: "")) {
             let code = ($0 as? CocoaError)?.code
-            XCTAssertEqual(code, .fileNoSuchFile)
+            XCTAssertEqual(code, .fileReadNoSuchFile)
```
jmschonfeld, line 1367: These weren't immediately caused by this change, I think this was an effect of an underlying swift-foundation change

parkera (approved):

PR #4973, opened 2024-06-06
Fix test_webSocket hang

jrflat:
Because the test server sends a close message shortly after sending a pong, the connection might close before our task handles the pong and calls its completion handler. In this case, the pong completion handler is never called, so `try await task.sendPing()` hangs indefinitely.

This PR cleans up `pongCompletionHandlers` on error/close and minimally fixes the test to allow the case where the server closes the connection before the pong is handled.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
+++ b/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
@@ -865,6 +865,12 @@ open class URLSessionWebSocketTask : URLSessionTask {
                     }
                 }
                 self.receiveCompletionHandlers.removeAll()
+                for handler in self.pongCompletionHandlers {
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

etcwilde (commented):
On initial inspection, this looks fine to me.
I've got a local 5.10 Fedora40 toolchain build going to verify and I'll be able to comment further once that finishes.

etcwilde (approved):
LGTM

Cherry-picking this back to 5.10 gets me past the hanging websocket test on Fedora40.

The macOS failure (`dyld[95291]: Symbol not found: _swift_getTypeByMangledNameInContext2`) is known. The OS on the test runners is too old for the latest standard library.

PR #4976, opened 2024-06-08
[wasm] Do not set permissions in `Data.write`

kkebo:
This fixes https://github.com/swiftwasm/swift/issues/5584.

This PR does not affect platforms other than WASI.

### DIFF
--- a/Sources/Foundation/NSData.swift
+++ b/Sources/Foundation/NSData.swift
@@ -457,7 +457,12 @@ open class NSData : NSObject, NSCopying, NSMutableCopying, NSSecureCoding {
         }
 
         let fm = FileManager.default
+#if os(WASI)
+        // WASI does not have permission concept
+        let permissions: Int? = nil
+#else
         let permissions = try? fm._permissionsOfItem(atPath: path)
+#endif
 
         if writeOptionsMask.contains(.atomic) {
             let (newFD, auxFilePath) = try _NSCreateTemporaryFile(path)


kateinoigakukun:
Good catch! Don't we need to fix `_setAttributes` too?

kkebo:
@kateinoigakukun 

> Don't we need to fix _setAttributes too?

I'm sorry, but I don't understand what you mean by "fix `_setAttributes`".

I changed `permissions` so that the condition of the `if` statement are always not satisfied instead of enclosing the following lines with `#if !os(WASI)`.

- https://github.com/apple/swift-corelibs-foundation/blob/d7c868a21d086711de2cca72e619d4e8ce6bed49/Sources/Foundation/NSData.swift#L471-L473
- https://github.com/apple/swift-corelibs-foundation/blob/d7c868a21d086711de2cca72e619d4e8ce6bed49/Sources/Foundation/NSData.swift#L503-L505

kateinoigakukun:
@kkebo Sorry, I meant it might be better to ignore the permission attribute within [`FileManager.setAttributes`]( https://github.com/apple/swift-corelibs-foundation/blob/d7c868a21d086711de2cca72e619d4e8ce6bed49/Sources/Foundation/FileManager.swift#L358C15-L358C28) because currently all users of the API also need to exclude `FileAttributeKey.posixPermissions` entry from the passing attributes manually as we do in this PR but it's just annoying for most of use cases...

kkebo:
@kateinoigakukun That makes sense and it can solve many problems with less code. On the other hand, since it is true that the `posixPermissions` key is not supported on WASI, I feel that we should throw an error and notify programmer, and also programmers should not try to set `posixPermissions` in the first place.

kateinoigakukun:
Okay, given that the API is not so widely used directly, asking users to change their code might be practically reasonable 👍

kateinoigakukun:
@swift-ci test

kateinoigakukun (approved):

parkera:
We're going to lose this diff once we delete this file in the package branch. Please open a corresponding patch for https://github.com/apple/swift-foundation.

parkera:
Actually, this one may persist because this change was only made to the class type `NSData` and not the struct - was that intentional?

kkebo:
@parkera 

> Actually, this one may persist because this change was only made to the class type NSData and not the struct - was that intentional?

At least in swift-corelibs-foundation, the `Data.write` method calls the `NSData.write` and does not have its own implementation.

parkera:
This will soon no longer be the case - see here: https://github.com/apple/swift-foundation/blob/main/Sources/FoundationEssentials/Data/Data%2BWriting.swift

kkebo:
Yes, I know swift-foundation's `Data.write`.

Besides, I have read [your comment](https://forums.swift.org/t/migration-plan-swift-corelibs-foundation-to-swift-foundation/70943/23) before, and I know we will eventually use the implementation of swift-foundation, but I don't know how much effort it takes to make swift-foundation compatible with Wasm.

PR #4977, opened 2024-06-10
Merge the main branch into package

parkera:
Incorporate latest updates from `main` into `package`.


Sources/FoundationNetworking/URLSession/URLSession.swift
```diff
@@ -666,6 +682,231 @@ internal extension URLSession {
     }
 }
 
+fileprivate struct Lock<State>: @unchecked Sendable {
```
jmschonfeld, line 685: Doesn't have to be a part of this PR, but maybe we should expose `FoundationEssentials.LockedState` as `@_spi` to avoid inventing yet another lock type

jmschonfeld (approved):
LGTM

Sources/FoundationNetworking/URLSession/URLSession.swift
parkera: I think we'll soon be able to use `Mutex` anyway.

PR #4981, opened 2024-06-16
Add a thread-safe implementation of Process.currentDirectoryURL

jakepetroules:
posix_spawn_file_actions_addchdir_np is the official way to set the working directory of a spawned process and is supported on both macOS and glibc (Linux, etc.). This makes Process.currentDirectoryURL thread-safe, as the current approach will result in the working directory being nondeterministically assigned when spawning processes across multiple threads, using different working directories.

### DIFF
--- a/Sources/CoreFoundation/CFPlatform.c
+++ b/Sources/CoreFoundation/CFPlatform.c
@@ -57,6 +57,10 @@
 #define kCFPlatformInterfaceStringEncoding	CFStringGetSystemEncoding()
 #endif
 
+#ifndef __GLIBC_PREREQ
+#define __GLIBC_PREREQ(maj, min) 0
+#endif
+
 extern void __CFGetUGIDs(uid_t *euid, gid_t *egid);
 
 #if TARGET_OS_MAC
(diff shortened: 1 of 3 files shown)


jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test

parkera:
@iCharlesHu can you review please?

Sources/CoreFoundation/CFPlatform.c
```diff
@@ -2274,6 +2274,23 @@ CF_EXPORT int _CFPosixSpawnFileActionsAddClose(_CFPosixSpawnFileActionsRef file_
...
+  //  - macOS 10.15 (October 2019)
+  //  - musl 1.1.24 (October 2019)
+  //  - FreeBSD 13.1 (May 2022)
+  //  - Android 14 (October 2023)
+  // standard variant available in:
+  //  - Solaris 11.4 (August 2018)
+  //  - NetBSD 10.0 (March 2024)
+  // missing in:
+  //  - OpenBSD
+  return posix_spawn_file_actions_addchdir_np((posix_spawn_file_actions_t *)file_actions, path);
```
iCharlesHu, line 2291: Do we need to provide support for both `_np` and `standard` variant here? Otherwise this code won't compile on Solaris, NetBSD, and OpenBSD right?

jakepetroules: Solaris has both variants, NetBSD has only the standard variant, OpenBSD/QNX have neither, and the remaining ones I listed have only the _np variant.

So the current code will compile fine on Solaris, but you're right that NetBSD will fail. I could change it to be more like:

```c
#if Solaris || Glibc || Apple || musl || FreeBSD || Android
posix_spawn_file_actions_addchdir_np
#else 
posix_spawn_file_actions_addchdir // NetBSD
#endif
```

...so that NetBSD works, and OpenBSD and QNX will "just work" once they update their POSIX conformance, though.

Fallback if neither variants are supported... let's leave that as an exercise for future contributors working to port Swift to those platforms? We don't have autotools-style configure checks so I'm not sure we can easily test for the presence of the function right now anyways, versus a bunch of hardcoded checks for specific platform macros.

And there isn't a good way to do a fallback while retaining thread safety by not mutating the calling process's working directory.

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci test windows

Sources/CoreFoundation/CFPlatform.c
iCharlesHu: Ahh sounds good. That's reasonable.

iCharlesHu (approved):

jakepetroules:
@swift-ci please test

PR #4996, opened 2024-06-28
[CMake] Include GNUInstallDirs

jmschonfeld:
The GNUInstallDirs module sets `CMAKE_INSTALL_BINDIR` which is used on Windows as the destination for installed `.dll` files. This addresses issues that I saw when building a Windows toolchain where the `.dll` files were not installed into the toolchain previously, but with this change are correctly installed into the bin directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,6 +134,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
         "-I/usr/lib/swift") # dispatch
 endif()
 
+include(GNUInstallDirs)
 include(FoundationSwiftSupport)
 
 add_subdirectory(Sources)


CMakeLists.txt
```diff
@@ -134,6 +134,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
         "-I/usr/lib/swift") # dispatch
 endif()
 
+include(GNUInstallDirs)
```
iCharlesHu, line 137: Just double check: it's okay to set it for all platforms (not just windows right?)

iCharlesHu (approved):

CMakeLists.txt
jmschonfeld: Yep! Linux tests + toolchain builds pass with this change too - I suspect this would have also led to an issue on Linux with `plutil` not being installed since it too would be installed into the runtime directory, so it's likely not only safe but necessary there too

PR #4998, opened 2024-06-28
Use SHGetKnownFolderPath to get path to localappdata for current user

### LINKED ISSUE #4997: UserDefaults doesn't work on Windows when USERNAME and UserDir don't match
Since UserDefaults [hand builds ](https://github.com/apple/swift-corelibs-foundation/blob/77af9c5a984f40e186c0ba127cff88f2a2588c15/CoreFoundation/Base.subproj/CFKnownLocations.c#L100) the path to the local app data location, this location is incorrect if the `USERNAME` doesn't match the `USERPFOILE` location directory. This means that all `UserDefaults` writes will fail.

stevenbrix:
On Windows, it's not guaranteed that `USERPROFILE` is equivalent to `%SYSTEMDRIVE%\Users\%USERNAME%`.  So it's possible that UserDefaults tries to write to a non-existent location, in which case it will never work. When this happens, all writes silently no-op.

This PR uses the Shell API `SHGetKnownFolderPath` to get the location of `%LOCALAPPDATA%` instead of manually building the path.

See [documentation](https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid) for this API:
image

Fixes #4997

### DIFF
--- a/CoreFoundation/Base.subproj/CFKnownLocations.c
+++ b/CoreFoundation/Base.subproj/CFKnownLocations.c
@@ -1,6 +1,6 @@
 /*	CFKnownLocations.c
 	Copyright (c) 1999-2017, Apple Inc. and the Swift project authors
- 
+
 	Portions Copyright (c) 2014-2017, Apple Inc. and the Swift project authors
 	Licensed under Apache License v2.0 with Runtime Library Exception
 	See http://swift.org/LICENSE.txt for license information
@@ -18,26 +18,27 @@
 
 #if TARGET_OS_WIN32
 #include <userenv.h>
+#include <shlobj_core.h>
 #endif
 
 CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUser user, CFStringRef _Nullable username) {
     CFURLRef location = NULL;
-    
+
 #if TARGET_OS_MAC
-    
+
 /*
  Building for a Darwin OS. (We use these paths on Swift builds as well, so that we can interoperate a little with Darwin's defaults(1) command and the other system facilities; but you want to use the system version of CF if possible on those platforms, which will talk to cfprefsd(8) and has stronger interprocess consistency guarantees.)
- 
+
  User:
  - Any: /Library/Preferences
  - Current: $HOME/Library/Preferences
  */
-    
+
     switch (user) {
         case _kCFKnownLocationUserAny:
             location = CFURLCreateWithFileSystemPath(kCFAllocatorSystemDefault, CFSTR("/Library/Preferences"), kCFURLPOSIXPathStyle, true);
             break;
-            
+
         case _kCFKnownLocationUserCurrent:
             username = NULL;
             // passthrough to:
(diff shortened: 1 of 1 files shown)


ShikiSuen:
@stevenbrix You might want to remove the format changes introduced on your computer.
This minifies the modifications that only necessary changes are kept.

P.S.: This PR is really important and I don't know why it gets abandoned here for months.

ShikiSuen:
@stevenbrix This PR is fallen behind the main branch too far.
The file to modified has been relocated to this folder:
`/Sources/CoreFoundation/CFKnownLocations.c`
And the part I guess you might want to modify is limited to this scope:
https://github.com/swiftlang/swift-corelibs-foundation/blob/0129358ceac4e587b6340a52977ea75f8b4601d5/Sources/CoreFoundation/CFKnownLocations.c#L101-L123

CoreFoundation/Base.subproj/CFKnownLocations.c
```diff
@@ -96,30 +97,12 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
...
-            CFRelease(usernameDir);
-            CFRelease(appdataDir);
-
-            CFRelease(profilesDir);
-            if (user == _kCFKnownLocationUserCurrent) {
-                CFRelease(username);
-            }
+            wchar_t* path = NULL;
+            SHGetKnownFolderPath(&FOLDERID_LocalAppData, 0, NULL, &path);
+            location = CFURLCreateWithFileSystemPath(kCFAllocatorSystemDefault, path, kCFURLWindowsPathStyle, true);
```
compnerd, line 104: This seems incorrect. The `path` parameter is a wide C String not a CFStrimg. This should do the creation of the CFString and pass that.

compnerd (changes requested):

ShikiSuen:
@compnerd I emailed stevenbrix on Feb 18 trying to bring his attention towards your previous response here, but he has no response.

I am less familiar with the moral culture context in FOSS projects like this one. Is it appropriate for me (or anyone else) to replicate this PR and patch this? (with credits of stevenbrix in the commit message whenever necessary.)

compnerd:
> @compnerd I emailed stevenbrix on Feb 18 trying to bring his attention towards your previous response here, but he has no response.
> 
> 
> 
> I am less familiar with the moral culture context in FOSS projects like this one. Is it appropriate for me (or anyone else) to replicate this PR and patch this? (with credits of stevenbrix in the commit message wherever necessary.)

Yes, maintaining the authorship/credit is key. It would be very much acceptable for someone to pick up the patch and help push it along if the original author is unable to do so.

ShikiSuen:
> Yes, maintaining the authorship/credit is key. It would be very much acceptable for someone to pick up the patch and help push it along if the original author is unable to do so.

Solid copy. Thanks for your response.

While I may see whether I can have some time to handle this issue within the next week, anyone can go ahead doing this without having to wait for me. This issue directly affects the usability of Swift (especially UserDefaults) on Windows.

stevenbrix:
Hi @ShikiSuen  sorry for not responding. I'm not sure these changes are needed anymore now that the code has moved. I'll need to do more investigation here.

ShikiSuen:
> Hi @ShikiSuen sorry for not responding. I'm not sure these changes are needed anymore now that the code has moved. I'll need to do more investigation here.

Thanks for your response. Good to see you get back online.
I'll wait for your investigation results.

parkera:
@swift-ci test

ShikiSuen:
@compnerd Sorry for bothering you.
This PR is on halt for 2 weeks since the previous comment left above, hence my question.
Could you please take a look at stevenbrix's most-recent commit to see whether there's still any issue?

Sources/CoreFoundation/CFKnownLocations.c
```diff
@@ -96,30 +97,16 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
         }
         case _kCFKnownLocationUserCurrent:
             username = CFGetUserName();
-            // fallthrough
+            break;
```
compnerd, line 100: This seems wrong. It is backwards no? `LocalAppData` would be for the current user, not a specified user. This seems like it would not return a location for the current user, and return the location for the current user when you specify a user name.

ShikiSuen: Quoting @stevenbrix for his attention.

ShikiSuen: @compnerd I wonder whether the entire case `case _kCFKnownLocationUserCurrent` can be written as the following instead:
```c
        case _kCFKnownLocationUserCurrent:
            username = CFGetUserName();
            if !(CFEqual(username, CFSTR(""))) {
                break;
            }
            // fallthrough.
```

ShikiSuen:
It takes too long time to see this PR gets done and I'll continue working on this:
https://github.com/swiftlang/swift-corelibs-foundation/pull/5201

compnerd:
Closing this in favour of #5201

ShikiSuen:
@stevenbrix #5201 is closed.

The **orthodox** solution lives in dotNET SDK 10. However, due to the license compatibility concerns, It'd be better to let Microsoft employees to make the PR against this issue.

At this point, I don't have any better idea than this last resort.

PR #5002, opened 2024-07-10
[Swift-6.0] Workaround broken glibc modulemap

etcwilde:
We have been running into modularization issues on newer versions of various Linux distros, resulting in the compiler saying that the Glibc module doesn't have a SOCK_STREAM or SOCK_DGRAM. From some poking around, the definition is now coming from the CoreFoundation module as far as Swift is concerned. This is ultimately because our modulemap for Glibc is bad, but also means that I can't bring up Swift 6 on all of the Linux distros that 5.10 has support for.

This workaround removes the explicit module name from `SOCK_STREAM` and renames it to `FOUNDATION_SOCK_STREAM` to avoid an ambiguous name, and completely removes `SOCK_DGRAM`.

Both SOCK_STREAM and SOCK_DGRAM are fileprivates, so changing them will have no visible external effect. It is true that if another header somewhere defines `SOCK_STREAM`, we may pick it up instead of the definition from the glibc module, but that will also cause some nasty surprises to anyone using that header in C, so it is unlikely.

Fixes: rdar://128079849

Cherry-pick of: https://github.com/apple/swift-corelibs-foundation/pull/5005

### DIFF
--- a/Sources/Foundation/Port.swift
+++ b/Sources/Foundation/Port.swift
@@ -90,18 +90,22 @@ open class SocketPort: Port {}
 
 #else
 
+#if canImport(Darwin)
+import Darwin
+fileprivate let FOUNDATION_SOCK_STREAM = SOCK_STREAM
+fileprivate let FOUNDATION_IPPROTO_TCP = IPPROTO_TCP
+#endif
+
 #if canImport(Glibc) && !os(Android) && !os(OpenBSD)
 import Glibc
-fileprivate let SOCK_STREAM = Int32(Glibc.SOCK_STREAM.rawValue)
-fileprivate let SOCK_DGRAM  = Int32(Glibc.SOCK_DGRAM.rawValue)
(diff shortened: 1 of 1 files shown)


etcwilde:
@swift-ci please test

etcwilde:
(Thanks Alexis for the idea to _just delete `Glibc`_ 😆)

etcwilde:
@swift-ci please test

etcwilde:
@swift-ci please test

etcwilde:
@swift-ci please test Windows

xymus (approved):

parkera:
I would rather leave main and 6.0 in sync with the workaround until that workaround is ready.

etcwilde:
Alright, cherry-picking to both places until we have an actual fix in place.

parkera:
@swift-ci test

parkera (approved):

PR #5007, opened 2024-07-13
Allocate buffer of correct size in `NSURL.fileSystemRepresentation` on Windows

lxbndr:
As the `String.withCString` function provides utf-8 encoded content, we should not use `String.count` for receiving buffer length. `NSURL.fileSystemRepresentation` is used in `FileHandle(forReadingFrom:)`, and it fails with file urls containing non-ascii (e.g. cyrillic) characters.


lxbndr:
cc @compnerd

jmschonfeld:

jmschonfeld (approved):
Makes sense to me!

PR #5010, opened 2024-07-17
[6.0] Android: add better nullability checks for nullability annotations added in NDK 26

finagolfin:
__Explanation:__ This is needed because [Bionic recently added a bunch of these annotations](https://android.googlesource.com/platform/bionic/+/25725717861813c18189c2699cf663bdb984d1b9%5E%21/#F0).

__Scope:__ Additional nullability checks and force unwraps only

__Issue:__ None

__Original PR:__ This is a cut-down #4850, which will have to be reworked for the `swift-foundation` merge.

__Risk:__ Low

__Testing:__ I made sure this pull doesn't break anything by testing it with the previous NDK 25c also. I used this patch with others to build the Swift toolchain for my Android CI, finagolfin/swift-android-sdk#122, and [the Termux app for Android](https://github.com/termux/termux-packages/commit/d10fd452662d3c39c0f19a901433bce8711b4688#diff-b8f3b540c354923485cb06eba2d66237536f89e88fd51e170495ddd8f5d32622R44), which now uses NDK 26b.

__Reviewer:__ @compnerd

### DIFF
--- a/Sources/Foundation/FileManager.swift
+++ b/Sources/Foundation/FileManager.swift
@@ -579,13 +579,13 @@ open class FileManager : NSObject {
 #elseif os(WASI)
         let type = FileAttributeType(statMode: mode_t(s.st_mode))
 #else
-        if let pwd = getpwuid(s.st_uid), pwd.pointee.pw_name != nil {
-            let name = String(cString: pwd.pointee.pw_name)
+        if let pwd = getpwuid(s.st_uid), let pwd_name = pwd.pointee.pw_name {
+            let name = String(cString: pwd_name)
             result[.ownerAccountName] = name
         }
 
-        if let grd = getgrgid(s.st_gid), grd.pointee.gr_name != nil {
-            let name = String(cString: grd.pointee.gr_name)
+        if let grd = getgrgid(s.st_gid), let grd_name = grd.pointee.gr_name {
+            let name = String(cString: grd_name)
             result[.groupOwnerAccountName] = name
         }
 
--- a/Sources/Foundation/FileHandle.swift
+++ b/Sources/Foundation/FileHandle.swift
@@ -321,10 +321,15 @@ open class FileHandle : NSObject {
             if options.contains(.alwaysMapped) {
                 // Filesizes are often 64bit even on 32bit systems
                 let mapSize = min(length, Int(clamping: statbuf.st_size))
+              #if canImport(Android)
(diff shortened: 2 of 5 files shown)


parkera:
We mentioned this in #4850 - but we will be merging that work here, so we'll only need one patch for this in the end.

Sources/Foundation/FileManager.swift
```diff
@@ -579,13 +579,13 @@ open class FileManager : NSObject {
 #elseif os(WASI)
         let type = FileAttributeType(statMode: mode_t(s.st_mode))
 #else
-        if let pwd = getpwuid(s.st_uid), pwd.pointee.pw_name != nil {
-            let name = String(cString: pwd.pointee.pw_name)
+        if let pwd = getpwuid(s.st_uid), let pwd_name = pwd.pointee.pw_name {
+            let name = String(cString: pwd_name)
```
compnerd, line 583:
```suggestion
        if let pwd = getpwuid(s.st_uid), let pw_name = pwd.pointee.pw_name {
            let name = String(cString: pw_name)
```

Sources/Foundation/FileManager.swift
```diff
@@ -579,13 +579,13 @@ open class FileManager : NSObject {
...
             result[.ownerAccountName] = name
         }
 
-        if let grd = getgrgid(s.st_gid), grd.pointee.gr_name != nil {
-            let name = String(cString: grd.pointee.gr_name)
+        if let grd = getgrgid(s.st_gid), let grd_name = grd.pointee.gr_name {
+            let name = String(cString: grd_name)
```
compnerd, line 588:
```suggestion
        if let grd = getgrgid(s.st_gid), let gr_name = grd.pointee.gr_name {
            let name = String(cString: gr_name)
```

compnerd:
@swift-ci please test

Sources/Foundation/FileManager.swift
finagolfin: Change is fine by me, but as mentioned in #4889, I will get this working with the `swift-foundation` re-core by next week and re-submit for trunk then.

parkera: https://github.com/apple/swift-corelibs-foundation/pull/5011 is the PR to track merging main to release/6.0.

finagolfin:
Some of these are no longer needed after the re-core and other changes, others have been submitted again in #5024 and my pull for that, hyp#1.

PR #5013, opened 2024-07-17
Do not depend on the HAVE_STRLCAT and HAVE_STRLCPY preprocessor definitions.

### LINKED ISSUE #5012: The swift-corelibs-foundation build is broken in glibc 2.38+ on the main branch.
swift-corelibs-foundation uses `strlcpy` and `strlcat` functions, which are originally from BSD and not available in glibc. However, glibc 2.38 added these functions as an extension, so swift-corelibs-foundation includes a CMake check for these functions and uses them from glibc if available; otherwise, it provides its own implementation like this:

```c
#if !TARGET_OS_MAC
#if !HAVE_STRLCPY
CF_INLINE size_t
strlcpy(char * dst, const char * src, size_t maxlen) {
    // omitting the implementation
}
#endif
```

See https://github.com/apple/swift-corelibs-foundation/commit/8d1a6794206637fee67f4a955853b0d1771746e2 and https://github.com/apple/swift-corelibs-foundation/commit/48076cb22c151b52b52e204328d097a2474f64b6 for more details.

However, when the new CMakeFiles were added in https://github.com/apple/swift-corelibs-foundation/commit/c772413c531526535b7bfa8dd446242d6c16f234, this check wasn't added to the new CMakeLists.txt, so the build fails when glibc 2.38+ is used.

For example, see my build log at https://download.copr.fedorainfracloud.org/results/leebc/swift-lang-main/fedora-40-x86_64/07741791-swift-lang/builder-live.log.gz. The build failed with:

<details open>
  <summary>Details</summary>

  ```
[472/573][ 82%][102.959s] /builddir/build/BUILD/swift-source/build/buildbot_linux/llvm-linux-x86_64/bin/clang  -I/builddir/build/BUILD/swift-source/swift-corelibs-foundation/Sources/CoreFoundation/include -I/builddir/build/BUILD/swift-source/swift-corelibs-foundation/Sources/CoreFoundation/internalInclude -I/builddir/build/BUILD/swift-source/swift-foundation-icu/icuSources/include -isystem /builddir/build/BUILD/swift-source/build/buildbot_linux/libdispatch-linux-x86_64 -isystem /builddir/build/BUILD/swift-source/swift-corelibs-libdispatch -isystem /builddir/build/BUILD/swift-source/swift-corelibs-libdispatch/src -isystem /builddir/build/BUILD/swift-source/build/buildbot_linux/libdispatch-linux-x86_64/src -isystem /builddir/build/BUILD/swift-source/swift-corelibs-libdispatch/src/BlocksRuntime -Wno-unknown-warning-option -Werror=unguarded-availability-new -fno-stack-protector -O3 -DNDEBUG -fPIC -DDEPLOYMENT_RUNTIME_SWIFT -DCF_BUILDING_CF -DDEPLOYMENT_ENABLE_LIBDISPATCH -DHAVE_STRUCT_TIMESPEC -DSWIFT_CORELIBS_FOUNDATION_HAS_THREADS -Wno-shorten-64-to-32 -Wno-deprecated-declarations -Wno-unreachable-code -Wno-conditional-uninitialized -Wno-unused-variable -Wno-unused-function -Wno-microsoft-enum-forward-reference -Wno-int-conversion -fblocks -fconstant-cfstrings -fdollars-in-identifiers -fno-common -fcf-runtime-abi=swift -fexceptions -D_GNU_SOURCE -I/usr/lib/swift -Winvalid-pch -fpch-instantiate-templates -Xclang -emit-pch -Xclang -include -Xclang /builddir/build/BUILD/swift-source/build/buildbot_linux/foundation-linux-x86_64/Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h -x c-header -MD -MT Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.pch -MF Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.pch.d -o Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.pch -c /builddir/build/BUILD/swift-source/build/buildbot_linux/foundation-linux-x86_64/Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.c
FAILED: Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.pch
/builddir/build/BUILD/swift-source/build/buildbot_linux/llvm-linux-x86_64/bin/clang  -I/builddir/build/BUILD/swift-source/swift-corelibs-foundation/Sources/CoreFoundation/include -I/builddir/build/BUILD/swift-source/swift-corelibs-foundation/Sources/CoreFoundation/internalInclude -I/builddir/build/BUILD/swift-source/swift-foundation-icu/icuSources/include -isystem /builddir/build/BUILD/swift-source/build/buildbot_linux/libdispatch-linux-x86_64 -isystem /builddir/build/BUILD/swift-source/swift-corelibs-libdispatch -isystem /builddir/build/BUILD/swift-source/swift-corelibs-libdispatch/src -isystem /builddir/build/BUILD/swift-source/build/buildbot_linux/libdispatch-linux-x86_64/src -isystem /builddir/build/BUILD/swift-source/swift-corelibs-libdispatch/src/BlocksRuntime -Wno-unknown-warning-option -Werror=unguarded-availability-new -fno-stack-protector -O3 -DNDEBUG -fPIC -DDEPLOYMENT_RUNTIME_SWIFT -DCF_BUILDING_CF -DDEPLOYMENT_ENABLE_LIBDISPATCH -DHAVE_STRUCT_TIMESPEC -DSWIFT_CORELIBS_FOUNDATION_HAS_THREADS -Wno-shorten-64-to-32 -Wno-deprecated-declarations -Wno-unreachable-code -Wno-conditional-uninitialized -Wno-unused-variable -Wno-unused-function -Wno-microsoft-enum-forward-reference -Wno-int-conversion -fblocks -fconstant-cfstrings -fdollars-in-identifiers -fno-common -fcf-runtime-abi=swift -fexceptions -D_GNU_SOURCE -I/usr/lib/swift -Winvalid-pch -fpch-instantiate-templates -Xclang -emit-pch -Xclang -include -Xclang /builddir/build/BUILD/swift-source/build/buildbot_linux/foundation-linux-x86_64/Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h -x c-header -MD -MT Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.pch -MF Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.pch.d -o Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.pch -c /builddir/build/BUILD/swift-source/build/buildbot_linux/foundation-linux-x86_64/Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h.c
In file included from <built-in>:1:
In file included from /builddir/build/BUILD/swift-source/build/buildbot_linux/foundation-linux-x86_64/Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h:4:
/builddir/build/BUILD/swift-source/swift-corelibs-foundation/Sources/CoreFoundation/internalInclude/CoreFoundation_Prefix.h:205:1: error: static declaration of 'strlcpy' follows non-static declaration
  205 | strlcpy(char * dst, const char * src, size_t maxlen) {
      | ^
/usr/include/string.h:506:15: note: previous declaration is here
  506 | extern size_t strlcpy (char *__restrict __dest,
      |               ^
In file included from <built-in>:1:
In file included from /builddir/build/BUILD/swift-source/build/buildbot_linux/foundation-linux-x86_64/Sources/CoreFoundation/CMakeFiles/CoreFoundation.dir/cmake_pch.h:4:
/builddir/build/BUILD/swift-source/swift-corelibs-foundation/Sources/CoreFoundation/internalInclude/CoreFoundation_Prefix.h:219:1: error: static declaration of 'strlcat' follows non-static declaration
  219 | strlcat(char * dst, const char * src, size_t maxlen) {
      | ^
/usr/include/string.h:512:15: note: previous declaration is here
  512 | extern size_t strlcat (char *__restrict __dest,
      |               ^
2 errors generated.
  ```
</details>

We should add the check for strlcpy and strlcat functions in the new CMakeLists.txt as well.

bc-lee:
While checking whether the system provides the `strlcat` and `strlcpy`
functions, using an autoconf-style check to determine their availability
does not seem to be a good idea. This is because the code is also
used with SwiftPM, which does not have a similar feature test.
Instead, we should directly check the availability of these functions
in the system.

Given that `strlcpy` and `strlcat` are not standard C functions,
we should check whether the system is macOS, Linux with Bionic or
Musl libc, or glibc version 2.38 or later. This patch adds the necessary
checks in the header file so that SwiftPM can use these checks as well.

Fixes #5012

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,17 +76,7 @@ else()
 endif()
 FetchContent_MakeAvailable(SwiftFoundationICU SwiftFoundation)
 
-include(CheckSymbolExists)
 include(CheckLinkerFlag)
-
-check_symbol_exists(strlcat "string.h" HAVE_STRLCAT)
-check_symbol_exists(strlcpy "string.h" HAVE_STRLCPY)
-if(HAVE_STRLCAT)
-    add_compile_definitions($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:HAVE_STRLCAT>)
-endif()
-if(HAVE_STRLCPY)
-    add_compile_definitions($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:HAVE_STRLCPY>)
-endif()
 
 check_linker_flag(C "LINKER:--build-id=sha1" LINKER_SUPPORTS_BUILD_ID)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,17 @@ else()
 endif()
 FetchContent_MakeAvailable(SwiftFoundationICU SwiftFoundation)
 
+include(CheckSymbolExists)
 include(CheckLinkerFlag)
+
+check_symbol_exists(strlcat "string.h" HAVE_STRLCAT)
+check_symbol_exists(strlcpy "string.h" HAVE_STRLCPY)
+if(HAVE_STRLCAT)
+    add_compile_definitions($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:HAVE_STRLCAT>)
+endif()
+if(HAVE_STRLCPY)
+    add_compile_definitions($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:HAVE_STRLCPY>)
+endif()
 
 check_linker_flag(C "LINKER:--build-id=sha1" LINKER_SUPPORTS_BUILD_ID)
 
--- a/Sources/CoreFoundation/internalInclude/CoreFoundation_Prefix.h
+++ b/Sources/CoreFoundation/internalInclude/CoreFoundation_Prefix.h
@@ -199,8 +199,14 @@ static dispatch_queue_t __ ## PREFIX ## Queue(void) {			\
     #endif
 #endif
 
-#if !TARGET_OS_MAC
-#if !HAVE_STRLCPY
+#if TARGET_OS_MAC ||                                                           \
+    (TARGET_OS_LINUX &&                                                        \
+     (defined(__GLIBC__) &&                                                    \
+          (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 38)) ||      \
+      !defined(__GLIBC__)))
+// strlcat and strlcpy are available in the system
+#else
+// Define local versions of strlcpy and strlcat for glibc < 2.38
 CF_INLINE size_t
 strlcpy(char * dst, const char * src, size_t maxlen) {
     const size_t srclen = strlen(src);
@@ -212,9 +218,7 @@ strlcpy(char * dst, const char * src, size_t maxlen) {
     }
     return srclen;
 }
-#endif
 
-#if !HAVE_STRLCAT
 CF_INLINE size_t
 strlcat(char * dst, const char * src, size_t maxlen) {
     const size_t srclen = strlen(src);
@@ -228,8 +232,8 @@ strlcat(char * dst, const char * src, size_t maxlen) {
     }
     return dstlen + srclen;
 }
-#endif
-#endif // !TARGET_OS_MAC
+#endif // TARGET_OS_MAC || (TARGET_OS_LINUX && defined(__GLIBC__) && (__GLIBC__
+       // > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 38)))
 
 #if TARGET_OS_WIN32
 // Compatibility with boolean.h


CMakeLists.txt
```diff
@@ -76,8 +76,18 @@ else()
 endif()
 FetchContent_MakeAvailable(SwiftFoundationICU SwiftFoundation)
 
+include(CheckSymbolExists)
 include(CheckLinkerFlag)
 
+check_symbol_exists(strlcat "string.h" HAVE_STRLCAT)
+check_symbol_exists(strlcpy "string.h" HAVE_STRLCPY)
```
jmschonfeld, line 83: Is there a way that we can replicate this behavior in the package manifest as well so that when building via SwiftPM we get the same behavior? Or a way that we can add this check into the code itself rather than on the CMake/SwiftPM side?

bc-lee: Generally, no, SwiftPM doesn't have a way to add such checks or generate "config.h" files.

However, in these cases (HAVE_STRLCAT and HAVE_STRLCPY in Swift), I think we can change swift-corelibs-foundation (and other projects) to use local versions of these functions only if certain conditions are not met. Below is an example of how we can do this:

```c
#if !TARGET_OS_MAC
#if defined(__linux__) && ((defined(__GLIBC__) && __GLIBC__ >= 2 && defined(__GLIBC_MINOR__) && __GLIBC_MINOR__ >= 38) || !defined(__GLIBC__))
// strlcat and strlcpy are available in the system
#else
// Define local versions of strlcpy and strlcat for glibc < 2.38
CF_INLINE size_t
strlcpy(char * dst, const char * src, size_t maxlen) {
  ...
}
#endif
#endif  // !TARGET_OS_MAC
```

This code assumes the following things:
- Swift is supported on macOS (and other Apple platforms), Linux, and Windows.
- Android's bionic supports strlcpy and strlcat. [Link](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/string.h;drc=c5d9795cf403febea9562c03ffdb397327189ab0;l=96)
- Other than Android, Linux uses glibc or musl. Musl supports strlcpy and strlcat. [Link](https://git.musl-libc.org/cgit/musl/tree/include/string.h?id=ab31e9d6a0fa7c5c408856c89df2dfb12c344039#n86)
- glibc 2.38 and later supports strlcpy and strlcat.
- Windows doesn't have strlcpy and strlcat.

This approach is somewhat error-prone and may not be the best way to handle this, but I think this is the only way to handle this in SwiftPM. Also, I don't think supported platforms change frequently, so this approach should be fine with appropriate compile flags. Do you think this approach is acceptable?

CMakeLists.txt
```diff
@@ -76,8 +76,18 @@ else()
...
+if(HAVE_STRLCAT)
+    add_compile_definitions($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:HAVE_STRLCAT>)
+endif()
+if(HAVE_STRLCPY)
+    add_compile_definitions($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:HAVE_STRLCPY>)
+endif()
```
compnerd, line 89: I think that we need to have an entry for Swift as well to pass this to the ClangImporter. Something like `$<$<COMPILE_LANGUAGE:Swift>:SHELL:-Xcc -DHAVE_STRLCPY>`

compnerd (commented):
This definitely seems like a good improvement once we add the Swift case as well.

CMakeLists.txt
jmschonfeld: I think this is an alright way to right this given that SwiftPM can't perform a check - assuming there's no other macro defined in the header that we can use to indicate the presence or lack of this function. @parkera had to do something similar for CURL: https://github.com/apple/swift-corelibs-foundation/blob/15ee5c32d2b808fd912c586c535feaf9af2eb2d5/Sources/_CFURLSessionInterface/include/CFURLSessionInterface.h#L39-L57

The only tweak I'd suggest - rather than doing this for `!TARGET_OS_MAC`, should we just only do this for `TARGET_OS_LINUX` (or whatever the right macro is for that)? In other words, do we need to define `strlcpy` for Windows/Android or does the source code in Foundation already handle not calling that function?

bc-lee: If you think it is a good idea to remove autoconf-style feature test preprocessor macros and check features in the code for compatibility with SwiftPM, I'll revisit this patch to do so.

jmschonfeld: 👍 yep since the alternative would break the SwiftPM build of this project, checking conditionally in the code is the way to go here to avoid breaking the build

CMakeLists.txt
bc-lee: All changes within CMakeLists.txt were reverted in the latest commit. Please check the header file for the changes.

parkera:
@swift-ci test

Sources/CoreFoundation/internalInclude/CoreFoundation_Prefix.h
```diff
@@ -199,8 +199,14 @@ static dispatch_queue_t __ ## PREFIX ## Queue(void) {			\
     #endif
 #endif
 
-#if !TARGET_OS_MAC
-#if !HAVE_STRLCPY
+#if TARGET_OS_MAC ||                                                           \
+    (TARGET_OS_LINUX &&                                                        \
```
parkera, line 203: Is this going to be a problem for platforms which are not Linux but do have `strlcat` (FreeBSD, e.g.)?

parkera: (maybe not depending on the definition of `__GLIBC__` but I'm unsure of where that's set)

bc-lee: That's the weakness of this approach, and that's also why every project has a list of supporting platforms. From what I know, Swift isn't officially supported on BSD variants (obviously, except for Darwin). So, I think it's not a problem. I don't even have a FreeBSD machine to test it. But if it is absolutely necessary, I may add another check for BSD variants based on https://sourceforge.net/p/predef/wiki/OperatingSystems/ and other sources, but honestly, I don't think it's necessary.

compnerd: FreeBSD does have a port as does Haiku, PS4, etc. I think that the target os macros do not work even on Linux. Does this cover uclibc and newlib? There is a plethora of libcs, and the fact that this "breaks" on SPM is fine - you can mirror this into SPM as:

```
cSettings: [
  .define("HAVE_STRLCAT", .when(platforms: [.linux]))),
  .define("HAVE_STRLCPY", .when(platforms: [.linux]))),
]
```

@jmschonfeld - I think that this is a sufficient means of mapping this in SPM, which will need to be changed if your environment is different as SPM does not have the necessary flexibility.

bc-lee: What if this (swift-corelibs-foundation) library is used as a transitive dependency in a Swift Package Manager project? Since SPM can't replace dependencies on the consumer side, it is extremely annoying to patch existing projects, as all dependent projects need to be patched as well. If we had a similar mechanism like `go mod replace` or npm's override mechanism, it would be much easier to handle this kind of situation, but that is a different topic.

compnerd: You would need to update the dependency. Fortunately, swift-corelibs-foundation cannot be used as a package dependency and is shipped as part of the toolchain.

bc-lee: It seems that the annoyance of replacing transitive dependencies is not a problem for the swift-corelibs-foundation repository. However, I'm still unsure about defining `HAVE_STRLCAT` and `HAVE_STRLCPY` in the SwiftPM manifest file. Even though it's the same OS (Linux), the actual availability of these functions can vary. I think it would be better to define these macros in the source code, even though it might be a bit messy.

I also wish Swift had clearer documentation on which platforms are supported and which are not, so we can avoid confusions with platforms like FreeBSD. I have made a post on the Swift forums to clarify this: https://forums.swift.org/t/clarifying-supported-platforms-for-swift/73309

aoberoi:
Hi there, I'm wondering if there are any updates regarding this change landing. I'm a little out of my depth wrt the discussion of multiple techniques of implementing the conditional checks required, but it sounds to me that there is some consensus on a preferred solution that would require a new capability of SwiftPM. If that is true, is there an issue or discussion of to spec out this new capability? If someone were able to volunteer some guidance, I may even be able to pitch or implement such a change.

I'm interested in using swift-corelibs-foundation as a package dependency instead of using the binaries within the toolchain so that I can more easily debug issues by stepping into the source. I believe this is aligned with the project and work group's goal of minimizing this compatibility layer over time, since it will help people debug issues, find inconsistencies, and contribute changes to swift-foundation.

jmschonfeld:
> Hi there, I'm wondering if there are any updates regarding this change landing. I'm a little out of my depth wrt the discussion of multiple techniques of implementing the conditional checks required, but it sounds to me that there is some consensus on a preferred solution that would require a new capability of SwiftPM. If that is true, is there an issue or discussion of to spec out this new capability? If someone were able to volunteer some guidance, I may even be able to pitch or implement such a change.
> 

This change was landed, in effect, in some other PRs I believe. The current state is:

- The CMake build sets `HAVE_STRLCAT`/`HAVE_STRLCPY` for platforms that define this symbol by performing the most accurate check (actually looking for the existence of the symbol). This means that the CMake build sets this correctly always
- In the source code, a few platforms have explicitly set these values unconditionally for their platform (like WASM IIRC) which will affect both CMake and SwiftPM builds and ensure this is always correct regardless of the build system for that platform
- SwiftPM unfortunately does not yet have the capability to check for the existence of the symbol, so the best it can do is make a best effort attempt (enabling/disabling on a per-platform basis), which is "good enough" for our uses of the SwiftPM build (at-desk building of this project on main supported platforms)

If you're hitting build failures related to this, I'd be interested to know your build setup (which platform/SDK, which build system - CMake or SwiftPM, etc.)

> I'm interested in using swift-corelibs-foundation as a package dependency instead of using the binaries within the toolchain so that I can more easily debug issues by stepping into the source. I believe this is aligned with the project and work group's goal of minimizing this compatibility layer over time, since it will help people debug issues, find inconsistencies, and contribute changes to swift-foundation.

A word of caution here: the SwiftPM build of this project is intended to be used for development of this project but not as a stable dependency for other projects. Other projects need to use the Foundation from the toolchain so that everybody "agrees" on which Foundation to use. If your project imports Foundation from a package dependency and uses another package dependency that also imports Foundation from the toolchain, you will end up with "two Foundations" that are not compatible with each other. You can definitely use the SwiftPM build as a dependency for small test projects to debug issues (I often use it as a dependency for a small CLI that I can use to test out reported problematic code) but using it as a dependency for a full project with a tree of dependencies (especially in any production use cases) is not supported and likely will not work due to the "two Foundations" problem. Hopefully that clears this up a bit just to make sure you don't accidentally end up in unsupported territory 🙂.

bc-lee:
I confirmed that this patch is no longer necessary, as the relevant PR has already landed in #5113. Closing this.

PR #5014, opened 2024-07-17
Add fcntl.h include, required for some linux platforms

parkera:
Amazon Linux, at least, and probably others, requires the `fcntl.h` include to use the various constants in `statx`.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #2897, opened 2020-10-10
suggestedFilename always returns a valid filename

Coeur:
The property `suggestedFilename` is documented with:
>This method always returns a valid filename.

The default initializer `init(url:mimeType:expectedContentLength:textEncodingName:)` had a correct implementation, but the required `init?(coder:)` was unsafe.

This is the same PR as #2643 created in February 2020, but for the **main** branch.

PR #2643, opened 2020-02-03
suggestedFilename always returns a valid filename

Coeur:
The property `suggestedFilename` is documented with:
>This method always returns a valid filename.

The default initializer `init(url:mimeType:expectedContentLength:textEncodingName:)` had a correct implementation, but the required `init?(coder:)` was unsafe.

PR #5019, opened 2024-07-20
suggestedFilename always returns a valid filename

Coeur:
The property `suggestedFilename` is documented with:
>This method always returns a valid filename.

The default initializer `init(url:mimeType:expectedContentLength:textEncodingName:)` had a correct implementation, but the required `init?(coder:)` was unsafe.

This is the same PR as #2643 created in February 2020, but for the **main** branch.
This is the same PR as #2897 created in October 2020, but it was closed for being unreviewed for too long.

### DIFF
--- a/Sources/FoundationNetworking/URLResponse.swift
+++ b/Sources/FoundationNetworking/URLResponse.swift
@@ -46,8 +46,11 @@ open class URLResponse : NSObject, NSSecureCoding, NSCopying, @unchecked Sendabl
         }
         
         // re-sanitizing with lastPathComponent because of supportsSecureCoding
-        let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename")?.lastPathComponent
-        self.suggestedFilename = encodedFilename?.isEmpty != false ? "Unknown" : encodedFilename
+        if let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename")?.lastPathComponent, !encodedFilename.isEmpty {
+            self.suggestedFilename = encodedFilename
+        } else {
+            self.suggestedFilename = "Unknown"
+        }
     }
     
     open func encode(with aCoder: NSCoder) {
--- a/Sources/FoundationNetworking/URLResponse.swift
+++ b/Sources/FoundationNetworking/URLResponse.swift
@@ -46,8 +46,11 @@ open class URLResponse : NSObject, NSSecureCoding, NSCopying, @unchecked Sendabl
             self.textEncodingName = encodedEncodingName as String
         }
         
-        if let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename") {
-            self.suggestedFilename = encodedFilename as String
+        if let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename")?.lastPathComponent,
+           !encodedFilename.isEmpty {
+            self.suggestedFilename = encodedFilename
+        } else {
+            self.suggestedFilename = "Unknown"
         }
     }
     
--- a/Sources/FoundationNetworking/URLResponse.swift
+++ b/Sources/FoundationNetworking/URLResponse.swift
@@ -35,7 +35,6 @@ open class URLResponse : NSObject, NSSecureCoding, NSCopying, @unchecked Sendabl
         guard let nsurl = aDecoder.decodeObject(of: NSURL.self, forKey: "NS.url") else { return nil }
         self.url = nsurl as URL
         
-        
         if let mimetype = aDecoder.decodeObject(of: NSString.self, forKey: "NS.mimeType") {
             self.mimeType = mimetype as String
         }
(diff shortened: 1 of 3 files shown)


parkera:
@swift-ci test

parkera:
@jrflat to review

Sources/FoundationNetworking/URLResponse.swift
```diff
@@ -46,8 +46,11 @@ open class URLResponse : NSObject, NSSecureCoding, NSCopying {
             self.textEncodingName = encodedEncodingName as String
         }
         
-        if let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename") {
-            self.suggestedFilename = encodedFilename as String
+        if let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename")?.lastPathComponent,
```
jrflat, line 49: Thanks for addressing this! I don't think we actually need to encode or decode the `suggestedFilename`, since given the `url` it can be calculated with
```
let c = url.lastPathComponent
self.suggestedFilename = c.isEmpty ? "Unknown" : c
```
such as in the `public init(url:mimeType:...)` below. (Darwin also considers `mimeType` and other logic described in `suggestedFilename`'s API comment below, but we could address that in another PR.)

Can we refactor `init?(coder:)` to only decode `url`, `mimeType`, `expectedContentLength`, and `textEncodingName`, then pass those in a single call to
```
public init(url: URL, mimeType: String?, expectedContentLength length: Int, textEncodingName name: String?)
```

Coeur: No we can't do that.
One that uses `public init?(url: URL, statusCode: Int, httpVersion: String?, headerFields: [String : String]?)` will do it with the goal to override the suggestedFilename with a value coming from headerFields. Encoding then Decoding the class must preserves that custom suggestedFilename.

jrflat: Yeah that makes sense we would want to preserve the custom filename from the header field. Though ideally, I think we'd want to prevent any state where `suggestedFilename` does not match the properties it was computed from (or otherwise propose API for setting it explicitly).

One solution could be to make `suggestedFilename` a computed var, which is not ABI breaking because the setter is fileprivate. For a `URLResponse`, this would return `url.lastPathComponent` or `"Unknown"` if empty. For `HTTPURLResponse`, this could check if the header fields contains the respective value, otherwise fallback to the super implementation. What are your thoughts on this?

Coeur: Your idea is good in the sense that it will save us some storage space (no more "NS.suggestedFilename").

And that makes me think:
**Do we want to maximise security or performance when encoding/decoding?**

A. If we want to maximise performance, then we shouldn't have to reprocess what has already been processed, and we should blindly recopy what is found is the archive. So I shouldn't add `.lastPathComponent` here.

B. If we want to maximise security, then I just figured out that not only `suggestedFilename` wasn't properly sanitised previously, but `_allHeaderFields` too: I can modify an archive to lowercase all the headers, then decode the archive and "NS.allHeaderFields" will be used as-is, with lowercase headers.

In a sense, with the sensibility that constitutes an `URLResponse`, I wouldn't want to have surprising results at decode time. So my plans:
1. yes, let's stop encoding "NS.suggestedFilename"
2. when decoding an URLResponse, we validate/sanitise the suggestedFilename
3. when decoding an HTTPURLResponse, we validate/sanitise the header fields

Coeur: Actually, we don't have the choice:
>/// Conforming to the `NSSecureCoding` protocol indicates that an object handles
/// encoding and decoding instances of itself in a manner that is robust against
/// object substitution attacks.

So we by conformance, we must sanitise `suggestedFilename` and `_allHeaderFields` in `init?(coder aDecoder: NSCoder)`

jrflat: I agree with your analysis, sounds like a good plan. I would prefer security regardless since maliciously substituting a `suggestedFilename` could lead to selectively overwriting a file on a user's device, which is concerning. Also, I think any performance (speed) drawbacks would likely be negligible, and we get the benefit of using less memory during coding.

One thing I'd note is that we still need to fallback to deriving it from the URL if the `HTTPURLResponse` doesn't contain the respective header. I still think computed vars could be a good pattern for this, but I'm not stuck on that idea if you'd prefer to sanitize and store in the coder init.

Computed vars could negligibly boost performance if the user doesn't need `suggestedFilename`, but computing it should be very quick anyway. We could cache the first computed var call using some private `_suggestedFilename` for quick access after, but I'm guessing the performance gain would be so small anyway, I wouldn't worry about it. Just food for thought :)

Coeur: >Initializer cannot both delegate ('self.init') and chain to a superclass initializer ('super.init')

So our idea isn't trivially doable. I'll need to extract some logic in their own functions.

Coeur: I did the best I could in e424b2ca61d7f28cb5ac3c3e37cb2328f73cd1e6

Sources/FoundationNetworking/URLResponse.swift
```diff
@@ -46,9 +45,9 @@ open class URLResponse : NSObject, NSSecureCoding, NSCopying, @unchecked Sendabl
             self.textEncodingName = encodedEncodingName as String
         }
         
-        if let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename") {
-            self.suggestedFilename = encodedFilename as String
-        }
+        // re-sanitizing with lastPathComponent because of supportsSecureCoding
+        let encodedFilename = aDecoder.decodeObject(of: NSString.self, forKey: "NS.suggestedFilename")?.lastPathComponent
+        self.suggestedFilename = encodedFilename?.isEmpty != false ? "Unknown" : encodedFilename
```
jrflat, line 50: I think the previous logic you had might be a little clearer here since it might not be obvious to a reader that `nil != false`:
```
if let encodedFilename = aDecoder.decodeObject(...)?.lastPathComponent, !encodedFilename.isEmpty { ... }
```

jrflat (approved):
Thanks for your work on this! I left one small nit but otherwise looks good!

Sources/FoundationNetworking/URLResponse.swift
Coeur: b24b0f235fc69751859bf1af6d73ff44ee443a6a

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

PR #5021, opened 2024-07-20
Foundation: install the import libraries for public dependencies

compnerd:
When building against Foundation in a multi-stage build, we would fail to build due to the import libraries being missing. These are required on platforms such as Windows and AIX (and theoretically on macOS if we generated TBDs).

### DIFF
--- a/Sources/Foundation/CMakeLists.txt
+++ b/Sources/Foundation/CMakeLists.txt
@@ -178,3 +178,7 @@ endif()
 
 set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS Foundation)
 _foundation_install_target(Foundation)
+install(FILES
+    $<TARGET_LINKER_FILE:FoundationEssentials>
+    $<TARGET_LINKER_FILE:FoundationInternationalization>
+  DESTINATION lib)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

compnerd:

Sources/Foundation/CMakeLists.txt
```diff
@@ -178,3 +178,7 @@ endif()
 
 set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS Foundation)
 _foundation_install_target(Foundation)
+install(FILES
```
jmschonfeld, line 181: Should we move this to the swift-corelibs-foundation repo where these modules are defined instead? It seems a bit odd to be installing pieces of swift-foundation from another repository. Also, is this safe to apply to all platforms or should this be conditionalized to Windows/other applicable platforms?

compnerd: The generator expression should evaluate to the `.lib` on Windows, `.tbd` on macOS (if present), and empty string on Linux. Installing it in swift-foundation makes sense, but we will need to copy it here as well I think.

jmschonfeld: Yeah it seems like this repo might need to do this for the swift-corelibs-foundation libraries like `Foundation`/`FoundationXML`/`FoundationNetworking` but it'd be great if we could move the treatment of the swift-foundation libraries to that repo. Perhaps we could just add an extra `install` call to the `_foundation_install_target` function so we can make sure this is appropriately handled for all of our targets?

compnerd:
I think that this is not needed currently.

PR #5022, opened 2024-07-20
Add default implementations for three default protocol conformances in the URLSessionDelegate family

jakepetroules:
Implementing these callbacks without calling the completion handler causes hangs in cases where these methods are called. Add reasonable default behaviors for all of them, to prevent this.

This issue has been in place for at least 8 years for one of these callbacks.


jakepetroules:
@swift-ci please test

parkera:
@jrflat to review (and also to verify behavior compatibility with Darwin)

parkera:
I think at the very least we need test cases to cover the desired new behavior.

Sources/FoundationNetworking/URLSession/URLSessionDelegate.swift
```diff
@@ -244,15 +246,19 @@ public protocol URLSessionDataDelegate : URLSessionTaskDelegate {
 
 extension URLSessionDataDelegate {
     
-    public func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse, completionHandler: @escaping (URLSession.ResponseDisposition) -> Void) { }
+    public func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse, completionHandler: @escaping (URLSession.ResponseDisposition) -> Void) {
+        completionHandler(.allow)
```
jrflat, line 250: There was a recent change that added the API for `URLSessionTask.delegate`. A delegate set on task has priority over the delegate set on the session, but if the task-specific delegate doesn't implement the function, then we see if the delegate set on the session does. We can do this here and in the other function below, using the pattern

```
if self === task.delegate, let sessionDelegate = session.delegate as? URLSessionDataDelegate, self !== sessionDelegate {
    sessionDelegate.urlSession(session, dataTask: dataTask, didReceive: response, completionHandler: completionHandler)
} else {
    completionHandler(.allow)
}
```

jrflat (commented):
Thanks for fixing this! These default behaviors all LGTM and match the Darwin behavior. I just left one comment about needing to call both the task and session delegate.

parkera (changes requested):
Marking as request changes per @jrflat feedback.

jakepetroules:
Thanks for the review @jrflat! Does this look right, now?

jakepetroules:
@swift-ci please test

jrflat (approved):
Looks good, thanks!

jrflat:

PR #5025, opened 2024-07-24
Allow custom dispatch include path in SwiftPM package build

jmschonfeld:
This adds a `DISPATCH_INCLUDE_PATH` environment variable that allows specifying the location at which the package build can find dispatch headers. On linux/android this path still defaults to `/usr/lib/swift`. This will allow us to build the tests of this package via SwiftPM in the toolchain build where we build against a just-built toolchain (and `/usr/lib/swift` doesn't exist).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,6 +3,19 @@
 
 import PackageDescription
 
+var dispatchIncludeFlags: CSetting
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

PR #5027, opened 2024-07-25
Re-enable tests requiring isoLatin1/macOSRoman

jmschonfeld:
Now that swift-foundation has support for `isoLatin1` and `macOSRoman`, we can enable this piece of the test that relied on those encodings.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #5030, opened 2024-07-26
Make completion-based send/receive functions public in `URLSessionWebSocketTask`

lxbndr:
Completion-based [send](https://developer.apple.com/documentation/foundation/urlsessionwebsockettask/3281790-send) and [receive](https://developer.apple.com/documentation/foundation/urlsessionwebsockettask/3281789-receive) are public in Apple Foundation.

Perhaps, availability should be added to the containing type, but I added it here for consistency with async counterparts.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
+++ b/Sources/FoundationNetworking/URLSession/URLSessionTask.swift
@@ -830,7 +830,8 @@ open class URLSessionWebSocketTask : URLSessionTask, @unchecked Sendable  {
         }
     }
     
-    private func send(_ message: Message, completionHandler: @Sendable @escaping (Error?) -> Void) {
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    public func send(_ message: Message, completionHandler: @Sendable @escaping (Error?) -> Void) {
         self.workQueue.async {
             self.sendBuffer.append((message, completionHandler))
             self.doPendingWork()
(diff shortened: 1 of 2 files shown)


parkera:
It would be a good idea to add a test case that calls into these (if we don't have one already).

lxbndr:
Test added. The test is the same as existing `async`-based `test_webSocket`, but ported to use completion handlers.

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

lxbndr:
Windows CI failed on checkout:
```
C:\Users\swift-ci\jenkins\workspace\swift-corelibs-foundation-PR-windows\swift-docc failed (ret=1): ['git', 'rebase', 'FETCH_HEAD']
b'error: cannot rebase: You have unstaged changes.\nerror: Please commit or stash them.\n'
```

At least Linux is green. But I am not sure if it actually runs these tests. I don't see any output related to URLSession tests.

jrflat:
LGTM thanks! Since web sockets are an experimental feature in curl, we'll need to run it with a curl where they're enabled to test it. I'll test it on a Fedora40 container to confirm

lxbndr:
Just a small note: these API is what is actually used by `async` variants, which are already public. So I'd say it is pretty safe to open up a little more.

jrflat (approved):
Yeah that's true, the tests are also all green for me on a container with web sockets enabled. Looks good and let's merge after Windows passes (I think that issue's been resolved). Thanks again!

jrflat:

PR #5032, opened 2024-07-26
Fix local CMake build

jmschonfeld:
This updates our CMake files to make it possible to build this project locally via CMake without building the entire toolchain. This required 2 main changes:

1. When `dispatch_DIR` is not provided, find dispatch headers from the SDK (looking at `DISPATCH_INCLUDE_PATH` and falling back to `/usr/lib/swift` for linux). This mirrors how the SwiftPM build finds dispatch.
2. Explicitly provide paths to the clang module map files so that the compiler does not encounter ambiguity when the module also exists in the SDK. This mirrors what SwiftPM does today.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,10 +92,21 @@ if(NOT SwiftFoundation_MODULE_TRIPLE)
     mark_as_advanced(SwiftFoundation_MODULE_TRIPLE)
 endif()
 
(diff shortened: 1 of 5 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #5033, opened 2024-07-26
[Windows] Add default dispatch header search path to standalone builds

jmschonfeld:
When building this package via SwiftPM on Windows, we need to include dispatch headers from the appropriate directory in the SDK. We cannot rely on an absolute path such as `/usr/lib/swift` on linux, so instead we use a reasonable default based on the SDKROOT environment variable. I believe this environment variable should be set automatically in a VS developer shell with Swift installed to point to the `Windows.sdk` directory within the currently active swift installation. This allows us to find dispatch in many cases and get closer to a working package build on Windows (while still allowing manual overriding via the `DISPATCH_INCLUDE_PATH` environment variable)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,14 +96,15 @@ endif()
 find_package(dispatch CONFIG)
 if(NOT dispatch_FOUND)
     if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
-        set(DISPATCH_INCLUDE_PATH "/usr/lib/swift" CACHE STRING "A path to where you can find libdispatch headers")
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
The SwiftPM build still fails because we don't (yet) have a way to find headers for libxml and curl, but this at least gets us a step closer

compnerd (approved):
Minor correction on the commit message: `SDKROOT` is setup by Swift and should be available irrespective of the VSDevCmd setup.

jmschonfeld:

iCharlesHu (approved):

jmschonfeld:
Updated to make the same change to the CMake file for standalone CMake builds

jmschonfeld:
@swift-ci please test

jmschonfeld:

PR #5038, opened 2024-07-30
Make sure that the once used for Process is static

parkera:
Do not re-add sources to the run loop for each call to `setup`. Regression from #5000.

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -226,9 +226,9 @@ private func quoteWindowsCommandLine(_ commandLine: [String]) -> String {
 #endif
 
 open class Process: NSObject, @unchecked Sendable {
+    static let once = Mutex(false)
+    
     private static func setup() {
-        let once = Mutex(false)
-        
         once.withLock {
             if !$0 {
                 let thread = Thread {


parkera:
@swift-ci test

Sources/Foundation/Process.swift
```diff
@@ -226,9 +226,9 @@ private func quoteWindowsCommandLine(_ commandLine: [String]) -> String {
 #endif
 
 open class Process: NSObject, @unchecked Sendable {
+    static let once = Mutex(false)
+    
     private static func setup() {
-        let once = Mutex(false)
-        
         once.withLock {
```
jmschonfeld, line 232: Not sure if this needs a `Self.` in front of once, but if it builds it looks good to me!

jmschonfeld (approved):

Sources/Foundation/Process.swift
parkera: Since the function is also static it is inferred.

parkera:
@swift-ci test windows

compnerd (approved):

parkera:
@shahmishal it looks like something might be wrong with the windows hosts?

```
Running ``update_single_repository`` with up to 72 processes.
======UPDATE FAILURES======
C:\Users\swift-ci\jenkins\workspace\swift-corelibs-foundation-PR-windows\swift failed (ret=1): ['git', 'checkout', 'main']
b"error: pathspec 'main' did not match any file(s) known to git\n"
update-checkout failed, fix errors and try again
Build step 'Execute Windows batch command' marked build as failure
```

parkera:
I'm going to go ahead and merge this despite the Windows CI issue.

PR #5037, opened 2024-07-30
Fix CoreFoundation module install path for static build

jmschonfeld:
The CoreFoundation target had a hard-coded module install path of `lib/swift/CoreFoundation`, instead of conditionalizing it for the static build (which goes to `lib/swift_static`). While we don't expect clients to use `CoreFoundation` directly (and they shouldn't), this restores the module to unbreak existing clients that are doing this.

This resolves https://github.com/apple/swift-http-types/issues/61

PR #5043, opened 2024-08-01
[6.0] Fix CoreFoundation install path

jmschonfeld:
Explanation: Re-adds the CoreFoundation module to the static swift SDK
Scope: Should only change the static swift build and fixes a regression
Original PR: https://github.com/apple/swift-corelibs-foundation/pull/5037
Risk: Minimal - minor change to fix a regression that should effectively be additive
Testing: Testing done via swift-ci toolchain builds and local, standalone CMake builds
Reviewer: @iCharlesHu


jmschonfeld:
@swift-ci please test

PR #5046, opened 2024-08-02
Disable test_connectTimeout for flakiness on Linux

jrflat:
`TestURLSession.test_connectTimeout` was failing on Ubuntu 22.04 CI and I also reproduced the failure on Amazon Linux 2. It generally only fails when all `TestURLSession` cases are run together, and I think it's related to stopping and starting the server during the test. Disable it for now while we figure out another way to test this specific connect timeout (a similar request timeout is still being tested by the test case above).

### DIFF
--- a/Tests/Foundation/TestURLSession.swift
+++ b/Tests/Foundation/TestURLSession.swift
@@ -609,6 +609,8 @@ final class TestURLSession: LoopbackServerTest, @unchecked Sendable {
     }
 
     func test_connectTimeout() async throws {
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

parkera (approved):

lxbndr:
I guess the only option is to create dedicated server instance for this test. I'll try implementing that.

jrflat:
Sounds good, thanks! I'm going to merge this for now while we look into that

PR #5048, opened 2024-08-02
Allow for more flexibility in what path SWIFTCI_USE_LOCAL_DEPS means

parkera:
This patch allows SWIFTCI_USE_LOCAL_DEPS to be either `1` (use `..`) or a string to be prepended to the path.

It also moves from the deprecated `swiftLanguageVersion` to `swiftLanguageMode`.

Both of these are required to work with [this benchmarking update in swift-foundation](https://github.com/apple/swift-foundation/pull/810).


parkera:
@swift-ci test

jmschonfeld (approved):

jmschonfeld:

PR #5052, opened 2024-08-05
build: Repair the build on WASI platform

MaxDesiatov:
Cherry picked from https://github.com/apple/swift-corelibs-foundation/pull/4934

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,6 +12,7 @@ let platformsWithThreads: [Platform] = [
     .driverKit,
     .android,
     .linux,
+    .windows,
 ]
 var dispatchIncludeFlags: [CSetting]
 if let environmentPath = Context.environment["DISPATCH_INCLUDE_PATH"] {
--- a/Package.swift
+++ b/Package.swift
@@ -3,6 +3,16 @@
 
 import PackageDescription
 
+let platformsWithThreads: [Platform] = [
+    .iOS,
+    .macOS,
+    .tvOS,
+    .watchOS,
+    .macCatalyst,
+    .driverKit,
+    .android,
+    .linux,
+]
 var dispatchIncludeFlags: [CSetting]
 if let environmentPath = Context.environment["DISPATCH_INCLUDE_PATH"] {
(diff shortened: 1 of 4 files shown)


MaxDesiatov:
@swift-ci test

iCharlesHu (approved):
Thanks Max!

Package.swift
```diff
@@ -31,8 +41,11 @@ let coreFoundationBuildSettings: [CSetting] = [
     .define("DEPLOYMENT_ENABLE_LIBDISPATCH"),
     .define("DEPLOYMENT_RUNTIME_SWIFT"),
     .define("HAVE_STRUCT_TIMESPEC"),
-    .define("SWIFT_CORELIBS_FOUNDATION_HAS_THREADS"),
+    .define("SWIFT_CORELIBS_FOUNDATION_HAS_THREADS", .when(platforms: platformsWithThreads)),
```
jmschonfeld, line 44: Can we make this change to the CMake files as well so that the CMake and SwiftPM builds are in-sync?

Package.swift
```diff
@@ -3,6 +3,16 @@
...
 
+let platformsWithThreads: [Platform] = [
+    .iOS,
+    .macOS,
+    .tvOS,
+    .watchOS,
+    .macCatalyst,
+    .driverKit,
+    .android,
+    .linux,
```
jmschonfeld, line 14: Does this intentionally exclude Windows? This would be a change in behavior for that platform (unless this is already an issue today for that platform)

MaxDesiatov, line 14:
```suggestion
    .linux,
    .windows,
```

MaxDesiatov: That was accidental, fixed now.

Package.swift
MaxDesiatov: Done

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

jmschonfeld (approved):

PR #5056, opened 2024-08-07
CMake: Fix missing librt target

etcwilde:
When building the static SDK, dispatch is built as a static archive, so its dependency on librt is exposed as part of its link interface. This shows up in the `dispatchConfig.cmake` file imported by `find_package(dispatch CONFIG)`.

```
CMake Error at build/x86_64/dispatch/cmake/modules/dispatchExports.cmake:71 (set_target_properties):
  The link interface of target "dispatch" contains:

    RT::rt

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

It would be great if the dispatch config could include that target to ensure that everyone is using the same librt, it does not, so we have to find and include it ourselves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,7 @@ if(NOT SwiftFoundation_MODULE_TRIPLE)
 endif()
 
 # System dependencies
+find_package(LibRT)
 find_package(dispatch CONFIG)
 if(NOT dispatch_FOUND)
     if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
--- a/cmake/modules/FindLibRT.cmake
+++ b/cmake/modules/FindLibRT.cmake
@@ -4,7 +4,7 @@
 #
 # Find librt library and headers.
 #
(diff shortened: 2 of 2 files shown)


etcwilde:
@swift-ci please test

etcwilde:
(There are more failures in the static SDK builds, figured I'd put this one up for now.)

CMakeLists.txt
```diff
@@ -93,6 +93,7 @@ if(NOT SwiftFoundation_MODULE_TRIPLE)
 endif()
 
 # System dependencies
+find_package(LibRT)
```
jmschonfeld, line 96: For my context, is there a reason that's preventing us from fixing this on the dispatch side instead of applying a workaround to Foundation (and presumably any other project that imports dispatch via CMake in the toolchain)? Is there a way we can just add this to the exported config for dispatch?

etcwilde:
This will be fixed as part of https://github.com/apple/swift-corelibs-foundation/pull/5069.
Closing this smaller PR.

PR #5057, opened 2024-08-07
build: Repair the build on WASI platform (take 3)

kateinoigakukun:
I forgot to tell @MaxDesiatov we need several additional changes other than https://github.com/apple/swift-corelibs-foundation/pull/5052


kateinoigakukun:
@swift-ci test

MaxDesiatov:
I merged that one, but that introduced conflicts unfortunately 😢

Sources/CoreFoundation/CFBundle.c
```diff
@@ -600,7 +600,7 @@ CFBundleRef CFBundleGetBundleWithIdentifier(CFStringRef bundleID) {
 #if TARGET_OS_WASI
     hint = NULL;
 #else
-    hint = __builtin_frame_address(0);
+    hint = __builtin_return_address(0);
```
kateinoigakukun, line 603: It was accidentally changed in https://github.com/apple/swift-corelibs-foundation/pull/5052

kateinoigakukun:
@swift-ci test

CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ endif()
 
 # System dependencies
 find_package(dispatch CONFIG)
-if(NOT dispatch_FOUND)
+if(NOT dispatch_FOUND AND NOT CMAKE_SYSTEM_NAME STREQUAL WASI)
```
jmschonfeld, line 97: If WASI doesn't have dispatch, can we wrap this whole section in an `if` block to avoid even trying `find_package(dispatch CONFIG)` if we know it won't (or shouldn't) exist?

CMakeLists.txt
```diff
@@ -126,16 +125,10 @@ list(APPEND _Foundation_common_build_flags
     "-Wno-switch"
     "-fblocks")
 
-if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
-  list(APPEND _Foundation_common_build_flags
-    "-D_WASI_EMULATED_SIGNAL"
```
jmschonfeld, line 131: Just confirming - is it intentional that we no longer provide this define to the C targets / we don't provide `HAVE_STLCPY`/`HAVE_STRLCAT` at all?

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+## See CONTRIBUTORS.md for the list of Swift project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+# Build the vendored version of the BlocksRuntime library, which is used by
+# platforms that don't support libdispatch.
+
+add_library(BlocksRuntime
```
jmschonfeld, line 18: We probably want to always make this a static library rather than a separate shared library that needs to be installed - just like CF itself

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+    data.c
+    runtime.c)
+
+target_include_directories(BlocksRuntime PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/include
```
jmschonfeld, line 23: @etcwilde - remind me, this is not necessary, right?

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+    # For CFTargetConditionals.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/../include)
+
+set_target_properties(BlocksRuntime PROPERTIES
+    POSITION_INDEPENDENT_CODE FALSE)
+
+add_library(BlocksRuntime::BlocksRuntime ALIAS BlocksRuntime)
```
jmschonfeld, line 30: For my context, could you explain a little bit more about why the above two directives are needed?

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+if(NOT BUILD_SHARED_LIBS)
+    set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS BlocksRuntime)
+    install(TARGETS BlocksRuntime
+        ARCHIVE DESTINATION lib/swift$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:_static>/${SWIFT_SYSTEM_NAME}
```
jmschonfeld, line 35: Since we're already in the `NOT BUILD_SHARED_LIBS` block, I think we can drop the additional check here to make this more readable/understandable

Sources/FoundationXML/XMLParser.swift
```diff
@@ -471,7 +471,7 @@ open class XMLParser : NSObject {
     open var allowedExternalEntityURLs: Set<URL>?
     
 #if os(WASI)
-    private static var _currentParser: XMLParser?
+    private nonisolated(unsafe) static var _currentParser: XMLParser?
```
jmschonfeld, line 474: This seems unsafe at first look, since any thread can call `setCurrentParser` and effectively traffic `XMLParser` between actors. Can you elaborate a bit more on why this is necessary?

Sources/_CFXMLInterface/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ target_include_directories(_CFXMLInterface
         ../CoreFoundation/include
     PRIVATE
         ../CoreFoundation/internalInclude
-        /usr/include/libxml2/)
+        ${LIBXML2_INCLUDE_DIR})
```
jmschonfeld, line 23: Where is this defined?

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/../include)
+
+set_target_properties(BlocksRuntime PROPERTIES
+    POSITION_INDEPENDENT_CODE FALSE)
+
+add_library(BlocksRuntime::BlocksRuntime ALIAS BlocksRuntime)
+
+if(NOT BUILD_SHARED_LIBS)
+    set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS BlocksRuntime)
+    install(TARGETS BlocksRuntime
```
jmschonfeld, line 34: Does WASI have a static SDK? If so, unfortunately just installing the library isn't sufficient. We need to ensure that it is auto linked from every swift module that imports the library. See the `autolink` flags added to Foundation/FoundationNetworking/FoundationXML for an example of this. That way the linker will link this library even though there isn't a public import of the module

Sources/_CFXMLInterface/CMakeLists.txt
kateinoigakukun: it's defined by [`find_package(LibXml2)`](https://cmake.org/cmake/help/latest/module/FindLibXml2.html)

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
kateinoigakukun: > Does WASI have a static SDK?
We don't support shared libraries but only support static libraries on WASI.

Given that BlocksRuntime is a private dependency of CoreFoundation, I think the best way is to build BlocksRuntime as object library and include the object files in libCoreFoundation.a. In this way, we don't need to worry about how to place it in the toolchain and autolinking stuff.

Sources/FoundationXML/XMLParser.swift
kateinoigakukun: WASI preview1 ABI is always single-threaded, so memory access to the same storage should not happen simultaneously. So any data storage should be safe to access from any isolation context on WASI, but  6 lang mode requires explicit marking here.

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
kateinoigakukun: We no longer need installation since we made BlocksRuntime as an object library

Sources/FoundationXML/XMLParser.swift
jmschonfeld: Perhaps the better question is what does actor isolation mean on WASI where everything is single threaded? This will allow this mutable state to pass between actor isolations on WASI, is that ok because we know it's single threaded? Or should this instead be some form of task-local storage to prevent it from traveling across actors?

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
jmschonfeld: Ah right sorry, had a different thought mixed up

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+# Build the BlocksRuntime as an object library, shipped as a part
+# of libCoreFoundation.
+add_library(BlocksRuntime OBJECT
+    data.c
+    runtime.c)
+
+target_include_directories(BlocksRuntime PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/include
+    # For CFTargetConditionals.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/../include)
```
jmschonfeld, line 27: Should this be `PRIVATE`?

Sources/FoundationXML/XMLParser.swift
kateinoigakukun: We don't need task local storage at all here because there is no suspension point during `XMLParser.parse` call. No suspension point means no Task switching, thus it's guaranteed that `XMLParser.setCurrentParser(self)` is not called twice before resetting the storage by `XMLParser.setCurrentParser(nil)`.
The only exception is the case where user provided `XMLParserDelegate` methods call `XMLParser.parse()`, but the reentrancy issue is not specific to `os(WASI)` branch.

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
kateinoigakukun: No, it needs to be public because a public header `Block.h` depends on `CFTargetConditionals.h`.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+set_target_properties(BlocksRuntime PROPERTIES
+    POSITION_INDEPENDENT_CODE FALSE)
```
jmschonfeld, line 30: I think you addressed the other directive in https://github.com/apple/swift-corelibs-foundation/pull/5057#discussion_r1707333900, but it looks like this one is still here - is there a particular reason why this needs to be disabled for this target, while CoreFoundation has it enabled?

Sources/FoundationXML/XMLParser.swift
jmschonfeld: Thanks for the explanation. The reentrancy possibility isn't unique to WASI, but its ability to manipulate the underlying `nonisolated(unsafe)` variable across actor isolations is unique to WASI with this change (on other platforms, that variable is thread-specific - which should probably be marked as unavailable from async for this reason).

The example I'm thinking of is: someone calls `XMLParser.parse()`, in a delegate callout they create a new detached task on a different actor calling `XMLParser.parse()` again. At this point, we have two actors manipulating the same underlying `nonisolated(unsafe)` variable. Typically, that'd be a concurrency violation because two actors are mutating the same shared state. Since WASI is single-threaded, we know that this can't data race in practice, but it's still a violation in the Swift 6 concurrency design and I'm not sure how to rationalize that with the thread model of WASI at the moment (cc @hborla in case you have any insight here)

kateinoigakukun: > The example I'm thinking of is: someone calls XMLParser.parse(), in a delegate callout they create a new detached task on a different actor calling XMLParser.parse() again.

Having a closer look, I noticed the problem exists even in non-WASI code path since jobs associated with two different Tasks can be scheduled on the same thread, and in that case two different tasks can manipulate the same thread local variable.

So we had two issues in the existing code, 1. reentrancy issue, 2. sharing mutable storage across Tasks. I'll send a fix for them in another PR and exclude the commit from this PR. Does it make sense?

Sources/CoreFoundation/BlockRuntime/CMakeLists.txt
kateinoigakukun: I couldn't find any reason to disable PIC. Removed it in 3b8485a75c5fd9a7f7877961ceadbc6521dcf563

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

kateinoigakukun:
@swift-ci test Windows

Sources/FoundationXML/XMLParser.swift
kateinoigakukun: @jmschonfeld Opened another PR to fix the issues https://github.com/apple/swift-corelibs-foundation/pull/5061

PR #5060, opened 2024-08-08
Fixing post-swift-foundation corelibs-foundation static SDK builds

etcwilde:
Putting back definitions and flags that were dropped when doing the Swift-Foundation re-core that are necessary when building for the static SDK. This is primarily checks for symbol availability.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,8 +77,28 @@ endif()
 FetchContent_MakeAvailable(SwiftFoundationICU SwiftFoundation)
 
 include(CheckLinkerFlag)
+include(CheckSymbolExists)
 
 check_linker_flag(C "LINKER:--build-id=sha1" LINKER_SUPPORTS_BUILD_ID)
+
+check_symbol_exists("strlcat" "string.h" HAVE_STRLCAT)
+check_symbol_exists("strlcpy" "string.h" HAVE_STRLCPY)
+check_symbol_exists("issetugid" "unistd.h" HAVE_ISSETUGID)
+add_compile_definitions(
+  $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${HAVE_STRLCAT}>>:HAVE_STRLCAT>
+  $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${HAVE_STRLCPY}>>:HAVE_STRLCPY>
+  $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${HAVE_ISSETUGID}>>:HAVE_ISSETUGID>)
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux OR ANDROID)
+  add_compile_definitions($<$<COMPILE_LANGUAGE:C>:_GNU_SOURCE>)
+endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+  check_symbol_exists(sched_getaffinity "sched.h" HAVE_SCHED_GETAFFINITY)
+  add_compile_definitions($<$<COMPILE_LANGUAGE:C>:HAVE_SCHED_GETAFFINITY>)
+endif()
+
+add_compile_options($<$<COMPILE_LANGUAGE:C>:-fblocks>)
 
 # Precompute module triple for installation
 if(NOT SwiftFoundation_MODULE_TRIPLE)
@@ -93,6 +113,7 @@ if(NOT SwiftFoundation_MODULE_TRIPLE)
 endif()
 
 # System dependencies
+find_package(LibRT)
 find_package(dispatch CONFIG)
 if(NOT dispatch_FOUND)
     if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Android")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,12 @@ include(CheckSymbolExists)
 
 check_linker_flag(C "LINKER:--build-id=sha1" LINKER_SUPPORTS_BUILD_ID)
 
+# Detect if the system libc defines symbols for these functions.
+# If it is not availble, swift-corelibs-foundation has its own implementations
+# that will be used. If it is available, it should not redefine them.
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -77,9 +77,29 @@ endif()
...
 
 check_linker_flag(C "LINKER:--build-id=sha1" LINKER_SUPPORTS_BUILD_ID)
 
+check_symbol_exists("strlcat" "string.h" HAVE_STRLCAT)
+check_symbol_exists("strlcpy" "string.h" HAVE_STRLCPY)
+check_symbol_exists("issetugid" "unistd.h" HAVE_ISSETUGID)
+add_compile_definitions(
+  $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${HAVE_STRLCAT}>>:HAVE_STRLCAT>
+  $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${HAVE_STRLCPY}>>:HAVE_STRLCPY>
+  $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${HAVE_ISSETUGID}>>:HAVE_ISSETUGID>)
```
jmschonfeld, line 90: We've discussed these before a handful of times (see https://github.com/apple/swift-corelibs-foundation/pull/5013) but SwiftPM does not support these types of definitions so this is incompatible with the SwiftPM build. Instead I believe we need to conditionally define these in the headers/implementation files rather than in the CMake/package manifest files

CMakeLists.txt
```diff
@@ -77,9 +77,29 @@ endif()
...
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux OR ANDROID)
+  add_compile_definitions($<$<COMPILE_LANGUAGE:C>:_GNU_SOURCE>)
```
jmschonfeld, line 93: This is already added to `_Foundation_common_build_flags` which is applied to all of our C targets - is there a reason this is needed additionally?

CMakeLists.txt
```diff
@@ -77,9 +77,29 @@ endif()
...
+endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+  check_symbol_exists(sched_getaffinity "sched.h" HAVE_SCHED_GETAFFINITY)
+  add_compile_definitions($<$<COMPILE_LANGUAGE:C>:HAVE_SCHED_GETAFFINITY>)
```
jmschonfeld, line 98: Same comment as above about the SwiftPM incompatibility

CMakeLists.txt
```diff
@@ -77,9 +77,29 @@ endif()
...
+endif()
+
+add_compile_options($<$<COMPILE_LANGUAGE:C>:-fblocks>)
```
jmschonfeld, line 101: Same comment as above about this already being added to `_Foundation_common_build_flags` which is applied to all C targets

CMakeLists.txt
```diff
@@ -93,6 +113,7 @@ if(NOT SwiftFoundation_MODULE_TRIPLE)
 endif()
 
 # System dependencies
+find_package(LibRT)
```
jmschonfeld, line 116: Copying my comment over from https://github.com/apple/swift-corelibs-foundation/pull/5056#discussion_r1706335870:

>For my context, is there a reason that's preventing us from fixing this on the dispatch side instead of applying a workaround to Foundation (and presumably any other project that imports dispatch via CMake in the toolchain)? Is there a way we can just add this to the exported config for dispatch?

CMakeLists.txt
etcwilde: SPM will need to learn how to query what is available in the SDK then...
From having worked with internal vs external SDKs, there are certain Linux distros (that I have to support) where certain symbols are only exposed if you're working internally. The version numbers, outward names, and other macro definitions are identical in both cases, but the available functions aren't the same, so trying to hard-code a mish-mash of definitions into the source file statically isn't going to work.

jmschonfeld: I agree with the sentiment and definitely understand that the workaround is not a long term solution. This is definitely an area where SwiftPM should grow to support these needs.

For the time being since we don't have that capability in SwiftPM yet, are we able to work around that limitation for this case specifically? SwiftPM builds are one of our main workflows for developing on this project, and is how we run tests for this project as well so unfortunately neither adding a new SwiftPM feature nor breaking the SwiftPM build is an option. But if we have a way to get these features to work specifically (or a way to detect this for only the static build since SwiftPM doesn't build the Static SDK) then that seems like a workable solution until SwiftPM is capable of doing this

etcwilde:
https://github.com/apple/swift-foundation/pull/838
@swift-ci please test

CMakeLists.txt
etcwilde: The checks for the macros are already in this codebase, this just added the check back to CMake, so as far as SPM is concerned, it will continue to behave exactly as it is today. This just means that folks who are building for systems that define these functions in their libc have a way to build corelibs-foundation as well without having duplicate-definitions.

CMakeLists.txt
jmschonfeld: Ok thanks for the explanation, this seems fine for now then

CMakeLists.txt
jmschonfeld: Ok, as it stands today, `HAVE_STRLCAT` and `HAVE_STRLCPY` are defined to `1` on WASI via:

```c
#if TARGET_OS_WASI
#define HAVE_STRLCPY 1
#define HAVE_STRLCAT 1
#endif
```

so it sounds like this is for the Static SDK based on Musl - since we don't currently expect to be able to build the Static SDK with SwiftPM, can you add a comment here that makes it explicitly noted that this is to support building against Musl for the Static SDK and therefore does not need to be replicated in the SwiftPM package manifest, just to avoid future confusion?

etcwilde:
@swift-ci please test

jmschonfeld:
I don't have any remaining comments here, but since this is to the `release/6.0` branch @itingliu will need to approve. @etcwilde do you have a PR open to `main` for this yet? Otherwise these changes won't make it to the next release.

etcwilde:

itingliu (approved):

al45tair (approved):

PR #5061, opened 2024-08-09
[XMLParser] Fix reentrancy issue around `currentParser`

kateinoigakukun:
Instead of thread-local storage, use `TaskLocal` to store the current
parser. This solves three issues:

1. If someone calls `XMLParser.parse()` with a new parser instance in
   a delegate method call, it overwrote the current parser and wrote
   it back after the call as `nil`, not the previous current parser.
   This reentrancy issue can be a problem especially when someone uses
   external entity resolving since the feature depends on the current
   parser tracking. Using `TaskLocal` solves this issue since it tracks
   values as a stack and restores the previous value at the end of the
   `withValue` call.
2. Since jobs of different tasks can be scheduled on the same thread,
   different tasks can refer to the same thread-local storage. This
   wouldn't be a problem for now since the `parse()` method doesn't
   have any suspention points and different tasks can't run on the same
   thread during the parsing. However, it's better to use `TaskLocal`
   to leverage the concurrency model of Swift.
3. The global variable `_currentParser` existed in the WASI platform
   path but it's unsafe in the Swift concurrency model. It wouldn't be a
   problem on WASI since it's always single-threaded, we should avoid
   platform-specific assumption as much as possible.

The issue was revealed in https://github.com/apple/swift-corelibs-foundation/pull/5057#discussion_r1709934773

### DIFF
--- a/Sources/FoundationXML/XMLParser.swift
+++ b/Sources/FoundationXML/XMLParser.swift
@@ -398,9 +398,7 @@ extension XMLParser : @unchecked Sendable { }
 
 open class XMLParser : NSObject {
     private var _handler: _CFXMLInterfaceSAXHandler
-#if !os(WASI)
     internal var _stream: InputStream?
-#endif
     internal var _data: Data?
 
     internal var _chunkSize = Int(4096 * 32) // a suitably large number for a decent chunk size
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

kateinoigakukun:
Hmm, the new test case fails only in CI...

Sources/FoundationXML/XMLParser.swift
```diff
@@ -469,33 +461,35 @@ open class XMLParser : NSObject {
...
+    private struct _CurrentParser: @unchecked Sendable {
+        let parser: XMLParser?
+
+        static var `default`: _CurrentParser {
+            return _CurrentParser(parser: nil)
         }
-#endif
+    }
+
+    @TaskLocal
```
jmschonfeld, line 484: Are we always guaranteed to have a Task? For example, what happens if someone creates a thread (outside of swift concurrency) and calls `XMLParser.parse` - will it allow us to use Task-local storage?

kateinoigakukun: Yeah, TaskLocal falls back to TLS if the current execution is not bound to a Task. https://github.com/swiftlang/swift-evolution/blob/main/proposals/0311-task-locals.md#binding-task-local-values-from-synchronous-functions

kateinoigakukun:
@swift-ci test Linux

kateinoigakukun:
@swift-ci test

kateinoigakukun:
Ok, a part of the new test depended on the behavior of libxml2 v2.9.11 and later, but the CI uses v2.9.10. Fixed it by adjusting test case.

Sources/FoundationXML/XMLParser.swift
```diff
@@ -469,33 +461,35 @@ open class XMLParser : NSObject {
...
-        } else {
-            return nil
+    /// The current parser is stored in a task local variable to allow for
+    /// concurrent parsing in different tasks with different parsers.
+    ///
+    /// Rationale for `@unchecked Sendable`:
+    /// While the ``XMLParser`` class itself is not `Sendable`, `TaskLocal`
+    /// requires the value type to be `Sendable`. The sendability requirement
+    /// of `TaskLocal` is only for the "default" value and values set with
+    /// `withValue` will not be shared between tasks.
```
jmschonfeld, line 472: I've actually learned recently while investigating replacing `FoundationEssentials._ThreadLocal` with `TaskLocal` that this is not the case sadly. `TaskLocal` values are automatically inherited by child tasks. For that reason, the value needs to be `Sendable` because it is possible for a detached child task to also reference this value. I think that means that if someone were to call an API that used this value from a detached task created during a delegate call that unfortunately we'd have a concurrency safety violation

kateinoigakukun: That's a really good point. So this means we have to keep it in TLS instead of TaskLocal...

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

MaxDesiatov:
@swift-ci test

kateinoigakukun:
Ok, test fixed. @jmschonfeld Could you take a look again? 🙏

kateinoigakukun:
@jmschonfeld Gentle ping :)

Sources/FoundationXML/XMLParser.swift
```diff
@@ -469,33 +461,58 @@ open class XMLParser : NSObject {
...
+            // Use the existing context if it exists
+            ctx = current
         } else {
-            Thread.current.threadDictionary.removeObject(forKey: "__CurrentNSXMLParser")
+            // Create a new context in TLS
+            ctx = _CurrentParserContext()
+            _currentParserContext = ctx
         }
-#endif
+        // Push the parser onto the stack
```
grynspan, line 509: Does this need to be a stack at all? Can't the actual call stack itself be used?

```swift
let oldValue = ctx._currentParser
ctx._currentParser = parser
defer {
  ctx._currentParser = oldValue
}
```

This may simplify your design overall.

kateinoigakukun: That totally makes sense!

kateinoigakukun:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

grynspan:
I'm not a maintainer/codeowner of this repo so I'm going to hold off on approving, but don't take that as a blocker for merging. :)

jmschonfeld (approved):
Thanks for working through this one! This one LGTM now

PR #5063, opened 2024-08-10
Remove the workaround for WASI errno conflict

kateinoigakukun:
As we have fixed the conflict in the WASILibc overlay module while porting swift-foundation, we can remove the workaround from corelibs-foundation too.


kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #5064, opened 2024-08-10
Make curl an optional dependency when not building FoundationNetworking

kateinoigakukun:
When building for WASI, FoundationNetworking is not supported, so we should not require curl to be present. This change makes curl an optional dependency when FoundationNetworking is not being built.


kateinoigakukun:
@swift-ci test

Sources/CMakeLists.txt
```diff
@@ -14,12 +14,12 @@
...
 add_subdirectory(CoreFoundation)
 add_subdirectory(_CFXMLInterface)
-if(NOT CMAKE_SYSTEM_NAME STREQUAL "WASI")
+if(FOUNDATION_BUILD_NETWORKING)
     add_subdirectory(_CFURLSessionInterface)
 endif()
 add_subdirectory(Foundation)
 add_subdirectory(FoundationXML)
-if(NOT CMAKE_SYSTEM_NAME STREQUAL "WASI")
+if(FOUNDATION_BUILD_NETWORKING)
```
jmschonfeld, line 22: Should we make an equivalent change to the Package manifest to not build `FoundationNetworking` targets for WASI?

kateinoigakukun: SwiftPM does not have a way to define a target conditionally based on the target triple, so developers need to specify `--target` or `--product` explicitly instead of excluding them in Package.swift manifest.

jmschonfeld (approved):

PR #5065, opened 2024-08-12
Include userInfo in Error descriptions

jakepetroules:
This more closely matches Darwin behavior, and makes it significantly easier to debug issues like "File was not found" which presently include no reference whatsoever to the actual file path.

### DIFF
--- a/Sources/Foundation/NSError.swift
+++ b/Sources/Foundation/NSError.swift
@@ -182,7 +182,11 @@ open class NSError : NSObject, NSCopying, NSSecureCoding, NSCoding, @unchecked S
     }
     
     override open var description: String {
-        return "Error Domain=\(domain) Code=\(code) \"\(localizedFailureReason ?? "(null)")\""
+        var result = "Error Domain=\(domain) Code=\(code) \"\(localizedFailureReason ?? "(null)")\""
+        if !userInfo.isEmpty {
+            result += "UserInfo={\(userInfo.map { "\($0)=\($1)"}.joined(separator: ", "))}"
+        }
+        return result
     }
     
     // -- NSObject Overrides --


jakepetroules:
@swift-ci please test

Sources/Foundation/NSError.swift
```diff
@@ -182,7 +182,11 @@ open class NSError : NSObject, NSCopying, NSSecureCoding, NSCoding, @unchecked S
     }
     
     override open var description: String {
-        return "Error Domain=\(domain) Code=\(code) \"\(localizedFailureReason ?? "(null)")\""
+        var result = "Error Domain=\(domain) Code=\(code) \"\(localizedFailureReason ?? "(null)")\""
+        if !userInfo.isEmpty {
+            result += "UserInfo={\(userInfo.map { "\($0)=\($1)"}.joined(separator: ", "))}"
```
jmschonfeld, line 187: Thanks for looking into this! I've been meaning to fix this for a while and I agree this is definitely the direction we should take. One note, in addition to the user info we have the `localizedFailureReason` here, but on Darwin we actually effectively a combination of `NSDebugDescriptionErrorKey`/`NSLocalizedDescriptionKey`/`NSLocalizedFailureReasonErrorKey` (in that order of preference). Do you think it'd be worth checking the debug description / localized description keys first here to potentially provide a description that's more helpful and/or more on par with Darwin?

jakepetroules: I'm not sure; I don't have a particularly strong opinion on this.

Do you think this is something we could land as-is as a first step, though?

jmschonfeld: Yeah that seems fine

jmschonfeld (approved):

PR #5067, opened 2024-08-13
Replace dirty hack with (void)isHFSPlusCanonical;

RSilicon:


RSilicon:
@compnerd Thoughts?

RSilicon:
@swift-ci please test

compnerd:
I'm not sure about the value of this change tbh. There is active work to re-core to swift-foundation and that will entirely remove CoreFoundation. In the mean time, this diverges from the internal copy of CoreFoundation which might make it more difficult to backport changes if critical.

PR #5069, opened 2024-08-14
[main] Fixing post-swift-foundation corelibs-foundation MUSL static SDK builds

etcwilde:
Putting back definitions and flags that were dropped when doing the Swift-Foundation re-core that are necessary when building for the static SDK. This is primarily checks for symbol availability.

This will not affect SwiftPM builds, which will continue to define `strlcpy` and `strlcat` everywhere, even on systems that define them in the libc, resulting in build failures. The CMake build will work in these environments.

6.0: https://github.com/apple/swift-corelibs-foundation/pull/5060

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,8 +95,28 @@ endif()
 FetchContent_MakeAvailable(SwiftFoundationICU SwiftFoundation)
 
 include(CheckLinkerFlag)
+include(CheckSymbolExists)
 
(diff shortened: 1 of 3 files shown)


jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

al45tair (approved):

etcwilde:
@swift-ci please test Linux

PR #5070, opened 2024-08-15
Fix CFTimeZone crashes on Windows

jmschonfeld:
CFTimeZone on Windows requires mappings between the Windows and Olson timezone identifiers. We unfortunately lost those mappings in the build for the re-core. This PR adds those mappings back which resolves crashes when calling into `CFTimeZone`. We add them back as literals in code rather than as plist resource files to make it easily compatible with both the CMake and SwiftPM builds. I've confirmed that when running tests on Windows the tests no longer crash with `CFTimeZone` issues


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:

compnerd (commented):
I'm okay with the approach, but I think that we should update the script that generates the mapping.

jmschonfeld:
> I'm okay with the approach, but I think that we should update the script that generates the mapping.

Ah I didn't realize there was a script that generates the mapping - do you know where that lives?

jmschonfeld:
@swift-ci please test

jmschonfeld:

compnerd:
> > I'm okay with the approach, but I think that we should update the script that generates the mapping.
> 
> Ah I didn't realize there was a script that generates the mapping - do you know where that lives?

I could've sworn that I checked that in >.< ... maybe I didn't - it was some XSLT horror. I might've just shoved it into the commit message :(. For now, lets go ahead and get this merged.

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
```diff
@@ -0,0 +1,552 @@
...
+};
+
+static CFStringRef _CFTimeZoneIdentifierPairCopyValue(__CFTimeZoneIdentifierPair *data, CFStringRef key) {
+    char buffer[128] = {0};
+    if (!CFStringGetCString(key, buffer, 128, kCFStringEncodingASCII)) {
+        return NULL;
+    }
+    __CFTimeZoneIdentifierPair *current = data;
+    while (current->source) {
+        if (strcmp(current->source, buffer) == 0) {
```
itingliu, line 536: We can probably do better than a linear search as a polish later

itingliu (approved):

itingliu:
Do we want this in release/6.0 too if it's crashing the tests?

jmschonfeld:
> Do we want this in release/6.0 too if it's crashing the tests?

Yeah, this causes crashes in things that call down to some `CFTimeZone` functions (like `DateFormatter`) so I'll cherry-pick this back to `release/6.0` too

PR #5075, opened 2024-08-17
Fix various typos

Coeur:


Sources/CoreFoundation/CFDateFormatter.c
```diff
@@ -122,21 +122,21 @@ static void _CFDateFormatterStripAMPMIndicators(UniChar **bpat, int32_t *bpatlen
     }
 }
 
-CFStringRef CFDateFormatterCreateDateFormatFromTemplate(CFAllocatorRef allocator, CFStringRef tmplate, CFOptionFlags options, CFLocaleRef locale) {
+CFStringRef CFDateFormatterCreateDateFormatFromTemplate(CFAllocatorRef allocator, CFStringRef template, CFOptionFlags options, CFLocaleRef locale) {
```
parkera, line 125: Usually this kind of thing is to avoid the reserved C++ name.

Coeur: good point: reverted in 921591b24b2c9a21c4d3817950501d40e7b4c743

parkera:
@swift-ci test

Tests/Foundation/TestJSONSerialization.swift
```diff
@@ -1026,7 +1026,7 @@ extension TestJSONSerialization {
             }
         }
         //test serialize values grater than 1 with maxFractionDigits = 15
-        func excecute_testSetGraterThanOne() {
+        func execute_testSetGraterThanOne() {
```
parkera, line 1029: Actually this misses a typo - `execute_testSetGreaterThanOne`. Same for comment above it.

parkera (approved):
There was a follow-up comment, but I'm going to merge this and the other one can be done separately.

Tests/Foundation/TestJSONSerialization.swift
Coeur: Oh my, good catch. Well, next time I do a round of verifications, maybe in a year or two.

Coeur: #5290

PR #5077, opened 2024-08-20
Add stub for Testing module

jmschonfeld:
Now that swift-testing is present in the toolchain, SwiftPM builds of the test target in this package will automatically import `Testing`, causing our test target to link the `Testing` library in the toolchain. This is problematic, because the `Testing` module links Foundation (from the toolchain) which results in a variety of issues and crashes when running our tests caused by the presence of two Foundations (classes of the same name are indistinguishable and one implementation is nondeterministically picked leading to failures when bridging). To resolve this, we will use the same solution we found for XCTest - to vendor a small copy of the library within our project that will be picked up instead of the toolchain version. In our case, we only need a small stub because we don't actually use the `Testing` module yet.

In the future, now that SwiftPM supports cycles in package dependencies as long as there is no target-level cycle, we could update the package build to use the package build of swift-testing (which we could have rely on the package build of Foundation thus ensuring only one Foundation is present) but for now this simple solution gets us building and testing again.

### DIFF
--- a/Sources/Testing/Testing.swift
+++ b/Sources/Testing/Testing.swift
@@ -1,0 +1,25 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+#if canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif os(WASI)
+import WASILibc
+#elseif canImport(CRT)
+import CRT
+#endif
+
+
+// This function is used to mimic a bare minimum of the swift-testing library. Since this package has no swift-testing tests, we simply exit.
+// The test runner will automatically call this function when running tests, so it must exit gracefully rather than using `fatalError()`.
+public func __swiftPMEntryPoint(passing _: (any Sendable)? = nil) async -> Never {
+		exit(0)
+}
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

grynspan (approved):
We hates it! :shipit:

jmschonfeld:
Confirmed that the tests still pass with this change, and they also pass with the linked change so this should resolve the issue that the linked PR was originally hitting when re-ordering the swift-testing build

Sources/Testing/Testing.swift
```diff
@@ -0,0 +1,25 @@
...
+import WASILibc
+#elseif canImport(CRT)
+import CRT
+#endif
+
+
+// This function is used to mimic a bare minimum of the swift-testing library. Since this package has no swift-testing tests, we simply exit.
+// The test runner will automatically call this function when running tests, so it must exit gracefully rather than using `fatalError()`.
+public func __swiftPMEntryPoint(passing _: (any Sendable)? = nil) async -> Never {
+		exit(0)
```
grynspan, line 24: Two tabs deep here?

jmschonfeld: Whoops, not sure how that happened

jmschonfeld:
@swift-ci please test

jmschonfeld:

rintaro (approved):

Sources/Testing/Testing.swift
```diff
@@ -0,0 +1,25 @@
...
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+
+#if canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif os(WASI)
+import WASILibc
+#elseif canImport(CRT)
```
iCharlesHu, line 16: nit I think we should might as well add `#if canImport(Bionic)` here as well. But def none blocking

iCharlesHu (approved):

Sources/Testing/Testing.swift
jmschonfeld: Fair point - I'm not sure if we've tried running SwiftPM tests on android for this repo yet so I'll follow up with that later once we try to do that

PR #5080, opened 2024-08-29
Add one more XCTAssert for isLeapMonth in TestDate.swift

lamtrinhdev:
`isLeapMonth` is a component in `Date`. Add one more XCTAssert for `isLeapMonth` in `test_recreateDateComponentsFromDate`.


parkera:
We haven't traditionally always set this value (sometimes it is nil, depending on calendar). Is there something in particular we are trying to validate here?

parkera:
@swift-ci test

lamtrinhdev:
> We haven't traditionally always set this value (sometimes it is nil, depending on calendar). Is there something in particular we are trying to validate here?

Thank you for your review, @parkera . No particular, just see the test case that cover almost paramaters without it, and I would like to add it into test case.

lamtrinhdev:
Dear @parkera ,

It seems `Swift Test macOS Platform` has the problem. I have observed some previous merged PRs that have also failed with the `Swift Test macOS Platform`.

Thanks,
Lam

PR #5081, opened 2024-08-30
Fix compiler error in CoreFoundation when building in C++ mode

egorzhdan:
`__CFAllocatorRespectsHintZeroWhenAllocating` has two declarations in different headers: `ForFoundationOnly.h` and `ForSwiftFoundationOnly.h`. One of the declarations was under `extern "C"` block, the other one wasn't. Clang accepts this in C language mode, but it is a hard compiler error in C++ language mode due to a linkage mismatch.

This was blocking clients from using CoreFoundation in Swift projects that enable C++ interoperability.

This change makes sure that both declarations of `__CFAllocatorRespectsHintZeroWhenAllocating` are under `extern "C"` blocks.

### DIFF
--- a/Sources/CoreFoundation/include/ForFoundationOnly.h
+++ b/Sources/CoreFoundation/include/ForFoundationOnly.h
@@ -57,6 +57,8 @@ CF_IMPLICIT_BRIDGING_DISABLED
 #include <mach/mach_time.h>
 #endif
 
+_CF_EXPORT_SCOPE_BEGIN
+
 #if __BLOCKS__
 /* These functions implement standard error handling for reallocation. Their parameters match their unsafe variants (realloc/CFAllocatorReallocate). They differ from reallocf as they provide a chance for you to clean up a buffers contents (in addition to freeing the buffer, etc.)
  
@@ -69,9 +71,12 @@ CF_EXPORT void *_Nonnull __CFSafelyReallocateWithAllocator(CFAllocatorRef _Nulla
 #endif
 
 Boolean __CFAllocatorRespectsHintZeroWhenAllocating(CFAllocatorRef _Nullable allocator);
+
 typedef CF_ENUM(CFOptionFlags, _CFAllocatorHint) {
     _CFAllocatorHintZeroWhenAllocating = 1
 };
+
+_CF_EXPORT_SCOPE_END
 
 #define NSISARGTYPE void * _Nullable
 


egorzhdan:
@swift-ci please test

Sources/CoreFoundation/include/ForFoundationOnly.h
```diff
@@ -68,7 +68,10 @@ CF_EXPORT void *_Nonnull __CFSafelyReallocate(void * _Nullable destination, size
 CF_EXPORT void *_Nonnull __CFSafelyReallocateWithAllocator(CFAllocatorRef _Nullable, void * _Nullable destination, size_t newCapacity, CFOptionFlags options, void (^_Nullable reallocationFailureHandler)(void *_Nonnull original, bool *_Nonnull outRecovered));
 #endif
 
+_CF_EXPORT_SCOPE_BEGIN
 Boolean __CFAllocatorRespectsHintZeroWhenAllocating(CFAllocatorRef _Nullable allocator);
+_CF_EXPORT_SCOPE_END
```
parkera, line 73: Will making this CF_EXPORT like the others also fix the issue?

egorzhdan: Good question! `CF_EXPORT` seems to expand into `extern` on Linux, not `extern "C"`:

https://github.com/apple/swift-corelibs-foundation/blob/a9ffa779b1fdaa767c9c6ea28acfca32f7efe460/Sources/CoreFoundation/include/CFBase.h#L139

I'm wondering if this is an oversight, I can tweak the `CF_EXPORT` macro to expand into `extern "C"` in C++ language mode, and then use `CF_EXPORT` for this function.

parkera: In truth no one is supposed to be using CoreFoundation except Foundation.

parkera: Usually our CF headers start with the macro the begin extern “C” but since this one is internal to Foundation it doesn’t have it.

egorzhdan: I think this header file is included by a public Foundation header, so it ends up being parsed when someone is including Foundation.

There are a few `_CF_EXPORT_SCOPE_BEGIN`/`_CF_EXPORT_SCOPE_END` blocks below in this file, so I thought that this is the preferred solution. But I'm happy to switch to `CF_EXPORT` if you think that's better.

parkera: To avoid this turning into a tar pit of changes, let's just extend the begin/end to cover this 'group' of functions. Lines 60 - 75

egorzhdan: Sounds good, done.

egorzhdan:
@swift-ci please test

egorzhdan:
@swift-ci please test Windows

parkera (approved):

PR #5083, opened 2024-09-08
Tests: prefer `XCTAssertEqual` over `XCTAssertTrue`

compnerd:
Use the `XCTAssertEqual` check for the equality over an inline equal check. This allows the message to describe both sides of the comparison and helps identify the issue more easily.


compnerd:
@swift-ci please test

PR #5084, opened 2024-09-19
CMake: Fix accidental variable expansion of `WASI`

kateinoigakukun:
CMake recently introduced a new variable `WASI` to check if the target platform is WASI (https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9659). However, the change led to `WASI` being expanded as a variable, which is not what we want in checking the platform name.

To have compatibility with older and newer versions of CMake, we should quote the string `WASI` to prevent it from being expanded as a variable.


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

PR #5090, opened 2024-09-28
Restore thread-unsafe fallback for setting the Process working directory

jakepetroules:
Swift still needs to support Amazon Linux 2 until it EoLs in mid-2025. So restore the thread-unsafe fallback for systems with glibc older than version 2.29, which was removed in #4981.

PR #5095, opened 2024-10-02
[6.0.2] Add a thread-safe implementation of Process.currentDirectoryURL

jakepetroules:
posix_spawn_file_actions_addchdir_np is the official way to set the working directory of a spawned process and is supported on both macOS and glibc (Linux, etc.). This makes Process.currentDirectoryURL thread-safe, as the current approach will result in the working directory being nondeterministically assigned when spawning processes across multiple threads, using different working directories.

---

This already landed in main via #4981 (and amendment in #5090) and it's pretty disruptive for us so I'd like to get it into 6.0.2 if possible.

### DIFF
--- a/Sources/CoreFoundation/CFPlatform.c
+++ b/Sources/CoreFoundation/CFPlatform.c
@@ -57,6 +57,10 @@
 #define kCFPlatformInterfaceStringEncoding	CFStringGetSystemEncoding()
 #endif
 
+#ifndef __GLIBC_PREREQ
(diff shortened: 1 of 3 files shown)


jakepetroules:
@swift-ci please test linux

jakepetroules:
@swift-ci please test windows

parkera (approved):

jakepetroules:
Also opened #5096 for release/6.0 to ensure this makes it into 6.0.3+

PR #5107, opened 2024-10-14
Fix a leak when mutating CharacterSet by moving away from _SwiftNSCharacterSet wrapper

parkera:
`CharacterSet` was the one remaining type in swift-corelibs-foundation using a wrapper type to handle moving between immutable/mutable and handling uniqueness checks. By removing the wrapper and simplifying the usage, we can also remove the leak from mismanaging an "Unmanaged" CF type reference.

rdar://137806932


parkera:
@swift-ci test

itingliu (approved):
Did you know if the boxing was introduced to address some issues, or was that preventive?

itingliu:
Also we probably want this for 6.0.2 as well

parkera:
> Did you know if the boxing was introduced to address some issues, or was that preventive?

It was an abstraction for the several types that used this pattern. But this is the only one left, so it has outlived its usefulness. The Unmanaged part of it was an attempt to avoid accidentally increasing the retain count so copy-on-write worked correctly.

PR #5115, opened 2024-10-17
Fix infinite loop when force casting NSError to CocoaError

nnabeyang:
This PR addresses an infinite loop issue that occurred when force casting `NSError` to `CocoaError`. The problem arose because `_forceBridgeFromObjectiveC` was calling `_unconditionallyBridgeFromObjectiveC`, and in turn, `_unconditionallyBridgeFromObjectiveC` was calling `_forceBridgeFromObjectiveC`, causing a recursive loop.


parkera:
@swift-ci test

parkera:
Thanks for this fix. We are strangely inconsistent about which bridging function is 'primary' for these types; sometimes it funnels to conditionally, sometimes unconditionally, sometimes none of the above.

parkera (approved):

PR #5118, opened 2024-10-22
Ensure that NSCharacterSet always returns an NSObject on copy

parkera:
Follow up to #5107, fixes an issue where an assert in the stdlib cast was failing in assertions-enabled mode.

rdar://138005684


parkera:
@swift-ci test

jmschonfeld (approved):

itingliu:
In case we forget: We'd want to cherry pick to 6.0 and 6.0.2 as well

PR #5132, opened 2024-11-04
Improve CustomNSError.errorDomain calculation

ChrisBenua:
Optimized `CustomNSError.errorDomain` calculation.

`String(reflecting:)` for `Any.Type` objects eventually calls `_type(:qualified:)`. But `String(reflecting:)` call has overhead because of `as?` checks. 

All details are listed in [issue](https://github.com/swiftlang/swift-foundation/issues/1031).

### DIFF
--- a/Sources/Foundation/NSError.swift
+++ b/Sources/Foundation/NSError.swift
@@ -300,7 +300,7 @@ public protocol CustomNSError : Error {
 public extension CustomNSError {
     /// Default domain of the error.
     static var errorDomain: String {
-        return String(reflecting: self)
+        return _typeName(self, qualified: true)
     }
 
     /// The error code within the given domain.


ChrisBenua:
@swift-ci please test

parkera:
This patch will help for non-Darwin platforms - the Foundation version is [here](https://github.com/swiftlang/swift-foundation/blob/b49bbc637a95787492ed8099ac10af30bbdfd463/Sources/FoundationEssentials/Error/CocoaError.swift#L232).

parkera:
@swift-ci test

Sources/Foundation/NSError.swift
```diff
@@ -300,7 +300,7 @@ public protocol CustomNSError : Error {
 public extension CustomNSError {
     /// Default domain of the error.
     static var errorDomain: String {
-        return String(reflecting: self)
+        return _typeName(type(of: self), qualified: true)
```
parkera, line 303: @jmschonfeld you've used this function for `Predicate` - does it seem right here too?

ChrisBenua:
Here is the same MR for Darwin platforms:

https://github.com/swiftlang/swift-foundation/pull/1032

Sources/Foundation/NSError.swift
ChrisBenua: I've got it little bit wrong. Because it is static var, there is no need to call `type(of:)`, it should be `_typeName(self, qualified: true)`
Screenshot 2024-11-04 at 21 14 25

jmschonfeld: Yeah `_typeName(self)` is a fine way to get the fully qualified name here - just to confirm, @Azoy is it always the case that `String(reflecting: someType)` will always just return the fully qualified type name here or do you know of any types that might somehow produce a different value via `String(reflecting:)` here (just to make sure we don't accidentally change behavior)?

jmschonfeld:
@swift-ci test

jmschonfeld:

parkera (approved):
One more thing - can we verify we have a unit test for this behavior?

ChrisBenua:
> One more thing - can we verify we have a unit test for this behavior?

Sure, here it is [TestNSError.swift](https://github.com/swiftlang/swift-corelibs-foundation/blob/8c0a11ee6f0e6a7872d765d6be2e252816123c6a/Tests/Foundation/TestNSError.swift#L40)

jmschonfeld:
It looks like the linux test failed in a SwiftPM test with

```
Test Case 'PackageCommandTests.testCommandPluginDiagnostics' started at 2024-11-04 20:21:39.518
/home/build-user/swiftpm/Tests/CommandsTests/PackageCommandTests.swift:2433: error: PackageCommandTests.testCommandPluginDiagnostics : XCTAssertTrue failed - unexpected failure matching '' against pattern equal("command plugin: print\n")
Test Case 'PackageCommandTests.testCommandPluginDiagnostics' failed (55.424 seconds)
```

It's not immediately apparent how that could be related to this change. @ChrisBenua any thoughts, or should we try a re-run to see if it was an intermittent issue

jmschonfeld:
Windows passed and this isn't platform specific so I'll just rerun Linux and see what happens 🙂

jmschonfeld:

ChrisBenua:
> It looks like the linux test failed in a SwiftPM test with
> 
> ```
> Test Case 'PackageCommandTests.testCommandPluginDiagnostics' started at 2024-11-04 20:21:39.518
> /home/build-user/swiftpm/Tests/CommandsTests/PackageCommandTests.swift:2433: error: PackageCommandTests.testCommandPluginDiagnostics : XCTAssertTrue failed - unexpected failure matching '' against pattern equal("command plugin: print\n")
> Test Case 'PackageCommandTests.testCommandPluginDiagnostics' failed (55.424 seconds)
> ```
> 
> It's not immediately apparent how that could be related to this change. @ChrisBenua any thoughts, or should we try a re-run to see if it was an intermittent issue

It's unclear to me how my changes can possibly affect SPM. Linux tests passed after re-run, so most likely it was intermittent issue

ChrisBenua:
@jmschonfeld, I was wondering if you could please ask someone from your colleagues to take a look at similar PR: https://github.com/swiftlang/swift/pull/77369? 

It would be greatly appreciated. Thank you in advance for your consideration.

PR #5136, opened 2024-11-19
[Windows] FileManager.enumerator(at:) default error handler should continue rather than exiting

jmschonfeld:
On Darwin + Linux, the default error handler for `FileManager.enumerator(at:)` continues on error (and this is the documented behavior as well) however on Windows the default error handler stops enumeration at an error. This updates the Windows behavior to match the Darwin + Linux behavior.


compnerd (approved):

jmschonfeld:
@swift-ci please test

PR #5147, opened 2024-12-11
int i should match count, which is a CFIndex

AreaZR:


compnerd (commented):
By that token, the type should be `size_t` as `sizeof(...) / sizeof(...)` is a `size_t` as `sizeof` returns `size_t`. I'm not sure that there is a good justification here per se. Why not convert the `int` to `CFIndex` at the use site?

AreaZR:
> By that token, the type should be `size_t` as `sizeof(...) / sizeof(...)` is a `size_t` as `sizeof` returns `size_t`. I'm not sure that there is a good justification here per se. Why not convert the `int` to `CFIndex` at the use site?

Because it fits in a CFIndex and long means more possible optimizations due to unsigned being defined overflow behavior even if it doesn't overflow (thanks LLVM)

compnerd:
Are all of these paths currently exercised? The current migration from swift-corelibs-foundation to swift-foundation means that much of CoreFoundation is unused.

parkera (approved):

parkera:
@swift-ci test

parkera:
I've been trying to delete CF code that is truly unused as we go. Even if I missed some, it seems fine to take this patch to me.

PR #5148, opened 2024-12-15
Add public initializer to HTTPCookieStorage

agisboye:
The Foundation [docs](https://developer.apple.com/documentation/foundation/httpcookiestorage) state that HTTPCookieStorage can be subclassed. This, however, is impossible to do with the [swift-corelibs-foundation](https://github.com/swiftlang/swift-corelibs-foundation) version since there are no publicly accessible initializers.

subclass

```swift
class MyCookieStorage: HTTPCookieStorage {
    init(someParameter: String) {
        super.init() // Accessible on Darwin but not on other platforms
    }
}
```

This PR adds an initializer. It has no parameters similar to the Darwin version.
From what I can gather, when a `HTTPCookieStorage` is initialized, it's completely separate from the shared storage and only lives as long as there is a reference to it (i.e. cookies are not persisted).

### DIFF
--- a/Sources/FoundationNetworking/HTTPCookieStorage.swift
+++ b/Sources/FoundationNetworking/HTTPCookieStorage.swift
@@ -67,6 +67,13 @@ open class HTTPCookieStorage: NSObject, @unchecked Sendable {
     private let syncQ = DispatchQueue(label: "org.swift.HTTPCookieStorage.syncQ")
 
     private let isEphemeral: Bool
+    
+    public init() {
+        _allCookies = [:]
+        cookieAcceptPolicy = .always
+        isEphemeral = true
+        super.init()
+    }
 
     private init(cookieStorageName: String, isEphemeral: Bool = false) {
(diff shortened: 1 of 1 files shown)


guoye-zhang:
Looks OK to me. `-init` is inherited from NSObject on Darwin I believe.

@zhenchaoli Does `HTTPCookieStorage()` also create an in-memory cookie storage on Darwin?

zhenchaoli:
Seems that on Darwin HTTPCookieStorage() does not have internal backing store, so setting/retrieving cookies are no-ops.

agisboye:
@guoye-zhang, @zhenchaoli I've changed the behavior such that setting cookies is a no-op (making all the other operations no-ops as well).

guoye-zhang (approved):

guoye-zhang:
@swift-ci please test

guoye-zhang:
@swift-ci please test

parkera:
@swift test windows

PR #5149, opened 2024-12-17
[android] Use Bionic imports instead where possible

finagolfin:
@compnerd asked that we [use this narrower import instead where possible](https://github.com/swiftlang/swift-package-manager/pull/7755#discussion_r1665997808), so this pull does so and was tested natively on Android.

### DIFF
--- a/Sources/xdgTestHelper/main.swift
+++ b/Sources/xdgTestHelper/main.swift
@@ -19,7 +19,7 @@ import FoundationNetworking
 #endif
 #if os(Windows)
 import WinSDK
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 #endif
 
--- a/Sources/Foundation/FileManager.swift
+++ b/Sources/Foundation/FileManager.swift
@@ -21,8 +21,8 @@ import WinSDK
 
 #if os(WASI)
 import WASILibc
-#elseif canImport(Android)
-import Android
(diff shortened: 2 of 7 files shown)


finagolfin:
@swift-ci please test

Sources/xdgTestHelper/main.swift
```diff
@@ -19,7 +19,7 @@ import FoundationNetworking
 #endif
 #if os(Windows)
 import WinSDK
-#elseif os(Android)
+#elseif canImport(Android)
```
compnerd, line 22: I think that this should still be an `os` check.

compnerd (commented):
The switch from Android -> Bionic looks good to me!

compnerd (approved):

Sources/xdgTestHelper/main.swift
finagolfin: The two are pretty much equivalent, but I use this to guard imports like this and `os(Android)` for genuine OS differences that don't involve importing these C functions. I have seen this done for other platforms in the Swift codebase, so I'm following that distinction for Android too.

compnerd: This is only done for Darwin platforms because there is a large set of them and we don't have a way to uniformly identify them at once.

PR #5150, opened 2024-12-17
Revert "[Linux] Enable build-ids. (#4995)"

compnerd:
This reverts commit f186ff8801cf2ec5e5e530d08207dd17375e1cbb. This is incorrectly being applied to Windows builds as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,8 +98,6 @@ FetchContent_MakeAvailable(SwiftFoundationICU SwiftFoundation)
 include(CheckLinkerFlag)
 include(CheckSymbolExists)
 
-check_linker_flag(C "LINKER:--build-id=sha1" LINKER_SUPPORTS_BUILD_ID)
-
 # Detect if the system libc defines symbols for these functions.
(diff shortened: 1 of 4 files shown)


compnerd:
@swift-ci please test

parkera:
Do we just need to conditionalize this more instead of reverting it?

compnerd:
I'd rather avoid the warning and then re-enable this once we have proper testing. I'm worried that we might need CMake changes.

compnerd:

al45tair:
Why is this a problem on Windows? Surely the `check_linker_flag` prevents the extra argument from being passed to the linker there?

compnerd:
It doesn't seem to be preventing that from being passed to the invocation.

al45tair:
Hmmm. In that case, what we should probably do is conditionalise the `check_linker_flag` test so it doesn't happen on Windows, in which case we then won't pass the build ID argument through. I'm guessing `link.exe` is treating the argument as a filename, which presumably means that the logic in CMake's `check_linker_flag` is too simplistic to catch this.

PR #5156, opened 2025-01-15
[6.1] Fix race condition in __CFStringGetEightBitStringEncoding

jmschonfeld:
- **Explanation**: Fixes a small behavior issue caused by a race condition in `__CFStringGetEightBitStringEncoding`
    
  - **Scope**: Impacts many of the low-level `CFString` functions that use CoreFoundation's "eight bit" string encoding
    
  - **Issues**: https://github.com/swiftlang/swift-corelibs-foundation/issues/5152
    
  - **Original PRs**: https://github.com/swiftlang/swift-corelibs-foundation/pull/5155
    
  - **Risk**: Medium - this is a low level function that is called from a lot of places in CoreFoundation, but the change is very small and is not supposed to change behavior here aside from the window of the race condition.
    
  - **Testing**: Local testing at desk
    
  - **Reviewers**: @parkera


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #5163, opened 2025-01-24
Use installed Android cacerts for URLSession

marcprux:
This PR will cause `URLSession` to use the installed Android cacerts folders (`/apex/com.android.conscrypt/cacerts` and `/system/etc/security/cacerts`), similar to how swift-nio supports TLS on Android (https://github.com/apple/swift-nio-ssl/pull/453). If no custom `URLSessionCertificateAuthorityInfoFile` environment variable is set, it will dynamically assemble a certificate file from the contents of those folders, and then instruct curl to use it with the [curl_easy_setopt](https://curl.se/libcurl/c/curl_easy_setopt.html) parameter [CURLOPT_CAINFO](https://curl.se/libcurl/c/CURLOPT_CAINFO.html).

This has the benefits of:

1. https URLs will work out-of-the-box on Android without an app needing to bundle it's own custom certificate file (typically from https://curl.haxx.se/ca/cacert.pem)
2. Android system updates that add and remove certificate authorities will be respected in apps that use Swift


marcprux:
This is ready for review. I've tested it out against the [swift-6.0.3-RELEASE-android-24-0.1.artifactbundle.tar.gz](https://github.com/marcprux/swift-android-sdk/actions/runs/12960699464/artifacts/2484373617) artifact from https://github.com/marcprux/swift-android-sdk/actions/runs/12960699464 and loading an https site works on Android.

Sources/FoundationNetworking/URLSession/libcurl/EasyHandle.swift
```diff
@@ -220,6 +220,66 @@ extension _EasyHandle {
...
+
+            // See https://github.com/apple/swift-nio-ssl/blob/main/Sources/NIOSSL/AndroidCABundle.swift
+            let certsFolders = [
+                "/apex/com.android.conscrypt/cacerts", // >= Android14
+                "/system/etc/security/cacerts" // < Android14
+            ]
+
+            let aggregateCertPath = NSTemporaryDirectory() + "/cacerts-\(UUID().uuidString).pem"
+
+            if FileManager.default.createFile(atPath: aggregateCertPath, contents: nil) == false {
```
parkera, line 235: Does this file have to be cleaned up at some point?

Sources/FoundationNetworking/URLSession/libcurl/EasyHandle.swift
```diff
@@ -220,6 +220,66 @@ extension _EasyHandle {
...
+            }
+
+            guard let fs = FileHandle(forWritingAtPath: aggregateCertPath) else {
+                return
+            }
+
+            // write a header
+            fs.write("""
+            ## Bundle of CA Root Certificates
+            ## Auto-generated on \(Date())
```
parkera, line 246: You may want to give this a specific format instead of just relying on the debug description.

Sources/FoundationNetworking/URLSession/libcurl/EasyHandle.swift
```diff
@@ -220,6 +220,66 @@ extension _EasyHandle {
...
+                    } catch {
+                        // ignore individual errors and soldier on…
+                        continue
+                    }
+                }
+            }
+
+            try! fs.close()
+
+            aggregateCertPath.withCString { pathPtr in
```
guoye-zhang, line 276: `aggregateCertPath` should probably be a URL and you can call `withUnsafeFileSystemRepresentation` here

Sources/FoundationNetworking/URLSession/libcurl/EasyHandle.swift
```diff
@@ -220,6 +220,66 @@ extension _EasyHandle {
...
+            for certsFolder in certsFolders {
+                let certsFolderURL = URL(fileURLWithPath: certsFolder)
+                if (try? certsFolderURL.resourceValues(forKeys: [.isDirectoryKey]).isDirectory) != true { continue }
+                let certURLs = try! FileManager.default.contentsOfDirectory(at: certsFolderURL, includingPropertiesForKeys: [.isRegularFileKey, .isReadableKey])
+                for certURL in certURLs {
+                    // certificate files have names like "53a1b57a.0"
+                    if certURL.pathExtension != "0" { continue }
+                    do {
+                        if try certURL.resourceValues(forKeys: [.isRegularFileKey]).isRegularFile != true { continue }
+                        if try certURL.resourceValues(forKeys: [.isReadableKey]).isReadable != true { continue }
```
jrflat, line 265: It's generally preferred to get these resource values in a single call to avoid multiple filesystem calls. You could do something like

```
guard certURL.pathExtension == "0",
  let values = try? certURL.resourceValues(forKeys: [.isRegularFileKey, .isReadableKey]),
  values.isRegularFile == true,
  values.isReadable == true,
  let data = try? Data(contentsOf: certURL) else {
    continue
}
try? fs.write(contentsOf: data)
```

jrflat: Actually, `Data(contentsOf: certURL)` will fail anyway if the URL isn't a regular file or if it isn't readable, so this can be consolidated to a single `let data = try? Data(contentsOf: certURL)`

marcprux: Good idea. I've updated it in bda9bd4

Sources/FoundationNetworking/URLSession/libcurl/EasyHandle.swift
marcprux: Good question. I'm not sure exactly *when* libcurl will read this file, and so we don't know when we can delete it. In the case of an Android app, `NSTemporaryDirectory()` will be the app's `caches` folder, and so will eventually be automatic cleanup of this folder, but still leaving stay files around like this is not good form.

I think the best solution would be rather than using [CURLOPT_CAINFO](https://curl.se/libcurl/c/CURLOPT_CAINFO.html) (as exposed through `CFURLSessionOptionCAINFO`), we would instead use [CURLOPT_CAINFO_BLOB](https://curl.se/libcurl/c/CURLOPT_CAINFO_BLOB.html), which would mean we wouldn't need to create any temporary file at all. But not only does this not currently have any equivalent `CFURLSessionOption*` property, it would also require bringing in the `curl_blob` struct from libcurl, which seemed a bit much just for this Android feature. OTOH, I don't see any other way to avoid the issue of leaving behind old aggregate certificate files.

If you think it is worth investigating, I can look into it…

parkera:
@swift-ci test

marcprux:
Closing in favor of https://github.com/swiftlang/swift-corelibs-foundation/pull/5283

PR #5166, opened 2025-02-07
[DON'T MERGE] Temporarily set CMake policy CMP0157 to OLD

weliveindetail:
We need the old setting as long as we use Swift's legacy driver on Windows. This is a workaround for cross-compiling the Swift Runtime to Android in https://github.com/swiftlang/swift/pull/79185. It's not supposed to land.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ endif()
 
 if(POLICY CMP0157)
     # New Swift build model: improved incremental build performance and LSP support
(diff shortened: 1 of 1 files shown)


weliveindetail:
This was just a utility change to unblock PR tests for https://github.com/swiftlang/swift/pull/79185. Since https://github.com/swiftlang/swift-corelibs-foundation/pull/5180 landed I don't need it anymore.

PR #5167, opened 2025-02-08
Ensure URLSessionDataTask response is not dropped

### LINKED ISSUE #4052: [SR-6406] Fatal error: Transfer completed, but there's no response.: file Foundation/URLSession/http/HTTPURLProtocol.swift, line 549

harlanhaskins:
This code seems to have been attempting to resolve the completed response at the end of the transfer, but it was creating a local copy and mutating that instead of mutating the transfer state we're currently looking at. This seems to happen when a header is truncated? I'm not exactly sure, but this code looked incorrect on inspection.

Still testing this fix locally.

This should fix #4052


harlanhaskins:
@swift-ci please test

harlanhaskins:
@swift-ci build toolchain linux

harlanhaskins:
@swift-ci please build linux toolchain

PR #5174, opened 2025-02-20
Disable a base64 test while we investigate

parkera:
The new base64 implementation (https://github.com/swiftlang/swift-foundation/pull/1157) caused this SCL-F test to start failing. Disable the test while we decide if the test is faulty or the new implementation.


parkera:
@swift-ci test

PR #5170, opened 2025-02-15
always @preconcurrency import Glibc/Musl

weissi:
Related issues:
- https://github.com/swiftlang/swift-format/issues/933
- https://github.com/swiftlang/swift/issues/79414

I've got a working program

```
@preconcurrency import Glibc
import Foundation

await Task {
  _ = fputs("hello\n", stderr)
}.value
```

but alas, if `swift-format` reformats it, it makes it (note that the two imports are reordered)

```
import Foundation
@preconcurrency import Glibc

await Task {
  _ = fputs("hello\n", stderr)
}.value
```

which unfortunately doesn't compile

```
$ swift format --in-place test.swift && swiftc -swift-version 6 test.swift && echo OK
test.swift:5:24: error: reference to var 'stderr' is not concurrency-safe because it involves shared mutable state
3 | 
4 | await Task {
5 |   _ = fputs("hello\n", stderr)
  |                        `- error: reference to var 'stderr' is not concurrency-safe because it involves shared mutable state
6 | }.value
7 | 

SwiftGlibc.stderr:1:12: note: var declared here
1 | public var stderr: UnsafeMutablePointer<FILE>!
  |            `- note: var declared here
```

the problem is that `Foundation` imports `Glibc`/`Musl` but doesn't mark them `@preconcurrency`.

PR #5177, opened 2025-02-26
always @preconcurrency import Glibc/Musl/Android/Bionic/WASILibc [6.1]

weissi:
Same as #5170 but for 6.1.


weissi:
@swift-ci please test

weissi:
CC @parkera

PR #5178, opened 2025-02-26
Revert "Remove incorrect `guard` check for Android"

hjyamauchi:
Reverts swiftlang/swift-corelibs-foundation#5171

This makes it at least buildable on Android.

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -943,6 +943,13 @@ open class Process: NSObject, @unchecked Sendable {
 #else
         var spawnAttrs: posix_spawnattr_t = posix_spawnattr_t()
 #endif
+#if os(Android)
+        guard var spawnAttrs else {
+            throw NSError(domain: NSPOSIXErrorDomain, code: Int(errno), userInfo: [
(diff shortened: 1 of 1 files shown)


hjyamauchi:
@swift-ci please test

finagolfin (changes requested):
This is clearly incorrect, as explained in detail in my subsequent comments in #5171.

Since you're having problems dealing with the nullability annotations in the latest Android NDKs, which I noted I haven't upgraded these `posix_spawn` APIs to yet in my own Android SDK, what may work is to update to NDK 27 and move this Android `guard` check down a line after the `posix_spawnattr_init()`, to unwrap the optional in Swift _only after_ you know `spawnAttrs` has been initialized. Of course, the only way to test that, if it compiles, is to actually run it too, which it appears you haven't been doing so far.

I suggest you experiment with such ways to fix this code with the latest NDKs, rather than simply sticking with whatever temporary compilation-only hack @hyp put in earlier.

hjyamauchi:
Saleem has this: https://github.com/swiftlang/swift-corelibs-foundation/pull/5179 for moving the guard code

PR #5179, opened 2025-02-27
Process: unwrap the `posix_spawnattr_t` on Android

compnerd:
This is required as while `posix_spawnattr_init` permits a nullable type, `posix_spawnattr_setflags` properly expects a non-null parameter. Unwrap the newly minted spawnattr or abort if the allocation failed.

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -944,6 +944,12 @@ open class Process: NSObject, @unchecked Sendable {
         var spawnAttrs: posix_spawnattr_t = posix_spawnattr_t()
 #endif
         try _throwIfPosixError(posix_spawnattr_init(&spawnAttrs))
+#if os(Android)
+        guard var spawnAttrs else {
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @finagolfin

The revert was incorrect, the unwrapping should always be present. However, the position of the unwrapping was incorrect. Even with the older SDK, the unwrapping should occur _after_ the `posix_spawnattr_init` call. I worked with Google to get the nullability corrected on that call, however, in the mean time, I had overridden the API with APINotes to ensure that we could get the correct code. It seems that the unwrapping was just a bit earlier than it should've been.

This has only been tested with r27c, but I expect it to be backwards compatible with r26b and the Android SDK that we are currently shipping on main.

compnerd:
@swift-ci please test

compnerd:

finagolfin (approved):
Right, this was always my point, that the `guard` made no sense right after setting this to `nil`, whereas this should work.

PR #5180, opened 2025-02-27
Set CMP0157 to OLD for Android

Steelskin:
There is no early swift-driver build for the Windows toolchain. As a result, swift-collections fails to build properly when CMP0157 is set to NEW due to object files not being generated.

This sets CMP0157 to OLD when targetting Android until the early swift-driver is available on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ if(POLICY CMP0156)
 endif()
 
 if(POLICY CMP0157)
-    if(CMAKE_SYSTEM_NAME STREQUAL Android)
+    if(CMAKE_HOST_SYSTEM_NAME STREQUAL Windows AND CMAKE_SYSTEM_NAME STREQUAL Android)
         # CMP0157 causes swift-collections to fail to compile when targetting
         # Android on Windows due to swift-driver not being present during the
         # toolchain build. Disable it for now.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,8 +22,15 @@ if(POLICY CMP0156)
 endif()
 
 if(POLICY CMP0157)
-    # New Swift build model: improved incremental build performance and LSP support
-    cmake_policy(SET CMP0157 NEW)
+    if(CMAKE_SYSTEM_NAME STREQUAL Android)
+        # CMP0157 causes swift-collections to fail to compile when targetting
+        # Android on Windows due to swift-driver not being present during the
+        # toolchain build. Disable it for now.
+        cmake_policy(SET CMP0157 OLD)
+    else()
+        # New Swift build model: improved incremental build performance and LSP support
+        cmake_policy(SET CMP0157 NEW)
+    endif()
 endif()
 
 if (NOT DEFINED CMAKE_C_COMPILER)


weliveindetail:
Facing this issue as well. This sounds like a good compromise. Is it happening on other platforms too? Otherwise we might want to add `CMAKE_HOST_SYSTEM_NAME STREQUAL Windows` to the condition.

Steelskin:
> Facing this issue as well. This sounds like a good compromise. Is it happening on other platforms too? Otherwise we might want to add `CMAKE_HOST_SYSTEM_NAME STREQUAL Windows` to the condition.

It works fine when building for Windows, but the Android build does not take kindly to not having the early swift-driver.

weliveindetail:
What I meant was: The issue doesn't reproduce when running Android build on Linux right?

Steelskin:
> What I meant was: The issue doesn't reproduce when running Android build on Linux right?

Ah sorry, I have no idea. I am assuming there is no early swift-driver on Linux either since that requires a static build.

finagolfin:
I am not that familiar with [this CMake policy, which appears to enable using the new `CMAKE_Swift_COMPILATION_MODE` flag](https://cmake.org/cmake/help/latest/policy/CMP0157.html), looks like it was just added with the Foundation rewrite merged last summer. Looking at the CMake config below it appears that [configuring that new `CMAKE_Swift_COMPILATION_MODE` flag here is disabled on Windows hosts](https://github.com/swiftlang/swift-corelibs-foundation/blob/70ce5131e7564bac8a005b023cd1439c39b23bac/CMakeLists.txt#L45), so I'm not sure how it is affecting you.

Is it that CMake sets `CMAKE_Swift_COMPILATION_MODE` to some default value anyway if `CMP0157` is enabled? If so, you'll want to narrow this pull to `if(POLICY CMP0157 AND CMAKE_HOST_SYSTEM_NAME STREQUAL Windows AND CMAKE_SYSTEM_NAME STREQUAL Android)`, as Stefan says.

> I am assuming there is no early swift-driver on Linux either since that requires a static build.

No, the linux CI builds and uses the early swift-driver, as you can see on [any recent CI run (search for earlyswiftdriver)](https://ci.swift.org/job/swift-PR-Linux/20214/consoleText), and you can always pass in a prebuilt linux toolchain that includes the new swift-driver to build this repo.

My Android CI builds this repo just fine from a linux host using such an official prebuilt linux toolchain.

Btw, what is up with @hyp, will he be back?

Steelskin:
> I am not that familiar with [this CMake policy, which appears to enable using the new `CMAKE_Swift_COMPILATION_MODE` flag](https://cmake.org/cmake/help/latest/policy/CMP0157.html), looks like it was just added with the Foundation rewrite merged last summer. Looking at the CMake config below it appears that [configuring that new `CMAKE_Swift_COMPILATION_MODE` flag here is disabled on Windows hosts](https://github.com/swiftlang/swift-corelibs-foundation/blob/70ce5131e7564bac8a005b023cd1439c39b23bac/CMakeLists.txt#L45), so I'm not sure how it is affecting you.

When `CMP0157` is set to `OLD`, the swiftmodule and static library are built in a single ninja step. The step also expects that the object files are built in the process but they are not consumed by any other target. When `CMP0157` is set to `NEW`, the generated ninja file includes 2 targets: one to build the swiftmodule + object files and one to build the static library.
The problem occurs when swift-driver is not present (as is the case on Windows when building the toolchain). When targeting Android, the object files are not built, resulting in the target to build the static library failing. As a workaround, we are using an older CMake version in [GHA swift-build](https://github.com/compnerd/swift-build/blob/409d1d9e4cba8562776d79b5005809de4268db8a/.github/workflows/swift-toolchain.yml#L2130).

> Is it that CMake sets `CMAKE_Swift_COMPILATION_MODE` to some default value anyway if `CMP0157` is enabled? If so, you'll want to narrow this pull to `if(POLICY CMP0157 AND CMAKE_HOST_SYSTEM_NAME STREQUAL Windows AND CMAKE_SYSTEM_NAME STREQUAL Android)`, as Stefan says.

Thanks, I did that.

> Btw, what is up with @hyp, will he be back?

He has gone to pursue a different field. He might pop up every once in a while, though :)

finagolfin (approved):

finagolfin:
@swift-ci please test

finagolfin:
> He has gone to pursue a different field.

OK, only reason I asked is because Tony and I were still pinging him on recent pulls, we'll stop doing that. 😉

weliveindetail:
The bot checked out the correct commit [4af0f8c](https://github.com/swiftlang/swift-corelibs-foundation/commit/4af0f8c1f4afe982a26376b84cbf376242eb9b0a), but then it reported the result for the outdated commit: 
```
Adding one-line test results to commit status...
Setting status of a563acec95f721223859b2e704843f6ca78611e7 to SUCCESS with url https://ci-external.swift.org/job/swift-corelibs-foundation-PR-windows/618/ and message [...]
```

[a563ace](https://github.com/swiftlang/swift-corelibs-foundation/pull/5180/commits/a563acec95f721223859b2e704843f6ca78611e7) did actually get accepted 🤷‍♂️

weliveindetail:
Let's try that again.

@swift-ci please test Windows

PR #5181, opened 2025-03-02
CoreFoundation: adjust declaration of private SPI

compnerd:
The SPI declaration for `libdispatch_init` was not attribtued properly. The result of this omission was an improper reference to the function on Windows when disaptch is _not_ linked statically. Correct the declaration by re-using the `DISPATCH_EXPORT` macro to decorate the declaration with the appropriate DLLStorage as necessary.

This was caught by running the Foundation tests with SPM with a refactoring of the toolchain layout during the build.


compnerd:
CC: @drexin @jmschonfeld

This gets us one step closer to fixing the test suites on Windows.

compnerd:
@swift-ci please test

PR #5184, opened 2025-03-04
CoreFoundation: Correctly construct user preferences directory

2xsaiko:
Putting an absolute path here makes the home directory get ignored and the final path ends up as /Library/Preferences instead of ~/Library/Preferences.


compnerd:
Technically, this change is fine. However,I wonder if it is currently required  or not. Where is this path exercised? It is slightly confusing as FileManager has migrated to swift-foundation and has an implementation which doesn’t include CoreFoundation.

CC: @parkera

2xsaiko:
UserDefaults class uses this, unless that has changed since I wrote this.

compnerd (commented):
Ah, okay, UserDefaults coming through here makes sense.

compnerd:
@swift-ci please test

AZero13:
Any update on this?

PR #5189, opened 2025-03-17
[FreeBSD] lookup process path via sysctl

michael-yuji:
Currently executable importing Foundation may crash when launched with certain shell as a platform specific way to get the path of the process's executable is not defined. This PR implements this missing piece to fetch the process path via `sysctl`

### DIFF
--- a/Sources/CoreFoundation/CFPlatform.c
+++ b/Sources/CoreFoundation/CFPlatform.c
@@ -136,6 +136,10 @@ static inline void _CFSetProgramNameFromPath(const char *path) {
 #include <sys/exec.h>
 #endif
 
+#if TARGET_OS_BSD && defined(__FreeBSD__)
+#include <sys/sysctl.h>
+#endif
+
 const char *_CFProcessPath(void) {
     if (__CFProcessPath) return __CFProcessPath;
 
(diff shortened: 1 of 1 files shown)


Sources/CoreFoundation/CFPlatform.c
```diff
@@ -230,6 +234,29 @@ const char *_CFProcessPath(void) {
...
+    int sysctl_ret = 0;
+    size_t len = PATH_MAX + 1;
+    argv0 = calloc(len, 1);
+
+    sysctl_ret = sysctl(mib, 4, argv0, &len, NULL, 0);
+
+    // in case for whatever reason the path is > PATH_MAX
+    if (sysctl_ret == -1 && errno == ENOMEM) {
+        // get size needed
+        sysctl_ret = sysctl(mib, 4, NULL, &len, NULL, 0);
```
compnerd, line 249: Any reason to not do this unconditionally to compute the required length?

michael-yuji: Since we should almost never the case when exec  path is > PATH_MAX, so we can save an extra syscall in most cases. (In fact, I'm not sure if it's even possible on generic FreeBSD, but there are many FreeBSD variants floating around...)

grynspan:
Swift Testing has an implementation of this that may be useful. Given that both Swift Testing and Foundation need this information, it may make sense to add it to the stdlib.

parkera (approved):

Issue #3885, opened 2017-02-22
[SR-4036] Memory leak with String.cString(using: Encoding)


pushkarnk:
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-4036      |
   |Radar             | rdar://problem/73153090         |
   |Original Reporter | @pushkarnk      |
   |Type              | Bug    |

Attachment: [Download](https://user-images.githubusercontent.com/2727770/164961918-0bfa0d13-b0c3-42a3-bec3-2f50090e6718.gz)

      
   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 1         |
   |Component/s       | Foundation    |
   |Labels            | Bug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: 342a1be011db14a8a68d02d0d35c641e

  </details>

**Issue Description:**

This code leaks memory

``` java
for i in 1...100000 {
    let s: String = "test\(i)"
    let c = s.cString(using: .utf8)
    print(c)
}
```

parkera:
I think we should probably deprecate this API rather than allow it to persist with a known leak.

PR #5190, opened 2025-03-17
Deprecate cString on non-Darwin (#3885)

parkera:
The extension on `String` to provide a C string works well enough on Darwin, when used with an `autoreleasepool { }`. However, on other platforms where we have no autorelease pools, it simply leaks. Even though this API is useful on Darwin, we should deprecate it in swift-corelibs-foundation because it has repeatedly been the source of some confusing behavior.

https://github.com/swiftlang/swift-corelibs-foundation/issues/3885

### DIFF
--- a/Sources/Foundation/NSStringAPI.swift
+++ b/Sources/Foundation/NSStringAPI.swift
@@ -632,6 +632,7 @@ extension StringProtocol {
 
     /// Returns a representation of the string as a C string
     /// using a given encoding.
+    @available(*, deprecated, message: "On platforms without Objective-C autorelease pools, use withCString instead")
     public func cString(using encoding: String.Encoding) -> [CChar]? {
         return withExtendedLifetime(_ns) {
             (s: NSString) -> [CChar]? in
--- a/Sources/Foundation/NSStringAPI.swift
+++ b/Sources/Foundation/NSStringAPI.swift
@@ -632,7 +632,7 @@ extension StringProtocol {
 
     /// Returns a representation of the string as a C string
     /// using a given encoding.
-    @available(*, deprecated, message: "On platforms without Objective-C autorelease pools, use withCString instead")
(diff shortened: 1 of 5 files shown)


Sources/Foundation/NSStringAPI.swift
```diff
@@ -632,6 +632,7 @@ extension StringProtocol {
 
     /// Returns a representation of the string as a C string
     /// using a given encoding.
+    @available(*, deprecated, message: "On platforms without Objective-C autorelease pools, use withCString instead")
```
DougGregor, line 635: There's an `NSString.cString(using:)` that has the same issue, FWIW

compnerd: I think that we should refer people to use `String.withCString(encodedAs:_)` instead of this message.

parkera: Sure. There is one place we use this API in keyed archiver that will have to get a follow-up.

parkera:
@swift-ci test

parkera:
@swift-ci test

Sources/Foundation/NSStringAPI.swift
```diff
@@ -632,6 +632,7 @@ extension StringProtocol {
 
     /// Returns a representation of the string as a C string
     /// using a given encoding.
+    @available(*, deprecated, message: "On platforms without Objective-C autorelease pools, use withCString(encodedAs:_) instead")
```
grynspan, line 635: This one is actually safe to use though, right?

parkera: Not unless we reimplement it to use a lifetime-limited version of the `NSString` `cString(using:)` function. Which I suppose we could do.

grynspan: I don't know if it's worth it, but it might mean less churn.

parkera:
@swift-ci test

jmschonfeld (approved):

PR #5196, opened 2025-03-30
Add -z,origin linker flag on OpenBSD.

3405691582:
This is required for $ORIGIN rpath processing here (without having to fiddle with workarounds like LD_LIBRARY_PATH).

### DIFF
--- a/Sources/Foundation/CMakeLists.txt
+++ b/Sources/Foundation/CMakeLists.txt
@@ -164,10 +164,6 @@ if(NOT BUILD_SHARED_LIBS)
         "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend swiftSynchronization>")
 endif()
 
-set_target_properties(Foundation PROPERTIES
-    INSTALL_RPATH "$ORIGIN"
-    INSTALL_REMOVE_ENVIRONMENT_RPATH ON)
-
 if(dispatch_FOUND)
     set_target_properties(Foundation PROPERTIES
         BUILD_RPATH "$<TARGET_FILE_DIR:swiftDispatch>")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,6 +212,12 @@ else()
         "/clang:-fcf-runtime-abi=swift")
 endif()
 
+set(CMAKE_INSTALL_REMOVE_ENVIRONMENT_RPATH ON)
+set(CMAKE_INSTALL_RPATH "$ORIGIN")
+if(CMAKE_SYSTEM_NAME MATCHES "OpenBSD|DragonFlyBSD")
+    add_link_options("LINKER:-z,origin")
+endif()
+
 if(CMAKE_BUILD_TYPE STREQUAL Debug)
     list(APPEND _Foundation_common_build_flags
         "-DDEBUG")
--- a/Sources/FoundationNetworking/CMakeLists.txt
+++ b/Sources/FoundationNetworking/CMakeLists.txt
@@ -73,10 +73,6 @@ if(NOT BUILD_SHARED_LIBS)
 
 endif()
 
-set_target_properties(FoundationNetworking PROPERTIES
-    INSTALL_RPATH "$ORIGIN"
-    INSTALL_REMOVE_ENVIRONMENT_RPATH ON)
-
 if(LINKER_SUPPORTS_BUILD_ID)
   target_link_options(FoundationNetworking PRIVATE "LINKER:--build-id=sha1")
 endif()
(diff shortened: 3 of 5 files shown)


Sources/Foundation/CMakeLists.txt
```diff
@@ -168,6 +168,10 @@ set_target_properties(Foundation PROPERTIES
     INSTALL_RPATH "$ORIGIN"
     INSTALL_REMOVE_ENVIRONMENT_RPATH ON)
 
+if(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
+  target_link_options(Foundation PRIVATE "LINKER:-z,origin")
```
jmschonfeld, line 172: All of the targets in the Foundation project (both those here as well as those in swift-foundation) use `ORIGIN`-based rpath values. Is there something special about the Foundation library, or should this be provided for all of these targets?

3405691582: On OpenBSD, -z origin is also required to properly resolve $ORIGIN.

I'm fairly sure this will be required for the other targets, and only the main Foundation library has posed a problem while trying to get the toolchain stood up end to end, so I've only added it for Foundation so far. 

It probably is the case I should add them for the other targets, but I need to double-check.

compnerd: Ah, right, this is Open, not Free. Open and DragonFly both need `-z origin`.

3405691582:
@swift-ci please test.

finagolfin (approved):
LGTM, especially since confined to OpenBSD.

Sources/Foundation/CMakeLists.txt
3405691582: I think I forgot to close the loop here and mention that I added the flag for the other targets.

compnerd: I think that is set as a global isn't it? `CMAKE_INSTALL_RPATH` controls exactly that.

3405691582: I mean the INSTALL_RPATH here.

The alternative is I go and figure out how to add this to the driver.

compnerd: I don't think that this belongs in the driver. Does clang implicitly add `-z origin` when using `$ORIGIN`? Doing what you recommend seems reasonable:

```cmake
if(CMAKE_SYSTEM_NAME MATCHES "OpenBSD|DragonFlyBSD")
  if(CMAKE_INSTALL_RPATH MATCHES "\\$ORIGIN")
    add_link_options("LINKER:-z,origin")
  endif()
endif()
```

3405691582: Done -- thanks for the suggestion.

3405691582:
@swift-ci please test.

Sources/Foundation/CMakeLists.txt
compnerd: Yes, it would be evaluated once. It _could_ be that my regex is incorrect. But, `CMAKE_INSTALL_RPATH` is user supplied.

3405691582: My point is that I am guessing that cmake is evaluating CMAKE_INSTALL_RPATH at the time this expression is evaluated -- before the INSTALL_RPATH target property is evaluated on the Foundation target. If CMAKE_INSTALL_RPATH doesn't _already_ mention $ORIGIN, this won't add the linker flag globally.

If we were going to apply the linker flag to all targets and only all targets at the top level, we'd need to iterate all targets and get the INSTALL_RPATH target property and add the linker flag for only those targets whose INSTALL_RPATH property mentions $ORIGIN -- but that's a whole lot of messing around.

Hence, it might just be simpler to just add the flag unconditionally for now and be done with it. It's not ideal but it'll work, and the scope of the diff will be smaller.

compnerd: Ah, I see your confusion now. `CMAKE_INSTALL_RPATH` is not a _target_ specific property. It is a _default value_ property. The user specifies that and the value is used to initialise the property on all targets. That is:

```
cmake -B out -G Ninja -S . -D CMAKE_INSTALL_RPATH="$ORIGIN" -D CMAKE_SYSTEM_NAME=OpenBSD -D CMAKE_SYSTEM_PROCESSOR=AMD64
```

compnerd:

Sources/Foundation/CMakeLists.txt
3405691582: The point I was missing was forgetting to bring the INSTALL_RPATH property on each target up to the top level and actually set that default. D'oh! I've done that now. plutil still needs a target-specific INSTALL_RPATH, but it doesn't hurt to just override the default for that specific target.

I think this should all makes sense now.

3405691582:
@swift-ci please test.

3405691582:

jmschonfeld (approved):

3405691582:
Please merge on my behalf, thank you!

compnerd (approved):
Thanks, this looks good to me.

PR #5204, opened 2025-04-22
Avoid implicitly linking RegexBuilder symbol

jmschonfeld:
Recently, `FoundationEssentials` began importing the `RegexBuilder` module. While it's an `internal` import, since we don't have the `MemberImportVisibility` feature flag enabled, this import affected overload resolution of this call to `split`. Eventually, we should enable that feature flag to avoid these types of issues, but for now we make the argument explicitly `Character` to pick the stdlib overload instead of the `RegexBuilder` overload to fix linking failures.


jmschonfeld:
@swift-ci please test

tshortli (approved):

PR #5206, opened 2025-05-05
Add UserDefaults.setStandardUserDefaults for Android

marcprux:
This PR adds `UserDefaults.setStandardUserDefaults(UserDefaults)` to provide the ability for a platform to override the standard defaults.

The existing Linux-specific defaults system doesn't work well on Android, whose "defaults" system are handled by the [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) Java class in the Android SDK. Rather than adding a JNI bridge here just for an Android-specific implementation, this PR takes the simpler route of just exposing the ability to override the standard user defaults with a public setter, which can be called at initialization time.

### DIFF
--- a/Sources/Foundation/UserDefaults.swift
+++ b/Sources/Foundation/UserDefaults.swift
@@ -90,7 +90,14 @@ open class UserDefaults: NSObject {
     }
     
     open class func resetStandardUserDefaults() {}
-    
+
+    #if os(Android)
+    /// Provide a custom implementation of `UserDefaults.standard` for platform integration.
+    open class func setStandardUserDefaults(_ defaults: UserDefaults) {
+        sharedDefaults = defaults
+    }
+    #endif
+
     public convenience override init() {
         self.init(suiteName: nil)!
     }


compnerd:
This is adding new API, so I think that this might need a Foundation evolution proposal.

parkera:
Is this just emblematic of some larger issue about compatibility of UserDefaults with Android in general?

I don't know as much about it as I would like, so I'm unclear on how this API would be used, who is supposed to call it and when, and other details. In general we try very hard not to expose a platform-specific workaround if we can avoid it.

parkera:
In the long run, do we want Foundation to use the swift-java bridge to just call into [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) directly?

jakepetroules:
> adding a JNI bridge here just for an Android-specific implementation

> In the long run, do we want Foundation to use the swift-java bridge to just call into [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) directly?

Even though it's more difficult to call Java from Swift than it is to call C and Objective-C, I think we should still do that work to provide an ideal API surface for users.

As an example of what other cross platform frameworks are doing, I recall we did lots of JNI in Qt (written in C++) to back various APIs with the appropriate Android platform equivalent.

PR #5207, opened 2025-05-06
Instead of crashing with unknown types in print, use description

parkera:
On Darwin, `plutil -p` can print keyed archiver archives. They contain `UUID`, which is a not a property list type, so `plutil` special cases them for ease of use.

Instead of throwing when we hit an unknown type, simply print a description - which is basically what the Objective-C implementation used to do.


parkera:
@swift-ci test

parkera:
@swift-ci test

parkera:
@swift-ci test

parkera:
@swift-ci test

PR #5210, opened 2025-05-09
[android] Support building `_CFPosixSpawnFileActionsChdir()` for older Android APIs too

finagolfin:
@jakepetroules, following on the similar swiftlang/swift-llbuild#985, this gets Foundation building with older Android APIs too, which I've had to [patch otherwise on my Android CI](https://github.com/finagolfin/swift-android-sdk/blob/43b07babb7e27bf48dc5afe0b9be56398fe22805/swift-android.patch#L101) that builds against API 24.

### DIFF
--- a/Sources/CoreFoundation/CFPlatform.c
+++ b/Sources/CoreFoundation/CFPlatform.c
@@ -1948,7 +1948,7 @@ CF_CROSS_PLATFORM_EXPORT void *_CFReallocf(void *ptr, size_t size) {
 #endif
 }
 
-#if TARGET_OS_ANDROID
+#if TARGET_OS_ANDROID && __ANDROID_API__ < 28
 
 #include <dlfcn.h>
 #include <spawn.h>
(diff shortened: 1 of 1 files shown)


finagolfin:
@swift-ci test

finagolfin:
Updated this to actually build when the Android API is greater than 27 and added a default `_CFPosixSpawnFileActionsChdir()` for the previous callback scheme.

@marcprux and @compnerd, I've only tested the old callback scheme on my Android CI, as it builds against API 24, so if you two are building against API 28 or later, please build this and let me know what you think.

parkera (approved):

finagolfin:
@swift-ci test

finagolfin:
I just built and tested this natively on Android against API 28, got the same test results.

parkera:
@swift-ci test

finagolfin:
@swift-ci test linux

PR #5215, opened 2025-05-22
build: adjust the autolink library for Windows

compnerd:
When building statically, ensure that we prefix the Swift libraries with the correct prefix to permit static linking of the libraries.

PR #5219, opened 2025-05-29
Build: Adjust linking for the static Windows build

Steelskin:
When linking against a static `swiftSynchronization`, the proper prefix needs to be specified.


compnerd:
Duplicate of: #5215

PR #5228, opened 2025-06-28
build with 64-bit time_t on 32-bit platforms

lhoward:
It is good practice to build with 64-bit time_t/timeval on 32-bit platforms to avoid the Y2038 issue. This is the default when building on Yocto for armv7, for example. Unfortunately suseconds_t is not an alias to a type of the correct width (unlike time_t), so for Glibc make it a private alias of time_t to fix the build.


lhoward:
If for process reasons you need a PR against main, can do, just lmk once this is reviewed. Sorry to do it back to front.

PR #5241, opened 2025-07-15
[Windows] Enable `TestURLSession.test_repeatedRequestsStress`

lxbndr:
https://github.com/swiftlang/swift-corelibs-libdispatch/pull/888 fixes what made `TestURLSession.test_repeatedRequestsStress` unstable (actually, the test was initially designed specifically for testing conditions described in https://github.com/swiftlang/swift-corelibs-libdispatch/issues/887). When https://github.com/swiftlang/swift-corelibs-libdispatch/pull/888 lands, we might want to enable `test_repeatedRequestsStress` and see is it still fails.

I am not sure if we want to merge this in the end, as the test could be quite slow on CI (especially on VM cloud runners), but if it IS stable, I would change the XCTSkip reason at least.


compnerd:
https://github.com/swiftlang/swift-corelibs-libdispatch/pull/888

compnerd:

lxbndr:
>Test Case 'TestURLSession.test_repeatedRequestsStress' passed (9.051 seconds)

9 seconds of stress testing. That's impressive. Before the fix it usually failed in first 2-3 seconds. I guess it worked. The only question now is: do we want to add 9 seconds to test run? If yes, then I have nothing to add here.

parkera:
I would rather not have stress tests be a part of the regular runs. Sometimes we comment them out so people working in the area can manually test. Another approach is to reduce the 'stress level' to something much shorter, and count instead on running it many times in all future CI runs.

lxbndr:
Reducing pressure (and test timings) seems reasonable. It still would be occasionally catching the issue in case of regression.

lxbndr:
@compnerd could you please trigger a new run? I am expecting about 1,5s duration now.

compnerd:
@swift-ci please test

compnerd:

lxbndr:
Seems like it is still crashing.
```
Test Case 'TestURLSession.test_repeatedRequestsStress' started at 2025-07-18 21:46:55.468
warning: Skipping test expected to fail with reason 'FileAttributes[.systemFreeNodes], FileAttributes[.systemNodes] not implemented'
2025-07-18 21:46:03.494 swift-corelibs-foundationPackageTests.xctest[9756:5b0] *** NSClassFromString(TestFoundation.SwiftClass.InnerClass): nested class names not yet supported
2025-07-18 21:46:39.997 swift-corelibs-foundationPackageTests.xctest[9756:1164] readAndRespond: requestTooShort
2025-07-18 21:46:40.000 swift-corelibs-foundationPackageTests.xctest[9756:1164] readAndRespond: requestTooShort
Error: Error: swift.exe exited with code 1.
```
I believe lines after test start are not related (timestamps are earlier than the test itself). But we have abnormal exit, so the test is still unstable, and the issue appeared to be more complex. Not sure though why the test run was triggered twice (while being requested once). It did complete successfully [here](https://ci-external.swift.org/job/swift-corelibs-foundation-PR-windows/721/) (and the test took 1.508 s, just as expected), but then it started again [here](https://ci-external.swift.org/job/swift-corelibs-foundation-PR-windows/722/). Perhaps it is because of new commit in the target branch.

@compnerd if that makes sense to you, I am going to close this PR. Guess I can look a bit deeper into this just to refresh the context we have about the issue.

hmelder:
Yea GNUstep's NSURLSession (which has a similar architecture to the swift impl) also experiences incomplete transfers. libdispatch still swallows some muxnotes, but less frequently. Seems like there is still a lifecycle issue with muxnotes. I'll investiagte this further.

lxbndr:
@hmelder Thanks!

PR #5251, opened 2025-07-30
Add brotli support to FoundationNetworking on Windows

zhenchaoli:
Add brotli as a dependency to enable brotli decoding for curl on Windows.

Only brotlicommon.lib and brotlidec.lib is linked.
brotlienc.lib is intentionally left out as curl does not need it. brotlienc.lib is also much larger in size (~3.8MB).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -58,7 +58,9 @@ if let environmentPath = Context.environment["CURL_INCLUDE_PATH"] {
 
 var curlLinkFlags: [LinkerSetting] = [
     .linkedLibrary("libcurl.lib", .when(platforms: [.windows])),
-    .linkedLibrary("zlibstatic.lib", .when(platforms: [.windows]))
+    .linkedLibrary("zlibstatic.lib", .when(platforms: [.windows])),
+    .linkedLibrary("brotlicommon.lib", .when(platforms: [.windows])),
+    .linkedLibrary("brotlidec.lib", .when(platforms: [.windows]))
 ]
 if let environmentPath = Context.environment["CURL_LIBRARY_PATH"] {
     curlLinkFlags.append(.unsafeFlags([
@@ -66,6 +68,11 @@ if let environmentPath = Context.environment["CURL_LIBRARY_PATH"] {
     ]))
 }
 if let environmentPath = Context.environment["ZLIB_LIBRARY_PATH"] {
+    curlLinkFlags.append(.unsafeFlags([
+        "-L\(environmentPath)"
+    ]))
+}
+if let environmentPath = Context.environment["BROTLI_LIBRARY_PATH"] {
     curlLinkFlags.append(.unsafeFlags([
         "-L\(environmentPath)"
     ]))
--- a/Tests/Foundation/HTTPServer.swift
+++ b/Tests/Foundation/HTTPServer.swift
@@ -481,6 +481,18 @@ class _HTTPServer: CustomStringConvertible {
                 "\r\n").data(using: .utf8)!
         try tcpSocket.writeRawData(responseData)
     }
+
+    func respondWithAcceptEncoding(request: _HTTPRequest) throws {
+        var responseData: Data
+        if let acceptEncoding = request.getHeader(for: "Accept-Encoding") {
+            let content = acceptEncoding.data(using: .utf8)!
+            responseData = "HTTP/1.1 200 OK\r\nContent-Type: text/html; charset=ISO-8859-1\r\nContent-Length: \(content.count)\r\n\r\n".data(using: .utf8)!
+            responseData.append(content)
+        } else {
+            responseData = "HTTP/1.1 200 OK\r\nContent-Type: text/html; charset=ISO-8859-1\r\nContent-Length: 0\r\n\r\n".data(using: .utf8)!
+        }
+        try tcpSocket.writeRawData(responseData)
+    }
 }
 
 struct _HTTPRequest: CustomStringConvertible {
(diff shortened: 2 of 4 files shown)


zhenchaoli:
https://github.com/swiftlang/swift/pull/83441
@swift-ci please test

zhenchaoli:
Seems build failed. I see that brotli is not checked out. Do I need a 'clean' test? 

According to https://github.com/swiftlang/swift/blob/main/docs/ContinuousIntegration.md, this is a thing, but unsure from reading it whether that will make a difference.

zhenchaoli:
https://github.com/swiftlang/swift/pull/83441
@swift-ci please clean test

zhenchaoli:
the related pull request is at https://github.com/swiftlang/swift/pull/83441

compnerd (changes requested):
Windows builds the toolchain without SPM and without the dependency automation in WindowsSwiftPMDependencies.cmake. We need this to be wired up into the CMake build properly.

jmschonfeld:
> Windows builds the toolchain without SPM and without the dependency automation in WindowsSwiftPMDependencies.cmake. We need this to be wired up into the CMake build properly.

What extra pieces would need to be wired up in this build here? I know it's expected that the Windows toolchain build wouldn't use `WindowsSwiftPMDependencies`, but I think that's fine? I don't see any references to zlib in this project and I believe the brotli setup is largely the same, are the `build.ps1` updates to pass the appropriate paths in as cmake variables insufficient for CMake to find the project in the dependency graph automatically?

zhenchaoli:
> Windows builds the toolchain without SPM and without the dependency automation in WindowsSwiftPMDependencies.cmake. We need this to be wired up into the CMake build properly.

My reading (and discussion with Jeremy) is that `cmake/modules/WindowsSwiftPMDependencies.cmake` is for local developments, and that the actual cmake changes are made in https://github.com/swiftlang/swift/pull/83441

The meaty piece in this PR is just the 2 test cases in my opinion.

compnerd:
Ah, so Foundation itself won't need to link to brotli because curl will automatically ensure that the paths are exported and passed along to the linker?

zhenchaoli:
Not sure about the automatic path passing part, but brotli (and zlib) are statically linked both here in `WindowsSwiftPMDependencies.cmake` for local development and in https://github.com/swiftlang/swift/pull/83441

My experience building Foundation on Windows is that the paths to dependencies (brotli, zlib, libxml, curl) are required to be manually set for the build to succeed, which is the reason for this change:

```
add_custom_command(TARGET WindowsSwiftPMDependencies POST_BUILD
  COMMAND echo BROTLI_LIBRARY_PATH=${BROTLI_LIBRARY_DIR})
```

compnerd:
Right, so that would occur by means of having an `IMPORTED` library (and I expect it to be `STATIC` linkage) that sets that configuration up and ties it into the target.

```cmake
add_library(brotli IMPORTED STATIC)
set_target_properties(brotli PROPERTIES
  IMPORTED_IMPLIB ...
  IMPORTED_LOCATION ...
  INTERFACE_INCLUDE_DIRECTORIES ...)
```

Then the `FoundationNetworking` target can `target_link_libraries(FoundationNetworking PRIVATE brotli)`.

zhenchaoli:
> Right, so that would occur by means of having an `IMPORTED` library (and I expect it to be `STATIC` linkage) that sets that configuration up and ties it into the target.
> 
> ```cmake
> add_library(brotli IMPORTED STATIC)
> set_target_properties(brotli PROPERTIES
>   IMPORTED_IMPLIB ...
>   IMPORTED_LOCATION ...
>   INTERFACE_INCLUDE_DIRECTORIES ...)
> ```
> 
> Then the `FoundationNetworking` target can `target_link_libraries(FoundationNetworking PRIVATE brotli)`.

@compnerd Now that https://github.com/swiftlang/swift/pull/83441 has wired up the windows toolchain build's CMake changes, does this change look reasonable?
1. added brotli tests
2. CMake changes for building dependencies for local development

I think initially your review comments are based on `Windows builds the toolchain without SPM and without the dependency automation in WindowsSwiftPMDependencies.cmake`

The changes here are for local development workflow outlined in https://github.com/swiftlang/swift-corelibs-foundation?tab=readme-ov-file#building-swift-corelibs-foundation-on-windows

compnerd (approved):

compnerd:
> I think initially your review comments are based on `Windows builds the toolchain without SPM and without the dependency automation in WindowsSwiftPMDependencies.cmake`

This still holds - we want to ensure that Foundation is built properly in this case, because this is the most important case as that is the shipping product.

zhenchaoli:
> > I think initially your review comments are based on `Windows builds the toolchain without SPM and without the dependency automation in WindowsSwiftPMDependencies.cmake`
> 
> This still holds - we want to ensure that Foundation is built properly in this case, because this is the most important case as that is the shipping product.

Yes.
Since the cmake changes here are for local development, I think we are good.
Tagging @jmschonfeld to get his thoughts

zhenchaoli:
@swift-ci please test

zhenchaoli:
@swift-ci please test

zhenchaoli:
@jmschonfeld I'm preparing to merge this PR and https://github.com/swiftlang/swift/pull/83441
and it looks like they need to be merged simultaneously.

https://github.com/swiftlang/swift/blob/main/docs/ContinuousIntegration.md#cross-repository-testing also mentions this case, but does not say how to actually merge them so that neither repo breaks.

Do I just merge them, and hope the ci updates to the latest main branch soon enough to not cause problems?

jmschonfeld:
@zhenchaoli Nope nothing special you need to do, merging them at the same time is sufficient to make sure they're picked up together. Since the swift repo change updates the checkout list, unfortunately there's no way to get CI on this PR to pass before the other PR is merged. But we can use the CI passing on the swift PR as signal that this change works and force merge this PR despite the failing Windows check when you're ready to merge both.

zhenchaoli:
@swift-ci please test

PR #5255, opened 2025-08-04
Fix WebSocket buffered read and add support for fragmented messages

andriydruk:
This PR fixes a bug in Foundation.URLSessionWebSocketTask where incoming WebSocket frames were incorrectly assumed to arrive in a single TCP read. Since TCP can arbitrarily split data, this led to dropped or malformed messages.

Fixes:
* Accumulates socket data until the entire WebSocket frame is received
* Adds support for fragmented WebSocket messages (multiple frames using opcode 0x0)

The WebSocket unit tests were extended to cover:
* Buffered sending – simulates server-side buffering by sending a message in small TCP chunks
* Fragmented message – splits a single message across multiple WebSocket frames


parkera:
@swift-ci test

guoye-zhang:
+@pseligman for review

PR #5256, opened 2025-08-04
Fix memory leak in NSAttributedString

andriydruk:
Currently, all attributes are leaking in `NSAttributedString`
This PR adds proper retain/release handling for `CFRunArray` in the appropriate places.

A test was added to validate that attribute values in `NSAttributedString` are no longer leaking.

### DIFF
--- a/Sources/CoreFoundation/CFRunArray.c
+++ b/Sources/CoreFoundation/CFRunArray.c
@@ -200,6 +200,14 @@ CFRunArrayRef CFRunArrayCreate(CFAllocatorRef allocator) {
     return array;
 }
 
+CFRunArrayRef CFRunArrayRetain(CFRunArrayRef array) {
+    return COPY(array);
+}
+
+void CFRunArrayRelease(CFRunArrayRef array) {
+    FREE(array);
+}
+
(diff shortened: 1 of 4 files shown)


Sources/CoreFoundation/CFRunArray.c
```diff
@@ -200,6 +200,14 @@ CFRunArrayRef CFRunArrayCreate(CFAllocatorRef allocator) {
     return array;
 }
 
+CFRunArrayRef CFRunArrayRetain(CFRunArrayRef array) {
```
parkera, line 203: Is `CFRunArray` not a regular CF type? (`CFRetain`, `CFRelease` should work)

andriydruk: When I try to use CFRetain, I get the following error:
> 'CFRetain' is unavailable: Core Foundation objects are automatically memory managed

parkera: So if it’s automatically memory managed, where is the leak coming from? We don’t have to do this with other CF types.

itingliu:
> Currently, all attributes are leaking in NSAttributedString

How did you find out? Do you have a leaks report?

PR #5272, opened 2025-09-27
Add brotli as a dependency

zhenchaoli:
(This is the 6.2 back port of https://github.com/swiftlang/swift-corelibs-foundation/pull/5251)

Add brotli as a dependency to enable brotli decoding for curl.

On Windows, only brotlicommon.lib and brotlidec.lib is linked.
brotlienc.lib is intentionally left out as curl does not need it. brotlienc.lib is also much larger in size (~3.8MB).


parkera (approved):

parkera:
@swift-ci test

parkera:
@swift-ci test windows

PR #5273, opened 2025-09-28
FreeBSD: Set TSD to NULL when maximum pthread dtor iterations reached

michael-yuji:
On FreeBSD, libthr emits debug message to stderr if a TSD is non-null after `PTHREAD_DESTRUCTOR_ITERATIONS` of dtor getting called. 

Currently, although we skipped setting the TSD to `CF_TSD_BAD_PTR` like other platforms, the TSD still remains non-null after max iterations of dtor calls and setting the TSD to NULL explicitly is needed


jakepetroules:
@swift-ci test

PR #5274, opened 2025-10-05
Fix incorrect comments for time range in CFCalendar.c

UInt2048:
The definitions of these macros do not match their documentation, which would have been the following:

```c
#define MIN_CALENDAR_TIME -211873320000.0 // Julian day 0 (-4713-01-01 12:00:00 +0000) in CFAbsoluteTime.
#define MAX_CALENDAR_TIME 1514702073600.0 // 50000-01-01 00:00:00 +0000, smaller than the max time ICU supported.
```

This pull request proposes to keep the existing behavior but properly document it


parkera:
@swift-ci test

UInt2048:
@parkera Sorry, I had to push a correction to the PR after checking the definition of Julian day 0. It appears both swift-foundation's `Date.swift` and this file both incorrectly use Julian year -4713 (4714 BC) instead of -4712 (4713 BC) as the date of Julian day 0 in the comment, despite the interval itself appearing to be correct.

https://github.com/swiftlang/swift-foundation/blob/abfcc288dfca47d47396d099ba8b5269d17f4b06/Sources/FoundationEssentials/Date.swift#L392

UInt2048:
Out of curiosity, where does the date 506713-02-07 00:00:00 +0000 come from? It seems like the [ICU universal time scale](https://unicode-org.github.io/icu/userguide/datetime/universaltimescale.html) only has a range back to 29,000 BC and up to 29,000 AD, so the only thing I can think of is that 25,000 years or so were added to the back half of the range since dates before Julian day 0 weren't supported and that number (probably 53,286 AD or so) was rounded down to get 50,000 AD like the current documentation says, but then that maximum was miscomputed.

On the other hand, things like the iOS Calendar app seem to work fine with the full range up to 506713 AD (as far as I can tell with the Gregorian calendar anyways) so it doesn't seem to be posing issues in practice if this theory is actually what happened.

itingliu:
> Out of curiosity, where does the date 506713-02-07 00:00:00 +0000 come from? It seems like the [ICU universal time scale](https://unicode-org.github.io/icu/userguide/datetime/universaltimescale.html) only has a range back to 29,000 BC and up to 29,000 AD, so the only thing I can think of is that 25,000 years or so were added to the back half of the range since dates before Julian day 0 weren't supported and that number (probably 53,286 AD or so) was rounded down to get 50,000 AD like the current documentation says, but then that maximum was miscomputed.

ICU has this implicit limit in their implementation that would cause some of their API to become no-op when the input exceeds the limit: https://github.com/swiftlang/swift-foundation-icu/blob/122f088ccda752062b4c44d4d8471ecb3638da32/icuSources/i18n/gregocal.cpp#L71
```
// setTimeInMillis() limits the Julian day range to +/-7F000000.
// This would seem to limit the year range to:
//  ms=+183882168921600000  jd=7f000000  December 20, 5828963 AD
//  ms=-184303902528000000  jd=81000000  September 20, 5838270 BC
// HOWEVER, CalendarRegressionTest/Test4167060 shows that the actual
// range limit on the year field is smaller (~ +/-140000). [alan 3.0]
```

We decided to set a cap on our own instead of relying on their implicit limits so we know exactly what we're doing. The upper bound is just a random day in the far far future that sits safely below ICU's upper bound.

itingliu:
@swift-ci please test

UInt2048:
I apologize for bothering you. The corresponding change to swift-foundation's `Date.swift` has been successfully merged 2 weeks ago. Can you please provide a status update on whether anything further is blocking this PR?

itingliu:
My bad. Thanks!

PR #5278, opened 2025-10-14
[test] Disable test to unblock CI

AnthonyLatsis:


AnthonyLatsis:
Disabled only for Linux instead in https://github.com/swiftlang/swift-corelibs-foundation/pull/5279

PR #5289, opened 2025-11-23
Ensure strsep operates on a copy of PATH.

3405691582:
strsep modifies its first argument, which means that if we don't make a copy, any PATH processing that may occur after _CFProcessPath executes breaks because strsep will effectively change PATH to be only the first list item. For example, swiftpm's tool finding fails because it searches PATH, and if the tool in question doesn't reside in the first PATH item, things break.

This only occurs on BSD and only when the program name, i.e., argv[0], is not an absolute path.

(As a heads up: this should be picked onto release branches.)

### DIFF
--- a/Sources/CoreFoundation/CFPlatform.c
+++ b/Sources/CoreFoundation/CFPlatform.c
@@ -266,10 +266,12 @@ const char *_CFProcessPath(void) {
     }
 
     // Search PATH.
-    if (argv0) {
-        char *paths = getenv("PATH");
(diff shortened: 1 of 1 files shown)


3405691582:
@swift-ci please test.

parkera:
We will need to wait until the branch is reopened to merge

3405691582:
@swift-ci please test.

3405691582:

PR #5290, opened 2025-11-25
Fix various typos

Coeur:
This is the promised follow up of #5075.
This is an exhaustive review.

`__CFCheckForExapendedSet` was renamed to `__CFDebugExpandedSet` because the symbol `__CFCheckForExpandedSet` already exists for something else.


parkera:
@swift-ci test

PR #5295, opened 2025-12-02
[6.3] Windows: Use `SHGetKnownFolderPath` to compute `LocalAppData`

ShikiSuen:
## Summary

**Target Branch:** [swiftlang:release/6.3](https://github.com/swiftlang/swift-corelibs-foundation/tree/release/6.3)

This PR replaces manual `LocalAppData` path construction with the Windows Shell API `SHGetKnownFolderPath` to reliably obtain the correct per-user local application data directory.   

### Problem

On Windows, the environment variable `USERPROFILE` is not guaranteed to equal `%SYSTEMDRIVE%\Users\%USERNAME%`. When the manually constructed `LocalAppData` path does not exist, `UserDefaults` writes silently no-op, causing preferences to never persist.

### Solution

Use `SHGetKnownFolderPath(&FOLDERID_LocalAppData, ...)` to programmatically retrieve the correct `LocalAppData` path instead of building it from environment variables or string concatenation.

---

## Changes

- Replace hand-built `LocalAppData` path construction with `SHGetKnownFolderPath` API call
- Add `#include <shlobj_core.h>` for the Shell API
- Use `CoTaskMemFree` to properly release the pointer returned by `SHGetKnownFolderPath`
- Tighten control flow for `_kCFKnownLocationUserCurrent` to prevent unintended fallthrough after successful username lookup

---

## Scope

| Aspect | Details |
|--------|---------|
| **Platform** | Windows only (`TARGET_OS_WIN32`) |
| **Files Changed** | `Sources/CoreFoundation/CFKnownLocations.c` |
| **Public API** | No changes |
| **Other Platforms** | Unaffected |

---

## Risk Assessment

**Severity:** Low to Moderate (Windows-only)

### Low Risk
- No public API changes
- Uses a well-documented Windows API to find the current user's `LocalAppData`
- Reduces failure modes from incorrect path assumptions

### Moderate Risk
- **ByName Path Lookup:** The previous implementation could potentially derive a `LocalAppData` path for users other than the current user (via `GetProfilesDirectoryW` + username).  The new approach uses `SHGetKnownFolderPath`, which returns the path for the *current* user context only. If support for other-user preference paths (`CFKnownLocationUserByName`) on Windows is required, an alternative approach (e.g., SID impersonation or `GetProfilesDirectoryW`) may be needed. 

### Build/Link Considerations
- Requires `shlobj_core.h` header and linking to `ole32` or `shell32` libraries
- Verify build system provides these dependencies for Windows targets

### Runtime Considerations
- `SHGetKnownFolderPath` is available on Windows Vista and later; Swift targets higher WinNT builds, hence no worries.
- Memory is correctly managed using `CoTaskMemFree`

---

## References

This approach is consistent with how .NET runtime handles `LocalAppData` path retrieval on Windows:

- [Environment.Win32.cs - GetFolderPathCore](https://github.com/dotnet/dotnet/blob/0dcb6ee24ca4d7c959ee55e09abc34c727799fb0/src/runtime/src/libraries/System.Private.CoreLib/src/System/Environment.Win32.cs#L225-L243)
- [Interop.SHGetKnownFolderPath.cs](https://github.com/dotnet/dotnet/blob/0dcb6ee24ca4d7c959ee55e09abc34c727799fb0/src/runtime/src/libraries/Common/src/Interop/Windows/Shell32/Interop.SHGetKnownFolderPath.cs#L170-L183)
- [Evaluator_Tests.cs - Related test cases](https://github.com/dotnet/dotnet/blob/0dcb6ee24ca4d7c959ee55e09abc34c727799fb0/src/msbuild/src/Build.UnitTests/Evaluation/Evaluator_Tests.cs#L2688-L2728)

---

## Testing

*TO BE FILLED IN*

---

## Reviewers

*TO BE FILLED IN*

### DIFF
--- a/Sources/CoreFoundation/CFKnownLocations.c
+++ b/Sources/CoreFoundation/CFKnownLocations.c
@@ -18,6 +18,7 @@
 
 #if TARGET_OS_WIN32
 #include <userenv.h>
+#include <shlobj_core.h>
 #endif
 
 CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUser user, CFStringRef _Nullable username) {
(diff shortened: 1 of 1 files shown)


compnerd:

compnerd:

compnerd:
Could you please include the filled out CCC template from https://github.com/swiftlang/.github/blob/main/PULL_REQUEST_TEMPLATE/release.md?plain=1

ShikiSuen:
@compnerd Does this work with you?:

```
- **Explanation**:
  - On Windows, it is not guaranteed that the environment variable `USERPROFILE` is equivalent to `%SYSTEMDRIVE%\Users\%USERNAME%`. Because of this, building the LocalAppData path using environment variables or string concatenation can result in a path that does not exist, and writes may silently no-op.
  - This change replaces hand-built LocalAppData path construction with a proper Windows Shell API call: `SHGetKnownFolderPath(&FOLDERID_LocalAppData, ...)` to get the correct per-user LocalAppData path.
  - Additional cleanup: include `<shlobj_core.h>` and use `CoTaskMemFree` to free the pointer from `SHGetKnownFolderPath`. The change also tightens a fallthrough control flow for `_kCFKnownLocationUserCurrent` so we do not accidentally fall through when the username lookup succeeded.

- **Scope**:
  - Platform-specific change: only affects Windows (TARGET_OS_WIN32).
  - File(s) changed: `Sources/CoreFoundation/CFKnownLocations.c`.
  - Behavior change: CFPreferences path lookup for the current user is now retrieved through the Windows shell API rather than manually constructing a path. This should make LocalAppData detection more reliable.
  - No public API surface changes; behavior for other platforms is unaffected.
  - Potential behavior change: logic for "ByName" (i.e., `CFKnownLocationUserByName`) may no longer compute the LocalAppData for a specified user other than the current user; instead it uses the OS API to get the path for the current account. If the intent is to support accessing preferences for other users on Windows, this must be assessed (see Risk section).

- **Issues**:
  - Fixes a scenario where writing preferences silently no-ops because the computed LocalAppData path does not exist on Windows due to assumptions about `USERPROFILE` being equivalent to `%SYSTEMDRIVE%\Users\%USERNAME%`.
  - Related behavior: If the code previously relied on manually-built path logic to support non-current user "ByName" paths on Windows, that functionality may require re-evaluation.

- Original PRs:
  Derived from mainline work: "use shgetknownfolderpath instead of hand building the path" and follow-up Windows fixes.

- Risk:
  - Risk severity: Low to Moderate (Windows-only).
  - Rationale:
    - Low: No API changes; logic cleanup reduces certain failure modes and uses a documented OS API to find the current user's LocalAppData.
    - Moderate: The previous code path that attempted to build a per-user path based on profiles may have been able to derive a `LocalAppData` path for a user other than the current user (e.g., by combining `GetProfilesDirectoryW` + username); the new approach uses `SHGetKnownFolderPath` which returns the path for the effective user context. If the project intends to support preference paths for other users (ByName) on Windows, then that feature may need an explicit approach (e.g., obtaining a path via SID impersonation or `GetProfilesDirectoryW` for other users).
  - Build/link risk:
    - Uses `SHGetKnownFolderPath` and `FOLDERID_LocalAppData` which requires appropriate headers (`shlobj_core.h`) and linking to Windows libraries (ole32 or shell32 may be required). Verify that the build system provides these for Windows targets.
  - Runtime risk:
    - `SHGetKnownFolderPath` may fail on older Windows versions; ensure fallback is appropriate if needed (i.e., graceful fallback to prior methods or to a sensible default).
    - Memory management must handle the pointer returned by `SHGetKnownFolderPath` correctly (using `CoTaskMemFree`) — the commit does this.

- Testing:
  - <TO BE FILLED-IN.>

- Reviewers:
  - <TO BE FILLED-IN.>
```

ShikiSuen:
@compnerd I added the filled template to the main post. There are still fields pending finish depending on the CI results.

compnerd:
> @compnerd I added the filled template to the main post. There are still fields pending finish depending on the CI results.

Thanks, yes, that seems fine. I will kick off the windows tests once more, but it might be related tot his change.

compnerd:

ShikiSuen:
@swift-ci Please test Windows platform.

ShikiSuen:
@swift-ci Please test Windows platform.

ShikiSuen:
I removed those commits generated by CfCCAgent and rebased to let this PR inherit recent changes from the merging target.

Lemme draft this PR for now. Gonna see whether we can align the behavior to what dotNET 10 does.

ShikiSuen:
(I intentionally squashed all commits and removed `@stevenbrix` 's credit. Reason: This PR now uses the behavior of dotNET 10 instead.)

ShikiSuen:
@swift-ci Please test on Windows.

ShikiSuen:
Shutting down this PR after serious discussion with some of my Windows-app-developing fellows.

The reason will be posted as a reply under PR#5295.

PR #5303, opened 2025-12-09
Fix socket lifecycle management

clintonpi:
Motivation:

When an `easy_handle` is added to a `multi_handle`, the callback set for `CURLOPT_CLOSESOCKETFUNCTION` constructs a `_MultiHandle` from an unretained reference to `self`. This could cause a crash in `URLSession` if the `_MultiHandle`'s `self` is deinitialised before the callback is called, and the socket will never be closed.

_Partially resolves #3813_

Modifications:

- Set the `CURLOPT_CLOSESOCKETFUNCTION` callback to do nothing, deferring closing the socket.
- Schedule a socket for closure in the cancellation handler of its dispatch sources.

Result:

Better management of socket lifecycle.

Testing:

Passing existing tests is sufficient.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
+++ b/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
@@ -203,16 +203,11 @@ private extension URLSession._MultiHandle {
             // we should cancel pending work when unregister action is requested.
             precondition(!socketReference.shouldClose, "Socket close was scheduled, but there is some pending work left")
             workItem.perform()
+
+            /// Marks this reference to close socket on deinit. This allows us
+            /// to extend socket lifecycle by keeping the reference alive.
+            socketReference.shouldClose = true
         }
-
-        self.scheduleClose(for: socketReference.socket)
-    }
-
-    /// Marks this reference to close socket on deinit. This allows us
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
```diff
@@ -195,22 +195,19 @@ private extension URLSession._MultiHandle {
...
 
-    /// Marks this reference to close socket on deinit. This allows us
-    /// to extend socket lifecycle by keeping the reference alive.
-    func scheduleClose(for socket: CFURLSession_socket_t) {
-        let reference = socketReferences[socket] ?? _SocketReference(socket: socket)
-        reference.shouldClose = true
+            /// Marks this reference to close socket on deinit. This allows us
+            /// to extend socket lifecycle by keeping the reference alive.
+            socketReference.shouldClose = true
+        }
```
jrflat: Should we set `.shouldClose = true` outside of the `if let workItem = ...` block? It looks like `tearDown(handle:socket:queue)` will call `handle.endOperation(for: socketReference)` where `socketReference.workItem` is `nil`.

PR #5304, opened 2025-12-09
Clean up `multi_handle`s on the same thread

clintonpi:
Motivation:

`URLSession` elusively crashes due to a race condition in OpenSSL <1.1.0 when calling `curl_multi_cleanup` concurrently in different threads.

_Partially resolves #3813_

Modifications:

- Adjust `_MultiHandle.deinit` to asynchronously queue `curl_multi_cleanup` calls in `_MultiHandle.queue`.

Result:

Saves OpenSSL from cleaning up concurrently in different threads, avoiding the crash.

Testing:

Passing existing tests is sufficient. However, without this patch, the code snippet below crashes about once in 20 runs on Amazon Linux 2
> _curl 8.3.0 (aarch64-koji-linux-gnu) libcurl/8.3.0 OpenSSL/1.0.2k-fips zlib/1.2.7 libidn2/2.3.0 libpsl/0.21.5 (+libidn2/2.3.0) libssh2/1.4.3 nghttp2/1.41.0 OpenLDAP/2.4.44_
> _Release-Date: 2023-09-13_

```
DispatchQueue.concurrentPerform(iterations: 100) { _ in
    let session: URLSession = URLSession(configuration: URLSessionConfiguration.default)
    DispatchQueue.concurrentPerform(iterations: 100) { _ in
        _ = session
    }
}
```

### DIFF
--- a/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
+++ b/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
@@ -58,7 +58,16 @@ extension URLSession {
             easyHandles.forEach {
                 try! CFURLSessionMultiHandleRemoveHandle(rawHandle, $0.rawHandle).asError()
             }
-            try! CFURLSessionMultiHandleDeinit(rawHandle).asError()
+
+            let rawHandleSendable = Int(bitPattern: rawHandle)
+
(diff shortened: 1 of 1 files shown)


clintonpi:
Asynchronously cleaning up multi handles causes an issue where the socket close callback runs after `_MultiHandle` has already been deallocated. #5303 attempts to preemptively fix that. However, both patches will be abandoned for a better solution (#5321).

PR #5306, opened 2025-12-10
Windows: correct `Thread.stackSize` computation

LumioseSil:
Yield the total return of the stack, not the bottom limit.


compnerd:
@swift-ci please test

compnerd (approved):

PR #5307, opened 2025-12-10
__CFBinaryPlistGetOffsetForValueFromDictionary3 returns null instead of false

LumioseSil:
I guess this has worked when compiled, but this is semantically wrong.


compnerd:
@swift-ci please test

PR #5311, opened 2025-12-11
Off-by-one error in precondition

LumioseSil:
for index <= position, if bytes.count == 0 and index == position, then we don't want this condition to be true at all.


AZero13:
@parkera Thoughts?

jmschonfeld:
@swift-ci please test

jmschonfeld:

jmschonfeld (approved):

PR #5312, opened 2025-12-11
Deinitialize all the keys instead of just the first one

LumioseSil:


compnerd:
@swift-ci please test

compnerd:

parkera (approved):

Sources/Foundation/NSCFDictionary.swift
```diff
@@ -71,7 +71,7 @@ internal final class _NSCFDictionary : NSMutableDictionary {
                 let key = unsafeBitCast(keys.advanced(by: idx).pointee!, to: NSObject.self)
                 keyArray.append(key)
             }
-            keys.deinitialize(count: 1)
+            keys.deinitialize(count: count)
```
parkera: Are we also missing a call to actually initialize the keys? I'm not sure what it means to deinitialize anything at all without first initializing it.

It's probably true that in practice it is not necessary to deinitialize raw pointers at all.

PR #5314, opened 2025-12-11
Allow estimate to be used on all platforms

LumioseSil:


compnerd:
@swift-ci please test

AZero13:
> This was done to avoid wasting memory on more memory-sensitive platforms. Why do you feel this should be removed now?

Okay so I guess we can remove it entirely. I guess you can count embedded swift a memory sensitive platform.

parkera:
@swift-ci test windows

PR #5293, opened 2025-11-26
[wasm] Use pthread API for NSLock for wasip1-threads target

kateinoigakukun:
`SWIFT_CORELIBS_FOUNDATION_HAS_THREADS` is not defined for all WASI targets now but we do have pthread support for wasip1-threads target. `_runtime(_multithreaded)` check is more appropriate here to detect multithreaded runtime support. We can remove `SWIFT_CORELIBS_FOUNDATION_HAS_THREADS` definition later.

PR #5319, opened 2025-12-13
[wasm] Remove no longer used threading compile-time defines

kateinoigakukun:
Depends on https://github.com/swiftlang/swift-corelibs-foundation/pull/5293

After #5293, we no longer rely on `SWIFT_CORELIBS_FOUNDATION_HAS_THREADS` definitions


compnerd (approved):

kateinoigakukun:
@swift-ci test

PR #5331, opened 2025-12-27
CoreFoundation: correct mode computation in `_NS_chmod` on Windows

LumioseSil:


compnerd (approved):

compnerd:
Is this path in use? I am unsure if swift-foundation uses `_NS_chmod`. However, it doesn't hurt to fix this while we are still using CF.

compnerd:
@swift-ci please test

PR #5334, opened 2025-12-27
Use the new count rather than the array itself to determine the new length

LumioseSil:
If the middle array element is remove for example, this code is wrong.

range.length = range.length - (idx - range.location) - 1; before updating range.location to idx is also a way to do this.


AZero13:
@parkera

PR #5335, opened 2025-12-27
Make dotLocation CFIndex

LumioseSil:
Having it be uint16_t for no reason is bad for ARM64 and ARM.


parkera:
Can you tell me more about why it is "bad" for ARM?

PR #5342, opened 2026-01-01
CFBurstTrie: Release resources when error happens

LumioseSil:
Error check and properly release when errors happen.


AZero13:
@parkera thoughts?

parkera:
@swift-ci test

PR #5343, opened 2026-01-01
Fix copy paste code error: allocate and assign to value

LumioseSil:
We mean to assign to value, not key


AZero13:
@parkera I know you don't do this normally, but you may want to fix this bug in Apple's corefoundation too.

compnerd (approved):

compnerd:
@swift-ci please test

AZero13:
@parkera It passed!

AZero13:
@parkera It passed!

etcwilde (commented):
Nice catch. Seems like something that should have a test.
I could be missing it, but I don't actually see where this is getting called or exposed.

AZero13:
> Nice catch. Seems like something that should have a test.
> I could be missing it, but I don't actually see where this is getting called or exposed.

Probably why this bug wasn't caught for the 26? Years this code shipped assuming it was when OS X did

AZero13:
@jmschonfeld It passed!

jmschonfeld:
Indeed it looks like this function isn't used anywhere. Since we can't test it and it isn't used, how about we just delete `__CFGetEnvironment()` entirely (and `__CFGetPid()` below it which also seems unused). If it's just dangling code there's no point in fixing bugs here when we can just remove the bug instead.

AZero13:
Vestigial code?

parkera:
This code never existed in the Darwin version of Foundation - it is only for swift-corelibs-foundation.

parkera:
If it's not referenced in this project anymore, we can just delete it.

AZero13:
Done

jmschonfeld:
@swift-ci test

jmschonfeld (approved):

jmschonfeld:
@swift-ci test Windows platform

jmschonfeld:

PR #5348, opened 2026-01-04
TARGET_OS_WASI is not possible

LumioseSil:
We check for it not being defined.


jmschonfeld:
This can be closed now that https://github.com/swiftlang/swift-corelibs-foundation/pull/5343 is deleting this function entirely, right?

AZero13:
> This can be closed now that https://github.com/swiftlang/swift-corelibs-foundation/pull/5343 is deleting this function entirely, right?

Yes

PR #5349, opened 2026-01-04
Incorrect overlap check causes performance regression by allocating even on empty buffers

LumioseSil:
We were checking the wrong direction for overlap. This is always true for non-empty buffers, causing false positives for disjoint buffers where srcBuf < dstBuf.


AZero13:
@parkera I know this is not meant to go into apple's upstream corelibs, but since this is a bug, can you consider this too?

AZero13:
The original bug had srcBuf < newBytes + newBytesLength, which is trivially always true (a pointer is always less than itself plus a positive length), causing every buffer with src < dst to be incorrectly flagged as overlapping.

AZero13:
@jmschonfeld What do you think?

compnerd:
@swift-ci please test

PR #5350, opened 2026-01-04
Move the entire buffer, not half the bytes

LumioseSil:
We were only moving half the buffer, as Unichar is 2 bytes.

Bug also affects macOS CoreFoundation.


AZero13:
@jmschonfeld

PR #5354, opened 2026-01-05
Prevent crash when calling CFStringGetMostCompatibleMacStringEncoding with 0x200

LumioseSil:
Affects system CoreFoundation on Mac too.

This should just return invalidID, but instead it does underflow. I could not create a real actual exploit chain with this, nor is it like serious enough to have a CVE, because chances are otherwise this would have been exploited a lot more considering how simple this was.

### DIFF
--- a/Tests/Foundation/TestCFStringEncoding.swift
+++ b/Tests/Foundation/TestCFStringEncoding.swift
@@ -8,6 +8,7 @@
 //
 
 import CoreFoundation
+
 class TestCFStringEncoding: XCTestCase {
 
     func test_mostCompatibleMacStringEncoding_0x200() {


AZero13:
@jmschonfeld

AZero13:
Maybe I can add the invalid id to __CFISO8859SimilarScriptList as an alternative, and then change the - 1 access instead?

AZero13:
@parkera and the others too. I know it's a minor code change but it's your codebase.

AZero13:
@chloe-yeo @glessard opinion?

AZero13:
@glessard done!

glessard:
@swift-ci please test

glessard:

glessard:
@swift-ci please test

AZero13:
> @swift-ci please test

Realized the test won't pass on Windows because the code is guarded by #if in the CoreFoundation source

Sources/CoreFoundation/CFStringEncodingDatabase.c
```diff
@@ -780,14 +780,18 @@ static const char * const __CFOtherNameList[] = {
 #endif /* TARGET_OS_OSX || TARGET_OS_LINUX || TARGET_OS_WIN32 || TARGET_OS_BSD */
 
 CF_PRIVATE CFStringEncoding __CFStringEncodingGetMostCompatibleMacScript(CFStringEncoding encoding) {
-#if TARGET_OS_OSX || TARGET_OS_LINUX
+#if TARGET_OS_OSX || TARGET_OS_LINUX || TARGET_OS_WIN32 || TARGET_OS_BSD
```
glessard: Is this what we want? Expanding the coverage here is probably harmless, but it's well beyond the bug fix that this should be. Let's not add these, and conditionalize the test instead. Our goal is to have all of this functionality implemented in swift-foundation, so it's okay to fix bugs but probably not to add functionality.

AZero13: You know what? Maybe that's best for another time.

I'll just revert that change.

AZero13:
@glessard okay, It's ready now?

glessard:

glessard:

AZero13:
Try one more time because I don't know how I messed that up @glessard

glessard:
@swift-ci please test

glessard:

glessard:

AZero13:
It passed!

AZero13:
Does Apple plan on fixing this on their end too since it happens there when I pass 0x200

AZero13:
Maybe I should file a radar

parkera:
We'll take care of it, thanks.

PR #5364, opened 2026-01-10
Move the equality tests together

AZero13:


jmschonfeld (changes requested):
I'm going to mark this as rejected because there's no real value to doing this. If we want to, we can fix this while updating the tests but we generally don't re-organize just for the sake of re-organizing

PR #5367, opened 2026-01-10
[6.3] Fix character equality check by checking that the lengths are equal, not both nonzero

LumioseSil:
* Fix character equality check by checking that the lengths are equal, not just nonzero.

This was causing differing sets to be incorrectly evaluated as equal.

* Remove superfluous message in test failure.


itingliu (approved):

jmschonfeld:
@swift-ci please test

PR #5373, opened 2026-01-11
Fix illegal character detection in Planes 15 and 16

LumioseSil:
This PR fixes a logic error in `CFUniCharIsMemberOf` that prevented the correct identification of noncharacters (illegal characters) in Planes 15 and 16.

The Unicode Standard designates the last two code points of every plane (ends in `FFFE` and `FFFF`) as noncharacters. In `CFUniChar.c`, the check for these illegal characters in Planes 15 and 16 contained a typo where only the bottom 8 bits were checked, causing all results to return false.

Also affects macOS CoreFoundation.


AZero13:
@parkera

PR #5374, opened 2026-01-11
Move return success out of the preprocessor block to avoid undefined behavior

LumioseSil:


jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #5376, opened 2026-01-11
We can end up truncating one digit by mistake in _CFURLComponentsCopyString

LumioseSil:
snprintf needs to be passed in the whole buffer length.


parkera:
@swift-ci test

PR #5378, opened 2026-01-11
Check should be for *bytes, not bytes

LumioseSil:
We need to check if the CFAllocator failed


jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

PR #5379, opened 2026-01-11
Call uncv close on non-empty slots

LumioseSil:
NULL != converter was always true.


jmschonfeld:
@swift-ci please test

PR #5382, opened 2026-01-12
Remove redundant OSMemoryBarrier from CFBundle.c

LumioseSil:
When a bundle is unique, it is allocated and initialized on the current thread and returned to the same thread.

Since the pointer isn't published to any other thread, no memory barrier is needed to ensure initialization visibility.

If the bundle is not unique, the code immediately acquires CFBundleGlobalDataLock.

If a user passes the bundle pointer to another thread in a way that leads to races, that is a bug with the user, and their problem.


parkera:
Hm, I need to think back to why we introduced this in the first place. The comment suggests it was not safe to use with `dispatch_once` otherwise, but it may be outdated.

PR #5386, opened 2026-01-13
[6.2] Prevent crash when calling CFStringGetMostCompatibleMacStringEncoding with 0x200

LumioseSil:
This should just return invalidID, but instead it does underflow, causing a crash on any process that calls this at best, and accesses memory it should not at worst.

Zero risk because we go from crashing to returning an invalidID.


parkera:
@jmschonfeld what do you think of backporting this?

parkera:
@swift-ci test

jmschonfeld:
> @jmschonfeld what do you think of backporting this?

Seems fine to me if there's another 6.2.x release and the tests pass. We haven't seen any fallout from this on the 6.3 or main branches so I think the risk is mitigated and it's a good fix to have

AZero13:
Seems like it's gonna end with 6.2.3 so it's not gonna happen.

PR #5391, opened 2026-01-17
Use C99 flexible arrays for CFBasicHash

LumioseSil:
This is so the compiler can know one can access pointers beyond the size.


jmschonfeld:
> This is so the compiler can know one can access pointers beyond the size.

Can you elaborate more on what impact this has on this project or what incorrect behavior this corrects?

PR #5392, opened 2026-01-17
Replace sprintf with snprintf for safer string formatting

LumioseSil:
Updated macros in CFString.c to use snprintf and snprintf_l instead of sprintf and fixed buffer size usage. This change improves safety by preventing potential buffer overflows during string formatting.


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #5393, opened 2026-01-17
Check that CFAllocatorAllocate works before operating on bytes

LumioseSil:
Otherwise we get a NULL deref


itingliu:
Isn't this the same as #5397?

PR #5394, opened 2026-01-17
Fix conditional definition of HAVE_SCHED_GETAFFINITY

LumioseSil:
HAVE_SCHED_GETAFFINITY is now only defined for C files when the symbol is actually available, preventing incorrect macro definition on unsupported systems.


jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

PR #5395, opened 2026-01-17
Fix file descriptor check in fileFinalize

LumioseSil:
Change file descriptor check from 'fd > 0' to 'fd >= 0' in fileFinalize to ensure file descriptors with value 0 are also properly handled.


parkera (approved):

parkera:
@swift-ci test

PR #5396, opened 2026-01-17
0 is a valid fd for open

LumioseSil:
Never say open() failed if fd is 0, even if this is unlikely to happen


parkera:
@swift-ci test

PR #5397, opened 2026-01-19
Check that bytes is not null before memset

LumioseSil:
Otherwise, we get NULL deref


jmschonfeld:
There are still some unanswered questions here and there have been no recent updates. Closing, but feel free to re-open to continue the discussion

PR #5403, opened 2026-02-01
XCTest: replace `String(reflecting:)` with `_typeName`

ChrisBenua:
Optimized XCTest test listing and filtering by replacing `String(reflecting:)` with `_typeName`

### Motivation:

Calculating type-name using heavy `String(reflecting:)` is inefficient. There is far more performant alternative - `_typeName(_:qualified: true)`

### Modifications:

Replaced `String(reflecting:)` with `_typeName(_:qualified: true)`

### Result:

Performance of XCTest test filtering and listing will be slightly improved

### Testing:

Tested locally. In my opinion, testing is not needed, because these calls are basically equivalent when its argument is `Any.Type`. The same change was made here: https://github.com/swiftlang/swift/pull/77369

### DIFF
--- a/Sources/XCTest/Private/TestFiltering.swift
+++ b/Sources/XCTest/Private/TestFiltering.swift
@@ -67,6 +67,6 @@ private extension SelectedTest {
     }
 
     init(testCaseClass: XCTestCase.Type, testCaseMethodName: String?) {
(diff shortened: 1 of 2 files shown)


ChrisBenua:
Thanks! Didn't notice it. I will reopen PR in swift-corelibs-xctest

PR #5406, opened 2026-02-06
Fix bounds handling with vsnprintf

LumioseSil:
Make sure to not go beyond the bounds.


jmschonfeld:
@swift-ci please test

PR #5411, opened 2026-02-17
[URLSession] Add more thread-safety fixes for OpenSSL <1.1.0

clintonpi:
### Motivation:

`curl_global_init` also initializes the SSL library, and `curl_multi_socket_action` triggers certificate loading, SSL handshake, etc. As these functions affect the SSL library state, their calls should be serialized for a thread-unsafe OpenSSL.

### Modifications:

- Refactor libcurl initialization to be idempotent.
- Ensure libcurl has been initialized before calling `curl_version_info`. See https://curl.se/libcurl/c/curl_version_info.html.
- Serialize calls to `curl_multi_socket_action`.

### Result:

A thread-safer URLSession.

### Testing:

Passing the existing tests is sufficient.

---

_Related to #5321_

_Note: These fixes do not anticipate a limitation in libcurl (for OpenSSL 1.1.0+) where "[libcurl may not be able to fully clean up after multi-threaded OpenSSL depending on how OpenSSL was built and loaded as a library.](https://curl.se/libcurl/c/threadsafe.html)" Likely causing https://github.com/swiftlang/swift-package-manager/issues/8171_

### DIFF
--- a/Sources/FoundationNetworking/URLSession/libcurl/libcurlHelpers.swift
+++ b/Sources/FoundationNetworking/URLSession/libcurl/libcurlHelpers.swift
@@ -16,13 +16,8 @@
 ///
 // -----------------------------------------------------------------------------
 
-#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-import SwiftFoundation
-#else
-import Foundation
-#endif
-
 @_implementationOnly import _CFURLSessionInterface
+internal import Synchronization
 
 //TODO: Move things in this file?
 
@@ -36,9 +31,7 @@ internal func ensureLibcurlIsInitialized() {
     _ = _initializeLibcurl // The lazy global is computed on only first access.
 }
 
-/// OpenSSL < 1.1.0 is not thread-safe.
-/// https://curl.se/libcurl/c/threadsafe.html
-internal let isThreadSafeOpenSSL: Bool = {
+internal let usingThreadSafeSSLLibrary: Bool = {
     // curl_version_info is not thread-safe until curl_global_init has been called.
     // https://curl.se/libcurl/c/curl_version_info.html
     ensureLibcurlIsInitialized()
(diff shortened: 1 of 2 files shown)


clintonpi:
@swift-ci please test

clintonpi:
@swift-ci please test

itingliu:
@swift-ci please test windows

Sources/FoundationNetworking/URLSession/libcurl/libcurlHelpers.swift
```diff
@@ -16,16 +16,51 @@
...
+}()
+
+/// Initializes libcurl idempontently.
+internal func ensureLibcurlIsInitialized() {
+    _ = _initializeLibcurl // The lazy global is computed on only first access.
 }
 
+/// OpenSSL < 1.1.0 is not thread-safe.
+/// https://curl.se/libcurl/c/threadsafe.html
+internal let isThreadSafeOpenSSL: Bool = {
```
jrflat: Maybe we could rename this to be slightly more general like `usingThreadSafeSSLLibrary`? Right now the name is a bit confusing since `isThreadSafeOpenSSL` would return `true` for non-OpenSSL libraries.

Sources/FoundationNetworking/URLSession/libcurl/libcurlHelpers.swift
```diff
@@ -16,16 +16,51 @@
...
+    // https://curl.se/libcurl/c/curl_version_info.html
+    ensureLibcurlIsInitialized()
+
+    let version = CFURLSessionSSLVersionInfo();
+    guard version.isOpenSSL else { return true }
+
+    return (version.major == 1 && version.minor >= 1) || version.major > 1
+}()
+
+private let _openSSLOperationLock = NSLock()
```
jrflat: Could we try to use `Mutex` from `Synchronization` here?

jrflat (commented):
Overall I think the approach looks good! Left some minor comments on naming and `NSLock` usage.

clintonpi:
@swift-ci please test

clintonpi:

PR #5434, opened 2026-03-16
Add support for `wasm32-unknown-emscripten` platform

MaxDesiatov:
Now that both `swiftc` and `swift-frontend` support Emscripten target triple we can follow up on [the previously pitched Emscripte support](https://forums.swift.org/t/pitch-emscripten-target-support-for-swift/85310) in libraries like `swift-corelibs-foundation`.

### DIFF
--- a/Sources/CoreFoundation/include/CFTargetConditionals.h
+++ b/Sources/CoreFoundation/include/CFTargetConditionals.h
@@ -18,7 +18,10 @@
   
 */
 
-#if __has_include(<TargetConditionals.h>)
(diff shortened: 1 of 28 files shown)


Sources/CoreFoundation/include/CFTargetConditionals.h
```diff
@@ -18,7 +18,10 @@
   
 */
 
-#if __has_include(<TargetConditionals.h>)
+// When cross-compiling for Emscripten or WASI on a macOS host, the system
```
parkera, line 21: Please avoid manually wrapping comments in Foundation's sources, thanks.

PR #5442, opened 2026-03-27
build: add a shim for WASI detection on older CMake

compnerd:
WASI was introduced into CMake as a platform in 3.31. For older CMake versions, we add a shim to allow cleanly supporting this platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,13 @@ endif()
 project(Foundation
     LANGUAGES C Swift)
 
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

kateinoigakukun (approved):

itingliu:
The Linux failure seems related?
```
'Block.h' file not found
```

compnerd:
Yeah, it does, I'll get back to this shortly.

PR #5448, opened 2026-04-02
use default CURL_BUFFERSIZE (CURL_MAX_READ_SIZE typically 10MB).

toffaletti:
stop explicitly setting the receive buffer size so we get the default value which is safe and won't crash.

closes #5445

### Motivation:

The current code uses `Int.max` which is out of bounds for what libcurl allows so this call always crashes.

### Modifications:

Removed the calls to set the size.

### Result:

The default buffer size will be used (typically 10MB, but can be changed at configure time of building libcurl)

### Testing:

None yet, I need a refresher on how to test foundation networking builds.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/FTP/FTPURLProtocol.swift
+++ b/Sources/FoundationNetworking/URLSession/FTP/FTPURLProtocol.swift
@@ -70,7 +70,6 @@ internal class _FTPURLProtocol: _NativeProtocol {
             failWith(error: nsError, request: request)
             return
         }
-        easyHandle.set(preferredReceiveBufferSize: Int.max)
         do {
             switch (body, try body.getBodyLength()) {
             case (.none, _):
--- a/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
+++ b/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
@@ -329,7 +329,6 @@ internal class _HTTPURLProtocol: _NativeProtocol {
         let _config = session._configuration
         easyHandle.set(sessionConfig: _config)
         easyHandle.setAllowedProtocolsToHTTPAndHTTPS()
-        easyHandle.set(preferredReceiveBufferSize: Int.max)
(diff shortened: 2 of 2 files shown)


xbhatnag (approved):

itingliu:
@swift-ci please test

clintonpi:
For posterity's sake [it will be helpful if the title and description of this PR is adjusted], and in addition to @xbhatnag's [comment on the referenced issue](https://github.com/swiftlang/swift-corelibs-foundation/issues/5445#issuecomment-4201097995):
- The usage of `Int.max` has no effect on the crash.
- The default buffer size is CURL_MAX_WRITE_SIZE (16kB).

From my findings, the problem appears to be that for easy handles that libcurl **_had to queue_** [due to `CURLMOPT_MAX_HOST_CONNECTIONS` i.e. `URLSessionConfiguration.httpMaximumConnectionsPerHost` (default: 6)], libcurl does not free the buffer of those handles after calling `curl_multi_remove_handle`. I confirmed this by trying to set the buffer size just after removing a previously queued easy handle; `CURLE_BAD_FUNCTION_ARGUMENT` was returned.

- For libcurl < 8.7.0
  - When `CURL_BUFFERSIZE` is set,`CURLE_BAD_FUNCTION_ARGUMENT` is returned if an easy handle buffer is already/still allocated. A single `URLSessionTask` may do multiple, consecutive transfers, reusing easy handles (in this case, for redirects). If the reused handle was previously queued in libcurl, then its buffer would still be allocated.
  - [Just for example, setting the buffer size in `_EasyHandle.init(delegate:)` as opposed to in `_HTTPURLProtocol.configureEasyHandle(for:body:)` avoids the crash.]

- For libcurl >= 8.7.0
  - A single transfer buffer is allocated per multi handle (as opposed to per easy handle), so this crash _should_ not occur.
  - [The above example fix should stay appropriate for these libcurl versions too, since the same fixed value is being set and not a lesser value. Even better, the buffer size could be set idempotently in `_MultiHandle`, as the first easy handle is added to it.]

PR #5454, opened 2026-04-14
Cleanup swapped item after renameat2 call in FileManager

jmschonfeld:
Fixes an issue identified by test failures in Amazon Linux 2 CI in `FileManager`'s `replaceItem` API

### Motivation:

I believe the `renameat2` path wasn't actually taken in CI (based on the version of Linux used in CI) on the original PR that enabled `test_replacement`, but it is taken on Amazon Linux 2. In that environment, the test validates that the file at the path of the source file is non-existent after the replace call. However, `renameat2` performs a swap so this check failed.

### Modifications:

If we swap via `renameat2`, then we should cleanup the original file if it was swapped to the old location.

### Result:

`test_replacement` passes on AL2

### Testing:

Existing unit tests on AL2 cover this

### DIFF
--- a/Sources/Foundation/FileManager+POSIX.swift
+++ b/Sources/Foundation/FileManager+POSIX.swift
@@ -517,26 +517,28 @@ extension FileManager {
             
             #if canImport(Glibc)
             do {
-                let finalErrno = originalItemURL.withUnsafeFileSystemRepresentation { (originalFS) -> Int32? in
-                    return newItemURL.withUnsafeFileSystemRepresentation { (newItemFS) -> Int32? in
+                let (finalErrno, performedSwap) = originalItemURL.withUnsafeFileSystemRepresentation { (originalFS) -> (Int32?, Bool) in
+                    return newItemURL.withUnsafeFileSystemRepresentation { (newItemFS) -> (Int32?, Bool) in
                         if let originalFS = originalFS,
                            let newItemFS = newItemFS {
                                 #if os(Linux)
                                 if _CFHasRenameat2 && kernelSupportsRenameat2 {
+                                    // Use RENAME_EXCHANGE when possible because it can support more combinations of renames when the type of the item at the old location and the type of the pre-existing item at the new location don't match
+                                    // If the swap is successful, we will remove the item at originalFS (the pre-existing item) below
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@WFT this is a followup to https://github.com/swiftlang/swift-corelibs-foundation/pull/5443 to fix the newly enabled test on Amazon Linux 2 - does this change make sense to you?

Sources/Foundation/FileManager+POSIX.swift
```diff
@@ -550,6 +552,11 @@ extension FileManager {
                 if let finalErrno = finalErrno, !nonFatalErrors.contains(finalErrno) {
                     throw _NSErrorWithErrno(finalErrno, reading: false, url: originalItemURL)
                 } else if finalErrno == nil {
+                    if performedSwap {
+                        // If we performed an atomic swap instead of an atomic move, remove the file at the new item URL
+                        // If there was already a file at the destination of the move, it will now exist at the source of the move (when swapping) and we should clean that up
+                        try? removeItem(at: newItemURL)
```
WFT: I'm assuming this is discarding the error because the only error that could occur would be if the item had been removed by something else in the meantime?

WFT (approved):
Yep this LGTM. I had one question about discarding the error, but I assume the error isn't generally useful. 

It actually looks very close to a fix I had incorporated into a separate PR to [re-enable renameat2 more generally](https://github.com/swiftlang/swift-corelibs-foundation/pull/5455); I will rework that one before requesting your review.

Sources/Foundation/FileManager+POSIX.swift
jmschonfeld: Actually this is a good point, we probably do want to report the error from here. I originally chose to ignore it because I suspected that there might be errors we'd want to ignore such as if the directory weren't empty, but `removeItem(at:)` actually handles that and I can't think of any cases where a failure to remove this file would be expected. `applyPostprocessingRequiredByOptions` below this does propagate errors caused by removing the backup file, so we should probably be consistent. Test pass without discarding the error, so I'll update this

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

parkera (approved):

PR #5455, opened 2026-04-14
Actually use renameat2 for renaming when it's available

### LINKED ISSUE #5451: Enable `_CFHasRenameat2` on kernels which support that operation
Per the discussion in this comment thread ( https://github.com/swiftlang/swift-corelibs-foundation/pull/5443/changes#r3060431815 ) , `_CFHasRenameat2` is disabled even on Linux versions which have support. This is because an include is in the wrong place. This disables renameat2 support in FileManager, making `replaceItemAt` non-atomic. 

We should re-enable renameat2 support.

But there's a fair bit of complexity  because moving the include breaks other things unless we define `_GNU_SOURCE` everywhere. But that's a big change.

I'm willing to take a first stab at this.

WFT:
Correct definition if `_CFHasRenameat2` to actually detect the presence of renameat2 on Ubuntu, among others.

As discussed on https://github.com/swiftlang/swift-corelibs-foundation/pull/5443.

I figured this is a good fix without opening the bucket of worms which is defining `_GNU_SOURCE` for everything and fixing all the STATX issues.

My only real question about this code: is there CI coverage for both these paths or are all the CI runners Linux-with-sufficiently-new-kernel to use renameat2?

### Motivation:

As discussed in this comment thread ( https://github.com/swiftlang/swift-corelibs-foundation/pull/5443#discussion_r3012721709 ) renameat2 + `RENAME_EXCHANGE` gives us atomic file replacement, which is definitely nice to have.

### Modifications:

- Move `<sys/syscall.h>` include.
  - renameat2 support was all wired up, but we weren't using this code path on Ubuntu (at least) because `<sys/syscall.h>` wasn't included before the test for `SYS_renameat2`, causing `_CFHasRenameat2` to always be false. 
  - Fixing this causes some other STATX code to compile which didn't before. That code needs `_GNU_SOURCE` defined, so I've tweaked the compilation condition on that to specify the `_GNU_SOURCE` requirement.

### Result:

We now use renameat2 for atomic file replacement on Linux kernels where renameat2 is supported.

fixes https://github.com/swiftlang/swift-corelibs-foundation/issues/5451

### Testing:

I have tested this commit on top of #5443 before rebasing onto main using the `swift:6.3.0-noble` docker image. Unfortunately I haven't been able to test since rebasing because that Docker image can't build main anymore & neither can `swiftlang/swift:nightly-noble`.

### DIFF
--- a/Sources/CoreFoundation/include/ForSwiftFoundationOnly.h
+++ b/Sources/CoreFoundation/include/ForSwiftFoundationOnly.h
@@ -90,10 +90,12 @@
 #include <termios.h>
 #include <fcntl.h>
 
+// Required for statx + renameat2 wrappers
+#include <sys/syscall.h>
+
 #ifdef __GLIBC_PREREQ
 #if __GLIBC_PREREQ(2, 28) == 0
 // required for statx() system call, glibc >=2.28 wraps the kernel function
-#include <sys/syscall.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <linux/stat.h>
@@ -536,7 +538,7 @@ static inline unsigned int _dev_minor(dev_t rdev) {
 
 
 #if TARGET_OS_LINUX
-#ifdef __NR_statx
+#if defined(__NR_statx) && defined(_GNU_SOURCE)
 
 // There is no glibc statx() function, it must be called using syscall().
 


WFT:
@jmschonfeld This enables renameat2 on more systems, as discussed in https://github.com/swiftlang/swift-corelibs-foundation/pull/5443. Does this look reasonable to you?

jmschonfeld (approved):
I validated locally that when building this branch in the nightly-noble container that `_CFHasRenameat2` is `1` now when previously it was `0`. Thanks for the fix! I agree this seems like a good way to enable the `renameat2` call without worrying about changing the statx stuff now (theoretically we probably want to enable it if we can, but that's a separate can of worms due to the `_GNU_SOURCE` aspect as you pointed out)

jmschonfeld:
I also think there probably isn't much we can do here from a unit test perspective - not all of our CI environments will support the `renameat2` behavior, so I can't think of a reliable way to validate the behavior works without just checking gating the test itself on `_CFHasRenameat2` which kind of defeats the point

jmschonfeld:
@swift-ci please test

WFT:
If some CI jobs support renameat2 and some don’t, then that seems like good coverage no? Unless I’m misunderstanding your comment @jmschonfeld .

WFT:
Windows failure seems unrelated: 

`update-checkout failed, fix errors and try again`

jmschonfeld:
> If some CI jobs support renameat2 and some don’t, then that seems like good coverage no? Unless I’m misunderstanding your comment @jmschonfeld .

Sorry - to rephrase, yeah we have coverage of `renameat2` _when it is used_ based on the CI hosts today. We just don't have a unit test that validates that `renameat2` is used when it can be used (in other words, there is not a unit test that fails before your change but passes after your change). But I don't think there's really a unit test that we could write that would do so, so it seems fine to land this without a paired unit test (given that the functionality itself is already tested by `test_replacement`

jmschonfeld:

WFT:
Ah yeah makes sense

Issue #5456, opened 2026-04-15
Requests that require authentication never complete when no credentials are available


plemarquand:
If I make a network request to a server that presents an authentication challenge and I have no credentials saved for the request to use, the request waits indefinitely.

I've traced the root bug back to [attemptProceedingWithDefaultCredential](https://github.com/swiftlang/swift-corelibs-foundation/blob/0b23e798e921fd14ff59b37c9145021cd60e0aa9/Sources/FoundationNetworking/URLSession/URLSessionTask.swift#L1270) in URLSessionTask.swift.

```swift
        @Sendable func attemptProceedingWithDefaultCredential() {
            if let credential = challenge.proposedCredential {
                let last = task._protocolLock.performLocked { task._lastCredentialUsedFromStorageDuringAuthentication }
                
                if last?.credential != credential {
                    proceed(using: credential)
                } else {
                    task.cancel()
                }
            }
        }
```

When `credential` evaluates to `nil` the task is never continued or cancelled, pausing the request indefinitely.

plemarquand:
Originally I encountered this attempting to pull a binary artifact from an authenticated package registry in swift-package-manager. Initially I put up https://github.com/swiftlang/swift-package-manager/pull/9945 until @jakepetroules pushed me to dig deeper into the root cause.

Seems like this may be a pretty longstanding issue; I found this forum post from 2020 where the symptoms match what I've been seeing: [forums.swift.org/t/urlsession-works-on-macos-fail-on-linux/35903](https://forums.swift.org/t/urlsession-works-on-macos-fail-on-linux/35903/)

jmschonfeld:
The PRs to `main` and `release/6.3` have both been merged

PR #5457, opened 2026-04-15
Fix URLSessionTask hanging on 401 when no credentials are available

plemarquand:
When a server returns a 401 with a WWW-Authenticate header and the client has no credentials to offer, `attemptProceedingWithDefaultCredential` would do nothing, leaving the task suspended indefinitely. Pass the 401 response through as a normal completion instead. This produces an error that matches what macOS produces when making a request under the same conditions.

Also add a regression test covering this scenario.

Issue: #5456


plemarquand:
@swift-ci test

xbhatnag (approved):

parkera:
@swift-ci test

PR #5471, opened 2026-05-06
Update TimeZone names for issue 5463

### LINKED ISSUE #5463: `DateFormatter.string(from:)` crashes on Windows for specific Asian time zones
On Windows, `DateFormatter.string(from:)` traps the process with exit code `0xC000001D` (`STATUS_ILLEGAL_INSTRUCTION`, surfaced as POSIX exit 132 / `SIGILL`) whenever its `timeZone` is set to one of a small set of IANA identifiers that have sub-minute historical LMT offsets in the tz database:

- `Asia/Kolkata` (India, `+05:30`)
- `Asia/Kathmandu` (Nepal, `+05:45`)
- `Asia/Yangon` (Myanmar, `+06:30`)

`ISO8601DateFormatter.string(from:)` crashes identically for the same zones. All other `TimeZone` APIs work correctly (construction, `secondsFromGMT(for:)`, `abbreviation()`), as do all `Calendar` operations with the same zones. The crash is specific to the `DateFormatter` family.

**To Reproduce**

```swift
// Build: swiftc repro.swift -o repro.exe
// Run:   ./repro.exe
import Foundation

let tz = TimeZone(identifier: "Asia/Kolkata")!
print("offset:", tz.secondsFromGMT(for: Date()))      // +19800 — fine

let f = DateFormatter()
f.locale = Locale(identifier: "en_US")
f.timeZone = tz
f.timeStyle = .short
f.dateStyle = .none

print("about to format…")
print(f.string(from: Date()))   // never reached on Windows
```

On Windows, this prints:

```
offset: 19800
about to format…
```

then the process exits with code `0xC000001D` (POSIX `132`). On macOS, the same code prints, e.g. `4:09 PM` and exits cleanly.

**Expected behavior**
The formatter should emit a time string and exit cleanly, same as macOS (e.g. `4:09 PM` for `Asia/Kolkata` at the time of writing). Even if the Underlying tzdata had malformed / sub-minute transitions that the formatter couldn't handle. The correct failure mode would be to throw or return an empty/unexpected string — not kill the host process with SIGILL.

**Configuration:**
 - Swift Version: Swift 6.3 release (`swift-6.3-RELEASE`), target `aarch64-unknown-windows-msvc`.
 - OS: Windows
 - OS Version: Windows 11 Pro, build 26200.0

**Additional context**

**Zones that work:**

Other zones with fractional-hour offsets work correctly, which rules out "fractional offset" as the underlying cause:

- `Asia/Tehran` (`+03:30`) ✅
- `Asia/Kabul` (`+04:30`) ✅
- `Asia/Karachi` (`+05:00`) ✅
- `Asia/Dhaka` (`+06:00`) ✅
- `Australia/Adelaide` (`+09:30`) ✅

**Could the issue be sub-minute LMT history?**

Every affected zone has **sub-minute historical LMT offsets** recorded in the IANA tz database; working zones do not. For example, `Asia/Kolkata`'s tzdata entries include `−05:53:28` (HMT, to 1906) and `+05:21:10` (MMT, to 1941) — both measured down to the second. Adelaide, Tehran, Kabul, and Karachi's LMT entries are all measured in minutes only. I haven't confirmed this, but my guess is that `DateFormatter`'s ICU pattern-generator path walks the zone's transition records and trips when it encounters a transition whose `secondsFromGMT % 60 != 0`.

**Impact**

Any Swift-on-Windows application that formats a `Date` through `DateFormatter` with one of these three IANA identifiers as its timeZone dies with `SIGILL`, no stack trace, no recovery.

itingliu:
CFTimeZone hardcoded deprecated IANA identifiers (Asia/Calcutta,
Asia/Katmandu, Asia/Rangoon) in its Windows mapping and abbreviation
defaults. Modern tzdata uses the canonical names (Asia/Kolkata,
Asia/Kathmandu, Asia/Yangon), causing CFTimeZoneCreateWithName to
return nil and crash in NSTimeZone._cfObject.

Update mappings to canonical names, retain deprecated names in the
Olson→Windows direction for backward compatibility, and surface any
future mapping gaps as a preconditionFailure.

Previously the test crashed on Windows CI. With the fix the test no longer crashes.


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
Windows build failed with timeout

itingliu:
@swift-ci please test Windows

itingliu:
The unit test failed! OK I think I know how to fix this

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@jmschonfeld mentioned that as a follow up task, we should re-core `NSTimeZone` to use struct `TimeZone`, and `NSCalendar` to use `struct Calendar` in this repo, and remove CFCalendar and CFTimeZone

jmschonfeld (approved):

PR #5474, opened 2026-05-08
Disable OpenSSL atexit handler before CURL init

xbhatnag:
Attempt to dynamically link `OPENSSL_init_crypto` and invoke it with `OPENSSL_INIT_NO_ATEXIT` before libcurl init, preventing OpenSSL from registering an atexit handler.

### Motivation:

URLSession may flakily cause crashes at program exit, which has been recorded in these issues:
* https://github.com/swiftlang/swift-corelibs-foundation/issues/5469
* https://github.com/swiftlang/swift-package-manager/issues/8171

This happens because OpenSSL's atexit handler races with Swift's object cleanup on libdispatch threads. It is possible that OpenSSL could have been cleaned up before Swift cleaned up CURL handles, causing a double-free of SSL objects, leading to a crash.

**Note**: OpenSSL is already in the process of disabling the atexit handler by default from version 4.0 onwards: https://openssl-library.org/post/2026-03-10-remove-atexit/

### Modifications:

`CFURLSessionInit` will first attempt to dynamically link the symbol `OPENSSL_init_crypto` and invoke it with `OPENSSL_INIT_NO_ATEXIT`. Then it will call `curl_global_init(CURL_GLOBAL_SSL)`.

### Result:

This change will ensure that OpenSSL's atexit handler is never installed, so there is no race at program exit.

**Note**: Disabling OpenSSL cleanup on program exit will cause memory leak checks to fail. OpenSSL also acknowledges this, but says that those warnings should be suppressed because the memory that would have been cleaned up was only allocated by OpenSSL once, so it is not strictly speaking a leak.

### Testing:

Confirmed with this sample program running in a loop that the spurious crash no longer occurs after the fix:

```
import Foundation
import FoundationNetworking

@main
struct Example {
    static func main() async throws {
        let session = URLSession(configuration: .default)
        try await session.data(from: .init(string: "https://swift.org/index.html")!)
        session.finishTasksAndInvalidate()
    }
}
```

### DIFF
--- a/Sources/_CFURLSessionInterface/CFURLSessionInterface.c
+++ b/Sources/_CFURLSessionInterface/CFURLSessionInterface.c
@@ -22,6 +22,7 @@
 #include "CFURLSessionInterface.h"
 #include "CFString.h"
 #include <curl/curl.h>
+#include <dlfcn.h>
 
 #if !defined(LIBCURL_VERSION_MAJOR)
 #error "LIBCURL_VERSION_MAJOR not defined, missing curlver.h"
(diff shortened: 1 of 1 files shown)


Sources/_CFURLSessionInterface/CFURLSessionInterface.c
```diff
@@ -22,6 +22,7 @@
 #include "CFURLSessionInterface.h"
 #include "CFString.h"
 #include <curl/curl.h>
+#include <dlfcn.h>
```
jrflat, line 27: I don't think this will compile on Windows. We'll need to condition on e.g. `#if !defined(_WIN32)` for this include and `dlsym` usage.

For Windows, we could `#include <windows.h>` and use some form of `GetModuleHandleW` and `GetProcAddress` against the common libcrypto DLL names.

jrflat (commented):
Excited to see this worked! Left a comment re: building on Windows

Sources/_CFURLSessionInterface/CFURLSessionInterface.c
jrflat: Since the crashes are reported on Linux I think it's totally fine to just add the compile guards for this PR and investigate Windows parity later

xbhatnag: Added a TODO for Windows.

jrflat:
@swift-ci please test

jrflat:

xbhatnag:

parkera (approved):

jrflat (approved):

PR #5475, opened 2026-05-08
[wasm] Fix unavailable calls in NSString write methods on WASI

tshortli:
Apply the same `#if os(WASI)` guard that was already applied to `_writeTo`'s body to `write(to:)` and `write(toFile:)`.

### Motivation:

`write(to:)` and `write(toFile:)` call `_writeTo`, which is marked `@available(*, unavailable)` on WASI. An upcoming Swift compiler fix (https://github.com/swiftlang/swift/pull/88942) corrects handling of calls to universally unavailable members of extensions, so these now need the same `#if os(WASI)` guard that was already applied to `_writeTo`'s body.

### Modifications:

Throw `.featureUnsupported` errors directly from `write(to:)` and `write(toFile:)` on WASI.

### Result:

The package builds with a Swift compiler that has https://github.com/swiftlang/swift/pull/88942.

### Testing:

Cross repo testing initiated from https://github.com/swiftlang/swift.

### DIFF
--- a/Sources/Foundation/NSString.swift
+++ b/Sources/Foundation/NSString.swift
@@ -1283,14 +1283,22 @@ extension NSString {
     @available(*, unavailable, message: "WASI does not support atomic file-writing as it does not have temporary directories")
     #endif
     public func write(to url: URL, atomically useAuxiliaryFile: Bool, encoding enc: UInt) throws {
+        #if os(WASI)
+        throw CocoaError(.featureUnsupported)
+        #else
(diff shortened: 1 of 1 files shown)


tshortli:
@swift-ci please test

jmschonfeld (approved):
I have a few questions about the approach in general (should it `fatalError` instead of `throw`, and should we only do so when `useAuxiliaryFile` is `true`) but those are more about the pre-existing annotations and this seems fine to me to keep it in-line with what `_writeTo` already does

PR #5481, opened 2026-05-20
Fix XMLNode name/stringValue/prefix for DTD and namespace nodes

itingliu:
Three bugs in XMLNode.swift:

1. name setter: .entityDeclaration, .attributeDeclaration, and .elementDeclaration all fell through to the default case which calls _CFXMLNodeSetName. libxml2 silently ignores that call for DTD nodes, so renames were lost. Extended the existing .notationDeclaration case (which already used _CFXMLNodeForceSetName) to cover all four DTD declaration kinds.

2. stringValue setter: The default path runs _CFXMLEncodeEntities(document, string) before storing the value. For a standalone XMLDTDNode with no owning document, _CFXMLNodeGetDocument returns nil, _CFXMLEncodeEntities(nil, …) returns nil, and the fallback "" is written — erasing the value. Element declaration content is a DTD validation string (e.g. "(#PCDATA | array)*"), not XML text that needs entity escaping, so it now goes directly to _CFXMLNodeSetContent.

3. prefix getter: _CFXMLNodeCopyPrefix returns nil for namespace nodes because libxml2 stores the namespace prefix in a different slot. Added a .namespace case that reads it via _CFXMLNamespaceCopyPrefix, mirroring the pattern already used by the name getter.

Fixes test_nodeNames (entityDecl/attrDecl/elementDecl rename), test_dtd (elementDecl name + stringValue on standalone node), and test_createElement (namespace.prefix).

### DIFF
--- a/Sources/FoundationXML/XMLNode.swift
+++ b/Sources/FoundationXML/XMLNode.swift
@@ -351,9 +351,9 @@ open class XMLNode: NSObject, NSCopying {
             case .document:
                 // As with Darwin, ignore the name when the node is document.
                 break
-            case .notationDeclaration:
+            case .notationDeclaration, .entityDeclaration, .attributeDeclaration, .elementDeclaration:
                 // Use _CFXMLNodeForceSetName because
-                // _CFXMLNodeSetName ignores the new name when the node is notation declaration.
+                // _CFXMLNodeSetName ignores the new name for DTD declaration nodes.
                 _CFXMLNodeForceSetName(_xmlNode, newValue)
             case .namespace:
                 _CFXMLNamespaceSetPrefix(_xmlNode, newValue, Int64(newValue?.utf8.count ?? 0))
@@ -429,7 +429,13 @@ open class XMLNode: NSObject, NSCopying {
                 
             case .comment, .text:
                 _CFXMLNodeSetContent(_xmlNode, newValue)
-                
+
+            case .elementDeclaration:
+                // Element declaration content is a validation string (e.g. "(#PCDATA | array)*"),
+                // not XML text — entity-encoding it would corrupt it, and _CFXMLEncodeEntities
+                // returns nil when there is no owning document, which would wipe the value.
+                _CFXMLNodeSetContent(_xmlNode, newValue)
+
             default:
                 _removeAllChildNodesExceptAttributes() // in case anyone is holding a reference to any of these children we're about to destroy
                 if let string = newValue {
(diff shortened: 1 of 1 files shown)


Sources/FoundationXML/XMLNode.swift
```diff
@@ -695,8 +701,13 @@ open class XMLNode: NSObject, NSCopying {
      @abstract Returns the prefix foo if this attribute or element's name if foo:bar
      */
     open var prefix: String? {
-        let returned = _CFXMLNodeCopyPrefix(_xmlNode)
-        return returned == nil ? nil : unsafeBitCast(returned!, to: NSString.self) as String
+        switch kind {
+        case .namespace:
+            return _CFXMLNamespaceCopyPrefix(_xmlNode).map({ unsafeBitCast($0, to: NSString.self) as String }) ?? ""
```
jmschonfeld, line 706: Is there a reason that this returns an empty string when it fails instead of `nil` like the `else` path?

itingliu: This is `_CFXMLNamespaceCopyPrefix`, while the default case is `_CFXMLNodeCopyPrefix`. 

In the `.namespace` case, `_CFXMLNamespaceCopyPrefix` returns nil if an xml node is using the **default** namespace. But with `_CFXMLNodeCopyPrefix`, nil means it genuinely **doesn't have** any prefix

See comment on line 973: 
```
// Default namespace is represented by `NULL` in libxml2.
```

This is also consistent with other places where we use `_CFXMLNamespaceCopyPrefix`. For example on line 344 above, we're also doing it this way

```
case .namespace:
                return _CFXMLNamespaceCopyPrefix(_xmlNode).map({ unsafeBitCast($0, to: NSString.self) as String }) ?? ""
```

itingliu:
@swift-ci please test

Sources/FoundationXML/XMLNode.swift
jmschonfeld: Ok cool thanks for the explanation

jmschonfeld (approved):

jmschonfeld:
LGTM assuming tests pass on both the existing libxml versions and the new version

itingliu:
@swift-ci please test

itingliu:
Windows CI failed with 
```
error: you need to resolve your current index first
.
update-checkout failed, fix errors and try again
```

itingliu:
Ubuntu 26.04 succeeded https://ci.swift.org/job/swift-PR-toolchain-ubuntu2604/2/console

itingliu:
@swift-ci please test Windows

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test Windows

itingliu:
@swift-ci please test Windows

PR #5482, opened 2026-05-21
Fix variable used in updateProgress in URLSessionTask

valerianb:
The wrong variable was used to populate `toBeReceived`


parkera:
Can we add a unit test to this one too? Thanks!

xbhatnag:
I agree with @parkera , this should be possible to unit-test. I imagine it would require crafting a request with `Content-Length` header in response such that `URLSessionTask.progress.totalUnitCount` gets set correctly.

valerianb:
I was on leave, I will try to give it a go soon.

PR #5483, opened 2026-05-21
Fix potential oob read in sendWebSocketsData

valerianb:
Underneath this calls `curl_ws_send` and from the documentation:

> sent is set to the number of payload bytes actually sent. If the return value is [CURLE_OK](https://curl.se/libcurl/c/libcurl-errors.html#CURLEOK) but sent is less than the given buflen, libcurl was unable to consume the complete payload in a single call. In this case the application must call this function again until all payload is processed. buffer and buflen must be updated on every following invocation to only point to the remaining piece of the payload.

But in this loop, the `buflen` isn't updated and this can result in an out of bounds read if curl wasn't able to consume the whole payload in a single call.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/libcurl/EasyHandle.swift
+++ b/Sources/FoundationNetworking/URLSession/libcurl/EasyHandle.swift
@@ -426,7 +426,7 @@ extension _EasyHandle {
                 var offset = 0
                 repeat {
                     var amountWritten = 0
-                    try CFURLSessionEasyHandleWebSocketsSend(rawHandle, bytesPtr.advanced(by: offset), data.count, &amountWritten, 0, cfurlSessionFlags).asError()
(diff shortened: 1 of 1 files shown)


parkera:
Can we add a new test case?

valerianb:
Hi,

I believe sending data is already tested https://github.com/swiftlang/swift-corelibs-foundation/blob/8d66e4f0070a2f3e03195beda46a267255406563/Tests/Foundation/TestURLSession.swift#L2159

I am unsure how I can trigger multiple `curl_ws_send` in the loop reliably ?
Especially as according to the comment in the tests (`the test HTTPServer implementation is barebones, and can't handle receiving more than one frame at a time`), it seems we are restricted to sending a small payload.

Would appreciate any guidance as I'm not very familiar with the project.

xbhatnag:
It's quite hard to convince an operating system to reliably create artificial socket buffer conditions. Unless someone can suggest a way to do so, I think this change is fine.

xbhatnag (approved):

xbhatnag:
@swift-ci please test

xbhatnag:
@swift-ci please test

PR #5484, opened 2026-05-22
Remove explicit `Connection: keep-alive` header set

xbhatnag:
Setting this header explicitly causes issues for WebSockets which need `Connection: upgrade`.

curl is capable of managing the `Connection` header by itself and we don't need to specify this header, which is implicitly assumed for HTTP/1.1 and not needed for HTTP/2 or HTTP/3

Tested that new versions of curl which support websockets don't fail the URLSession websocket tests with this change.

### DIFF
--- a/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
+++ b/Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
@@ -662,9 +662,7 @@ internal class _HTTPURLProtocol: _NativeProtocol {
     /// These will only be set if not already part of the request.
     /// - SeeAlso: https://curl.haxx.se/libcurl/c/CURLOPT_HTTPHEADER.html
     var curlHeadersToSet: [(String,String)] {
-        var result = [("Connection", "keep-alive"),
-                      ("User-Agent", userAgentString),
-                      ]
+        var result = [("User-Agent", userAgentString)]
         if let language = NSLocale.current.languageCode {
             result.append(("Accept-Language", language))
         }


Sources/FoundationNetworking/URLSession/HTTP/HTTPURLProtocol.swift
```diff
@@ -662,9 +662,7 @@ internal class _HTTPURLProtocol: _NativeProtocol {
     /// These will only be set if not already part of the request.
     /// - SeeAlso: https://curl.haxx.se/libcurl/c/CURLOPT_HTTPHEADER.html
     var curlHeadersToSet: [(String,String)] {
-        var result = [("Connection", "keep-alive"),
```
jmschonfeld, line 665: In addition to testing with new versions of curl, just confirming, does this also work with older versions of curl that might exist on older Linux versions that we still support?

xbhatnag: Good point!

Tested on container `swift:bookworm` (swift 6.3.2 with curl 7.88.1).

All `TestURLSession*` tests pass.

[bookworm_curl_old.txt](https://github.com/user-attachments/files/28159674/bookworm_curl_old.txt)

jmschonfeld: It would probably be good to validate this on even older versions of curl. For example we still ship a Swift toolchain for Ubuntu 20.04 which I believe ships with curl 7.68.0.

Change seems fine to me if it works the same on older curl versions like that one, but if this change breaks with those older curl versions we may need to conditionalize it

xbhatnag: Tested Ubuntu 22.04 LTS as well: `swift:jammy` (swift 6.3.2 with curl 7.81.0)

All `TestURLSession*` tests pass.

[jammy_curl_old.txt](https://github.com/user-attachments/files/28159886/jammy_curl_old.txt)

xbhatnag: Happy to let the Swift CI test all the rest.

xbhatnag: Just saw your comment for 20.04. Do we have a docker container for that with 6.3.2 or later?

jmschonfeld: Looks like my info on 20.04 might be outdated, so given that you tested against 22.04 seems like we've validated it across a wide enough range that hopefully should be ok

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #5486, opened 2026-05-28
NSNull: fix === returning false when both operands are nil

### LINKED ISSUE #5248: Inconsistencies with NSNull and `===`
```swift
  1> import Foundation
  2> let x: NSNull? = nil
x: Foundation.NSNull? = nil
  3> x === x
$R0: Bool = false
  4> x !== x
$R1: Bool = false
  5> NSNull() === NSNull()
$R2: Bool = true
  6> NSNull() !== NSNull()
$R3: Bool = true
```

ayush-that:
Fixes #5248.

The hack in `===(NSNull?, NSNull?)` returns false when either operand is nil, so `nil === nil` returns false. Replace the guard with an explicit switch on the optional pair: `(nil, nil)` is true, mixed-nil is false, and two non-nil NSNull instances remain identical (preserving the singleton behavior).

Also add a matching `!==` overload so the symmetric operator agrees with `===` for this signature. Without it, Swift's default `!==` for `AnyObject?` returns true for two distinct NSNull instances even though the custom `===` says they are identical.

The new operator is a small public-API addition. Happy to drop it and land only the bug fix if a swift-evolution discussion is preferred.

### Tests

Added `test_identityOperator()` covering the full truth table (nil/nil, mixed, non-nil/non-nil for both `===` and `!==`). Existing `test_alwaysEqual()` still passes; added one supplemental `!==` assertion to it.

Verified on Linux via `swiftlang/swift:nightly-main-jammy`:

```
Test Suite 'TestNSNull' passed
   Executed 3 tests, with 0 failures (0 unexpected) in 0.106s
```

### DIFF
--- a/Sources/Foundation/NSNull.swift
+++ b/Sources/Foundation/NSNull.swift
@@ -44,6 +44,16 @@ open class NSNull : NSObject, NSCopying, NSSecureCoding, @unchecked Sendable {
 }
 
 public func ===(lhs: NSNull?, rhs: NSNull?) -> Bool {
-    guard let _ = lhs, let _ = rhs else { return false }
-    return true
+    switch (lhs, rhs) {
+    case (nil, nil):
+        return true
+    case (nil, _), (_, nil):
+        return false
+    default:
+        return true
+    }
+}
+
+public func !==(lhs: NSNull?, rhs: NSNull?) -> Bool {
+    return !(lhs === rhs)
 }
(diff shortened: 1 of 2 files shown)


parkera:
In Objective-C, NSNull's allocator will always return the singleton. There is no way to get two distinct instances of `NSNull`.

parkera (approved):

parkera:
@swift-ci test

Issue #5487, opened 2026-05-30
NSAttributedString should conform to CVarArg on non-Darwin Foundation


Kyle-Ye:
## Background

On Linux / swift-corelibs-foundation, `NSAttributedString` does not conform to `CVarArg`, so it cannot be used with `%@` formatting APIs even though it is a CF-backed Foundation object.

For example:

```swift
import Foundation

let value = NSAttributedString(string: "hello")
let result = NSString(format: "%@", value)
print(result)
```

Expected behavior:

This should compile and format the attributed string using `%@`, matching Darwin Foundation behavior for this Foundation object.

Actual behavior on Linux:

```text
error: argument type 'NSAttributedString' does not conform to expected type 'CVarArg'
```

There has been previous work in this area:

- Swift Forums: [CVarArg support for NSString and String](https://forums.swift.org/t/cvararg-support-for-nsstring-and-string/37404)
- swift-corelibs-foundation PR: [#2821 Added CVarArg support to NSString and String](https://github.com/swiftlang/swift-corelibs-foundation/pull/2821)
- Swift PR: [swiftlang/swift#32311 Added protocol to support CVarArg objects that need to be retained](https://github.com/swiftlang/swift/pull/32311)

That fixed `NSString` / `String`, but `NSAttributedString` still appears to be missing.

~~That fixed `NSString` / `String`, but the broader `NSObject` subclass case still appears to be missing.~~

~~A possible implementation is to mirror the object CVarArg encoding used by Apple's open-source Objective-C runtime support by making `NSObject` itself conform to `CVarArg`:~~

```swift
extension NSObject: CVarArg {
    public var _cVarArgEncoding: [Int] {
        _autorelease(self)
        return _encodeBitsAsWords(self)
    }
}
```

~~Inside swift-corelibs-foundation, this would also make the existing explicit `NSString: CVarArg` conformance redundant, since `NSString` inherits from `NSObject`.~~

~~This would allow `NSAttributedString` and other `NSObject` subclasses to be passed as `%@` arguments consistently on non-Darwin Foundation.~~

## Workaround

~~My current workaround was adding the following code to my library from apple-oss-distributions/objc4:~~

```swift
// Non-Darwin compatibility layer
#if !canImport(Darwin)
// Linux Foundation does not provide libswiftObjectiveC's NSObject CVarArg conformance yet.
extension NSObject: @retroactive CVarArg {
    public var _cVarArgEncoding: [Int] {
        _autorelease(self)
[...]

PR #5488, opened 2026-05-31
Make NSAttributedString conform to CVarArg

Kyle-Ye:
### Motivation:

Related to #5487.

`NSAttributedString` is CF-backed in swift-corelibs-foundation, but it does not currently conform to `CVarArg` on non-Darwin platforms. That prevents it from being passed to `%@` formatting APIs.

For example:

```swift
let value = NSAttributedString(string: "hello")
let result = NSString(format: "%@", value)
```

### Modifications:

Make `NSAttributedString` conform to `CVarArg` using the same object word encoding used by other Foundation object varargs. This keeps the conformance scoped to a CF-compatible Foundation type instead of applying it to all `NSObject` subclasses.

### Result:

The example above compiles on non-Darwin platforms and formats through the existing CFAttributedString description path.

### Testing:

- Added `test_initializeWithFormatNSAttributedStringCVarArg`
- Verified a Linux overlay of the same conformance formats `NSAttributedString` through `%@`
- Verified a plain Swift `NSObject` subclass remains rejected as `CVarArg`

### DIFF
--- a/Sources/Foundation/NSAttributedString.swift
+++ b/Sources/Foundation/NSAttributedString.swift
@@ -422,6 +422,13 @@ extension NSAttributedString {
     internal var _cfObject: CFAttributedString { return unsafeBitCast(self, to: CFAttributedString.self) }
 }
 
+extension NSAttributedString : CVarArg {
+    @inlinable // c-abi
+    public var _cVarArgEncoding: [Int] {
+        return _encodeBitsAsWords(self)
+    }
+}
+
 extension NSAttributedString {
 
     public struct EnumerationOptions: OptionSet, Sendable {
--- a/Tests/Foundation/TestNSString.swift
+++ b/Tests/Foundation/TestNSString.swift
@@ -862,6 +862,13 @@ class TestNSString: LoopbackServerTest {
         }
     }
 
+    func test_initializeWithFormatNSAttributedStringCVarArg() {
+        let value = NSAttributedString(string: "hello")
+        let result = NSString(format: "%@", value)
+        XCTAssertTrue(result.hasPrefix("hello "))
+        XCTAssertTrue(result.hasSuffix("{} Len 5\n"))
+    }
+
     func test_appendingPathComponent() {
         do {
             let path: NSString = "/tmp"


Kyle-Ye:
@swift-ci please test

Sources/Foundation/NSAttributedString.swift
```diff
@@ -422,6 +422,13 @@ extension NSAttributedString {
     internal var _cfObject: CFAttributedString { return unsafeBitCast(self, to: CFAttributedString.self) }
 }
 
+extension NSAttributedString : CVarArg {
```
jmschonfeld, line 425: I think just conforming `NSAttributedString` to `CVarArg` seems like too narrow of a solution here - there are plenty of other types that clients would likely want to interpolate into format strings. I think the right solution is conforming `NSObject` to `CVarArg` like it does on `Darwin`.

You did note that this isn't safe yet because Foundation doesn't provide callbacks to CF to correctly call `-description` on objects. Have you investigated providing those callbacks? This is what Foundation on Darwin does - it provides callbacks that call `-description` or `-descriptionWithLocale:` on objects

Kyle-Ye: >  I think the right solution is conforming NSObject to CVarArg like it does on Darwin.

The issue and PR is originally target at NSObject's level fix. But I changed my mind after digging into the detail.

Check 5487 body's update part.

> This is what Foundation on Darwin does - it provides callbacks that call -description or -descriptionWithLocale: on objects

We can consider do it if we agree. The current impl detail is it depends on _CFInfo to get CF layout and call CF's description callback.

If we want to add `NSObject` to `CVarArg`, a more complicated PR is required to somehow make NSObject's description accessible at C level.

Kyle-Ye: My ideal is we can merge this at it is and do a follow up PR to refactor the NSObject's description system on non-Darwin platform if we agree.

> Have you investigated providing those callbacks

I'll try open the new draft PR with the idea this weekend.

Kyle-Ye: Opened the follow-up draft PR here: https://github.com/swiftlang/swift-corelibs-foundation/pull/5490  @jmschonfeld. (If you agree this direction, I can close this and let's continue the review on that PR)

I tried the callback approach suggested above. The new draft makes `NSObject` conform to `CVarArg`, but also changes `NSString(format:)` to call `_CFStringCreateWithFormatAndArgumentsAux` with a Foundation `copyDescFunc`.

That means in the `%@` path, CoreFoundation calls back into Foundation before falling back to `__CFCopyFormattingDescription` / `CFCopyDescription`. For ordinary Swift `NSObject` subclasses, the callback consumes `object.description`, so they are not interpreted as CF runtime objects. (Or maybe we should try CF first?)

The callback also routes known locale-aware Foundation types through their existing description APIs (`NSArray`, `NSDictionary`, `NSSet`, `NSOrderedSet`, `NSNumber`, `NSDate`). `NSAttributedString` then works through the general `NSObject` path and produces the Swift/Foundation description instead of the CF one.

jmschonfeld: Thanks for opening the other PR, I think that's a better approach that we should aim for - will leave a few more detailed comments there but if you're ok with moving forward with that approach that sounds great to me!

Kyle-Ye: Got it. Let's continue the review conversation in that PR.

Kyle-Ye:
Close in favor of #5490

PR #5489, opened 2026-06-06
Reduce Unsafe Pointer Usage in Process argv and envp

juozasvalancius:
Modified the implementation of constructing argv and envp arrays for `posix_spawn`, to be more readable and use less unsafe pointers.

### Motivation:

Recently published [swift-foundation Contribution Guidelines](https://github.com/swiftlang/swift-foundation/blob/3019981601232666270dd250f2a1f4e496d6dec7/CONTRIBUTION_GUIDELINE.md) suggest avoiding unsafe APIs.

### Modifications:

`posix_spawn` needs an array of C-strings. We can't avoid using unsafe APIs for the C-strings themselves, but the array can be a regular Swift array, and Swift supports passing that into C. We just have to make sure our array ends with a null-pointer.

### Result:

No functional changes.

### Testing:

I've run the tests on Ubuntu 24.04 arm64.

It would be good to also run the tests on Android (I don't have one).

### DIFF
--- a/Sources/Foundation/Process.swift
+++ b/Sources/Foundation/Process.swift
@@ -744,45 +744,19 @@ open class Process: NSObject, @unchecked Sendable {
                 throw _NSErrorWithErrno(errno, reading: true, path: launchPath)
             }
         })
+
         // Convert the arguments array into a posix_spawn-friendly format
-        
-        var args = [launchPath]
-        if let arguments = self.arguments {
-            args.append(contentsOf: arguments)
-        }
-        
-        let argv : UnsafeMutablePointer<UnsafeMutablePointer<Int8>?> = args.withUnsafeBufferPointer {
-            let array : UnsafeBufferPointer<String> = $0
-            let buffer = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: array.count + 1)
-            buffer.initialize(from: array.map { $0.withCString(strdup) }, count: array.count)
-            buffer[array.count] = nil
-            return buffer
-        }
-        
-        defer {
-            for arg in argv ..< argv + args.count {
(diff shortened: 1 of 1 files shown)


Sources/Foundation/Process.swift
```diff
@@ -744,45 +744,19 @@ open class Process: NSObject, @unchecked Sendable {
...
 
-        var env: [String: String]
-        if let e = environment {
-            env = e
-        } else {
-            env = ProcessInfo.processInfo.environment
-        }
+        // Convert the arguments array into a posix_spawn-friendly format
+        let args: [String] = [launchPath] + (self.arguments ?? [])
+        var argv: [UnsafeMutablePointer<Int8>?] = args.map { strdup($0) }
```
itingliu, line 750: since we're already here, do you want to use this chance to switch to `withTemporaryAllocation` to reduce heap allocated memory? https://github.com/swiftlang/swift/pull/85866/changes

juozasvalancius: Do you mean for the value of `argv` that contains the list? I'm not sure if that's necessary, because the whole `run()` function already results in a lot of allocations; and even in this line, a `strdup` for every argument results in a bunch of allocations.

Since we have `argv` and `envp`, we would need two nested calls to `withTemporaryAllocation`, and then the actual usage of those values is 200 lines below, inside yet another closure for `_fileSystemRepresentation`.

I was also thinking of another optimization: create a single flat buffer that would contain all the strings. However, that would mean playing with unsafe pointers again, and reduced readability of the code.

So in the end, I feel like a simple array map is the best balance.

iCharlesHu (approved):

iCharlesHu:
@swift-ci please test

iCharlesHu:
@swift-ci please test

PR #5490, opened 2026-06-07
Make NSObject conform to CVarArg

Kyle-Ye:
## Summary

- Make `NSObject` conform to `CVarArg` so Foundation objects can be passed to `%@` formatting on non-Darwin platforms.
- Route `NSString(format:)` through `_CFStringCreateWithFormatAndArgumentsAux` with a Foundation description callback, matching the CoreFoundation hook intended for `descriptionWithLocale:` / `description`.
- Remove the narrower `NSAttributedString`-only conformance and cover ordinary `NSObject`, `NSAttributedString`, and locale-aware formatting in `TestNSString`.

## Details

`NSObject: CVarArg` only supplies the object pointer to the varargs list. The important behavioral change is that `NSString(format:)` now provides CoreFoundation with a `copyDescFunc`, so `%@` formatting asks Foundation for an object description instead of falling back to treating every object pointer as a CF runtime object.

For known Foundation collection/number/date types, the callback uses the existing locale-aware description entry points. Other `NSObject` subclasses use `object.description`.

### DIFF
--- a/Sources/Foundation/NSString.swift
+++ b/Sources/Foundation/NSString.swift
@@ -1315,32 +1315,18 @@ extension NSString {
     }
     
     public convenience init(format: String, arguments argList: CVaListPointer) {
-        let str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, nil, format._cfObject, argList)!
+        let str = _NSCreateStringWithFormatAndArguments(format._cfObject, nil, argList)
         self.init(str._swiftObject)
     }
     
     public convenience init(format: String, locale: AnyObject?, arguments argList: CVaListPointer) {
-        let str: CFString
-        if let loc = locale {
-            if type(of: loc) === NSLocale.self {
-                // Create a CFLocaleRef
-                let cf = (loc as! NSLocale)._cfObject
-                str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, unsafeBitCast(cf, to: CFDictionary.self), format._cfObject, argList)
-            } else if type(of: loc) === NSDictionary.self {
-                let dict = (loc as! NSDictionary)._cfObject
-                str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, dict, format._cfObject, argList)
-            } else {
-                fatalError("locale parameter must be a NSLocale or a NSDictionary")
-            }
-        } else {
-            str = CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, nil, format._cfObject, argList)
-        }
+        let str = _NSCreateStringWithFormatAndArguments(format._cfObject, locale, argList)
         self.init(str._swiftObject)
     }
-    
+
     public convenience init(format: NSString, _ args: CVarArg...) {
         let str = withVaList(args) { (vaPtr) -> CFString? in
-            CFStringCreateWithFormatAndArguments(kCFAllocatorSystemDefault, nil, format._cfObject, vaPtr)
+            _NSCreateStringWithFormatAndArguments(format._cfObject, nil, vaPtr)
         }!
         self.init(str._swiftObject)
     }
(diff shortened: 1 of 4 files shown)


Sources/Foundation/NSString.swift
```diff
@@ -1478,6 +1464,66 @@ extension NSString {
...
+        formatOptions = nil
+    }
+
+    return _CFStringCreateWithFormatAndArgumentsAux(kCFAllocatorSystemDefault, _NSCopyFormattingDescription, formatOptions, format, arguments)
+}
+
+private func _NSCopyFormattingDescription(_ object: UnsafeMutableRawPointer, _ formatOptions: UnsafeRawPointer) -> Unmanaged<CFString> {
+    let value = Unmanaged<AnyObject>.fromOpaque(object).takeUnretainedValue()
+    if let nsObject = value as? NSObject {
+        return Unmanaged.passRetained(_NSFormattingDescription(nsObject, locale: _NSFormattingLocale(formatOptions))._cfObject)
```
jmschonfeld: Instead of the `as? NSObject` and various `as NSX` casts, instead can we perform a selector check here? Ideally this code would look something like the following pseudocode:

```
if value responds to selector "descriptionWithLocale:" {
    value.description(withLocale: formatOptions)
} else if value responds to selector "description" {
    value.description
} else {
    "*nil description*"
}
```

Sources/Foundation/NSString.swift
```diff
@@ -1478,6 +1464,66 @@ extension NSString {
...
+    }
+
+    let cfObject = unsafeBitCast(object, to: CFTypeRef.self)
+    return Unmanaged.passRetained(CFCopyDescription(cfObject))
```
jmschonfeld: I think we can remove this path and just assume everything here will be an `NSObject`. `CFTypeRef`s don't conform to `CVarArg` so we know one can never be provided as an argument when using the NSString formatting functions (and CF functions that do accept the CF types will still work because this up-call isn't provided)

Tests/Foundation/TestNSString.swift
```diff
@@ -862,6 +867,26 @@ class TestNSString: LoopbackServerTest {
         }
     }
 
+    func test_initializeWithFormatNSAttributedStringCVarArg() {
+        let value = NSAttributedString(string: "hello")
+        let result = NSString(format: "%@", value)
+        XCTAssertEqual(result, "hello{\n}")
+    }
+
+    func test_initializeWithFormatNSObjectCVarArg() {
```
jmschonfeld, line 876: Would you mind adding a test here that passes `nil` as an object just to make sure optional handling works ok? I believe it does, but I don't know if we have a test for it yet.

Sources/Foundation/NSString.swift
jmschonfeld: Oh, I suppose without the Objective-C runtime, do we have a way to do a responds-to-selector check? Perhaps not...

Tests/Foundation/TestNSString.swift
Kyle-Ye: No problem

Sources/Foundation/NSString.swift
Kyle-Ye: Got it.

Sources/Foundation/NSString.swift
Kyle-Ye: Yeah. That's issue here. Unless we add a new internal protocol and make those types conform to the new protocol we added. But the code is not less than the current status.

Tests/Foundation/TestNSString.swift
Kyle-Ye: `NSObject?` does not conform to CVarArg too on macOS. Do you mean something like this?

```swift
func test_initializeWithFormatNilObjectCVarArg() {
    let value = unsafeBitCast(0, to: NSObject.self)
    let result = NSString(format: "%@", value)
    XCTAssertEqual(result, "(null)")
}
```

Sources/Foundation/NSString.swift
jmschonfeld: Ok then this seems fine to me

Tests/Foundation/TestNSString.swift
jmschonfeld: Oh you're right! From ObjC/C callers you can provide null objects, but Swift prevents it since `Optional` isn't a `CVarArg`, even on macOS. Ok in that case no need to test it since it's not valid to do anyways

jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
Sorry I didn't catch that you had committed the `test_initializeWithFormatNilObjectCVarArg` - looks like the CI failures are due to that test

Tests/Foundation/TestNSString.swift
Kyle-Ye: Yes. I'll remove this and trigger test again.

jmschonfeld:
@swift-ci please test

jmschonfeld:

jmschonfeld (approved):

Kyle-Ye:

Issue #5203, opened 2025-04-19
Crash when making a network request on Android: _MultiHandle deallocated with non-zero retain count


marcprux:
```
2025-04-19 13:04:34.678  6468-6518  SwiftRuntime            com.mediquo.main                     A  Object 0xb400007ab8eaf830 of class _MultiHandle deallocated with non-zero retain count 2. This object's deinit, or something called from it, may have created a strong reference to self which outlived deinit, resulting in a dangling reference.
2025-04-19 13:04:34.680  6468-6518  libc                    com.mediquo.main                     A  Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 6518 (DefaultDispatch), pid 6468 (om.mediquo.main)
---------------------------- PROCESS STARTED (6544) for package com.mediquo.main ----------------------------
2025-04-19 13:04:35.127  6542-6542  DEBUG                   crash_dump64                         A  Cmdline: com.mediquo.main
2025-04-19 13:04:35.127  6542-6542  DEBUG                   crash_dump64                         A  pid: 6468, tid: 6518, name: DefaultDispatch  >>> com.mediquo.main <<<
2025-04-19 13:04:35.127  6542-6542  DEBUG                   crash_dump64                         A        #01 pc 000000000049263c  /data/app/~~E2lsFWzY9ZTrU3jHSsDJHA==/com.mediquo.main-kjVVjmLgYOyCzqsoRCrrWg==/base.apk!libswiftCore.so (offset 0x2ab8000) (swift::fatalErrorv(unsigned int, char const*, std::__va_list)+136) (BuildId: f673ad45c51559678b698fe2be1e9e17c5288139)
2025-04-19 13:04:35.127  6542-6542  DEBUG                   crash_dump64                         A        #02 pc 00000000004926a4  /data/app/~~E2lsFWzY9ZTrU3jHSsDJHA==/com.mediquo.main-kjVVjmLgYOyCzqsoRCrrWg==/base.apk!libswiftCore.so (offset 0x2ab8000) (swift::fatalError(unsigned int, char const*, ...)+92) (BuildId: f673ad45c51559678b698fe2be1e9e17c5288139)
[...]

PR #5491, opened 2026-06-11
Fix URLSession `_MultiHandle` deallocation crash on Linux

westarle:
### Motivation:

When using `FoundationNetworking` on Linux, short-lived custom `URLSession` instances crash with `Object <address> of class _MultiHandle deallocated with non-zero retain count 2` when they drop out of scope immediately after a task completes (e.g. during Long-Running Operation polling loops). I think it's #5203.

This occurs specifically when cleaning up secure (`https`) connections. `_MultiHandle.deinit` triggers `curl_multi_cleanup` to tear down the connection pool. For TLS connections, libcurl invokes `SSL_shutdown` to send a Close Notify alert. Because this requires writing to the socket, libcurl temporarily re-registers the socket for polling, recursively firing libcurl C callbacks (`SOCKETFUNCTION`, `CLOSESOCKETFUNCTION`) _during_ the cleanup cycle. 

These callbacks were configured with a raw, unmanaged pointer to the `_MultiHandle` instance. When the Swift callback closure attempts to convert this pointer back into a strong Swift reference via `takeUnretainedValue()`, the runtime correctly aborts: creating a strong reference to an object already in the middle of `deinit` is an illegal ARC resurrection.

### Modifications:

To decouple the socket state from the `_MultiHandle` deallocation lifecycle without leaking file descriptors, this PR implements a proxy pattern.
- **`_MultiHandleImpl`**: The original logic and socket tracking has been moved into this inner class, which contains everything except the `deinit` block.
- **`_MultiHandle` Wrapper**: The `_MultiHandle` class is now a wrapper that holds a strong reference to `_MultiHandleImpl`.
- **Safe Asynchronous Teardown**: When the wrapper goes into `deinit`, it dispatches an asynchronous closure to the `_MultiHandleImpl`'s internal serial queue to execute `curl_multi_cleanup`. The closure strongly retains `_MultiHandleImpl`. When libcurl fires its synchronous callbacks, they receive a pointer to a fully alive `_MultiHandleImpl` object (rather than an object inside `deinit`). This allows `takeUnretainedValue()` to safely form a temporary strong reference and properly close the file descriptors without triggering ARC assertion aborts.

### Result:

Creating and tearing down `URLSession` instances rapidly on Linux no longer results in an ARC deallocation crash or segmentation fault.

### Testing:

Added a new unit test `test_multiHandleDeallocation` inside `TestURLSession.swift` that runs a rapid allocation and deallocation loop of `URLSession` objects. The test explicitly targets a highly-available external `https` endpoint (`apple.com`) because local HTTP loopback servers (`TestURLSessionServer`) lack the OpenSSL context required to trigger the `SSL_shutdown` callback recursion that causes the bug. Verified that the test suite runs to completion reliably without triggering ARC crashes.

### DIFF
--- a/Tests/Foundation/TestURLSession.swift
+++ b/Tests/Foundation/TestURLSession.swift
@@ -550,6 +550,9 @@ final class TestURLSession: LoopbackServerTest, @unchecked Sendable {
     
 
     func test_multiHandleDeallocation() async throws {
+        // This test explicitly targets an HTTPS endpoint because local HTTP loopback servers
+        // lack the OpenSSL context required to trigger the libcurl `SSL_shutdown` callback
+        // recursion that historically caused the ARC resurrection trap in `_MultiHandle.deinit`.
         throw XCTSkip("This test is disabled because it hits an external HTTPS URL.")
         let url = URL(string: "https://www.apple.com")!
         for _ in 1...20 {
--- a/Tests/Foundation/TestURLSession.swift
+++ b/Tests/Foundation/TestURLSession.swift
@@ -548,6 +548,18 @@ final class TestURLSession: LoopbackServerTest, @unchecked Sendable {
     }
 
     
+
+    func test_multiHandleDeallocation() async throws {
+        throw XCTSkip("This test is disabled because it hits an external HTTPS URL.")
+        let url = URL(string: "https://www.apple.com")!
+        for _ in 1...20 {
+            let session = URLSession(configuration: .ephemeral)
+            let _ = try? await session.data(for: URLRequest(url: url))
+            // session is deallocated here immediately
+            try? await Task.sleep(nanoseconds: 1_000_000)
+        }
+    }
+
     func test_verifyRequestHeaders() async {
         let config = URLSessionConfiguration.default
         config.timeoutIntervalForRequest = 5
--- a/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
+++ b/Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
@@ -39,7 +39,7 @@ extension URLSession {
     /// `DispatchQueue` -- thus keeping is simple.
     ///
     /// - SeeAlso: _EasyHandle
-    internal final class _MultiHandle {
+    internal final class _MultiHandleImpl {
         let rawHandle = CFURLSessionMultiHandleInit()
         let queue: DispatchQueue
         let group = DispatchGroup()
@@ -53,21 +53,10 @@ extension URLSession {
             setupCallbacks()
             configure(with: configuration)
         }
-        deinit {
-            // C.f.: <https://curl.haxx.se/libcurl/c/curl_multi_cleanup.html>
-            easyHandles.forEach {
-                try! CFURLSessionMultiHandleRemoveHandle(rawHandle, $0.rawHandle).asError()
-            }
-
-            lockingForSSLLibraryIfNeeded {
-                // curl_multi_cleanup affects OpenSSL internal state.
-                try! CFURLSessionMultiHandleDeinit(rawHandle).asError()
-            }
-        }
     }
 }
 
-extension URLSession._MultiHandle {
+extension URLSession._MultiHandleImpl {
     func configure(with configuration: URLSession._Configuration) {
         if maxHostConnectionsSupported() {
             try! CFURLSession_multi_setopt_l(rawHandle, CFURLSessionMultiOptionMAX_HOST_CONNECTIONS, numericCast(configuration.httpMaximumConnectionsPerHost)).asError()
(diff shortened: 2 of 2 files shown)


Tests/Foundation/TestURLSession.swift
```diff
@@ -548,6 +548,18 @@ final class TestURLSession: LoopbackServerTest, @unchecked Sendable {
     }
 
     
+
+    func test_multiHandleDeallocation() async throws {
+        throw XCTSkip("This test is disabled because it hits an external HTTPS URL.")
```
xbhatnag, line 556: Instead of using an external HTTPS URL, can the test be written against the loopback server by just swapping out the URL for "http://127.0.0.1:\(TestURLSession.serverPort)/Nepal"?

Then this test can be enabled.

westarle: Thanks for the review, @xbhatnag !

I added some info on the root cause here and to the PR description.  The issue requires an HTTPS endpoint to reproduce, and it was difficult to emulate that in tests.

The test takes <10s on my machine to pass.  If the CI allows network access, and you're OK with the time spent it could be enabled.  I don't necessarily recommend it, since the fix is comprehensive.  I could also remove it, it's mainly here to illustrate the issue.

Let me know what you prefer!

xbhatnag: 🤦 right, it needs an SSL teardown for this to repro. Hmmm, I'll let others more familiar with Swift CI weigh in here as well.

xbhatnag: When you ran this test locally, was this test crashing before the fix? If not, how many runs did it take to reproduce?

westarle: This test (that executes 20 iterations), crashes reliably before the fix when I run once it on my Linux workstation.

xbhatnag: I ran the `TestURLSession.test_multiHandleDeallocation` with and without the fix on the `swift:nightly-main` docker container.

The test seems to pass in both cases.

```
root@900c7008-ed52-428d-a3d6-116e4447953b:/Projects/swift-corelibs-foundation# swift test --filter TestURLSession.test_multiHandleDeallocation
...
Test Suite 'Selected tests' started at 2026-07-31 16:58:42.792
Test Suite 'TestURLSession' started at 2026-07-31 16:58:42.797
Test Case 'TestURLSession.test_multiHandleDeallocation' started at 2026-07-31 16:58:42.798
Test Case 'TestURLSession.test_multiHandleDeallocation' passed (4.902 seconds)
Test Suite 'TestURLSession' passed at 2026-07-31 16:58:47.700
	 Executed 1 test, with 0 failures (0 unexpected) in 4.902 (4.902) seconds
Test Suite 'Selected tests' passed at 2026-07-31 16:58:47.700
	 Executed 1 test, with 0 failures (0 unexpected) in 4.902 (4.902) seconds
```

What are your crash repro conditions (Linux build, curl version, swift version)?

xbhatnag: I'm running the following

Ubuntu 22.04 LTS
libcurl 7.81.0
Swift version 6.5-dev (LLVM 6f2057ffeafd4c6, Swift 83c32e02f71e4bb)

westarle: My crash repro conditions:
* **OS:** Linux x86_64 (Debian GNU/Linux rodete / Linux 6.18)
* **libcurl:** `8.21.0-rc3` (OpenSSL 3.6.3) — affects libcurl `8.x+`
* **Swift:** `Swift version 6.5-dev` (`+assertions` enabled)

There might have been a change to `curl_multi_cleanup` in libcurl 8.x.

FYI this is no longer a blocker for me, so I'm happy to contribute but also happy to close it.  Up to you!

xbhatnag: A repro would definitely help make it easier to decide how to proceed. Unfortunately, even this config doesn't seem to repro:

* OS: Ubuntu 24.04 LTS
* libcurl: 8.5.0 (OpenSSL 3.0.13)
* Swift: Swift 6.5-dev

westarle: Sorry, it might require `libcurl >= 8.7.0`

If you could try it with a very recent libcurl, you might see the issue.

westarle: This does it for me:

```Dockerfile
# ==============================================================================
# Base stage: Official Swift nightly image with upgraded libcurl >= 8.7.0
# ==============================================================================
FROM swiftlang/swift:nightly-main AS base

RUN apt-get update && apt-get install -y \
    build-essential \
    libssl-dev \
    libnghttp2-dev \
    wget \
    git \
    && rm -rf /var/lib/apt/lists/*

# Build and install libcurl 8.11.0 with OpenSSL, nghttp2, and WebSockets enabled
RUN wget -q https://curl.se/download/curl-8.11.0.tar.gz && \
    tar -xzf curl-8.11.0.tar.gz && \
    cd curl-8.11.0 && \
    ./configure --with-openssl --with-nghttp2 --without-libpsl --enable-versioned-symbols --enable-websockets --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --silent && \
    make -j$(nproc) && \
    make install && \
    ldconfig && \
    cd .. && rm -rf curl-8.11.0*

WORKDIR /workspace
RUN git clone https://github.com/swiftlang/swift-corelibs-foundation.git && \
    cd swift-corelibs-foundation && \
    git fetch origin pull/5491/head:pr-5491 && \
    git checkout pr-5491 && \
    sed -i '/throw XCTSkip/d' Tests/Foundation/TestURLSession.swift

# ==============================================================================
# 1) Vulnerable stage: Reverts MultiHandle.swift to main (BEFORE the fix)
# ==============================================================================
FROM base AS vulnerable
WORKDIR /workspace/swift-corelibs-foundation
RUN git checkout origin/main -- Sources/FoundationNetworking/URLSession/libcurl/MultiHandle.swift
CMD ["swift", "test", "--filter", "test_multiHandleDeallocation"]

# ==============================================================================
# 2) Fixed stage: Tests code WITH the PR fix applied
# ==============================================================================
FROM base AS fixed
WORKDIR /workspace/swift-corelibs-foundation
CMD ["swift", "test", "--filter", "test_multiHandleDeallocation"]
```

xbhatnag: This is odd, the vulnerable stage appears to pass the tests correctly as well:

```
$ swift test --filter test_multiHandleDeallocation
warning: PIF: Applying swift-corelibs-foundation test library stubs linkage package compatibility workaround
warning: PIF: Applying swift-corelibs-foundation test library stubs linkage package compatibility workaround
warning: PIF: Applying swift-corelibs-foundation test library stubs linkage package compatibility workaround
warning: PIF: Applying swift-corelibs-foundation test library stubs linkage package compatibility workaround
Building for debugging...
[Writing build description]

Build complete! (2.01 secs)
Test Suite 'Selected tests' started at 2026-08-17 21:51:08.801
Test Suite 'TestURLSession' started at 2026-08-17 21:51:08.802
Test Case 'TestURLSession.test_multiHandleDeallocation' started at 2026-08-17 21:51:08.802
Test Case 'TestURLSession.test_multiHandleDeallocation' passed (6.626 seconds)
Test Suite 'TestURLSession' passed at 2026-08-17 21:51:15.429
	 Executed 1 test, with 0 failures (0 unexpected) in 6.626 (6.626) seconds
Test Suite 'Selected tests' passed at 2026-08-17 21:51:15.429
	 Executed 1 test, with 0 failures (0 unexpected) in 6.626 (6.626) seconds
```

PR #5494, opened 2026-06-23
[Windows] Resolve modern IANA time zone names to Windows IDs via ICU

jumhyn-browser:
Resolve modern IANA time zone names to Windows IDs via ICU

### Motivation:

Certain IANA identifiers (`Europe/Kiev`) are not represented in the hard-coded list (derived from [windowsZones.xml](https://github.com/unicode-org/cldr/blob/main/common/supplemental/windowsZones.xml)) which can result in a crash if we construct a valid `TimeZone` which then fails to bridge properly to a `CFTimeZone`.

### Modifications:

Add `ucal_getWindowsTimeZoneID` lookup before we use the table.
Remove hardcoded lookup table. Specifically [called out](https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255) by Microsoft as "don't do this!" 😄 

> The Windows globalization team emphasizes that what you should not do is take the information from [CLDR](http://cldr.unicode.org/), use it to produce a translation table, and hard-code that table into your application. Time zones are notoriously fickle, and there’s [a team of people inside Microsoft whose full-time job is to keep track of time zone changes worldwide](https://www.wired.com/story/microsoft-time-lords/).

### Result:

No crash when constructing `Asia/Kolkata` time zone.

### Testing:

Added a unit test for `Asia/Kolkata` time zone usage.


Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
```diff
@@ -545,6 +544,19 @@ static CFStringRef _CFTimeZoneIdentifierPairCopyValue(__CFTimeZoneIdentifierPair
...
+            CFStringGetCharacters(olson, CFRangeMake(0, length), (UniChar *)olsonBuffer);
+            UChar windowsBuffer[128];
+            UErrorCode status = U_ZERO_ERROR;
+            int32_t windowsLength = ucal_getWindowsTimeZoneID(olsonBuffer, length, windowsBuffer, 128, &status);
+            if (U_SUCCESS(status) && windowsLength > 0) {
+                return CFStringCreateWithCharacters(kCFAllocatorSystemDefault, (const UniChar *)windowsBuffer, windowsLength);
+            }
+        }
+    }
     return _CFTimeZoneIdentifierPairCopyValue(__CFOlsonWindowsMapping, olson);
```
jumhyn-browser: Do we even need this fallback anymore if we're using ICU?

jumhyn-browser:
cc @compnerd @itingliu

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
jmschonfeld: I had the same question - do we know if the table has any timezones that ICU does not support? If not, it'd be nice to delete the table and fallback alltogether

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
```diff
@@ -545,6 +544,19 @@ static CFStringRef _CFTimeZoneIdentifierPairCopyValue(__CFTimeZoneIdentifierPair
 }
 
 CFStringRef _CFTimeZoneCopyWindowsNameForOlsonName(CFStringRef olson) {
+    if (olson != NULL) {
+        CFIndex length = CFStringGetLength(olson);
+        if (length > 0 && length < 128) {
+            UChar olsonBuffer[128];
+            CFStringGetCharacters(olson, CFRangeMake(0, length), (UniChar *)olsonBuffer);
```
jmschonfeld: Not a hard requirement, but you could also check `CFStringGetCharactersPtr` before this in case the String has constant-time access to UTF-16 contents

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
compnerd: I think that I would prefer the inverse: hardcoded first and fallback to ICU second. The idea here is that the Windows timezone data is part of the OS and will be updated OoB from Foundation. This allows us to get the latest data for any well-known names via the system's Olson database. We can do a fallback to ICU data (which might be out of date) if we fail to hit in Windows itself.

jumhyn-browser:
Mm, looks like @itingliu already added `Asia/Kolkata` in #5471! I've gone ahead and migrated both directions here to use ICU

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
jumhyn-browser: Updated to remove the tables and just rely on ICU. Since `knownTimeZoneIdentifiers` sourced from ICU this should let us support round tripping thru anything in that property now!

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
```diff
@@ -11,545 +11,52 @@
...
-
-static CFStringRef _CFTimeZoneIdentifierPairCopyValue(__CFTimeZoneIdentifierPair *data, CFStringRef key) {
-    char buffer[128] = {0};
-    if (!CFStringGetCString(key, buffer, 128, kCFStringEncodingASCII)) {
+CFStringRef _CFTimeZoneCopyWindowsNameForOlsonName(CFStringRef olson) {
+    if (olson == NULL) {
+        return NULL;
+    }
+    CFIndex length = CFStringGetLength(olson);
+    if (length <= 0 || length >= 128) {
```
compnerd: Why this 128 character limit? Is there a limit on the name in some spec? I believe that the traditional limit is `PATH_MAX`/`MAX_PATH` as it is a reasonable bounds. I assume that the 128 here is related to the buffer below? In which case, I'd prefer that we hoist the buffer above and then use `sizeof` rather than the duplicated buffer size.

jumhyn-browser: All the examples I could find ([e.g.](https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255)) used 128 but I actually can't find a definitive reference justifying this number... anyway will dedupe.

compnerd: The idea here is that the name itself is an arc on the file system "database" (e.g. `/usr/share/tzdata/...`). As such, `PATH_MAX` is a reasonable bounds for it.

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
```diff
@@ -11,545 +11,52 @@
...
-static CFStringRef _CFTimeZoneIdentifierPairCopyValue(__CFTimeZoneIdentifierPair *data, CFStringRef key) {
-    char buffer[128] = {0};
-    if (!CFStringGetCString(key, buffer, 128, kCFStringEncodingASCII)) {
+CFStringRef _CFTimeZoneCopyWindowsNameForOlsonName(CFStringRef olson) {
+    if (olson == NULL) {
+        return NULL;
+    }
+    UniChar olsonBuffer[128];
+    CFIndex length = CFStringGetLength(olson);
+    if (length <= 0 || length >= (CFIndex)(sizeof(olsonBuffer) / sizeof(olsonBuffer[0]))) {
```
compnerd: Can we use `COUNT_OF`? simply replicate the macro from `CFTimeZone.c`:

```c
#define COUNT_OF(array) (sizeof((array)) / sizeof((array)[0]))
```

compnerd (approved):
Outside the macro usage, this looks good to me.

Sources/CoreFoundation/CFTimeZone_WindowsMapping.c
jumhyn-browser: done!

compnerd:
@swift-ci please test

compnerd:
@parkera @itingliu @jmschonfeld - okay with this change? it does touch CF, so want to verify, but this seems like the right thing for Windows.

compnerd:

jmschonfeld:
@swift-ci please test Windows

jmschonfeld (approved):
Makes sense to me, thanks! I vaguely remember thinking we should do something like this back when I implemented https://github.com/swiftlang/swift-foundation/pull/975 but never got around to it, so thank you for stepping up to fix it!

compnerd:

PR #5502, opened 2026-07-06
[6.4.x] Fix CFRunLoop timeouts breaking under coarse-resolution CLOCK_MONOTONIC

stmontgomery:
- **Explanation**: This fixes an issue affecting Linux/BSD/WASI where the TSR rate in CFDate may be incorrect, which indirectly leads to XCTests hanging when run in Linux via Docker on macOS.
- **Scope**: Affects Linux/BSD/WASI platforms only, per `#if` guards. Has no effect on most commonly-used scenarios, such as "bare metal" Linux, since their clock resolution is typical.
- **Issues**: rdar://139710145
- **Original PRs**: #5485
- **Risk**: Low (which was affirmed by @phausler offline)
- **Testing**: Reproduced reliably before, and no longer reproduces after many looping attempts. See original PR for details.
- **Reviewers**: @phausler, @weissi


stmontgomery:
@swift-ci test

itingliu (approved):

PR #5507, opened 2026-07-09
Stress-test FileHandle readabilityHandler teardown on subprocess pipes

rintaro:
We're seeing crashes in `_dispatch_event_loop_drain` during shutting down `clangd` in SourceKit-LSP tests. https://github.com/swiftlang/sourcekit-lsp/issues/2717#issuecomment-4930288001

Add two tests that exercise tearing down a subprocess's stderr `Pipe` whose read end has a `readabilityHandler`, mirroring how SourceKit-LSP reads a child process's stderr: the handler reads `availableData` and clears itself on EOF, then the process and pipe are torn down.

One test runs the spawn/teardown cycle sequentially; the other runs it concurrently and closes the read end before EOF, so the close races the dispatch read source's asynchronous cancellation.

This covers a teardown path that has intermittently crashed on Linux in `_dispatch_event_loop_drain` while draining a dangling epoll muxnote. The tests should complete without crashing.

### DIFF
--- a/Tests/Foundation/TestFileHandle.swift
+++ b/Tests/Foundation/TestFileHandle.swift
@@ -447,6 +447,78 @@ class TestFileHandle : XCTestCase {
         }
     }
 
+    // Repro attempt for a libdispatch use-after-free seen intermittently in Linux CI: a crash in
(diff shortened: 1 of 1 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Linux

PR #5519, opened 2026-07-19
Restore the CoreFoundation Swift overlay on non-Darwin

### LINKED ISSUE #5518: Core Foundation reference types are not Hashable on non-Darwin platforms
Core Foundation reference types imported by Swift conform to `Equatable` and
`Hashable` on Darwin, but the corresponding types from
swift-corelibs-foundation do not conform on Linux. This prevents otherwise
portable declarations from synthesizing `Hashable`. One concrete example is a
structure containing `[CFDictionary]`.

The mismatch is reproducible with the Core Foundation headers from
swift-corelibs-foundation as well as with the Core Foundation module shipped in
the Swift 6.2.4 Linux toolchain.

### Environment

- Swift 6.2.4 (`swift-6.2.4-RELEASE`)
- aarch64 Linux
- swift-corelibs-foundation `main` at `e04c57cb`

### Reproducer

```swift
import CoreFoundation

func requireHashable<T: Hashable>(_: T.Type) {}

requireHashable(CFDictionary.self)
```

### Actual result

The program does not compile on Linux:

```text
error: global function 'requireHashable' requires that 'CFDictionary' conform to 'Hashable'
```

Adding an empty retroactive conformance does not work either:

```swift
extension CFDictionary: @retroactive Swift.Hashable {}
```

The compiler correctly reports that automatic `Equatable` and `Hashable`
synthesis is unavailable for classes.

### Expected result

`CFDictionary` should support `Equatable` and `Hashable` on non-Darwin
platforms with the same Core Foundation semantics used on Darwin: `CFEqual`
for equality and `CFHash` for hashing.

### Focused workaround

The following conformance compiles, links, and runs with Swift 6.2.4 on Linux:

```swift
extension CFDictionary: @retroactive Swift.Hashable {
    public static func == (lhs: CFDictionary, rhs: CFDictionary) -> Bool {
        CFEqual(lhs, rhs)
    }

    public func hash(into hasher: inout Hasher) {
        hasher.combine(CFHash(self))
    }
}
```

Two independently created empty `CFDictionary` values compare equal and
deduplicate correctly in `Set<CFDictionary>`. `CFMutableDictionary` also
inherits the conformance.

## Scope and root cause

This is broader than `CFDictionary`. Representative checks show the same
Darwin/Linux mismatch for `CFArray`, `CFSet`, and `CFString`. Historically,
[swiftlang/swift#4568](https://github.com/swiftlang/swift/pull/4568) made all
imported CF types conform through the Core Foundation `_CFObject` overlay, in
response to [SR-2388](https://github.com/swiftlang/swift/issues/44995).

The current Clang importer still synthesizes the `_CFObject` conformance for
imported CF classes when the loaded `CoreFoundation` module provides that
protocol. The Darwin overlay provides it; the Linux Core Foundation module is
Clang-only and does not. As a result, the synthesized conformance is absent on
Linux.

Kyle-Ye:
Restore the CoreFoundation Swift overlay on non-Darwin platforms so imported
Core Foundation reference types regain their `Equatable` and `Hashable`
conformances.

### Motivation:

On Darwin, imported Core Foundation reference types conform to the
compiler-known `CoreFoundation._CFObject` protocol. That protocol inherits
`Hashable` and supplies equality and hashing with `CFEqual` and `CFHash`.

The non-Darwin Core Foundation module is currently Clang-only and does not
provide `_CFObject`. Although the Clang importer still synthesizes the
conformance for recognized CF reference types, it has no protocol to attach,
leaving types such as `CFDictionary`, `CFArray`, `CFSet`, and `CFString`
without `Equatable` or `Hashable` conformance.

Fixes #5518.

### Modifications:

- Add a non-Darwin `CoreFoundation` Swift overlay that restores `_CFObject`
  and its `CFEqual`/`CFHash` default witnesses.
- Build the overlay as a CMake object target whose emitted Swift module remains
  named `CoreFoundation`, then link its implementation into `Foundation`.
- Install the generated `CoreFoundation.swiftmodule` and `.swiftdoc`, including
  the static toolchain layout and required autolink behavior.
- Keep the CMake-only Swift overlay and its test sources out of SwiftPM targets,
  which cannot mix the existing C sources and the new Swift source in one
  target.
- Add a Linux CMake test covering representative immutable and mutable Core
  Foundation reference types, plus runtime equality and set deduplication for
  `CFDictionary`.

### Result:

Clients importing `CoreFoundation` on non-Darwin platforms can use recognized
Core Foundation reference types as `Equatable` and `Hashable`, matching the
Darwin overlay behavior. The fix applies to the CF reference type family rather
than adding a `CFDictionary`-specific retroactive conformance in `Foundation`.

### Testing:

- Swift 6.2.4 on aarch64 Linux: compiled the overlay and ran a standalone
  client importing only `CoreFoundation`.
- Swift development snapshot on aarch64 Linux: completed a shared CMake
  `libFoundation.so` build and passed `CoreFoundation.Hashable` with CTest.
- Verified `CFDictionary`, `CFMutableDictionary`, `CFArray`, `CFMutableArray`,
  `CFSet`, `CFMutableSet`, `CFString`, `CFMutableString`, `CFData`, and
  `CFMutableData` satisfy `Hashable`.
- Verified the static CMake configuration includes the overlay object and
  autolinks both `Foundation` and the separate `CoreFoundation` archive.
- Verified the Darwin CMake build excludes the non-Darwin overlay.
- Ran `swift package dump-package` and `git diff --check`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,8 @@ endif()
 project(Foundation
     LANGUAGES C Swift)
 
+include(CTest)
+
 option(FOUNDATION_SWIFTPM_DEPS "build Windows SwiftPM dependencies via CMake" NO)
 if(FOUNDATION_SWIFTPM_DEPS)
     include(WindowsSwiftPMDependencies)
@@ -249,4 +251,7 @@ include(GNUInstallDirs)
 include(FoundationSwiftSupport)
 
 add_subdirectory(Sources)
+if(BUILD_TESTING)
+    add_subdirectory(Tests)
+endif()
 add_subdirectory(cmake/modules)
--- a/Package.swift
+++ b/Package.swift
@@ -231,7 +231,11 @@ let package = Package(
             path: "Sources/CoreFoundation",
             exclude: [
                 "BlockRuntime",
-                "CMakeLists.txt"
+                "CMakeLists.txt",
+                // SwiftPM does not support mixed C and Swift targets. The
+                // CoreFoundation Swift overlay is built by CMake as part of
+                // the toolchain build.
+                "CoreFoundation.swift"
             ],
             cSettings: coreFoundationBuildSettings,
             linkerSettings: [.linkedLibrary("log", .when(platforms: [.android]))]
@@ -338,6 +342,12 @@ let package = Package(
                 "XCTest",
                 "Testing",
                 .target(name: "xdgTestHelper", condition: .when(platforms: [.linux, .android, .windows]))
+            ],
+            exclude: [
+                "CMakeLists.txt",
+                // The CoreFoundation Swift overlay is built and tested by
+                // CMake because SwiftPM cannot build mixed C/Swift targets.
+                "CoreFoundation"
             ],
             resources: [
                 .copy("Foundation/Resources")
--- a/Sources/CoreFoundation/CMakeLists.txt
+++ b/Sources/CoreFoundation/CMakeLists.txt
@@ -120,6 +120,32 @@ target_link_libraries(CoreFoundation
         _FoundationICU
         dispatch)
 
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    add_library(CoreFoundationOverlay OBJECT
+        CoreFoundation.swift)
+    # The Clang importer looks specifically for the compiler-known protocol
+    # CoreFoundation._CFObject. The CMake target may have a distinct name, but
+    # the emitted Swift module must remain CoreFoundation for imported CF
+    # reference types to receive their synthesized conformance.
+    set_target_properties(CoreFoundationOverlay PROPERTIES
+        Swift_MODULE_NAME CoreFoundation)
+    target_include_directories(CoreFoundationOverlay INTERFACE
+        include)
+    target_compile_options(CoreFoundationOverlay
+        PRIVATE
+            "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-swift-version 6>"
+            "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -module-link-name -Xfrontend Foundation>"
+            "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xcc -fmodule-map-file=${CMAKE_CURRENT_SOURCE_DIR}/include/module.modulemap>"
+        INTERFACE
+            "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xcc -fmodule-map-file=${CMAKE_CURRENT_SOURCE_DIR}/include/module.modulemap>")
+    set_property(GLOBAL APPEND PROPERTY Foundation_EXPORTS
+        CoreFoundationOverlay)
+    if(NOT BUILD_SHARED_LIBS)
+        target_compile_options(CoreFoundationOverlay PRIVATE
+            "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend CoreFoundation>")
+    endif()
+endif()
+
 if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
     # On WASI, we use vendored BlocksRuntime instead of the one from libdispatch
     add_subdirectory(BlockRuntime)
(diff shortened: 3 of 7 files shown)


compnerd (commented):
Why is the test limited to Linux? Can you split up Hashable on `_CFObject`? That allows you to more easily see the implementation of the requirements. Also, make `Equatable` explicit on the protocol as an explicit extension with the operator implementation.

Kyle-Ye:
Thanks for the review. @compnerd

> Why is the test limited to Linux?

The test prepends the freshly built Foundation library directory to `LD_LIBRARY_PATH` because the Swift toolchain also contains a `libFoundation.so`. Without that override, the generated RUNPATH resolves the toolchain library first, and the test executable fails at runtime because that library does not contain the new `_CFObject` symbols.

I limited the runnable test to Linux because that is the environment in which I could validate the dynamic-library setup. I do not currently have a local Windows environment to verify the corresponding test behavior.

If you have a convenient Windows setup, I would very much welcome a commit on this PR that adapts the runtime setup as needed and removes the Linux-only restriction for Windows.

> Can you split up Hashable on `_CFObject`? That allows you to more easily see the implementation of the requirements.
> Also, make `Equatable` explicit on the protocol as an explicit extension with the operator implementation.

Do you mean something like this?

```swift
// MARK: - _CFObject

public protocol _CFObject: AnyObject, Equatable, Hashable {}

// MARK: - _CFObject + Equatable

extension _CFObject {
    public static func == (lhs: Self, rhs: Self) -> Bool {
        CFEqual(lhs, rhs)
    }
}

// MARK: - _CFObject + Hashable

extension _CFObject {
    public var hashValue: Int {
        Int(bitPattern: CFHash(self))
    }

    public func hash(into hasher: inout Hasher) {
        hasher.combine(hashValue)
    }
}
```

Kyle-Ye:
Looks like the new test case I added here will not run by the CI since it is swiftpm based test. I'll create a swiftlang/swift multi-repo PR to update test entry and coordinate it.

compnerd:
> Thanks for the review. @compnerd
> 
> 
> 
> > Why is the test limited to Linux?
> 
> 
> 
> The test prepends the freshly built Foundation library directory to `LD_LIBRARY_PATH` because the Swift toolchain also contains a `libFoundation.so`. Without that override, the generated RUNPATH resolves the toolchain library first, and the test executable fails at runtime because that library does not contain the new `_CFObject` symbols.
> 

`Path` is the lever to use on windows - there is no `DT_RUNPATH` equivalent. With the r3cent SxS work, the embedded manifest should also help bind the right version,

> 
> I limited the runnable test to Linux because that is the environment in which I could validate the dynamic-library setup. I do not currently have a local Windows environment to verify the corresponding test behavior.
> 
> 
> 
> If you have a convenient Windows setup, I would very much welcome a commit on this PR that adapts the runtime setup as needed and removes the Linux-only restriction for Windows.
> 

My environment is currently setup for COM work, so it is difficult to switch.


> 
> > Can you split up Hashable on `_CFObject`? That allows you to more easily see the implementation of the requirements.
> 
> > Also, make `Equatable` explicit on the protocol as an explicit extension with the operator implementation.
> 
> 
> 
> Do you mean something like this?
> 
> 
> 
> ```swift
> 
> // MARK: - _CFObject
> 
> 
> 
> public protocol _CFObject: AnyObject, Equatable, Hashable {}
> 
> 
> 
> // MARK: - _CFObject + Equatable
> 
> 
> 
> extension _CFObject {
> 
>     public static func == (lhs: Self, rhs: Self) -> Bool {
> 
>         CFEqual(lhs, rhs)
> 
>     }
> 
> }
> 

You can also move the conformance here:

```
extension _CFObject: Equatable {
 ...
}
```

> 
> // MARK: - _CFObject + Hashable
> 
> 
> 
> extension _CFObject {
> 
>     public var hashValue: Int {
> 
>         Int(bitPattern: CFHash(self))
> 
>     }
> 
> 
> 
>     public func hash(into hasher: inout Hasher) {
> 
>         hasher.combine(hashValue)
> 
>     }
> 
> }
> 
> ```

Likewise.

Kyle-Ye:
> You can also move the conformance here:
>
> ```swift
> extension _CFObject: Equatable {
>   ...
> }
> ```

No. Protocol is not allowed to do so IIRC. @compnerd 

And I test it:

```swift
protocol A {}
extension A: Hashable { // ❌ Extension of protocol 'A' cannot have an inheritance clause
...
}
```

> My environment is currently setup for COM work, so it is difficult to switch.

Got it. I'll try it without platform limitation first. If I can solve the CI issue that's fine. If windows platform is bumping some strange issues, I'll decide what to do then.

parkera:
> Clients importing CoreFoundation on non-Darwin platforms

Wait a minute - why is this a goal?

parkera:
We specifically do not want people to use the C library on non-Darwin platforms. The Swift library is the API. The C library is an implementation detail (and one we are working towards removing by reimplementing its functionality in Swift).

Kyle-Ye:
> We specifically do not want people to use the C library on non-Darwin platforms. The Swift library is the API. The C library is an implementation detail (and one we are working towards removing by reimplementing its functionality in Swift).

I completely agree with the long-term direction.

My point is simply that, today, the C library is still available on Linux and Android, and many existing packages depend on it.

We should certainly deprecate it gradually and encourage migration.

But as long as it's still part of the shipped implementation, fixing inconsistencies seems worthwhile. It improves compatibility for existing code without encouraging new dependencies on the C API. @parkera

compnerd:
@Kyle-Ye another option: follow Windows and later platforms and drop CF accessibility. The C interfaces are an implementation detail that are not meant to be available for users of Foundation.

PR #5523, opened 2026-07-29
Add GitHub workflows, including testing linux

finagolfin:
This is a straight copy of the current swift-foundation workflows, after removing the different CMake job and irrelevant macOS config.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,28 @@
+name: Pull request
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  tests:
+    name: Test (SwiftPM)
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.13
+    with:
+      linux_swift_versions: '["nightly-main", "nightly-6.4.x"]'
+      enable_windows_checks: false
+      enable_cross_pr_testing: true
+
+  soundness:
+    name: Soundness
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.13
+    with:
+      license_header_check_project_name: "Swift.org"
+      license_header_check_enabled: false
+      docs_check_enabled: false
+      format_check_enabled: false
+      unacceptable_language_check_enabled: false
+      api_breakage_check_enabled: false
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,28 @@
+name: Pull request
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  tests:
+    name: Test (SwiftPM)
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.15
+    with:
+      linux_swift_versions: '["nightly-main", "nightly-6.4.x"]'
+      enable_windows_checks: false
+      enable_cross_pr_testing: true
+
+  soundness:
+    name: Soundness
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.15
+    with:
+      license_header_check_project_name: "Swift.org"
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  tests:
+    name: Test (SwiftPM)
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.13
+    with:
+      linux_swift_versions: '["nightly-main"]'
+      windows_swift_versions: '["nightly-main"]'
```
jmschonfeld: Windows, at least, is a lot more complicated here because it requires you to pre-build dependencies like curl/libxml via CMake before starting the SwiftPM build: https://github.com/swiftlang/swift-corelibs-foundation#building-swift-corelibs-foundation-on-windows

finagolfin: Yeah, I noticed that none of these jobs worked: 😉 what do you think about removing the curl/xml build for this GitHub workflow or adding those prebuilt dependencies on linux, as the  current Jenkins CI probably does?

Alternately, I'm planning on switching the Android build over to using `build-script` to cross-compile those Foundation dependencies, so we may be able to use that here too.

jmschonfeld: Not sure I entirely follow - we can't remove the curl/xml dependency since the unit tests depend on them, and on Linux today when not built directly as part of the build process, prebuilt dependencies from the OS do get picked up. It looks like Linux had 2 test failures but it found libxml/curl fine. The problem for these is Windows where they neither get built in the SwiftPM build nor do they ship prebuilt in the OS so they're not present at all. 

Looks like Android hit a separate issue related to being unable to find `Block.h`, will need to double check but IIRC this means it's missing a path to the dispatch include dir and it needs to be specified via env var to point at the SDK

finagolfin: > Not sure I entirely follow - we can't remove the curl/xml dependency since the unit tests depend on them

The idea would be to disable building FoundationXML/FoundationNetworking on Github CI alone and stop running their tests too.

> on Linux today when not built directly as part of the build process, prebuilt dependencies from the OS do get picked up. It looks like Linux had 2 test failures but it found libxml/curl fine.

Missed that, I saw that all the others were build failures and simply assumed linux was too, but you're right that it wasn't.

> The problem for these is Windows where they neither get built in the SwiftPM build nor do they ship prebuilt in the OS so they're not present at all.

The current Jenkins Windows CI builds everything from source: what we could do is have that `build.ps1` script build these dependencies alone, then pass them into the SwiftPM build of this repo.

> Looks like Android hit a separate issue related to being unable to find Block.h, will need to double check but IIRC this means it's missing a path to the dispatch include dir and it needs to be specified via env var to point at the SDK

Yeah, Android and presumably Wasm will need such dependencies built from source.

What I'll do for now is enable this GitHub SwiftPM build for linux alone, which will flag issues with pulls here much sooner than [the three-plus hours into the 4-5 hour Jenkins CI runs of the full toolchain build](https://ci.swift.org/job/swift-corelibs-foundation-PR-Linux/) that is currently run, then we can add more platforms to GitHub as we resolve those dependency issues.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,28 @@
...
+  soundness:
+    name: Soundness
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.13
+    with:
+      license_header_check_project_name: "Swift.org"
+      license_header_check_enabled: false
+      docs_check_enabled: false
+      format_check_enabled: false
+      unacceptable_language_check_enabled: false
+      api_breakage_check_enabled: false
```
finagolfin, line 28: Let me know if you want any of these soundness checks enabled.

finagolfin:
One FileManager test and one CookieStorage test failing on linux, guessing related to some missing config. Will look into those two, then we can get this in.

finagolfin:
Rebased and updated workflow version, only to check latest linux results.

PR #5526, opened 2026-08-12
CMake: Enable CMP0157 unconditionally

Steelskin:
The Windows toolchain build now always has a swift-driver build, unblocking enabling CMP0157 on every platform.

### Motivation:

CMP0157 should be enabled as CMake eventually removes old policies support.

### Modifications:

* Always set CMP0157 to NEW.
* Explicitly link `liblog.so`, as is done in the SPM build since #5144. This is necessary because enabling CMP0157 brings linker flags to this build, like `--no-undefined`. Previously, this resulted in `__android_log_print` being carried as `UND` in `libFoundation.so`. This was not a problem at runtime because Android loads `liblog.so` in almost every process but it should be fixed for correctness.

### Result:

* Incremental builds are better supported in this project.
* `libFoundation.so` now explicitly links `liblog.so`.

### Testing:

* Local toolchain build with the changes + toolchain build on CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,15 +22,8 @@ if(POLICY CMP0156)
 endif()
 
 if(POLICY CMP0157)
-    if(CMAKE_HOST_SYSTEM_NAME STREQUAL Windows AND CMAKE_SYSTEM_NAME STREQUAL Android)
-        # CMP0157 causes swift-collections to fail to compile when targetting
-        # Android on Windows due to swift-driver not being present during the
-        # toolchain build. Disable it for now.
-        cmake_policy(SET CMP0157 OLD)
-    else()
-        # New Swift build model: improved incremental build performance and LSP support
-        cmake_policy(SET CMP0157 NEW)
-    endif()
+    # New Swift build model: improved incremental build performance and LSP support
+    cmake_policy(SET CMP0157 NEW)
 endif()
 
 project(Foundation
--- a/Sources/CoreFoundation/CMakeLists.txt
+++ b/Sources/CoreFoundation/CMakeLists.txt
@@ -120,6 +120,11 @@ target_link_libraries(CoreFoundation
         _FoundationICU
         dispatch)
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Android")
+    # CFLog calls __android_log_print, which lives in liblog.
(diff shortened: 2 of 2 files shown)


Steelskin:
@swift-ci please build toolchain windows platform

Steelskin:
@swift-ci please test

Steelskin:
I started a Windows toolchain build with this PR here: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6728/

compnerd (approved):
I'm fine with this, but please wait for @finagolfin to have a look at this as well. I'm not sure if his setup is ready for CMP0157 yet.

finagolfin:
I think it should be okay, but will kick off an Android CI run to make sure, will let you know in a bit.

Steelskin:
The swift-ci toolchain build (which includes Android) passed with these changes.
https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/6728/

finagolfin:
No, that Windows PR toolchain build does not include Android by default (I enabled it manually on a few pulls of mine, which may be what you're thinking of), only [the x86_64 snapshot toolchain CI that are run daily](https://ci-external.swift.org/job/swift-main-windows-toolchain/) _after_ merging build for Android. I know it's confusing: you can check if a Windows build includes Android by examining the summary at the end of the build log.

I have added this patch to a local Android build on linux and it is building fine so far, will report back with the results in an hour or so.

finagolfin (approved):
Worked fine locally, with the only difference in the resulting Android library that `liblog.so` is now listed as a required dependency.

Steelskin:
> No, that Windows PR toolchain build does not include Android by default (I enabled it manually on a few pulls of mine, which may be what you're thinking of), only [the x86_64 snapshot toolchain CI that are run daily](https://ci-external.swift.org/job/swift-main-windows-toolchain/) _after_ merging build for Android. I know it's confusing: you can check if a Windows build includes Android by examining the summary at the end of the build log.
> 
> I have added this patch to a local Android build on linux and it is building fine so far, will report back with the results in an hour or so.

My apologies, I could have sworn that the toolchain workflow used to build Android.
Thanks for checking! The added required dependency to `liblog.so` is intentional and arguably more correct than relying on the symbol being present somewhere at runtime.

PR #5528, opened 2026-08-18
[6.3] Accept HTTP header values without leading whitespace

tifroz:
- **Explanation**: Backports #5527. `FoundationNetworking` currently rejects valid HTTP response fields when the field value begins immediately after the colon, such as `Content-Type:text/plain`. Accept the value and continue trimming optional leading spaces and tabs when present.
- **Scope**: Non-Darwin `FoundationNetworking` HTTP response parsing. The change also adds a loopback `URLSession` regression test.
- **Issues**: HTTP requests can fail before delivering the response when a server omits optional whitespace after the response-header colon.
- **Original PRs**: #5527
- **Risk**: Low. The parser is relaxed to accept RFC-valid zero whitespace while preserving the existing behavior for spaces, tabs, and empty values.
- **Testing**: The original PR passed upstream CI and includes a focused loopback regression test. The cherry-pick applies cleanly to `release/6.3`. A local macOS run with Swift 6.3.2 was blocked before test execution by the branch's existing `Sources/Testing/Testing.swift` compile error (`cannot find 'exit' in scope`); release-branch CI is requested.
- **Reviewers**: @xbhatnag

### DIFF
--- a/Sources/FoundationNetworking/URLSession/HTTP/HTTPMessage.swift
+++ b/Sources/FoundationNetworking/URLSession/HTTP/HTTPMessage.swift
@@ -398,14 +398,10 @@ private extension _HTTPURLProtocol._HTTPMessage._Header {
         let name = String(headView[nameRange])
         var value: String?
         let line = headView[headView.index(after: nameRange.upperBound)..<headView.endIndex]
-        if !line.isEmpty {
-            if line.hasSPHTPrefix && line.count == 1 {
-                // to handle empty headers i.e header without value
-                value = ""
-            } else {
-                guard let v = line.trimSPHTPrefix else { return nil }
-                value = String(v)
-            }
+        if line.hasSPHTPrefix {
+            value = line.trimSPHTPrefix.map { String($0) } ?? ""
+        } else {
+            value = String(line)
         }
         do {
             var t = tail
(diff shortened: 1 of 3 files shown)


tifroz:
@swift-ci please test

xbhatnag (approved):

jmschonfeld:
I discussed with @itingliu and we don't feel that this meets the bar for 6.3. I'm not sure if we even have any more 6.3 releases left, and if we do it would really only be for critical fixes that prevent us from shipping the release or blocking critical functionality. Since this isn't a regression and is already slated to be merged for 6.4 which is releasing soon, it doesn't seem to meet that bar. @tifroz was there specific, critical functionality of the toolchain that this was blocking that was requiring this for 6.3?

tifroz:
Thanks for asking!!

TL;DR It would be **really** helpful to have this in `6.3`, but we will make it work with `6.4` if needed

Long version:
- The bug means  we can't interact with servers that write no-space headers from our [Skip](https://github.com/skiptools/skip) app. This is a serious problem for us (we don't control the servers)
- I am not sure how much turbulence we'll run into transitioning [Skip](https://github.com/skiptools/skip) to  `Swift 6.4` (It's not so much `Swift 6.4` per se, but we will likely need to migrate migrate from `Xcode 26` - > `27` for the `iOS` platform to run everything on `Swift 6.4`).

This said, the worst case scenario is that we will need to build an ugly way to bypass the `Swift Android SDK` when talking to specific servers. Ugly, but not deadly

PR #5535, opened 2026-08-28
Recore the ISO8601DateFormatter type on top of FoundationEssentials instead of CoreFoundation

parkera:
Instead of using CoreFoundation for ISO8601, use the Swift implementation in FoundationEssentials.

### Motivation:

Removes unsafe CF usage, plus it's faster.

### Modifications:

Import FoundationEssentials and use ISO8601FormatStyle.

### Result:

Should be no visible effect on the output.

### Testing:

All the existing tests work as-is.


Sources/Foundation/ISO8601DateFormatter.swift
```diff
@@ -126,31 +121,49 @@ open class ISO8601DateFormatter : Formatter, NSSecureCoding {
...
+            style = style.month()
+        }
+        if formatOptions.contains(.withWeekOfYear) {
+            style = style.weekOfYear()
+        }
+        if formatOptions.contains(.withDay) {
+            style = style.day()
+        }
+        if formatOptions.contains(.withTime) {
+            style = style.time(includingFractionalSeconds: formatOptions.contains(.withFractionalSeconds))
```
jmschonfeld, line 156: Just confirming, `withFractionalSeconds` doesn't imply `withTime` with the current behavior right? (that's the only thing that I can think of that might be different between how the options are parsed here and how they may have been parsed before

jmschonfeld (approved):

Sources/Foundation/ISO8601DateFormatter.swift
parkera: I think we're good here, the test case at https://github.com/swiftlang/swift-corelibs-foundation/blob/main/Tests/Foundation/TestISO8601DateFormatter.swift#L98 verifies it

jmschonfeld (approved):

Issue #5534, opened 2026-08-28
XMLParser traps on invalid UTF-8 in a tag or attribute name


lrmoorejr:
`XMLParser` force-unwraps a failable UTF-8 conversion when building element and attribute
**names**, so a document containing an invalid UTF-8 byte in a name **kills the process**
instead of reporting a parse error. Character data and attribute *values* are handled
correctly, which is what makes this surprising: the same byte is tolerated in one position
and fatal in another.

There is no way for a caller to defend against this — the trap happens inside the SAX
callback, `parse()` never returns, and there is no error to catch. Any program that parses
XML it did not author (a downloaded feed, a `.docx`/`.xlsx` part, a user-supplied file) can
be terminated by four bytes of input.

## Reproducer

```swift
import Foundation
#if canImport(FoundationXML)
import FoundationXML
#endif

final class D: NSObject, XMLParserDelegate {}

let d = D()
let p = XMLParser(data: Data([0x3C, 0x61, 0xFF, 0x3E]))  // `<a\xFF>`
p.delegate = d
print("parsed:", p.parse())   // never reached
```

Result: `Illegal instruction` (SIGILL, exit 132). No message, no `parserError`, no delegate
callback.

`<a r\xFF="1"/>` — an invalid byte in an *attribute* name — does the same.

## Environment

- Swift 6.3.3 (swift-6.3.3-RELEASE), `x86_64-unknown-windows-msvc`
- Not tested on Linux, but the code below is platform-independent, so it very likely
  reproduces there too.

## Cause

`Sources/FoundationXML/XMLParser.swift` converts C strings through:

```swift
private func UTF8STRING(_ bytes: UnsafePointer<UInt8>?) -> String? {
    guard let bytes = bytes else { return nil }
    if let (str, _) = String.decodeCString(bytes, as: UTF8.self,
                                           repairingInvalidCodeUnits: false) {
        return str
    }
    return nil
}
```

`repairingInvalidCodeUnits: false` makes this correctly return `nil` for invalid UTF-8 — and
then the SAX handlers force-unwrap it. In `_NSXMLParserStartElementNs`:

```swift
let elementName: String = UTF8STRING(localname)!
...
let attrLocalNameString = UTF8STRING(attrLocalName)!
```

The same `UTF8STRING(...)!` pattern appears in at least nine handlers:
`_NSXMLParserGetEntity`, `_NSXMLParserNotationDecl`, `_NSXMLParserAttributeDecl`,
[...]

PR #5536, opened 2026-08-31
Do not terminate the process when an XML name is not valid UTF-8

nanbu:
### Motivation:

On non-Darwin platforms `XMLParser` terminates the process — `Illegal instruction`, not a parse error
— when a document contains an element, attribute, entity or processing-instruction **name** whose
bytes are not valid UTF-8. Darwin returns a parse failure for the same input.

A delegate is required to reach it: `parseData(_:lastChunkOfData:)` installs the SAX handler only when
one is set, and the force-unwrap is inside those callbacks.

```swift
let handler: _CFXMLInterfaceSAXHandler? = (delegate != nil ? _handler : nil)
```

Every program that parses untrusted XML has a delegate, so every one of them can be killed by a
byte-flipped file.

**Reproduction**

```swift
import Foundation
#if canImport(FoundationXML)
import FoundationXML
#endif

final class Recorder: NSObject, XMLParserDelegate {}

var bytes = Array("<?xml version=\"1.0\" encoding=\"UTF-8\"?><doc><De".utf8)
bytes += [0xFF, 0xFF, 0xFF, 0xFF]      // inside the element name
bytes += Array("t/></doc>".utf8)

let parser = XMLParser(data: Data(bytes))
let recorder = Recorder()
parser.delegate = recorder             // required: no delegate, no SAX handler, no crash
withExtendedLifetime(recorder) {
    print("parse()     ->", parser.parse())
    print("parserError ->", parser.parserError as Any)
}
```

Observed on Linux (`swift:6.2-noble`, Ubuntu 24.04.4, x86_64):

```
*** Program crashed: Illegal instruction at 0x... ***
0  specialized _NSXMLParserStartElementNs(...) + 2838 in libFoundationXML.so
1  @objc _NSXMLParserStartElementNs(...) in libFoundationXML.so
2  <unknown> in libxml2.so.2.9.14
5  _CFXMLInterfaceParseChunk + 12 in libFoundationXML.so
7  XMLParser.parseData(_:lastChunkOfData:) + 140 in libFoundationXML.so
8  XMLParser.parse() + 302 in libFoundationXML.so
Illegal instruction (core dumped)          # exit 132
```

Observed on Linux with the delegate removed — no SAX handler is installed, so nothing traps, and the
parse fails the way it should:

```
parse()     -> false
parserError -> NSXMLParserErrorDomain code 9
```

Observed on Darwin (macOS 26, Swift 6.3.3), with and without a delegate:

```
parse()     -> false
parserError -> NSXMLParserErrorDomain code 111
```

The codes differ between platforms; the point is only that Darwin returns rather than dies.

**Cause**

`parseData(_:lastChunkOfData:)` asks libxml2 to recover from errors:

```swift
// Prepare options (substitute entities, recover on errors)
var options = _kCFXMLInterfaceRecover | _kCFXMLInterfaceNoEnt
```

So after libxml2 raises the fatal encoding error it carries on and still calls `startElementNs`, with
the raw bytes of the name. `UTF8STRING` decodes with `repairingInvalidCodeUnits: false` and returns
`nil` for those bytes, and the name call sites force-unwrap it. There are eleven direct ones — entity
name, notation name, element and attribute declarations, unparsed entity, internal subset, attribute
local name, start element, end element, PI target, comment — plus one more a line after a namespace
name is decoded:

```swift
namespaceNameString = UTF8STRING(ns)
asAttrNamespaceNameString = "xmlns:" + namespaceNameString!
```

The optional call sites in the same functions — prefix, publicId, systemId, namespace URI — already
accept `nil`, because for them `nil` means "absent".

### Modifications:

`UTF8STRING` decodes with repair, which is what the attribute **value** a few lines above already does
(`String(decoding: buffer, as: UTF8.self)`):

```diff
-    if let (str, _) = String.decodeCString(bytes, as: UTF8.self,
-                                           repairingInvalidCodeUnits: false) {
+    if let (str, _) = String.decodeCString(bytes, as: UTF8.self,
+                                           repairingInvalidCodeUnits: true) {
         return str
     }
```

A comment records why. A regression test is added to `TestXMLParser`.

The stricter alternative — keep the strict decode and abort the parse at each site — is a much larger
diff, needs an error code chosen for a condition libxml2 has already reported, and would have to be
repeated in twelve places. I am happy to write that instead if it is preferred.

### Result:

`UTF8STRING` returns `nil` only when the pointer itself is `nil` — the meaning every optional caller
tests for — so those force-unwraps can no longer fail because of malformed UTF-8 while libxml2 hands
back a non-null pointer.

The document does not start looking valid: the parse still fails with the error libxml2 raised. What
changes is that the caller gets that failure back instead of losing the process. This matches Darwin in
the outcome that is externally observable — malformed input produces a parse failure rather than
process termination — not necessarily in error codes or in which callbacks fire.

One consequence worth stating: while libxml2 is recovering, a delegate may now be handed a name that
contains U+FFFD where the undecodable bytes were, just before the parse fails. Today it is handed
nothing, because the process is gone.

### Testing:

`test_malformedUTF8InNamesDoesNotTrap` in `Tests/Foundation/TestXMLParser.swift` parses the document
above with a delegate set and asserts that the parse fails and reports an error. It does not assert the
error code, which is 9 on Linux and 111 on Darwin.

The reproduction above was run on Linux (`swift:6.2-noble`) and on Darwin (macOS 26 / Swift 6.3.3),
with and without a delegate, before and without the change. The post-change Linux values are inferred
from the no-delegate run rather than measured directly: the callback returns `Void` and cannot affect
the result of the parse, so removing the trap leaves the same `false` / non-nil error the no-delegate
run already produces. This project's CI will confirm it.

**How it was found**

Fuzzing the readers of [SwiftSheets](https://github.com/nanbu/SwiftSheets), a spreadsheet library, on
Linux CI: a byte-flipped `[Content_Types].xml` inside an `.xlsx` had put four `0xFF` bytes inside the
element name `Default`, and the test process died rather than reporting a malformed file
([CI run](https://github.com/nanbu/SwiftSheets/actions/runs/33388474946/job/99476371674)).

That library now [checks parts for UTF-8 validity before handing them to
`XMLParser`](https://github.com/nanbu/SwiftSheets/blob/c9317e0bb1fe9f933f7ab3bdc4982ab472b43cd9/Sources/SheetCore/Utils/XML.swift#L123-L144),
which is a reasonable thing for it to do in any case — but it is not something every caller of
`XMLParser` should have to discover by crashing.


nanbu:
Closing in favor of #5537, which fixes the same crash (#5534) and is already under review with the stop-and-report approach.

PR #5537, opened 2026-09-02
Avoid crashing on invalid UTF-8 in XML names

### LINKED ISSUE #5534: XMLParser traps on invalid UTF-8 in a tag or attribute name

EhsanAzish80:
Fixes #5534.

Stop parsing and report `invalidCharacterError` when a required XML name contains invalid UTF-8, instead of force-unwrapping a failed conversion.

Adds regression coverage for malformed UTF-8 in element and attribute names.

Validation:
- `FoundationXML` compiled successfully with the current Swift nightly Linux toolchain.
- The complete test runner reached its final packaging stage but failed due to a nightly build-system file-copy error unrelated to this change.
- The package intentionally does not support native macOS SwiftPM builds (`macOS("99.9")`), so the focused test cannot be run directly on macOS.
- `git diff --check` passes.

### DIFF
--- a/Sources/FoundationXML/XMLParser.swift
+++ b/Sources/FoundationXML/XMLParser.swift
@@ -47,6 +47,17 @@ private func UTF8STRING(_ bytes: UnsafePointer<UInt8>?) -> String? {
     return nil
 }
 
+private func requiredUTF8String(_ bytes: UnsafePointer<UInt8>?, parser: XMLParser) -> String? {
+    guard let string = UTF8STRING(bytes) else {
+        let error = NSError(domain: XMLParser.errorDomain, code: XMLParser.ErrorCode.invalidCharacterError.rawValue)
+        parser._parserError = error
+        _CFXMLInterfaceStopParser(parser._parserContext)
+        parser.delegate?.parser(parser, parseErrorOccurred: error)
+        return nil
+    }
+    return string
+}
+
 internal func _NSXMLParserCurrentParser() -> _CFXMLInterface? {
     if let parser = XMLParser.currentParser() {
         return parser.interface
@@ -154,7 +165,7 @@ internal func _NSXMLParserGetEntity(_ ctx: _CFXMLInterface, name: UnsafePointer<
     }
     if entity == nil {
         if let delegate = parser.delegate {
-            let entityName = UTF8STRING(name)!
+            guard let entityName = requiredUTF8String(name, parser: parser) else { return nil }
             // if the systemID was valid, we would already have the correct entity (since we're loading external dtds) so this callback is a bit of a misnomer
             let result = delegate.parser(parser, resolveExternalEntityName: entityName, systemID: nil)
             if _CFXMLInterfaceHasDocument(context) != 0 {
@@ -175,7 +186,7 @@ internal func _NSXMLParserGetEntity(_ ctx: _CFXMLInterface, name: UnsafePointer<
 internal func _NSXMLParserNotationDecl(_ ctx: _CFXMLInterface, name: UnsafePointer<UInt8>, publicId: UnsafePointer<UInt8>, systemId: UnsafePointer<UInt8>) -> Void {
     let parser = ctx.parser
     if let delegate = parser.delegate {
-        let notationName = UTF8STRING(name)!
+        guard let notationName = requiredUTF8String(name, parser: parser) else { return }
         let publicIDString = UTF8STRING(publicId)
         let systemIDString = UTF8STRING(systemId)
         delegate.parser(parser, foundNotationDeclarationWithName: notationName, publicID: publicIDString, systemID: systemIDString)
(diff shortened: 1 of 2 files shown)


Sources/FoundationXML/XMLParser.swift
```diff
@@ -41,7 +41,7 @@ private func UTF8STRING(_ bytes: UnsafePointer<UInt8>?) -> String? {
         return nil
     }
     if let (str, _) = String.decodeCString(bytes, as: UTF8.self,
-                                           repairingInvalidCodeUnits: false) {
+                                           repairingInvalidCodeUnits: true) {
```
jmschonfeld: Can you elaborate on your decision to repair the invalid code units rather than throwing an error while parsing when invalid code units are detected?

EhsanAzish80: Good point. I initially chose repair because it was the smallest change and matched character-data handling, but invalid UTF-8 in an XML name should make parsing fail. I’ve updated this to report a parser error instead.

jmschonfeld: Can you elaborate on the character-data handling you're referring to here that you were matching?

EhsanAzish80: _NSXMLParserCharacters (and attribute-value decoding), which uses String(decoding:as:)
those are content values, while XML names are syntax and should reject invalid encoding

Issue #5390, opened 2026-01-15
Upgrade libxml2 dependency


marcprux:
The libxml2 dependency we [build against](https://github.com/swiftlang/swift/blob/main/utils/update_checkout/update-checkout-config.json#L743) is [2.11.5](https://github.com/GNOME/libxml2/releases/tag/v2.12.5) from Feb 4, 2024. There have been a number (~25) of releases since, the latest being [2.15.1](https://github.com/GNOME/libxml2/releases/tag/v2.15.1) on Oct 16, 2025. Many of these releases have addressed security flaws. Given that XML parsing often deals with untrusted input, we should make an effort to be on the most recent version we can be (unless someone want to re-write it in Swift 😉)

I tried to bump the dependency in https://github.com/swiftlang/swift/pull/86321, but encountered build errors. And then it was mentioned that there were compatibility concerns (https://github.com/swiftlang/swift-corelibs-foundation/pull/5082) with recent changes.

It would be interesting to know what version Foundation on Darwin platforms is building against, and if there were any issues in keeping it up to date.

finagolfin:
@jmschonfeld, is this something you could look at? I think I have seen the semantic breaks mentioned there in the Termux app on Android, where I build trunk Swift against the latest libxml2 and some tests have been failing.

jmschonfeld:
I personally haven't looked into the semantic breaks mentioned here yet. Updating to a newer libxml version certainly seems beneficial for security reasons, so I think the main blocker would be finding a way to resolve the compatibility fallout of the change.

Since @compnerd you made the original attempt to upgrade and found those problematic changes, perhaps you could share some insight on what problems you encountered or what issues there are with adopting it? It sounds like that's a good place to start to figure out what needs to be investigated

compnerd:
@jmschonfeld the pending draft change on swift-corelibs-foundation has a start. But the XML parsing callbacks have changed and that would require changing what Foundation returns from the XML parser.

itingliu:
The draft #5082 changes the return value of this function, right?
`CFStringRef _Nullable _CFXMLNodeCopyPrefix(_CFXMLNodePtr node);`
It looks like it's already returning a nullable result, so I don't think it's an actual API breaker as the callers already need to handle a null result.

compnerd:
The nullability isn't the issue, the results that you get back from libxml2 are different. This led to test failures for the XML parsing.

marcprux:
[...]

PR #5541, opened 2026-09-05
[6.4] Fix XMLNode name/stringValue/prefix for DTD and namespace nodes (#5481)

finagolfin:
6.4 cherry-pick of fix for newer libxml2 in linux distros like Ubuntu 26.04, as discussed in #5390.

Three bugs in XMLNode.swift:

1. name setter: .entityDeclaration, .attributeDeclaration, and .elementDeclaration all fell through to the default case which calls _CFXMLNodeSetName. libxml2 silently ignores that call for DTD nodes, so renames were lost. Extended the existing .notationDeclaration case (which already used _CFXMLNodeForceSetName) to cover all four DTD declaration kinds.

2. stringValue setter: The default path runs _CFXMLEncodeEntities(document, string) before storing the value. For a standalone XMLDTDNode with no owning document, _CFXMLNodeGetDocument returns nil, _CFXMLEncodeEntities(nil, …) returns nil, and the fallback "" is written — erasing the value. Element declaration content is a DTD validation string (e.g. "(#PCDATA | array)*"), not XML text that needs entity escaping, so it now goes directly to _CFXMLNodeSetContent.

3. prefix getter: _CFXMLNodeCopyPrefix returns nil for namespace nodes because libxml2 stores the namespace prefix in a different slot. Added a .namespace case that reads it via _CFXMLNamespaceCopyPrefix, mirroring the pattern already used by the name getter.

Fixes test_nodeNames (entityDecl/attrDecl/elementDecl rename), test_dtd (elementDecl name + stringValue on standalone node), and test_createElement (namespace.prefix).


parkera (approved):
Approved for 6.4.x

PR #5546, opened 2026-09-08
[6.4] _CFXMLInterface: account for possible `nullptr` return

iCharlesHu:
Backport https://github.com/swiftlang/swift-corelibs-foundation/pull/5082 to 6.4.x

`xmlSplitQName2` may return `nullptr` for the result, which when passed to `CFStringCreateWithCString` would attempt to perform `strlen(nullptr)` which is ill-defined. When updating libxml2 on Windows, we would perform an invalid memory access due to the `strlen` invocation inside `CFStringCreateWithCString`. Protect against this case, returning `NULL` instead.

Resolves: rdar://186569766


parkera (approved):
Approved for 6.4.x

PR #5555, opened 2026-09-14
Use Windows known folders for current-user preferences

sebmarchand:
Use Windows' LocalAppData known folder to persist current-user preferences when the account name differs from the profile directory.

### Motivation:

CoreFoundation constructs the current user's preferences directory from the profiles root and account name. Renaming an account does not rename its profile directory, so this can select an inaccessible or nonexistent location and cause `UserDefaults.synchronize()` to fail. This addresses https://github.com/swiftlang/swift-corelibs-foundation/issues/4997.

Windows preference writes also pass a UTF-8 path to narrow `chmod`, which can fail for Unicode paths even after the preference file is written.

### Modifications:

- Resolve current-user LocalAppData through `SHGetKnownFolderPath`, retaining the existing named-user and any-user behavior.
- Propagate known-folder lookup failure safely.
- Use the existing Unicode-safe `_NS_chmod` helper for Windows file permission updates.

### Result:

Current-user preferences use the directory selected by Windows, and permission updates support Unicode paths. Preference domains and file formats remain compatible.

### Testing:

- Added Windows regression cases for the LocalAppData destination and saving/deleting preferences under a Unicode path.
- The three modified C files compile individually with a Windows Swift SDK, and the added Swift test methods type-check.
- A WinSDK probe using the Windows folder APIs links without explicit Windows library options. The updated SwiftPM manifest evaluates successfully.
- A separate synthetic-profile probe reproduced the existing failure: an account/profile mismatch causes synchronization to fail and values to disappear in a fresh process.

The modified Foundation runtime has not been linked or executed, so the added regression tests still need to run against it. Full Windows runtime validation and non-Windows builds remain pending; this PR is a draft.


Package.swift
```diff
@@ -234,7 +234,12 @@ let package = Package(
                 "CMakeLists.txt"
             ],
             cSettings: coreFoundationBuildSettings,
-            linkerSettings: [.linkedLibrary("log", .when(platforms: [.android]))]
+            linkerSettings: [
+                .linkedLibrary("log", .when(platforms: [.android])),
+                .linkedLibrary("shell32", .when(platforms: [.windows])),
+                .linkedLibrary("ole32", .when(platforms: [.windows])),
+                .linkedLibrary("uuid", .when(platforms: [.windows]))
+            ]
```
compnerd, line 242: These shouldn't be needed I think? The use of modules should do the autolinking.

Sources/CoreFoundation/CFKnownLocations.c
```diff
@@ -94,9 +95,19 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
...
             break;
         }
-        case _kCFKnownLocationUserCurrent:
-            username = CFGetUserName();
-            // fallthrough
+        case _kCFKnownLocationUserCurrent: {
+            PWSTR path = NULL;
+            HRESULT result = SHGetKnownFolderPath(&FOLDERID_LocalAppData, 0, NULL, &path);
+            if (SUCCEEDED(result) && path) {
+                CFStringRef pathRef = CFStringCreateWithCharacters(kCFAllocatorSystemDefault, (const UniChar *)path, wcslen(path));
```
compnerd, line 102: Might be worth doing `CFStringCreateWithCharactersNoCopy`

Sources/CoreFoundation/CFURLAccess.c
```diff
@@ -246,7 +246,11 @@ static Boolean _CFFileURLWritePropertiesToResource(CFURLRef url, CFDictionaryRef
                 const MODE_TYPE *modePtr = (const MODE_TYPE *)CFDataGetBytePtr((CFDataRef)value);
                 mode = *modePtr;
             }
+#if TARGET_OS_WIN32
+            err = _NS_chmod(cPath, mode);
```
compnerd, line 250: I wonder if using `_wchmod` is clearer even if we need to jump between encodings.

compnerd (commented):
This is a good idea, we should prefer using the system folders over creating the path by hand.

Sources/CoreFoundation/CFKnownLocations.c
```diff
@@ -94,9 +95,19 @@ CFURLRef _Nullable _CFKnownLocationCreatePreferencesURLForUser(CFKnownLocationUs
             CFRelease(allUsersPath);
             break;
         }
-        case _kCFKnownLocationUserCurrent:
-            username = CFGetUserName();
-            // fallthrough
+        case _kCFKnownLocationUserCurrent: {
+            PWSTR path = NULL;
+            HRESULT result = SHGetKnownFolderPath(&FOLDERID_LocalAppData, 0, NULL, &path);
```
jmschonfeld, line 100: It's a bit unfortunate we can't easy call up to `FileManager` here because it already implements this for the `.applicationSupportDirectory` with the user domain mask, but calling up to FileManager is probably more trouble than it's worth here right now

Sources/CoreFoundation/CMakeLists.txt
```diff
@@ -120,6 +120,10 @@ target_link_libraries(CoreFoundation
         _FoundationICU
         dispatch)
 
+if(WIN32)
+    target_link_libraries(CoreFoundation PRIVATE shell32 ole32 uuid)
```
jmschonfeld: Can you elaborate on where these libraries come from / why they're required? I'm guessing that `shell32` is required for the declarations in `shlobj.h` but `uuid` seems odd given none of the code here references UUIDs

Tests/Foundation/TestUserDefaults.swift
```diff
@@ -7,8 +7,51 @@
...
 class TestUserDefaults : XCTestCase {
 
+#if os(Windows)
+    func test_currentUserPreferencesUseLocalAppData() throws {
+        var folderID = FOLDERID_LocalAppData
+        var path: UnsafeMutablePointer<WCHAR>?
+        let result = SHGetKnownFolderPath(&folderID, 0, nil, &path)
+        defer { CoTaskMemFree(path) }
+        XCTAssertGreaterThanOrEqual(result, 0)
+        let directory = URL(fileURLWithPath: String(decodingCString: try XCTUnwrap(path), as: UTF16.self), isDirectory: true)
```
jmschonfeld: It might be easier to just use `FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)[0]` here

compnerd:

Sources/CoreFoundation/CFURLAccess.c
sebmarchand: Switched this to the existing `_NS_chmod` helper, which converts the UTF-8 path to UTF-16 and calls `_wchmod`.

Sources/CoreFoundation/CMakeLists.txt
sebmarchand: Removed uuid. shell32 provides SHGetKnownFolderPath, and ole32 provides CoTaskMemFree. Added a comment explaining both dependencies.

Tests/Foundation/TestUserDefaults.swift
sebmarchand: Done. The test gets the expected directory from `FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)` and unwraps `.first` with `XCTUnwrap`

compnerd:

sebmarchand:
@compnerd all tests are green!

compnerd (approved):

